PR #19396, opened 2022-11-14
convert container-orchestrator to micronaut

colesnodgrass:
## What
- fix #13271

## How
- remove unused dependencies
  - apache-commons
  - jetty
  - config-persistence
  - db-lib
  - protocol-models
  - job-persistence
- add additional `deserialize` methods to the `Jsons` helper class
  - exposes more methods of the underlying jackson object-mapper
    - primarily for avoiding _unchecked casts_ issues
- rename `ContainerOrchestratorApp` to `Application`
  - remove busy-wait on `SUCCESS_FILE_NAME`
    - replace with k8s initContainer (#19088)
  - move logging initialization to EventListener classes
  - move dependency creation to Factory classes
- remove `WorkerHeartbeatServer` with its `HttpServlet`
  - replace with Micronaut `HeartbeatController` 
- combine `AsyncStateManager` interface with `DefaultAsyncStateManager` class into `AsyncStateManager` class 
- move `JobOrchestrator` classes to a new package
  - replace file reading methods of `JobOrchestrator` interface with injected dependencies to the implementations 
- add `application.yml`
- add `micronaut-banner.txt`
- add unit-tests
  - most of existing code did not contain any unit-tests as the code wasn't easily testable, some improvements have been made here and a lot of the new code now has tests   
- general code clean up
  - change visibility of methods to better match their expected usage (mostly moving them from the public scope)
  - remove unnecessary lambda usage (mostly moving to method references)
  - `final _type_` -> `final var`

## Oddities
- the `Application` class does a little too much currently and should probably be split into two different classes
- the `EventListeners.setEnvVars` method should probably be replaced with the env-vars being set on the container itself (in the worker when it creates this pod and containers)
- the `@Secured` annotation on the `HeartbeatController` could (should?) be replaced with something else

PR #19501, opened 2022-11-16
Fix MDC scope

benmoriceau:
## What
Address: https://github.com/airbytehq/oncall/issues/1066

We were using the Default MDC scope which made all the logs to be tagged with the wrong prefix. This is fixing it.


colesnodgrass:
As some of these changes are going to cause issues with #19396, can I instead incorporate the container-orchestrator changes into that PR?

benmoriceau:
@colesnodgrass I didn't have time to look at your PR yet. I Think that we should merge this first since the code freeze is approaching and I don't know how long the migration to micronaut PR will take.

colesnodgrass (approved):

gosusnp (approved):

Issue #18582, opened 2022-10-28
Fix how we store null json objects in the database


gosusnp:
## What

We need to manipulate and store optional JSON blobs in our database. For example, looking at the `connector_operation` table, we have `operator_normalization`, `operator_dbt` and `operator_webhook`.

Our storage is currently error prone, because the columns that are storing JSON blob configs are typed as `optional` `JSONB` which is consistent w.r.t the definition. However, in practice, a null config will be stored as a `"null"` (without the quotes but to emphasize that we have a "null" string) rather than an actual `NULL` in SQL. The other problem is that most of our code is not meant to handle `NULL` values and will crash if we have a null config.

This lead to an issue when we introduced the `operator_webhook` column.
* The schema was declared as `optional` `JSONB` (as expected)
* The tests were fine because when we add new data in the tests, data is saved as `"null"` strings.
* When running on an instanced that has been migrated, any data read would crash on a null pointer exception, coming from older data. It has been caught and fixed by a manual test before we went to production.

Our data model should be more resilient to those issues.

## How

Suggestion: handle null configs as null rather than the serialize as string that seems to be coming from our default `Jsons.serialialize`.

PR #19505, opened 2022-11-16
add null check when reading resource_requirements from the db

cgardens:
## What
* @colesnodgrass and I were talking about the new APM dashboard that @jdpgrailsdev built and noticed this very frequent NPE ([link](https://app.datadoghq.com/apm/trace/2177460164262014560?colorBy=service&env=dev&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=5729246165658222820&spanViewType=errors&timeHint=1668633434233.255))
* Looks like we are assuming a field is non-null in the db when actually it can be null.
* Adds a null check. @gosusnp any idea of this is a regression? Does this break all connections where we don't explicitly set a resource requirement? This seems like it might be stopping the show, but I might be missing some handling.

### DIFF
--- a/airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/DbConverter.java
+++ b/airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/DbConverter.java
@@ -77,7 +77,8 @@ public class DbConverter {
                 : Jsons.deserialize(record.get(CONNECTION.SCHEDULE_DATA).data(), ScheduleData.class))
         .withOperationIds(connectionOperationId)
         .withResourceRequirements(
-            Jsons.deserialize(record.get(CONNECTION.RESOURCE_REQUIREMENTS).data(), ResourceRequirements.class))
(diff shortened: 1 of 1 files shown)


gosusnp (approved):
I don't know if it is a "new" regression, we already had an issue in the past for something similar.
https://github.com/airbytehq/airbyte/issues/18582

Last time I ran into this issue, the bad data came from a SQL migration that would default to NULL.

cgardens:
Interesting. I wonder if this NPE is misleading then.

cgardens:
```
select count(*) from connection;
-- 10786
select count(*) from connection where resource_requirements = '{}';
-- 551
select count(*) from connection where resource_requirements = 'null';
-- 10125
select count(*) from connection where resource_requirements IS NULL;
-- 0
```

yeah. issue isn't that the it is a sql null but something else.

jdpgrailsdev (approved):
:shipit:  APM FTW!

colesnodgrass (approved):

cgardens:
likely not the root cause.

PR #12400, opened 2022-04-27
🎉 JDBC source: adjust streaming query fetch size dynamically

tuliren:
## What
- This PR resolves https://github.com/airbytehq/airbyte/issues/12192 and https://github.com/airbytehq/airbyte/issues/4314.

## How
- Firstly, set the fetch size to 10. Fetch the 10 rows, measure the mean serialized size of those rows, and use that size to estimate the best fetch size (N).
- Secondly, set the fetch size to N. Sample the serialized size for every 100 rows to adjust the fetch size.

## Recommended reading order
1. `InitialSizeEstimator.java`
2. `SamplingSizeEstimator.java`
3. `TwoStageSizeEstimator.java`
4. `AdaptiveStreamingQueryConfig.java`
5. `StreamingJdbcDatabase.java`

## 🚨 User Impact 🚨
- This PR fixes the issue that JDBC source connector will fail to fetch database tables with extremely large rows.

PR #19514, opened 2022-11-16
Increase Database Source SELECT Batch Size

evantahler:
https://github.com/airbytehq/airbyte/pull/12400 added dynamic batch sizes to our SQL sources to prevent OOM problems.  However, I believe that we are currently creating batch sizes that are too small for _optimal speed_.  This is validated in 2 ways:
* We have shown that Airbyte's [speed increases with wider rows](https://airbytehq-team.slack.com/archives/C048E3BK1KM/p1668639622311289)
* We are not seeing OOM errors on our database sources any more, and we are not utilizing more than 60% of available memory in our source containers

unnamed

This PR does 2 things:
1. Bumps the total allocation of the source's available memory for the JDBC buffer from 50% to 60%
2. Rather than assuming the worst-case UTF8 byte-size for RECORD strings (4-byte/char), we move to 3-byte/char which is likely more common - assuming most chars fit in the ASCII space.


evantahler:
@edgao I'm network-bound at home testing this out... what's the process for getting a connector tested on cloud or a dev server?

airbyte-db/db-lib/src/main/java/io/airbyte/db/jdbc/streaming/FetchSizeConstants.java
```diff
@@ -10,7 +10,7 @@ public final class FetchSizeConstants {
   // This size is not enforced. It is only used to calculate a proper
   // fetch size. The max row size the connector can handle is actually
   // limited by the heap size.
-  public static final double TARGET_BUFFER_SIZE_RATIO = 0.5;
+  public static final double TARGET_BUFFER_SIZE_RATIO = 0.6;
```
davinchia, line 13: curious, why not go higher?

if primary memory usage is due to this buffer, I think we can go as high as 80%. Usual java overhead is ~~10% so a 20% buffer seems good enough.

evantahler: I think we can keep ramping this up... but in an environment where I can't robustly test this, I want to go step-by-step

evantahler:
@davinchia / @edgao I'm struggling to figure out how to test this.  Any ideas?

evantahler:
I've done some local testing and thing seem ok... @edgao / @davinchia please let me know what I should do to be confident that this is safe to merge in

davinchia:
@evantahler I might regret saying this - I'm pretty confident this is safe to merge in. We might see some OOM if not. We give all jobs up to 2GB now. This is plenty of airspace, which makes me think the chance is low.

evantahler:
@davinchia i feel more or less the same.  @edgao how do you feel about YOLOing this one (after code freeze, of course)?

edgao:
Seems fine to me. Vast majority of source-postgres cloud syncs are way below 2G, with just a handful of outliers

image

fyi @subodh1810 as random db sources representative

edgao (approved):
:shipit:

PR #19528, opened 2022-11-17
add some tests for manifests with references and some additional invalid test cases and error handling refinement

brianjlai:
## What
We need to use the latest version of `airbyte-cdk` to correctly handle dereferencing the incoming YAML from the connector builder UI. The logic to properly resolve references is all handled in https://github.com/airbytehq/airbyte/pull/19517 . We just need to have the builder server to use the latest version.

## How
This will be a patch or minor bump so technically this should work automatically, but I also wanted to take this time to fix and add a few tests for coverage. This should be merged after #19517


brianjlai:
retested now that the latest version of `airbyte-cdk` (0.9.4) has been deployed and the new tests that would fail w/ the old version are passing now

airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_low_code_cdk_adapter.py
```diff
@@ -94,6 +96,69 @@ def parse_response(
     ],
     "check": {"stream_names": ["hashiras"], "class_name": "airbyte_cdk.sources.declarative.checks.check_stream.CheckStream"},
 }
+
+MANIFEST_WITH_REFERENCES = {
```
girarda, line 100: me gusta

airbyte-connector-builder-server/connector_builder/impl/default_api.py
```diff
@@ -199,6 +198,6 @@ def _create_response_from_log_message(self, log_message: AirbyteLogMessage) -> O
     def _create_low_code_adapter(manifest: Dict[str, Any]) -> LowCodeSourceAdapter:
         try:
             return LowCodeSourceAdapter(manifest=manifest)
-        except ValidationError as error:
+        except Exception as error:
             # TODO: We're temporarily using FastAPI's default exception model. Ideally we should use exceptions defined in the OpenAPI spec
-            raise HTTPException(status_code=400, detail=f"Invalid connector manifest with error: {error.message}")
+            raise HTTPException(status_code=400, detail=f"Invalid connector manifest with error: {error.args[0]}")
```
girarda, line 203: are the any advantages to using `error.args[0]` instead of `str(error)`? The later seems safer since args could be an empty array

girarda (approved):

airbyte-connector-builder-server/connector_builder/impl/default_api.py
brianjlai: probably not and agree, str is safer, fixed!

PR #19538, opened 2022-11-17
Fix format

suhomud:
## What
Fix format


etsybaev (approved):
Approved assuming there are only code-style changes

PR #19539, opened 2022-11-17
🪟🎉 Remove the onboarding page

timroes:
## What

Remove the onboarding page from the product. We've run some extensive experiment and had the onboarding page disabled for 50% of all users in the past month(s). We did not see that the onboarding page had any impact on users actually succeeding to setup their connection (or other metrics we track). Thus we decided to remove it from the product and instead focus on improving onboarding flows in the future via other parts of the app (e.g. the already improved empty Connection page). This change was signed off by Natalie and Andy J.

## How?

I also ran `ts-prune` and compared the before state to after to make sure there are no new orphaned components that are no longer needed.


joeykmh (approved):
Nice! One fewer set of routes to refactor 😆 tested locally with a new account and set up a new connection - all worked as expected.

PR #19564, opened 2022-11-17
🪟🔧 Add isCorporate flag to analytics identify

timroes:
## What

Adds a new `isCorporate` flag to the user traits we send to analytics, so it's easier to filter on that in different analytic systems.


joeykmh (approved):
LGTM ✨ did not test locally

PR #19577, opened 2022-11-18
Fix bumpversion for airbyte-connector-builder-server

lmossman:
## What
The bumpversion.cfg was incorrectly configured for the connector builder server, because the module was renamed at one point and bumpversion.cfg was not updated accordingly.

## How
This PR fixes the bumpversion.cfg to point to the right directory, removes a pointer to a non-existent file, and fixes the versions in the airbyte-connector-builder-server to match the platform version


lmossman:
Merging without review because this change shouldn't break any modules, as it is just changing bumpversion.cfg and modifying the version of the connector builder server files, and it is currently blocking OSS release which is needed to fix the current state of OSS master (see [slack thread](https://airbytehq-team.slack.com/archives/C02U9R3AF37/p1668752600151109) for context)

PR #19586, opened 2022-11-18
🎉Source Monday: migration to low-code

roman-yermilov-gl:
## What
*Migrate to low-code*


airbyte-integrations/connectors/source-monday/acceptance-test-config.yml
```diff
@@ -1,34 +1,51 @@
 # See [Source Acceptance Tests](https://docs.airbyte.com/connector-development/testing-connectors/source-acceptance-tests-reference)
 # for more information about how to configure these tests
 connector_image: airbyte/source-monday:dev
-tests:
+acceptance_tests:
```
davydov-d, line 4:
```suggestion
test_strictness_level: high
acceptance_tests:
```

airbyte-integrations/connectors/source-monday/acceptance-test-config.yml
```diff
@@ -1,34 +1,51 @@
...
-      empty_streams: ["updates"]
-    - config_path: "secrets/config_api_token.json"
-      configured_catalog_path: "integration_tests/configured_catalog.json"
-      empty_streams: ["updates"]
-    - config_path: "secrets/config_oauth.json"
-      configured_catalog_path: "integration_tests/configured_catalog.json"
-      empty_streams: ["updates"]
+    tests:
+      - config_path: "secrets/config_api_token.json"
+        configured_catalog_path: "integration_tests/configured_catalog.json"
```
davydov-d, line 29: if I'm not mistaken, you should omit this for high strictness level

airbyte-integrations/connectors/source-monday/acceptance-test-config.yml
```diff
@@ -1,34 +1,51 @@
...
+          extra_records: yes
+        empty_streams:
+          - name: updates
+            bypass_reason: "unable to populate"
   full_refresh:
-    - config_path: "secrets/config.json"
-      configured_catalog_path: "integration_tests/configured_catalog.json"
+    tests:
+      - config_path: "secrets/config_api_token.json"
+        configured_catalog_path: "integration_tests/configured_catalog.json"
```
davydov-d, line 51: what about incremental tests?

airbyte-integrations/connectors/source-monday/source_monday/monday.yaml
```diff
@@ -0,0 +1,90 @@
...
+    request_options_provider:
+      class_name: "source_monday.GraphQLRequestOptionsProvider"
+      limit: "{{ options['items_per_page'] }}"
+    error_handler:
+      response_filters:
+        - predicate: "{{ 'error_code' in response and response['error_code'] == 'ComplexityException' }}"
+          action: RETRY
+      backoff_strategies:
+        - type: ExponentialBackoffStrategy
+          factor: 30
```
davydov-d, line 29: why so much?

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
```diff
@@ -0,0 +1,48 @@
...
+from dataclasses import dataclass
+from typing import Optional, Mapping, Any, Union, MutableMapping
+
+from airbyte_cdk.sources.declarative.interpolation import InterpolatedString
+from airbyte_cdk.sources.declarative.requesters.request_options import InterpolatedRequestOptionsProvider
+from airbyte_cdk.sources.declarative.types import StreamState, StreamSlice
+
+
+@dataclass
+class CustomOptionsProviderMixin:
```
davydov-d, line 12: why do you prefer implementing it as a separate class?

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
```diff
@@ -0,0 +1,48 @@
...
+            return schema_dict["properties"]
+
+    def get_fields(self, x: dict):
+        query = ','.join([f'{col}' + (f'{{{self.get_fields(x[col]["properties"])}}}' if 'properties' in x[col] else '') for col in x])
+        return query
+
+    def get_request_params(self, *, stream_state: Optional[StreamState] = None, stream_slice: Optional[StreamSlice] = None,
+                           next_page_token: Optional[Mapping[str, Any]] = None) -> MutableMapping[str, Any]:
+        graphql_params = {}
+        limit = self.limit.eval(self.config)
```
davydov-d, line 39: can't it raise an error if it's None?

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
```diff
@@ -0,0 +1,48 @@
...
+    NEXT_PAGE_TOKEN_FIELD_NAME = "next_page_token"
+
+    def __post_init__(self, options: Mapping[str, Any]):
+        super(GraphQLRequestOptionsProvider, self).__post_init__(options)
+
+        self.limit = InterpolatedString.create(self.limit, options=options)
+        self.name = options.get('name', '').lower()
+
+    def get_schema_root_properties(self):
+        schema_path = os.path.join(os.path.abspath(os.curdir), "source_monday", f"schemas/{self.name}.json")
```
davydov-d, line 27: we do hardcode the path but do not hardcode the name, that's very weird

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
```diff
@@ -0,0 +1,48 @@
...
+    def get_request_params(self, *, stream_state: Optional[StreamState] = None, stream_slice: Optional[StreamSlice] = None,
+                           next_page_token: Optional[Mapping[str, Any]] = None) -> MutableMapping[str, Any]:
+        graphql_params = {}
+        limit = self.limit.eval(self.config)
+        if limit:
+            graphql_params["limit"] = limit
+        if next_page_token:
+            graphql_params.update({"page": next_page_token[self.NEXT_PAGE_TOKEN_FIELD_NAME]})
+
+        graphql_query = ",".join([f"{k}:{v}" for k, v in graphql_params.items()])
```
davydov-d, line 45: if I do understand things right, `graphql_params` consists of max 2 items. For the sake of readability I'd ask to fill `graphql_query` it in the `if` statements above

airbyte-integrations/connectors/source-monday/unit_tests/test_graphql_request_options_provider.py
```diff
@@ -0,0 +1,29 @@
...
+        "key_b": {"type": ["null", "string"]},
+        "key_c": {"type": ["null", "string"]},
+    }
+    mocker.patch.object(GraphQLRequestOptionsProvider, "get_schema_root_properties", return_value=schema)
+    provider = GraphQLRequestOptionsProvider(
+        limit="{{ options['items_per_page'] }}",
+        options={"name": "items", "items_per_page": 100}
+    )
+    assert provider.get_request_params(
+        stream_state=mocker.MagicMock(),
```
davydov-d, line 26: could you replace these mocks with empty lists or dicts for instance? mocks are very tolerant to types/signature errors

airbyte-integrations/connectors/source-monday/unit_tests/test_graphql_request_options_provider.py
```diff
@@ -0,0 +1,29 @@
+from source_monday import GraphQLRequestOptionsProvider
+
+
+def test_get_request_params(mocker):
+    schema = {
```
davydov-d, line 5: I suggest you move it to parametrize(), it would enable us extend the test in the future

airbyte-integrations/connectors/source-monday/source_monday/pagination_workaround.py
```diff
@@ -0,0 +1,55 @@
...
+    page_size: Optional[Union[InterpolatedString, str]]
+    options: InitVar[Mapping[str, Any]]
+
+    def __post_init__(self, options: Mapping[str, Any]):
+        self._page = 1
+        self._page_size = None
+
+        self.page_size = InterpolatedString.create(self.page_size, options=options)
+
+    def eval_page_size(self, possible_record_len: int):
```
davydov-d, line 31: why do we need this method to be public?

airbyte-integrations/connectors/source-monday/unit_tests/test_pagination_workaround.py
```diff
@@ -0,0 +1,108 @@
...
+    assert getattr(page_increment, "_page_size") is None
+    assert page_increment.page_size.eval(page_increment.config) == ""
+
+    page_increment.next_page_token(
+        response=mocker.MagicMock(),
+        last_records=[{}, {}, {}, {}]
+    )
+
+    assert getattr(page_increment, "_page") == 2
+    assert getattr(page_increment, "_page_size") == 4
```
davydov-d, line 89: what's the reason you're asserting private attributes instead of the `next_page_token` return value?

davydov-d (changes requested):

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
roman-yermilov-gl: Is there other way of implementing custom logic? Could you please clarify your question or resolve it?

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
```diff
@@ -0,0 +1,48 @@
...
+        self.name = options.get('name', '').lower()
+
+    def get_schema_root_properties(self):
+        schema_path = os.path.join(os.path.abspath(os.curdir), "source_monday", f"schemas/{self.name}.json")
+        with open(schema_path) as f:
+            schema_dict = json.load(f)
+            return schema_dict["properties"]
+
+    def get_fields(self, x: dict):
+        query = ','.join([f'{col}' + (f'{{{self.get_fields(x[col]["properties"])}}}' if 'properties' in x[col] else '') for col in x])
```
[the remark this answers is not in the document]
roman-yermilov-gl: I was trying to do it but the structure is not complex. The cause of poor readability is in curly braces `{`, so splitting will not be helpful

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
roman-yermilov-gl: What do you mean to be None? Could you please describe any scenario of how `self.limit` to be `None`?

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
roman-yermilov-gl: Do you mean something like that?
```
graphql_query = ""
if limit:
    graphql_query = f"limit:{limit}"
if next_page_token:
    if graphql_query:
        graphql_query += ","
    graphql_query += f"page:{next_page_token[self.NEXT_PAGE_TOKEN_FIELD_NAME]}"
```

airbyte-integrations/connectors/source-monday/unit_tests/test_graphql_request_options_provider.py
roman-yermilov-gl: There is no any reason to discuss about `stream_state` and `stream_slice`:
- I do not see how it can affect the test
- fields are not a part of the test
- `get_request_params` does not use it at all

airbyte-integrations/connectors/source-monday/unit_tests/test_pagination_workaround.py
roman-yermilov-gl: `next_page_token` mutates protected variable so it is convenient for me to check its value after each call. In some cases I need to check it before the `next_page_token` method to be called. If to dive into logic of the test, then you will see that this way is visually better explains what is going on

airbyte-integrations/connectors/source-monday/acceptance-test-config.yml
roman-yermilov-gl: Done

airbyte-integrations/connectors/source-monday/source_monday/monday.yaml
roman-yermilov-gl: Because of rate limits. I need ask again after 1 minute

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
davydov-d: I have probably took the name of the stream for the name of the source, nevermind

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
davydov-d: Sorry, I haven't noticed you're evaluating it in `__post_init__`

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
```diff
@@ -0,0 +1,53 @@
...
+        self.name = options.get('name', '').lower()
+
+    def _get_schema_root_properties(self):
+        schema_path = os.path.join(os.path.abspath(os.curdir), "source_monday", f"schemas/{self.name}.json")
+        with open(schema_path) as f:
+            schema_dict = json.load(f)
+            return schema_dict["properties"]
+
+    def _get_query_by_stream_properties(self, props: dict):
+        if self.name == "teams" and self.config.get("debug"):
```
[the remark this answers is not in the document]
roman-yermilov-gl: Rate limits force us to make custom query when doing testing. One query of Teams stream costs more then all point limits we have

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
roman-yermilov-gl: Right, I did't understand to what class was a question. I thought it was for `GraphQLRequestOptionsProvider`. It was a part of some experiments with inheritance and, so it should be cleared.

airbyte-integrations/connectors/source-monday/unit_tests/test_pagination_workaround.py
roman-yermilov-gl: It is tested, we have special test for it. For this test the method is only a driver for some internal logic.
"*but it's an implementation detail, not the exposed API*" - this is exactly what I am testing

airbyte-integrations/connectors/source-monday/unit_tests/test_graphql_request_options_provider.py
roman-yermilov-gl: After some thinking I decided that maybe MagicMock is too much for the args and we don't need most of functionality it proposes. So we can switch to Mock or just basic types as you suggested by the same reasons from comment above

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
davydov-d: The question was about `CustomOptionsProviderMixin`. As for me, the `limit` field could be better defined in `GraphQLRequestOptionsProvider` itself

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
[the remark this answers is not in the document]
davydov-d: can we move this logic to another property to make it testable?

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
```diff
@@ -0,0 +1,50 @@
...
+        self.limit = InterpolatedString.create(self.limit, options=options)
+        self.name = options.get('name', '').lower()
+
+    def _get_schema_root_properties(self):
+        schema_path = os.path.join(os.path.abspath(os.curdir), "source_monday", f"schemas/{self.name}.json")
+        with open(schema_path) as f:
+            schema_dict = json.load(f)
+            return schema_dict["properties"]
+
+    def _get_query_by_stream_properties(self, props: dict):
```
davydov-d, line 29: please add some docstrings with example(s) here

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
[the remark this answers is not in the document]
roman-yermilov-gl: So we made this terrifying injection for internal use just for test to pass successfully, but you ask me to make it a testable feature and to write a special test for it? First of all it sounds crazy. But I would say even more: I made a unittest that take it into account. Please take a look at unittests one more time

airbyte-integrations/connectors/source-monday/acceptance-test-config.yml
roman-yermilov-gl: Done

airbyte-integrations/connectors/source-monday/source_monday/monday.yaml
```diff
@@ -0,0 +1,98 @@
...
+streams:
+  - "*ref(definitions.items_stream)"
+  - "*ref(definitions.boards_stream)"
+  - "*ref(definitions.teams_stream)"
+  - "*ref(definitions.updates_stream)"
+  - "*ref(definitions.users_stream)"
+
+check:
+  stream_names:
+    - "users"
```
davydov-d, line 98:
```suggestion
    - "users_stream"
```

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        fields = []
+        for field, nested_schema in field_schema.items():
+            nested_fields = nested_schema.get(
+                "properties",
+                nested_schema.get("items", {}).get("properties")
+            )
+            if nested_fields:
+                # preconfigured_arguments = get properties from schema or any other source ...
+                # fields.append(self._build_query(field, nested_fields, **preconfigured_arguments))
```
davydov-d, line 56: looks like a debug code that probably should be removed

davydov-d (approved):

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
roman-yermilov-gl: This is not debug code. This is how I want to extend it but can`t for now

airbyte-integrations/connectors/source-monday/source_monday/monday.yaml
roman-yermilov-gl: This is not correct suggestion which makes impossible to pass the check. I will not be applying it

airbyte-integrations/connectors/source-monday/source_monday/dpath_string_extractor.py
```diff
@@ -0,0 +1,31 @@
...
+from airbyte_cdk.sources.declarative.decoders.decoder import Decoder
+from airbyte_cdk.sources.declarative.decoders.json_decoder import JsonDecoder
+from airbyte_cdk.sources.declarative.extractors.record_extractor import RecordExtractor
+from airbyte_cdk.sources.declarative.interpolation.interpolated_string import InterpolatedString
+from airbyte_cdk.sources.declarative.types import Config, Record
+from dataclasses_jsonschema import JsonSchemaMixin
+
+
+@dataclass
+class DpathStringExtractor(RecordExtractor, JsonSchemaMixin):
```
girarda, line 19: @roman-yermilov-gl can you add a comment explaining how this behaves differently from the default `DPathExtractor`?

airbyte-integrations/connectors/source-monday/source_monday/pagination_workaround.py
```diff
@@ -0,0 +1,54 @@
...
+
+import requests
+from airbyte_cdk.sources.declarative.interpolation import InterpolatedString
+from airbyte_cdk.sources.declarative.requesters.paginators.strategies.pagination_strategy import PaginationStrategy
+from airbyte_cdk.sources.declarative.types import Config
+from dataclasses_jsonschema import JsonSchemaMixin
+
+
+@dataclass
+class PageIncrementWorkaround(PaginationStrategy, JsonSchemaMixin):
```
girarda, line 12: why do we need this workaround?

airbyte-integrations/connectors/source-monday/source_monday/schemas/boards.json
```diff
@@ -27,7 +27,7 @@
         "title": { "type": ["null", "string"] }
       }
     },
-    "id": { "type": ["null", "integer"] },
+    "id": { "type": ["null", "string"] },
```
girarda, line 30: changing the field types is a breaking change. we'll need to coordinate with TCS on releasing this if it's used in Cloud. cc @YowanR

airbyte-integrations/connectors/source-monday/source_monday/dpath_string_extractor.py
roman-yermilov-gl: Done

airbyte-integrations/connectors/source-monday/source_monday/pagination_workaround.py
roman-yermilov-gl: We need `page_size` to be optional and to be calculated on fly for those cases when we don't know exactly how many records we have for some endpoint. We also need `page` to start from 1

roman-yermilov-gl: Removed custom component. Managed it with default pagination strategy

airbyte-integrations/connectors/source-monday/source_monday/graphql_request_options_provider.py
girarda: can you add a comment explaining how you would want to use the commented code? Otherwise I think we should remove it if there's no context

girarda (approved):

PR #19599, opened 2022-11-18
🪟 🧹 Removes old statuses enum

timroes:
## What
Removes an old shared `Status` enum that should no longer be used, but instead the enums from the API generated types should be used.


krishnaglick (approved):
Did not test locally, LGTM if build is green

PR #19395, opened 2022-11-14
Add structured dbt cloud information to the operations api

mfsiega-airbyte:
## What
Add structured dbt Cloud information to the Operations API

## How
- Add the structured info to the openapi spec
- In the API/persistence layer converter, add logic to convert between persistence layer and API format

NOTE: this PR has dual-writing when we return results, both to the old and new formations for a "generic webhook". When the frontend switches to the new format, we'll remove the old one.

## Recommended reading order
1. `airbyte-api/src/main/openapi/config.yaml`
2. `airbyte-server/src/main/java/io/airbyte/server/converters/OperationsConverter.java`

PR #19600, opened 2022-11-18
🪟🔧 Change frontend dbt transformation to use structured fields

joeykmh:
## What
- Changes the body of the API request we send to create dbt transformation operations to match the [recent backend changes](https://github.com/airbytehq/airbyte/pull/19395)
- Makes a few styling tweaks
- Adds i18n for a few hard-coded strings

Styling changes:

Before:
image

After:
image

### DIFF
--- a/airbyte-webapp/src/packages/cloud/services/dbtCloud.ts
+++ b/airbyte-webapp/src/packages/cloud/services/dbtCloud.ts
@@ -11,7 +11,13 @@
 import isEmpty from "lodash/isEmpty";
 import { useMutation, useQuery } from "react-query";
 
-import { OperatorType, WebBackendConnectionRead, OperationRead, WebhookConfigRead } from "core/request/AirbyteClient";
+import {
+  OperatorType,
+  WebBackendConnectionRead,
+  OperationRead,
+  OperatorWebhookWebhookType,
+  WebhookConfigRead,
+} from "core/request/AirbyteClient";
 import { useWebConnectionService } from "hooks/services/useConnectionHook";
 import { useCurrentWorkspace } from "hooks/services/useWorkspace";
 import {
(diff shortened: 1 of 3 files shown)


airbyte-webapp/src/packages/cloud/services/dbtCloud.ts
```diff
@@ -25,47 +31,35 @@ import { useUpdateWorkspace } from "services/workspaces/WorkspacesService";
 import { useConfig } from "./config";
 
 export interface DbtCloudJob {
-  account: string;
-  job: string;
+  accountId: number;
+  jobId: number;
   operationId?: string;
   jobName?: string;
 }
```
josephkmh, line 38: @ambirdsall I wasn't quite sure if we even need this type anymore. Could we maybe just use `OperatorWebhookDbtCloud` from the Orval generated `AirbyteClient.ts`?

airbyte-webapp/src/packages/cloud/services/dbtCloud.ts
```diff
@@ -25,47 +31,35 @@ import { useUpdateWorkspace } from "services/workspaces/WorkspacesService";
...
+  if (operationRead.operatorConfiguration.webhook?.webhookType === "dbtCloud") {
+    const dbtCloud = operationRead.operatorConfiguration.webhook.dbtCloud as DbtCloudJob;
+    return {
+      accountId: dbtCloud.accountId,
+      jobId: dbtCloud.jobId,
+    };
   }
+  throw new Error(
+    `Cannot convert operationRead of type ${operationRead.operatorConfiguration.operatorType} to DbtCloudJob`
+  );
```
timroes, line 55: @josephkmh Is this something that can happen in reality, or just a dev catch? In either case, should we log this somehow to monitoring?

timroes (approved):
Code LGTM. Have tested basic operations locally. Seem to work.

ambirdsall (approved):
The logic changes are sound, the feature works, the UI looks better with the styling tweaks. :pray:

airbyte-webapp/src/packages/cloud/services/dbtCloud.ts
ambirdsall: There isn't really a way to do this more cleverly without maintaining handwritten types for this API: the java code generation doesn't support `anyOf` because java doesn't have proper union types, so we don't get to associate specific enum values with the presence or absence of fields in the auto-generated code.

PR #19606, opened 2022-11-18
feat: remove limit to amount of gradle workers and enable pytest parallelism

c-p-b:
## What
Removes the limit to amount of threads that gradle can spawn in order to get better cpu utilization, possibly with larger machines. Also add pytest parallelism

## How
Remove the gradle.properties setting and add pytest-xdist, which parallelizes pytest unit tests

### DIFF
--- a/.github/actions/build-branch/action.yml
+++ b/.github/actions/build-branch/action.yml
@@ -42,7 +42,6 @@ runs:
           --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED \
           --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED \
           --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
-        org.gradle.workers.max=8
         org.gradle.vfs.watch=false
         EOF
       shell: bash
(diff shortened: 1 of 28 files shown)


cpdeethree:
We confirmed that gradle was smart enough to work without setting the max limit (no difference between large/8 workers and large/no limit workers) so we will go ahead and remove the limit

alafanechere (commented):
Thanks for this initiative!
Could you please share how you spotted dependencies on the python tests?
I'm wondering if we could group some of these tests under common Test classes that you could mark directly, instead of decorating each test with the marker.

cpdeethree:
@alafanechere I think that would be more ideal. I'm using the marker because it's a quick and easy solution but I think it's not as maintainable to ask devs to remember to mark tests, but also I think in most cases you should probably not be having unit tests that have dependencies on each other.

I found them by enabling parallelization and then marking the tests which failed after doing that. It's not a perfect science, because obviously with multithreading you can have race conditions and only have it fail sometimes (so there might be some tests slipping through the cracks here) but it seems reliable enough. I would want to have it run green like 5-10 times before merging this to master though just to make sure, because of the nature of the multithreading it would be extremely confusing for a developer to get an error after I merged this and not know it was because of the multithreading.

alafanechere:
>  but also I think in most cases you should probably not be having unit tests that have dependencies on each other.
You are right. I'll log a tech dept issue to investigate why this deps exists and I'm fine using these marker as they are now to remind us even more of the problem :)

cpdeethree:
Since we are tackling this with dagger now I'm going to close this

PR #19610, opened 2022-11-18
Adding retry attempts in OSS repo workflows

Mykyta-Serbynevskyi:
As per https://github.com/airbytehq/airbyte-cloud/issues/3109 this PR adds retry attempts in GH workflows in `airbytehq/airbyte` repo.
The following retry actions were used:

https://github.com/marketplace/actions/retry-action (can be used for actions & steps)
https://github.com/marketplace/actions/retry-step (can be used for steps only)


.github/workflows/publish-command.yml
```diff
@@ -288,18 +288,25 @@ jobs:
           echo "IMAGE_VERSION=$(_get_docker_image_version ${DOCKERFILE})" >> $GITHUB_ENV
       - name: Prepare Sentry
         if: startsWith(matrix.connector, 'connectors')
-        run: |
-          curl -sL https://sentry.io/get-cli/ | bash || echo "sentry cli already installed"
+        uses: nick-fields/retry@v2
+        with:
+          timeout_seconds: 15
```
git-phu, line 293: would it be fine to retry without a timeout for this?

supertopher (approved):
this looks solid to me, I think we are safe to merge as long as you have seen this run once.

git-phu (commented):
looks great!

just one question

.github/workflows/publish-command.yml
Mykyta-Serbynevskyi: Done, I have removed timeout for this.

davydov-d:
Hey team, this PR brings in an unexpected behavior as for me. 
Looks like `fail-fast: false` + `retries` (correct me if I'm wrong) make long failing builds even longer. I bet we should only retry transient errors.
This one was killed due to a timeout (240 mins):
https://github.com/airbytehq/airbyte/actions/runs/3648925714/jobs/6162969779 - the reason is the tests are not passing.

Is it possible to alter the workflows so they would not run that long?

PR #19611, opened 2022-11-18
:window: :wrench: Show success message and clear form on successful dbt cloud token submisison

ambirdsall:
2022-11-18-10:34:49-screenshot

Wrote this quickly; I'm guessing that `components/ui/Toast` would be the conventional way to show the success message, rather than a `message` prop on the input, is that correct?

### DIFF
--- a/airbyte-webapp/src/packages/cloud/views/settings/integrations/DbtCloudSettingsView.tsx
+++ b/airbyte-webapp/src/packages/cloud/views/settings/integrations/DbtCloudSettingsView.tsx
@@ -1,6 +1,6 @@
 import { Field, FieldProps, Form, Formik } from "formik";
 import React, { useState } from "react";
-import { FormattedMessage } from "react-intl";
+import { FormattedMessage, useIntl } from "react-intl";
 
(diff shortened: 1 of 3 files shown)


airbyte-webapp/src/packages/cloud/views/settings/integrations/DbtCloudSettingsView.tsx
```diff
@@ -21,15 +22,20 @@ export const DbtCloudSettingsView: React.FC = () => {
           initialValues={{
             serviceToken: "",
           }}
-          onSubmit={({ serviceToken }) => {
+          onSubmit={({ serviceToken }, { resetForm }) => {
             setHasValidationError(false);
```
timroes, line 26: Just noticed: we likely also want to add a `setValidationMessage("")` here before querying. At the moment the error message stays during querying still visible but turned grey suddenly.

ambirdsall: Good call, will do.

PR #19617, opened 2022-11-18
Disallow deleting jobs while an update is pending

ambirdsall:
Prevents the Magically Reappearing Job bug shown in [this Loom recording](https://www.loom.com/share/7f8e583153394a6d86c8364dc198ac56). If the user clicks the "X"/"remove job" button several times in quick succession, they can see jobs reappear when a prior delete's successful response resets the form to the remaining jobs (some of which were already removed from the DOM when a subsequent update was cued).

### DIFF
--- a/airbyte-webapp/src/packages/cloud/locales/en.json
+++ b/airbyte-webapp/src/packages/cloud/locales/en.json
@@ -104,6 +104,7 @@
   "settings.integrationSettings.dbtCloudSettings.form.singleTenantUrl": "Single-tenant URL",
   "settings.integrationSettings.dbtCloudSettings.form.testConnection": "Test connection",
   "settings.integrationSettings.dbtCloudSettings.form.submit": "Save changes",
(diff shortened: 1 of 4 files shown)


joeykmh (approved):
Looks good! Tested deletion locally and worked well 👍

ambirdsall:
In all cases, deletion is directly handled by the remove function from formik's FieldArrayHelper; I'm a little surprised by its behavior. Specifically, these behaviors are a head-scratching combination:
- adding a new job does not trigger an update; the new, larger job list is only in memory until the user manually clicks the save button
- removing an unsaved new job does not trigger an update, it only modifies the unsubmitted, in-memory list
- removing one of multiple saved jobs does immediate trigger form submission (thus the bug Andy saw)
- removing the last saved job does not trigger form submission; the final deletion is only in memory until the user manually clicks the save button

A glance at [the FieldArray docs](https://formik.org/docs/api/fieldarray#fieldarray-helpers) doesn't show a way to configure the behavior of `remove`.

Issue #18709, opened 2022-10-31
Source Zenloop: Migrate connector from Alpha (Python) to Beta (YAML)


lazebnyi:
credentials - Zenloop Primary Dev Account

lazebnyi:
Migrated in https://github.com/airbytehq/airbyte/pull/19624

PR #19624, opened 2022-11-18
🎉Source Zenloop: migrate to lowcode

lazebnyi:
## What
https://github.com/airbytehq/airbyte/issues/18709

## How
Migrate to low code, set up tests, edit docs etc


davydov-d (approved):

airbyte-integrations/connectors/source-zenloop/integration_tests/configured_catalog.json
```diff
@@ -0,0 +1,256 @@
+{
+  "streams": [
+    {
+      "stream": {
+        "name": "answers",
+        "json_schema": {
```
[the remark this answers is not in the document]
girarda: +1

airbyte-integrations/connectors/source-zenloop/source_zenloop/components.py
```diff
@@ -0,0 +1,73 @@
...
+class ZenloopDatetimeStreamSlicer(DatetimeStreamSlicer):
+    def get_request_params(
+        self,
+        *,
+        stream_state: Optional[StreamState] = None,
+        stream_slice: Optional[StreamSlice] = None,
+        next_page_token: Optional[Mapping[str, Any]] = None,
+    ) -> Mapping[str, Any]:
+        params = super().get_request_params(stream_state=stream_state, stream_slice=stream_slice, next_page_token=next_page_token)
+        date_shortcut = {"date_shortcut": "custom"}
```
girarda, line 28: I don't think we need a custom stream slicer here. It looks like we can use the default `DatetimeStreamSlicer` and add a request param

```
        request_options_provider:
          request_parameters:
            order_type: "desc"
            order_by: "inserted_at"
            date_shortcut: "custom"
```

airbyte-integrations/connectors/source-zenloop/source_zenloop/components.py
```diff
@@ -0,0 +1,73 @@
...
+        params.update(date_shortcut)
+        return params
+
+
+@dataclass
+class ZenloopSubstreamSlicer(SubstreamSlicer):
```
girarda, line 34: can you document the custom behavior so we can better understand the intent and how we can generalize it in the future?

airbyte-integrations/connectors/source-zenloop/source_zenloop/components.py
```diff
@@ -0,0 +1,73 @@
...
+        if not custom_stream_state_value:
+            yield from super().stream_slices(sync_mode, stream_state)
+        else:
+            for parent_stream_config in self.parent_stream_configs:
+                stream_state_field = parent_stream_config.stream_slice_field or None
+                yield {stream_state_field: custom_stream_state_value, "parent_slice": {}}
+
+
+@dataclass
+class ZenloopPageIncrement(PageIncrement):
```
girarda, line 49: is the only difference between this and PageIncrement that the initial page is 1 instead of 0? If that's the case, you should be able to use [this field](https://github.com/airbytehq/airbyte/pull/19712/files#diff-e8ae2253fe4503731a489ac82e1a44326a59fe5a2eb27dcba8b93eea7b0bd52eR25) @davydov-d added

girarda (changes requested):

airbyte-integrations/connectors/source-zenloop/source_zenloop/components.py
lazebnyi: Updated.

airbyte-integrations/connectors/source-zenloop/acceptance-test-config.yml
```diff
@@ -13,31 +13,12 @@ tests:
     - config_path: "secrets/config.json"
   basic_read:
     - config_path: "secrets/config.json"
-      configured_catalog_path: "integration_tests/configured_catalog_answers.json"
```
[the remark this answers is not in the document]
lazebnyi: yes

airbyte-integrations/connectors/source-zenloop/integration_tests/configured_catalog.json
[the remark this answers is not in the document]
lazebnyi: Updated.

airbyte-integrations/connectors/source-zenloop/source_zenloop/components.py
lazebnyi: Added

airbyte-integrations/connectors/source-zenloop/source_zenloop/streams.py
```diff
@@ -0,0 +1,229 @@
+#
```
girarda, line 1: are the classes in this file used?

airbyte-integrations/connectors/source-zenloop/source_zenloop/components.py
```diff
@@ -0,0 +1,53 @@
...
+import requests
+from airbyte_cdk.models import SyncMode
+from airbyte_cdk.sources.declarative.requesters.paginators.strategies.page_increment import PageIncrement
+from airbyte_cdk.sources.declarative.stream_slicers.datetime_stream_slicer import DatetimeStreamSlicer
+from airbyte_cdk.sources.declarative.stream_slicers.substream_slicer import SubstreamSlicer
+from airbyte_cdk.sources.declarative.types import StreamSlice, StreamState
+
+
+@dataclass
+class ZenloopDatetimeStreamSlicer(DatetimeStreamSlicer):
```
girarda, line 19: do we need this class or could we use the default `DatetimeStreamSlicer` and add a request param in the requester's config?

```
request_options_provider:
  request_parameters:
    order_type: "desc"
    order_by: "inserted_at"
    "date_shortcut": "custom"
```

girarda (changes requested):

airbyte-integrations/connectors/source-zenloop/source_zenloop/components.py
lazebnyi: Yes, we need this class. The reason is that we need ` "date_shortcut": "custom"` param only during incremental sync. More context above - https://github.com/airbytehq/airbyte/pull/19624#discussion_r1029881973

airbyte-integrations/connectors/source-zenloop/source_zenloop/streams.py
YowanR: I'm using this [doc](https://docs.google.com/document/d/1YcSpPmGkKWXtrTh_BNx8L2L6vcuSk_KSkPqezTjtnkA/edit#) as reference. I think that we would need both until our confidence is low-code is higher and we won't need to fall back to Python. So both should be good for now but happy to hear your thoughts, @girarda.

airbyte-integrations/connectors/source-zenloop/source_zenloop/components.py
lazebnyi: Updated.

airbyte-integrations/connectors/source-zenloop/source_zenloop/components.py
lazebnyi: Updated

girarda (approved):

airbyte-integrations/connectors/source-zenloop/source_zenloop/streams.py
girarda: it's pretty easy to revert a pull request so I'd prefer we delete them, but I'm also fine with keeping them for a bit so long as we create issue to keep track of the debt

PR #19625, opened 2022-11-18
RecordSchemaValidator can resolve $ref schemas

edgao:
Wire up the RecordSchemaValidator with the new data types protocol typedefs. Specifically:
* Dockerfile now copies in WellKnownTypes.yaml and converts it to JSON
* JsonSchemaValidator correctly resolves the `$ref` (the changes are necessary because otherwise it would require `file:WellKnownTypes.json`, which breaks in Python schema validation)

Existing behavior should remain unchanged.

Example jsonl file -> dev-null sync: (note that it correctly detects the `int` field as being the wrong type)
Screen Shot 2022-11-18 at 2 22 59 PM

This was the schema: (I ran this locally and edited the `connection` record manually)
```json
{
  "int": {
    "$ref": "WellKnownTypes.json#/definitions/Integer"
  },
  "str": {
    "$ref": "WellKnownTypes.json#/definitions/Boolean"
  }
}
```

Reading order:
1. build.gradle + Dockerfile
2. JsonSchemaValidator.java
3. JsonSchemaValidatorTest


airbyte-workers/Dockerfile
```diff
@@ -20,5 +20,11 @@ WORKDIR /app
 # Move worker app
 ADD bin/${APPLICATION}-${VERSION}.tar /app
 
+# Grab well-known types file
+RUN wget https://github.com/mikefarah/yq/releases/download/v4.30.4/yq_linux_${DOCKER_BUILD_ARCH}.tar.gz -O - | tar xz \
```
gosusnp, line 24: nit: This could be done earlier for a better caching of the steps. Probably around line 10.

airbyte-workers/Dockerfile
```diff
@@ -20,5 +20,11 @@ WORKDIR /app
...
+        && mv yq_linux_${DOCKER_BUILD_ARCH} /usr/bin/yq
+COPY well_known_types.yaml /app
+RUN yq -o json /app/well_known_types.yaml > /app/WellKnownTypes.json
```
gosusnp, line 27: It looks like we only need the json. Can we have this as a gradle task rather than having file transformation in the Dockerfile?

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
```diff
@@ -26,12 +31,24 @@ public class JsonSchemaValidator {
...
 
   public JsonSchemaValidator() {
-    this.schemaValidatorsConfig = new SchemaValidatorsConfig();
+    // This URI just needs to point at any path in the same directory as /app/WellKnownTypes.json
+    // It's required for the JsonSchema#validate method to resolve $ref correctly.
+    this("file:///app/nonexistent_file.json");
+  }
+
+  @VisibleForTesting
+  protected JsonSchemaValidator(String baseUri) {
```
gosusnp, line 44: Passing an `URI` here instead of a String would be a clearer contract and it removes the need for the exception wrapping in the constructor.

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
```diff
@@ -26,12 +31,24 @@ public class JsonSchemaValidator {
...
 
-  private final SchemaValidatorsConfig schemaValidatorsConfig;
   private final JsonSchemaFactory jsonSchemaFactory;
+  private final URI BASE_URI;
 
   public JsonSchemaValidator() {
-    this.schemaValidatorsConfig = new SchemaValidatorsConfig();
+    // This URI just needs to point at any path in the same directory as /app/WellKnownTypes.json
+    // It's required for the JsonSchema#validate method to resolve $ref correctly.
+    this("file:///app/nonexistent_file.json");
```
gosusnp, line 40: What is this `BASE_URI` for? Could be good to have some explanation here, it's not obvious why we only want to expose this for tests given its current usage.

airbyte-workers/Dockerfile
edgao: removed this entirely since the file gets generated in gradle

airbyte-workers/Dockerfile
edgao: done

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
edgao: done - there's a static initializer for the default uri but I think that's fine since it's hardcoded to a valid value

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
edgao: added some comments - lmk if this is still unclear

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
```diff
@@ -25,13 +30,37 @@
 public class JsonSchemaValidator {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(JsonSchemaValidator.class);
+  // This URI just needs to point at any path in the same directory as /app/WellKnownTypes.json
+  // It's required for the JsonSchema#validate method to resolve $ref correctly.
+  private static final URI DEFAULT_BASE_URI;
+
+  static {
+    try {
+      DEFAULT_BASE_URI = new URI("file:///app/nonexistent_file.json");
```
gosusnp, line 39: Why not use `file:///app/WellKnownTypes.json` instead?

airbyte-workers/Dockerfile
```diff
@@ -3,6 +3,9 @@ FROM ${JDK_IMAGE} AS worker
 
 ARG DOCKER_BUILD_ARCH=amd64
 
+# Grab well-known types file
+COPY WellKnownTypes.json /app
```
gosusnp, line 7: I would suggest moving this COPY to line 23. This would avoid rebuilding the complete image whenever we change this file.

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
gosusnp: So if I understand correctly, we pass a value so that the validation can generate local references that we ignore anyway?

gosusnp (approved):
LGTM.

Left a couple suggestions. I don't think the docker would matter a lot given this file shouldn't change a lot. In general, I think it's a better practice to keep the copy of build artifacts closer to the end for better caching.

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
edgao: just to make it clear that it doesn't need to be the exact file being referenced. E.g. if we later added a second file, we wouldn't need to update this URI.

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
edgao: the references aren't ignored - see the new unit test where the validator _does_ read the referenced file

airbyte-workers/Dockerfile
edgao: done

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
gosusnp: Yes, the part that always feels a bit unusual is referencing a non existing file.

edgao: hm. what if we appended the `nonexistent_file.json` in the constructor? so callers would just need `new JsonSchemaValidator(new URI("file:///app/"))`

edgao: bump @gosusnp

gosusnp: Not sure that appending this in the constructor helps understanding that much. It probably boils down to why do we actually need to pass a full URI rather than just the actual base URI.
I'd say that it doesn't feel ideal, at the same time, it doesn't seem important enough to block this PR. The comments helps with the understanding.

PR #19221, opened 2022-11-09
🪟🔧 Use workspace-scoped endpoints for connector CRUD

joeykmh:
## What
Closes https://github.com/airbytehq/airbyte/issues/11606
Closes https://github.com/airbytehq/airbyte/issues/11607

## How
- Updates API methods to use workspace-scoped endpoints
- Fixes visual bug where loading state was not visible on `CreateConnectorModal`
- Adds a migration to convert all existing custom connector definitions in oss deployments to be correctly configured as custom definitions based on the new schema changes

## Recommended reading order
1. [src/core/domain/connector/DestinationDefinitionService.ts](https://github.com/airbytehq/airbyte/pull/19221/files#diff-a5ce3ca8b4b07c3d44aab322ec5bfaee363217f9033a3d1f8a045c50b83bb59a) and [src/core/domain/connector/SourceDefinitionService.ts](https://github.com/airbytehq/airbyte/pull/19221/files#diff-974f2a2c0efb257880efe98d7b3a468961cce2f46df9ce1cbd26edf6e24813a3) contain the updated API methods
2. [src/services/connector/DestinationDefinitionService.ts](https://github.com/airbytehq/airbyte/pull/19221/files#diff-6090f0111a82c28bf7ea0f71c24e3bd075c78a9a4f5967db8b4bc42c52136051) and [src/services/connector/SourceDefinitionService.ts](https://github.com/airbytehq/airbyte/pull/19221/files#diff-de422c03ac5c766c63eb05946dce0e2db6a39e7fe3b1c7bbaefca6ca010854f8) contain the updated hooks
3. I also fixed a small bug in [CreateConnectorModal.tsx](https://github.com/airbytehq/airbyte/pull/19221/files#diff-21311e51332dbb2fc443c2702e738d6ac4d34ecd3f0c40590db68cd123f8757b) that I discovered while testing

## 🚨 User Impact 🚨
This PR includes a db migration that addresses https://github.com/airbytehq/airbyte/issues/11607

PR #19626, opened 2022-11-18
Revert "🪟🔧 Use workspace-scoped endpoints for connector CRUD"

krishnaglick:
Reverts airbytehq/airbyte#19221
Relevant [slack](https://airbytehq-team.slack.com/archives/C03AS1GAQV6/p1668798021553059) conversation.
This PR makes all source and destination definition update calls qualify as custom which is not desired behavior.


airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_40_18_004__BackfillActorDefinitionWorkspaceGrant.java
```diff
@@ -1,50 +0,0 @@
...
-      for (final var existingWorkspaceRecord : existingWorkspaces) {
-        // Populate a record for new table;
-        var customActorDefinitionIdValue = customActorDefinitionIdRecord.getValue("id", UUID.class);
-        var existingWorkspaceIdValue = existingWorkspaceRecord.getValue("id", UUID.class);
-
-        ctx.execute("INSERT INTO actor_definition_workspace_grant(workspace_id, actor_definition_id) VALUES ({0}, {1})",
-            existingWorkspaceIdValue, customActorDefinitionIdValue);
-      }
-    }
-  }
```
ambirdsall, line 48: I don't really have enough context on the system that runs our database migrations to say if this is safe to revert this way. Naively, I don't see a specific reason to expect troubles from reading the SQL--it's not changing the db structure, just making an idempotentish* write to one specific field--but it seems like someone who does have that context should weigh in.

\* the "ish" is just a nod to the possibility of an `updatedAt` column

krishnaglick:
Closed in favor of a more surgical revert.

ambirdsall (commented):
Submitting the in-progress review for posterity, but stopping here since this PR has been closed in favor of a new one.

PR #19627, opened 2022-11-18
No longer forces custom source/destination definition always

krishnaglick:
## What
This [PR](https://github.com/airbytehq/airbyte/pull/19221) introduced a bug that made all source and destination definitions create or update a custom.
Rather than a full revert, this PR simply returns the old functionality.


ambirdsall (approved):
This fix makes sense in light of the bug as reported/diagnosed (create/update calls failed for all non-custom connectors since they used the custom connector create/update calls but lacked the corresponding grants). I did confirm that this correctly reverts all usages of these methods, but did not fully test locally.

PR #19640, opened 2022-11-21
Create Config Database test base class

cgardens:
## What
* As a prerequisite to getting rid of the `DatabaseConfigPersistence` I realized that over time the `BaseDatabaseConfigPersistenceTest` had been misused. This was making it so that test run much slower (because the full database migration was being run between each test) _and_ that it was much harder to write test classes for checking our database queries because lots of boilerplate was being needlessly copied and pasted because people didn't understand how to use the base class.

## How
* Replaced `BaseDatabaseConfigPersistenceTest` with `BaseConfigDatabaseTest`. This is a much more opinionated abstraction that I think will be harder to misuse. The javadocs in the class explain it in detail, so I will not repeat it here.
* The rest of the changes in this PR are moving existing test classes to use the new abstraction.
* This new abstraction will get used heavily in the next PR that removes `DatabaseConfigPersistence` which relies on us splitting out our tests more clearly. Decided to separate out this change for clarity.

## Results
Time to run `airbyte-config:config-persistence:test` before this change:
Screen Shot 2022-11-20 at 4 55 01 PM

Time to run  `airbyte-config:config-persistence:test`  after this change:
Screen Shot 2022-11-20 at 4 50 52 PM

A neat 55% speed increase!! 😄 

## Recommended reading order
1. `BaseConfigDatabaseTest.java`
2.  the rest

## 🚨 User Impact 🚨
* This PR does NOT touch any production code. It is just touching tests and test abstraction. It will not have any effect on production behavior.


airbyte-config/config-persistence/src/test/java/io/airbyte/config/persistence/DatabaseConfigPersistenceTest.java
```diff
@@ -50,29 +55,75 @@
  * methods.
  */
 @SuppressWarnings({"PMD.SignatureDeclareThrowsException", "PMD.ShortVariable", "PMD.JUnitTestsShouldIncludeAssert"})
-class DatabaseConfigPersistenceTest extends BaseDatabaseConfigPersistenceTest {
+class DatabaseConfigPersistenceTest extends BaseConfigDatabaseTest {
+
+  public static final String DEFAULT_PROTOCOL_VERSION = "0.2.0";
+  protected static final StandardSourceDefinition SOURCE_GITHUB = new StandardSourceDefinition()
```
cgardens, line 61: all of this noise is copied from `BaseDatabaseConfigPersistenceTest`. It will all get removed in the next PR which gets rid of `DatabaseConfigPersistence` entirely.

airbyte-config/config-persistence/src/test/java/io/airbyte/config/persistence/StatePersistenceTest.java
```diff
@@ -55,6 +48,36 @@ class StatePersistenceTest extends BaseDatabaseConfigPersistenceTest {
...
 
+  @BeforeEach
+  void beforeEach() throws DatabaseInitializationException, IOException, JsonValidationException, SQLException {
+    truncateAllTables();
+
+    setupTestData();
+    statePersistence = new StatePersistence(database);
+  }
+
+  private void setupTestData() throws JsonValidationException, IOException {
```
cgardens, line 59: this diff looks bigger than it is. this is just pulled up from lower down in the class.

jdpgrailsdev (approved):
:shipit:

benmoriceau (approved):

gosusnp (approved):

PR #19646, opened 2022-11-21
airbyte-cdk offset pagination strategy: page_size to be interpolated …

### LINKED ISSUE #20291: airbyte-cdk offset pagination strategy: page_size to be interpolated
It is often happen that page_size need to be specified per stream. With current changes we can set the page_size in stream options while define it only once:

`page_size: "{{ options['items_per_page'] }}"`

roman-yermilov-gl:
It is often happen that `page_size` need to be specified per stream. With current changes we can set the `page_size` in stream options while define it only once:
```
page_size: "{{ options['items_per_page'] }}"

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
@@ -3,10 +3,12 @@
 #
 
 from dataclasses import InitVar, dataclass
-from typing import Any, List, Mapping, Optional
+from typing import Any, List, Mapping, Optional, Union
 
 import requests
 from airbyte_cdk.sources.declarative.requesters.paginators.strategies.pagination_strategy import PaginationStrategy
+from airbyte_cdk.sources.declarative.interpolation import InterpolatedString
+from airbyte_cdk.sources.declarative.types import Config
 from dataclasses_jsonschema import JsonSchemaMixin
 
 
@@ -16,14 +18,17 @@ class OffsetIncrement(PaginationStrategy, JsonSchemaMixin):
     Pagination strategy that returns the number of records reads so far and returns it as the next page token
 
     Attributes:
-        page_size (int): the number of records to request
+        page_size (InterpolatedString): the number of records to request
     """
 
-    page_size: int
+    config: Config
+    page_size: Union[InterpolatedString, str, int]
     options: InitVar[Mapping[str, Any]]
 
     def __post_init__(self, options: Mapping[str, Any]):
         self._offset = 0
+        self.page_size.string = str(self.page_size.string)
+        self.page_size = self.page_size.eval(self.config)
 
     def next_page_token(self, response: requests.Response, last_records: List[Mapping[str, Any]]) -> Optional[Any]:
         if len(last_records) < self.page_size:
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
@@ -16,6 +16,21 @@ from dataclasses_jsonschema import JsonSchemaMixin
 class OffsetIncrement(PaginationStrategy, JsonSchemaMixin):
     """
     Pagination strategy that returns the number of records reads so far and returns it as the next page token
+    Examples:
+        # page_size to be a constant integer value
+        pagination_strategy:
+          type: OffsetIncrement
+          page_size: 2
+
+        # page_size to be a constant string value
+        pagination_strategy:
+          type: OffsetIncrement
+          page_size: "2"
+
+        # page_size to be an interpolated string value
+        pagination_strategy:
+          type: OffsetIncrement
+          page_size: "{{ options['items_per_page'] }}"
 
     Attributes:
         page_size (InterpolatedString): the number of records to request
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by sherifnada — CHANGES_REQUESTED** [review]
@roman-yermilov-gl could you indicate why you're making this change, include a snippet of how this changes the input YAML, and add tests that prove the new functionality works? It's hard to give useful feedback on a PR without that information :)

**Comment by roman-yermilov-gl** [comment]
> @roman-yermilov-gl could you indicate why you're making this change, include a snippet of how this changes the input YAML, and add tests that prove the new functionality works? It's hard to give useful feedback on a PR without that information :)

Thanks for comments. I updated description and made fixes in tests

File: airbyte-cdk/python/unit_tests/sources/declarative/requesters/paginators/test_offset_increment.py [hunk 1]
```diff
@@ -6,18 +6,21 @@
...
+from airbyte_cdk.sources.declarative.interpolation import InterpolatedString
 from airbyte_cdk.sources.declarative.requesters.paginators.strategies.offset_increment import OffsetIncrement
 
 
 @pytest.mark.parametrize(
     "test_name, page_size, expected_next_page_token, expected_offset",
     [
-        ("test_same_page_size", 2, 2, 2),
-        ("test_larger_page_size", 3, None, 0),
+        ("test_same_page_size", InterpolatedString(string="2", options={}), 2, 2),
```
girarda, line 17: the page_size parameter can be an InterpolatedString, a str, and an int. we should have a test case for all 3 input types

airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
```diff
@@ -16,14 +18,17 @@ class OffsetIncrement(PaginationStrategy, JsonSchemaMixin):
...
     """
 
-    page_size: int
+    config: Config
+    page_size: Union[InterpolatedString, str, int]
     options: InitVar[Mapping[str, Any]]
 
     def __post_init__(self, options: Mapping[str, Any]):
         self._offset = 0
+        self.page_size.string = str(self.page_size.string)
```
girarda, line 30: this won't work if the page_size isn't already an InterpolatedString. It's a little annoying to do, but you'll need to check the type of the field and update it accordingly

girarda (changes requested):

airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
roman-yermilov-gl: I have tested it with three value types and it works without errors. As I see we always have `self.page_size` as interpolated string type. The difference is in `.string` attribute which has int/str/ type depending on value in .yaml config. This was test value in my config:
int:
```
pagination_strategy:
      type: OffsetIncrement
      page_size: 2
```
str
```
pagination_strategy:
      type: OffsetIncrement
      page_size: "2"
```
Interpolated string
```
pagination_strategy:
      type: OffsetIncrement
      page_size: "{{ options['items_per_page'] }}"
```

airbyte-cdk/python/unit_tests/sources/declarative/requesters/paginators/test_offset_increment.py
roman-yermilov-gl: I made this fixes accordion to explanation to the comment above

girarda (approved):
🚢

airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
```diff
@@ -3,27 +3,49 @@
...
+    config: Config
+    page_size: Union[InterpolatedString, str, int]
     options: InitVar[Mapping[str, Any]]
 
     def __post_init__(self, options: Mapping[str, Any]):
         self._offset = 0
+        # InterpolatedString page_size may contain one of int/str types,
+        # so we need to ensure that its `.string` attribute is of *string* type
+        self.page_size.string = str(self.page_size.string)
+        self.page_size = self.page_size.eval(self.config)
```
brianjlai, line 48: When possible, we try to perform the interpolation evluation `eval()` at runtime instead of parse time in `post_init`. So in this case, self.page_size would be the interpolated string (as you mentioned above) but we should move evaluation of it into the `get_page_size()`.

Now that this could be an interpolated string or just a normal string, the type hints in the getter method `get_page_size() -> Optional[int]` is now a bit misleading. I think we should also add some error handling in the event that this does not evaluate into an integer since it won't behave correctly when `get_page_size()` gets invoked by parent classes.

roman-yermilov-gl: Fixed

airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
```diff
@@ -3,30 +3,51 @@
...
 
     def __post_init__(self, options: Mapping[str, Any]):
         self._offset = 0
+        # InterpolatedString page_size may contain one of int/str types,
+        # so we need to ensure that its `.string` attribute is of *string* type
+        self.page_size.string = str(self.page_size.string)
 
     def next_page_token(self, response: requests.Response, last_records: List[Mapping[str, Any]]) -> Optional[Any]:
-        if len(last_records) < self.page_size:
+        if len(last_records) < self.page_size.eval(self.config):
```
brianjlai, line 50: instead of calling `self.page_size.eval(self.config)` we can replace this with `self.get_page_size()` so that we can reuse the method without needing to duplicate the type check

brianjlai (approved):
looks good! just one comment about reusing some code

sherifnada (approved):

girarda:
gentle reminder to run `SUB_BUILD=CONNECTORS_BASE ./gradlew format` to make sure CI passes before merging https://github.com/airbytehq/airbyte/actions/runs/3630457570/jobs/6123859471.

PR #19660, opened 2022-11-21
Extend template for the Java JDBC template by SAT SPEC

### LINKED ISSUE #19130: Extend source-java-jdbc by files required for SAT SPEC (Example #18779)
We integrated SAT SPEC into the Java sources in PR #18779.
We need to update our `source-java-jdbc` module to generate these files in the new java source connectors by default.

DoNotPanicUA:
## What
Update Java JDBC source template to enable SAT SPEC by default for generated sources.
airbytehq/airbyte#19130

## Test
Steps to reproduce:
1. Generate Java JDBC source
2. Build the image (exclude unit tests from the build by `-x test` as there is no implementation inside)
3. Run SAT locally `acceptance-test-docker.sh`

Result:
Screenshot from 2022-11-21 18-42-49


airbyte-integrations/connector-templates/source-java-jdbc/src/test-integration/resources/expected_spec.json
```diff
@@ -0,0 +1,61 @@
...
+        "description": "Replication method to use for extracting data from the database. STANDARD replication requires no setup on the DB side but will not be able to represent deletions incrementally. CDC uses the Binlog to detect inserts, updates, and deletes. This needs to be configured on the source database itself.",
+        "type": "string",
+        "order": 6,
+        "default": "STANDARD",
+        "enum": ["STANDARD", "CDC"]
+      }
+    }
+  },
+  "supported_destination_sync_modes": []
+}
```
alafanechere, line 61: Missing new line 😄

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-config.yml
```diff
@@ -0,0 +1,7 @@
+# See [Source Acceptance Tests](https://docs.airbyte.com/connector-development/testing-connectors/source-acceptance-tests-reference)
+# for more information about how to configure these tests
+connector_image: airbyte/source-{{dashCase name}}:dev
```
alafanechere, line 3: The template files are usually using `.hsb` extension

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-config.yml
```diff
@@ -0,0 +1,7 @@
...
+tests:
```
alafanechere, line 4: Please re-use the files of https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connector-templates/source_acceptance_test_files/acceptance-test-config.yml.hbs#L1 . It has the latest configuration structure. I think it's fine to expose all the tests and let the connector developer decide which one they can enable.

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-docker.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env sh
+
```
alafanechere, line 2: Please reuse https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connector-templates/source_acceptance_test_files/acceptance-test-config.yml.hbs#L1

airbyte-integrations/connector-templates/source-java-jdbc/integration_tests/acceptance.py
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (c) 2022 Airbyte, Inc., all rights reserved.
```
alafanechere, line 2: I think you can copy all the content of this folder `airbyte-integrations/connector-templates/source-python/integration_tests`

alafanechere (changes requested):
I think you can re-use existing templates instead of re-creating files like `acceptance-test-config.yml` and `acceptance-test-docker.sh`.

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-config.yml
DoNotPanicUA: The original scope was to add only tests we already enabled for the existing sources. Because we are sure what exactly is required and how it will work for the Java sources.
Of course, it's not a big deal just put everything from the doc here as I do not need to work with it, but someone who wants to create a new source will have to do his investigation of how to make it work and what is missing.
I propose to add only what we can guarantee here.

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-docker.sh
DoNotPanicUA: let discuss it here https://github.com/airbytehq/airbyte/pull/19660/files/eea4d4b555919f46d790572f3a1359450e930337#r1028406403

airbyte-integrations/connector-templates/source-java-jdbc/integration_tests/acceptance.py
DoNotPanicUA: I worry that we can make this generator worth it if we add things we don't use. At least until we enabled DISCOVER and READ tests for the existing source and specify what we need and why.
I remember the expectation of enabling SPEC and what we have. Here we might have even more changes required besides just turning it on.

airbyte-integrations/connector-templates/source-java-jdbc/src/test-integration/resources/expected_spec.json
DoNotPanicUA: Luckily I have a few unused new lines for this case. :smile:

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-config.yml
alafanechere: I'd suggest to comment the disabled test then. IMO it's developer's job to tweak the test config according to what's possible to run in the context of their connector.

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-docker.sh
alafanechere: Sorry I meant this file https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connector-templates/source_acceptance_test_files/acceptance-test-docker.sh#L1

Your new file is not different from this one isn't it?

airbyte-integrations/connector-templates/source-java-jdbc/integration_tests/acceptance.py
alafanechere: > Here we might have even more changes required besides just turning it on.

I don't get which change you mean.
IMO these files are only examples that could be used in the SAT context.
Developers are free to remove these if it's not used. 
When a new connector gets generated it's not expected they directly pass all SATs. So showing devs what can be tested remains valuable. We have no clue that a new JDBC connector can't pass tests other than SPEC, so let's offer them the opportunity to do so. Wdyt?

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-docker.sh
DoNotPanicUA: :+1:

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-config.yml
DoNotPanicUA: Done

airbyte-integrations/connector-templates/source-java-jdbc/integration_tests/acceptance.py
DoNotPanicUA: I've managed to integrate the BigQuery source with the SAT in the PR. As a result, I got a list of files and required content to add here.

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-docker.sh.hbs
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env sh
+
```
alafanechere, line 2: This file is not templated, so I think the `.hbs` extension is unnecessary.

airbyte-integrations/connector-templates/source-java-jdbc/integration_tests/acceptance.py.hbs
```diff
@@ -0,0 +1,16 @@
+#
```
alafanechere, line 1: This file is not templated, so I think the `.hbs` extension is unnecessary.

airbyte-integrations/connector-templates/source-java-jdbc/integration_tests/configured_catalog.json.hbs
```diff
@@ -0,0 +1,19 @@
+{
```
alafanechere, line 1: This file is not templated, so I think the `.hbs` extension is unnecessary.

airbyte-integrations/connector-templates/source-java-jdbc/integration_tests/configured_catalog_inc.json.hbs
```diff
@@ -0,0 +1,16 @@
+{
```
alafanechere, line 1: This file is not templated, so I think the `.hbs` extension is unnecessary.

airbyte-integrations/connector-templates/source-java-jdbc/integration_tests/expected_records.json.hbs
```diff
@@ -0,0 +1,2 @@
+{"stream": "stream_name", "data": {"column_1": "value_1", "column_2": -128 }, "emitted_at": 1626172757000 }
```
alafanechere, line 1: This file is not templated, so I think the `.hbs` extension is unnecessary.

airbyte-integrations/connector-templates/source-java-jdbc/integration_tests/sat-config.json.hbs
```diff
@@ -0,0 +1,7 @@
+{
```
alafanechere, line 1: This file is not templated, so I think the `.hbs` extension is unnecessary.

airbyte-integrations/connector-templates/source-java-jdbc/src/test-integration/resources/dummy_config.json.hbs
```diff
@@ -0,0 +1,7 @@
+{
```
alafanechere, line 1: This file is not templated, so I think the `.hbs` extension is unnecessary.

alafanechere (changes requested):
As I mentionned, most of the file you created are already existing in an exact same format under `airbyte-integrations/connector-templates/source-python/integration_tests`.
For simpler maintenance, I would prefer you pick one of the following approach:
1. Change the `plopfile.js` to make `Java JDBC Source` use the `source-python/integration_tests` files
2. Create a symlink in `source-java-jdbc/integration-tests` to `source-python/integration-tests`
3. Rework the `connector-templates` directory structure to declare files common to both `source-python` and `source-java-jdbc` and handle this with the `plopfile.js` 

I don't want to be a blocker on this. I'd love to get the opinion of someone else on the team.

(please format the files 😄 I know you still have new lines in stock to add these to the end of your files)

DoNotPanicUA:
@alafanechere 
As we agreed, I've left only SPEC-related files and created [the follow-up issue](https://github.com/airbytehq/airbyte/issues/19989).
Please let me know if some changes are still required.

alafanechere (approved):
Please fix the missing new line 😄 Otherwise LGTM

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-docker.sh
```diff
@@ -0,0 +1,15 @@
...
+# Pull latest acctest image
+docker pull airbyte/source-acceptance-test:latest
+
+# Run
+docker run --rm -it \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    -v /tmp:/tmp \
+    -v $(pwd):/test_input \
+    airbyte/source-acceptance-test \
+    --acceptance-test-config /test_input
```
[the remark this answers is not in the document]
DoNotPanicUA: they found a way how to hide from me :eyes:

PR #19664, opened 2022-11-21
DO NOT MERGE: build: use cached docker image layers

c-p-b:
## What
Cache docker image layers for usage in future builds and distribute them to the AMI runners via S3

## How
Save the docker outputs to tar file, upload to s3, then when we boot up a runner pull the tar file in and load it


cpdeethree:
This is a test PR that is purposefully failing to test whether merge on red is enabled properly

perangel (approved):
👍🏼

PR #19678, opened 2022-11-21
🪟 🎉 Add datepicker for date/date-time fields in connector form

joeykmh:
## What
- Adds a datepicker component
- Displays this component in the connector form when a field with format `date` or `date-time` is rendered

https://user-images.githubusercontent.com/7550957/203813360-2f765d96-5262-423a-b4c3-bb1e1bd9099d.mov

## How
- Adds `react-datepicker` and styles it with our design tokens
- Adds a new `<Datepicker />` component which renders a text input with a datepicker button.
- Only UI widgets with the correct json-schema `format` of `date` or `date-time` will see the datepicker. Many connectors today have date fields that are typed as `string`. In these cases a normal text input will be rendered (see "End date" in the video above).

## Recommended reading order
Top to bottom

## 🚨 User Impact 🚨
The datepicker is currently behind the `connector.form.useDatepicker` experiment flag. It's enabled by default on OSS, and can be toggled easily on LaunchDarkly for cloud.

The plan is to roll out the datepicker to all users on cloud as well, and remove the flag if all goes well. Even if there is a bug with the datepicker itself, a normal `<Input />` is rendered where the user can enter the UTC timestamp manually.


airbyte-webapp/craco.config.js
```diff
@@ -2,6 +2,14 @@
 const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 
 module.exports = {
+  jest: {
+    configure: (jestConfig) => {
+      return {
+        ...jestConfig,
+        setupFiles: [...jestConfig.setupFiles, "./__test__/globalSetup.ts"],
```
josephkmh, line 9: Is `__test__` a good place for this to live? I could not find an existing alternative.

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/PropertyLabel.tsx
```diff
@@ -24,7 +24,13 @@ export const PropertyLabel: React.FC<React.PropsWithChildren<PropertyLabelProps>
...
   htmlFor,
 }) => {
-  const examples = property._type === "formItem" || property._type === "formGroup" ? property.examples : undefined;
+  const examples =
+    // Show examples for individual items and groups, unless its a date field, since we don't
+    // want to confuse the user around potential date format examples
+    (property._type === "formItem" && property.format !== "date-time" && property.format !== "date") ||
+    property._type === "formGroup"
+      ? property.examples
+      : undefined;
```
josephkmh, line 33: @timroes I don't quite get this - why would users be confused by an example date format like this?
image

timroes: I removed this when I was still planning to do present the date inside the field in a different format. Since we're not planning this at the moment. With us sticking to present the ISO format, I agree that this can still be shown.

airbyte-webapp/craco.config.js
timroes: Jest (as configured in CRA) already reads by default the `strc/setupTests.ts` file. I think we should move the setup code just into this file instead?

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/Control.tsx
```diff
@@ -33,6 +35,7 @@ export const Control: React.FC<ControlProps> = ({
   error,
 }) => {
   const [field, meta, helpers] = useField(name);
+  const canUseDatepicker = useFeature(FeatureItem.AllowUseDatepicker);
```
timroes, line 38: This should not be a feature, but instad should just be an experiment using `useExperiment`. There is no reason why we'd want to have this environment dependant, what features are meant for, but all we want is an experiment toggle for this. `useExperiment` would be the right place for this.

timroes:
Two small additional notes: Could we maybe add a Storybook story for this component. Also I think for users to always find it, it would be nice if we would open the popup of the datepicker still when the input gains focus. This should be possible via `setOpen(true)` on the ref (https://github.com/Hacker0x01/react-datepicker/issues/685).

josephkmh:
@timroes I've added a storybook story and also made the datepicker open on input focus. The story is kind of buggy, since storybook unmounts/remounts the component when args change. There is an [issue open](https://github.com/storybookjs/storybook/issues/11657) on storybook but none of the approaches there are working. It's not a priority, so going to ignore it for now 🙂

airbyte-webapp/src/components/ui/DatePicker/DatePicker.module.scss
```diff
@@ -0,0 +1,147 @@
...
+  }
+}
+
+:global(.react-datepicker__day--outside-month) {
+  color: colors.$grey-300;
+}
+
+:global(.react-datepicker__day--today) {
+  background-color: colors.$grey-50;
+  font-weight: 7 00;
```
lmossman, line 87:
```suggestion
  font-weight: 700;
```
I think this was a typo?

airbyte-webapp/src/components/ui/DatePicker/DatePicker.module.scss
```diff
@@ -0,0 +1,147 @@
...
+  display: flex;
+  color: colors.$dark-blue-900;
+  font-family: fonts.$primary;
+  border: variables.$border-thick solid colors.$grey-100;
+  border-radius: variables.$border-radius-md;
+  box-shadow: 2px 2px 20px -6px colors.$grey-200;
+}
+
+/** Main calendar area **/
+:global(.react-datepicker__header) {
```
lmossman, line 46: I saw that there are a few `xClassName` props on the ReactDatePicker component, which we generally prefer over using `:global`, correct? But, I'm guessing those don't give you enough granular control over the styling of each part of the datepicker, so that is why you needed to use `:global` here?

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/Control.tsx
```diff
@@ -66,6 +69,26 @@ export const Control: React.FC<ControlProps> = ({
     );
   }
 
+  if (
+    property.type === "string" &&
+    (property.format === "date-time" || property.format === "date") &&
```
lmossman, line 74: Something I noticed when testing locally was that the `End Date` field in the Facebook Marketing connector did not have a datepicker rendered, and that's because the `end_date` field doesn't have this `format` field set in the spec: https://github.com/airbytehq/airbyte/blob/5354b54c86d4ea12ce4848f5ccdabe0817f1eb79/airbyte-integrations/connectors/source-facebook-marketing/integration_tests/spec.json#L22

So that should be fixed (in a separate PR)

airbyte-webapp/src/components/ui/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,155 @@
...
+      onChange(e.target.value);
+    },
+    [onChange]
+  );
+
+  const localDate = useMemo(() => toEquivalentLocalTime(value), [value]);
+
+  return (
+    <div className={styles.wrapper}>
+      <Input
```
lmossman, line 129: When testing locally, I noticed that if the datepicker is opened by focusing the input, and then a date is clicked in order to fill it in, the UI quickly flashes the `Required` error on the field before going away once it is filled. Here is a screen recording of this:

https://user-images.githubusercontent.com/22731524/204665634-cffbce1d-140f-4ba1-9980-59500dcad7de.mov

This doesn't seem to happen when just clicking on the Datepicker button directly. So my guess is that clicking into the input to focus it, then clicking somewhere in the datepicker popup causes the field to be marked as "touched", and since it doesn't have any value at that point, we show the empty Formik value error (you can reproduce the same situation if you just click on another required input then click away). Then the datepicker actually fills in the field, making it no longer have an error, so it goes away.

Maybe can somehow keep the Input as "focused" or prevent it from being "blurred" to prevent this? Or somehow render the datepicker pop "inside" the Input so that it is not being clicked out of?

lmossman (commented):
A couple small comments and one visual bug I found when testing

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/Control.tsx
timroes: This is not just that field. Quiet a few of connector specs are missing the `format` for date fields. We should try to have all specs double checked and fixed once this is in.

airbyte-webapp/src/components/ui/DatePicker/DatePicker.tsx
timroes: I think we should solve that by making sure that we only trigger the external `onBlur` method if actually the focus left the wrapping element of the date picker. I.e. we should treat focus within that wrapping element as still focusing this component.

josephkmh: Thanks for catching that @lmossman! I took @timroes advice and attached a blur listener to the wrapper div:

```ts
  const handleWrapperBlur = (event: React.FocusEvent<HTMLDivElement>) => {
    if (onBlur && !wrapperRef.current?.matches(":focus-within")) {
      onBlur(event);
    }
  };
```

airbyte-webapp/src/components/ui/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,163 @@
...
+    },
+    [onChange]
+  );
+
+  const localDate = useMemo(() => toEquivalentLocalTime(value), [value]);
+
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  const handleWrapperBlur = (event: React.FocusEvent<HTMLDivElement>) => {
+    if (onBlur && !wrapperRef.current?.matches(":focus-within")) {
+      onBlur(event);
```
timroes, line 133: Small note: the `event` here, will usually be the bubbled up one, so not actually of the `HTMLDivElement` (i.e. the `event.target` will be the actual element that lost focus, not the wrapper `div`). Because that could mean we're passing random elements' blur events around here, I feel we should instead just not pass the `event` forward at all. Instead I'd just recommend we `stopPropagation` here and just call `onBlur` without any event.

airbyte-webapp/src/components/ui/DatePicker/DatePicker.module.scss
josephkmh: Yes, definitely 🙂

airbyte-webapp/src/components/ui/DatePicker/DatePicker.tsx
lmossman: @josephkmh how do you feel about this comment from Tim? I'll hold off on approval and locally testing once more until this is addressed

josephkmh: @timroes thanks! That is indeed the wrong generic for the event we're expecting, and you're right, we don't really need to pass the event at all. I've just removed that.

I don't think stopping blur events from bubbling is the best choice here, though. What benefit would that bring? Seems to me it's better to leave the default browser behavior there.

flash1293:
It seems like error states for missing value and malformed input are not shown:
Screenshot 2022-12-05 at 16 55 25
Screenshot 2022-12-05 at 16 55 29

flash1293:
I understand that it would be a difficult task to show the time in the users local time zone, however what about setting the `timeCaption` property on the picker to "Time (UTC)" to emphasize this? Otherwise I'm concerned this way of setting the time makes it very easy for users to miss they need to convert to UTC before selecting the time:
Screenshot 2022-12-05 at 16 58 10
Screenshot 2022-12-05 at 17 03 20

josephkmh:
> It seems like error states for missing value and malformed input are not shown: Screenshot 2022-12-05 at 16 55 25 Screenshot 2022-12-05 at 16 55 29

@flash1293 looks like you're using the Bing Ads source connector, right? That connector spec unfortunately doesn't specify a `pattern` for the `start_date`:

```json
"reports_start_date": {
  "type": "string",
  "order": 5,
  "title": "Reports replication start date",
  "format": "date",
  "default": "2020-01-01",
  "description": "The start date from which to begin replicating report data. Any data generated before this date will not be replicated in reports. This is a UTC date in YYYY-MM-DD format."
}
```

Which is what we depend on to do regex validation [here](https://github.com/airbytehq/airbyte/blob/316cdd88d6cf99500331f8ea26909ac4c9bc4cee/airbyte-webapp/src/core/jsonSchema/schemaToYup.ts#L59-L61) when generating the yup schema. So any old string can be passed in 😕  Unfortunately quite a few connector specs are missing `pattern` or `format: date | date-time` which means the field won't be validated or the datepicker won't be shown respectively.

flash1293:
Should we auto-add this pattern if it doesn't exist or will it break things? If it's too dangerous to do it, we should probably extend the SAT to cover this situation (out of scope for this PR I'd guess)

lmossman:
Yeah I think the problem with trying to enforce the regex pattern in the frontend if it is missing on the connector is that different connectors may require different timestamp formats, so we can't implement a blanket logic to cover all connectors correctly. Though @sherifnada or @girarda correct me if I'm wrong here - I assume some timestamp formats are API-specific, but if it actually can be made consistent across all APIs that would be great

flash1293:
>  if it actually can be made consistent across all APIs that would be great

+1 for that, I think going with the current logic for now and extending the SAT is a good approach to get to a clean state without breaking things in the meantime.

josephkmh:
@flash1293 thanks for the idea about adding `Time (UTC)`! Just pushed that [here](https://github.com/airbytehq/airbyte/pull/19678/commits/e0226b16fe580e09d646e191eb3ec003aae6b8c9).

lmossman (approved):
Changes LGTM, tested locally once more and it seems to be working as expected. Nice work!

sherifnada:
>  if it actually can be made consistent across all APIs that would be great

agreed. But this will take a little bit longer than we want to merge this PR :) so I think we shouldn't assume the regex on the frontend

PR #19696, opened 2022-11-22
Speed Up Publish OSS Artifacts for Cloud Workflow

git-phu:
## What
Attempt to speed up Publish OSS Artifacts for Cloud Workflow

## How
- Use `gradle test` instead of `gradle build` to skip unnecessary steps before publishing jars
- Do not run the `buildDockerImage` tasks (since this will cause the workflow to build the same images twice).  
    - Since we are using `docker buildx`, just run `copyGeneratedTar` and only have `buildx` actually build images.

Others:
- enable remote gradle caching
- fix workflow concurrency to work as intended
- fix not stopping the runner at the end of the workflow

## Tests

Successful run: https://github.com/airbytehq/airbyte/actions/runs/3520087241
`gw test` build scan: https://scans.gradle.com/s/vfs6gdoaq3ue4/timeline

Actual Speedup:
Not very definitive since this workflow seems to have high variance in run time (can take ~18 minutes or ~28 minutes, not sure what affects this) but it seems like going from `gradle build` -> `gradle test` saves us about ~3 minutes.

Also copying this from a comment below:
> looks when switching from gradle build to gradle test we save some time, but the tests still take up a large amount of the total run time. From the build scan it looks like :airbyte-webapp:test (and its dependencies) are what make the test step take so long

We could also just try merging this in and seeing what run times look like after this change


.github/workflows/publish-oss-for-cloud.yml
```diff
@@ -144,3 +157,29 @@ jobs:
       - name: Cleanup Docker buildx
         run: docker buildx rm oss-buildx
         shell: bash
+
+  stop-runner:
```
davinchia, line 161: oh did we previously not stop runners?

.github/workflows/publish-oss-for-cloud.yml
```diff
@@ -10,6 +12,9 @@ on:
         required: false
 jobs:
   find_valid_pat:
+    concurrency:
```
davinchia, line 15: for my understanding: what is this for?

.github/workflows/publish-oss-for-cloud.yml
```diff
@@ -134,6 +140,13 @@ jobs:
           echo "GIT_REVISION=${GIT_REVISION}" >> $GITHUB_ENV
         shell: bash
 
+      - name: Prepare Runner for Building
+        uses: ./.github/actions/runner-prepare-for-build
+
+      - name: Prepare Docker context
+        run: VERSION=${{ needs.generate-tags.outputs.dev_tag }} SUB_BUILD=PLATFORM ./gradlew copyGeneratedTar airbyte-db:db-lib:copyInitSql
```
davinchia, line 147: maybe good idea to add a comment here saying this is required for buildx? Putting on my reader cap, I'm not sure this is 100% obvious without you and I discussing this yesterday.

.github/workflows/publish-oss-for-cloud.yml
git-phu: yup, seems like a bug that went unnoticed

.github/workflows/publish-oss-for-cloud.yml
git-phu: this is so only one instance of this workflow can be actively running for a specific oss sha (so we don't attempt to push artifacts for the same sha from multiple concurrent workflows)

.github/workflows/publish-oss-for-cloud.yml
git-phu: ah ok sure I'll add a comment

davinchia (approved):
Looks good!
 
Comments are non-blocking. So feel free to merge after they are addressed.

I'm not sure I see where the gradle remote cache is enabled in the changeset. Can you point me to it for my knowledge?

.github/workflows/publish-oss-for-cloud.yml
davinchia: ah ok! can we also add a comment here?

git-phu: added comments

.github/workflows/publish-oss-for-cloud.yml
git-phu: added comments

.github/workflows/publish-oss-for-cloud.yml
```diff
@@ -1,6 +1,8 @@
 name: Publish OSS Artifacts for Cloud
-concurrency:
-  group: ${{ github.workflow }}-${{ inputs.oss_ref || github.sha }}
+
+env:
+  S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.SELF_RUNNER_AWS_ACCESS_KEY_ID }}
+  S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.SELF_RUNNER_AWS_SECRET_ACCESS_KEY }}
```
git-phu, line 5: @davinchia these env vars enable gradle remote caching to work properly

davinchia: ah ok!

.github/workflows/publish-oss-for-cloud.yml
```diff
@@ -1,6 +1,8 @@
 name: Publish OSS Artifacts for Cloud
-concurrency:
-  group: ${{ github.workflow }}-${{ inputs.oss_ref || github.sha }}
+
+env:
+  S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.SELF_RUNNER_AWS_ACCESS_KEY_ID }}
```
cpdeethree, line 4: Maybe a comment is nice here that says something along the lines of 

//these environment variables are necessary for gradle cache task

git-phu:
forgot to merge this in after the code freeze but I'll do so now

One more verification that it's still working: https://github.com/airbytehq/airbyte/actions/runs/3699244652

Issue #19669, opened 2022-11-21
Consolidate update/delete custom definitions to regular update/delete definitions


xiaohansong:
FE should use the same API when user wants to update or delete their source or destination definition.

This ticket requires to consolidate 4 custom APIs into the original API (FE will use the one without `custom` in it, such as updateSourceDefinition)

Also as part of the effort we should remove the legacy creation path to avoid misuse:

POST /v1/source_definitions/create
POST /v1/destination_definitions/create

pmossman:
Refining notes:
- Before removing any routes, we should check to see if there are other places that call them. For instance, does Octavia CLI call `POST /v1/source_definitions/create` or `POST /v1/destination_definitions/create`? @alafanechere perhaps you know the answer to this?

marcelopio:
octavia do not call it yet, and my implementation was using the custom prefix

https://github.com/airbytehq/airbyte/pull/15058

pmossman:
Closing this as a duplicate in favor of https://github.com/airbytehq/airbyte/issues/19670

PR #19702, opened 2022-11-22
🪟 🐛 Fix custom connection creation endpoint

### LINKED ISSUE #19637: Differentiate between custom and non-custom connectors when updating
[PR 19221](https://github.com/airbytehq/airbyte/pull/19221) updated the following API methods in the webapp globally:

1. `createSourceDefinition()` => `createCustomSourceDefinition()`
2. `createDestinationDefinition()` => `createCustomDestinationDefinition()`
3. `updateDestinationDefinition()` => `updateCustomDestinationDefinition()`
4. `updateSourceDefinition()` => `updateCustomSourceDefinition()`

While 1. and 2. were correct, 3. and 4. were causing OSS users to no longer be able to update connectors that are not custom.

This resulted in an [oncall issue](https://github.com/airbytehq/oncall/issues/1078). A targeted revert PR was introduced ([PR 19627](https://github.com/airbytehq/airbyte/pull/19627)) that reverted all 4 methods above. This means:

1. Non-custom connectors can now be updated again (good!)
2. New custom connectors are now being created with the old methods again (impact unclear, probably a new migration similar to the [first one](https://github.com/airbytehq/airbyte/blob/master/airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_40_18_004__BackfillActorDefinitionWorkspaceGrant.java) will be needed to update these)
3. Custom connectors created before the revert commit are now being updated with the old methods (unclear whether this works, needs to be tested)

Relates to https://github.com/airbytehq/airbyte/issues/9652

joeykmh:
## What
Closes https://github.com/airbytehq/airbyte/issues/19637

PR that introduced this bug was a precision revert of another issue: https://github.com/airbytehq/airbyte/pull/19627

This PR removes the `createSourceDefinition()` and `createDestinationDefinition()` methods in favor of the new `createCustomSourceDefinition()` and `createCustomDestinationDefinition()` methods. `createSourceDefinition()` and `createDestinationDefinition()` are to be deprecated.

This PR does not address _updating_ custom connectors. There is a backend issue for fixing this behavior here: https://github.com/airbytehq/airbyte/issues/19669


timroes (approved):
Code LGTM in accordance what we've discussed and outlined as a plan. Did not test locally, since the BE fix isn't in yet, so I'd anyway expect weird behavior still.

krishnaglick (commented):
So the backend needs an update to correctly handle custom vs not? Because this looks like we're always creating custom now which was the issue before.

josephkmh:
@krishnaglick the issue before was actually only related to updating: non-custom connectors needed to be updated with `POST /v1/destination_definitions/update` while custom connectors needed to be updated with `POST /v1/destination_definitions/update_custom`

Custom connectors should always be created with `POST /v1/destination_definitions/create_custom`, and `POST /v1/destination_definitions/create` should no longer be used in the webapp at all.

krishnaglick:
> @krishnaglick the issue before was actually only related to updating: non-custom connectors need to be updated with `POST /v1/destination_definitions/update` while custom connectors need to be updated with `POST /v1/destination_definitions/update_custom`. A separate backend ticket will fix this, where `POST /v1/destination_definitions/update` will update a connector whether it is custom or not.
> 
> Custom connectors should always be created with `POST /v1/destination_definitions/create_custom`, and `POST /v1/destination_definitions/create` should no longer be used in the webapp at all.

My confusion stems from `DestinationDefinitionService` no longer having a non-custom endpoint as it was completely removed for the custom one. Is the codepath there always custom?

josephkmh:
> My confusion stems from DestinationDefinitionService no longer having a non-custom endpoint as it was completely removed for the custom one. Is the codepath there always custom?

Yes exactly - creating a connector via the UI is now always a "custom" connector which is scoped to the current workspace.

krishnaglick (approved):
Have not tested locally, but code looks good.

PR #19712, opened 2022-11-22
CDK Lowcode: allow parametrizing initial page when using PageIncrement

### LINKED ISSUE #20296: CDK Lowcode: allow parametrizing initial page when using PageIncrement
## What
PageIncrement starts counting pages from 0 which is not always a good choice

davydov-d:
## What
PageIncrement starts counting pages from 0 which is not always a good choice 

## How
Allow configuring the number of the start page

Use case: pagination starts from 1 in [Zenloop](https://github.com/airbytehq/airbyte/pull/19624), Delighted(work in progress)


girarda (approved):

airbyte-cdk/python/setup.py
```diff
@@ -15,7 +15,7 @@
 
 setup(
     name="airbyte-cdk",
-    version="0.9.5",
+    version="0.9.6",
```
brianjlai, line 18: since this is a new backwards compatible feature not a bug, we should increment the minor version

brianjlai (approved):
looks good to me! one note about the minor version increment instead of patch

grubberr:
@davydov-d 
do we wait Monday 28.11 here ?

PR #19713, opened 2022-11-22
🪟 🎉 new streams table edit modal namespace and stream

### LINKED ISSUE #17831: Update new streams table to edit namespace and stream name options with modals
1. Add modals for editing stream name and namespaces
2. Remove options from connection (?) and replication settings

Prototype - Click on the gear icon next to namespace and stream names on the streams table

https://www.figma.com/proto/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?node-id=865%3A52984&scaling=min-zoom&page-id=857%3A24319&starting-point-node-id=865%3A52984&show-proto-sidebar=1

m19berger:
## What
Closes [#17831](https://github.com/airbytehq/airbyte/issues/17831)

## How

- Added new modal window "Destination namespace" and "Destination stream names"
- The previous implementation is hidden by a local variable 
- Fixed bug with Tooltip (block-level element inside inline element)

## Loom
https://www.loom.com/share/61713527a7334a61af9c05a067ae9114


airbyte-webapp/src/components/connection/CatalogTree/next/DestinationNamespaceModal.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+  return (
+    <Formik
+      initialValues={{
+        namespaceDefinition: initialValues?.namespaceDefinition ?? NamespaceDefinitionType.source,
+        namespaceFormat: initialValues.namespaceFormat,
+      }}
+      enableReinitialize
+      validateOnBlur
+      validateOnChange={false}
```
dizel852, line 247: Any reason why we don't wanna validate on change?) 😃 
Since I noticed even after we've entered the valid input, the forms remain invalid, and only after the `onBlur` event occurs, the form become valid.
IMO we need to give the user faster feedback, don't make him think about what to do next.
The same question for 
airbyte-webapp/src/components/connection/CatalogTree/next/DestinationStreamNamesModal.tsx

dizel852 (approved):
Code LGTM 👍 
Also tested locally - have not found any issues.
Left some not significant comments, except `Formik` `onChange` behavior.
But need also another review

@matter-q could you please create an issue on missed doc link for "Learn more" and mention it in this PR to keep it trackable. Thanks!

@edmundito could you please remind whom should we ping regarding doc links?

airbyte-webapp/src/components/connection/CatalogTree/next/DestinationNamespaceModal.tsx
```diff
@@ -0,0 +1,346 @@
...
+    switch (namespaceDefinition) {
+      case NamespaceDefinitionType.source:
+        return <FormattedMessage id="connectionForm.modal.destinationNamespace.option.source.description" />;
+      case NamespaceDefinitionType.destination:
+      case NamespaceDefinitionType.customformat:
+        return <FormattedMessage id="connectionForm.modal.destinationNamespace.option.destination.description" />;
+    }
+  };
+
+  const getExampleTableData = (namespaceDefinition: NamespaceDefinitionType): ExampleSettingsTableProps => {
```
edmundito, line 89: This could be extracted out of the component function and into its own, same for `getInfo`

dizel852:
@matter-q Could you please extract the structure of the tables to separate files?

airbyte-webapp/src/components/connection/CatalogTree/next/DestinationNamespaceModal.tsx
```diff
@@ -0,0 +1,346 @@
...
+import { FormikConnectionFormValues } from "../../../../views/Connection/ConnectionForm/formConfig";
+import LabeledRadioButton from "../../../LabeledRadioButton";
+import styles from "./DestinationNamespaceModal.module.scss";
+
+interface ExampleSettingsTableProps {
+  columns: Array<{ id: string; displayName: string }>;
+  data: Array<Record<string, string>>;
+}
+
+const ExampleSettingsTable: React.FC<ExampleSettingsTableProps> = ({ columns, data }) => {
```
edmundito, line 21: This should be extracted to its own file along with the example table data generation. It would make this file easier to maintain

edmundito (changes requested):
Tested. Verified that the tooltips still look good in other areas.

One thing I'd like see is move the modals out of the `connection/CatalogTree` folder and instead provide their own folders (`connection/DestinationNamespaceModal/` and `connection/DestinationStreamNamesModal/`) This would allow the files more easier to maintain since the `CatalogTree` has  too many files.

edmundito:
One more thing I could not figure out from the review: When I open the namespace modal, I see a loading indicator in the catalog tree table. Would you know why that is happening?

edmundito:
Also, some tests are failing, it looks like the snapshots need to be updated.

airbyte-webapp/src/components/connection/CatalogTree/next/DestinationNamespaceModal.tsx
```diff
@@ -0,0 +1,346 @@
...
+import { FormattedMessage, useIntl } from "react-intl";
+import * as yup from "yup";
+
+import { Button } from "components/ui/Button";
+import { Input } from "components/ui/Input";
+import { ModalBody, ModalFooter } from "components/ui/Modal";
+import { Text } from "components/ui/Text";
+
+import { NamespaceDefinitionType } from "../../../../core/request/AirbyteClient";
+import { FormikConnectionFormValues } from "../../../../views/Connection/ConnectionForm/formConfig";
```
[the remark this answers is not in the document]
matter-q: Fixed! Thanks :slightly_smiling_face:

airbyte-webapp/src/components/connection/CatalogTree/next/DestinationNamespaceModal.tsx
matter-q: Good idea. Thanks :slightly_smiling_face:

matter-q:
> One more thing I could not figure out from the review: When I open the namespace modal, I see a loading indicator in the catalog tree table. Would you know why that is happening?

I found the problem with this behavior. Buttons in the header that open new modals trigger submit an event for the parent form (saveConnection). Fixed by attribute `type="button"`

airbyte-webapp/src/components/connection/CatalogTree/next/DestinationNamespaceModal.tsx
matter-q: But I have a dependency with `formatMessage` for those functions :worried:

airbyte-webapp/src/components/connection/CatalogTree/next/DestinationNamespaceModal.tsx
dizel852: Verified - fixed

dizel852 (approved):
Did another review after @edmundito 's comments, seems like all mentioned issues are fixed. 
But some questions still are opened: https://github.com/airbytehq/airbyte/pull/19713#discussion_r1035454520
Let's wait for reply.

edmundito (approved):

PR #19717, opened 2022-11-22
Parallel attempt for e2e

krishnaglick:
Enable parallel for cypress tests


cpdeethree:
This did improve wall clock time of E2E tests by 3 minutes. I approved this, and once we do bump the machine sizes we will realize the gain, but can go ahead and mainline this change now.

c-p-b (approved):

krishnaglick:
Based on the Cypress docs, we shouldn't parallelize unless we can setup multiple machines.

krishnaglick:
> > Based on the Cypress docs, we shouldn't parallelize unless we can setup multiple machines.
> 
> Interesting - maybe that's why we didn't see much improvement from turning it on here.
> 
> That's probably part of a longer story about having workflows spawn up multiple runners themselves - currently the pattern is one runner per workflow.

Is there an epic or ticket to add that functionality?

PR #19719, opened 2022-11-22
deprecate config repository use in web backend

cgardens:
## What
* it was a mistake that `ConfigRepository` was ever added to the web backend. 
* it is now now not easy to remove, but we need to more clearly signal to people that they should not use it for new cases (which keeps happening).
* add java doc explaining the purpose of the webbackend.

## How
* rename the field to make it clear that it shouldn't be used.
* deprecate the field

### DIFF
--- a/airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
+++ b/airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
@@ -68,6 +68,11 @@ import java.util.stream.Collectors;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
+/**
+ * The web backend is an abstraction that allows the frontend to structure data in such a way that
+ * it is easier for a react frontend to consume. It should NOT have direct access to the database.
+ * It should operate exclusively by calling other endpoints that are exposed in the API.
+ */
 @AllArgsConstructor
 @Slf4j
 public class WebBackendConnectionsHandler {
--- a/airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
+++ b/airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
@@ -80,14 +80,16 @@ public class WebBackendConnectionsHandler {
   private final SchedulerHandler schedulerHandler;
   private final OperationsHandler operationsHandler;
   private final EventRunner eventRunner;
+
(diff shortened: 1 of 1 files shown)


airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
```diff
@@ -68,6 +68,11 @@
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
+/**
+ * The web backend is an abstraction that allows the frontend to structure data in such a way that
+ * it is easier for a react frontend to consume. It should NOT have direct access to the database.
+ * It should operate exclusively by calling other endpoints that are exposed in the API.
```
benmoriceau, line 74: I think that the last sentence is missleading because we are not calling other endpoints but other handler. Moving the calls we do to the repository to another handler will still add a direct call to the DB. If we want to call the other endpoint through this handler we should use the generated clients via open API (like `io.airbyte.api.client.generated.HealthApi`). I don't think that calling the repository in this class will prevent us from having the server to be the gateway of all of our DB access since this handler is not use outside of the server module.

airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
```diff
@@ -80,14 +85,16 @@ public class WebBackendConnectionsHandler {
   private final SchedulerHandler schedulerHandler;
   private final OperationsHandler operationsHandler;
   private final EventRunner eventRunner;
+
   // todo (cgardens) - this handler should NOT have access to the db. only access via handler.
-  private final ConfigRepository configRepository;
+  @Deprecated
```
benmoriceau, line 90: The main issue I saw when trying that is that IntelliJ don't give feedback about this deprecated (No strike through or red color). That's why I went with an abstract class and getter and setter since using this field will look like an error. That being say, the new naming is pretty explicit so I'm fine with both solution.

benmoriceau (approved):
LGTM, I am just thinking that even if we are removing the repositiory, we will still do direct call to the DB in this class. It will just be hidden in the call to the handler.

airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
cgardens: @benmoriceau yeah. i noticed it too. let's merge your PR since it gets the highlighting right. i'll merge mine after with the comment and the name change

cgardens:
done elsewhere.

PR #19240, opened 2022-11-09
Data types update: Implement protocol message migrations

edgao:
closes https://github.com/airbytehq/airbyte/issues/17639 and https://github.com/airbytehq/airbyte/issues/18254

In most catalogs, this is a one-to-one swap (i.e. detect an existing schema and replace it with an updated schema). The one exception is in `type: [foo, bar]` schema declarations, which are now expanded to `oneOf` declarations.

`mutateSchemas` is written slightly generically, because I plan to use it in the downgrade path as well.

Reading order:
* V0ToV1MigrationTest.java
* TODO other stuff after rebasing against https://github.com/airbytehq/airbyte/pull/20745

PR #19721, opened 2022-11-22
🎉Updated normalization to handle new datatypes

### LINKED ISSUE #12658: Destination BigQuery: map "airbyte_type": "big_number" to BIGNUMERIC
Due to new updates in the building of JSON schema, a new property `"airbyte_type"` was added to the field description. 
One of the possible values for `"airbyte_type"` is `"big_number"`. Now `"airbyte_type"` is ignored and all `"big_number"` fields are mapped as `"STRING"` in destination BigQuery. 
According to #11090 and this [comment](https://github.com/airbytehq/airbyte/issues/11090#issuecomment-1119113122) all fields with `"airbyte_type": "big_number"` should be mapped as `BIGNUMERIC` in destination BigQuery. 

**Input JSON schema example:**
```
{
  "type": "object",
  "properties": {
    "type_big_number": {"type": "string", "airbyte_type": "big_number"},
  }
}
```
**Input data example:**
```
{
  "type_big_number": "1000000000000000000000000000000000.1234"
}
```

### LINKED ISSUE #17642: Update normalization to handle the new `$ref` data types (including binary data)
https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py needs to be updated to check for `$ref` values instead of `type`/`airbyte_type`/`format`. This updated version of normalization will never see old-style catalogs, because it will be released after the platform protocol versioning migrations, and normalization is released as part of platform (i.e. it's impossible for someone to upgrade normalization without also upgrading platform).

Normalization also should add handling for BinaryData. A minimal implementation is to just treat this as string; a less-minimal implementation is to use destination-specific `BINARY` columns.

etsybaev:
## What
Airbyte introduces a new schema with centralized "wellKnownDatatypes" (https://github.com/airbytehq/airbyte/blob/master/airbyte-protocol/protocol-models/src/main/resources/airbyte_protocol/well_known_types.yaml). 

Note: this cannot be merged until https://github.com/airbytehq/airbyte/pull/19240 is merged

## How
1. Updated normalization to handle new datatypes
2. Added processing for newly introduced Binary datatype to save it as binary object. Only Oracle and Redshift handle it, but save as a base64 string

## Recommended reading order
1. All python files
3. All .json and .txt files (updated schemas and test messages for unit and integration tests)
4. dbt-project-template/macros/cross_db_utils/datatypes.sql
5. other ".sql" and ".yaml" files

## 🚨 User Impact 🚨
Breaking changes! Need to merge only after logic for schema upgrade once connectors would be able to work with new datatypes

## Pre-merge Checklist

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>

</details>

## Tests

</details>

</details>

</details>


etsybaev:
Locally all integration and unit tests have passed
Selection_130

airbyte-integrations/bases/base-normalization/integration_tests/resources/redshift_normalization_migration/destination_catalog.json
```diff
@@ -8,22 +8,22 @@
           "$schema": "http://json-schema.org/draft-07/schema#",
           "properties": {
             "id": {
-              "type": ["null", "integer"]
+              "$ref" : "WellKnownTypes.json#/definitions/Integer"
```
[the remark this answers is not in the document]
etsybaev: @grubberr the answer is yes for both questions. Please check out this ticket's description and useful links from its description https://github.com/airbytehq/airbyte/issues/17642

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
```diff
@@ -16,76 +17,125 @@ def remove_jinja(command: str) -> str:
...
+    for prop in one_of_property:
+        plain_datatypes_list.append(prop[data_type.REF_TYPE_VAR_NAME])
+    return plain_datatypes_list
+
+
 def is_string(property_type) -> bool:
-    return property_type == "string" or "string" in property_type
+    if data_type.ONE_OF_VAR_NAME in property_type and data_type.ONE_OF_VAR_NAME in property_type:
+        plain_datatypes_list = get_plain_list_from_one_of_array(property_type)
+        return plain_datatypes_list == data_type.STRING_TYPE or data_type.STRING_TYPE in plain_datatypes_list
```
grishick, line 32: The [integration test](https://github.com/airbytehq/airbyte/actions/runs/3614412995/jobs/6090834357) seems to be complaining about this line
```
BUILD FAILED in 26s
[python] .venv/bin/python -m mypy normalization --config-file /actions-runner/_work/airbyte/airbyte/pyproject.toml
	 normalization/transform_catalog/utils.py: note: In function "is_string":
	 normalization/transform_catalog/utils.py:32:16: error: Non-overlapping equality
	 check (left operand type: "List[Any]", right operand type: "str") 
	 [comparison-overlap]
	             return plain_datatypes_list == data_type.STRING_TYPE or data_t...
	                    ^
```

grishick (changes requested):
1. need to make sure normalization tests are passing before publishing and merging
2. I think we should keep tests for old format until it is completely deprecated

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
etsybaev: Hi @grishick 
I believe that integration tests on CI fail because It generates thousands of files, but I haven't committed autogenerated files yet to make the review process easier.  Obviously, I will have to make them pass before merging :) 

Just to make sure, do you want me to make it works for both old and new protocols for a while? I may do that, but it would take me lots of additional time.  The original ticket says: "This updated version of normalization will never see old-style catalogs, because it will be released after the platform protocol versioning migrations, and normalization is released as part of platform (i.e. it's impossible for someone to upgrade normalization without also upgrading platform).". 

So I didn't spend time for keep supporting old-style also.

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/stream_processor.py
```diff
@@ -574,7 +596,40 @@ def cast_property_type(self, property_name: str, column_name: str, jinja_column:
...
             replace_operation = jinja_call(f"empty_string_to_null({jinja_column})")
             return f"cast({replace_operation} as {sql_type}) as {column_name}"
-        elif is_string(definition["type"]):
+        elif (data_type.REF_TYPE_VAR_NAME in definition and is_binary_datatype(definition[data_type.REF_TYPE_VAR_NAME])) or (
+            data_type.ONE_OF_VAR_NAME in definition and is_binary_datatype(definition)
+        ):
+            if self.destination_type.value == DestinationType.POSTGRES.value:
+                # sql_type = "bytea"
+                sql_type = jinja_call("type_binary()")
+                return f"cast(decode({column_name}, 'base64') as {sql_type}) as {column_name}"
```
edgao, line 605: maybe worth defining a `decode_base64` macro, to move the destination-specific logic out of this file. Then if we make the `type_binary()` macro return `dbt_utils.type_string()` for destinations without a binary type, then we don't need so much destination-specific logic here.

I.e. we could do
```python
decode_call = jinja_call(f"decode_base64({column_name})")
return f"cast({decode_call} as {sql_type}) as {column_name}"
```
for all destination types

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/stream_processor.py
```diff
@@ -574,7 +596,40 @@ def cast_property_type(self, property_name: str, column_name: str, jinja_column:
...
+                sql_type = jinja_call("type_binary()")
+                return f"CAST({column_name} as XML ).value('.','{sql_type}') as {column_name}"
+            elif self.destination_type.value == DestinationType.SNOWFLAKE.value:
+                # sql_type = "VARBINARY"
+                sql_type = jinja_call("type_binary()")
+                return f"cast(BASE64_DECODE_BINARY({column_name}) as {sql_type}) as {column_name}"
+            elif self.destination_type.value == DestinationType.CLICKHOUSE.value:
+                # sql_type = "VARBINARY"
+                sql_type = jinja_call("type_binary()")
+                trimmed_column_name = f"trim(BOTH '\"' from {column_name})"
```
edgao, line 625: what does this do?

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/stream_processor.py
```diff
@@ -584,11 +639,27 @@ def cast_property_type(self, property_name: str, column_name: str, jinja_column:
...
                 sql_type = f"{sql_type}(1024)"
         else:
-            print(f"WARN: Unknown type {definition['type']} for column {property_name} at {self.current_json_path()}")
+            if data_type.REF_TYPE_VAR_NAME in definition:
+                print(
+                    f"WARN: Unknown ref type {definition[data_type.REF_TYPE_VAR_NAME]} for column {property_name} at {self.current_json_path()}"
+                )
+            elif data_type.ONE_OF_VAR_NAME in definition:
+                print(
+                    f"WARN: Unknown onOf simple type {definition[data_type.ONE_OF_VAR_NAME]} for column {property_name} at {self.current_json_path()}"
```
edgao, line 648:
```suggestion
                    f"WARN: Unknown oneOf simple type {definition[data_type.ONE_OF_VAR_NAME]} for column {property_name} at {self.current_json_path()}"
```

edgao (commented):
took a first pass at this. Definitely a complicated PR! I think the general approach  makes sense; I just had some questions about specific implementation bits.

Also had a comment for a potential refactor - totally reasonable to not want to do that since this is already a big change, but I think it would simplify the logic overall.

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
grishick: Yes, my reasoning for supporting both versions was just as a fallback in case something weird happens. However, I think that's not necessary, so I am OK with not supporting the older version. If somehow an older platform ends up loading new normalization image it'll fail, but that's not a reasonable scenario given how we manage normalization version right now

edgao: it'll become more of a possibility, since normalization version is moving into destination_definitions.yaml. I don't think there's any safeguard against downgrading normalization to a pre-protocol-v1 version other than "PR reviewer is careful"

though maybe there should be such a safeguard

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/stream_processor.py
```diff
@@ -456,13 +466,15 @@ def extract_json_column(property_name: str, json_column_name: str, definition: D
...
+        if data_type.REF_TYPE_VAR_NAME in definition or data_type.TYPE_VAR_NAME in definition or data_type.ONE_OF_VAR_NAME in definition:
+            if data_type.TYPE_VAR_NAME in definition and is_array(definition[data_type.TYPE_VAR_NAME]):
                 json_extract = jinja_call(f"json_extract_array({json_column_name}, {json_path}, {normalized_json_path})")
-                if is_simple_property(definition.get("items", {"type": "object"})):
+                if is_simple_property(definition.get("items", {data_type.TYPE_VAR_NAME: "object"})):
                     json_extract = jinja_call(f"json_extract_string_array({json_column_name}, {json_path}, {normalized_json_path})")
-            elif is_object(definition["type"]):
+            elif data_type.TYPE_VAR_NAME in definition and is_object(definition[data_type.TYPE_VAR_NAME]):
                 json_extract = jinja_call(f"json_extract('{table_alias}', {json_column_name}, {json_path}, {normalized_json_path})")
+            elif data_type.REF_TYPE_VAR_NAME in definition and (is_date(definition) or is_time(definition) or is_datetime(definition)):
```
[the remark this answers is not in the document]
etsybaev: These were there before my changes. I just adopted it to handle new ref case

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/stream_processor.py
etsybaev: This is how clickhouse works. It complains about " symbols in value. So need to trim it explicitly.

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
etsybaev: The problem here is that there are also lots of other places where the old "type" were used, not only in case method. To find all cases I had to do dozens run in debug mode for different connectors and check what actually comes here. Some of the methods also appear to be called in very unpredictable ways. So adding backward compatibility will take me some time (can't estimate it yet), as for now I will not know if the received "type" is an object or still should be treated as simple object. Also will need to dive deeper in test sqls. As not my goal will be not only make them pass, but also to extend to support versioning.
By the way, locally all tests pass, maybe something is not passed on Ci as I haven't committed autogenerated files to make the review process easier.

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
```diff
@@ -16,76 +17,125 @@ def remove_jinja(command: str) -> str:
...
+    else:
+        return property_type == data_type.STRING_TYPE or data_type.STRING_TYPE in property_type
+
+
+def is_binary_datatype(property_type) -> bool:
+    if data_type.ONE_OF_VAR_NAME in property_type and data_type.ONE_OF_VAR_NAME in property_type:
+        plain_datatypes_list = get_plain_list_from_one_of_array(property_type)
+        return plain_datatypes_list == data_type.BINARY_DATA_TYPE or data_type.BINARY_DATA_TYPE in plain_datatypes_list
+    else:
+        return property_type == data_type.BINARY_DATA_TYPE or data_type.BINARY_DATA_TYPE in property_type
```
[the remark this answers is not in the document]
etsybaev: This was just a generic verification to check for object. As depends on the place where this is called, It may be a single object, it may be a dict, or it may be inside of oneOf. But if we have an assumption that binary datatype could not be a part of oneOf (as per your comment above) - then yes,  this may be simplified.

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
grishick: Note from grooming session: we don't need to keep support for the old protocol version in normalization. The only  breaking scenario is old platform + new normalization, which is not possible unless someone deliberately updates hard-coded normalization version w/o updating the platform.

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
```diff
@@ -16,76 +17,125 @@ def remove_jinja(command: str) -> str:
     return str(command).replace("{{ ", "").replace(" }}", "")
 
 
+# transform oneOf nested list ot plain types list
+def get_plain_list_from_one_of_array(one_of_definition) -> list:
+    one_of_property = one_of_definition[data_type.ONE_OF_VAR_NAME]
+    plain_datatypes_list = list()
+    for prop in one_of_property:
+        plain_datatypes_list.append(prop[data_type.REF_TYPE_VAR_NAME])
```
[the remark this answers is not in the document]
etsybaev: @edgao In this case, it will take a String type by already existing logic. If the string is in the list, then other types would be ignored. As far as I see from Normalization logic -  it always takes a wider type for SQL column-> string

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
```diff
@@ -16,76 +17,125 @@ def remove_jinja(command: str) -> str:
...
+    if data_type.ONE_OF_VAR_NAME in property_type and data_type.ONE_OF_VAR_NAME in property_type:
+        plain_datatypes_list = get_plain_list_from_one_of_array(property_type)
+        return plain_datatypes_list == data_type.STRING_TYPE or data_type.STRING_TYPE in plain_datatypes_list
+    else:
+        return property_type == data_type.STRING_TYPE or data_type.STRING_TYPE in property_type
+
+
+def is_binary_datatype(property_type) -> bool:
+    if data_type.ONE_OF_VAR_NAME in property_type and data_type.ONE_OF_VAR_NAME in property_type:
+        plain_datatypes_list = get_plain_list_from_one_of_array(property_type)
```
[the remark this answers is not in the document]
etsybaev: Updated, thanks

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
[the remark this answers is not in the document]
etsybaev: @edgao This is what I mean by saying that some of the methods may be called from different places and sometimes passing a simple object and sometimes a "dict". This was also mindblowing for me after strict types in java :) 


Here is an example where it's passed like just a string. That is why some checks may look redundant at the first glance

Selection_139

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
```diff
@@ -16,76 +17,125 @@ def remove_jinja(command: str) -> str:
...
 def is_string(property_type) -> bool:
-    return property_type == "string" or "string" in property_type
+    if data_type.ONE_OF_VAR_NAME in property_type and data_type.ONE_OF_VAR_NAME in property_type:
+        plain_datatypes_list = get_plain_list_from_one_of_array(property_type)
+        return plain_datatypes_list == data_type.STRING_TYPE or data_type.STRING_TYPE in plain_datatypes_list
+    else:
+        return property_type == data_type.STRING_TYPE or data_type.STRING_TYPE in property_type
+
+
+def is_binary_datatype(property_type) -> bool:
```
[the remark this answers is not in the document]
etsybaev: Sometimes this may a simple string. Sometimes dict, sometimes string :( 
Selection_139

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/stream_processor.py
etsybaev: updated, thanks

etsybaev:
Integration tests passed locally
Selection_144

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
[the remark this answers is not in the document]
edgao: hmmmmmmmmmmmmmmmm

how hard would it be to make all of these methods _always_ expect a dict? It feels like this is causing us a lot of confusion :/ and this seems like a good opportunity to clean up the code a little bit

grishick:
LGTM, although I am confused about the discussion around `property_type`

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
```diff
@@ -17,100 +17,88 @@ def remove_jinja(command: str) -> str:
...
     return is_datetime_with_timezone(definition) or is_datetime_without_timezone(definition)
 
 
 def is_datetime_without_timezone(definition: dict) -> bool:
-    if data_type.ONE_OF_VAR_NAME in definition and data_type.ONE_OF_VAR_NAME in definition:
-        plain_datatypes_list = get_plain_list_from_one_of_array(definition)
-        return (
-            plain_datatypes_list == data_type.TIMESTAMP_WITHOUT_TIMEZONE_TYPE
-            or data_type.TIMESTAMP_WITHOUT_TIMEZONE_TYPE in plain_datatypes_list
+    if data_type.ONE_OF_VAR_NAME in definition:
```
edgao, line 36: should all the `is_whatever_type` methods have this oneof handling?

maybe worth doing something like this, to reduce the amount of copypasta (I didn't test this at all, hopefully it's clear enough. Also maybe would be more readable using `lambda` syntax in `get_reftype_function`, idk)
```python
def is_type_included(definition: dict, is_type: Callable[[dict], bool]) -> bool:
    if data_type.ONE_OF_VAR_NAME in definition:
        return bool(
            any(
                is_type(option)
                for option in definition[data_type.ONE_OF_VAR_NAME]
            )
        )
    else:
        return is_type(definition)

def get_reftype_function(type: string) -> Callable[[dict], bool]:
    def is_reftype(definition: dict) -> bool:
        return data_type.REF_TYPE_VAR_NAME in definition and type == definition[data_type.REF_TYPE_VAR_NAME]
    return is_reftype

def is_binary_datatype(definition: dict) -> bool:
    return is_type_included(definition, get_reftype_function(data_type.BINARY_DATA_TYPE))

def is_datetime_with_timezone(definition: dict) -> bool:
    return is_type_included(definition, get_reftype_function(data_type. TIMESTAMP_WITHOUT_TIMEZONE_TYPE))

# etc
```

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
```diff
@@ -119,21 +107,22 @@ def is_big_integer(definition: dict) -> bool:
 
 
 def is_long(property_type, definition: dict) -> bool:
```
edgao, line 109: can we remove the `property_type` param? Since it's redundant with the `definition` param

edgao (commented):
Added some more comments around code structure - feel free to say I'm increasing the scope too much :) 

Can you commit the generated files for `airbyte-integrations/bases/base-normalization/integration_tests/normalization_test_output/postgres`? That should be a representative sample for what's actually changing.

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
etsybaev: I've seen these methods are also called in other parts of code. So even a simple change may cause unexpected side effects and I've seen it many times during my work on this PR. So since we may have one not only in ref types I really afraid to touch it :)

etsybaev:
@grishick I think that the discussion above related to property_type is not more relevant since I've re-implemented this part of code to fix static code check gradle tasks. Please correct me if I'm wrong

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
etsybaev: @edgao this method may be called in different places. for some of them it's important to still have this `property_type`

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
edgao: I think this code change is transparent though? I.e. it keeps the same method signature, just changes the implementation to be consistent across methods. So other than the methods that take both `definition` and `property_type`, it shouldn't matter?

mostly just want to doublecheck whether the `oneOf` handling is intentionally excluded from some methods.

edgao (commented):
I completely forgot about this earlier - it would be good to add a test case for the binary data stuff, since there's some potential weird things that could happen. I.e.:
1. add a binary data field to https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-normalization/integration_tests/resources/test_simple_streams/data_input/catalog.json#L269
2. populate that column in https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-normalization/integration_tests/resources/test_simple_streams/data_input/messages.txt#L63
3. add the expected values to https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-normalization/integration_tests/resources/test_simple_streams/dbt_test_config/dbt_data_tests_tmp/types_testing_incorrect_values.sql

Looks like the generated diff is _mostly_ unrelated to this PR (e.g. adding the SCD drop table hook, fixing schema quoting). The stuff that _is_ related looks correct (i.e. adding `property_binary_data` and parsing it).

So once we have some good tests for binary data values, I think this will be good to go! (but still can't be merged until all the platform-side stuff is ready)

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/utils.py
etsybaev: @Ohh...got your point about Callable. Updated. Also updated methods to get rid of unused property_types. Thanks

etsybaev:
> I completely forgot about this earlier - it would be good to add a test case for the binary data stuff, since there's some potential weird things that could happen. I.e.:
> 
> 1. add a binary data field to https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-normalization/integration_tests/resources/test_simple_streams/data_input/catalog.json#L269
> 2. populate that column in https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-normalization/integration_tests/resources/test_simple_streams/data_input/messages.txt#L63
> 3. add the expected values to https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-normalization/integration_tests/resources/test_simple_streams/dbt_test_config/dbt_data_tests_tmp/types_testing_incorrect_values.sql
> 
> Looks like the generated diff is _mostly_ unrelated to this PR (e.g. adding the SCD drop table hook, fixing schema quoting). The stuff that _is_ related looks correct (i.e. adding `property_binary_data` and parsing it).
> 
> So once we have some good tests for binary data values, I think this will be good to go! (but still can't be merged until all the platform-side stuff is ready)

I've already added it to simple streams here:
https://github.com/airbytehq/airbyte/pull/19721/files#diff-e9c133b3274f5bb507241a922f41dc5eac40943e74dda9cc2ea7fec336f8ea7cR49

https://github.com/airbytehq/airbyte/pull/19721/files#diff-a4a03c2f52f9f77cd8da3d7e8ebe4c3f5102f18b2fd42fc924eabdf3e5427d3dR1


But will have a look at that possible "weird things that could happen." :) thanks

edgao:
ah maybe I just missed it - the links won't load because github is dumb :( can you link to the files directly?

the main thing is that only the `types_testing` stream actually verifies values - the other streams just verify record count, so if the data is written incorrectly, it'll still pass the test. That's the purpose of the `types_testing_incorrect_values` file.

etsybaev:
> ah maybe I just missed it - the links won't load because github is dumb :( can you link to the files directly?
> 
> the main thing is that only the `types_testing` stream actually verifies values - the other streams just verify record count, so if the data is written incorrectly, it'll still pass the test. That's the purpose of the `types_testing_incorrect_values` file.

Updated, thanks. But had to create dedicated scripts for each DBs.

edgao (approved):
nice! LGTM - we'll figure out how to get this merged in the new year :)

grishick (approved):

edgao:
ran a quick manual protocol v1 postgres -> snowflake sync, and got these results:
image

notably:
* binarydata ended up as a varchar
* time_with_timezone also was varchar
* timestamp_with_timezone somehow ended up as variant (probably stream_processor didn't recognize the type correctly)

I'll take a look at this PR to try and understand why those mismappings happened

airbyte-integrations/bases/base-normalization/normalization/data_type.py
```diff
@@ -0,0 +1,20 @@
...
+ONE_OF_VAR_NAME = "oneOf"
+
+STRING_TYPE = "WellKnownTypes.json#/definitions/String"
+BINARY_DATA_TYPE = "WellKnownTypes.json#/definitions/BinaryData"
+BOOLEAN_TYPE = "WellKnownTypes.json#/definitions/Boolean"
+INTEGER_TYPE = "WellKnownTypes.json#/definitions/Integer"
+NUMBER_TYPE = "WellKnownTypes.json#/definitions/Number"
+DATE_TYPE = "WellKnownTypes.json#/definitions/Date"
+TIMESTAMP_WITHOUT_TIMEZONE_TYPE = "WellKnownTypes.json#/definitions/TimestampWithoutTimezone"
+TIMESTAMP_WITH_TIMEZONE_TYPE = "WellKnownTypes.json#definitions/TimestampWithTimezone"
```
edgao, line 18:
```suggestion
TIMESTAMP_WITH_TIMEZONE_TYPE = "WellKnownTypes.json#/definitions/TimestampWithTimezone"
```

edgao (commented):
I think we should actually change the ordering of the if-else chain in stream_processor.py to something like:
```python
if is_array
elif is_object
elif is_string
elif is_binary
elif is_number
elif is_long
elif ...
```

i.e. move the "wider" types to the top. The older version had to put them at the bottom, because `is_string({type: string, airbyte_type: timestampwithtimezone})` would return true, even though it should be a timestamptz. But now that we're checking for exact matches against `#/definitions/String` vs `#/definitions/TimestampWithTImezone`, we don't have that problem anymore.

Currently this PR is actually prioritizing narrower types over wider types. E.g. `{"oneOf": [{"$ref": "WellKnownTypes.json#/definitions/Integer"}, {"$ref": "WellKnownTypes.json#/definitions/String"}]}` will be treated as an integer, even though it should be widened to string.

(I found this by manually editing the catalog in my local OSS DB to have a oneOf field)

edgao:
nice! I'll run a test against this later.

though now I'm wondering if we should write proper unit tests for this part of stream_processor - it looks like test_stream_processor only validates some stuff related to primary key / cursor :/

PR #19730, opened 2022-11-22
Disable broken connections when feature flag is on

alovew:
## What
Check the feature flag in the discover read endpoint and if it is enabled, then disable connections that have breaking schema changes or where the user has specified that non-breaking changes should cause the connection to be disabled

## Recommended reading order

Adding ConnectionStatus to SchemaRead object and updating ConnectionRead object:
1. `openapi/config.yaml`
2. `SchedulerHandler.java` -> line 384
3. `WebBackendConnectionHandler.java`
4. `WebBackendConnectionHandlerTest.java`
5. `ConnectionHelpers.java` <- helpers to support test objects

SchedulerHandler logic to disable broken connections:
1. `SchedulerHandler.java` -> lines 376-382
2. `SchedulerHandlerTest.java`

### DIFF
--- a/airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
+++ b/airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
@@ -15,6 +15,7 @@ import io.airbyte.api.model.generated.CatalogDiff;
 import io.airbyte.api.model.generated.CheckConnectionRead;
 import io.airbyte.api.model.generated.CheckConnectionRead.StatusEnum;
 import io.airbyte.api.model.generated.ConnectionIdRequestBody;
+import io.airbyte.api.model.generated.ConnectionRead;
 import io.airbyte.api.model.generated.ConnectionStatus;
 import io.airbyte.api.model.generated.ConnectionUpdate;
 import io.airbyte.api.model.generated.DestinationCoreConfig;
@@ -28,6 +29,7 @@ import io.airbyte.api.model.generated.JobConfigType;
 import io.airbyte.api.model.generated.JobIdRequestBody;
 import io.airbyte.api.model.generated.JobInfoRead;
 import io.airbyte.api.model.generated.LogRead;
+import io.airbyte.api.model.generated.NonBreakingChangesPreference;
 import io.airbyte.api.model.generated.SourceCoreConfig;
 import io.airbyte.api.model.generated.SourceDefinitionIdWithWorkspaceId;
 import io.airbyte.api.model.generated.SourceDefinitionSpecificationRead;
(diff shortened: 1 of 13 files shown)


airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
```diff
@@ -114,7 +119,8 @@ public SchedulerHandler(final ConfigRepository configRepository,
         jobPersistence,
         eventRunner,
         new JobConverter(workerEnvironment, logConfigs),
-        connectionsHandler);
+        connectionsHandler,
+        new EnvVariableFeatureFlags());
```
benmoriceau, line 123: Can we inject that instead of creating a new instance here, it will help for the migration to micronaut.

airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
```diff
@@ -360,16 +368,32 @@ private void discoveredSchemaWithCatalogDiff(SourceDiscoverSchemaRead discovered
...
+    discoveredSchema.catalogDiff(diff).breakingChange(containsBreakingChange).connectionStatus(connectionStatus);
+
+  }
+
+  private boolean shouldDisableConnection(boolean containsBreakingChange, NonBreakingChangesPreference preference) {
+    if (!envVariableFeatureFlags.autoDetectSchema()) {
+      return false;
+    }
 
+    return containsBreakingChange || preference == NonBreakingChangesPreference.DISABLE;
```
benmoriceau, line 396: should this be an && instead of an ||? Otherwise we will disable all connection where the preference is `NonBreakingChangesPreference.DISABLE`.

We need an UTest to catch that.

benmoriceau (changes requested):

airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
```diff
@@ -360,16 +369,32 @@ private void discoveredSchemaWithCatalogDiff(SourceDiscoverSchemaRead discovered
...
     boolean containsBreakingChange = containsBreakingChange(diff);
     ConnectionUpdate updateObject =
         new ConnectionUpdate().breakingChange(containsBreakingChange).connectionId(discoverSchemaRequestBody.getConnectionId());
+    ConnectionStatus connectionStatus;
+    if (shouldDisableConnection(containsBreakingChange, connectionRead.getNonBreakingChangesPreference(), diff)) {
+      connectionStatus = ConnectionStatus.INACTIVE;
+    } else {
+      connectionStatus = ConnectionStatus.ACTIVE;
+    }
+    updateObject.status(connectionStatus);
```
benmoriceau, line 386: We need a test that check that we are setting the expected status.

alovew: I have a bunch of new SchedulerHandlerTests -- is there another case you think I should be testing?

alovew: The status is being tested on each new test, and I also added it to existing tests

alovew: yes - you have to expand SchedulerHandlerTest in github to see it because I think it is collapsed due to the large number of changes to the file

benmoriceau: Oh my bad, I missed them because of them being collapsed 🤦

benmoriceau (approved):

PR #19731, opened 2022-11-22
Parse connector manifest JSON schema $refs in webapp

lmossman:
## What
The next phase of the Connector Builder project requires importing the connector manifest JSON schema and resolving it's $refs.

We chose to do this through the [json-schema-ref-parser](https://www.npmjs.com/package/@apidevtools/json-schema-ref-parser) library, because it seems to be by far the most well-supported (> 1.7M weekly downloads).
However, this required us to make some changes to our craco.config.js, because it requires some polyfills and other webpack configurations to get it to work properly.

## How
Left comments in the craco.config.js indicating the changes that were made. 
We also refactored the craco.config.js file to use lodash `merge` so that we didn't need to keep spreading the same config object.


lmossman:
Converted this back to a draft, because I am not certain that we will want to use this yet

airbyte-cdk/python/airbyte_cdk/sources/declarative/connector_manifest_schema_openapi.json
```diff
@@ -0,0 +1,1493 @@
+{
```
girarda, line 1: did you generate this from config_component_schema.json?

lmossman: Yeah this was copied over from that file and then modified a bit to convert it into an "openapi spec" type file. This is only being done right now as a workaround due to orval (our openapi typescript-generator library) still having issues with importing external json schemas. This workaround lets me just treat this entire file as its own openapi spec so I can [use orval to generate typescript directly from it](https://github.com/airbytehq/airbyte/pull/19731/files#diff-748ed30ef7eb13e66c7f64efcff9462812e56f3c50247731b613acb3a9cb10f9).

But I am still hopeful that orval will be properly fixed so that this isn't necessary, as the maintainer has been going back and forth with me and making incremental progress towards the right solution

PR #19732, opened 2022-11-22
🪟 🧪 Make sure to filter out connectors correctly

timroes:
## What

Fix the connector sign up experiment to properly filter connectors in that file. Earlier the `filter` method was simply checking if the return value of the `getExcludedConnectorIds` method was falsey for filtering out (which it never would be).

cc @letiescanciano sorry I apparently reviewed this sloppily and this slipped the review. I disabled the experiment until we have this merged next week.


airbyte-webapp/src/packages/cloud/components/experiments/SignupSourceDropdown/useGetSourceDefinitions.tsx
```diff
@@ -16,10 +16,11 @@ const fetchCatalog = async (): Promise<Catalog> => {
 };
 
 export const useGetSourceDefinitions = () => {
-  return useQuery<Catalog, Error, Catalog["sources"]>("cloud_catalog", fetchCatalog, {
+  return useQuery("cloud_catalog", fetchCatalog, {
     select: (data) => {
+      const filteredConnectors = getExcludedConnectorIds();
```
lmossman, line 21: nit: rename this variable to `excludedConnectorIds`. When I read the word `filtered`, I'm never sure if it means the things that were "filtered out", or if it means what is remaining after _other_ things have been filtered out.

lmossman (approved):
One small nit comment, otherwise LGTM. Did not test locally

PR #19751, opened 2022-11-23
Source Pinterest: changed AD_ACCOUNT_ID type to integer, updated acceptance-test-config…

darynaishchenko:
changed AD_ACCOUNT_ID type to integer, updated acceptance-test-config


girarda:
do you have context for this change? Looking at a few response samples from https://developers.pinterest.com/docs/api/v5/#operation/campaigns, it looks like ad_account_id should be a string

```
{
"items": [
{
"data": {
"ad_account_id": "549755885175",
"name": "ACME Tools",
"status": "ACTIVE",
<...>
```

darynaishchenko:
fixed in https://github.com/airbytehq/airbyte/pull/20242

Issue #19684, opened 2022-11-21
Custom connector related jobs should run in a different node pool


xiaohansong:
For all activities, we want to create a bit to indicate if that's custom connector related; and if so, in KubeProcessFactory.java we should assign a different node selector to run the jobs.

The custom connector bit could also be useful for metrics tracking purpose.

pmossman:
Refining notes:
- Kube process factory controls our node selectors, which is called for every Temporal workflow. We need to pass in a flag that indicates if the source or destination is a custom connector. There are 16 different Temporal job types that will need this flag, so the work here involves updating the 16 different job configs to have a flag for isCustom.

PR #19770, opened 2022-11-23
Use a separate node pool to run custom connector jobs

xiaohansong:
## What
https://github.com/airbytehq/airbyte/issues/19684

Custom connectors - user can upload their own docker image as the source or destination of the sync. Such sourceDefinition or destinationDefinition will be classified as custom connectors. Since they are not vetted by airbyte, we want to run it in a separate node pool for safety reasons to avoid failing jobs running with airbyte vetted connectors.

The goal of the PR is to:
1) Identify which activity are involving custom connectors.
2) pass along custom connector bit via temporal and eventually when starting a kube process we know it's a custom connector and thus run it on a different node pool.

## How
Passing the value in launchConfig - launcher will read launchConfig and launch a kube process with correct node selector.

For jobs spawned by Orchestrator, we also need to pass down the env variable to containers.


airbyte-commons-worker/src/main/java/io/airbyte/workers/normalization/DefaultNormalizationRunner.java
```diff
@@ -138,6 +138,8 @@ private boolean runProcess(final String jobId,
           attempt,
           jobRoot,
           normalizationImageName,
+          // custom connector does not use normalization
```
xiaohansong, line 141: @davinchia Can you help me confirm if this is true? (seems NormalizationRunner will only be running if it's called out here: https://github.com/airbytehq/airbyte/blob/master/airbyte-commons-worker/src/main/java/io/airbyte/workers/normalization/NormalizationRunnerFactory.java#L18-L38)

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DbtTransformationRunner.java
```diff
@@ -99,6 +99,7 @@ public boolean transform(final String jobId,
               jobRoot,
               dbtConfig.getDockerImage(),
               false,
```
xiaohansong, line 101: @davinchia or @mfsiega-airbyte I don't have much context on this one - from the comment it seems transformation will only be provided by user, but for custom connector are we allowing them to use Dbt related features?

pmossman: We don't currently support custom DBT transformations in Cloud (and we deprioritized the work to enable custom transformations on kube this quarter) but I think DBT transformations are only supported in a specific list of destinations, just like with normalization. So for custom destination connectors, I don't think we need to worry about DBT transformations or Normalization (@davinchia should confirm though to be sure)

airbyte-commons-worker/src/main/java/io/airbyte/workers/normalization/DefaultNormalizationRunner.java
pmossman: I believe this is true, normalization is supported in a specific list of destination connectors, so I don't think we would be able to support normalization for custom destination connectors (at least for V1)

davinchia: Parker is correct!

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DbtTransformationRunner.java
davinchia: yeap, Parker is correct here too!

davinchia:
@xiaohansong apart from the 2 questions, are there specific bits you would like me to review? I think my review is not needed if Parker is going over this - putting it out here if there you want a second eye.

pmossman (changes requested):
@xiaohansong I just finished going through this, great stuff! I made some suggestions and flagged a couple of questions/thoughts but I think this looks pretty close to me. 

@davinchia I think you could just look through my comments and weigh in if you have thoughts, rather than reviewing everything. I think I commented everywhere that is worth taking a look, the rest of the PR is straightforward and/or boilerplate

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/DockerProcessFactory.java
```diff
@@ -83,6 +83,7 @@ public Process create(final String jobType,
                         final int attempt,
                         final Path jobRoot,
                         final String imageName,
+                        final boolean usesIsolatedPool,
```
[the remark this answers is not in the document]
xiaohansong: I didn't know docker has this pool concept as well. Changed to use `usesIsolatedPool`.

airbyte-commons-worker/src/main/java/io/airbyte/workers/sync/ReplicationLauncherWorker.java
```diff
@@ -54,7 +54,8 @@ public ReplicationLauncherWorker(final UUID connectionId,
         activityContext,
         serverPort,
         temporalUtils,
-        workerConfigs);
+        workerConfigs,
+        sourceLauncherConfig.getIsCustomConnector() || destinationLauncherConfig.getIsCustomConnector());
```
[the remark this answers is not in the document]
xiaohansong: Fixed!

airbyte-container-orchestrator/src/main/java/io/airbyte/container_orchestrator/ReplicationJobOrchestrator.java
```diff
@@ -106,15 +107,17 @@ public Optional<String> runJob() throws Exception {
...
 
     log.info("Setting up destination launcher...");
     final IntegrationLauncher destinationLauncher = new AirbyteIntegrationLauncher(
         destinationLauncherConfig.getJobId(),
         Math.toIntExact(destinationLauncherConfig.getAttemptId()),
         destinationLauncherConfig.getDockerImage(),
         processFactory,
-        syncInput.getDestinationResourceRequirements());
+        syncInput.getDestinationResourceRequirements(),
+        destinationLauncherConfig.getIsCustomConnector());
```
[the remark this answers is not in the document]
xiaohansong: yup done.

airbyte-workers/src/main/java/io/airbyte/workers/config/WorkerConfigurationBeanFactory.java
```diff
@@ -85,6 +85,18 @@ public Map<String, String> specJobKubeAnnotations(@Value("${airbyte.worker.spec.
     return splitKVPairsFromEnvString(kubeAnnotations);
   }
 
+  @Singleton
+  @Named("customNodeSelectors")
```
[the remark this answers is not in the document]
xiaohansong: decide to use `useIsolatedNodeSelector` to keep align with other names (such as specJobKubeNodeSelectors). Since this is to define the actual kube selector I think we should just include `node` in the name

airbyte-commons-worker/src/main/java/io/airbyte/workers/normalization/DefaultNormalizationRunner.java
xiaohansong: Perfect, thanks!

airbyte-commons-worker/src/main/java/io/airbyte/workers/WorkerConfigs.java
```diff
@@ -41,6 +42,7 @@ public WorkerConfigs(final Configs configs) {
             .withMemoryLimit(configs.getJobMainContainerMemoryLimit()),
         configs.getJobKubeTolerations(),
         configs.getJobKubeNodeSelectors(),
+        configs.getUseCustomKubeNodeSelector() ? configs.getIsolatedJobKubeNodeSelectors() : configs.getJobKubeNodeSelectors(),
```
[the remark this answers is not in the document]
xiaohansong: I agree. Changed!

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubeProcessFactory.java
```diff
@@ -107,6 +108,8 @@ public Process create(
 
       final var allLabels = getLabels(jobId, attempt, customLabels);
 
+      final var nodeSelectors = usesIsolatedPool ? workerConfigs.getWorkerIsolatedKubeNodeSelectors() : workerConfigs.getworkerKubeNodeSelectors();
```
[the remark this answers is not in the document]
xiaohansong: it can be empty if the flag was set to false - if they are using custom connectors but didn't turn on the useIsolatedNodePool flag, they might hit this

xiaohansong:
@pmossman when you get a chance can you take another look at this PR? thanks!

airbyte-commons-worker/src/main/java/io/airbyte/workers/sync/LauncherWorker.java
```diff
@@ -176,13 +176,19 @@ public OUTPUT run(final INPUT input, final Path jobRoot) throws WorkerException
           log.info("Creating " + podName + " for attempt number: " + jobRunConfig.getAttemptId());
           killRunningPodsForConnection();
 
+          // custom connectors run in an isolated node pool from airbyte-supported connectors
+          // to reduce the blast radius of any problems with custom connector code.
+          final var nodeSelectors =
+              isCustomConnector ? workerConfigs.getWorkerIsolatedKubeNodeSelectors().orElse(workerConfigs.getworkerKubeNodeSelectors())
```
pmossman, line 182: Similar comment here, I think it's more dangerous to fall back on our primary node selectors, and perhaps we should throw an exception instead.

airbyte-server/src/main/java/io/airbyte/server/handlers/SourceDefinitionsHandler.java
```diff
@@ -285,16 +289,6 @@ public void deleteSourceDefinition(final SourceDefinitionIdRequestBody sourceDef
     configRepository.writeStandardSourceDefinition(persistedSourceDefinition);
   }
 
-  public void deleteCustomSourceDefinition(final SourceDefinitionIdWithWorkspaceId sourceDefinitionIdWithWorkspaceId)
```
pmossman: is this change supposed to be part of this PR? Or did it sneak in from your other branch for consolidating API endpoints?

airbyte-server/src/test/java/io/airbyte/server/handlers/DestinationDefinitionsHandlerTest.java
```diff
@@ -336,52 +336,6 @@ void testGetDefinitionWithGrantForWorkspace() throws JsonValidationException, Co
     assertEquals(expectedDestinationDefinitionRead, actualDestinationDefinitionRead);
   }
 
-  @Test
-  @DisplayName("createDestinationDefinition should correctly create a destinationDefinition")
```
pmossman: similar question here, is this change sneaking in from your other PR?

airbyte-server/src/test/java/io/airbyte/server/handlers/SourceDefinitionsHandlerTest.java
```diff
@@ -324,54 +324,6 @@ void testGetDefinitionWithGrantForWorkspace() throws JsonValidationException, Co
     assertEquals(expectedSourceDefinitionRead, actualSourceDefinitionRead);
   }
 
-  @Test
```
pmossman, line 327: similar question here, is this change sneaking in from your other PR?

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityImpl.java
```diff
@@ -111,7 +111,9 @@ public GeneratedJobInput getSyncWorkflowInput(final SyncInput input) {
           .withAttemptId((long) attempt)
           .withDockerImage(config.getDestinationDockerImage())
           .withProtocolVersion(config.getDestinationProtocolVersion())
-          .withIsCustomConnector(config.getIsDestinationCustomConnector());
+          .withIsCustomConnector(config.getIsDestinationCustomConnector())
```
pmossman, line 114: These changes look unrelated, maybe due to a merge/rebase with master?

pmossman (approved):
@xiaohansong I left a couple more comments but overall this looks good to me! I'll approve the PR assuming we don't disagree about any of my comments, let me know if you want to sync about any of them!

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubeProcessFactory.java
```diff
@@ -108,7 +108,11 @@ public Process create(
 
       final var allLabels = getLabels(jobId, attempt, customLabels);
 
-      final var nodeSelectors = usesIsolatedPool ? workerConfigs.getWorkerIsolatedKubeNodeSelectors() : workerConfigs.getworkerKubeNodeSelectors();
+      // If using isolated pool, check workerConfigs has isolated pool set. If not set, fall back to use
+      // regular node pool.
+      final var nodeSelectors =
+          usesIsolatedPool ? workerConfigs.getWorkerIsolatedKubeNodeSelectors().orElse(workerConfigs.getworkerKubeNodeSelectors())
```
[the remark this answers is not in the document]
xiaohansong: good call! fixed

airbyte-commons-worker/src/main/java/io/airbyte/workers/sync/LauncherWorker.java
xiaohansong: I added this check in `WorkerConfigurationBeanFactory.java` instead - because 1) workerConfigs does not carry `useIsolatedPool` flag, 2) should throw this error as early as possible

airbyte-server/src/main/java/io/airbyte/server/handlers/SourceDefinitionsHandler.java
xiaohansong: ah true

airbyte-server/src/test/java/io/airbyte/server/handlers/SourceDefinitionsHandlerTest.java
xiaohansong: ditto

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityImpl.java
xiaohansong: yep!

airbyte-commons-worker/src/main/java/io/airbyte/workers/sync/LauncherWorker.java
pmossman: Nice!

airbyte-workers/src/main/java/io/airbyte/workers/config/WorkerConfigurationBeanFactory.java
```diff
@@ -228,6 +228,9 @@ public WorkerConfigs checkWorkerConfigs(
                                           @Value("${airbyte.worker.job.kube.images.busybox}") final String busyboxImage,
                                           @Value("${airbyte.worker.job.kube.images.curl}") final String curlImage,
                                           @Named("jobDefaultEnvMap") final Map<String, String> jobDefaultEnvMap) {
+    if (useIsolatedNodeSelector && isolatedNodeSelectors.isEmpty()) {
```
pmossman, line 231: since this gets repeated so many times, I'd probably move it into a private method that takes in `useIsolatedNodeSelector` and `isolatedNodeSelectors` as arguments

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubeProcessFactory.java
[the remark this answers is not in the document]
xiaohansong: Realized this is not true - usesIsolatedPool here is actually passed by callers (isCustomConnector flag). Thus for OSS users who do not run them in a separate pool, usesIsolatedPool might still be true because they are custom connectors. I should rename this to isCustomConnectors instead.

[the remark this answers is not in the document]
pmossman: Oh good call, OSS users should be able to use custom connectors without an isolated pool indeed

xiaohansong:
Connector base build has been failing for unrelated reasons. Submitting this.

PR #19771, opened 2022-11-23
Add schema change attrs to connection create

alovew:
Allow users to specify non-breaking schema change preference on Connection Create

### DIFF
--- a/airbyte-api/src/main/openapi/config.yaml
+++ b/airbyte-api/src/main/openapi/config.yaml
@@ -3371,6 +3371,8 @@ components:
           format: uuid
         geography:
           $ref: "#/components/schemas/Geography"
+        nonBreakingChangesPreference:
+          $ref: "#/components/schemas/NonBreakingChangesPreference"
     ConnectionStateCreateOrUpdate:
       type: object
       required:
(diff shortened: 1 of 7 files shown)


gosusnp (approved):
LGTM.
Probably out of scope for this PR, but I'll ask since you may have more context. A long list of duplicated properties like this feels very error prone, how complex do you think it would be to refactor `WebBackendConnectionCreate` so that it can use `ConnectionCreate`?

alovew:
I'm really not sure why it was built this way. But it seems we have this pattern for all of our objects that are passed to the WebBackendConnectionHandler - there's a version of the object for WebBackendConnectionHandler (e.g. WebBackendConnectionUpdate, WebBackendConnectionPatch, etc) and then a new object is created to pass to the ConnectionHandler. @benmoriceau do you know why this was done this way?

cgardens:
>  how complex do you think it would be to refactor WebBackendConnectionCreate so that it can use ConnectionCreate?

we've looked in the past and it's hard to merge the definitions of 2 structs in openapi. we have tried to use `allof` but apparently that's considered bad practice. 🤷‍♀️

gosusnp:
> > how complex do you think it would be to refactor WebBackendConnectionCreate so that it can use ConnectionCreate?
> 
> we've looked in the past and it's hard to merge the definitions of 2 structs in openapi. we have tried to use `allof` but apparently that's considered bad practice. 🤷‍♀️

What about nesting? (like having `WebBackendConnectionCreate` contain a `ConnectionCreate`) It may not be as elegant but I'd say it would be probably more practical for us.

cgardens:
sure. given our new public api strategy, i feel like we can be a little less precious about these webbackend apis.

PR #19773, opened 2022-11-23
:window: :wrench: Handle LD feature flags besides `featureService.overwrites`

### LINKED ISSUE #19726: Spike: Update cloud FE's LaunchDarkly wrapper to handle arbitrary feature flags
Right now, all the feature overrides we use are defined as variants of a single catchall feature flag, `featureService.overwrites`. Because LaunchDarkly user targeting (which we use to selectively release new features and for initial releases which can be "rolled back" with no deploy) is mutually exclusive (i.e. each user can only be assigned one single variant), independently toggling different features in parallel gets complex and fussy very quickly.

Instead, the frontend code which parses the LaunchDarkly dynamic flag assignments and provides flag values to application code should be able to find the values for new, independent feature flags with arbitrary names. For backwards compatibility, it must continue to properly query variants defined within the `featureService.overwrites` namespace until and unless all such feature assignments are deprecated.

ambirdsall:
## Background
Feature flags in LaunchDarkly serve multiple uses in airbyte cloud at various points in the lifecycle of a feature release. Sometimes they are used to opt specific people into the feature for testing; sometimes they are used to "soft launch" a feature in a way that can be rolled out and rolled back without the lag of a code update and deployment; sometimes they are used to grandfather certain users into the ability to continue using otherwise deprecated code.

The original implementation of our LaunchDarkly frontend integration put all feature toggles into a single LD "feature flag", with various experiences defined as variants on that single flag. Because all of a given flag's variants are mutually exclusive to a given user, this means we could not independently toggle different features for the whole userbase without auditing and possibly editing every other more targeted rule and worrying about rule precedence in LD. This added complexity and hassle scales up exponentially with the number of distinct feature toggles in use at a given moment.

## External context
A chunk of the work for this feature happened in LaunchDarkly itself: for each feature toogle that has been managed to date with the shared `featureService.overwrites` feature flag needed to be ported to a new feature flag using an `{ enabled?: boolean }` payload, along with all targeting rules used to apply that variant. More on this below.

## What this changes
This adds logic to the LaunchDarkly integration that populates the FeatureService's overwrites to accept single-toggle feature flags. It hardcodes the assumption that all such flags will use a key that begins with the same `featureService.` prefix as the shared feature flag: this (otherwise arbitrary) criterion lets us quickly and reliably filter the list of feature flags in LD to the set that is specifically relevant to the frontend codebase for visibility and auditing. Closes https://github.com/airbytehq/airbyte/issues/19726.

Here's a recording of a manual test of the new code, toggling the presence of two different sub-nav link via new, independent feature flags. At the time of recording this screencap, https://github.com/airbytehq/airbyte/pull/20105 had not yet been merged into the feature branch, which is kind of ideal for demonstration purposes: explicitly using the "enable feature" and "disable feature" variants work identically for each feature, while using the "use application default" variants enables the "Data residency" link and disables the "dbt Cloud integration" one, using the cloud build's default features [as of commit `7b19f59`](https://github.com/airbytehq/airbyte/blob/7b19f59f7c1d07f1daf8c972078b97444b531819/airbyte-webapp/src/packages/cloud/App.tsx#L42-L48):
independent-feature-flags

## 🚨 User Impact 🚨
Existing feature flags are handled exactly as before, so there should be zero user impact. There is one cohort of users who have been grandfathered into continued use of sub-one-hour sync scheduling, so some care should be taken to ensure that deploying this change causes no interruption to that targeting logic.

## Manual testing
Unfortunately, there's not a great way to add automated tests for this feature: writing jest tests would require lots of time consuming effort to mock out large swaths of the application context before any assertions can be written; cypress tests would be ideal, but we only use `LDExperimentService` in cloud, and we don't yet have a test suite for that. So manual testing it is.

Tests done:
- [X] *In the feature branch*, each existing feature can be toggled on, off, or to the application default in isolation via the new feature flags (tested via the LD development environment and a local frontend)
- [X] *In the feature branch*, toggling a feature via the shared `featureService.overwrites` feature flag does not affect the UI (same testing setup)
- [X] *In the existing prod app*, creating and toggling the new feature flags does not affect the UI (tested via [a targeting rule matching my personal email](https://app.launchdarkly.com/default/production/features/featureService.ALLOW_DBT_CLOUD_INTEGRATION/targeting) in the LD production environment and cloud.airbyte.io)

## Documentation
This seems worth documenting for the Airbyte team! That is not, strictly speaking, a concern for this PR, since it happens outside of the codebase; just calling it out to keep myself honest.

### DIFF
--- a/airbyte-webapp/src/packages/cloud/services/thirdParty/launchdarkly/LDExperimentService.tsx
+++ b/airbyte-webapp/src/packages/cloud/services/thirdParty/launchdarkly/LDExperimentService.tsx
@@ -12,10 +12,14 @@ import { useAnalyticsService } from "hooks/services/Analytics";
 import { useAppMonitoringService, AppActionCodes } from "hooks/services/AppMonitoringService";
 import { ExperimentProvider, ExperimentService } from "hooks/services/Experiment";
 import type { Experiments } from "hooks/services/Experiment/experiments";
-import { FeatureSet, useFeatureService } from "hooks/services/Feature";
+import { FeatureSet, FeatureItem, useFeatureService } from "hooks/services/Feature";
 import { User } from "packages/cloud/lib/domain/users";
 import { useAuthService } from "packages/cloud/services/auth/AuthService";
 import { rejectAfter } from "utils/promises";
+
+type LDFeatureName = `${FeatureItem}` | "overwrites";
+type RawLDFeatureName = `${typeof FEATURE_FLAG_PREFIX}.${LDFeatureName}`;
+type LDInitState = "initializing" | "failed" | "initialized";
 
 /**
  * The maximum time in milliseconds we'll wait for LaunchDarkly to finish initialization,
@@ -23,9 +27,18 @@ import { rejectAfter } from "utils/promises";
  */
 const INITIALIZATION_TIMEOUT = 5000;
 
-const FEATURE_FLAG_EXPERIMENT = "featureService.overwrites";
-
-type LDInitState = "initializing" | "failed" | "initialized";
+// Originally, all feature toggles were contained within one single feature flag in
+// LaunchDarkly, using internal conventions to represent multiple feature toggles from a
+// single variant. `FEATURE_FLAG_EXPERIMENT` is the LaunchDarkly key for that shared
+// feature flag.
+//
+// This service hardcodes the convention that all feature overrides in LaunchDarkly use
+// keys that follow the format `${FEATURE_FLAG_PREFIX}.${FeatureItem}`. The primary
+// benefit of requiring the prefix in code is to provide a reliable search term which can
+// be used in LaunchDarkly to retrieve the complete set of flags which can be used to
+// dynamically toggle UI features.
+const FEATURE_FLAG_PREFIX = "featureService";
+const FEATURE_FLAG_EXPERIMENT: RawLDFeatureName = "featureService.overwrites";
 
 function mapUserToLDUser(user: User | null, locale: string): LDClient.LDUser {
   return user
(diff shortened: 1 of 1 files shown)


ambirdsall:
That's a pair of very good points, @timroes, thank you.

- I hadn't thought through the option of swapping the implementation of _existing_ feature flags to something nicer, but there's nothing blocking that: the alternate, independent feature flag definitions will be straightforward to create in advance and will be silently ignored until we deploy updated code that can use them. They will also make manual e2e testing much easier, since we already have conditional UI toggled by those flags (in particular, the multi-cloud and dbt Cloud integration flags both control the cloud build's settings page content).
- For representing all three possible states of an override flag (set on, set off, no override), we can define our LD flags to return a JSON object with an optional boolean property `overwrite` instead of a top-level boolean value (cf. [the `connector.orderOverwrite` flag](https://app.launchdarkly.com/default/test/features/connector.orderOverwrite/variations)); then we can represent the three valid states like so:
```typescript
// enable feature from LD
{ overwrite: true }

// disable feature from LD
{ overwrite: false }

// no override; default variation
{}
```

ambirdsall:
The most important external context from a business perspective: the [new](https://app.launchdarkly.com/default/production/features/featureService.ALLOW_SYNC_SUB_ONE_HOUR_CRON_EXPRESSIONS/targeting) targeting rule to grandfather certain users into sub-one-hour sync schedules matches the [old one](https://app.launchdarkly.com/default/production/features/featureService.overwrites/targeting), so those users won't see any disruption of their current usage.

joeykmh (approved):
Very nice work @ambirdsall! This is a solid improvement to working with feature service overwrites 👏🏻 I tested locally with the data geography & cron sync mode flags, and everything worked as expected. I also checked the LD flags on production, and the targeting for ThriveCap on `featureService.ALLOW_SYNC_SUB_ONE_HOUR_CRON_EXPRESSIONS` looks good.

The only thought I had while reading through - and this is definitely out of scope for this PR - is why do we have different semantics for "experiments" and the "featureService" flags? `Experiments` uses keys of an interface while `FeatureItem` uses an enum, experiments uses an rxjs observable while featureService overwrites use a callback function, experiments are camelCase while featureService is ALL_CAPS. Anyway, not pertinent to the core changes here, just strikes me as strangely inconsistent.

airbyte-webapp/src/packages/cloud/services/thirdParty/launchdarkly/LDExperimentService.tsx
```diff
@@ -12,21 +12,45 @@ import { useAnalyticsService } from "hooks/services/Analytics";
...
+ * reliable search term which can be used in LaunchDarkly to filter the list of feature
+ * flags to all of, and only, the ones which can dynamically toggle features in the UI.
+ *
+ * LDFeatureToggle objects can take three forms, representing the three possible decision
+ * states LaunchDarkly can provide for a user/feature pair:
+ * |--------------------------+-----------------------------------------------|
+ * | `{}`                     | use the application's default feature setting |
+ * | `{ "enabled": true }`    | enable the feature                            |
+ * | `{ "enabled": false }`   | disable the feature                           |
+ * |--------------------------+-----------------------------------------------|
```
timroes, line 38: Just a thought, not a necessary blocker for this PR: Would it maybe be nicer if we'd use a `string` for the tristate, that can either be empty, `"true"` or `"false"`? That would be easier to type in LD and would have the benefit, that the default value for string experiments already is the empty string, while using the JSON flag here, we'll require to type `{}` for each of the new features to not break the code below that destructures the object?

PR #19806, opened 2022-11-25
SAT: optionally disable discovered catalog caching

alafanechere:
## What
I think the ongoing build failure of the `source-tiktok-marketing` connectors are due to a default caching of discovered catalog in SAT. 

This connector catalog is dynamic according to connector configurations. Multiple acceptance tests are run with different configurations hence different catalogs. But some tests fail because they try to retrieve a stream defined in the configured catalog absent from the cached discovered catalog.

After investigation, there is nothing badly configured on TikTok's acceptance test config. 
I figured a catalog discovered in the previous tests was reused and caused the error.

## How
* Expose a new configuration field on SAT to disable discovered catalog caching.


airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/config.py
```diff
@@ -189,6 +189,9 @@ class TestStrictnessLevel(str, Enum):
         high = "high"
         low = "low"
 
+    cache_discovered_catalog: bool = Field(
+        default=True, description="Enable or disable caching of discovered catalog for reuse in multiple tests."
+    )
```
evantahler, line 194: Should the default be false to prevent these kind of errors?  What's the slow-down introduced if we don't cache?

alafanechere: If we don't cache the discovered catalog, every test using it will trigger a `discover` command run as the scope of the `discovered_catalog` fixture is `function` by default. As this caching feature did not bother us until now, a specific edge case, I would suggest keeping it as a default. 
We could try to make it smarter and leverage another fixture scope  for more granular caching (e.g. `class` scope).

evantahler: 👍

evantahler (approved):

pedroslopez (approved):
Makes sense!

If the issue is that the cache is different per config, could there be a way to cache per-config rather than globally? This way it wouldn't have to be a user-defined thing they have to figure out how to set and the SAT would just be a bit smarter about how it caches the schemas.

alafanechere:
+1 @pedroslopez . I created an issue to track this improvement: https://github.com/airbytehq/airbyte/issues/19872

alafanechere:

PR #19693, opened 2022-11-22
Add Retry Logic to Airbyte API calls.

davinchia:
## What
Today we often see `HTTP/1.1 header parser received no bytes'` during syncs, especially in the Data Plane.

This PR attempts to fix this by adding naive retries.

~If this works, we will explore using a more fully featured HTTP Client. I will **not** merge this in.~

It turns out this is the best way for what we are trying to do. More details in How section.

## How
Add a basic retry wrapper with the unique ability to retry for a much longer period on the last retry. This is particularly useful for us as most of our jobs are long running workflows, and the benefit of not having to restart the entire job outweighs the added wait time.

Alternative solutions I explored:
* Switching the underlying HTTP client to a more fully featured HTTP client. E.g. Apache or OkHttp. Issues with this:
  * These clients do not support the ability to configure the retry policy we want.
  * These clients do not support the ability to inject application aware logging.
  * Most importantly, because this changes the interface, the resulting change set is big and affects many unrelated classes. I do think we eventually want to switch the underlying libraries out. However I don't think we should do this as part of OC work.
* Exploring pairing retry libraries such as https://resilience4j.readme.io/docs with the native http clients. The main issue here is the lack of ability to configure the last retry period. 

Since the hand-rolled wrapper is simple + gets the job done, my thoughts are to run with this for the time being and revisit this if additional requirements around the clients come up.

## Recommended reading order
1. `AirbyteApiClient.java` for implementation and `AirbyteApiClientTest.java` for tests.
2. All others are usage changes.

## 🚨 User Impact 🚨
Are there any breaking changes? What is the end result perceived by the user? If yes, please merge this PR with the 🚨🚨 emoji so changelog authors can further highlight this if needed.

## Pre-merge Checklist
Expand the relevant checklist and delete the others.

<details><summary><strong>New Connector</strong></summary>

### Community member or Airbyter

- [ ] **Community member?** Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
[...]

PR #19811, opened 2022-11-26
feat(Platform): update actor configuration when receiving control messages from connectors during sync

### LINKED ISSUE #17909: The Airbyte Platform can accept and store updated configurations from syncs
_Part of the [Single Use Refresh Tokens](https://github.com/airbytehq/airbyte/issues/3990) Epic_

The Protocol Change to add `AirbyteControlMessage.ConnectorConfig` was done via https://github.com/airbytehq/airbyte/pull/17907. 

```js
{
  type: CONTROL,
  emitted_at: 123456,
  connectorConfig: {
    config: : {
       "oAuthToken": "abc123"
    }
  }
}
```

Now, we need to use these messages to store updated configurations sent by the connector:

1. Update the AirbyteMessageTracker to process the new messages; release a new version of Airbyte OSS & Cloud
    * ~~Config updates should be stored in-memory until the end of the sync~~
    * We should only write config updates once (to avoid undo stress on the database)
    * We should write config updates regardless of if the sync was successful or not 
2. Ensure that AirbyteConnectorConfigMessage is checked for all types of activities - CHECK, DISCOVER, and READ/WRITE

Follow the pattern for STATE updates to be persisted at the end of the sync. There is some nuance for workers running on Cloud (e.g we don't have database access in the middle of a sync).  The risk here is that if the worker crashes the update is lost... but that's OK for now. We can resist this in later work. 

Look here - https://github.com/airbytehq/airbyte/blob/b05a5b2a6a7c01fe4ee8f3c627fa8f31a02332fa/airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/AirbyteMessageTracker.java#L236-L245

pedroslopez:
## What

Process connector's config control messages and update the stored configuration accordingly when this kind of message is received.

This PR focuses on making this work for the replication, and a follow-up PR will do the same for the other commands e.g. check, discover)

Here's a[ Loom demo](https://www.loom.com/share/956ef99da8e449f7b5ff5e42478a5957) of it in action.

Closes airbytehq/airbyte#17909
Closes https://github.com/airbytehq/airbyte/issues/17909
Closes https://github.com/airbytehq/airbyte-internal-issues/issues/206

## How

- Create new UpdateConnectorConfigHelper class to call APIs to update configs for a source / destination
- Inject and use the above in DefaultReplicationWorker
- Accept CONTROL messages from the destination / source so we can process them
- In the DefaultReplicationWorker, when we see a config control message call the helper class to store the updated config
- Update the Source/Destination update APIs to handle masking global OAuth parameters

TODO:
- [x] handle initializing the AirbyteApiClient / UpdateConnectorConfigHelper for use in the container-orchestrator in kubernetes environments
- [x] test the above in a real kube env

## Recommended reading order
1. `x.java`
2. `y.python`

## 🚨 User Impact 🚨
No breaking changes - connectors will now be able to update their configs in the middle of a sync.


airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteDestination.java
```diff
@@ -81,8 +82,9 @@ public void start(final WorkerDestinationConfig destinationConfig, final Path jo
 
     writer = messageWriterFactory.createWriter(new BufferedWriter(new OutputStreamWriter(destinationProcess.getOutputStream(), Charsets.UTF_8)));
 
+    final List<Type> acceptedMessageTypes = List.of(Type.STATE, Type.TRACE, Type.CONTROL);
     messageIterator = streamFactory.create(IOs.newBufferedReader(destinationProcess.getInputStream()))
-        .filter(message -> message.getType() == Type.STATE || message.getType() == Type.TRACE)
+        .filter(message -> acceptedMessageTypes.contains(message.getType()))
         .iterator();
```
pedroslopez, line 88: Enable reading control messages from destinations

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteSource.java
```diff
@@ -86,9 +87,10 @@ public void start(final WorkerSourceConfig sourceConfig, final Path jobRoot) thr
 
     logInitialStateAsJSON(sourceConfig);
 
+    final List<Type> acceptedMessageTypes = List.of(Type.RECORD, Type.STATE, Type.TRACE, Type.CONTROL);
     messageIterator = streamFactory.create(IOs.newBufferedReader(sourceProcess.getInputStream()))
         .peek(message -> heartbeatMonitor.beat())
-        .filter(message -> message.getType() == Type.RECORD || message.getType() == Type.STATE || message.getType() == Type.TRACE)
+        .filter(message -> acceptedMessageTypes.contains(message.getType()))
```
pedroslopez, line 93: Enable reading control messages from sources

airbyte-commons-worker/src/test/java/io/airbyte/workers/general/DefaultReplicationWorkerTest.java
```diff
@@ -231,14 +242,112 @@ void testReplicationRunnableSourceFailure() throws Exception {
...
+        source,
+        mapper,
+        destination,
+        messageTracker,
+        recordSchemaValidator,
+        workerMetricReporter,
+        persistConfigHelper);
+
+    final ReplicationOutput output = worker.run(syncInput, jobRoot);
+    assertEquals(ReplicationStatus.FAILED, output.getReplicationAttemptSummary().getStatus());
```
pedroslopez, line 295: Note: we will fail the sync if there was an error when persisting the new config.

airbyte-container-orchestrator/src/main/java/io/airbyte/container_orchestrator/ReplicationJobOrchestrator.java
```diff
@@ -149,6 +153,15 @@ public Optional<String> runJob() throws Exception {
...
   }
 
+  private AirbyteApiClient getAirbyteApiClient() {
+    // TODO make this actually functional
+    return new AirbyteApiClient(
+        new ApiClient().setScheme("http")
+            .setHost("airbyte-server-svc")
+            .setPort(8001)
+            .setBasePath("/api"));
+  }
```
pedroslopez, line 163: This is the main missing piece right now: getting it working and trying it out in a kubernetes environment with the container orchestrator.

airbyte-persistence/job-persistence/src/main/java/io/airbyte/persistence/job/factory/OAuthConfigSupplier.java
```diff
@@ -85,31 +117,21 @@ public JsonNode injectDestinationOAuthParameters(final UUID destinationDefinitio
...
-      return false;
-    }
-    // TODO: if we write a migration to flatten persisted configs in db, we don't need to flatten
-    // here see https://github.com/airbytehq/airbyte/issues/7624
-    boolean result = false;
-    final JsonNode flatOAuthParameters = MoreOAuthParameters.flattenOAuthConfig(oAuthParameters);
+  /**
+   * Gets the OAuth parameter paths as specified in the connector spec and traverses through them
+   */
+  private static void traverseOAuthOutputPaths(final ConnectorSpecification spec,
```
pedroslopez, line 123: Extracted this logic from the existing injection method so it can be re-used for both injection and masking

airbyte-persistence/job-persistence/src/main/java/io/airbyte/persistence/job/factory/OAuthConfigSupplier.java
```diff
@@ -47,6 +49,36 @@ public static boolean hasOAuthConfigSpecification(final ConnectorSpecification s
     return spec != null && spec.getAdvancedAuth() != null && spec.getAdvancedAuth().getOauthConfigSpecification() != null;
   }
 
+  public JsonNode maskSourceOAuthParameters(final UUID sourceDefinitionId, final UUID workspaceId, final JsonNode sourceConnectorConfig)
```
pedroslopez, line 52: Added some new methods to _mask_ oauth parameters from a provided config and ensure we don't write our global oauth params to the connector-specific config / secrets.

airbyte-server/src/main/java/io/airbyte/server/handlers/SourceHandler.java
```diff
@@ -295,14 +301,14 @@ private void persistSourceConnection(final String name,
                                        final JsonNode configurationJson,
                                        final ConnectorSpecification spec)
       throws JsonValidationException, IOException {
+    final JsonNode oAuthMaskedConfigurationJson = oAuthConfigSupplier.maskSourceOAuthParameters(sourceDefinitionId, workspaceId, configurationJson);
```
pedroslopez, line 304: New: Source and Destination update APIs now mask global oAuth parameters if they're present. For OSS users without any oauth parameters in their db this will have no effect.

These endpoints were already handling splitting out the secrets to store separately from the config - we're just adding to this and _also_ masking global oauth params.

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/PersistConfigHelper.java
```diff
@@ -0,0 +1,90 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Helper class for workers to persist updates to Source/Destination configs emitted from
+ * AirbyteControlMessages.
+ */
+public class PersistConfigHelper {
```
pedroslopez, line 34: Extracted the methods to persist config updates for a job into this helper class since there are a few api calls here that we'll very likely re-use when bringing this functionality to the other connector commands.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/AirbyteMessageTracker.java
```diff
@@ -238,12 +238,8 @@ private void handleEmittedOrchestratorMessage(final AirbyteControlMessage contro
   @SuppressWarnings("PMD") // until method is implemented
   private void handleEmittedOrchestratorConnectorConfig(final AirbyteControlConnectorConfigMessage configMessage,
                                                         final ConnectorType connectorType) {
-    // TODO: Update config here
-    /**
-     * Pseudocode: for (key in configMessage.getConfig()) { validateIsReallyConfig(key);
-     * persistConfigChange(connectorType, key, configMessage.getConfig().get(key)); // nuance here for
-     * secret storage or not. May need to be async over API for replication orchestrator }
-     */
+    // Currently being persisted as part of the DefaultReplicationWorker
```
davinchia, line 241: note: we shouldn't persist the config in the airbyte message tracker - the tracker is really for stats/metadata tracking and should not have side effects

pedroslopez:
Still TODO is proper integration with the container-orchestrator. The PR that migrated it to micronaut has just been merged - I'll tackle this soon but want to start getting some eyes on the approach 😄

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/AirbyteMessageTracker.java
pedroslopez: Yep, agree! Updated the comment to no longer be a TODO 😛

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -184,7 +190,7 @@ private void replicate(final Path jobRoot,
       // note: `whenComplete` is used instead of `exceptionally` so that the original exception is still
       // thrown
       final CompletableFuture<?> readFromDstThread = CompletableFuture.runAsync(
-          readFromDstRunnable(destination, cancelled, messageTracker, mdc, timeTracker),
+          readFromDstRunnable(destination, cancelled, messageTracker, persistConfigHelper, mdc, timeTracker, Long.valueOf(jobId)),
```
evantahler, line 193: Reasonable information to have in a job!

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -250,8 +259,19 @@ private static Runnable readFromDstRunnable(final AirbyteDestination destination
...
+            messageTracker.acceptFromDestination(message);
+
+            try {
+              if (message.getType() == Type.CONTROL && message.getControl().getType() == AirbyteControlMessage.Type.CONNECTOR_CONFIG) {
+                persistConfigHelper.persistDestinationConfig(jobId, message.getControl().getConnectorConfig().getConfig());
+              }
+            } catch (final ApiException e) {
+              LOGGER.error("Error trying to save updated destination config", e);
+              throw new RuntimeException(e);
+            }
```
evantahler, line 273: 👍 that if we can't persist a config update we should fail the sync.
As a note, I read this as a blocking activity - processing updated configs will block all other message processing.  This is probably fine as it won't happen that often in the sync

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/PersistConfigHelper.java
evantahler: cc @davinchia think we should also have a `PersistStateHelper` and `PersistMetricsAndEstimates` helper like this?

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/PersistConfigHelper.java
```diff
@@ -0,0 +1,90 @@
...
+    final UUID sourceId = connection.getSourceId();
+
+    final SourceApi sourceApi = apiClient.getSourceApi();
+    final SourceRead source = sourceApi.getSource(new SourceIdRequestBody().sourceId(sourceId));
+
+    final SourceRead updatedSource = sourceApi
+        .updateSource(new SourceUpdate()
+            .sourceId(sourceId)
+            .name(source.getName())
+            .connectionConfiguration(Jsons.jsonNode(config.getAdditionalProperties())));
```
evantahler, line 64: I don't see anything in here about authentication... I assume that's handled for us?

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/PersistConfigHelper.java
```diff
@@ -0,0 +1,90 @@
...
+
+    LOGGER.info("Persisted updated configuration for source {}. New config hash: {}.", sourceId,
+        Hashing.sha256().hashString(updatedSource.getConnectionConfiguration().asText(), StandardCharsets.UTF_8));
```
evantahler, line 67: 👍 not logging the actual secrets in the config

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/AirbyteMessageTracker.java
```diff
@@ -238,12 +238,8 @@ private void handleEmittedOrchestratorMessage(final AirbyteControlMessage contro
...
   private void handleEmittedOrchestratorConnectorConfig(final AirbyteControlConnectorConfigMessage configMessage,
                                                         final ConnectorType connectorType) {
-    // TODO: Update config here
-    /**
-     * Pseudocode: for (key in configMessage.getConfig()) { validateIsReallyConfig(key);
-     * persistConfigChange(connectorType, key, configMessage.getConfig().get(key)); // nuance here for
-     * secret storage or not. May need to be async over API for replication orchestrator }
-     */
+    // Config updates are being persisted as part of the DefaultNormalizationWorker.
+    // In the future, we could add tracking of these kinds of messages here. Nothing to do for now.
```
evantahler, line 242: That's a great idea! Story added for this: https://github.com/airbytehq/airbyte/issues/19951

evantahler (commented):
👍 from me! Probably best to wait for someone on the Platform teams to approve

airbyte-container-orchestrator/src/main/resources/application.yml
```diff
@@ -3,4 +3,18 @@ micronaut:
...
+    plane:
+      service-account:
+        credentials-path: ${DATA_PLANE_SERVICE_ACCOUNT_CREDENTIALS_PATH:}
+        email: ${DATA_PLANE_SERVICE_ACCOUNT_EMAIL:}
+  internal:
+    api:
+      auth-header:
+        name: ${AIRBYTE_API_AUTH_HEADER_NAME:}
+        value: ${AIRBYTE_API_AUTH_HEADER_VALUE:}
+      host: ${INTERNAL_API_HOST}
```
pedroslopez, line 20: This probably means we need to add more environment variables to the container-orchestrator. I'm in the process of trying this out on a dev cloud environment and will make a PR over there to add the env vars as needed. Haven't been able to fully test due to some incompatibility with the latest OSS branch contents and Cloud that are currently being worked through.

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/PersistConfigHelper.java
pedroslopez: The ApiClientBeanFactory used by micronaut to initialize the ApiClient takes care of setting up the client with any auth-related / evironment-dependent functionality:

https://github.com/airbytehq/airbyte/blob/c30e8f64002d4042121b3b9b6284d75cfc054ea1/airbyte-commons-worker/src/main/java/io/airbyte/workers/config/ApiClientBeanFactory.java#L33

evantahler: ☕

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -250,8 +259,19 @@ private static Runnable readFromDstRunnable(final AirbyteDestination destination
...
+
```
davinchia, line 274: let's this this into private function named along the lines of `updateConnectorConfig` to maintain the same levels of abstraction here.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -236,8 +243,10 @@ private void replicate(final Path jobRoot,
   private static Runnable readFromDstRunnable(final AirbyteDestination destination,
                                               final AtomicBoolean cancelled,
                                               final MessageTracker messageTracker,
+                                              final PersistConfigHelper persistConfigHelper,
                                               final Map<String, String> mdc,
-                                              final ThreadedTimeTracker timeHolder) {
+                                              final ThreadedTimeTracker timeHolder,
+                                              final Long jobId) {
```
davinchia, line 249: to avoid adding more parameters to this function, can we make this a part of the PersistConfigHelper constructor?

Why I like this:
- this information is already available when the when the PersistConfigHelper is created,
- the PersistConfigHelper is created per job, so it's ok for it to have job specific information.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
davinchia: same thought here on pulling job id into the PersistConfigHelper constructor to simplify the interface

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/PersistConfigHelper.java
davinchia: Yeap having a various persist helpers sounds generally right.

One thought - what about changing the name to `UpdateConnectorConfigHelper`? Update seems better as it's clear we are persisting to update and not persisting for the first time. We would propagate that name change to the functions here too e.g `updateSource` and `updateDestination`. Can probably drop the `config` suffix since `UpdateConnectorConfigHelper.updateSource` is pretty self-explanatory.

Can we also have a better explanation of why in the javadocs? This describes what and not really the why. We want to make it as easy as possible for a readers to piece the context together without talking to someone.

airbyte-commons-worker/build.gradle
```diff
@@ -10,6 +10,7 @@ dependencies {
     implementation libs.bundles.micronaut
 
     implementation 'io.fabric8:kubernetes-client:5.12.2'
+    implementation 'com.auth0:java-jwt:3.19.2'
```
davinchia, line 13: curious, what is this for?

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -307,6 +329,15 @@ private static Runnable readFromSrcAndWriteToDstRunnable(final AirbyteSource sou
...
 
+            if (message.getType() == Type.CONTROL && message.getControl().getType() == AirbyteControlMessage.Type.CONNECTOR_CONFIG) {
+              try {
+                persistConfigHelper.persistSourceConfig(jobId, message.getControl().getConnectorConfig().getConfig());
+              } catch (final ApiException e) {
+                LOGGER.error("Error trying to save updated source config", e);
+                throw new RuntimeException(e);
+              }
+            }
+
```
davinchia, line 340: same though here on pulling this out into a private method.

airbyte-server/src/main/java/io/airbyte/server/handlers/SourceHandler.java
davinchia: for my understanding, how is this related to persisting configs?

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/PersistConfigHelper.java
```diff
@@ -0,0 +1,90 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(PersistConfigHelper.class);
+
+  private final AirbyteApiClient apiClient;
+
+  public PersistConfigHelper(final AirbyteApiClient apiClient) {
+    this.apiClient = apiClient;
+  }
+
+  private UUID getConnectionIdFromJobId(final Long jobId) throws ApiException {
```
davinchia, line 44: We've seen some failures in the api calls from the data plane to the control plane and are still trying to figure out why.

In the mean time, I've added a [retry wrapper in this PR](https://github.com/airbytehq/airbyte/pull/19693/files#diff-81b9db1a46774bfa5dc46a3800a820b7c6596c799822e2012761fb37d65c9eb5R147). We probably want to use this wrapper for all the api calls in this class.

davinchia (commented):
Thanks for the tag @pedroslopez . This looks reasonable.

I left some suggestions. I might not be able to sneak time to take another look during PTO so please feel free to merge in after thinking about these suggestions. We can always revisit in later PR.

I focused my review on the platform classes. I did not scrutinise the tests or the oauth pieces.

I would also like someone from the workflows team to take a look. @gosusnp tagging you since you've expressed interest in side effects during replication. Please tag someone from workflows in if you feel there is a better reviewer.

airbyte-commons-worker/build.gradle
pedroslopez: This was necessary in order to move the ApiClientBeanFactory into commons-worker (to reuse it in both airbyte-workers and container-orchestrator). It does some token generation that needs this dependency.

airbyte-server/src/main/java/io/airbyte/server/handlers/SourceHandler.java
pedroslopez: This is necessary because when we receive a config control message from a connector, it will include those global oauth parameters as part of the config (e.g. our oauth app's client_id + secret). These oauth params are not provided by the user, and are instead injected by the platform. We don't want to persist those as part of the user's secrets when writing back the updated configuration, so we need to mask them again.

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/PersistConfigHelper.java
pedroslopez: I like it! Renamed and added some docs in [bff1d44](https://github.com/airbytehq/airbyte/pull/19811/commits/bff1d44900f3e682c37eb4adc6f9df6db77b2248)

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
pedroslopez: 👍🏻 done in [51f1976](https://github.com/airbytehq/airbyte/pull/19811/commits/51f19765184523e0ae88d8ffceaaad6e39a772e1)

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
pedroslopez: 👍🏻 done in [51f1976](https://github.com/airbytehq/airbyte/pull/19811/commits/51f19765184523e0ae88d8ffceaaad6e39a772e1)

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/PersistConfigHelper.java
pedroslopez: 👍🏻 done in [8203432](https://github.com/airbytehq/airbyte/pull/19811/commits/8203432df13dd43adf060dc7704ce68de0da27ef)

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
pedroslopez: I decided not to add this as part of the helper constructor because I'd like to use this in the other connector commands like check/discover, which don't have a jobId like this. In that follow-up PR I'll pull out some of the logic into a different method for re-use.

airbyte-container-orchestrator/src/main/resources/application.yml
alafanechere, line 20: Could you explain the purpose of this change?

airbyte-server/src/main/java/io/airbyte/server/handlers/SourceHandler.java
alafanechere: Shall we mask these global oAuth parameters on the connector side in the CDK?

alafanechere (approved):
Nice work! Loved the demo.

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/UpdateConnectorConfigHelper.java
```diff
@@ -0,0 +1,115 @@
...
+    LOGGER.info("Persisted updated configuration for source {}. New config hash: {}.", sourceId,
+        Hashing.sha256().hashString(updatedSource.getConnectionConfiguration().asText(), StandardCharsets.UTF_8));
+
+  }
+
+  /**
+   * Updates the Destination from a sync job ID with the provided Configuration. Secrets and OAuth
+   * parameters will be masked when saving.
+   */
+  public void updateDestination(final Long jobId, final Config config) {
```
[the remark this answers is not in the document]
pedroslopez: Yeah, it's tricky because even though Sources/Destinations are equivalent in the DB, they're handled  as completely different in the java classes. Unfortunately there's tons of this kind of duplication in the java codebase for handling sources and destinations. All the methods for interacting with the classes are also type-specific, e.g. `.sourceId`, `.updateSource` etc.

airbyte-container-orchestrator/src/main/resources/application.yml
pedroslopez: This is to be able to initialize and use AirbyteApiClient from the container orchestrator. It's a micronaut configuration file that is used when configuring the API client. An equivalent configuration is present in airbyte-workers.

airbyte-server/src/main/java/io/airbyte/server/handlers/SourceHandler.java
pedroslopez: No, the connector doesn't know what's a global auth param and what's provided directly by the user. Note that in the case of OSS users can provide their own oauth client ids and secrets as part of the config, and global auth params are really only enabled in Cloud. I think considering that the control message contains secrets and handling that on the platform side makes the interface easier to work with, and less to have connector devs worry about.

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/UpdateConnectorConfigHelper.java
```diff
@@ -0,0 +1,115 @@
...
+        "get job info");
+    return UUID.fromString(jobInfo.getJob().getConfigId());
+  }
+
+  /**
+   * Updates the Source from a sync job ID with the provided Configuration. Secrets and OAuth
+   * parameters will be masked when saving.
+   */
+  public void updateSource(final Long jobId, final Config config) {
+    final UUID connectionId = getConnectionIdFromJobId(jobId);
```
gosusnp, line 61: This part feels very inefficient, taking a step back, when we setup the jobs, we have all the pieces, `connectionId`, `sourceId`, `destinationId`, *but* since we only pass `jobId`, we need to have an extra API call to retrieve the sourceId/destinationId.
Can we pass those instead of just the jobId and remove the need of the `connectionApi`.

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/UpdateConnectorConfigHelper.java
```diff
@@ -0,0 +1,115 @@
...
+ * useful for migrating configuration to a new version or for enabling connectors that require
+ * single-use or short-lived OAuth tokens.
+ */
+public class UpdateConnectorConfigHelper {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(UpdateConnectorConfigHelper.class);
+
+  private final AirbyteApiClient apiClient;
+
+  public UpdateConnectorConfigHelper(final AirbyteApiClient apiClient) {
```
gosusnp, line 43: We're moving towards a more fine grained injection, can we injection `sourceApi` and `destinationApi` rather than the `AirbyteApiClient`? (I have a comment down that aims at removing the dependency on `connectionApi`

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/UpdateConnectorConfigHelper.java
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Helper class for workers to persist updates to Source/Destination configs emitted from
+ * AirbyteControlMessages.
+ *
+ * This is in order to support connectors updating configs when running commands, which is specially
+ * useful for migrating configuration to a new version or for enabling connectors that require
+ * single-use or short-lived OAuth tokens.
+ */
+public class UpdateConnectorConfigHelper {
```
gosusnp, line 37: nit: is this just a `ConnectorConfigUpdater` rather than Helper?

gosusnp (commented):
Change overall looks good, most of my comments are around the config update piece.

airbyte-commons-worker/build.gradle
pedroslopez: @gosusnp Could you expand a bit on what the issue might be? My change moved the `AirbyteAPiClientBeanFactory` to `airbyte-commons-worker`. I don't see `airbyte-workers` nor `airbyte-commons-worker` being a dependency of `airbyte-cron`.

gosusnp: I was looking for an issue to link for better context but I cannot find it...
I think the problem we had is that micronaut discovers all singleton upon starts, so by having some services in a shared lib led us to start more services than intended. For example, the cron would pick up some beans from the workers and would effectively behave as an extra worker.

@jdpgrailsdev or @benmoriceau might have better context since they were more closely working on this.

airbyte-commons-worker/src/main/java/io/airbyte/workers/config/ApiClientBeanFactory.java
```diff
@@ -67,6 +68,11 @@ public SourceApi sourceApi(final ApiClient apiClient) {
     return new SourceApi(apiClient);
   }
 
+  @Singleton
+  public DestinationApi destinationApi(final ApiClient apiClient) {
+    return new DestinationApi(apiClient);
+  }
```
pedroslopez, line 74: Added so we have the singleton available for injection directly - followed what was already being done for `SourceApi`

airbyte-config/config-models/src/main/resources/types/StandardSyncInput.yaml
```diff
@@ -19,6 +21,14 @@ properties:
...
     description: Prefix that will be prepended to the name of each stream when it is written to the destination.
     type: string
+  sourceId:
+    description: Actor ID for the source used in the sync - this is used to update the actor configuration when requested.
+    type: string
+    format: uuid
+  destinationId:
+    description: Actor ID for the destination used in the sync - this is used to update the actor configuration when requested.
+    type: string
+    format: uuid
```
pedroslopez, line 31: sync inputs now have sourceId and destinationId so we can refer to them directly when updating configs

airbyte-container-orchestrator/src/main/resources/application.yml
pedroslopez: See https://github.com/airbytehq/airbyte/pull/20724 and https://github.com/airbytehq/airbyte-cloud/pull/3902

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityImpl.java
```diff
@@ -90,25 +90,18 @@ public GeneratedJobInput getSyncWorkflowInput(final SyncInput input) {
...
+
+      final StandardDestinationDefinition destinationDefinition =
+          configRepository.getDestinationDefinitionFromDestination(standardSync.getDestinationId());
+      final String destinationNormalizationDockerImage = destinationDefinition.getNormalizationConfig() != null
+          ? DockerUtils.getTaggedImageName(destinationDefinition.getNormalizationConfig().getNormalizationRepository(),
+              destinationDefinition.getNormalizationConfig().getNormalizationTag())
+          : null;
+      final String normalizationIntegrationType =
+          destinationDefinition.getNormalizationConfig() != null ? destinationDefinition.getNormalizationConfig().getNormalizationIntegrationType()
+              : null;
```
pedroslopez, line 104: While I was in here, I refactored how we were looking up the destination definition from the sync. We need to get the StandardSync anyway to retrieve the sourceId and destinationId for config updating now, so we can simplify to get the definition from the destinationId directly. 

This used to be looking up all destination definitions and filtering based on docker image + tag combo, which also seemed a bit error prone if we decide to move into a world where multiple definitions could have the same docker image (which I think _will_ be the case)

The functionality should be equivalent.

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/UpdateConnectorConfigHelper.java
pedroslopez: Renamed in https://github.com/airbytehq/airbyte/pull/19811/commits/c20f61e673363d293d33ea3ccf7b5c6fb38c0fea

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/UpdateConnectorConfigHelper.java
pedroslopez: Done in https://github.com/airbytehq/airbyte/pull/19811/commits/6ecac44ac2f7c0d1e2ea4be33cb5079fce913b00

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/UpdateConnectorConfigHelper.java
pedroslopez: Agree! I've moved to specifying the sourceId and destinationId as part of the StandardSyncInput and passing that along so we have the information available. This means we can call get/update directly using these IDs rather than have to look up the job, then the connection, then the source/destination.

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/UpdateConnectorConfigHelper.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  /**
+   * Updates the Source from a sync job ID with the provided Configuration. Secrets and OAuth
+   * parameters will be masked when saving.
+   */
+  public void updateSource(final Long jobId, final Config config) {
+    final UUID connectionId = getConnectionIdFromJobId(jobId);
+
+    final ConnectionRead connection = AirbyteApiClient.retryWithJitter(
```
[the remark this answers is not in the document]
pedroslopez: Discussed offline - we will be blocking for now and revisit later since control messages should not be a common thing during the sync (_maybe_ 1 per hour depending on the connector)

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -306,6 +319,7 @@ private static Runnable readFromSrcAndWriteToDstRunnable(final AirbyteSource sou
             final AirbyteMessage message = mapper.mapMessage(airbyteMessage);
 
             messageTracker.acceptFromSource(message);
+            acceptAndUpdateSrcConfig(jobId, message, updateConnectorConfigHelper);
```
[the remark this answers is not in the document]
pedroslopez: Updated to `acceptSrcControlMessage` and `acceptDstControlMessage` methods so we know what's. being passed through

pedroslopez:
@gosusnp There's a separate piece in https://github.com/airbytehq/airbyte/pull/20724 that will be merged into this branch to set things up properly for the container orchestrator in kube environments, but I've addressed your comments and would love another look!

Notably the sourceId and destinationId are now being passed as part of the StandardSyncInput.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -267,8 +276,14 @@ private static Runnable readFromDstRunnable(final AirbyteDestination destination
...
           if (messageOptional.isPresent()) {
-            LOGGER.info("State in DefaultReplicationWorker from destination: {}", messageOptional.get());
-            messageTracker.acceptFromDestination(messageOptional.get());
+            final AirbyteMessage message = messageOptional.get();
+            LOGGER.info("State in DefaultReplicationWorker from destination: {}", message);
+
+            messageTracker.acceptFromDestination(message);
+
+            if (message.getType() == Type.CONTROL) {
+              acceptDstControlMessage(destinationConfig, message.getControl(), connectorConfigUpdater);
```
gosusnp, line 285: We may want to have a `try...catch` here, otherwise, if we fail to update the config, we'd end up failing the sync.
Same comment for the source.

gosusnp (commented):
Looks good, one last question about how we want to handle config update errors.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
pedroslopez: Originally I had intentionally made syncs fail if we couldn't update the config, but after thinking about this some more I agree we shouldn't fail because:

- Syncing data is our top priority and what we care about. We should keep moving data.
- These messages would be emitted for two main reasons at the moment: config migrations and handling short-lived or one-time-use tokens for oauth. 
    - For config migrations, we should be able to try again on the next sync. 
    - For the OAuth case, we could see another CONTROL message later on that might get correctly updated, or if it's really an issue we'll get an error on the next sync anyway (so we should let the current sync run its course).

The benefit of failing is we know right away something didn't go quite right, and we would want to know about these errors. However it might be better surfaced as some kind of notice rather than a full failure.

Updated to just log the error in [e7f4736](https://github.com/airbytehq/airbyte/pull/19811/commits/e7f4736e343d0ccc2fc1b8716eda30588a462fa8)

gosusnp (approved):

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -391,6 +420,22 @@ private static Runnable readFromSrcAndWriteToDstRunnable(final AirbyteSource sou
     };
   }
 
+  private static void acceptSrcControlMessage(final WorkerSourceConfig sourceConfig,
+                                              final AirbyteControlMessage controlMessage,
+                                              final ConnectorConfigUpdater connectorConfigUpdater) {
+    if (controlMessage.getType() == AirbyteControlMessage.Type.CONNECTOR_CONFIG) {
```
davinchia, line 426: are there other kinds of control messages today?

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -391,6 +420,22 @@ private static Runnable readFromSrcAndWriteToDstRunnable(final AirbyteSource sou
     };
   }
 
+  private static void acceptSrcControlMessage(final WorkerSourceConfig sourceConfig,
```
davinchia, line 423: we can simplify this method signature by passing in the id instead of the entire config

davinchia: this simplification also extends to the various runnable method interfaces e.g ` readFromDstRunnable` doesn't need the `destinationConfig`. Improves readability since it's clearer what scope requires what data.

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/ConnectorConfigUpdater.java
```diff
@@ -0,0 +1,87 @@
...
+import io.airbyte.api.client.model.generated.SourceRead;
+import io.airbyte.api.client.model.generated.SourceUpdate;
+import io.airbyte.commons.json.Jsons;
+import io.airbyte.protocol.models.Config;
+import java.nio.charset.StandardCharsets;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
```
davinchia, line 24: +1 for the javadocs!

davinchia (approved):
Thanks Pedro!

Some minor comments on the code readability in the DefaultReplicationWorker class. Otherwise looks good!

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
pedroslopez: 👍🏻 updated in [d6eb05f](https://github.com/airbytehq/airbyte/pull/19811/commits/d6eb05f4a9a4edfee6c566ed888a7711f9c41449)

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
pedroslopez: Nope, this is the only one for now!

PR #19820, opened 2022-11-28
SAT: stricter integer validation

### LINKED ISSUE #19800: SAT: perform extra validation on integer fields.
# What
SAT perform schema validation of records against the catalog schema with the `jsonschema` library.
According to `jsonschema` a float value with a zero fractional part is a valid `integer` : 

```bash
>>> from jsonschema import Draft7Validator
>>> schema = {'$schema': 'http://json-schema.org/draft-07/schema#', 'type': 'object', 'properties': {'metrics.conversion': {'type': ['null', 'integer']},}}
>>> record = {'metrics.conversion': 1.0}
>>> Draft7Validator(schema).is_valid(record)
True
>>> record = {'metrics.conversion': 1.1}
>>> Draft7Validator(schema).is_valid(record)
False
```

This relaxed definition is problematic because database destinations are more strict in terms of typing.
This situation prevents us from identifying schema inconsistencies in SAT when a field is declared as `integers` but the actual record values in our test account are `1.0, 2.0, 3.0` etc. It led this [PR](https://github.com/airbytehq/airbyte/pull/18069) that introduced a schema inconsistency to be merged with successful acceptance tests.

# How
A. Implement a custom check in the [verify_records_schema  function](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/utils/asserts.py#L42) to make sure that a field defined as integer is an integer in python term's...
B. Find a way to force this validation with `jsonschema`

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/19800

According to JSON schema a float value with a zero fractional part is a valid integer. We don't want to have this behavior and make field that are declared as `integer` fail if their value is `1.0, 2.0` etc...

## How
Leverage `jsonschema` [capabilities](https://python-jsonschema.readthedocs.io/en/latest/creating/) at redefining type checker of validators, and extending the off the shelf validators (`Draft7` in our case).

### DIFF
--- a/airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/utils/asserts.py
+++ b/airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/utils/asserts.py
@@ -67,7 +67,3 @@ def verify_records_schema(
             stream_errors[record.stream][str(error.schema_path)] = error
 
     return stream_errors
-
-
-def is_int(checker, value):
-    return isinstance(value, int)
--- a/airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/utils/asserts.py
+++ b/airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/utils/asserts.py
@@ -9,7 +9,7 @@ from typing import List, Mapping
 
 import pendulum
 from airbyte_cdk.models import AirbyteRecordMessage, ConfiguredAirbyteCatalog
-from jsonschema import Draft7Validator, FormatChecker, FormatError, ValidationError
+from jsonschema import Draft7Validator, FormatChecker, FormatError, ValidationError, validators
 
 # fmt: off
 timestamp_regex = re.compile((r"^\d{4}-\d?\d-\d?\d"  # date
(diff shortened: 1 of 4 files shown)


bazarnov (approved):

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/utils/asserts.py
```diff
@@ -18,6 +18,11 @@
                               r".*$"))  # timezone
 # fmt: on
 
+# In Json schema, numbers with a zero fractional part are considered integers. E.G. 1.0 is considered a valid integer
+# For stricter type validation we don't want to keep this behavior. We want to consider integers in the Pythonic way.
+strict_integer_type_checker = Draft7Validator.TYPE_CHECKER.redefine("integer", lambda _, value: isinstance(value, int))
```
pedroslopez, line 23: oh nice, it's cool that jsonschema lets you redefine the validators like this 👍🏼

pedroslopez (approved):
LGTM!

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/utils/asserts.py
```diff
@@ -62,3 +67,7 @@ def verify_records_schema(
             stream_errors[record.stream][str(error.schema_path)] = error
 
     return stream_errors
+
+
+def is_int(checker, value):
+    return isinstance(value, int)
```
erohmensing, line 73:
```suggestion
```
Is this still used anywhere? Looks leftover

erohmensing (approved):
Jsonschema will truly be the death of us all. Looks good

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/utils/asserts.py
alafanechere: Thanks for spotting this :)

alafanechere:

PR #19824, opened 2022-11-28
Helm Chart: fix indentation of securityContext block

proprefenetre:
## What
Fix indentation of the `securityContext` block

## How
Decrease indentation with one level.


marcosmarxm (approved):
Thanks @proprefenetre as described here https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ the security context belongs to container level.

PR #19826, opened 2022-11-28
🪟 🔧 Add ConnectorIds utility

timroes:
## What

Adds a `ConnectorIds` map to `utils/connectors` to make sure we're not hard-coding connector specification ids all around the code base. Also adds a eslint rule that fails if a connector ID would be used hard-coded in the code anywhere.


airbyte-webapp/package.json
```diff
@@ -81,6 +81,7 @@
     "yup": "^0.32.11"
   },
   "devDependencies": {
+    "@airbyte/eslint-plugin": "file:./packages/eslint-plugin",
```
timroes, line 84: :information_source: Add a local eslint plugin, so we can easier write custom linting rules in the future.

airbyte-webapp/scripts/classname-serializer.js
```diff
@@ -1,8 +1,7 @@
 import { prettyDOM } from "@testing-library/react";
 
 /**
- * Traverse a tree of nodes and replace all class names with
- * the count of classnames instead, e.g. "<3 classnames>"
+ * Traverse a tree of nodes and replace all class names with "<removed-for-snapshot-test>"
```
timroes, line 4: :information_source: Unrelated but very small fix in the docs, I noticed while writing this.

airbyte-webapp/src/utils/connectors/destinations.json
```diff
@@ -0,0 +1,50 @@
+{
```
timroes, line 1: :information_source: This (and the sources.json) are JSON files, so I can import them in JS in the eslint rule, since I can't import TypeScript there, unless setting up a full tyepscript build for that plugin as well.

airbyte-webapp/src/core/domain/connector/constants.ts
```diff
@@ -15,26 +16,26 @@ export const DEV_IMAGE_TAG = "dev";
...
+        ConnectorIds.Destinations.MariaDbColumnStore, // hide MariaDB Destination https://github.com/airbytehq/airbyte-cloud/issues/2611
+        ConnectorIds.Destinations.Mqtt, // hide MQTT Destination https://github.com/airbytehq/airbyte-cloud/issues/2613
+        ConnectorIds.Destinations.Pulsar, // hide Pulsar Destination https://github.com/airbytehq/airbyte-cloud/issues/2614
+        ConnectorIds.Destinations.Rockset, // hide Rockset Destination https://github.com/airbytehq/airbyte-cloud/issues/2615
+        ConnectorIds.Sources.SalesforceSinger, // Salesforce Singer
+        ConnectorIds.Destinations.Scylla, // hide Scylla Destination https://github.com/airbytehq/airbyte-cloud/issues/2617
+        ConnectorIds.Destinations.MeiliSearch, // hide MeiliSearch Destination https://github.com/airbytehq/airbyte/issues/16313
+        ConnectorIds.Destinations.RabbitMq, // hide RabbitMQ Destination https://github.com/airbytehq/airbyte/issues/16315
+        ConnectorIds.Destinations.AmazonSqs, // hide Amazon SQS Destination https://github.com/airbytehq/airbyte/issues/16316
         ...(workspaceId !== "54135667-ce73-4820-a93c-29fe1510d348" // Shopify workspace for review
```
krishnaglick, line 33: Feels strange to have this workspace id sitting in here.

airbyte-webapp/src/utils/connectors/destinations.json
krishnaglick: Is there any chance these GUIDs change elsewhere?

airbyte-webapp/src/core/domain/connector/constants.ts
timroes: totally, didn't want to change this in this PR though.

airbyte-webapp/src/utils/connectors/destinations.json
lmossman: +1 to what Tim said - I don't believe these ever change and if they did it would break other parts of the system as well.

airbyte-webapp/src/utils/connectors/destinations.json
```diff
@@ -0,0 +1,50 @@
...
+  "AzureBlobStorage": "b4c5d105-31fd-4817-96b6-cb923bfc04cb",
+  "AmazonSqs": "0eeee7fb-518f-4045-bacc-9619e31c43ea",
```
lmossman, line 3: It would be nicer if we could pull this mapping directly from the [source_definitions.yaml](https://github.com/airbytehq/airbyte/blob/master/airbyte-config/init/src/main/resources/seed/source_definitions.yaml) and [destination_definitions.yaml](https://github.com/airbytehq/airbyte/blob/master/airbyte-config/init/src/main/resources/seed/destination_definitions.yaml) files, since those are the real source of truth for these IDs, and that way we wouldn't need to maintain these separate JSON files or add to them whenever a new connector is added.

However, I think pulling YAML files in is harder than JSON, and those files are outside the airbyte-webapp which makes them harder to pull in as well. Maybe we can make a symlink, and use some library to convert to JSON?

airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/auth/AuthButton.tsx
```diff
@@ -14,19 +15,21 @@ import GoogleAuthButton from "./GoogleAuthButton";
...
+      ConnectorIds.Sources.GoogleAnalyticsUniversalAnalytics,
+      ConnectorIds.Sources.GoogleDirectory,
+      ConnectorIds.Sources.GoogleSearchConsole,
+      ConnectorIds.Sources.GoogleSheets,
+      ConnectorIds.Sources.GoogleWorkspaceAdminReports,
+      ConnectorIds.Sources.YouTubeAnalytics,
+      ConnectorIds.Destinations.GoogleSheets,
+      // TODO: revert me
+      ConnectorIds.Sources.YouTubeAnalyticsBusiness, // Youtube analytics Business
+      //
```
lmossman, line 30:
```suggestion
      ConnectorIds.Sources.YouTubeAnalyticsBusiness,
```
any reason to keep those comments?

airbyte-webapp/src/test-utils/mock-data/mockConnection.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable no-template-curly-in-string */
+/* eslint-disable no-template-curly-in-string, @airbyte/no-hardcoded-connector-ids */
```
lmossman, line 1: Why disable the rule here instead of using the ConnectorIds object?

airbyte-webapp/packages/eslint-plugin/no-hardcoded-connector-ids.js
```diff
@@ -0,0 +1,38 @@
...
+
+// Create a map for connector id to variable name (i.e. flip the direction of the map)
+const sourceIds = Object.fromEntries(Object.entries(sources).map((entry) => [entry[1], entry[0]]));
+const destinationIds = Object.fromEntries(Object.entries(destinations).map((entry) => [entry[1], entry[0]]));
+
+const validateStringContent = (context, node, string) => {
+  if (string in destinationIds) {
+    context.report({ node, messageId: "destinationId", data: { id: string, name: destinationIds[string] } });
+  } else if (string in sourceIds) {
+    context.report({ node, messageId: "sourceId", data: { id: string, name: sourceIds[string] } });
```
lmossman, line 12:
```suggestion
const sourceIdToName = Object.fromEntries(Object.entries(sources).map((entry) => [entry[1], entry[0]]));
const destinationIdToName = Object.fromEntries(Object.entries(destinations).map((entry) => [entry[1], entry[0]]));

const validateStringContent = (context, node, string) => {
  if (string in destinationIdToName) {
    context.report({ node, messageId: "destinationId", data: { id: string, name: destinationIdToName[string] } });
  } else if (string in sourceIdToName) {
    context.report({ node, messageId: "sourceId", data: { id: string, name: sourceIdToName[string] } });
```
nit: rename variables to `xIdToName` instead of just `xId`

airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/auth/AuthButton.tsx
timroes: I think this is still meant to be removed, and currently only enabled temporarily, for Google to validate the connector. My understanding was, that the plan is to remove that connector altogether again.

airbyte-webapp/src/utils/connectors/destinations.json
timroes: Good call. I'll try to play around. We could potentially symlink as you mentioned, and use a yaml webpack loader to import them.

timroes: Okay problem here is: if we go this way, we'd lose strong typing on `ConnectorIds.Sources/Destinations`, i.e. because it's not coming from a JSON, we'd either need to manually type it, so it knows which connector names exist on it (which can then also always run async with the actual content of the yaml), or just type it as `{ [key: string]: string }`, which basically means TS won't be able to check anymore if the key you're trying to access actually exists.

I personally would find the loss of the typing here more severe than having this file separate from the specification yamls. We also don't NEED to keep those JSONs constantly up to date, since we mainly need to refresh them in case we actually want to use a (new) connector's id in code, that isn't yet in the JSON. I'd not necessarily try to keep it constantly updated when adding new connectors, if we're not accessing it's ID anywhere.

What do you think?

airbyte-webapp/packages/eslint-plugin/no-hardcoded-connector-ids.js
```diff
@@ -2,14 +2,14 @@ const destinations = require("../../src/utils/connectors/destinations.json");
 const sources = require("../../src/utils/connectors/sources.json");
 
 // Create a map for connector id to variable name (i.e. flip the direction of the map)
-const sourceIds = Object.fromEntries(Object.entries(sources).map((entry) => [entry[1], entry[0]]));
-const destinationIds = Object.fromEntries(Object.entries(destinations).map((entry) => [entry[1], entry[0]]));
+const sourceIdToName = Object.fromEntries(Object.entries(sources).map((entry) => [entry[1], entry[0]]));
+const destinationIdToName = Object.fromEntries(Object.entries(destinations).map((entry) => [entry[1], entry[0]]));
 
 const validateStringContent = (context, node, string) => {
```
krishnaglick, line 8: Nit: Can we call this something other than string?

timroes: Happy for any suggestions :D Couldn't come up with anything better, since this is literally just the content of any `string` in the UI, we don't have any more knowledge what this is or isn't.

krishnaglick: Maybe `nodeContent`?

airbyte-webapp/src/test-utils/mock-data/mockConnection.ts
timroes: My original thought was, that these are just test mocks, we're really not caring that those are real IDs in there. That said there's also absolutely not reason NOT to use it here, so I've changed it.

airbyte-webapp/src/utils/connectors/destinations.json
lmossman: That's a good point - the strong typing here seems useful, and we don't need to add an entry until it actually needs to be used somewhere in the code, which I don't expect to happen _too_ often.

So I think I'm fine with this approach

airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/auth/AuthButton.tsx
lmossman: I think it makes sense to keep the `// TODO: revert me` comment, but I just meant we can remove the `// Youtube analytics Business` comment, because we now have the name of it in the ConnectorIds value reference, and we can also remove that empty `//` on the line below

lmossman (approved):
Just one outstanding comment about removing some unnecessary comments in the code, otherwise LGTM.

Did not test locally.

airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/auth/AuthButton.tsx
timroes: Ah thank you. Misunderstood that. Yeah they are useless now. Will remove.

krishnaglick (approved):

PR #19840, opened 2022-11-28
Optimize all connector icons

timroes:
## What

Puts all our SVG connector icons through `svgo` to optimize their size (e.g. remove unnecessary editor specific tags and attributes, etc.). This should not have any affect on the actual rendering of the icon.


krishnaglick (approved):
LGTM

Upmitt (approved):

Issue #3608, opened 2021-05-26
Add total count of connectors to connector dashboard


sherifnada:
## Tell us about the problem you're trying to solve
We'd like to have a very clear indication of how many connectors we have shipping with Airbyte out of the box. 

The easiest way to do this is to query `source_definitions.yaml` and `destination_definitions.yaml` and count the number of entries in them (or count the number of files in the corresponding JSON file directories).

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200368060087111/1200382956041760) by [Unito](https://www.unito.io)

sherifnada:
done in data studio

marcosmarxm:
you are awesome

PR #19845, opened 2022-11-28
fix: add action that remove unattended-upgrades from the os [#3608]

xpuska513:
## What
* Remove unattended-upgrades package from ubuntu so it won't lock apt frontend during workflow execution
* Fixes https://github.com/airbytehq/airbyte-cloud/issues/3608


supertopher (approved):
this looks like a decent change based on the slack context, what's with the failing tests though?

PR #19847, opened 2022-11-28
Bmoric/convert attempt micronaut

benmoriceau:
## What
Convert the Attempt Controller to micronaut.
This allow all post request to not require authentications

## How
Use micronaut annotations


jdpgrailsdev (approved):
:shipit:

PR #19851, opened 2022-11-28
exclude integrations from airbyte-maintainers group

cgardens:


cgardens:
merging without code review as this adjusts a codowners to which only i am subscribed.

PR #19860, opened 2022-11-29
Move handling TIMESTAMP_WITH_TIMEZONE to common JdbcSourceOperations

### LINKED ISSUE #16838: `JdbcSourceOperations` method `setJsonField` misses a branch for `TIMESTAMP_WITH_TIMEZONE`
## Current Behavior
- Currently `TIMESTAMP_WITH_TIMEZONE` is not handled by `JdbcSourceOperations#setJsonField`. This time falls under the default branch, and the output is not the same as what we want (e.g. it does not have microsecond precision).
- This issue is discovered in https://github.com/airbytehq/airbyte/issues/16552 and https://github.com/airbytehq/airbyte/pull/16766.
  - When the Snowflake JDBC version is bumped from `3.13.9` to `3.13.22`, the timestamp with timezone column type from JDBC changed from `timestamp` to `timestamp with timezone`. The hypothesis is that the old version of the Snowflake JDBC is not returning the `timestamp` column type correctly. But it is unclear which version fixed this behavior according to the Snowflake JDBC [changelog](https://github.com/snowflakedb/snowflake-jdbc/blob/master/CHANGELOG.rst).
  - Because `JdbcSourceOperations` is processing `timestamp with timezone` with the default branch, the output changed from `2018-03-22T07:00:00.123000Z` to `2018-03-22 12:00:00.123 +0500`.
  - The fix in https://github.com/airbytehq/airbyte/pull/16766 is to override the `setJsonField` method in `SnowflakeSourceOperations`, and handle both types with the `putTimestamp` method, which works for timestamp:
    ```java
    case TIMESTAMP, TIMESTAMP_WITH_TIMEZONE -> putTimestamp(json, columnName, resultSet, colIndex);
    ```
  - The fix is limited to Snowflake because we don't want to increase the scope of that PR.
- The same issue could happen to `time with timestamp` as well, because that is also a standard JDBC type, and currently it falls under the default branch.

## Expected Behavior
- `timestamp with timezone` type should be handled correctly for all JDBC sources.
- We need to investigate whether the solution for Snowflake works for other JDBC sources.

akashkulk:
Closes https://github.com/airbytehq/airbyte/issues/16838


VitaliiMaltsev (approved):
LGTM

rodireich (approved):

ryankfu (approved):

PR #19864, opened 2022-11-29
Allow resets for disabled connections

### LINKED ISSUE #19861: Don't skip reset when saving and resetting disabled connection
Allow disabled connections to reset. Just need to remove the check from here:

https://github.com/airbytehq/airbyte/blob/5755dd84b9809b6851a05492f02b771a484c0cb6/airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionReplicationTab.tsx#L124

alovew:
Resolves #19861 

This is fixing a bug that was introduced during the connection form refactor. skipReset should be set to false if there is any change to the configured catalog, regardless of whether the connection is disabled.


krishnaglick (approved):
I pushed a small linting fix so it builds. Code changes LGTM

edmundito (approved):

PR #19882, opened 2022-11-29
Invoke handleChatUs button

krishnaglick:
## What
A change was introduced that broke the Chat With Us button a few weeks ago.

## How
This is fixed by correctly invoking the chat with us call when expected.


timroes (approved):
Code LGTM. Have not tested locally.

PR #19888, opened 2022-11-29
refresh before syncs when feature flag is on

alovew:
Refresh schemas before each sync using feature flag

The main logic here is in `SyncWorkflowImpl.java`

### DIFF
--- a/airbyte-workers/src/main/java/io/airbyte/workers/config/ApiClientBeanFactory.java
+++ b/airbyte-workers/src/main/java/io/airbyte/workers/config/ApiClientBeanFactory.java
@@ -10,6 +10,7 @@ import com.auth0.jwt.algorithms.Algorithm;
 import com.google.auth.oauth2.ServiceAccountCredentials;
 import io.airbyte.api.client.AirbyteApiClient;
 import io.airbyte.api.client.generated.SourceApi;
+import io.airbyte.api.client.invoker.generated.ApiClient;
 import io.airbyte.commons.temporal.config.WorkerMode;
 import io.micronaut.context.BeanProvider;
 import io.micronaut.context.annotation.Factory;
@@ -36,49 +37,34 @@ public class ApiClientBeanFactory {
   private static final int JWT_TTL_MINUTES = 5;
 
   @Singleton
-  public AirbyteApiClient airbyteApiClient(
-                                           @Value("${airbyte.internal.api.auth-header.name}") final String airbyteApiAuthHeaderName,
-                                           @Value("${airbyte.internal.api.host}") final String airbyteApiHost,
-                                           @Named("internalApiAuthToken") final BeanProvider<String> internalApiAuthToken,
-                                           @Named("internalApiScheme") final String internalApiScheme) {
-    return new AirbyteApiClient(
-        new io.airbyte.api.client.invoker.generated.ApiClient()
-            .setScheme(internalApiScheme)
-            .setHost(parseHostName(airbyteApiHost))
-            .setPort(parsePort(airbyteApiHost))
-            .setBasePath("/api")
-            .setHttpClientBuilder(HttpClient.newBuilder().version(Version.HTTP_1_1))
-            .setRequestInterceptor(builder -> {
-              builder.setHeader("User-Agent", "WorkerApp");
-              // internalApiAuthToken is in BeanProvider because we want to create a new token each
-              // time we send a request.
-              if (!airbyteApiAuthHeaderName.isBlank()) {
-                builder.setHeader(airbyteApiAuthHeaderName, internalApiAuthToken.get());
-              }
-            }));
-  }
-
-  @Singleton
-  public SourceApi sourceApi(
-                             @Value("${airbyte.internal.api.auth-header.name}") final String airbyteApiAuthHeaderName,
+  public ApiClient apiClient(@Value("${airbyte.internal.api.auth-header.name}") final String airbyteApiAuthHeaderName,
                              @Value("${airbyte.internal.api.host}") final String airbyteApiHost,
                              @Named("internalApiAuthToken") final BeanProvider<String> internalApiAuthToken,
                              @Named("internalApiScheme") final String internalApiScheme) {
-    return new SourceApi(
-        new io.airbyte.api.client.invoker.generated.ApiClient()
-            .setScheme(internalApiScheme)
-            .setHost(parseHostName(airbyteApiHost))
-            .setPort(parsePort(airbyteApiHost))
-            .setBasePath("/api")
-            .setHttpClientBuilder(HttpClient.newBuilder().version(Version.HTTP_1_1))
-            .setRequestInterceptor(builder -> {
-              builder.setHeader("User-Agent", "WorkerApp");
-              // internalApiAuthToken is in BeanProvider because we want to create a new token each
-              // time we send a request.
-              if (!airbyteApiAuthHeaderName.isBlank()) {
-                builder.setHeader(airbyteApiAuthHeaderName, internalApiAuthToken.get());
-              }
-            }));
+    return new io.airbyte.api.client.invoker.generated.ApiClient()
+        .setScheme(internalApiScheme)
+        .setHost(parseHostName(airbyteApiHost))
+        .setPort(parsePort(airbyteApiHost))
+        .setBasePath("/api")
+        .setHttpClientBuilder(HttpClient.newBuilder().version(Version.HTTP_1_1))
+        .setRequestInterceptor(builder -> {
+          builder.setHeader("User-Agent", "WorkerApp");
+          // internalApiAuthToken is in BeanProvider because we want to create a new token each
+          // time we send a request.
+          if (!airbyteApiAuthHeaderName.isBlank()) {
+            builder.setHeader(airbyteApiAuthHeaderName, internalApiAuthToken.get());
+          }
+        });
+  }
+
+  @Singleton
+  public AirbyteApiClient airbyteApiClient(ApiClient apiClient) {
+    return new AirbyteApiClient(apiClient);
+  }
+
+  @Singleton
+  public SourceApi sourceApi(final ApiClient apiClient) {
+    return new SourceApi(apiClient);
   }
 
   @Singleton
--- a/airbyte-workers/src/main/java/io/airbyte/workers/config/ApiClientBeanFactory.java
+++ b/airbyte-workers/src/main/java/io/airbyte/workers/config/ApiClientBeanFactory.java
@@ -9,6 +9,7 @@ import com.auth0.jwt.JWTCreator;
 import com.auth0.jwt.algorithms.Algorithm;
 import com.google.auth.oauth2.ServiceAccountCredentials;
 import io.airbyte.api.client.AirbyteApiClient;
+import io.airbyte.api.client.generated.SourceApi;
 import io.airbyte.commons.temporal.config.WorkerMode;
 import io.micronaut.context.BeanProvider;
 import io.micronaut.context.annotation.Factory;
@@ -41,6 +42,29 @@ public class ApiClientBeanFactory {
                                            @Named("internalApiAuthToken") final BeanProvider<String> internalApiAuthToken,
                                            @Named("internalApiScheme") final String internalApiScheme) {
     return new AirbyteApiClient(
+        new io.airbyte.api.client.invoker.generated.ApiClient()
+            .setScheme(internalApiScheme)
+            .setHost(parseHostName(airbyteApiHost))
+            .setPort(parsePort(airbyteApiHost))
+            .setBasePath("/api")
+            .setHttpClientBuilder(HttpClient.newBuilder().version(Version.HTTP_1_1))
+            .setRequestInterceptor(builder -> {
+              builder.setHeader("User-Agent", "WorkerApp");
+              // internalApiAuthToken is in BeanProvider because we want to create a new token each
+              // time we send a request.
+              if (!airbyteApiAuthHeaderName.isBlank()) {
+                builder.setHeader(airbyteApiAuthHeaderName, internalApiAuthToken.get());
+              }
+            }));
+  }
+
+  @Singleton
+  public SourceApi sourceApi(
+                             @Value("${airbyte.internal.api.auth-header.name}") final String airbyteApiAuthHeaderName,
+                             @Value("${airbyte.internal.api.host}") final String airbyteApiHost,
+                             @Named("internalApiAuthToken") final BeanProvider<String> internalApiAuthToken,
+                             @Named("internalApiScheme") final String internalApiScheme) {
+    return new SourceApi(
         new io.airbyte.api.client.invoker.generated.ApiClient()
             .setScheme(internalApiScheme)
             .setHost(parseHostName(airbyteApiHost))
--- a/airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/RefreshSchemaActivityImpl.java
+++ b/airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/RefreshSchemaActivityImpl.java
@@ -10,6 +10,7 @@ import datadog.trace.api.Trace;
 import io.airbyte.api.client.AirbyteApiClient;
 import io.airbyte.api.client.invoker.generated.ApiException;
 import io.airbyte.api.client.model.generated.SourceDiscoverSchemaRequestBody;
+import io.airbyte.commons.features.EnvVariableFeatureFlags;
 import io.airbyte.config.ActorCatalogFetchEvent;
 import io.airbyte.config.persistence.ConfigRepository;
 import jakarta.inject.Singleton;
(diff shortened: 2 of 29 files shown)


airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/RefreshSchemaActivityImpl.java
```diff
@@ -40,10 +44,15 @@ public boolean shouldRefreshSchema(UUID sourceCatalogId) throws IOException {
...
+  public void refreshSchema(UUID sourceCatalogId, UUID connectionId) throws ApiException {
     SourceDiscoverSchemaRequestBody requestBody =
-        new SourceDiscoverSchemaRequestBody().sourceId(sourceCatalogId).disableCache(true);
-    sourceApi.discoverSchemaForSource(requestBody);
+        new SourceDiscoverSchemaRequestBody().sourceId(sourceCatalogId).disableCache(true).connectionId(connectionId);
+
+    try {
+      airbyteApiClient.getSourceApi().discoverSchemaForSource(requestBody);
+    } catch (final Exception e) {
+      log.info("Attempted schema refresh, but failed.");
```
jdpgrailsdev, line 54: This should probably be an `error` log with the exception itself logged as well.

alovew: oops yep, meant to update that! I also have a question about logging in general. These logs (and the ones in SyncWorkflowImpl) are not showing up in the UI logs and I'm not sure why.

alovew:
Two questions on this. I'm seeing this CI error:

```
___________ test_read_stream_returns_error_if_stream_does_not_exist ____________

    def test_read_stream_returns_error_if_stream_does_not_exist():
        expected_status_code = 400
        expected_detail = "Could not perform read with with error: \"The requested stream not_in_manifest was not found in the source. Available streams: dict_keys(['hashiras', 'breathing-techniques'])\""
    
        api = DefaultApiImpl()
        loop = asyncio.get_event_loop()
        with pytest.raises(HTTPException) as actual_exception:
            loop.run_until_complete(
                api.read_stream(StreamReadRequestBody(manifest=MANIFEST, config=***, stream="not_in_manifest"))
            )
    
        assert actual_exception.value.status_code == expected_status_code
>       assert expected_detail in actual_exception.value.detail
E       assert 'Could not perform read with with error: "The requested stream not_in_manifest was not found in the source. Available streams: dict_keys([\'hashiras\', \'breathing-techniques\'])"' in "Could not perform read with with error: The requested stream not_in_manifest was not found in the source. Available streams: dict_keys(['hashiras', 'breathing-techniques'])"
E        +  where "Could not perform read with with error: The requested stream not_in_manifest was not found in the source. Available streams: dict_keys(['hashiras', 'breathing-techniques'])" = HTTPException(status_code=400, detail="Could not perform read with with error: The requested stream not_in_manifest was not found in the source. Available streams: dict_keys(['hashiras', 'breathing-techniques'])").detail
E        +    where HTTPException(status_code=400, detail="Could not perform read with with error: The requested stream not_in_manifest was not found in the source. Available streams: dict_keys(['hashiras', 'breathing-techniques'])") = <ExceptionInfo HTTPException(status_code=400, detail="Could not perform read with with error: The requested stream not_in_manifest was not found in the source. Available streams: dict_keys(['hashiras', 'breathing-techniques'])") tblen=3>.value
```

and I'm not seeing how that's related. 

Also - the logs I added in SyncWorkflowImpl are not showing up in the UI. Does anyone have insight into which logs end up in the UI?

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
```diff
@@ -72,6 +86,23 @@ public StandardSyncOutput run(final JobRunConfig jobRunConfig,
 
     final int version = Workflow.getVersion(VERSION_LABEL, Workflow.DEFAULT_VERSION, CURRENT_VERSION);
     final String taskQueue = Workflow.getInfo().getTaskQueue();
+
+    if (version > Workflow.DEFAULT_VERSION) {
```
benmoriceau, line 90: We need a dedicated version for that. Re-using the existing version won't work.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
```diff
@@ -55,14 +64,19 @@ public class SyncWorkflowImpl implements SyncWorkflow {
...
 
   @Trace(operationName = WORKFLOW_TRACE_OPERATION_NAME)
   @Override
   public StandardSyncOutput run(final JobRunConfig jobRunConfig,
                                 final IntegrationLauncherConfig sourceLauncherConfig,
                                 final IntegrationLauncherConfig destinationLauncherConfig,
                                 final StandardSyncInput syncInput,
-                                final UUID connectionId) {
+                                final UUID connectionId)
+      throws JsonValidationException, ConfigNotFoundException, IOException, ApiException {
```
benmoriceau, line 79: I think that those exception shouldn't be in the method signature, we should catch them and set the failure reason in the return `StandardSyncOutput`

benmoriceau (dismissed):

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
alovew: oh I actually had this locally but didn't push up the change - doing that now

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
alovew: Just updated this -- I actually think we shouldn't fail the sync just because a schema refresh fails. I logged the error, but I'm not sure if that would be confusing to users. What do you think?

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
```diff
@@ -42,6 +51,8 @@ public class SyncWorkflowImpl implements SyncWorkflow {
   private static final int CURRENT_VERSION = 2;
   private static final String NORMALIZATION_SUMMARY_CHECK_TAG = "normalization_summary_check";
   private static final int NORMALIZATION_SUMMARY_CHECK_CURRENT_VERSION = 1;
+  private static final String AUTO_DETECT_SCHEMA_TAG = "auto_detect_schema";
```
benmoriceau, line 54: NIt: make sure that we have an issue to disable that after it is deployed.

airbyte-workers/src/test/java/io/airbyte/workers/temporal/sync/SyncWorkflowTest.java
```diff
@@ -423,4 +450,13 @@ private static void verifyDbtTransform(final DbtTransformationActivity dbtTransf
         operatorDbtInput);
   }
 
+  private static void verifyShouldRefreshSchema(final RefreshSchemaActivity refreshSchemaActivity) throws IOException {
```
benmoriceau, line 453: I don't think that having method to wrap one line is helping. It is only call once. We should add a level of indirection if it is not needed.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
```diff
@@ -72,6 +87,39 @@ public StandardSyncOutput run(final JobRunConfig jobRunConfig,
 
     final int version = Workflow.getVersion(VERSION_LABEL, Workflow.DEFAULT_VERSION, CURRENT_VERSION);
     final String taskQueue = Workflow.getInfo().getTaskQueue();
+
+    final int autoDetectSchemaVersion =
+        Workflow.getVersion(AUTO_DETECT_SCHEMA_TAG, Workflow.DEFAULT_VERSION, AUTO_DETECT_SCHEMA_VERSION);
+    if (autoDetectSchemaVersion >= AUTO_DETECT_SCHEMA_VERSION) {
+      UUID sourceId = null;
+      try {
```
benmoriceau, line 95: Should this be in a single try catch? I think that the stack trace will be enough to say which step failed.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
```diff
@@ -72,6 +87,29 @@ public StandardSyncOutput run(final JobRunConfig jobRunConfig,
...
+        final Status status = configFetchActivity.getStandardSync(connectionId).getStatus();
+        if (Status.INACTIVE == status) {
+          LOGGER.info("Connection is disabled. Cancelling run.");
+          final StandardSyncOutput output =
+              new StandardSyncOutput()
+                  .withStandardSyncSummary(new StandardSyncSummary().withStatus(ReplicationStatus.CANCELLED).withTotalStats(new SyncStats()));
+          return output;
+        }
+      } catch (JsonValidationException | ConfigNotFoundException | IOException | ApiException e) {
+        LOGGER.error("An error occurred during schema refresh processing. Skipping schema refresh. ", e);
```
gosusnp, line 109: Does it make sense to continue in this case? If we were unable to refresh the schema, I wonder if we're just risking getting into more errors.

airbyte-workers/src/test/java/io/airbyte/workers/temporal/sync/SyncWorkflowTest.java
gosusnp: +1

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/RefreshSchemaActivityImpl.java
```diff
@@ -7,43 +7,61 @@
...
 
   private final Optional<ConfigRepository> configRepository;
 
-  private final SourceApi sourceApi;
+  private final AirbyteApiClient airbyteApiClient;
+  private final EnvVariableFeatureFlags envVariableFeatureFlags;
 
-  public RefreshSchemaActivityImpl(Optional<ConfigRepository> configRepository, SourceApi sourceApi) {
+  public RefreshSchemaActivityImpl(Optional<ConfigRepository> configRepository,
+                                   AirbyteApiClient airbyteApiClient,
```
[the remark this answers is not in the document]
alovew: I tried this but was getting a micronaut bean error, so I followed how you did it in the PersistStateActivity

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
alovew: I think it does make the most sense to continue. Refreshing the schema every time isn't a hard requirement (we're not doing it now, for example), and I don't think we should block syncs just because we had an issue refreshing the schema.

airbyte-workers/src/test/java/io/airbyte/workers/temporal/sync/SyncWorkflowTest.java
alovew: I updated this to be called multiple times in other tests

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/RefreshSchemaActivityImpl.java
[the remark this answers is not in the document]
gosusnp: At the time `PersistStateActivity` was added, I don't think we had the independent clients yet.
The way to solve the micronaut bean error should be to add the client to the API factory: https://github.com/airbytehq/airbyte/blob/8c780bd3aaf905fa0fe2fa7cdac2299b17bac7c9/airbyte-workers/src/main/java/io/airbyte/workers/config/ApiClientBeanFactory.java#L38

@benmoriceau, any specific reasons we do not have those yet other than did not have time to get to it?

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
gosusnp: I wonder if we are taking the risk of running into not determinism errors here. Assuming it was a transient error, upon replay, we would get a different result here which I suspect may end up aborting the active sync.
If we want to be able to continue in this case, I think it would be safer to handle those errors within the activity and return a response with an error state so that temporal caches it. Unless temporal is caching errors, in which case, this not be an issue.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/RefreshSchemaActivityImpl.java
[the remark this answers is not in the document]
benmoriceau: Not that I know of. WE should add them

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
benmoriceau: To be tested but I think that temporal have the exception that is being thrown in its history and re-throw it.

airbyte-workers/src/test/java/io/airbyte/workers/temporal/sync/SyncWorkflowTest.java
benmoriceau: For a one line, I still don't think that having on indirection is needed. It doesn't really with code readability and just makes you wondering what is in the implementation.

alovew: The reason I did it was this was the pattern we seem to be using for syncWorkflowTests - all the activities are in separate methods just testing one method call

benmoriceau: Some other methods factorize the call of some static variables. Which is not the case here. Also what was done in the past might not have been the best. I would have the same comment for verifyPersistState function but it's already committed. I don't want to block the PR for that but I think that we should stick to the existing pattern if we think it is not the best. @jdpgrailsdev @colesnodgrass any though on that?

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
alovew: We have error handling within workflows already - I think we would've seen issues with this if this were a problem, or are you saying something different?

benmoriceau: Errors within the connection manaager workflow amke the workflow to continue as new so there is no risk of having an history replay issue. I think that we won't face an issue here but I think that it would worth testing it manually.

alovew: We already have one within the SyncWorkflowImpl too on line 135 - i can test manually though

alovew: How should I test this? I mean how can I 'replay' a run of a sync? Just trying syncing again?

airbyte-workers/src/main/java/io/airbyte/workers/config/ApiClientBeanFactory.java
```diff
@@ -57,6 +58,29 @@ public AirbyteApiClient airbyteApiClient(
             }));
   }
 
+  @Singleton
+  public SourceApi sourceApi(
```
gosusnp, line 62: I think you should be able to reuse the airbyteApiClient from above here:
```
@Singleton
public SourceApi sourceApi(final AirbyteApiClient airbytceApiClient) {
    return new SourceApi(airbyteApiClient);
}
```

gosusnp: Got it. In this case, we should extract the the `ApiClient` into its own singleton, and inject it in both `AirbyteApiClient` and `SourceApi`. We can pair quickly to speed up the process if you want.

`ApiClient` is the client used to make the actual calls. It should come from the API code generator we use.

`AirbyteApiClient` is a class that currently consolidate all the Apis within one. The usage of this one should decrease, moving forward, we should inject the specific Apis to highlight dependencies rather than passing a catch all with everything we have.

alovew: just updated

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
gosusnp: Looking at the temporal [retry policy doc](https://docs.temporal.io/retry-policies#default-behavior), I'd expect that temporal would catch the exception and retry the activity by default.

There is a way to set the retry policy for the activity (see [how-to-set-activity-retry-options-in-java](https://docs.temporal.io/java/how-to-set-activity-retry-options-in-java)).
```
ActivityOptions.newBuilder()
        .setRetryOptions(RetryOptions.newBuilder().setDoNotRetry().build())
```

One way to test might be to just throw the exceptions and see how the errors are currently handled.

alovew: Temporal does retry the activity by default - I moved the exception handling into the activities instead

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/RefreshSchemaActivityImpl.java
```diff
@@ -8,52 +8,72 @@
...
-    Optional<ActorCatalogFetchEvent> mostRecentFetchEvent = configRepository.get().getMostRecentActorCatalogFetchEventForSource(sourceCatalogId);
+    SourceDiscoverSchemaRequestBody requestBody =
+        new SourceDiscoverSchemaRequestBody().sourceId(sourceCatalogId).disableCache(true).connectionId(connectionId);
 
-    if (mostRecentFetchEvent.isEmpty()) {
-      return false;
+    try {
+      sourceApi.discoverSchemaForSource(requestBody);
+    } catch (final Exception e) {
+      log.error("Attempted schema refresh, but failed with error: ", e);
```
gosusnp, line 62: nit: Could be worth adding a comment explaining that we swallow the exception here to avoid blocking the replication if we fail to refresh.

gosusnp (approved):

PR #19889, opened 2022-11-29
Stop sync on a null value in a cursor column

### LINKED ISSUE #17150: Database Sources: Handle when a cursor value is NULL
[This PR](https://github.com/airbytehq/airbyte/pull/17131) adds the ability for nullable columns to be the cursor. This is to handle the case where the user doesn't really have control over the schema there (or is a view). However, we still need actual values in these columns to reliably sync.

For example, Airbyte sorts the table based on the cursor field and have a where clause `> lastCursorValue`, and different SQL databases put null values differently in sorting. Postgres puts them after non-null values, while SQL Server and MySQL puts them before non-null values.
So it means if there are new null values:
* Postgres connector can sync them, at the price of duplicating all null values each time.
* SQL Server and MySQL cannot sync them, because of the “>” where clause will filter them all out.

We should consider throwing an error to be less surprising and prevent users from getting into a bad state.

rodireich:
## What
This change will cause a sync to stop if one of the streams contains a column designated as a cursor that contains a null value.
## How
This is adding another step to the verification at the beginning of a `read` in source postgres.
We look for each stream's cursor column and if that column is NULLABLE, we look for a null value .

Because the lookup can be expensive on a large table, we are only going to do that if the column allows null values. 
Otherwise it will immediately pass this test.

## Recommended reading order
1. `PostgresSource.java` for query implementation
2. `AbstractDbSource.java` for how the new logic is added to the read process


airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -90,6 +92,14 @@ public class PostgresSource extends AbstractJdbcSource<PostgresType> implements
   public static final String SSL_KEY = "sslkey";
   public static final String SSL_PASSWORD = "sslpassword";
   public static final String MODE = "mode";
+  public static final String NULL_CURSOR_VALUE_WITH_SCHEMA = "SELECT "
```
subodh1810, line 95: Will this query also cover VIEWS with null values?

airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/PostgresSourceTest.java
```diff
@@ -544,4 +544,44 @@ private JsonNode buildConfigEscapingNeeded() {
         JdbcUtils.SSL_KEY, "false"));
   }
 
+  @Test
+  public void tableWithNullValueCursorShouldThrowException() throws SQLException {
```
subodh1810, line 548: Lets add a similar test where we have a VIEW with null value for cursor column

subodh1810 (changes requested):

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
rodireich: Yes. I verified this with views as well.
Good idea to add a test of view. working on it

airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/PostgresSourceTest.java
rodireich: Added

airbyte-integrations/connectors/source-relational-db/src/main/java/io/airbyte/integrations/source/relationaldb/InvalidCursorInfoUtil.java
```diff
@@ -15,14 +15,15 @@ public static String getInvalidCursorConfigMessage(final List<InvalidCursorInfo>
             .collect(Collectors.joining(","));
```
akashkulk, line 15: Do we want to update the messaging here? Probably want to seperate out tables that have an invalid cursor altogether vs NULL values so that the user can quickly get back into a reasonable state

rodireich: I did that - now giving a cause for failure. Either "Unsupported cursor type" or "Cursor column contains NULL value" which is formatted to the error shown in UI.
This the only place this code is used so I didn't bother with creating subclasses structure.
Added a new cause field instead

akashkulk: My suggestion would be to change 
"The following tables have invalid columns selected as cursor, please select a column with a well-defined ordering as a cursor" -> "The following tables have invalid columns selected as cursor or a NULL value in the cursor field. Please select a column with a well-defined ordering as a cursor and make sure it doesn't contain NULL values: "

WDYT? Bit verbose and the message you added to the individual `InvalidCursorInfo` would help pinpoint the exact table

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -499,4 +515,24 @@ protected static String toSslJdbcParamInternal(final SslMode sslMode) {
...
+    // Query: Only if cursor column allows null values, query whether it contains one
+    if (StringUtils.isNotBlank(schema)) {
+      query = String.format(NULL_CURSOR_VALUE_WITH_SCHEMA,
+          schema, tableName, columnName, schema, tableName, columnName, resultColName);
+    } else {
+      query = String.format(NULL_CURSOR_VALUE_NO_SCHEMA,
+          tableName, columnName, tableName, columnName, resultColName);
+    }
+    LOGGER.debug("null value query: {}", query);
+    final List<JsonNode> jsonNodes = database.bufferedResultSetQuery(conn -> conn.createStatement().executeQuery(query),
```
akashkulk, line 531: Just curious - why not a prepared statement here?

akashkulk (commented):
Just a couple small comments, looks good in general!

airbyte-integrations/connectors/source-relational-db/src/main/java/io/airbyte/integrations/source/relationaldb/InvalidCursorInfoUtil.java
rodireich: > invalid columns selected as cursor **or** a NULL value in the cursor field
I tried to specify what the exact error is so the user will be able to know the exact reason we failed.
See at the end of the [loom](https://www.loom.com/share/f410403f19324a40b0d0f51c80d2f87c) I recorded. 
Maybe I'm not getting it, but is your suggestion improve on current behavior?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
rodireich: I _think_ it is safer with a prepared statement with regards to escaping special characters?
But I may remember it incorrectly from the past.
Anything I can improve here?

airbyte-integrations/connectors/source-relational-db/src/main/java/io/airbyte/integrations/source/relationaldb/InvalidCursorInfoUtil.java
akashkulk: Maybe just add this line : "The following tables have invalid columns selected as cursor, please select a column with a well-defined ordering as a cursor and **a non-null value**"

Just a quick update to the existing string in line 13. But I'll leave it to you

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
akashkulk: They're normally preferred since:

- They are pre-compiled (so generally faster)
- Prevent SQL injection - so protect against bad table name/column name/etc in this case

airbyte-integrations/connectors/source-relational-db/src/main/java/io/airbyte/integrations/source/relationaldb/InvalidCursorInfoUtil.java
rodireich: Added

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
rodireich: Prepared statement is a pain here as the query also parameterizes schema/table/column names which are not compatible - PS adds a single quote which causes a syntax error.

rodireich:
test connector=connectors/source-postgres

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
akashkulk: Ac

akashkulk (approved):

rodireich:

PR #19909, opened 2022-11-30
Data types update: Implement protocol message downgrade path

### LINKED ISSUE #18254: data types update - implement protocol versioning (downgrades)
Implement message downgrading. See https://github.com/airbytehq/airbyte/issues/17639 for mappings between new and old types.

edgao:
closes https://github.com/airbytehq/airbyte/issues/18254

based on https://github.com/airbytehq/airbyte/pull/19240

for reviewers:
* definitely hide whitespace changes ;)
* take a quick look at: (misc changes)
    * build.gradle
    * JsonSchemaValidator.java
    * JsonSchemaReferenceTypes.java
* WellKnownTypes.json (this is just needed for the sake of testing; it's equivalent to the YAML file. I'll generate this as part of the build later, but for now just wanted to get the tests running)
* MigrationV1
    * downgrade is the top-level method, which just invokes everything else
    * downgradeSchema + isPrimitiveReferenceTypeDeclaration + getSubschemas + downgradeTypeDeclaration are for downgrading catalog messages
    * downgradeNode is for downgrading record messages; it calls into the various `downgradeXyzNode` methods
* V1MigrationTest
    * couple minor formatting/refactoring changes in the upgrade tests, but no actual behavior changes
    * scroll down to `CatalogDowngradeTest` for the actual downgrade stuff

### DIFF
--- a/airbyte-commons-protocol/build.gradle
+++ b/airbyte-commons-protocol/build.gradle
@@ -6,6 +6,7 @@ dependencies {
     testImplementation libs.bundles.micronaut.test
 
     implementation project(':airbyte-protocol:protocol-models')
+    implementation project(':airbyte-json-validation')
 }
 
 Task publishArtifactsTask = getPublishArtifactsTask("$rootProject.ext.version", project)
--- a/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
+++ b/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
@@ -34,7 +34,6 @@ public class AirbyteMessageMigrationV1 implements AirbyteMessageMigration<io.air
 
   @Override
   public io.airbyte.protocol.models.v0.AirbyteMessage downgrade(AirbyteMessage oldMessage) {
-    // TODO implement downgrade
     io.airbyte.protocol.models.v0.AirbyteMessage newMessage = Jsons.object(
         Jsons.jsonNode(oldMessage),
         io.airbyte.protocol.models.v0.AirbyteMessage.class);
@@ -239,7 +238,8 @@ public class AirbyteMessageMigrationV1 implements AirbyteMessageMigration<io.air
             replacement.setAll(subschema);
           }
         } else {
-          // TODO maybe if this oneOf has boolean entries, then it's doing something funky, and we shouldn't attempt to combine it into a single type entry
+          // TODO maybe if this oneOf has boolean entries, then it's doing something funky, and we shouldn't
+          // attempt to combine it into a single type entry
         }
       }
 
--- a/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
+++ b/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
@@ -203,7 +203,13 @@ public class AirbyteMessageMigrationV1 implements AirbyteMessageMigration<io.air
   }
 
   private void downgradeTypeDeclaration(JsonNode schema) {
-    // TODO
+    if (schema.hasNonNull("$ref")) {
+      String referenceType = schema.get("$ref").asText();
+      ((ObjectNode) schema).removeAll();
+      ((ObjectNode) schema).setAll(JsonSchemaReferenceTypes.REFERENCE_TYPE_TO_OLD_TYPE.get(referenceType));
+    } else {
+      // TODO handle oneOf
+    }
   }
 
   private static void copyKey(ObjectNode source, ObjectNode target, String key) {
--- a/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
+++ b/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
@@ -38,9 +38,9 @@ public class AirbyteMessageMigrationV1 implements AirbyteMessageMigration<io.air
   private final ConfiguredAirbyteCatalog catalog;
   private final JsonSchemaValidator validator;
 
-  public AirbyteMessageMigrationV1(ConfiguredAirbyteCatalog catalog) {
+  public AirbyteMessageMigrationV1(ConfiguredAirbyteCatalog catalog, JsonSchemaValidator validator) {
     this.catalog = catalog;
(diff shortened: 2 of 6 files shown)


airbyte-commons-protocol/build.gradle
```diff
@@ -6,6 +6,7 @@ dependencies {
     testImplementation libs.bundles.micronaut.test
 
     implementation project(':airbyte-protocol:protocol-models')
+    implementation project(':airbyte-json-validation')
```
edgao, line 9: added this dependency because the downgrade migration needs to be able to test whether a record (or part of a record) matches a schema (or subschema).

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
```diff
@@ -58,7 +58,7 @@ public JsonSchemaValidator() {
    * @param baseUri The base URI for schema resolution
    */
   @VisibleForTesting
```
suhomud, line 60: nit: `@VisibleForTesting` can be removed. Also if class has the same package structure with the unit test all protected method is accessible for unit test

edgao: unfortunately I'm using this in a separate test case from a different package (probably not a best practice, but it simplified writing the test cases) https://github.com/airbytehq/airbyte/pull/19909/files#diff-6829fbf8db73883ede3e47d869e6f5756a882c2d94d74c13e0b2128b07666216R36

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
```diff
@@ -4,30 +4,69 @@
...
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.StreamSupport;
 
 public class AirbyteMessageMigrationV1 implements AirbyteMessageMigration<io.airbyte.protocol.models.v0.AirbyteMessage, AirbyteMessage> {
 
+  private final ConfiguredAirbyteCatalog catalog;
+  private final JsonSchemaValidator validator;
+
+  public AirbyteMessageMigrationV1(ConfiguredAirbyteCatalog catalog, JsonSchemaValidator validator) {
```
gosusnp, line 39: Why do you need a catalog here? The migration is supposed to be stateless and reusable, if we need to pass a catalog here, we should probably go over this requirement.

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
gosusnp: With the constructor being public now, I don't think we need the annotation anymore.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
```diff
@@ -4,30 +4,69 @@
...
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.StreamSupport;
 
 public class AirbyteMessageMigrationV1 implements AirbyteMessageMigration<io.airbyte.protocol.models.v0.AirbyteMessage, AirbyteMessage> {
```
gosusnp, line 34: Just seeing now that the previous PR should have added an `@Singleton` annotation here.
I'll add it to the base PR which may change some behaviors, mostly, `JsonSchemaValidator` will need a factory method.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
```diff
@@ -4,30 +4,69 @@
...
 import com.fasterxml.jackson.databind.node.TextNode;
 import io.airbyte.commons.json.Jsons;
 import io.airbyte.commons.version.AirbyteProtocolVersion;
 import io.airbyte.commons.version.Version;
+import io.airbyte.protocol.models.ConfiguredAirbyteCatalog;
+import io.airbyte.protocol.models.ConfiguredAirbyteStream;
 import io.airbyte.protocol.models.JsonSchemaReferenceTypes;
+import io.airbyte.protocol.models.v0.AirbyteRecordMessage;
 import io.airbyte.protocol.models.v1.AirbyteMessage;
+import io.airbyte.protocol.models.v1.AirbyteMessage.Type;
```
gosusnp, line 21: Is there some unexpected changes from the branch rebase? I think this should still be using v-null as v1 instead of the actual v1 namespace.

gosusnp (changes requested):

airbyte-json-validation/src/main/java/io/airbyte/validation/json/JsonSchemaValidator.java
edgao: I'd mildly prefer to keep it since it's still only being used by tests? I.e. production code paths should just use the no-args constructor

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
edgao: I guess I botched a rebase conflict. looks like this also happened on the upgrades pr (https://github.com/airbytehq/airbyte/pull/19240/files#diff-bca6d63bb992a16edcbc829c7faf7e4321adf0e8b45704ce464453918ea1c961R15); will fix there also

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
edgao: by factory method - just something like `public static JsonSchemaValidator build() {...}` ? or does it need to have a specific signature? and/or literally have some `class JsonSchemaValidatorFactory {...}`

edgao:
merging this into https://github.com/airbytehq/airbyte/pull/19240 for my convenience. code should be reasonably easy to differentiate between up vs downgrades, for anyone who only wants to review the downgrade half.

(I'm about to do some git nonsense, so only having one PR will simplify that process)

PR #19910, opened 2022-11-30
Fix the underlying component schema to use check instead of checker

brianjlai:
## What
The manifest component schema that is generated dynamically based on the low-code data model and used to validate manifests expects `checker` but it should really be validating that `check` is present.

## How
In the class ConcreteDeclarativeSource, the dataclass field is called `checker` which then formats the manifest by that name. This is causing an issue for the form based connector builder UI. For some reason in a manifest we take the contents of `check` and place them under `checker`. We should just do away with this entirely and format everything under checker. This PR does that and also regenerates the static schema  as well.

Given that all manifests were already defining the `check`, this isn't a breaking change, we're just doing away with remapping it to checker which seems unnecessary.

## Recommended reading order


airbyte-cdk/python/airbyte_cdk/sources/declarative/config_component_schema.json
```diff
@@ -111,72 +128,37 @@
...
-    },
     "SimpleRetriever": {
       "allOf": [
         {
           "$ref": "#/definitions/Retriever"
         },
         {
           "type": "object",
-          "required": ["requester", "record_selector"],
+          "required": ["requester", "record_selector", "config"],
```
lmossman, line 138: Do you know why this PR is adding `config` as a required field to several of these objects? I think ideally this would not be required, since we are not planning to set this in the connector builder

brianjlai: This is because we haven't regenerated the schema in at least a few weeks and there were some component classes that had the config object added as a field. So that gets reflected in the auto-generated schema. I did some spot checks and this is accurate to the current fields of the SimpleRetriever class.

Sadly this is just another case of the downsides of the autogenerated schema since we're coupled to the underlying implementation which passes things like config under the hood

sherifnada (approved):

lmossman (approved):

PR #19914, opened 2022-11-30
🪟 🔧 Optimize lodash imports

timroes:
## What

This changes all named imports from lodash e.g. `import { isEqual } from 'lodash';` to imports from the submodule directly i.e. `import isEqual from 'lodash/isEqual';`.

This allows our build system to only need to include the corresponding lodash submodules we really use. This stripped off ~500kB (21kB gziped) of our bundle size.

Also this adds a linting rule that will prevent those types of imports in the future.


joeykmh (approved):
Wow, I was not aware that named imports were importing the whole library. Good fix! 👍

PR #19720, opened 2022-11-22
Enable full SAT for the BigQuery source

DoNotPanicUA:
## What
Enable SAT for the BigQuery source

## Added secrets:
- SECRET_SOURCE-BIGQUERY_SAT__CREDS

## Configured test instances
- [BigQuery] sat_test_dataset

### SAT related issues:
- SAT can't check null values due to column type validation fail
### Source related issues:
- [BigQuery] Complex type like REPEATED or OBJECT fail validation. The root cause of the issue requires investigation. Is it SAT issue, or is the source work wrong?
- [BigQuery] Fails the future incremental check. An investigation is required. Or the abnormal cursor value format is wrong, or the source doesn't accept the initial state message.

PR #19915, opened 2022-11-30
 Enable full SAT for the Redshift source

### LINKED ISSUE #19895: Enable SAT for hosted sources - Redshift
Enable the SAT for the Redshift source. See epic description for more context airbytehq/airbyte#19708.

DoNotPanicUA:
## What
Enable SAT for the Redshift source

## Configured test instances
- sat_test_dataset

## Found issues
SAT related issues:
- in case of default value for the `configured_catalog_path`, the SAT doesn't take the file by default, but sync all available streams from the source.


airbyte-integrations/connectors/source-redshift/acceptance-test-config.yml
```diff
@@ -4,4 +4,26 @@ connector_image: airbyte/source-redshift:dev
 tests:
   spec:
     - spec_path: "src/test-integration/resources/expected_spec.json"
+      timeout_seconds: "1200"
```
evantahler, line 7: lol, redshift.

evantahler (commented):
👍 

Similar question to [BQ](https://github.com/airbytehq/airbyte/pull/19720#pullrequestreview-1200009858) - should we document the test instance's details in some way

DoNotPanicUA:
> +1
> 
> Similar question to [BQ](https://github.com/airbytehq/airbyte/pull/19720#pullrequestreview-1200009858) - should we document the test instance's details in some way

[the doc](https://docs.google.com/document/d/1k5TvxaNhKdr44aJIHWWtLk14Tzd2gbNX-J8YNoTj8u0/edit#).

alafanechere:
> > +1
> > Similar question to [BQ](https://github.com/airbytehq/airbyte/pull/19720#pullrequestreview-1200009858) - should we document the test instance's details in some way
> 
> [the doc](https://docs.google.com/document/d/1k5TvxaNhKdr44aJIHWWtLk14Tzd2gbNX-J8YNoTj8u0/edit#).

I think we shall create connector-specific docs for better visibility for our community. I would suggest adding a `README` under `integration-tests` + a source specific script

airbyte-integrations/connectors/source-redshift/acceptance-test-config.yml
```diff
@@ -4,4 +4,26 @@ connector_image: airbyte/source-redshift:dev
 tests:
```
alafanechere, line 4: Please use the latest acceptance-test-config format. [Example here.](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-zendesk-support/acceptance-test-config.yml#L1). I'd be curious to check the test output under `high` test strictness level, we might get this for free 😄

alafanechere (changes requested):

airbyte-integrations/connectors/source-redshift/integration_tests/README.md
```diff
@@ -0,0 +1,3 @@
+# Seeding the dataset
+You can find the SQL scripts in this folder if you need to create or fix the SAT dataset.
+For more instructions and information about valid scripts, please check this [doc](https://docs.google.com/document/d/1k5TvxaNhKdr44aJIHWWtLk14Tzd2gbNX-J8YNoTj8u0/edit#heading=h.ls9oiedt9wyy).
```
alafanechere, line 3: I don't think we should link a Google Doc here, it's not or usual approach to documentation and can break accessibility. 

Moreover I would suggest you add the following to the README:
- Share the requirements in terms of Redshift setup to run the tests on it.
- Explain the role of each SQL script.
- Explain technically how to seed the DB.

I think these instructions will differ for each source, so I'm not in favor of a central document. If there are common instructions that can be useful for all database sources, feel free to edit https://docs.google.com/document/d/1k5TvxaNhKdr44aJIHWWtLk14Tzd2gbNX-J8YNoTj8u0/edit#heading=h.ls9oiedt9wyy

alafanechere (changes requested):

airbyte-integrations/connectors/source-redshift/integration_tests/README.md
DoNotPanicUA: That is the point, all instructions are the same. The difference is only in the credentials required for the connection and the SQL syntax. It's common for any SQL source.
@evantahler created the doc and said to put all related instructions there.

P.S. If we have a new demand to improve or rework something, it's better to create a new issue and discuss it at the planning meeting with the whole team. No need to hold the integration tests for the source as a hostage of this discussion.

evantahler: If I'm being asked to be a tie-breaker, in the spirt of "done is better than perfect", I think the google doc is OK for now.  We can move the contents into the public repo once we are done with all the Java sources, and we can see how similar/different the instructions will be.  Also, the doc currently links to secrets, which is helpful, and shouldn't be in this repo.

alafanechere (approved):
I updated your branch to get the test working again. Feel free to merge once they pass.

Issue #19733, opened 2022-11-22
[CDK] Refactor reading JSON inputs for syncs under one shared method


brianjlai:
## Tell us about the problem you're trying to solve
This ticket is a precursor to https://github.com/airbytehq/airbyte/issues/18946 . We currently attempt to read in JSON file inputs during connector syncs like configs, configured catalogs, and state. Before tackling that issue, we should clean up the code related to file reads and improve code reuse. The naming of the `BaseConnector.read_config()` is also not accurate since it is currently being used to read other types of inputs too since its a generic json file reader.

## Describe the solution you’d like
We should rename `BaseConnector.read_config()` to `read_json_file()` to accurately reflect what it is doing. The existing usages of `read_config()` should be updated.

The `Source.read_state()` method should start using this newly renamed `read_json_file()` to read the state JSON file instead of duplicating all the same file reading logic.

## Acceptance Criteria

- [ ] When running a `check` operation on a connector, the operation succeeds with valid json
- [ ] When running a `read` operation on a connector, the operation succeeds with valid json
- [ ] When running a `read` where the configured catalog doesn't exist, an error should be thrown

maxi297:
This has been solved as part of https://github.com/airbytehq/airbyte/pull/20019

PR #19931, opened 2022-11-30
🚨🚨 [ISSUE #19733] Executing a source with an invalid JSON file should display a clear error

maxi297:
## What
This change address https://github.com/airbytehq/airbyte/issues/19733 which would cause the error message from reading config files to be unclear.

## How
As https://github.com/airbytehq/airbyte/issues/19733 states, `We should rename BaseConnector.read_config() to read_json_file() to accurately reflect what it is doing. The existing usages of read_config() should be updated.` and `The Source.read_state() method should start using this newly renamed read_json_file() to read the state JSON file instead of duplicating all the same file reading logic.`

## 🚨 User Impact 🚨
Previous error message: _Expecting value: line 1 column 1 (char 0)_
New error message: _Could not read json file secrets/config.json: `Expecting value: line 1 column 1 (char 0)`_

The solution proposed in https://github.com/airbytehq/airbyte/issues/19733 is a breaking change since the public method `read_config` is changed to `read_json_file`.

- [X] Unit & integration tests added and passing
- [ ] Code reviews completed

## Tests

</details>

</details>

</details>

## Other Notes
Running `mypy airbyte_cdk` (as mentioned in the README.md) before creating changes to the repository gave me the following output `Found 134 errors in 60 files (checked 140 source files)`. Is this expected?

### DIFF
--- a/airbyte-cdk/python/CHANGELOG.md
+++ b/airbyte-cdk/python/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.12.0
+Improve error readability when error while reading JSON config files
+
 ## 0.11.1
 Low-code: Fix the component manifest schema to and validate check instead of checker
 
--- a/airbyte-cdk/python/airbyte_cdk/connector.py
+++ b/airbyte-cdk/python/airbyte_cdk/connector.py
@@ -47,10 +47,14 @@ class BaseConnector(ABC, Generic[TConfig]):
         """
 
     @staticmethod
-    def read_config(config_path: str) -> TConfig:
+    def read_json_file(config_path: str) -> TConfig:
         with open(config_path, "r") as file:
             contents = file.read()
-        return json.loads(contents)
+
+        try:
+            return json.loads(contents)
+        except json.JSONDecodeError as error:
+            raise ValueError(f"Could not read json file {config_path}: `{error}`")
 
     @staticmethod
     def write_config(config: TConfig, config_path: str):
--- a/airbyte-cdk/python/airbyte_cdk/connector.py
+++ b/airbyte-cdk/python/airbyte_cdk/connector.py
@@ -54,7 +54,7 @@ class BaseConnector(ABC, Generic[TConfig]):
         try:
             return json.loads(contents)
         except json.JSONDecodeError as error:
-            raise ValueError(f"Could not read json file {config_path}: `{error}`")
+            raise ValueError(f"Could not read json file {config_path}: {error}. Please ensure that it is a valid JSON.")
 
     @staticmethod
     def write_config(config: TConfig, config_path: str):
@@ -78,7 +78,10 @@ class BaseConnector(ABC, Generic[TConfig]):
         if yaml_spec:
             spec_obj = yaml.load(yaml_spec, Loader=yaml.SafeLoader)
         elif json_spec:
-            spec_obj = json.loads(json_spec)
+            try:
+                spec_obj = json.loads(json_spec)
+            except json.JSONDecodeError as error:
+                raise ValueError(f"Could not read json spec file: {error}. Please ensure that it is a valid JSON.")
         else:
             raise FileNotFoundError("Unable to find spec.yaml or spec.json in the package.")
 
--- a/airbyte-cdk/python/airbyte_cdk/destinations/destination.py
+++ b/airbyte-cdk/python/airbyte_cdk/destinations/destination.py
@@ -94,7 +94,7 @@ class Destination(Connector, ABC):
         if cmd == "spec":
             yield AirbyteMessage(type=Type.SPEC, spec=spec)
             return
-        config = self.read_config(config_path=parsed_args.config)
+        config = self.read_json_file(config_path=parsed_args.config)
         if self.check_config_against_spec or cmd == "check":
             try:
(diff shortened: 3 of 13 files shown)


airbyte-cdk/python/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.12.0
```
maxi297, line 3: Having to change the public method `read_config` is a breaking change in the API so I bumped the version. However, I feel like the change I made shouldn't affect the version number and maybe this means that we would benefit for better separation of concerns. For example, if the validation of input (JSON file in this case) would not be part of the domain, we would probably be shielded from breaking change in that case.

airbyte-cdk/python/airbyte_cdk/connector.py
```diff
@@ -47,10 +47,14 @@ def configure(self, config: Mapping[str, Any], temp_dir: str) -> TConfig:
...
-    def read_config(config_path: str) -> TConfig:
+    def read_json_file(config_path: str) -> TConfig:
         with open(config_path, "r") as file:
             contents = file.read()
-        return json.loads(contents)
+
+        try:
+            return json.loads(contents)
+        except json.JSONDecodeError as error:
+            raise ValueError(f"Could not read json file {config_path}: `{error}`")
```
sherifnada, line 57:
```suggestion
            raise ValueError(f"Could not read json file {config_path}: {error}. Please ensure that it is a valid JSON.")
```

sherifnada (changes requested):
Nice work! The main change looks good to me. Request for change: more connectors use this method, so we should make sure to update them as well: 

```

airbyte-integrations/connectors/source-s3/source_s3/source.py:
  60      def read_config(self, config_path: str) -> Mapping[str, Any]:
  61:         config: Mapping[str, Any] = super().read_config(config_path)
  62          if config.get("format", {}).get("delimiter") == r"\t":

airbyte-integrations/connectors/source-s3/unit_tests/conftest.py:
  51      source = SourceS3()
  52:     config = source.read_config(config_file)
  53      return config

airbyte-integrations/connectors/source-s3/unit_tests/test_source.py:
  21      source = SourceS3()
  22:     config = source.read_config(config_file)
  23      assert config["format"]["delimiter"] == "\t"

```

airbyte-cdk/python/airbyte_cdk/connector.py
maxi297: I'll take the action item to see in which case JSONDecodeError might get raised. As of now, we have identified:
* File with invalid JSON
* File does not exist
Based on that (and a research to see if there are more cases), I'll update the error message to be more precise. Thanks Sherif

airbyte-cdk/python/CHANGELOG.md
brianjlai: just a heads up that I merged a patch version for a fix as `0.11.1`. Your version should be fine, but just a reminder to get the latest master to make sure our changelog is up to date

brianjlai (commented):
changes and additional tests look good. great work! 

Will approve pending the update to the s3 source which appears to be the only source that was using `read_config()`. I think we also discussed async, but the json file reading in `BaseConnector.spec()` to reuse what you built here will be in a follow up PR?

airbyte-cdk/python/airbyte_cdk/connector.py
```diff
@@ -74,7 +78,10 @@ def spec(self, logger: logging.Logger) -> ConnectorSpecification:
         if yaml_spec:
             spec_obj = yaml.load(yaml_spec, Loader=yaml.SafeLoader)
         elif json_spec:
-            spec_obj = json.loads(json_spec)
+            try:
+                spec_obj = json.loads(json_spec)
+            except json.JSONDecodeError as error:
+                raise ValueError(f"Could not read json spec file: {error}. Please ensure that it is a valid JSON.")
```
maxi297, line 84: I would have like to reuse `read_json_file` but if we want to keep which file we were not able to parse, this would have become tricky since we don't read the file the same way for the spec. I'm open to suggestion as how we could reuse in a clean way

airbyte-cdk/python/airbyte_cdk/connector.py
```diff
@@ -47,10 +47,14 @@ def configure(self, config: Mapping[str, Any], temp_dir: str) -> TConfig:
         """
 
     @staticmethod
-    def read_config(config_path: str) -> TConfig:
+    def read_json_file(config_path: str) -> TConfig:
```
maxi297, line 50: I was wondering here: should we change the return type to `Mapping[str, Any]` since this method has become more generic than just reading configs? 

Also, the override in `SourceS3. read_json_file` still use `Mapping[str, Any]` so base on your answer, I'll re-align the typing for the return types

brianjlai: Mapping is a more accurate for this method so it's probably okay to change this signature

brianjlai (approved):
looks good! just responses to the comments you left

maxi297:
Since this merge request was from a forked repo, it had to be reopened in the current airbyte repo. Here is the new PR for this change: https://github.com/airbytehq/airbyte/pull/20019

PR #19950, opened 2022-11-30
Fix margin and padding issues with CatalogTree table

### LINKED ISSUE #19874: Connection streams vertical and right margin do not align with other cards
Seen in master, the top, bottom, and right margin of the streams card is too large and does not align with the rest of the

Image

This is related to the changes in #18187

edmundito:
## What
Resolves #19874
Closes #19924

Fixes many issues with the margin and padding in the CatalogTree table:

* Fixes the field table left margin not quite aligning with the row
* Fixes the bottom margin of the field table
* Fixes padding issues around the table and simplifies check between old table and new table

Before:
image

Problem areas:
fixes

After:

image

## How
This approach favors the parent pushing the content inward with padding over the content making margin for themselves.
There is a hotfix for specific situation that was too convoluted to fix. Noted in comments.

### DIFF
--- a/airbyte-webapp/src/components/connection/CatalogTree/StreamFieldTable.module.scss
+++ b/airbyte-webapp/src/components/connection/CatalogTree/StreamFieldTable.module.scss
@@ -1,8 +1,15 @@
 @use "scss/colors";
 @use "scss/variables";
 
+.container {
+  padding-bottom: variables.$spacing-md;
+
+  // HACK: Hot-fixes an issue in StreamHeader SCSS where the .catalogSectionRow margin-top is creating a spacing between rows
+  margin-bottom: -9px;
+}
+
 .rowsContainer {
   background: colors.$white;
   border-radius: variables.$border-radius-md;
-  margin: 0 10px 5px;
+  margin: 0 variables.$spacing-md;
 }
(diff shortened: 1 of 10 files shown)


krishnaglick (commented):
Code looks good. Would like to test locally before approval.

airbyte-webapp/src/components/connection/CatalogTree/StreamFieldTable.module.scss
```diff
@@ -1,8 +1,15 @@
 @use "scss/colors";
 @use "scss/variables";
 
+.container {
+  padding-bottom: variables.$spacing-md;
+
+  // HACK: Hot-fixes an issue in StreamHeader SCSS where the .catalogSectionRow margin-top is creating a spacing between rows
+  margin-bottom: -9px;
```
josephkmh, line 8: Hot-fix in the sense that this whole table will be deprecated soon, right?

joeykmh (approved):
Looks good to me - tested locally and did not see any issues ✨

airbyte-webapp/src/components/connection/CatalogTree/StreamFieldTable.module.scss
edmundito: @josephkmh Yes, in the sense that fixing the root issue is not as simple as adjusting the margins in the StreamHeader and that this table is on its way out in the next few months.

krishnaglick:
Build failing on snapshot changes. Maybe test should be its own runner?

PR #19953, opened 2022-11-30
`socat` gets more resources

evantahler:
We currently only allocate 0.2 CPU shares to the socat processes, but we've observed that we are hitting that limit:

unnamed

Rather than remove socat and enforce the same pod affinity for all the pods in the sync, lets start by simply throwing money at the problem!

Future work would be to adjust this dynamically as needed at runtime, or for only those syncs with "fast enough" sources


colesnodgrass (approved):

evantahler:
@davinchia can you confirm that this file is what is used by Cloud, and we don't overwrite these values?

evantahler:
@supertopher are there any worries about exhausting our cluster resources?

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubePodProcess.java
```diff
@@ -109,9 +109,14 @@ public class KubePodProcess extends Process implements KubePod {
   private static final String DEFAULT_MEMORY_LIMIT = "50Mi";
   private static final String DEFAULT_CPU_REQUEST = "0.1";
   private static final String DEFAULT_CPU_LIMIT = "0.2";
+  private static final String DEFAULT_SOCAT_CPU_REQUEST = "1.0";
+  private static final String DEFAULT_SOCAT_CPU_LIMIT = "1.5";
```
davinchia, line 113: we probably need to provide a way to override these via env vars since these are high for our dev envs and also high for a portion of our OSS users - it would also make it easier for us to test different configurations and the performance effect.

evantahler:
I *think* this is the pattern to add ENV configs?  I also think we shouldn't add this to the OSS helm charts for now, sort of like how `CONTROL_PLANE_AUTH_ENDPOINT` isn't there either. 

@davinchia once this is merged, I have no idea how to try our a change in the cloud repo...

colesnodgrass (approved):

airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
```diff
@@ -73,7 +73,17 @@ public class EnvConfigs implements Configs {
   public static final String JOB_KUBE_TOLERATIONS = "JOB_KUBE_TOLERATIONS";
   public static final String JOB_KUBE_NODE_SELECTORS = "JOB_KUBE_NODE_SELECTORS";
   public static final String JOB_KUBE_ANNOTATIONS = "JOB_KUBE_ANNOTATIONS";
+  private static final String DEFAULT_SIDECAR_MEMORY_REQUEST = "25Mi";
```
davinchia, line 76: nit: if you scroll down to line 200, you'll see a bunch of defaults there. should move all the defaults there when we merge. ok to leave here if this is for testing.

airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
```diff
@@ -725,11 +735,41 @@ public String getJobKubeMainContainerImagePullSecret() {
...
+    return getEnvOrDefault(SIDECAR_KUBE_CPU_LIMIT, DEFAULT_SIDECAR_KUBE_CPU_LIMIT);
+  }
+
+  @Override
+  public String getSidecarKubeMemoryLimit() {
+    return getEnvOrDefault(SIDECAR_KUBE_MEMORY_LIMIT, DEFAULT_SIDECAR_KUBE_MEMORY_LIMIT);
+  }
+
+  @Override
+  public String getSidecarMemoryRequest() {
```
davinchia, line 754:
```suggestion
  public String getSidecarKubeMemoryRequest() {
```

davinchia: To follow existing convention

davinchia (approved):
Evan, I'm going to approve. Feel free to merge after addressing the comments.

evantahler:
Merged with a red test - that's a connector base format error, will be solved by https://github.com/airbytehq/airbyte/pull/20163

PR #19976, opened 2022-12-01
make flaky test less sensitive

mfsiega-airbyte:
## What
Make flaky test less sensitive.

## How
Don't insist that only a single job be run. If the sync needed a retry for any other reason then the test will fail. This is too sensitive, and the flakiness is interfering with normal dev process.

### DIFF
--- a/airbyte-tests/src/acceptanceTests/java/io/airbyte/test/acceptance/ContainerOrchestratorAcceptanceTests.java
+++ b/airbyte-tests/src/acceptanceTests/java/io/airbyte/test/acceptance/ContainerOrchestratorAcceptanceTests.java
@@ -104,7 +104,8 @@ class ContainerOrchestratorAcceptanceTests {
   @Test
   @SuppressWarnings("PMD.JUnitTestsShouldIncludeAssert")
   void testDowntimeDuringSync() throws Exception {
-    // NOTE: PMD assert warning suppressed because the assertion was flaky. The test will throw if the sync does not succeed.
+    // NOTE: PMD assert warning suppressed because the assertion was flaky. The test will throw if the
+    // sync does not succeed.
     final String connectionName = "test-connection";
     final UUID sourceId = testHarness.createPostgresSource().getSourceId();
     final UUID destinationId = testHarness.createPostgresDestination().getDestinationId();
--- a/airbyte-tests/src/acceptanceTests/java/io/airbyte/test/acceptance/ContainerOrchestratorAcceptanceTests.java
+++ b/airbyte-tests/src/acceptanceTests/java/io/airbyte/test/acceptance/ContainerOrchestratorAcceptanceTests.java
@@ -102,7 +102,9 @@ class ContainerOrchestratorAcceptanceTests {
   }
 
   @Test
+  @SuppressWarnings("PMD.JUnitTestsShouldIncludeAssert")
   void testDowntimeDuringSync() throws Exception {
+    // NOTE: PMD assert warning suppressed because the assertion was flaky. The test will throw if the sync does not succeed.
     final String connectionName = "test-connection";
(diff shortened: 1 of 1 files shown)


supertopher:
this looks great to me, but I'm pretty short on context for this part of the codebase so I'm shying away from approval.  
thanks for working on this so quickly

me and Michael had a DM conversation about this.  I think my team can do more to set expectations around responsibility.  I'll have a more clear DRI and better process going forward.  Thanks for helping us with the first run

airbyte-tests/src/acceptanceTests/java/io/airbyte/test/acceptance/ContainerOrchestratorAcceptanceTests.java
```diff
@@ -129,14 +132,6 @@ void testDowntimeDuringSync() throws Exception {
...
 
     waitForSuccessfulJob(apiClient.getJobsApi(), connectionSyncRead.getJob());
-
-    final long numAttempts = apiClient.getJobsApi()
-        .getJobInfo(new JobIdRequestBody().id(connectionSyncRead.getJob().getId()))
-        .getAttempts()
-        .size();
-
-    // it should be able to accomplish the resume without an additional attempt!
-    assertEquals(1, numAttempts);
```
lmossman, line 139: I think the reason this assertion was being made was that we explicitly wanted to test that if the worker is scaled down and back up in the middle of a sync (e.g. deploying a new change to cloud prod), then those syncs will NOT be failed but will instead continue on as normal and succeed.

If we get rid of this assertion, it could mean that a regression where scaling down a working in the middle of a sync actually causes that sync to fail will go unnoticed.

@mfsiega-airbyte Do we have a sense of why these syncs are failing in CI? I feel like we should be able to rely on the fact that our acceptance tests syncs are succeeding, otherwise other tests will likely be breaking as well.

mfsiega-airbyte: Here's my understanding - bear in mind that I don't actually have any context on this test, beyond looking at it today.

This test (even after this PR) *does* validate that the sync succeeds even if the worked is scaled down and back up. Note that in the preceding line we `waitForSuccessfulJob` - if the sync fails, `waitForSuccessfulJob` will eventually throw and the test will fail.

What the assertion -- deleted by this PR -- is verifying is that it succeeds within a single attempt. Is there a reason that it's important that it succeeds within a single attempt? And in other places where our acceptance tests look for successful syncs, do they also verify that it succeeds within a single attempt?

If not, it's hard to see why we're verifying that here (except I guess that it's a cool implementation detail).

lmossman: That's a good point about our other acceptance tests.

I think this was trying to ensure that scaling down and up doesn't actually result in any "failures" at all, whether at the attempt or the overall sync level (i.e. it should be invisible to the user that the scaling down/up happened at all). But since our other tests are not ensuring this, it probably doesn't make sense to do that here. And as long as the overall sync succeeds, that's the more important part.

Given that, I think I'm fine with this change. But it might be worthwhile to separately look into why these attempts are failing, since I think that is still unexpected.

lmossman (approved):

airbyte-tests/src/acceptanceTests/java/io/airbyte/test/acceptance/ContainerOrchestratorAcceptanceTests.java
mfsiega-airbyte: Agree on all of it. I see what you mean about scaling being invisible.

I'll go ahead and merge this but agree we should understand why some attempts are failing, and ideally we'd be able to re-enable these.

Issue #17853, opened 2022-10-11
Remove custom code for checking 403s* to skip unavailable streams for a connector, identify other affected connectors


evantahler:
* What are these connectors?  Make a list.
  * `source-recharge`
  * .... more! (at least 2)
* Migrate custom code to new CDK methods

erohmensing:
[WIP]

Connectors handling 403s already:
* [Stripe](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-stripe/source_stripe/streams.py#L101-L104) - skip stream
* [Zendesk-sunshine](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-zendesk-sunshine/source_zendesk_sunshine/source.py#L70) - removed unavailable stream by commenting it out. Could be being skipped unnecessarily if this stream is now readily available


Connectors potentially expecting 403s for other reasons: 
* Amazon ads - Unclear because of [these mocked tests](https://github.com/airbytehq/airbyte/pull/5023/files?diff=unified&w=0#diff-422d940bee3a02c052097d4bdd28d9a7cea06596c8d6048e5e3ba10b2563714bR193-R206) that still exist, but look to be made redundant by changes more effectively handling errors [here](1406e0a66972002f1b440afd7d6f3cddde884507c0da056d48ce8c9422e7569aR128)?

pedroslopez:
grooming:
- do one connector to make sure it works
- if it's many connectors, rescope this and make a list on connectors that need to be updated, potentially have GL take a look

evantahler:
Check out https://github.com/airbytehq/oncall/issues/973 - The Facebook Marketing source is particularly bad right now with access permission failures.  Perhaps this is one of the first connectors we upgrade

cc @YowanR

erohmensing:
I created a Google Doc with an audit of custom error handling for 403s [here](https://docs.google.com/document/d/1ghkeIhlfqx7yf6fi3sv3nbpEk7BUpIrVlPqkBmrXR0U/edit). 

TL;DR:
* I found 14 sources using HttpStreams that do custom error handling for 403s 
* Most of these were on the record-reading level, some had special cases for connection checking 
* A common spot for handling was adjusting `should_retry` to `False`. As far as I know that would be per-record, so wouldn't stop the source from trying to read more records from the stream. 
[...]

PR #19977, opened 2022-12-01
CDK: `AbstractSource.read()` skips syncing stream if its unavailable (add `AvailabilityStrategy` concept)

### LINKED ISSUE #17617: Skip stream syncing when stream is unavailable
Moving towards https://github.com/airbytehq/airbyte/issues/14906 and the stream availability proposal outlined in [this doc](https://docs.google.com/document/d/1PA170k3YkLSe2aqvldPZwAxQJ8troulj_7wmNtdArdM/edit#), we need a way for connectors to declare whether a stream is available or not. Currently, connectors are doing this on their own and the behavior is not standardized, so this is about adding functionality in the CDK to make it easy for connector developers to do the right thing.  This ticket will use this feature skip the stream when syncing, but can later be used to set unavailability info on the stream catalog once the protocol has been updated to support this.

## Implementation proposal

I would propose adding two new methods in the cdk, one at the source level and one at the stream level:
- **AbstractSource.get_stream_availability_data()**: returns a python dictionary with any data the connector may need to determine stream availability. For example, get the authorized scopes for the currently used api key or issue a request to determine the account level (e.g. manager vs regular account type).
- **Stream.is_unavailable(availability_data)**: is passed in the source-level availability data returned by the previous method, and returns a tuple `is_unavailable, unavailability_reason` (e.g. `False, "This stream is only available for manager accounts`)

The CDK can use these methods to automatically skip these streams and log a message when syncing. Some pseudo-code for `AbstractSource.read()`:

```
AbstractSource.read:
        availability_data = source.get_stream_availability_data()
	before calling the stream's read:
		is_unavailable, unavailable_reason = stream_instance.is_unavailable(asdfasdf)
                if is_unavailable:
                       log(Skipped syncing stream XYZ: {unavailable_reason})
                 else:
                       do the sync
```

**Implementation example (hubspot)**:

```python
class Source:
	def get_stream_avaiability_data:
		granted_scopes = fetch()
		return {'granted_scopes': granted_scopes} 

class Stream:
	def is_unavailable(availability_data):
		if "my_scope" in availability_data.get("granted_scopes", []):
			return False, None
		return True, "my_scope not granted"
```

## Acceptance criteria
- CDK exposes new methods to determine whether a stream is available
- CDK auto skips streams if they are not available

erohmensing:
## What
* Closes airbytehq/airbyte#17617 
* Fixes issue where an entire sync would fail if one or more streams were unavailable due to 403: Forbidden permission errors (usually due to missing scopes in an OAuth connection)

## How
* Adds a `check_availability` method to the `Stream` class to check availability
* Adds an `AvailabilityStrategy` class to the `Stream` class to perform this check
* Adds an `HttpAvailabilityStrategy` class to the `HTTPStream` class, and sets it as the default `AvailabilityStrategy` for this class
  * This strategy tries to read the first record, excepts a 403:Forbidden HTTPError, and returns that the stream is unavailable, why, and where to learn more to fix this issue
* Adds logic to `AbstractSource.read()` to skip syncing the stream if the stream is unavailable
* Updates `CheckStream` `ConnectionChecker` class to first try using the availability strategy, if there is one

## Recommended reading order
Implementation:
1. `airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py`
2. `airbyte-cdk/python/airbyte_cdk/sources/streams/core.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/availability_strategy.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/streams/http/http.py`
4. `airbyte-cdk/python/airbyte_cdk/sources/streams/http/availability_strategy.py`

Testing:
1. `airbyte-cdk/python/unit_tests/sources/streams/test_availability_strategy.py`
2. `airbyte-cdk/python/unit_tests/sources/streams/http/test_availability_strategy.py`
3. `airbyte-cdk/python/unit_tests/sources/test_source.py`

Refactoring/small improvements:
1. `airbyte-cdk/python/airbyte_cdk/sources/utils/stream_helpers.py`
2. Remaining files (typos, deprecation warnings etc)

## 🚨 User Impact 🚨
End users:
* For a given source, if CDK version is updated and source streams subclasses `HttpStream`, the `HTTPAvailabilityStrategy` will now be in effect for that stream, which means:
  * Syncs that fail due to an expected HTTP error (403) while reading a stream will no longer fail. Information about the stream that is skipped is currently only listed as a warning in the logs. The skipped stream will not sync, the other streams will.

Source developers:
* Source developers will now be able to write AvailabilityStrategies if they update their CDK version. 
* If CDK version is updated and source streams only subclasses `Stream`, no change.
* If CDK version is updated and source streams subclasses `HttpStream`, the `HTTPAvailabilityStrategy` will now be in effect for that stream. 

**Note**: Sources that already handle 403 errors in their `read()` methods should move that handling to their own `AvailabilityStrategy` (if handling is more specific than the default implementation). That will be addressed in airbytehq/airbyte#17853


airbyte-cdk/python/airbyte_cdk/sources/availability_strategy.py
```diff
@@ -0,0 +1,97 @@
...
+          is available. Otherwise, the stream is unavailable for some reason and
+          the str should describe what went wrong.
+        """
+        try:
+            # Some streams need a stream slice to read records (e.g. if they have a SubstreamSlicer)
+            stream_slice = self._get_stream_slice(stream)
+            records = stream.read_records(sync_mode=SyncMode.full_refresh, stream_slice=stream_slice)
+            next(records)
+        except HTTPError as error:
+            return False, repr(error)
```
erohmensing, line 47: Move exception handling to separate method so that instances of this class don't need to worry about the try: part
see: [github](https://github.com/airbytehq/airbyte/pull/19978/files#diff-895c7e2fc3c7a29570263791b1d031ceb5309127e9e42609ddd9ed425f548e47R241)

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -130,6 +136,10 @@ def read(
 
         logger.info(f"Finished syncing {self.name}")
 
+    @property
+    def availability_strategy(self) -> Optional[AvailabilityStrategy]:
+        return None
```
erohmensing, line 141: TODO: make default, need to mock the `check_availability` method in a lot of tests that break

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -106,6 +107,11 @@ def read(
                         f"The requested stream {configured_stream.stream.name} was not found in the source."
                         f" Available streams: {stream_instances.keys()}"
                     )
+                if self.availability_strategy is not None:
+                    stream_is_available, error = self.availability_strategy.check_availability(stream_instance)
+                    if not stream_is_available:
+                        logger.info(f"Skipped syncing stream '{stream_instance.name}' because it was unavailable. Error: {error}")
```
erohmensing, line 113: @evantahler what was it you said about info we want in the error message? Also is this only for logging to the stream sync info, or should we output an AirbyteConnectionStatusMessage or something similar?

airbyte-cdk/python/airbyte_cdk/sources/availability_strategy.py
```diff
@@ -0,0 +1,97 @@
...
+                sync_mode=SyncMode.full_refresh,
+            )
+        )
+        try:
+            return next(slices)
+        except StopIteration:
+            return {}
+
+
+class ScopedAvailabilityStrategy(AvailabilityStrategy):
```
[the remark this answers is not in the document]
erohmensing: The plan is to remove this before merging and keep it on another branch. As we review PRs for connectors implementing availabilitystrategies, if we see enough common functionality, we can make a specific availabilitystrategy for scoped availability.

I'll keep this here while still working though, so that it gets refactored along with anything else so it's easier to slot back in if necessary.

airbyte-cdk/python/airbyte_cdk/sources/availability_strategy.py
```diff
@@ -0,0 +1,145 @@
...
+
+        :param source: source
+        :param logger: source logger
+        :param stream: stream
+        :param error: HTTPerror raised while checking stream's availability.
+        :return: A tuple of (boolean, str). If boolean is true, then the stream
+          is available. Otherwise, the stream is unavailable for some reason and
+          the str should describe what went wrong.
+        """
+        if error.response.status_code == requests.codes.FORBIDDEN:
```
erohmensing, line 70: Any others we want to handle out of the box? E.g. 401 may mean that a single-use token didn't get replaced with a valid one?

airbyte-cdk/python/airbyte_cdk/sources/availability_strategy.py
```diff
@@ -0,0 +1,145 @@
...
+            error_message = "This is most likely due to insufficient permissions on the credentials in use. "
+            error_message += self._visit_docs_message(source, logger)
+            return False, error_message
+
+        error_message = repr(error)
+        return False, error_message
```
bleonard, line 76: Similar to the other PR. Is it safer to default this to `True`? Otherwise ,we would be hiding lots of errors. Or maybe we want to skip ahead on 500. I would expect a failed sync.

airbyte-cdk/python/airbyte_cdk/sources/availability_strategy.py
```diff
@@ -0,0 +1,145 @@
...
+        :param source: source
+        :param logger: source logger
+        :param stream: stream
+        :param error: HTTPerror raised while checking stream's availability.
+        :return: A tuple of (boolean, str). If boolean is true, then the stream
+          is available. Otherwise, the stream is unavailable for some reason and
+          the str should describe what went wrong.
+        """
+        if error.response.status_code == requests.codes.FORBIDDEN:
+            error_message = "This is most likely due to insufficient permissions on the credentials in use. "
```
alafanechere, line 71: What do you think about mentioning we considered unavailability because we received a `403` HTTP response?

alafanechere (commented):
Nice work here and great readability 👏

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -106,6 +107,11 @@ def read(
                         f"The requested stream {configured_stream.stream.name} was not found in the source."
                         f" Available streams: {stream_instances.keys()}"
                     )
+                if self.availability_strategy is not None:
+                    stream_is_available, error = self.availability_strategy.check_availability(logger, stream_instance)
```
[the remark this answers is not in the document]
erohmensing: > I'd be interested to see what it would take to manage availability at the stream level (I think this is an option you're considering).

Yep, looking into that! 

> Moreover, what do you think about building a newly configured catalog from which we've stripped the unavailable streams in a specific availability_check step? The availability check could occur between discover and read.

Not sure I understand here, is the idea to pull the logic completely out of read, and not read sources that we've just previously determined to be available? Is this assuming we run discover and get a new configured catalog before every read, or just the one time when we create the initial catalog for the user to configure?

airbyte-cdk/python/airbyte_cdk/sources/availability_strategy.py
```diff
@@ -0,0 +1,145 @@
...
+        except StopIteration:
+            return {}
+
+    def _visit_docs_message(self, source: Source, logger: logging.Logger) -> str:
+        """
+        :param source:
+        :return: A message telling the user where to go to learn more about the source.
+        """
+        try:
+            connector_spec = source.spec(logger)
```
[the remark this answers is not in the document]
erohmensing: Haha, I don't really either. I guess we want to pass it along to anything that might hypothetically want to log something? 🤷🏻

airbyte-cdk/python/airbyte_cdk/sources/availability_strategy.py
erohmensing: Ooh, I like that idea!

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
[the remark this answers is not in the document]
alafanechere: You could change  `read_catalog` [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/source.py#L89)  to return a catalog with only the available streams.  This method is called in the entrypoint [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/entrypoint.py#L119).

airbyte-cdk/python/airbyte_cdk/sources/availability_strategy.py
erohmensing: To answer both you and @alafanechere's same question, I had initially thought we wanted to except all HTTPErrors, since otherwise this would fix 403 issues but nothing else. I think the reason I thought this was that I was thinking that the main goal of this PR was to stop failing streams from failing a whole sync, when it's really more about gracefully handling the case of 403ing.

I've changed it to default to `True` (aka still available, don't catch errors, things will probably fail) for other HTTPErrors that might pop up. This leads us to more failed syncs (less than before, but more than if we excepted them all), but we can check in on which HTTPError is the new most common cause of stream sync failures and handle that code, and iterate that way. 

Alternatively we could except all HTTPErrors so that the "partial sync" (not yet a concept but you get the idea) still succeeds, and directly report the errors that are excepted so that we can still do that iteration, though its worth noting in this case that we're less motivated by errored syncs to fix those...

>  Or maybe we want to skip ahead on 500. I would expect a failed sync.

Could you elaborate on this? I'd definitely expect a failed sync on a 500, but that doesn't mean that all of the streams have to fail if their endpoints are still working. That's the main argument toward my initial implementation. 

Something that could cause false positives would be something like rate limiting and 429s, but we should already have code that handles that with retries -- one thing on my TODO is to make sure that works as expected, i.e. that the amount of retrying on the `next(records)` call in `check_connection` is the same as when reading records in the read function itself. We don't want to have to deal with retryable error code in this class, I think.

alafanechere: > I had initially thought we wanted to except all HTTPErrors, since otherwise this would fix 403 issues but nothing else
I think, for now, only handling a specific set of HTTPError is safer: we know precisely that a stream is unavailable for a permission error and can support the user accordingly. 

I don't think transient HTTP errors like 429 or 503 should be used to characterize stream availability at the moment as these errors can disappear during the read.

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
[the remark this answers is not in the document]
erohmensing: Great idea, makes sense, will implement in a separate PR! Made a [new issue](https://app.zenhub.com/workspaces/connector-operations-6318ede5eda32e0010ef8566/issues/airbytehq/airbyte/20252) for it

airbyte-cdk/python/airbyte_cdk/sources/streams/availability_strategy.py
```diff
@@ -0,0 +1,78 @@
...
+        :param logger: source logger
+        :param source: (optional) source
+        :return: A tuple of (boolean, str). If boolean is true, then the stream
+          is available, and no str is required. Otherwise, the stream is unavailable
+          for some reason and the str should describe what went wrong and how to
+          resolve the unavailability, if possible.
+        """
+
+
+class ScopedAvailabilityStrategy(AvailabilityStrategy):
```
erohmensing, line 36: This class will be removed before merging. As we review PRs for connectors implementing availabilitystrategies, if we see enough common functionality, we can make a specific availabilitystrategy for scoped availability.

airbyte-cdk/python/unit_tests/sources/utils/test_schema_helpers.py
```diff
@@ -12,14 +13,13 @@
...
 import jsonref
-from airbyte_cdk.logger import AirbyteLogger
 from airbyte_cdk.models.airbyte_protocol import ConnectorSpecification, FailureType
 from airbyte_cdk.sources.utils.schema_helpers import ResourceSchemaLoader, check_config_against_spec_or_exit
 from airbyte_cdk.utils.traced_exception import AirbyteTracedException
 from pytest import fixture
 from pytest import raises as pytest_raises
 
-logger = AirbyteLogger()
+logger = logging.getLogger("airbyte")
```
erohmensing, line 22: This was to address a deprecation warning I noticed while running tests

airbyte-cdk/python/unit_tests/sources/test_availability.py
```diff
@@ -0,0 +1,193 @@
...
+
+    stream_1_is_available, _ = stream_1.check_availability(logger)
+    assert stream_1_is_available is True
+
+    stream_2_is_available, reason = stream_2.check_availability(logger)
+    assert stream_2_is_available is False
+    assert "Could not reach stream 'unavailable_stream'" in reason
+
+
+def test_scoped_availability_strategy():
```
erohmensing, line 97: This test will be removed before merging. As we review PRs for connectors implementing availabilitystrategies, if we see enough common functionality, we can make a specific availabilitystrategy for scoped availability.

airbyte-cdk/python/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.13.0
```
erohmensing, line 3: Open Question: Do I need to describe this feature in the documentation anywhere? Tried to follow `ConnectionChecker` but don't see anything about it in any markdown files.

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -106,6 +106,10 @@ def read(
                         f"The requested stream {configured_stream.stream.name} was not found in the source."
                         f" Available streams: {stream_instances.keys()}"
                     )
+                stream_is_available, error = stream_instance.check_availability(logger, self)
+                if stream_is_available is False:
+                    logger.warning(f"Skipped syncing stream '{stream_instance.name}' because it was unavailable. Error: {error}")
+                    continue
```
erohmensing, line 112: This logic, as well as the logic above it checking whether streams in the configured catalog are in the source's streams, is due to be moved to a `filter_unavailable_streams` method in a separate PR. See #20252

airbyte-cdk/python/CHANGELOG.md
erohmensing: Maybe add a section here? https://docs.airbyte.com/connector-development/cdk-python/http-streams

Probably worth documenting `HttpAvailabilityStrategy` but I'm not sure about the superclass.

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -106,6 +106,10 @@ def read(
                         f"The requested stream {configured_stream.stream.name} was not found in the source."
                         f" Available streams: {stream_instances.keys()}"
                     )
+                stream_is_available, error = stream_instance.check_availability(logger, self)
+                if stream_is_available is False:
```
alafanechere, line 110: nit for extra readability
```suggestion
                if not stream_is_available:
```
I find it safe as you documented check_availability returns `Tuple[bool, str]`. Feel free to ignore it if you think it's not safe :)

airbyte-cdk/python/unit_tests/sources/test_source.py
```diff
@@ -462,3 +473,116 @@ def test_source_config_transform_and_no_transform(abstract_source, catalog):
...
+
+    # Don't set anything for read_records return value for HttpStream, since
+    # it should be skipped due to the stream being unavailable
+    non_http_stream.read_records.return_value = iter([{}] * 3)
+
+    # Patch HTTP request to stream endpoint to make it unavailable
+    req = requests.Response()
+    req.status_code = 403
+    mocker.patch.object(requests.Session, "send", return_value=req)
+
```
alafanechere, line 572: Nit: you could use `requests-mock` ([example](https://github.com/airbytehq/airbyte/blob/master/tools/ci_credentials/tests/test_secrets_manager.py#L183))

airbyte-cdk/python/unit_tests/sources/test_source.py
```diff
@@ -462,3 +473,116 @@ def test_source_config_transform_and_no_transform(abstract_source, catalog):
...
+        def path(self, **kwargs) -> str:
+            return ""
+
+        def parse_response(self, response: requests.Response, **kwargs) -> Iterable[Mapping]:
+            stub_response = {"data": self.resp_counter}
+            self.resp_counter += 1
+            yield stub_response
+
+    class MockStream(MagicMock, Stream):
```
alafanechere, line 548: nit: `MagicMock` is also available under `mocker.MagicMock` (allows lighter imports)

alafanechere (approved):
Nice work 👏 I left minor comments. LGTM, but please address @girarda questions before  publishing the CDK and merging 😄

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
erohmensing: Yes true! I had to change some of my pytest assertions to `is True` instead of `== True`, and then coming back to this part days later my brain must have mapped that onto here. But I could update those tests with this pattern too -- will do

airbyte-cdk/python/airbyte_cdk/sources/declarative/checks/check_stream.py
```diff
@@ -36,26 +36,10 @@ def check_connection(self, source: Source, logger: logging.Logger, config: Mappi
             if stream_name in stream_name_to_stream.keys():
                 stream = stream_name_to_stream[stream_name]
                 try:
-                    # Some streams need a stream slice to read records (eg if they have a SubstreamSlicer)
-                    stream_slice = self._get_stream_slice(stream)
-                    records = stream.read_records(sync_mode=SyncMode.full_refresh, stream_slice=stream_slice)
-                    next(records)
+                    stream_helper = StreamHelper()
+                    stream_helper.get_first_record(stream)
```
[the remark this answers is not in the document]
erohmensing: Yes that actually makes a lot of sense! The only issue I see is that they don't _quite_ have the same functionality -- the `HTTPAvailabilityStrategy` currently handles 403 errors, but excepts other HTTPErrors and returns that the stream is still available. The check_stream behavior at the moment would still raise other HTTPErrors.  

@alafanechere I think this brings up a good point - right now the default behavior is other non-403 HTTPErrors will still cause issues and fail the whole sync - but while syncing (and potentially after having synced other streams), not during this check. Maybe we should make the default behavior do something like (pseudocode)
```
except HTTPError as http_error:
  if 403: 
    return False, "got 403"
  else:
    raise http_error
```

That would make us be able to do what @girarda suggested here, and make unexpected HTTPErrors fail the sync before it starts rather than after some streams have already synced.

airbyte-cdk/python/airbyte_cdk/sources/streams/http/availability_strategy.py
```diff
@@ -0,0 +1,113 @@
...
+        :param source: optional (source)
+        :param error: HTTPError raised while checking stream's availability.
+        :return: A tuple of (boolean, str). If boolean is true, then the stream
+          is available, and no str is required. Otherwise, the stream is unavailable
+          for some reason and the str should describe what went wrong and how to
+          resolve the unavailability, if possible.
+        """
+        try:
+            status_code = error.response.status_code
+            reason = self.reasons_for_unavailable_status_codes(stream, logger, source)[status_code]
```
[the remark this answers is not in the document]
erohmensing: Definitely, thanks for pointing out that method

airbyte-cdk/python/unit_tests/sources/test_source.py
erohmensing: I pulled this version from this [example](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/unit_tests/sources/streams/http/test_http.py#L144-L147), will keep it since we already import requests for type checking

girarda (approved):

airbyte-cdk/python/airbyte_cdk/sources/utils/stream_helpers.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        Gets the first record for a stream.
+
+        :param stream: stream
+        :return: StreamData containing the first record in the stream
+        """
+        # Some streams need a stream slice to read records (e.g. if they have a SubstreamSlicer)
+        stream_slice = self.get_stream_slice(stream)
+        records = stream.read_records(sync_mode=SyncMode.full_refresh, stream_slice=stream_slice)
+        next(records)
```
bazarnov, line 23: This part should be replaced with this to avoid `StopIteration` error on empty streams:

```python
class StreamHelper:
    def get_first_record(self, stream: Stream) -> StreamData:
        """
        Gets the first record for a stream.

        :param stream: stream
        :return: StreamData containing the first record in the stream
        """
        try:
            # Some streams need a stream slice to read records (e.g. if they have a SubstreamSlicer)
            stream_slice = self.get_stream_slice(stream)
            records = stream.read_records(sync_mode=SyncMode.full_refresh, stream_slice=stream_slice)
            return next(records)
        except StopIteration:
            return {}
```

PR #19983, opened 2022-12-01
Add options to tooltip of oneOf properties

### LINKED ISSUE #17975: Show descriptions of oneOf field choices in connector setup form
As a result of [this OC issue](https://github.com/airbytehq/oncall/issues/740), it was discovered that for `oneOf` fields in a connector spec, we currently only display the description of the `oneOf` field in the connector form, but don't display descriptions of the actual choices themselves anywhere.

For example, the Replication Method field in the Postgres source spec has the following (simplified) structure:
```
"replication_method": {
  "type": "object",
  "title": "Replication Method",
  "description": "Replication method for extracting data from the database.",
  "oneOf": [
    {
      "title": "Standard",
      "description": "Standard replication requires no setup on the DB side but will not be able to represent deletions incrementally.",
      ...
      }
    },
    {
      "title": "Logical Replication (CDC)",
      "description": "Logical replication uses the Postgres write-ahead log (WAL) to detect inserts, updates, and deletes. This needs to be configured on the source database itself. Only available on Postgres 10 and above. Read the <a href=\"https://docs.airbyte.com/integrations/sources/postgres\">docs</a>.",
      "properties": {
        "plugin": {
          "type": "string",
          "title": "Plugin",
          "description": "A logical decoding plugin installed on the PostgreSQL server. The `pgoutput` plugin is used by default. If the replication table contains a lot of big jsonb values it is recommended to use `wal2json` plugin. Read more about <a href=\"https://docs.airbyte.com/integrations/sources/postgres#step-2-select-a-replication-plugin\">selecting replication plugins</a>.",
          ...
        },
        "replication_slot": {
          "type": "string",
          "title": "Replication Slot",
          "description": "A plugin logical replication slot. Read about <a href=\"https://docs.airbyte.com/integrations/sources/postgres#step-3-create-replication-slot\">replication slots</a>.",
        },
        ...
      }
    }
  ]
}
```

But in the UI, we only have a tooltip showing the `Replication Method`'s description:
Screen Shot 2022-10-13 at 2 24 27 PM
and tooltips on the _properties_ of the `Logical Replication (CDC)` object:
Screen Shot 2022-10-13 at 2 36 10 PM

We don't currently have any place to show the description of the `Logical Replication (CDC)` choice itself. We should figure out a way to display these oneOf-choice descriptions, as they are helpful for users and could help us avoid OC issues like the one linked above.

For oneOf fields in connector specs, display the descriptions of the oneOf-choices somewhere in the connector form UI.

Ideally, these descriptions would be viewable even before a user actually chooses one of the options in the dropdown, so that they can use the descriptions to help them decide which option to choose. This may require us to put InfoTooltips on each choice in the dropdown menu itself, so that users can hover over them and read descriptions of each choice.

**Feedback from design**: Render out the descriptions in the overall field tooltip, similar to how this connector description is set up
image

## Acceptance Criteria
- Descriptions of oneOf field choices in connector specs are displayed somewhere in the connector setup form UI

flash1293:
## What

Closes https://github.com/airbytehq/airbyte/issues/17975 by adding the options of condition nodes as a list to the tooltip of the form group.

Before: 
Screenshot 2022-12-01 at 16 15 02

After:
Screenshot 2022-12-01 at 15 54 56

## How

When rendering the `PropertyLabel` for a condition node, build a list of options with their descriptions (use title if available, otherwise fall back to key which is used for the dropdown as well) and pass it down to the `LabelInfo` component where it`s rendered similar to the description.

I debated a bit where to put the logic for turning the options of a condition into a list of descriptions - I settled on the `PropertyLabel` because it's the last component in the stack that still knows about condition properties. Putting it into `ConditionSection` would be closer to other code concerning conditions, but it makes it necessary to introduce a new type and pass it through a few layers which seems less clean eventually.

## 🚨 User Impact 🚨

This looks a little weird if the description of the group has been used to describe the options already (looks like this only affects a few connectors which I updated in a separate PR (https://github.com/airbytehq/airbyte/pull/19984), but it might be an issue for custom connectors, too): 
Screenshot 2022-12-01 at 15 57 56

Also, I'm not 100% sure about the h5 for the "Options" title because the font size is so small, but it seems like the right thing semantically (as the title of the tooltip is a h4)

### DIFF
--- a/airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.module.scss
+++ b/airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.module.scss
@@ -18,6 +18,17 @@
   margin-top: 5px;
 }
 
+.optionsHeader {
+  margin-top: 0;
+  margin-bottom: 5px;
+}
+
+.options {
+  padding-left: 25px;
+  margin-bottom: 5px;
+  margin-top: 5px;
+}
+
 .exampleHeader {
   margin-block-start: 0;
   margin-bottom: 5px;
--- a/airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.tsx
+++ b/airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.tsx
@@ -10,6 +10,7 @@ interface LabelInfoProps {
   label: React.ReactNode;
   examples?: JSONSchema7Type;
   description?: string;
+  options?: Array<{ title: string; description?: string }>;
 }
 
 const Description: React.FC<Pick<LabelInfoProps, "label" | "description">> = ({ label, description }) => {
@@ -22,6 +23,33 @@ const Description: React.FC<Pick<LabelInfoProps, "label" | "description">> = ({ 
       {/* don't use <Text as=h4> here, because we want the default parent styling for this header */}
       <h4 className={styles.descriptionHeader}>{label}</h4>
       <TextWithHTML className={styles.description} text={description} />
+    </div>
+  );
+};
+
+const Options: React.FC<Pick<LabelInfoProps, "options">> = ({ options }) => {
+  if (!options) {
+    return null;
+  }
+
+  return (
+    <div>
+      <h5 className={styles.optionsHeader}>
+        <FormattedMessage id="connector.optionsHeader" />
+      </h5>
+      <ul className={styles.options}>
+        {options.map((option) => (
+          <li key={option.title}>
+            <strong>{option.title}</strong>
+            {option.description && (
+              <>
+                {" "}
+                - <TextWithHTML className={styles.description} text={option.description} />
+              </>
+            )}
+          </li>
+        ))}
+      </ul>
     </div>
   );
 };
@@ -48,10 +76,11 @@ const Examples: React.FC<Pick<LabelInfoProps, "examples">> = ({ examples }) => {
   );
 };
 
-export const LabelInfo: React.FC<LabelInfoProps> = ({ label, examples, description }) => {
+export const LabelInfo: React.FC<LabelInfoProps> = ({ label, examples, description, options }) => {
   return (
     <div className={styles.container}>
       <Description label={label} description={description} />
+      <Options options={options} />
       <Examples examples={examples} />
     </div>
   );
--- a/airbyte-integrations/connectors/destination-postgres-strict-encrypt/src/test/resources/expected_spec.json
+++ b/airbyte-integrations/connectors/destination-postgres-strict-encrypt/src/test/resources/expected_spec.json
@@ -56,14 +56,14 @@
       },
       "ssl_mode": {
         "title": "SSL modes",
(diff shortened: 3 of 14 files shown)


airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.module.scss
```diff
@@ -18,6 +18,17 @@
   margin-top: 5px;
 }
 
+.optionsHeader {
+  margin-top: 0;
+  margin-bottom: 5px;
```
lmossman, line 23:
```suggestion
  margin-bottom: variables.$spacing-sm;
```

nit: we should replace the px values in this file with our `_variables.scss` spacing values: https://github.com/airbytehq/airbyte/blob/5da3a7c2a80814bef8cc10e18f4faad93229d3da/airbyte-webapp/src/scss/_variables.scss#L16-L20

For ones that don't have a matching value (e.g. 25px), we should try out the closest value that matches, e.g. `spacing-xl` and see if that looks fine.

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.tsx
```diff
@@ -26,6 +27,33 @@ const Description: React.FC<Pick<LabelInfoProps, "label" | "description">> = ({
...
 };
 
+const Options: React.FC<Pick<LabelInfoProps, "options">> = ({ options }) => {
+  if (!options) {
+    return null;
+  }
+
+  return (
+    <div>
+      <h5 className={styles.optionsHeader}>
```
lmossman, line 37: I agree that the `Options` title feels too small - maybe we should use a larger heading for the tooltip tile (e.g. h3) and use h4 here instead

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.tsx
```diff
@@ -26,6 +27,33 @@ const Description: React.FC<Pick<LabelInfoProps, "label" | "description">> = ({
...
+        <FormattedMessage id="connector.optionsHeader" />
+      </h5>
+      <ul className={styles.options}>
+        {options.map((option) => (
```
lmossman, line 41: Could we add a bit of spacing in between these options? I think it would look a bit nicer if each option was slightly separated from the others

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.module.scss
```diff
@@ -18,6 +18,17 @@
   margin-top: 5px;
 }
```
lmossman, line 20: It looks like you forked the `airbytehq` repo to create this PR. Our pattern is to use feature branches instead of forks for our PRs, since we have a bunch of CI checks that only run on feature branches and not on forks.

We may actually need to recreate this PR on a feature branch instead of a fork to get those CI checks to run

flash1293:
Thanks for your comments @lmossman - I opened a new PR over here with all of them addressed: https://github.com/airbytehq/airbyte/pull/20015

Closing this one.

PR #20000, opened 2022-12-01
MySQL Source : Check for `REQUIRE CLIENT` privilege in CDC

akashkulk:
Currently, we do not check whether the user has `REQUIRE CLIENT` privilege which is necessary to access the binary logs. For CDC, the sync fails if the user doesn't have this permission. This PR adds that to the check. 

This has been added to the check method of MySQL source

Closes https://github.com/airbytehq/alpha-beta-issues/issues/506


airbyte-integrations/connectors/source-mysql/src/main/java/io/airbyte/integrations/source/mysql/helpers/CdcConfigurationHelper.java
```diff
@@ -38,12 +41,35 @@ public class CdcConfigurationHelper {
...
+  // the binary log files, which are needed for CDC.
+  private static CheckedConsumer<JdbcDatabase, Exception> getMasterStatusOperation() {
+    return database -> {
+      try {
+        database.unsafeResultSetQuery(
+            connection -> connection.createStatement().executeQuery("SHOW MASTER STATUS"),
+            resultSet -> {
+              final String file = resultSet.getString("File");
+              final int position = resultSet.getInt("Position");
+              if (file == null || position == 0) {
```
rodireich, line 61: Is it possible that file exists but position is legitimately 0, as in at the top?

akashkulk: Yes - actually this is logic from `MySqlTargetPosition`. I'm changing this now to just attempt to execute the query (and not to fail if file or position is empty/null.

rodireich (approved):

akashkulk:
/publish connector=connectors/source-mysql-strict-encrypt run-tests=false

PR #20004, opened 2022-12-01
:window: :wrench: Generify GroupControls to make it more reusable

lmossman:
## What
The Connector Builder needs to include inputs similar to the ConditionSection of the ConnectorForm, in which there is a label with a dropdown to select which fields are shown inside of the box. However, the styling and positioning of the dropdown is currently contained in the `ConditionSection` component, which has other assumptions and logic specific to the ConnectorForm.

## How
Instead, we would like to reuse the generic GroupControls component to implement this in the connector builder, so to avoid repeating the dropdown styling, this PR makes the label and dropdown explicit props, rather than taking in a single `title` prop that contains both, and moves the styling of these into the GroupControls scss file. 

This way, the Connector Builder can reuse the GroupControls component to get the same look as that in the ConnectorForm, without needing to repeat styling, such as in [this feature branch](https://github.com/airbytehq/airbyte/blob/b17fe0511aae3ef929fb8edffcd893ea5fc7a245/airbyte-webapp/src/components/connectorBuilder/Builder/BuilderOneOf.tsx#L39).

## 🚨 User Impact 🚨
There should be no user-visible changes as a result of this PR.


airbyte-webapp/src/components/GroupControls/GroupControls.module.scss
```diff
@@ -6,7 +6,6 @@ $group-spacing: variables.$spacing-xl;
 $border-width: variables.$border-thick;
 
 .container {
-  margin-bottom: 27px;
```
lmossman, line 9: Removed this because we don't want this spacing in the Connector Builder. To not affect the look of the ConnectorForm, I opted to wrap the ArraySection and ConditionSection components in `SectionContainers` instead: https://github.com/airbytehq/airbyte/pull/20004/files#diff-359ca8d25eceab9f588e4ca17095865f532588fecd0d7d9f2fc76458c1b1f10bR66

This is similar to how we add the correct spacing in FormSection

airbyte-webapp/src/components/GroupControls/GroupControls.module.scss
```diff
@@ -27,6 +26,23 @@ $border-width: variables.$border-thick;
...
+  padding-right: variables.$spacing-md;
+  display: flex;
+  align-items: center;
+  background-color: colors.$white;
+  white-space: nowrap;
+}
+
+.dropdown {
+  margin-left: auto;
+  padding: 0 2px;
```
flash1293, line 41: nit: should we use a size variable here?

flash1293 (approved):
Tested and seems to work fine, no regressions that I noticed. LGTM

PR #20008, opened 2022-12-01
[Connector Builder] Configuration UI MVP

lmossman:
## What
This PR adds the first iteration of the Configuration UI to the Connector Builder, allowing users to fill out UI forms rather than hand-writing a YAML file.

I have recorded a loom where I go through what it looks like to build a simple connector in the new UI, and the various other features that have been added:
https://www.loom.com/share/9cf4ddb8ad894503bbc82cda5606aa2b

The main thing left to address for this first iteration is to better handle error cases, e.g. if a required field is not filled out and the user tries to test, currently they will see a somewhat confusing error message. This will be addressed in a later PR.

## How
This PR is doing quite a bit -- apologies for such a large changeset. The motivation behind putting all of these changes in one PR was that I didn't want to add the new UI view until it reached at least minimum usability for building a simple connector like the Poke API.

However, I could have still created incremental PRs into this feature branch to get review as I went, so I will aim to do things that way in the future.

## Recommended reading order
1. Start at the top of the builder component structure and go down: ConnectorBuilderPage.tsx -> Builder.tsx -> BuilderSidebar.tsx -> Global/StreamConfigView.tsx -> BuilderField.tsx...
2. Also important is reviewing the changes to the ConnectorBuilderStateService, as that is now handling both the UI form state and the yaml editor state.

## 🚨 User Impact 🚨
Users can now configure connectors through UI forms in the connector builder. More functionality to come soon!


airbyte-webapp/src/components/ui/ListBox/ListBox.module.scss
```diff
@@ -33,6 +33,7 @@
   padding: variables.$spacing-lg variables.$spacing-md;
   border-radius: variables.$border-radius-lg;
   cursor: pointer;
+  overflow: auto;
```
lmossman, line 36: So that very long entries in the listbox menu have a horizontal scroll

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+        <AddStreamButton
+          className={styles.addStreamButton}
+          numStreams={numStreams}
+          onAddStream={(addedStreamNum, addedStreamName) => onViewSelect(addedStreamNum, addedStreamName)}
+        />
+      </div>
+
+      <div className={styles.streamList}>
+        {Array.from(Array(numStreams).keys()).map((streamNum) => {
```
flash1293, line 134: This structure with numStreams seems more complex than necessary - as you don't access anything out of the stream except for the value of name anyway, it seems easier to do a `useField("streams")` here in the sidebar and then simply map it here like this:
```js
streams.map(stream => (<StreamSelectButton stream={stream} {/* ... */} />))
```

airbyte-webapp/src/components/connectorBuilder/Builder/UiYamlToggleButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <button className={classnames(styles.button, className)} onClick={onClick}>
+      <Text
+        className={classnames(styles.ui, styles.text, {
+          [styles.selected]: !yamlSelected,
+          [styles.unselected]: yamlSelected,
+        })}
+        size="xs"
+        bold
+      >
+        UI
```
flash1293, line 24: i18n

airbyte-webapp/src/components/connectorBuilder/Builder/UiYamlToggleButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+      </Text>
+      <Text
+        className={classnames(styles.yaml, styles.text, {
+          [styles.selected]: yamlSelected,
+          [styles.unselected]: !yamlSelected,
+        })}
+        size="xs"
+        bold
+      >
+        YAML
```
flash1293, line 34: i18n

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+interface StreamConfigViewProps {
+  streamNum: number;
+}
+
+export const StreamConfigView: React.FC<StreamConfigViewProps> = ({ streamNum }) => {
+  const streamPath = (path: string) => `streams[${streamNum}].${path}`;
+
+  return (
+    <BuilderCard>
```
flash1293, line 12: various i18ns

airbyte-webapp/src/components/connectorBuilder/Builder/UiYamlToggleButton.module.scss
```diff
@@ -0,0 +1,36 @@
...
+
+.button {
+  cursor: pointer;
+  border: variables.$border-thin solid colors.$dark-blue;
+  background-color: colors.$dark-blue;
+  display: flex;
+  border-radius: 6px;
+  padding: 0;
+  overflow: hidden;
+  height: 21px;
```
flash1293, line 12: do we need to hardcode the size like this? It seems like we can instead work with paddings

airbyte-webapp/src/components/connectorBuilder/Builder/UiYamlToggleButton.module.scss
```diff
@@ -0,0 +1,36 @@
...
+  height: 100%;
+  flex: 1;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  font-weight: 700;
+}
+
+.selected {
+  background-color: colors.$white;
```
flash1293, line 29: Maybe it's just me, but it feels like these are flipped. I perceive the darker one as the "selected" one

airbyte-webapp/src/components/connectorBuilder/Builder/Builder.tsx
```diff
@@ -0,0 +1,55 @@
+import { Form } from "formik";
+import { useEffect, useState } from "react";
+
+import { useConnectorBuilderState } from "services/connectorBuilder/ConnectorBuilderStateService";
+import { usePatchFormik } from "views/Connector/ConnectorForm/useBuildForm";
```
flash1293, line 5: As we use the formik patch trick in multiple places now, should we pull it into `core/form`?

airbyte-webapp/src/components/connectorBuilder/YamlEditor/YamlEditor.tsx
```diff
@@ -1,26 +1,37 @@
...
+import { useEffect, useMemo, useRef, useState } from "react";
 
 import { CodeEditor } from "components/ui/CodeEditor";
 
 import { ConnectorManifest } from "core/request/ConnectorBuilderClient";
-import { useManifestTemplate } from "services/connectorBuilder/ConnectorBuilderApiService";
+import { useConfirmationModalService } from "hooks/services/ConfirmationModal";
 import { useConnectorBuilderState } from "services/connectorBuilder/ConnectorBuilderStateService";
 
+import { convertToManifest } from "../Builder/types";
```
flash1293, line 13: As these are used in both the builder and the yaml editor, should we pull them into the top level connector builder directory?

airbyte-webapp/src/components/connectorBuilder/Builder/AddStreamButton.tsx
```diff
@@ -0,0 +1,92 @@
...
+        onClick={() => {
+          setIsOpen(true);
+        }}
+        icon={<PlusIcon className={styles.plus} />}
+      />
+      {isOpen && (
+        <Formik
+          initialValues={{ streamName: "", urlPath: "" }}
+          onSubmit={(values: AddStreamValues) => {
+            setFieldValue(`streams[${numStreams}]`, {
```
flash1293, line 42: nit: I'm not 100% sure whether it's a problem, but it looks like a code smell to me that adding a new item to an array is done by setting a numeric property of the next index. The more idiomatic way seems to set the field value of `streams` and spread the existing array of streams and adding the new one:
```
setFieldValue("streams", [...oldStreams, {
name: values.streamName,
              urlPath: values.urlPath,
              fieldPointer: [],
              httpMethod: "GET",
}])
```

Feel free to disregard

flash1293 (commented):
Left some small comments in the code, but they are mostly nitpicks.

Some general observations:

* Do we want to allow to test the connector if base url is not defined / invalid in some way we know?:

Screenshot 2022-12-08 at 11 27 20

* Weird idea - I found myself trying to do command+enter to re-test, what do you think?
* I stopped commenting on every part but it seems like there are a bunch of phrases not i18n-ified yet in various components

airbyte-webapp/src/components/connectorBuilder/Builder/UiYamlToggleButton.module.scss
flash1293: I think I realized why it was done this way - it's to make sure both sides are the same size. You can do this with grid layout:
```
  display: grid; 
  grid-template-columns: 1fr 1fr; 
  grid-template-rows: 1fr; 
  gap: 0px 0px; 
  grid-template-areas: 
    ". ."; 
```

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
lmossman: Thanks for calling these out. I originally didn't do i18n for the BuilderField strings because I wanted to keep them easy to maintain and add to in the future. If we put them all behind i18n, then every time we add a new field we would also need to update the i18n strings and point it to there which sounded cumbersome to me for a process that we will need to do a lot.

But since we are i18ning all of our other text in the webapp, maybe we should just bite the bullet and do it here as well

airbyte-webapp/src/pages/ConnectorBuilderPage/ConnectorBuilderPage.tsx
```diff
@@ -1,36 +1,64 @@
...
-        minWidth: 60,
-        overlay: {
-          displayThreshold: 325,
-          header: formatMessage({ id: "connectorBuilder.testConnector" }),
-          rotation: "counter-clockwise",
-        },
+    <Formik
+      initialValues={builderFormValues}
+      onSubmit={(values: BuilderFormValues) => {
+        console.log(values);
```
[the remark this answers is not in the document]
lmossman: It is not. This is actually a slightly weird form because we are never really "submitting" it, so I've just changed this to be a no op (since Formik still requires you to specify an onSubmit function)

airbyte-webapp/src/components/connectorBuilder/Builder/AddStreamButton.module.scss
```diff
@@ -0,0 +1,23 @@
+@use "scss/variables";
+@use "scss/colors";
+
+.body {
+  display: flex;
+  flex-direction: column;
+  gap: variables.$spacing-xl;
+}
+
+.addButton {
```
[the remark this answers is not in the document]
lmossman: I really need the 9px padding to make the `+` look right in the add button, and we don't have a value in `_variables` for component sizes. But I did capture the 26px in a local variable for this file so that the px value isn't repeated

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.tsx
lmossman: Good suggestion - I actually was able to simplify this even further and completely eliminate the `StreamSelectButton` component based on your suggestion

airbyte-webapp/src/components/connectorBuilder/Builder/UiYamlToggleButton.module.scss
lmossman: I actually felt that way too when I first added this, and convinced myself that the designer knew better. But since you have also voiced this, I think I'll swap them

airbyte-webapp/src/components/connectorBuilder/Builder/Builder.tsx
lmossman: I took a closer look at this and I think we actually don't need it at all in the builder since this form doesn't get submitted. So I'm just going to remove it from the builder instead

lmossman: Ah but we need it in the AddStreamButton component. I'll move it to core/form then

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
lmossman: I thought about this a bit more - I feel I should fix all of the i18n issues in this PR _except_ for the labels and descriptions on the BuilderField components. My reasoning is that once the connector manifest JSON schema is decoupled from the python classes, we should be able to capture the titles and descriptions for all of the various fields in the JSON schema, and at that point we can pull those into the UI.

So it feels like a bit of a waste of time to migrate all of these to the en.json file, and require any new BuilderFields being added to also update that i18n file and point to it every time we add a new field. Thoughts?

airbyte-webapp/src/components/connectorBuilder/Builder/UiYamlToggleButton.module.scss
lmossman: That definitely seems like a better way to do it. Thanks for the suggestion!

airbyte-webapp/src/components/connectorBuilder/Builder/AddStreamButton.tsx
lmossman: I think this makes a lot of sense, I made this change and it allowed me to simplify a few different components 👍 thanks for the suggestion!

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
flash1293: If we plan to pull in this copy from the schema in the near term I agree it doesn’t make sense to i18n-ify them right now. If possible we should add it as a task to the appropriate issue / create an issue about it as this is one of the things that’s easily missed

lmossman: Done: https://github.com/airbytehq/airbyte/issues/20271

airbyte-webapp/src/services/connectorBuilder/connector_manifest_openapi.yaml
```diff
@@ -0,0 +1,9 @@
+openapi: 3.0.0
+info:
+  title: Connector Manifest schema
+  version: 1.0.0
+paths: {}
+components:
+  schemas:
+    ConnectorManifest:
+      $ref: "../../../../airbyte-cdk/python/airbyte_cdk/sources/declarative/config_component_schema.json"
```
lmossman, line 9: FYI I added this file so that we could use orval to pull in the Connector Manifest json schema and generate typescript types from it. This is necessary to do rather than referencing the schema directly in the connector builder server openapi spec, because of the issue described here: https://github.com/airbytehq/airbyte/issues/20248

lmossman:
> Do we want to allow to test the connector if base url is not defined / invalid in some way we know?

No we do not -- this is what I am currently working on fixing and will be addressing in a follow-up PR. Basically, I will be making disabling the test button if they click it and there are any errors in the formik form, and will work on making it clear which views have errors that need to be fixed.

> Weird idea - I found myself trying to do command+enter to re-test, what do you think?

Interesting, I hadn't considered hotkeys for testing, but I think that could definitely be useful since re-testing is pretty common. Created an issue for that here: https://github.com/airbytehq/airbyte/issues/20273

---

I think with that I have addressed all of your comments. Could you take another look @flash1293

airbyte-webapp/src/components/connectorBuilder/Builder/UiYamlToggleButton.module.scss
timroes: Haven't tried this locally, but shouldn't it be possible to just shorten this to?

```css
  display: grid; 
  grid-template-columns: 1fr 1fr; 
```

airbyte-webapp/src/components/connectorBuilder/Builder/AddStreamButton.module.scss
```diff
@@ -0,0 +1,20 @@
...
+
+$buttonWidth: 26px;
+
+.body {
+  display: flex;
+  flex-direction: column;
+  gap: variables.$spacing-xl;
+}
+
+.addButton {
```
flash1293, line 12: I think it's OK for now, but I don't like how much you need to fight the stylings of the underlying button component.

It seems like it would be possible to make the round icon button a prop on the regular button - this way you don't need to overrule styles with `!important` but you can selectively apply the right styles in the right moment.

If you think that makes sense I can create an issue for it.

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.tsx
```diff
@@ -0,0 +1,136 @@
...
+        <FormattedMessage id="connectorBuilder.globalConfiguration" />
+      </ViewSelectButton>
+
+      <div className={styles.streamsHeader}>
+        <Text className={styles.streamsHeading} size="xs" bold>
+          <FormattedMessage id="connectorBuilder.streamsHeading" values={{ number: values.streams.length }} />
+        </Text>
+
+        <AddStreamButton
+          className={styles.addStreamButton}
```
flash1293, line 112: You are adding some styles here and then there are additional styles in the AddStreamButton component itself. Can we consolidate these in a single place?

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.module.scss
```diff
@@ -0,0 +1,102 @@
...
+  margin-left: variables.$spacing-md;
+  color: colors.$blue;
+}
+
+.addStreamButton {
+  margin-left: auto;
+  margin-right: variables.$spacing-sm;
+}
+
+.downloadButton {
```
flash1293, line 54: It seems like this rule (and the whole div around the button itself) is only needed to push the buttons to the bottom of the screen.

The more idiomatic way of doing this is to give the element in the flex container you want to take the remaining size a `flex-grow: 1` (in this case it would be the stream list). This is how it looks like:
Screenshot 2022-12-09 at 15 12 38

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.module.scss
```diff
@@ -0,0 +1,102 @@
...
+  margin-top: auto;
+  width: 90%;
+}
+
+.resetButton {
+  margin-top: variables.$spacing-sm;
+  margin-bottom: variables.$spacing-md;
+  width: 90%;
```
flash1293, line 62: I'm not so sure about these 90% widths - is there a reason the buttons are not as wide as the sidebar allows, just like the stream list items itself? It looks cleaner to me:
Screenshot 2022-12-09 at 15 12 38

airbyte-webapp/src/components/connectorBuilder/YamlEditor/YamlEditor.module.scss
```diff
@@ -5,17 +5,22 @@
...
   flex-direction: row;
-  justify-content: flex-end;
 }
 
 .editorContainer {
+  padding-top: variables.$spacing-xl;
   flex: 1 1 0;
 }
+
+.downloadButton {
```
flash1293, line 24: You don't need this, you can simply give the `.control` class a `flex-direction: row-reverse` instead of row so it starts from the right

airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/StreamSelector.module.scss
```diff
@@ -11,14 +11,19 @@
 .button {
   padding: variables.$spacing-md;
   background-color: transparent;
-  border-radius: variables.$border-radius-lg;
+  border-radius: variables.$border-radius-md;
   border: none;
   display: flex;
```
flash1293, line 16: In this whole file: Instead of using margins to position stuff, you can use flex layout:
on the button container, set
```
justify-content: center;
gap: variables.$spacing-md;
```

This will put both elements in the center of the flex container with the specified gap. Then you don't need any of the other margins.

airbyte-webapp/src/pages/ConnectorBuilderPage/ConnectorBuilderPage.module.scss
```diff
@@ -1,16 +1,23 @@
...
 .leftPanel {
   overflow: hidden;
 }
 
 .rightPanel {
-  border-top-left-radius: variables.$border-radius-lg;
-  border-bottom-left-radius: variables.$border-radius-lg;
+  @include mixins.left-shadow;
+
+  border-top-left-radius: variables.$border-radius-md;
```
flash1293, line 12: Why are there border radiuses on the right panel? The left sidebar doesn't have them and the setup guide that's shown in a very similar fashion in the connector form doesn't have them either.

airbyte-webapp/src/components/connectorBuilder/Builder/Builder.tsx
```diff
@@ -0,0 +1,45 @@
...
+  useEffect(() => {
+    setBuilderFormValues(values);
+  }, [values, setBuilderFormValues]);
+
+  const [selectedView, setSelectedView] = useState<BuilderView>("global");
+
+  const handleViewSelect = (selectedView: BuilderView, streamName?: string) => {
+    setSelectedView(selectedView);
+    if (selectedView !== "global" && streamName !== undefined) {
+      setSelectedStream(streamName);
```
flash1293, line 28: We are auto-selecting the testing panel for the stream when switching the view, but we don't do it the other way around (switching the view when switching the `selectedStream`). Would it make sense to do so?

Related to that - would it make sense to lift the selected view into the ConnectorBuilderStateService as well? It's conceptually very similar to the selected stream, it confused me at first why it's handled in such a different fashion.

Not a must-have though.

flash1293 (commented):
I focused mostly on small css stuff. In general I think we should invest in general layout components that make it easier to achieve common layouts to make these things much easier, but I will bring that up separately with the frontend chapter.

Besides that and a few nits it starts to look pretty good.

Phrase is not translated in modal:
Screenshot 2022-12-09 at 15 38 34

A more general one - while testing I noticed the UI sends a request for every keystroke which seems like a lot for production systems:
Kapture 2022-12-09 at 15 56 06

Can we throttle these calls?

airbyte-webapp/src/components/connectorBuilder/Builder/UiYamlToggleButton.module.scss
flash1293: Seems like that works too, good catch. I need to update my knowledge of CSS grid, it has been a while since I used it :)

airbyte-webapp/src/pages/ConnectorBuilderPage/ConnectorBuilderPage.module.scss
lmossman: I was trying to match the design here: https://www.figma.com/file/F1ljvrJerLoVbNvtejBWid/LOW-CODE-CONNECTOR-BUILDER?node-id=2643%3A109796&t=VcXAAc8881mHE2Rx-0

Personally I like the border radiuses because they make the right panel feel more like a slide-out view, which differentiates it a bit from the left sidebar since the size of that one cannot be adjusted / it is static.

I wouldn't be opposed to adding the border radiuses to the documentation panel in the setup guide as well

flash1293: > I wouldn't be opposed to adding the border radiuses to the documentation panel in the setup guide as well

good idea, sounds great!

lmossman:
> Can we throttle these calls?

I've pushed this commit to debounce setting the json manifest from the yaml editor, which seems to solve that problem: https://github.com/airbytehq/airbyte/pull/20008/commits/5ffe20e6702cf36bf629ec9acaa5ae60168761fb

airbyte-webapp/src/components/connectorBuilder/Builder/AddStreamButton.module.scss
lmossman: Yeah I think that makes sense - having a prop on the button to easily make it round would be really nice here. If you could create an issue for that, that would be great!

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.tsx
lmossman: This is a pattern I've been doing for a little while, but let me know if you think I shouldn't be doing it:
- When I create a custom component, I've usually been trying to keep positional styling (e.g. the `addStreamButton` style in BuilderSidebar.module.scss which just sets margin values) in the same `.module.scss` file as the parent component that holds the custom component, because that feels like the most natural place to control the position. 
- I then put the styling code for the internals of the custom component (e.g. the `addButton` style in the AddStreamButton.module.scss file) in the custom component's own scss file, since that feels like the most natural place to style internals of a component.

Basically it felt weird to me to put positional styling inside the custom component's own styles, because if I want to use that custom component in another place in the future then I would need to lift those positional styles back up to the parent anyway. What do you think?

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.module.scss
lmossman: Thanks for the suggestion, updated and I agree that feels like a better approach

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.module.scss
lmossman: I agree, I think using the full width is better. Fixed

airbyte-webapp/src/components/connectorBuilder/YamlEditor/YamlEditor.module.scss
lmossman: Done!

airbyte-webapp/src/components/connectorBuilder/Builder/Builder.tsx
lmossman: Good suggestion -- I've added this behavior. The one slightly nuanced aspect of it is that if the Global Configuration view is currently selected while the user changes streams in the testing panel dropdown, I don't think we want it to switch the view to that stream, because the user might want to test out different streams while iterating on global configuration values. So I've encoded that behavior as well.

lmossman:
@flash1293 I made a couple more changes in the last few commits, which do the following:
1. If a stream's name is set to an empty/pure whitespace string, it now shows an (empty) placeholder in the view select button and test panel stream dropdown
2. I switched the tracking of the selected test panel stream to track the index instead of the stream name, as this was necessary to ensure that if a user changes the name of a stream, that stream stays selected in the test panel

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.tsx
flash1293: > But what do you think of that pattern overall?

Thanks for the explanation, that makes a lot of sense to me as a pattern - Ideally I hope we can abstract away most of these concerns into library components, but I'll bring it up in the chapter sync.

flash1293 (approved):
For the scope of this PR it LGTM, great work!

Two notes that should be addressed, can happen later on:

* The "one request per keystroke" also happens for the test input - this hasn't been touched here, so we can address it separately, just something I noticed. Not sure whether we even need to re-fetch the streams in this situation, but I might miss something
* The "add stream" modal doesn't really add any value, it's just making the user flow more complex - it seems like it would be easier to add an empty, untouched stream and switch to the view so the user can fill it in without any modal involved. I realize that this is not according to the current designs so happy to discuss it in some forum before implementing. Let me know if I should create an issue for it.

lmossman:
> The "one request per keystroke" also happens for the test input - this hasn't been touched here, so we can address it separately, just something I noticed. Not sure whether we even need to re-fetch the streams in this situation, but I might miss something

One reason to refetch the streams in this situation is if the user has implemented a stream in which the name or the URL use some interpolated user-input value. Then the name/url of that stream depend on the value they type into the test input menu, so they need to be refetched to resolve those interpolated references.

I won't address it in this PR but maybe that is something you can improve in your PR to use the Connector Form for the test input menu?

> The "add stream" modal doesn't really add any value, it's just making the user flow more complex - it seems like it would be easier to add an empty, untouched stream and switch to the view so the user can fill it in without any modal involved. I realize that this is not according to the current designs so happy to discuss it in some forum before implementing. Let me know if I should create an issue for it.

I think the benefits of using a modal are that simplifies the set of fields that the user is presented with upon creating a new stream to only the ones that are absolutely required (name and path), and doesn't allow them to create a stream unless they supply those values. If we just create an empty stream and focus that view, it may be a bit more overwhelming (especially once we add more components to the stream view). Though that may be lessened if we have collapsible components. 

@flash1293 could you create a ticket to capture this question? We can discuss it in the next low-code builder sync

flash1293:
> One reason to refetch the streams in this situation is if the user has implemented a stream in which the name or the URL use some interpolated user-input value. Then the name/url of that stream depend on the value they type into the test input menu, so they need to be refetched to resolve those interpolated references.

Makes sense, thanks

> I won't address it in this PR but maybe that is something you can improve in your PR to use the Connector Form for the test input menu?

Sounds good, I'm gonna take care of it.

> I think the benefits of using a modal are that simplifies the set of fields that the user is presented with upon creating a new stream to only the ones that are absolutely required (name and path), and doesn't allow them to create a stream unless they supply those values. If we just create an empty stream and focus that view, it may be a bit more overwhelming (especially once we add more components to the stream view). Though that may be lessened if we have collapsible components.

I see your point, I still think we can shape the general form in a way to make it more approachable without resorting to modals.

flash1293:
> I see your point, I still think we can shape the general form in a way to make it more approachable without resorting to modals.

Added it to the next sync

PR #20009, opened 2022-12-01
make our unit tests less brittle and not compare against messages out of our control

brianjlai:
## What
We had some tests in the connector builder server that asserted on very specific error messages. Something mightve changed upstream with additional quotes. This is breaking the build.

## How
I'm not 100% what changed slightly upstream, but at the end of the day we really shouldn't be having tests that look for specific text and especially when that text is dependent on an external package (in this case `airbyte-cdk`). We should just test what is mostly in our control like status code and part of the error text that our server is emitted.

### DIFF
--- a/airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
+++ b/airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
@@ -8,8 +8,6 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 from airbyte_cdk.models import AirbyteLogMessage, AirbyteMessage, AirbyteRecordMessage, Level, Type
-from fastapi import HTTPException
-
 from connector_builder.generated.models.http_request import HttpRequest
 from connector_builder.generated.models.http_response import HttpResponse
 from connector_builder.generated.models.stream_read import StreamRead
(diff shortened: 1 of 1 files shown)


airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
```diff
@@ -371,24 +368,20 @@ def test_read_stream_invalid_group_format():
...
         loop = asyncio.get_event_loop()
         with pytest.raises(HTTPException) as actual_exception:
-            loop.run_until_complete(
-                api.read_stream(StreamReadRequestBody(manifest=MANIFEST, config=CONFIG, stream="hashiras"))
-            )
+            loop.run_until_complete(api.read_stream(StreamReadRequestBody(manifest=MANIFEST, config=CONFIG, stream="hashiras")))
 
         assert actual_exception.value.status_code == 400
-        assert actual_exception.value.detail == "Could not perform read with with error: Every message grouping should have at least one request and response"
+        assert "Could not perform read with with error" in actual_exception.value.detail
```
brianjlai, line 374: we could even do away with checking any error text at all since text is often changing and error code tends to be more important

girarda: +1 to not testing the error message

alovew (approved):

girarda (approved):

PR #19275, opened 2022-11-10
Remove production use of DatabaseConfigPersistence

cgardens:
Depends on PR to add new db base test calss: https://github.com/airbytehq/airbyte/pull/19640

## What
* Removes `DatabaseConfigPersistence`!!! 

## How
* It does this by moving relevant database queries into the `ConfigRepository`
* Took this opportunity to start to split tests into their own classes by resource (e.g. workspace, actor definition). I did _not_ break resources into their own persistences. That's too much to do in one PR, though we should do that next. I did the tests, because otherwise I couldn't wade through what was going on safely enough.
* Removes test classes that were just testing `ConfigPersistence`. Moved any relevant business logic tests that were hidden in there (which there were a few, which is kinda scary because those tests were testing the old `ConfigPersistence` implementations and not the ones that are currently used) and moved them into appropriate classes.
* Even with all the code that is just being moved, we are still managing to drop nearly 2K lines of code. 😄 

## Future Work
* `ConfigRepositoryE2EReadWriteTest.java` - this class is a monster. Decided not to tackle it now, but as we split persistences and their tests by resource, we should pick at this. It basically has tests for CRUD operations for every single resource.
* Split out test suites for workspace, actor definitions, and health checks. It should make it relatively straight forward to write separate persistences for them as a follow up.

## Recommended reading order
1. `ConfigRepository.java`
2. `ActorDefinitionMigratorTest.java` -- this is the one where there were a lot of test that got move to it from ConfigPersistence. 
3. `WorkspacePeristenceTest.java`
4. `ActorDefinitionPersistenceTest.java`
5. `HealthCheckPersistenceTest.java`
6. the rest...

## 🚨 User Impact 🚨
This PR does move around a lot of the queries from the `ConfigPersistence` to the `ConfigRepository`. There is a lot of testing but it's not that methodical or perfect coverage. There is some risk here that there is some mistake in a query that isn't caught by testing.

PR #20011, opened 2022-12-02
bug: add workspace_id to query

cgardens:
## What
* the `workspaceQuery()` should take in a workspace id in order to service get workspace, but right now it does not.
* this is a bug introduced by https://github.com/airbytehq/airbyte/pull/19275

### DIFF
--- a/airbyte-config/config-persistence/src/test/java/io/airbyte/config/persistence/WorkspacePersistenceTest.java
+++ b/airbyte-config/config-persistence/src/test/java/io/airbyte/config/persistence/WorkspacePersistenceTest.java
@@ -42,6 +42,12 @@ class WorkspacePersistenceTest extends BaseConfigDatabaseTest {
   }
 
   @Test
+  void testGetWorkspace() throws ConfigNotFoundException, IOException, JsonValidationException {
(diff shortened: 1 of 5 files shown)


airbyte-config/config-persistence/src/test/java/io/airbyte/config/persistence/WorkspacePersistenceTest.java
```diff
@@ -41,6 +41,12 @@ void setup() {
         null));
   }
 
+  @Test
+  void testGetWorkspace() throws ConfigNotFoundException, IOException, JsonValidationException {
+    configRepository.writeStandardWorkspaceNoSecrets(createBaseStandardWorkspace().withWorkspaceId(UUID.randomUUID()));
```
pmossman, line 46: So before your bug fix, would this test fail because the returned workspace would have the `UUID.randomUUID()` instead of the `WORKSPACE_ID`? Just making sure I follow the logic of the test

pmossman (approved):
Code looks good to me, just had a question to make sure I follow the logic of the new test case that you added!

airbyte-config/config-persistence/src/test/java/io/airbyte/config/persistence/WorkspacePersistenceTest.java
cgardens: exactly.

PR #19984, opened 2022-12-01
Update connectors one of descriptions

flash1293:
## What

Related to https://github.com/airbytehq/airbyte/pull/19983

Updates the specs to use the new ability of listing options in the tooltip

## 🚨 User Impact 🚨

Should be released after the linked change is released as it makes the descriptions worse otherwise.

## Pre-merge Checklist
Expand the relevant checklist and delete the others.

<details><summary><strong>New Connector</strong></summary>

### Community member or Airbyter

- [ ] **Community member?** Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
- [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`
    - [ ] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - [ ] `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
    - [ ] `docs/integrations/README.md`
    - [ ] `airbyte-integrations/builds.md`
- [ ] PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests)

### Airbyter

If this is a community PR, the Airbyte engineer reviewing this PR is responsible for the below items.

- [ ] Create a non-forked branch based on this PR and test the below items on it
- [ ] Build is successful
- [ ] If new credentials are required for use in CI, add them to GSM. [Instructions](https://docs.airbyte.io/connector-development#using-credentials-in-ci).
- [ ] [`/test connector=connectors/<name>` command](https://docs.airbyte.io/connector-development#updating-an-existing-connector) is passing
[...]

PR #20015, opened 2022-12-02
🪟 🎉 Add options to tooltip of oneOf properties

### LINKED ISSUE #17975: Show descriptions of oneOf field choices in connector setup form

flash1293:
## What

Closes https://github.com/airbytehq/airbyte/issues/17975 by adding the options of condition nodes as a list to the tooltip of the form group.

Before: 
Screenshot 2022-12-01 at 16 15 02

After:
Screenshot 2022-12-02 at 09 32 32

## How

When rendering the `PropertyLabel` for a condition node, build a list of options with their descriptions (use title if available, otherwise fall back to key which is used for the dropdown as well) and pass it down to the `LabelInfo` component where it`s rendered similar to the description.

I debated a bit where to put the logic for turning the options of a condition into a list of descriptions - I settled on the `PropertyLabel` because it's the last component in the stack that still knows about condition properties. Putting it into `ConditionSection` would be closer to other code concerning conditions, but it makes it necessary to introduce a new type and pass it through a few layers which seems less clean eventually.

## 🚨 User Impact 🚨

This looks a little weird if the description of the group has been used to describe the options already (looks like this only affects a few connectors which I updated in a separate PR (https://github.com/airbytehq/airbyte/pull/19984), but it might be an issue for custom connectors, too): 
Screenshot 2022-12-02 at 09 34 53

### DIFF
--- a/airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.tsx
+++ b/airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.tsx
@@ -10,6 +10,7 @@ interface LabelInfoProps {
   label: React.ReactNode;
   examples?: JSONSchema7Type;
   description?: string;
+  options?: Array<{ title: string; description?: string }>;
 }
 
 const Description: React.FC<Pick<LabelInfoProps, "label" | "description">> = ({ label, description }) => {
@@ -22,6 +23,33 @@ const Description: React.FC<Pick<LabelInfoProps, "label" | "description">> = ({ 
       {/* don't use <Text as=h4> here, because we want the default parent styling for this header */}
       <h4 className={styles.descriptionHeader}>{label}</h4>
       <TextWithHTML className={styles.description} text={description} />
+    </div>
+  );
+};
+
+const Options: React.FC<Pick<LabelInfoProps, "options">> = ({ options }) => {
+  if (!options) {
+    return null;
+  }
+
+  return (
+    <div>
+      <h5 className={styles.optionsHeader}>
+        <FormattedMessage id="connector.optionsHeader" />
+      </h5>
+      <ul className={styles.options}>
+        {options.map((option) => (
+          <li key={option.title}>
+            <strong>{option.title}</strong>
+            {option.description && (
+              <>
+                {" "}
+                - <TextWithHTML className={styles.description} text={option.description} />
+              </>
+            )}
+          </li>
+        ))}
+      </ul>
     </div>
   );
 };
@@ -48,10 +76,11 @@ const Examples: React.FC<Pick<LabelInfoProps, "examples">> = ({ examples }) => {
   );
 };
 
-export const LabelInfo: React.FC<LabelInfoProps> = ({ label, examples, description }) => {
+export const LabelInfo: React.FC<LabelInfoProps> = ({ label, examples, description, options }) => {
   return (
     <div className={styles.container}>
       <Description label={label} description={description} />
+      <Options options={options} />
       <Examples examples={examples} />
     </div>
   );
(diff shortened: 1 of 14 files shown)


airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.tsx
```diff
@@ -20,12 +21,39 @@ const Description: React.FC<Pick<LabelInfoProps, "label" | "description">> = ({
...
+    return null;
+  }
+
+  return (
+    <div>
+      <h4 className={styles.optionsHeader}>
+        <FormattedMessage id="connector.optionsHeader" />
+      </h4>
+      <ul className={styles.options}>
+        {options.map((option) => (
```
lmossman, line 41: Do you think it would make sense to only render this options list if one of the options has a `description`?

I'm suggesting this due to cases like this:
Screen Shot 2022-12-02 at 10 37 37 AM
where none of the options have any descriptions, in which the `Options` list in the tooltip doesn't really provide any more info than clicking on the dropdown to see which options are available, so it feels somewhat redundant.

lmossman (approved):
Had one comment about rendering the option descriptions conditionally, but I don't feel it is a hard blocker. Let me know if you'd like me to take another look there based on your response.

Tested locally and it seems to work properly! Nice work

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/LabelInfo.tsx
lmossman: You're reasoning makes sense to me, I'm on board with keeping it! Thanks

PR #20019, opened 2022-12-02
Issue 19733 cdk clarify config error message for config files

maxi297:
## What
This is a duplicate of https://github.com/airbytehq/airbyte/pull/19931 but since it was opened from a fork, here is the correct pull request from the branch directly on airbyte repository. Some very relevant comments have already been added to https://github.com/airbytehq/airbyte/pull/19931 and should be fixed in this pull request.

This change address https://github.com/airbytehq/airbyte/issues/19733 which would cause the error message from reading config files to be unclear.

## How
As https://github.com/airbytehq/airbyte/issues/19733 states, `We should rename BaseConnector.read_config() to read_json_file() to accurately reflect what it is doing. The existing usages of read_config() should be updated.` and `The Source.read_state() method should start using this newly renamed read_json_file() to read the state JSON file instead of duplicating all the same file reading logic.`

## 🚨 User Impact 🚨
Previous error message: _Expecting value: line 1 column 1 (char 0)_
New error message: _Could not read json file secrets/config.json: `Expecting value: line 1 column 1 (char 0)`_

~~The solution proposed in https://github.com/airbytehq/airbyte/issues/19733 is a breaking change since the public method `read_config` is changed to `read_json_file`.~~ We have decided to avoid this breaking change since source-s3 re-implements this method and we therefore can't reuse it for parsing the state.

- [X] Unit & integration tests added and passing
- [ ] Code reviews completed

## Tests

</details>

</details>

</details>

## Other Notes
Running `mypy airbyte_cdk` (as mentioned in the README.md) before creating changes to the repository gave me the following output `Found 134 errors in 60 files (checked 140 source files)`. Is this expected?


brianjlai (approved):
changes look good! just a reminder to bump the version after pulling the latest master and there a formatting error in the CI run since `/sources/source.py` no longer uses the json.loads() method so we don't need to import it anymore

airbyte-cdk/python/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Changelog
 
-## 0.11.2
+## 0.12.2
+Revert breaking change on `read_config` while keeping the improvement on the error message
+
```
maxi297, line 5: There seems to be a changelog missing here https://github.com/airbytehq/airbyte/commit/23b35ba27c7eca2859373b2b2cc79bbaad21a943

airbyte-cdk/python/unit_tests/test_connector.py
```diff
@@ -75,6 +83,11 @@ def test_read_config(nonempty_file, integration: Connector, mock_config):
     assert mock_config == actual
 
 
+def test_read_non_json_config(nonjson_file, integration: Connector):
```
brianjlai, line 86: +1 to the test using the invalid config. But can we also add a test when we run `read_state()` where the incoming state comes in the form of a list to make sure the underlying `_read_json_file()` handles it correctly as expected

brianjlai (approved):
approving again! good work on the second iteration that makes it non-breaking but still let's us reuse the try/catch

airbyte-cdk/python/unit_tests/test_connector.py
maxi297: Is it what you were looking for @brianjlai ? https://github.com/airbytehq/airbyte/pull/20019/commits/67c55b8b1d2b0b6e0e16fae4f9f54dd224e83c19

brianjlai: ah yeah okay sounds like we already got tests. so don't mind me then

PR #20020, opened 2022-12-02
🪟 🐛 Fix tag input styles

### LINKED ISSUE #18692: Tag input in connector configuration no longer matches the styles of the rest of the inputs
## Environment
- **Airbyte version**:  0.40.17

## Current Behavior
Tag input no longer matches the style of the rest of the inputs. Some things I noticed:

* Placeholder text with "Select..."
* Focus state has thick border
* Background is not grey

Screen Shot 2022-10-31 at 10 06 37

Likely caused by #16027

## Expected Behavior
Tag input should match the style of the rest of the inputs

## Steps to Reproduce
1. Create a new Postgres source
2. Scroll down to the schemas component

flash1293:
## What
Fixes https://github.com/airbytehq/airbyte/issues/18692

Before:
Screen Shot 2022-10-31 at 10 06 37

After (regular, hovered, focused, error):
Screenshot 2022-12-02 at 15 30 15
Screenshot 2022-12-02 at 15 30 19
Screenshot 2022-12-02 at 15 30 23
Screenshot 2022-12-02 at 15 31 08

## How

Upgrading react-select made it necessary to overwrite a bunch of additional styles because react-select became more opinionated. This was necessary before, so the existing custom styles could easily be extended.

I couldn't figure out a native way to trigger the error state (I most likely missed it, but even if not we might want to add it at a later point in time), making it required by adding

```
          .required("you need this!")
```

to https://github.com/airbytehq/airbyte/blob/e3ef3fdbe192a3cb24ac235c62dd23f29260c5b9/airbyte-webapp/src/core/jsonSchema/schemaToYup.ts#L85

makes it easy to do so in a synthetic way.

## Drive-by changes

* Removed some any-s, seems like the react-select typings got better.
* Set aria-invalid on the react-select if it's in error state.

## 🚨 User Impact 🚨

Fixes the visual bug, no changes in behavior.


lmossman:
> I couldn't figure out a native way to trigger the error state

For the Postgres source, it is not showing the `Required` error state when empty because it is an optional field, so the yup `.required()` is not being applied here: https://github.com/airbytehq/airbyte/blob/f3d4830cf5d679e0a1540d5a10ab8764e8a5c566/airbyte-webapp/src/core/jsonSchema/schemaToYup.ts#L145-L147

We may have other connectors with string array fields that are required, which should be using this error state, but I don't know off-hand which ones

airbyte-webapp/src/components/ui/TagInput/TagInput.tsx
```diff
@@ -20,17 +19,29 @@ const customStyles = {
...
     color: `${styles.fontColor}`,
     fontWeight: 500,
   }),
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- same as above
-  multiValueRemove: (provided: any) => ({
+  multiValueRemove: (provided) => ({
     ...provided,
     borderRadius: `${styles.borderRadius}`,
   }),
+  control: (provided, state) => {
```
lmossman, line 31: One thing I noticed when testing locally was that when I hovered my cursor over the TagInput it stays as the `default` cursor, whereas when I hover over the other text input fields it changes to a `text` cursor. Could we also make this TagInput have a `text` cursor, so that it is consistent? (And I feel a text cursor also makes sense here because the user is still expected to type values into the field)

lmossman, line 31: Another thing I noticed while testing locally: this TagInput requires users to type in a value and then hit Enter for that value to be actually added to the field as a tag. If the users just types a value and then clicks away, the value they typed is erased.

I feel like this is not a very clear UX that the user has to press Enter in order for their changes to actually be saved, and it would be better if unfocusing the input also added their text as a tag. I think that behavior actually might have existed before we upgraded the react-select version.

Could you explore adding that functionality here? Also fine if you'd rather address that in a separate PR.

lmossman (commented):
Had a couple comments coming from testing locally. Otherwise the code changes look good

airbyte-webapp/src/components/ui/TagInput/TagInput.tsx
flash1293: good catch, fixed

flash1293: I agree, it can be very annoying to a user, especially as it works fine when navigating via keyboard. I didn't find a "supported" way, but I think I came up with an acceptable workaround.

flash1293:
@lmossman could you check again?

airbyte-webapp/src/components/ui/TagInput/TagInput.tsx
lmossman: @flash1293 one more small nit here: when hovering over the `X` close button of one of the tags, could we go back to the `default` cursor (or maybe `pointer`, since that seems to be what we use for buttons)? It currently uses the text cursor which looks a little odd.

We should also use the same pointer style for the whole-component clear `X` button on the right side

lmossman: Tried it out and I think this looks good! One question:
> This needs to be implemented outside of the onBlur prop of react-select because it's not default behavior.

What do you mean by "it's not default behavior"? Isn't the `handleDelete` method we are passing to the `onBlur` prop also not default behavior?

lmossman (approved):
Had one more small comment about cursor styling but otherwise LGTM!

Tested out locally and looks to be working well

flash1293:
> What do you mean by "it's not default behavior"? Isn't the handleDelete method we are passing to the onBlur prop also not default behavior?

I just mean that the `onBlur` prop of the component is not called when you are actually bluring, that's why I had to add it to the surrounding div and wait for the event to bubble up. Just a technical detail.

Issue #17155, opened 2022-09-26
[Frontend] Move connections components


edmundito:
Move all connection-related pages and views to `components/connection` and cleanup the internals based on what's described in the doc:

https://docs.google.com/document/d/1sz7M4aHqtq1leDwkNSN1Mufn2IBT-Ljlezb3LiS32lU/edit

octavia-squidington-iii:
cc @airbytehq/frontend

PR #20024, opened 2022-12-02
🪟 🧹  Move CatalogDiff modal from views to components

edmundito:
## What
Related to #17155

Moves CatalogDiff modal from views/ to components/ to follow the new naming convention.

## How
* Moves the files
* Removes components subfolder
* Clean up imports
* Moves hook to open modal to hooks/connection


krishnaglick (approved):
LGTM, build would fail if anything was missed

PR #20033, opened 2022-12-02
test for behavior when a column is removed in an incremental sync

davinchia:
## What
Adds a destination acceptance test for removing a column in an incremental sync.

The goal is to verify that it's safe to go ahead with column selection, particularly for some concerning destinations.


airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java
```diff
@@ -614,12 +618,65 @@ public void testIncrementalSync() throws Exception {
...
+
+    final String defaultSchema = getDefaultSchema(config);
+    List<AirbyteRecordMessage> actualMessages = retrieveNormalizedRecords(catalog,
+        defaultSchema);
+    assertSameMessages(messages, actualMessages, true);
+
+    // remove one field
+    final JsonNode jsonSchema = configuredCatalog.getStreams().get(0).getStream().getJsonSchema();
+    ((ObjectNode) jsonSchema.findValue("properties")).remove("HKD");
+    configuredCatalog.getStreams().get(0).getStream().setJsonSchema(jsonSchema);
```
edgao, line 657: nitpick: this isn't technically necessary, since you're modifying the ObjectNode in-place

airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java
```diff
@@ -561,6 +564,7 @@ public void specDBTValueShouldBeCorrect() throws WorkerException {
    * append records to the datastore instead of overwriting the previous run.
    */
   @Test
+  @Disabled
```
mfsiega-airbyte, line 567: Just a note - don't forget that these need to be un-disabled!

mfsiega-airbyte:
Quick summary: passing tests for:

1. BigQuery
2. Snowflake
3. Postgres
4. BigQuery Denormalized
5. Databricks
6. S3

No green test for AWS Datalake but from the test run, the actual new test case is passing and it appears that the failure is an unrelated test dating back to July, so I think we can ignore that one.

airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java
```diff
@@ -614,6 +613,73 @@ public void testIncrementalSync() throws Exception {
...
+  public void testIncrementalSyncWithNormalizationDropOneColumn(final String messagesFilename, final String catalogFilename)
+      throws Exception {
+    if (!normalizationFromSpec()) {
+      return;
+    }
+
+    final AirbyteCatalog catalog = Jsons.deserialize(MoreResources.readResource(catalogFilename),
+        AirbyteCatalog.class);
+
+    if (!catalog.getStreams().get(0).getName().equals("exchange_rate")) {
```
edgao, line 627: rather than using `@ParameterizedTest`, you can just reference `DataArgumentsProvider.EXCHANGE_RATE_CONFIG` directly

edgao (approved):
looks good! one comment about simplifying the test structure but :shipit: once that's resolved

airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java
davinchia: @mfsiega-airbyte what about this comment?

mfsiega-airbyte: @davinchia me wanting to but not getting a chance to spend the time to address this comment is why this PR isn't already merged :) it's not a trivial context switch, so I haven't found time.

edgao: I think this entire if statement can be removed now?

mfsiega-airbyte: I'll send a follow-up with that cleanup.

airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java
```diff
@@ -1462,7 +1529,6 @@ public String toString() {
    * your_containers_id" (ex. docker container attach 18cc929f44c8) to see the container's output
    */
   @Test
-  @Disabled
```
etsybaev, line 1465: Enabling this test may break some connectors and will keep lots of trash in DB. Basically, it wasn't even a test, but just a script for manual load emulation.

Issue #19670, opened 2022-11-21
Prepare update/delete definition endpoints for custom connectors in Cloud


xiaohansong:
In https://github.com/airbytehq/airbyte/issues/19637#issuecomment-1322485100, we captured notes from a discussion about these API endpoints between FE and BE.
> Synced offline about this. Our outcome and action items
> 
> * createCustomSource/DestinationDefinition should be used (with workspaceId)
> * Once in UI, we want to remove the old (non workspace scoped) create APIs
> * updateSource/DestinationDefinition (does not need workspaceId)
>   
>   * Cloud will check if the passed connector ID, is actually assigned to a workspace that the user has access to, if not throws 404 in cloud if connector ID
> * Same logic as for `update` should apply for the `delete` endpoint
> * No migration will be needed, we accept orphaned rows in the actor table for the rare cases users ran into that case

In order to achieve this, we'll perform the following steps:
- Remove the custom-specific blocks from config.yaml for update/delete
  - source_definitions/update_custom: https://github.com/airbytehq/airbyte/blob/master/airbyte-api/src/main/openapi/config.yaml#L472-L493
  - destination_definitions/update_custom: https://github.com/airbytehq/airbyte/blob/master/airbyte-api/src/main/openapi/config.yaml#L994-L1015
  - CustomSourceDefinitionUpdate: https://github.com/airbytehq/airbyte/blob/master/airbyte-api/src/main/openapi/config.yaml#L2683-L2692
  - CustomDestinationDefinitionUpdate: https://github.com/airbytehq/airbyte/blob/master/airbyte-api/src/main/openapi/config.yaml#L3053-L3062
  - source_definitions/delete_custom: https://github.com/airbytehq/airbyte/blob/master/airbyte-api/src/main/openapi/config.yaml#L494-L512
  - destination_definitions/delete_custom: https://github.com/airbytehq/airbyte/blob/master/airbyte-api/src/main/openapi/config.yaml#L1016-L1034
- Remove the custom-specific update/delete handler methods:
  - updateCustomSourceDefinition method in handler: https://github.com/airbytehq/airbyte/blob/master/airbyte-server/src/main/java/io/airbyte/server/handlers/SourceDefinitionsHandler.java#L283-L291
  - updateCustomDestinationDefinition method in handler: https://github.com/airbytehq/airbyte/blob/master/airbyte-server/src/main/java/io/airbyte/server/handlers/DestinationDefinitionsHandler.java#L284-L292
[...]

PR #20034, opened 2022-12-02
remove un-used update/delete custom connector api

xiaohansong:
## What
https://github.com/airbytehq/airbyte/issues/19670
https://github.com/airbytehq/airbyte-cloud/issues/3642

remove un-used update/delete custom connector api and create connector api


pmossman:
@marcelopio I know you mentioned working on CLI custom connectors in an issue comment earlier, can you verify that this PR is ok with your changes as well?

pmossman (approved):
Looks good to me, assuming builds pass!

xiaohansong:
@pmossman Took another pass and I think it should be straightforward to remove the creation in the same PR - PTAL if you got a chance!

And for @marcelopio's PR - I commented in your PR, the update path needs to be changed

marcelopio:
> @marcelopio I know you mentioned working on CLI custom connectors in an issue comment earlier, can you verify that this PR is ok with your changes as well?

It is ok for me. Actually this solves a doubt I was having if I should use custom or not. 

Now I just have the question if I should manage connectors version for non custom connectors, but that is to be discussed on my PR. Thanks!

pmossman:
@xiaohansong looks like there may be some broken tests in the build after removing those endpoints, maybe some tests need to be converted to the `createCustom` version of the endpoint?

josephkmh:
Just confirming that we are not using any of these endpoints on the frontend anymore 👍

PR #20036, opened 2022-12-02
Add Airbyte Protocol V1 support.

### LINKED ISSUE #16814: Create Airbyte Protocol v1
## What

Create the v1 of Airbyte Protocol.

## How

* Update documentation
* Clone `v0/airbyte_protocol.yaml` into `v1/airbyte_protocol.yaml`
* Make sure `airbyte_protocol.yaml` is a symlink to `v1/airbyte_protocol.yaml`

gosusnp:
## What

Enable Airbyte Protocol V1 in the platform.

Closes #16814 

## How

* Use the default objects as v1 and the v0 objects for v0 (as expected)
* Update the config to enable v0 and v1 in the platform

This leverages the fact that v1 doesn't change the structure of v0. However, we won't be fully using v1 until https://github.com/airbytehq/airbyte/pull/19240 is merged.
In order to properly use the v1 objects, we need to have a few tasks to decouple Java connectors from the default version of the protocol objects.

## Recommended reading order
Any

## 🚨 User Impact 🚨

Following this change, as expected but worth highlighting, we will be converting all messages v0->v1->v0 for all connectors until they have been migrated to v1.


airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  @Override
+  public Version getPreviousVersion() {
+    return new Version("0.0.0");
+  }
+
+  @Override
+  public Version getCurrentVersion() {
+    return new Version("1.0.0");
```
jdpgrailsdev, line 35: Does it make sense to define these versions somewhere in a constants class?  Something like PREVIOUS and CURRENT, as well as specific version numbers so that when we want to change, we can just basically update that class (e.g. point CURRENT to a newly added VERSION_X constant).  Feels like otherwise we are going to have a few versions of these strings floating around.  Not sure exactly how to reduce the duplication, but it feels like centralizing these would help.

gosusnp: Those values are tied to migration itself. Currently, we can add V2 that would have a previous v1, current v2.

We currently have https://github.com/airbytehq/airbyte/blob/4974855b415ea4812e0d6647efb367ad837d8733/airbyte-commons/src/main/java/io/airbyte/commons/version/AirbyteProtocolVersion.java#L7, I could add some constants `V0`, `V1` in there.

gosusnp: Updated

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
```diff
@@ -0,0 +1,39 @@
...
+import io.airbyte.commons.version.AirbyteProtocolVersion;
+import io.airbyte.commons.version.Version;
+import io.airbyte.protocol.models.v0.AirbyteMessage;
+import jakarta.inject.Singleton;
+
+/**
+ * Placeholder AirbyteMessage Migration from v0 to v1
+ */
+@Singleton
+public class AirbyteMessageMigrationV1 implements AirbyteMessageMigration<AirbyteMessage, io.airbyte.protocol.models.AirbyteMessage> {
```
edgao, line 17: is it intentional to upgrade to `io.airbyte.protocol.models.AirbyteMessage` rather than `io.airbyte.protocol.models.v1.AirbyteMessage`?

gosusnp: Yes, it is due to the fact that Java connectors and Platform share common libraries that are currently not versioning friendly.

In order to use `io.airbyte.protocol.models.v1.AirbyteMessage`, we need to update pretty much all the protocol objects used in the platform to `io.airbyte.protocol.models.v1.*` which forces us to currently bump some of the Java connectors to use `v1` objects as well. (see the slack thread from yesterday).

The idea here is to take a two step approach https://github.com/airbytehq/airbyte/issues/20037 and https://github.com/airbytehq/airbyte/issues/20038 rather than a giant change to bump platform to v1 and connectors to v0.

benmoriceau (commented):
LGTM, I'll let previous review to approve

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/AirbyteMessageMigrationV1.java
gosusnp: Yes, the idea here is to have a two step change, enabling v1 on the default objects allows us to have a smaller PR focused on the actual changes to enable v1.
I will have another PR to switch the objects to v1.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/VersionedProtocolSerializer.java
```diff
@@ -0,0 +1,32 @@
...
+  private final Version protocolVersion;
+
+  public VersionedProtocolSerializer(final ConfiguredAirbyteCatalogMigrator configuredAirbyteCatalogMigrator, final Version protocolVersion) {
+    this.configuredAirbyteCatalogMigrator = configuredAirbyteCatalogMigrator;
+    this.protocolVersion = protocolVersion;
+  }
+
+  @Override
+  public String serialize(final ConfiguredAirbyteCatalog configuredAirbyteCatalog) {
+    return Jsons.serialize(configuredAirbyteCatalogMigrator.downgrade(configuredAirbyteCatalog, protocolVersion));
```
grishick, line 29: Shy is `serialize` always calling `downgrade` and what happens if `protocolVersion` the latest version?

gosusnp: The Migrator figure out the migration path, if the version is current, nothing happens.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/ConfiguredAirbyteCatalogMigrationV1.java
```diff
@@ -0,0 +1,37 @@
...
+    implements ConfiguredAirbyteCatalogMigration<ConfiguredAirbyteCatalog, io.airbyte.protocol.models.ConfiguredAirbyteCatalog> {
+
+  @Override
+  public ConfiguredAirbyteCatalog downgrade(io.airbyte.protocol.models.ConfiguredAirbyteCatalog message) {
+    return Jsons.object(Jsons.jsonNode(message), ConfiguredAirbyteCatalog.class);
+  }
+
+  @Override
+  public io.airbyte.protocol.models.ConfiguredAirbyteCatalog upgrade(ConfiguredAirbyteCatalog message) {
+    return Jsons.object(Jsons.jsonNode(message), io.airbyte.protocol.models.ConfiguredAirbyteCatalog.class);
```
grishick, line 24: trying to understand what's going on here, is `io.airbyte.protocol.models.ConfiguredAirbyteCatalog` identical to the latest `io.airbyte.protocol.models.v1.ConfiguredAirbyteCatalog`? Should anyone be using `io.airbyte.protocol.models.v1.ConfiguredAirbyteCatalog` directly outside of migration or should people always be using `io.airbyte.protocol.models.ConfiguredAirbyteCatalog`?

gosusnp: This is currently a transition phase. Eventually, we should be using `io.airbyte.protocol.models.vX` and the objects without an explicit versions will be removed.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/VersionedProtocolSerializer.java
grishick: Is there a scenario under which `serialize` will have to upgrade rather than downgrade a catalog? E.g. if the connector supports only the newer version? And if that's the scenario - does `downgrade` then essentially upgrade instead of `downgrading`?

grishick (approved):

PR #20041, opened 2022-12-03
Add /v1/web_backend/check_updates

### LINKED ISSUE #19853: Create "Out of date definitions" endpoint
The frontend currently makes a request to source_definitions, destination_definitions, as well as their "latest" endpoint counterparts solely to determine if any definitions are out of date.

This is done on _every_ initial page load.

By creating an endpoint that handles all that logic on the backend will save us an immense amount of transfered bytes.

Frontend code triggering this lives [here](https://github.com/airbytehq/airbyte/blob/09d1c06cff2c5b385eef597d664a8a49db73200f/airbyte-webapp/src/views/layout/SideBar/NotificationIndicator.tsx#L15)

gosusnp:
## What
* Add a new endpoint to retrieve the number of updatable connectors.

Relates to #19853 

## How
* Add new endpoint.


airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendCheckUpdatesHandler.java
```diff
@@ -0,0 +1,101 @@
...
+    final Map<UUID, String> newActorDefToDockerImageTag;
+
+    try {
+      currentActorDefToDockerImageTag = configRepository.listStandardDestinationDefinitions(false)
+          .stream()
+          .map(def -> Map.entry(def.getDestinationDefinitionId(), def.getDockerImageTag()))
+          .toList();
+    } catch (final IOException e) {
+      log.error("Failed to get current list of standard destination definitions", e);
+      return 0;
```
jdpgrailsdev, line 49: Nit:  consider extracting to a constant to avoid re-use/magic number.

jdpgrailsdev (approved):
:shipit: with one small nit.

airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendCheckUpdatesHandler.java
gosusnp: done

airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendCheckUpdatesHandler.java
```diff
@@ -0,0 +1,103 @@
...
+    } catch (final InterruptedException e) {
+      log.error("Failed to get latest list of standard source definitions", e);
+      return NO_CHANGES_FOUND;
+    }
+
+    return getDiffCount(currentActorDefToDockerImageTag, newActorDefToDockerImageTag);
+  }
+
+  private int getDiffCount(final List<Entry<UUID, String>> initialSet, final Map<UUID, String> newSet) {
+    int diffCount = 0;
```
benmoriceau, line 93: nit:
```
initialSet.stream()
.filter(entry -> {
  final String newDockerImageTag = newSet.get(entry.getKey());
  return newDockerImageTag != null && !entry.getValue().equals(newDockerImageTag))
})
.count();
```

airbyte-server/src/test/java/io/airbyte/server/handlers/WebBackendCheckUpdatesHandlerTest.java
```diff
@@ -0,0 +1,190 @@
...
+class WebBackendCheckUpdatesHandlerTest {
+
+  ConfigRepository configRepository;
+  AirbyteGithubStore githubStore;
+  WebBackendCheckUpdatesHandler webBackendCheckUpdatesHandler;
+
+  final static boolean INCLUDE_TOMBSTONE = false;
+
+  @BeforeEach
+  void beforeEach() {
```
benmoriceau, line 33: Nit: this could be replace by the annotation @Mock and @InjectMock

benmoriceau (approved):
2 nit comments

krishnaglick (approved):
I did a test implementation on this branch and it's looking good! Seems to correctly update the count when the user upgrades 👍

airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendCheckUpdatesHandler.java
gosusnp: as discussed, using count sends us into type issues so we'll stick to old fashioned for loop count here.

airbyte-server/src/test/java/io/airbyte/server/handlers/WebBackendCheckUpdatesHandlerTest.java
gosusnp: Played with the annotations, I think a setup function feels clearer than having annotations on both members + class to figure out how the objects are initialized.
Might be related to the fact that I am not a big fan of the lombok type of annotations.

PR #20055, opened 2022-12-05
Add column.include.list for debezium column selection support

### LINKED ISSUE #19701: Use column filter for CDC sources in Debezium
We should use the [column.include.list](https://debezium.io/documentation/reference/stable/connectors/postgresql.html#postgresql-property-column-include-list) filter for CDC implementation in Debezium so that we only query for the columns that are present in the catalog.

rodireich:
## What
This change adds column selection to CDC sync in mysql, mssql and postgres sources.

## How
By Configuring the column.include.list during CDC sys startup and adding a configuration matching whatever is in the catalog, debezium will only sync the requested columns of tables.

In order to be able to support all allowed characters, I had to get all the escaping correctly.
Case in point: Postgres allows using any character in a name of a db, schema, table or column. This include also regex control characters that may mix up with debezium matching of table/columns.

## 🚨 User Impact 🚨
Anything that works today should work the same going forward.
When future UI supports selective configuration of tables columns in catalog, CDC sync should support it out of the box, similarly to standard sync.


rodireich:
The only characters that won't work as a name for a table or a column is a backslash `\`.
I couldn't find any way to escape it in a way that will then be successfully processed as column or table include list prop.
There's an issue in debezium I think.
I'm checking on their support forum

subodh1810 (commented):
Can we add a test where a table has multiple columns but we only pass in a subset of those columns in the catalog and assert that the data read only contains the columns that we passed (and this test should be for cdc (snapshot+incremental) and non-cdc) as well

akashkulk (approved):

Issue #17352, opened 2022-09-28
Consider default values for oneOf fields in connector specs


lmossman:
## What

See original [slack thread](https://airbytehq-team.slack.com/archives/C03088BTMFC/p1664320332231419) for context.

Currently our frontend code does not consider the `default` field if it is set on a oneOf property in the connector specification. We should update our code to use these default values if they are set when a user first creates a connector, and set the dropdowns to those default values automatically in that case.

## How

This would likely require changes in several places. Here is a draft PR with some progress started on these changes, though this still likely requires more testing and iteration: https://github.com/airbytehq/airbyte/pull/17351/files 

One open question: In the case of oneOfs of objects, does the `default` need to contain default values for every sub-field as well? Or is there some way in the json schema that we can just say `default: X` where X is one choice in the oneOf, without needing to also provide defaults for all of its children?

octavia-squidington-iv:
cc @airbytehq/frontend

sw-yx:
more broadly i wonder if we can use a jsonschema interpreter library that will at least ensure we are "up to spec" whatever that means

timroes:
We are using types for jsonschema in Typescript via a library (there's not much to parse since JSON schema is already plain JSON itself). This though won't solve the problem that we need to generate our UI and forms from that spec, which a library won't help us with.

flash1293:
In the context of https://github.com/airbytehq/airbyte/issues/14250 I would like to close this one because a connector developer can always achieve the same by reordering their oneOfs to put the first one on top.

This is a preferable solution because it makes the UI more consistent - if there is a oneOf, the default selection will always be the top one. Preselecting another one that's not the default could be slightly unexpected/confusing for a user.

Instead I would like to enforce no `default` values on `oneOf`properties via SAT (already added it to this issue: https://github.com/airbytehq/airbyte/issues/20063)

What do you think @lmossman ?

flash1293:
Discussed with @rodireich and we concluded that there are good use cases for this functionality. I checked through our existing specs and it turns out a bunch of them are already implementing `default` values. All of them specify the value of the shared const field, so I'm going to implement this behavior.

rodireich:
Thanks @flash1293 
This was in the context of #16664 from a few months back.
Once a `default` is implemented I can go back and improve they way our config UI behaves.

flash1293:
Approved, waiting for the code freeze

flash1293:
@rodireich This should work now

Issue #14250, opened 2022-06-29
(Resolve) Tech debt of the ConnectorForm


timroes:
Brainstomring issue about problem with the ServiceForm component. Please feel free to list all issues that should be addressed to make this component more stable and easier to maintain:

- [x] https://github.com/airbytehq/airbyte/issues/13056
- [x] Only render connector form component if the selected connector definition specification is loaded (move loading state and spinner into ConnectorCard) https://github.com/airbytehq/airbyte/pull/20127
- [x] Remove component override for name field and make it a normal field in the schema with title and description https://github.com/airbytehq/airbyte/pull/20132
- [x] Move unfinished flows into local state in ConfirmationControls (reset from the outside if formik state becomes non-dirty) https://github.com/airbytehq/airbyte/pull/20135
- [ ] Remove uiWidgetsInfo state by finding the enum/const field defining the selected oneOf and adding it to the condition form block to check the selected value by looking at the formik state
- [ ] Only do default value calculation once before rendering the formik component the first time (remove `PatchInitialValuesWithWidgetConfig` hack)
- [ ] Build yup schema once by flattening the oneOf conditions into a single object and adding `when` conditions on the found condition key (remove `RevalidateOnValidationSchemaChange` hack)
- [ ] Add https://github.com/airbytehq/airbyte/issues/17352 to default value calculation

* ~Pull out auth field hiding from `FormSection` to a higher level when generating from the spec which fields should be rendered.~
* ~`values.connectionConfiguration` plus flattened values are confusing (for a flattened form)~
* ~Rethink `default` behavior (should it be sending the value if the field is empty?), since it's working different depending on your field type atm~ -> extra issue

octavia-squidington-iii:
cc @airbytehq/frontend

flash1293:
Discussed getting rid of re-calculating the yup schema based on the selected conditions for `oneOf` properties, but as there is no straight forward way to encode the logic of "oneOf" using yup we decided against this route for now. This means the `RevalidateOnValidationSchemaChange` hack will need to stay for now.

flash1293:
current selection for `oneOf` does not need additional state because the documentation states that each oneOf has a const property which is the same for each of the conditions (also checked via SAT): https://docs.airbyte.com/connector-development/connector-specification-reference/#using-oneof

[...]

PR #20057, opened 2022-12-05
🪟🧹 Improve connector form types

flash1293:
## What
Improve some types in relation to https://github.com/airbytehq/airbyte/issues/17352

https://github.com/airbytehq/airbyte/issues/17352 itself can be tackled once https://github.com/airbytehq/airbyte/issues/14250 is done.

## How
* All `FormItem` types inherit some properties from the underlying JSON schema, and the types should reflect that (also we need the extra `default` for the linked related task)
* Remove custom `pick` implementation and use the lodash one to reduce the number of places where typing is lost in the intermediate processing (Object.fromEntries returns an object with almost no type information)


airbyte-webapp/src/core/form/types.ts
```diff
@@ -2,7 +2,16 @@ import { JSONSchema7Type, JSONSchema7TypeName } from "json-schema";
...
+/**
+ * When turning the JSON schema into `FormBlock`s,
+ * some often used props are copied over for easy access.
+ */
+type FormRelevantJSONSchema = Pick<
+  AirbyteJSONSchema,
+  "default" | "examples" | "description" | "pattern" | "order" | "const" | "title" | "airbyte_hidden" | "enum"
+>;
+
+interface FormItem extends FormRelevantJSONSchema {
```
lmossman, line 14: Is there a reason to keep `order`, `title`, `description`, and `airbyte_hidden` in this interface, now that they are coming from `FormRelevantJSONSchema`?

airbyte-webapp/src/core/jsonSchema/schemaToUiWidget.ts
```diff
@@ -119,12 +121,12 @@ const defaultFields: Array<keyof AirbyteJSONSchema> = [
...
   // airbyte specific fields
   "airbyte_hidden",
-];
+] as const;
 
-const pickDefaultFields = (schema: AirbyteJSONSchema): Partial<AirbyteJSONSchema> => {
-  const partialSchema: Partial<AirbyteJSONSchema> = {
-    ...Object.fromEntries(Object.entries(schema).filter(([k]) => defaultFields.includes(k as keyof AirbyteJSONSchema))),
-  };
+type DefaultFields = Pick<AirbyteJSONSchema, typeof defaultFields[number] | "enum">;
```
lmossman, line 126: A couple questions here:
1. What is `typeof defaultFields[number]` doing? It seems like this is just trying to pick all of the fields in the `defaultFields` array from `AirbyteJSONSchema`, but I don't know what that `number` in the brackets is being defined. Is this some functionality specific to `Pick`?
2. Why not just add `"enum"` to the `defaultFields` array? This way you wouldn't need to do ` | "enum"` unless I've missed something

lmossman (commented):
Had a couple questions

airbyte-webapp/src/core/form/types.ts
flash1293: There is not, good catch.

flash1293:
Based on your comments I simplified a bit further, let me know if there are other questions.

lmossman (approved):
Changes LGTM. Tested with connectors that have conditional fields and didn't notice any regression

Issue #20064, opened 2022-12-05
Improve schema checking in the UI


flash1293:
JSON schema can express a wide array of constraints on values. Not all of them are checked right in the UI even though they are straightforward to implement.

As the connector can produce error messages as part of the `check` implementation as well, it's not necessary to get to 100% here, but everything that can easily be handled should be handled by the UI.

List of checks to implement (can probably be extended):
* [ ] array minItems, maxItems (natively in yup)
* [ ] array uniqueItems (easy to implement manually)
* [ ] number exclusiveMinimum and exclusiveMaximum (natively in yup)
* [ ] number multipleOf (easy to implement manually)
* [ ] string minLength maxLength (natively in yup)
* [ ] string various formats (email, hostname, ip, uuid, uri, ...) (partially native in yup, regex for the others)

octavia-squidington-iv:
cc @airbytehq/frontend

PR #20068, opened 2022-12-05
🪟🐛 Fix type number in schema

### LINKED ISSUE #20062: type: "number" properties in spec do not work in the UI
## Current Behavior
When using `type: "number"` in the spec, the UI returns a string for this field rendering the spec invalid.

This can be observed for example in the e2e destination connector: https://github.com/airbytehq/airbyte/blob/938436bcc9bd05efb02b1937dd1b5ce02f0e6052/airbyte-integrations/connectors/destination-e2e-test/src/main/resources/spec.json#L41

## Expected Behavior
Should work by parsing the user input into a number like it happens for `type: "integer"` 

## Steps to Reproduce
1. Create E2E Testing destination
2. Change the "N" config
3. Check connection (fails with weird error message)

flash1293:
## What
Fixes https://github.com/airbytehq/airbyte/issues/20062

by treating type `number` like treating `integer`

## How
Just adding another case to the switch/case handling different types of form fields

### DIFF
--- a/airbyte-webapp/src/core/jsonSchema/schemaToYup.ts
+++ b/airbyte-webapp/src/core/jsonSchema/schemaToYup.ts
@@ -68,6 +68,7 @@ export const buildYupFormForJsonSchema = (
       schema = yup.boolean();
       break;
     case "integer":
+    case "number":
       schema = yup.number().transform((value) => (isNaN(value) ? undefined : value));
 
       if (jsonSchema?.minimum !== undefined) {
--- a/airbyte-webapp/src/core/jsonSchema/schemaToYup.test.ts
+++ b/airbyte-webapp/src/core/jsonSchema/schemaToYup.test.ts
@@ -73,6 +73,9 @@ it("should build schema for conditional case", () => {
         start_date: {
           type: "string",
         },
+        max_objects: {
+          type: "number",
+        },
         credentials: {
           type: "object",
           oneOf: [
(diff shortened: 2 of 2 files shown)


airbyte-webapp/src/core/jsonSchema/schemaToYup.ts
```diff
@@ -68,6 +68,7 @@ export const buildYupFormForJsonSchema = (
       schema = yup.boolean();
       break;
     case "integer":
```
lmossman, line 70: Slightly tangential to this PR, but it looks like this code is not actually validating that the `"integer"` case is actually an integer. It seems to just treat it as a number, so non-integer values are actually allowed to be set (I just confirmed this by testing it out locally:
Screen Shot 2022-12-05 at 11 38 48 AM

So we should likely update that case to add a `yup.integer()` validation

lmossman (approved):
Changes LGTM, tested locally and it seems to have fixed the bug.

Added a comment about handling `"integer"` types more strictly, but that can also be addressed in a separate PR

airbyte-webapp/src/core/jsonSchema/schemaToYup.ts
flash1293: I added it to this related issue I created - there are a bunch of these low-hanging fruits: https://github.com/airbytehq/airbyte/issues/20064

Issue #17828, opened 2022-10-11
Add auto-detect schema changes flow to connection replication settings page


edmundito:
When changes are auto-detected:

1. Show banner below connection widget
2. Block the streams table
3. User clicks on review changes to look at diffs
4. User can save or cancel the changes

Prototype - Click on "Slack Community" to see the flow:

https://www.figma.com/proto/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?node-id=972%3A37260&scaling=min-zoom&page-id=972%3A36208&starting-point-node-id=972%3A36212&show-proto-sidebar=1

octavia-squidington-iv:
cc @airbytehq/frontend

PR #20070, opened 2022-12-05
🪟 🎨 Auto-detect schema changes UI fixes

edmundito:
## What
Related to #17828 

Fixes small issues in the auto-detect schema change UI:

* Makes "Review changes" button variant dark to match the design
* Adds matching hover state to source link when there is a breaking or non-breaking change
* Removes periods from breaking/non-breaking messages


teallarson (approved):
Code lgtm.  Tested changes locally.

PR #20072, opened 2022-12-05
🪟🐛 Warn on navigating away from unsaved connector form with error

### LINKED ISSUE #18246: User won't get "Unsaved changes" modal after moving to another page if testing connector has failed
## Environment
- **Airbyte version**:  0.40.15
- **OS Version / Instance**: all
- **Deployment**: all
- **Step where error happened**: Source / Create new source / Set up source

## Steps to repro:
1. Go to the "Sources" tab
2. Click the "New source" button
3. Choose "Poke API" in the dropdown
4. Fill in the `Pokemon Name` field with a random string
5. Click the "Set up source" button  - Testing connector should fail
6. On the left sidebar click on the "Destinations" tab

## Current Behavior
The "Destinations" page is opened

## Expected Behavior
The "Unsaved changes" modal appears

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/18246 

When navigating away from an unsaved connector form with an error, the warning is shown.

## How
Make sure exceptions are thrown up the promise chain from the actual react-query call to the submit callback so the connector form is not clearing the dirty state of the form change tracker.

### DIFF
--- a/airbyte-webapp/src/views/Connector/ConnectorForm/ConnectorForm.tsx
+++ b/airbyte-webapp/src/views/Connector/ConnectorForm/ConnectorForm.tsx
@@ -89,7 +89,7 @@ export interface ConnectorFormProps {
   formId?: string;
   selectedConnectorDefinition?: ConnectorDefinition;
   selectedConnectorDefinitionSpecification?: ConnectorDefinitionSpecification;
-  onSubmit: (values: ConnectorFormValues) => Promise<void> | void;
+  onSubmit: (values: ConnectorFormValues) => Promise<void>;
   isLoading?: boolean;
   isEditMode?: boolean;
   formValues?: Partial<ConnectorFormValues>;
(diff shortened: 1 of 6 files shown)


airbyte-webapp/src/views/Connector/ConnectorForm/ConnectorForm.tsx
```diff
@@ -89,7 +89,7 @@ export interface ConnectorFormProps {
   formId?: string;
   selectedConnectorDefinition?: ConnectorDefinition;
   selectedConnectorDefinitionSpecification?: ConnectorDefinitionSpecification;
-  onSubmit: (values: ConnectorFormValues) => Promise<void> | void;
```
lmossman, line 92: This looks much cleaner now 🙌

lmossman (approved):
Changes LGTM. Tested locally and bug looks to be resolved!

PR #20073, opened 2022-12-05
Add test for checking stream view

SofiiaZaitseva:
In scope of PR was added test for checking stream view (test case number c_19)
https://docs.google.com/spreadsheets/d/1OMKYYoTAFeCAk_ZXUlFmiKpfnt7x5edRGxRQB4XMyQM


airbyte-webapp-e2e-tests/cypress/integration/connection.spec.ts
```diff
@@ -541,3 +361,212 @@ describe("Connection main actions", () => {
...
+  });
+
+  it("Stream view", () => {
+    const sourceName = appendRandomString("Test connection Postgres source cypress");
+    const destName = appendRandomString("Test connection Postgres destination cypress");
+
+    const collectionNames = new Array();
+    collectionNames.push("Field name", "col1", "id");
+    const collectionTypes = new Array();
+    collectionTypes.push("Data type", "String", "Integer");
```
dizel852, line 555: Can be simplified
```suggestion
    const collectionNames = ["Field name", "col1", "id"];
    const collectionTypes = ["Data type", "String", "Integer"];
```

airbyte-webapp-e2e-tests/cypress/pages/replicationPage.ts
```diff
@@ -9,6 +9,9 @@ const destinationNamespaceSource = "div[data-testid='namespaceDefinition-source'
 const destinationNamespaceCustomInput = "input[data-testid='input']";
 const syncModeDropdown = "div[data-testid='syncSettingsDropdown'] input";
 const cursorFieldDropdown = "button[class^='PathPopoutButton_button']";
+const streamFieldNames = "[class^='TreeRowWrapper_rowWrapper'] span";
+const streamDataTypes = "[class^='TreeRowWrapper_rowWrapper'] div:nth-child(2)";
+const arrowStream = "[class^='Arrow_container__']";
```
dizel852, line 14: what is an arrow stream? 🙂

dizel852 (changes requested):
Overall - happy to see more tests 🥳 
Left some comments.
Also, I'd like to ask to change test suites structure to smth like this one:
image

airbyte-webapp-e2e-tests/cypress/pages/replicationPage.ts
SofiiaZaitseva: Screenshot 2022-12-06 at 21 59 24
button for open stream details

dizel852 (changes requested):
Did another review - LGTM 👍 
Left one small comment. 
Would be great to fix it.

dizel852 (approved):
LGTM
Passing to @edmundito for review

airbyte-webapp-e2e-tests/cypress/pages/replicationPage.ts
```diff
@@ -9,6 +9,9 @@ const destinationNamespaceSource = "div[data-testid='namespaceDefinition-source'
 const destinationNamespaceCustomInput = "input[data-testid='input']";
 const syncModeDropdown = "div[data-testid='syncSettingsDropdown'] input";
 const cursorFieldDropdown = "button[class^='PathPopoutButton_button']";
+const streamFieldNames = "[class^='TreeRowWrapper_rowWrapper'] span";
+const streamDataTypes = "[class^='TreeRowWrapper_rowWrapper'] div:nth-child(2)";
```
edmundito, line 13: These selectors work for now, but ideally, we should try to use testIds instead because the implementation can always change and break this. For this PR, it's OK to leave because we also have the problem in other areas.

edmundito (approved):

PR #20077, opened 2022-12-05
Fail on internal errors

benmoriceau:
## What
If a runtime exception happens in an activity launched by the sync workflow, we should fail the workflow instead of making it stuck.

### DIFF
--- a/airbyte-workers/src/main/java/io/airbyte/workers/ApplicationInitializer.java
+++ b/airbyte-workers/src/main/java/io/airbyte/workers/ApplicationInitializer.java
@@ -317,7 +317,7 @@ public class ApplicationInitializer implements ApplicationEventListener<ServiceR
       final Worker syncWorker = factory.newWorker(taskQueue,
           getWorkerOptions(maxWorkersConfig.getMaxSyncWorkers()));
       final WorkflowImplementationOptions options = WorkflowImplementationOptions.newBuilder()
-          .setFailWorkflowExceptionTypes(NonDeterministicException.class).build();
+          .setFailWorkflowExceptionTypes(NonDeterministicException.class, RuntimeException.class).build();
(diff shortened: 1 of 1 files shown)


jdpgrailsdev (approved):
:shipit:

airbyte-workers/src/main/java/io/airbyte/workers/ApplicationInitializer.java
```diff
@@ -317,7 +317,7 @@ private void registerSync(final WorkerFactory factory, final MaxWorkersConfig ma
       final Worker syncWorker = factory.newWorker(taskQueue,
           getWorkerOptions(maxWorkersConfig.getMaxSyncWorkers()));
       final WorkflowImplementationOptions options = WorkflowImplementationOptions.newBuilder()
-          .setFailWorkflowExceptionTypes(NonDeterministicException.class).build();
+          .setFailWorkflowExceptionTypes(NonDeterministicException.class, RuntimeException.class).build();
```
gosusnp, line 320: Any reason to not just fail any exceptions?

gosusnp (approved):

airbyte-workers/src/main/java/io/airbyte/workers/ApplicationInitializer.java
pmossman: Found a thread about this, seems like Temporal generally recommends against this because typically, workflow failures are something to be avoided: https://community.temporal.io/t/throwing-exception-vs-failure-in-workflow/1059

But Airbyte tends to prefer explicit workflow failures instead of retries, so I wonder if we have a good reason to go against their recommendation?

pmossman:
@benmoriceau I am going through old PRs, should this be merged or is it obsolete by now?

benmoriceau:
@pmossman We decided to not doing it, I closed the PR

PR #20082, opened 2022-12-05
Add README info about using your local CDK locally and in docker

erohmensing:
Make it easier to test new CDK features by giving instructions on how to install the local CDK into a connector.


airbyte-cdk/python/README.md
```diff
@@ -70,6 +70,61 @@ pip install -e ".[dev]" # [dev] installs development-only dependencies
...
+}
+```
+
+You should be able to build your connector with
+```bash
+# from the airbytehq/airbyte base directory
+./gradlew build :airbyte-integrations:connectors:<connector-directory>
+```
+and the installation should use your local CDK. Note that the local CDK is injected at build time, so if you make changes, you will have to run the build command again to see them reflected.
+
```
sherifnada, line 127:
```suggestion
**Note:** if your connector uses a `.dockerignore` file, it cannot have exclude-all or exclude-except patterns i.e: the `.dockerignore` must specifically say which files to ignore without using any regex. 
```

sherifnada (approved):
Awesome! Thanks Ella

airbyte-cdk/python/README.md
erohmensing: Thanks for catching this!

PR #20095, opened 2022-12-05
🪟 🔧 Add testid to error view

timroes:
## What

Adds a `data-testid` to the "Oops" error view. This allows us to filter on it's appearance in Fullstory to better investigate those errors.


krishnaglick (approved):
Code LGTM, did not test locally.

PR #20097, opened 2022-12-05
Added stylelintignore for generated coverage files.

krishnaglick:
Added stylelintignore for generated coverage files.

### DIFF
--- a/airbyte-webapp/.stylelintignore
+++ b/airbyte-webapp/.stylelintignore
@@ -1,0 +1,1 @@
+coverage/**/*.css
--- a/airbyte-webapp/.stylelintignore
+++ b/airbyte-webapp/.stylelintignore
@@ -1,1 +1,1 @@
-coverage/**/*.css
+coverage/**/*


teallarson (approved):
confirmed that this does fix the local OSS build

airbyte-webapp/.stylelintignore
```diff
@@ -0,0 +1 @@
+coverage/**/*.css
```
ambirdsall, line 1: nit of all nits, but could you put a final newline in this file? `.editorconfig` is agnostic except for python files, where `insert_final_newline = true` is explicitly set, but our files nearly universally have them. Perhaps most importantly, though, is that I think github's big red "no final newline here" icon is kind of ugly :shrug: :sweat_smile:

ambirdsall: Is there a reason to scope this specifically to `*.css` files? IIUC, everything in `coverage/` is autogenerated, so why not ignore `coverage/**/*`?

ambirdsall (approved):
Minor suggestions, but this improves our setup with no apparent downsides :raised_hands: :sparkles: :rocket: :guitar:

airbyte-webapp/.stylelintignore
krishnaglick: I used the inline editor, I think it got the newline

ambirdsall (approved):
:sparkles: :guitar: :shipit: :guitar: :sparkles:

PR #20105, opened 2022-12-05
Add dbt cloud integration to default cloud features

ambirdsall:
For the initial rollout, we chose to enable dbt Cloud integration via a feature override in LaunchDarkly so we could roll back immediately if something went awry. This enables the feature in the cloud build by default, so a LaunchDarkly outage would no longer prevent people from managing synced dbt jobs.


timroes (approved):
Code LGTM. Have not tested locally.

PR #20123, opened 2022-12-06
🪟 🔧 Add comment about feature/experiment usage

joeykmh:
## What
Adds some clarifying comments about `FeatureItem` vs. `Experiment`

### DIFF
--- a/airbyte-webapp/src/hooks/services/Experiment/experiments.ts
+++ b/airbyte-webapp/src/hooks/services/Experiment/experiments.ts
@@ -1,5 +1,8 @@
 /* eslint sort-keys: "error" */
 /**
+ * Experiments are short-term flags for A/B testing or staged rollouts of features.
+ * They are only available in Airbyte Cloud.
+ *
  * When adding a new feature flag in LaunchDarkly to consume in code you'll need to make
  * sure to update the typing here.
  */
--- a/airbyte-webapp/src/hooks/services/Experiment/experiments.ts
+++ b/airbyte-webapp/src/hooks/services/Experiment/experiments.ts
@@ -1,7 +1,6 @@
 /* eslint sort-keys: "error" */
 /**
  * Experiments are short-term flags for A/B testing or staged rollouts of features.
- * They are only available in Airbyte Cloud.
  *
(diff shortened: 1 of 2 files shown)


airbyte-webapp/src/hooks/services/Experiment/experiments.ts
```diff
@@ -1,5 +1,8 @@
 /* eslint sort-keys: "error" */
 /**
+ * Experiments are short-term flags for A/B testing or staged rollouts of features.
+ * They are only available in Airbyte Cloud.
```
timroes, line 4: I find this part of the comment a bit confusing in this file. Since the `useExperiment` hook has a default value which will be used in OSS always, I feel this might confuse new developers to read this as: "don't use experiments in OSS code", which isn't the case.

josephkmh: Makes sense, I just removed that part since how the default/fallback experiment value works is not actually relevant to the interface here.

airbyte-webapp/src/hooks/services/Feature/types.tsx
```diff
@@ -1,3 +1,8 @@
+/**
+ * FeatureItems are for permanent flags to differentiate features between environments (e.g. Cloud vs. OSS),
+ * workspaces, specific user groups, etc. The corresponding flag should be added to LaunchDarkly.
```
timroes, line 3: > The corresponding flag should be added to LaunchDarkly.

Not entirely sure we should put that in here? For features we currently don't have that PR that makes each feture an individual flag (https://github.com/airbytehq/airbyte/pull/19773) merged yet.

Did you mean to put this on top of the experiments.ts file, where we will need a LD flag for everything?

timroes (approved):

airbyte-webapp/src/hooks/services/Feature/types.tsx
josephkmh: I've just removed it now. I guess it's not necessary to be added to LD (only if you're overriding a default `FeatureItem`), so the explanation about how LD works with `FeatureService` is probably better off in the Notion docs.

PR #20126, opened 2022-12-06
🪟 🎉 Change default sync mode

### LINKED ISSUE #15818: Discuss changing default sync mode
We currently set a default sync mode as we discussed in April which is documented in https://github.com/airbytehq/airbyte/issues/9625#issuecomment-1105320983. We've now seen and heard from user feedback, that what we outlined that time, might not be the ideal solution. Thus I wanted to discuss two potential changes to the default sync mode we're selecting for new connections:

#### Change 1

We currently prefer using "incremental, append_dedupe" if the source has a source defined cursor. We though don't check if it has a source defined primary key. Thus this can become the default mode and users will still require to select primary keys for those streams manually. Since our original intend was to make the setup as low friction as possible, I'd suggest that we only select this as default, if the source has a source defined cursor AND a source defined primary key.

Pros:
- Less friction with the setup, i.e. if we select a sync mode all fields will already be selected, and we'd not force the user to have any manual field selection

Cons:
- We give the user less often the hint that a stream could be incremental (and they'd "just" need to select a primary key to enable it)

_My opinion:_ I think we should make sure we only select incremental, append_dedupe if we have all fields by the source defined so the user doesn't need to configure anything manually.

#### Change 2

We currently prefer "full_refresh, overwrite" over "incremental, append", and since full_refresh, overwrite should be supported by all connections, "incremental, append" will effectively never be selected. We could select this over "full_refresh, overwrite" if we have a source defined cursor.

Pros:
- We'd give users easier incremental data (e.g. the google analytics source would atm fall in this category).

Cons:
- `incremental, append` has a change of creating duplicates (without dedupe). This might be a dangerous default behavior, since it might not entirely clear to the user, that our default setting might leave them with duplicate data.

@andyjih @sherifnada @nataliekwong Would be happy for your input on that.

timroes:
## What

Closes https://github.com/airbytehq/airbyte/issues/15818 (Change 1 in this issue, since we did not agree to make Change 2 suggested)

This will make sure we're no longer choosing "Incremental | Append Dedupe" if we don't have a source defined primary key. The idea is, that we should try to only change default values that won't require the user to select anything by themselves (which will also be a requirement to get to a more simplified form eventually).

### DIFF
--- a/airbyte-webapp/src/views/Connection/ConnectionForm/calculateInitialCatalog.test.ts
+++ b/airbyte-webapp/src/views/Connection/ConnectionForm/calculateInitialCatalog.test.ts
@@ -79,11 +79,8 @@ describe("calculateInitialCatalog", () => {
               name: "test",
               sourceDefinedCursor: true,
               defaultCursorField: ["id"],
-              supportedSyncModes: [SyncMode.incremental],
-              sourceDefinedPrimaryKey: [
-                ["primary", "field1"],
-                ["primary", "field2"],
-              ],
+              supportedSyncModes: [SyncMode.full_refresh, SyncMode.incremental],
+              sourceDefinedPrimaryKey: undefined,
             },
             config: {
               ...config,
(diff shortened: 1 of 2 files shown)


krishnaglick (approved):
Have confirmed locally this works as expected.
Faker -> Postgres no longer does incremental+dedup, but mailchimp -> postgres continues to do so.

Issue #17674, opened 2022-10-06
Generate config menu modal from yaml connector spec block


lmossman:
## Tell us about the problem you're trying to solve
As described in the [Tech Spec](https://docs.google.com/document/d/11HrieUnA7oa6YsDhOVZpAVkOoZt2RWQQWJLPf7xQodY/edit#heading=h.x4dhu0sbkibe), something that is needed for the testing of a connector is a place to enter the "config" values, i.e. fill out the user inputs required to run the connector, as described in the connector spec.

## Describe the solution you’d like
The frontend should pull the connector specification out of the yaml editor contents, and use that schema to generate an input form, similar to how connector forms are generated from connector specifications today (in the ConnectorForm component)

Many parts of the ConnectorForm can likely be reused for the purpose, but there will undoubtedly be some changes required to port it over to this other part of the application, which is where most of the complexity in this issue will be.

## Acceptance Criteria
- A button exists that opens the Config menu in a modal, which contains a form generated from the connector specification
- When the connector specification is modified, the Config menu form is also updated
- Changes to the config menu form update the connector builder state that is keeping track of the config

lmossman:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Connector-Extensibility-6257455decf3920012f4e872/issues/airbytehq/airbyte/17674?planning-poker) with Zenhub @girarda

flash1293:
PR is ready to review, but stacked on top of the refactoring PR

PR #20127, opened 2022-12-06
🪟🔧 Move connector loading state into connector card

flash1293:

## What
This PR moves the loading spinner which is shown when the connection specification definition is loaded into the connector card component. This is done for a bunch of reasons:
* Simplifying the connector form code - it's now only rendered if the specification is available
* Remove complexity by not passing the loading prop deep into the component tree
* Preparing the component to be used in other places as well

## How
In the current state, the formik form is always initialized and rendered, even if there is no specification. The form component is also responsible for rendering the loading spinner itself.

This PR lifts the loading spinner to the connector card level and only renders the form component if all the prerequisites are present.

## Drive-by fixes

* The loading spinner requires the name of the connector that's currently loaded. This doesn't work correctly in the current state as the loading spinner pulls the name from the specification definition which isn't loaded yet. This PR changes this by passing the chosen connector id down to the connector card even if it's not loaded yet so the name can be shown in the loading spinner
* In case the loading of the specification definition fails, the current state is not showing any error message because the component handling this (EditControls/CreateControls) is only showing the error if the specification definition is loaded. This PR pulls this error case to the connector card level and doesn't even attempt to render the form in this situation
* There were unused props `errorMessage` and `successMessage` on the connector card level, those were removed

## 🚨 User Impact 🚨
Fixes a bunch of small bugs but otherwise no user-visible change.


airbyte-webapp/src/views/Connector/ConnectorCard/ConnectorCard.tsx
```diff
@@ -38,14 +41,16 @@ interface ConnectorCardBaseProps {
 
   // used in ConnectorCard and ConnectorForm
   formType: "source" | "destination";
+  /**
+   * id of the selected connector definition id - might be available even if the specification is not loaded yet
+   * */
+  selectedConnectorDefinitionId?: string | null;
```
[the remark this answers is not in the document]
flash1293: I tried to find a good cutoff point to stop with the cleanup. The null is coming from here: https://github.com/airbytehq/airbyte/blob/7d6f9770503d1d6f34697513529d0fa13e0040a3/airbyte-webapp/src/components/destination/DestinationForm/DestinationForm.tsx#L46

I thought it would be OK for now because the `jobId` is doing the same thing, but if you feel strongly about it I can  change it.

krishnaglick (commented):
I haven't done a deep dive into the code but I feel like there could be an argument for adding a context to hold some of these reused and prop-drilled values. Don't hold this up on my account, merely an observation.

airbyte-webapp/src/views/Connector/ConnectorForm/FormRoot.tsx
```diff
@@ -1,25 +1,20 @@
...
 
 interface FormRootProps {
   formFields: FormBlock;
   hasSuccess?: boolean;
   isTestConnectionInProgress?: boolean;
   errorMessage?: React.ReactNode;
-  fetchingConnectorError?: Error | null;
   successMessage?: React.ReactNode;
   onRetest?: () => void;
   onStopTestingConnector?: () => void;
```
lmossman, line 20: I can't comment on it but I think the `| undefined` on `selectedConnector` prop type (the line below this one) can now be removed, because it is now being passed in from the ConnectorForm in a branch where it is verified to not be undefined. 

This should also allow us to remove the `selectedConnector && ()` around the `<CreateControls>` component near the bottom of this file

airbyte-webapp/src/views/Connector/ConnectorForm/ConnectorForm.tsx
```diff
@@ -87,14 +87,12 @@ const RevalidateOnValidationSchemaChange: React.FC<{ validationSchema: unknown }
 export interface ConnectorFormProps {
   formType: "source" | "destination";
   formId?: string;
-  selectedConnectorDefinition?: ConnectorDefinition;
-  selectedConnectorDefinitionSpecification?: ConnectorDefinitionSpecification;
+  selectedConnectorDefinition: ConnectorDefinition;
+  selectedConnectorDefinitionSpecification: ConnectorDefinitionSpecification;
```
lmossman, line 91: A side note that doesn't need to be addressed in this PR: when we eventually want to use the component in the Connector Builder (https://github.com/airbytehq/airbyte/issues/17674), there won't be any concept of "connector definitions" to pass to that component. Instead we will just have a JSON schema object that we need to pass in and get back a rendered connector form (or something similar). So it would be great if this component could be generified in that way

airbyte-webapp/src/views/Connector/ConnectorCard/ConnectorCard.tsx
```diff
@@ -175,25 +183,34 @@ export const ConnectorCard: React.FC<ConnectorCardCreateProps | ConnectorCardEdi
...
+          {props.isLoading && (
+            <div className={styles.loaderContainer}>
+              <Spinner />
+              <div className={styles.loadingMessage}>
+                <ShowLoadingMessage connector={selectedConnectorDefinition?.name} />
+              </div>
+            </div>
+          )}
+          {fetchingConnectorError && <FetchingConnectorError />}
+          {selectedConnectorDefinition && selectedConnectorDefinitionSpecification && (
```
lmossman, line 195: Is it possible that one of these is nullish even though a user has clicked on a connector from the dropdown? If so we may want to show some sort of error message here indicating that the connector could not be loaded or something. Because based on this code it still seems possible for `fetchingConnectorError` to be nullish, _and_ for one of these values to be nullish as well, in which case we just won't render anything here which feels like a not great UX

airbyte-webapp/src/views/Connector/ConnectorCard/ConnectorCard.tsx
```diff
@@ -175,25 +183,34 @@ export const ConnectorCard: React.FC<ConnectorCardCreateProps | ConnectorCardEdi
...
+            <ConnectorForm
```
lmossman, line 196: Also I'm getting a little confused here: your new code makes sense to me in that it only renders a `<ConnectorForm>` if `selectedConnectorDefinition` and `selectedConnectorDefinitionSpecification` are not nullish, but I'm confused how the old code was handling that case -- it seems like the old code was always rendering a ConnectorForm which was always rendering a FormRoot which was always rendering a FormSection, so how does the UI currently look empty when no connector is selected?

airbyte-webapp/src/views/Connector/ConnectorForm/FormRoot.tsx
flash1293: Good catch, removed that stuff as well!

airbyte-webapp/src/views/Connector/ConnectorForm/ConnectorForm.tsx
flash1293: Yeah, I'm not sure yet at which level we can lift out the functionality required for the connector builder, I guess we need to move a few things around. I will go through with the rest of the cleanup and take a look then.

airbyte-webapp/src/views/Connector/ConnectorCard/ConnectorCard.tsx
flash1293: I don't think there is a way to run into this situation besides the desired one when the user navigates to a completely empty form.

If you look at the code where the component is used in `SourceForm` and `DestinationForm`, the props `isLoading`,`fetchingConnectorError` and `selectedConnectorDefinition` are coming from a react-query hook, so one of them will always be set if there is a selection that has been made.

flash1293:
Thanks for the review @lmossman , could you check again? Removed the additional bit you found.

lmossman (approved):
Changes LGTM. Tested locally and seems to work as expected

PR #20132, opened 2022-12-06
🪟🔧 Remove connector form name override

flash1293:

Based on https://github.com/airbytehq/airbyte/pull/20127 (wait until merged)

## What
This PR removes the special component for the connector name and replaces it with regular schema definition. It also moves this logic into the `useBuildForm` hook.

## How
As the regular form builder can handle name fields just fine, it's simple to do the name and description resolution in the schema modification instead of handling this via a special component. This also allows to remove the "uiOverride" concept from the uiWidget state. In a follow-up PR this will be removed completely.

## 🚨 User Impact 🚨
No user-visible changes


airbyte-webapp/src/views/Connector/ConnectorForm/useBuildForm.tsx
```diff
@@ -13,21 +14,35 @@ import { buildYupFormForJsonSchema } from "core/jsonSchema/schemaToYup";
...
+          title: formatMessage({ id: `form.${formType}Name` }),
+          description: formatMessage({ id: `form.${formType}Name.message` }),
+        },
+        connectionConfiguration:
+          selectedConnectorDefinitionSpecification.connectionSpecification as JSONSchema7Definition,
+      },
+      required: ["name"],
+    }),
+    [formType, formatMessage, selectedConnectorDefinitionSpecification.connectionSpecification]
+  );
```
lmossman, line 45: One potential future issue with putting this in the `useBuildForm` hook is that when we want to use the connector form in the Connector Builder for the config menu, we don't want to have this "connector name" field in there.

Can definitely be addressed in a later PR but I wanted to call it out since I thought of it

airbyte-webapp/src/views/Connector/ConnectorForm/useBuildForm.tsx
```diff
@@ -54,19 +70,18 @@ interface BuildUiWidgetsContextHook {
...
-  const [overriddenWidgetState, setUiWidgetsInfo] = useState<WidgetConfigMap>(uiOverrides ?? {});
+  const [overriddenWidgetState, setUiWidgetsInfo] = useState<WidgetConfigMap>({});
 
   // As schema is dynamic, it is possible, that new updated values, will differ from one stored.
   const mergedState = useMemo(
     () =>
       merge(
         buildPathInitialState(Array.isArray(formFields) ? formFields : [formFields], formValues),
-        merge(overriddenWidgetState, uiOverrides)
+        merge(overriddenWidgetState)
```
lmossman, line 82:
```suggestion
        overriddenWidgetState
```
This `merge()` call is now unnecessary

airbyte-webapp/src/views/Connector/ConnectorForm/useBuildForm.tsx
```diff
@@ -54,19 +70,18 @@ interface BuildUiWidgetsContextHook {
 
 export const useBuildUiWidgetsContext = (
   formFields: FormBlock[] | FormBlock,
-  formValues: ConnectorFormValues,
-  uiOverrides?: WidgetConfigMap
+  formValues: ConnectorFormValues
 ): BuildUiWidgetsContextHook => {
-  const [overriddenWidgetState, setUiWidgetsInfo] = useState<WidgetConfigMap>(uiOverrides ?? {});
+  const [overriddenWidgetState, setUiWidgetsInfo] = useState<WidgetConfigMap>({});
```
lmossman, line 75: Since there is no more concept of "ui overrides" here, I think we can just name this something like `uiWidgetsInfo` instead of `overriddenWidgetState`

lmossman (approved):
A couple of small suggestions, but overall this looks like a good incremental improvement to me.

I don't think I'll need to take another look after you address my comments bc they are pretty straightforward, so approving now.

Tested locally and didn't notice any regressions from master, though I did find a bug in master which I captured in an issue here: https://github.com/airbytehq/airbyte/issues/20332

airbyte-webapp/src/views/Connector/ConnectorForm/useBuildForm.tsx
flash1293: I will take care of this in a follow-up PR

PR #20135, opened 2022-12-06
🪟🔧  Remove top level state for unfinished flows in connector form

flash1293:

Based on https://github.com/airbytehq/airbyte/pull/20132 (wait until merged)

## What

This PR removes the top level state for "unfinished flows" (the confirm control to "open" a secret field and submit it) and pulls it into local component state.

This is one step towards removing the ui-widget state completely from the connector form to make it easier to reason about state handling as well as integrating it in other places.

## How

* Remove all "unfinished flow" state handling and passing around of props related to it
* Introduce a new local state in the confirmation control to do the same thing
  * Consolidate the difference between secret text area and password input into the confirmation control (naming it `SecretConfirmationControl` as it's not used anywhere else) to avoid the local state leaking out of that component
  * Listen to the global formik state - if the dirty flag flips to `false` but there is a "previous value", reset that as well

## 🚨 User Impact 🚨

There is one small detail which changed based on this - the "Save changes and test" button would be disabled as long as the "unfinished flow" still exists. As there is no global state anymore to inform the button a secret is in the process of being edited, it will allow to "retest" even if the secret editing is not "finished" yet.

I don't see a UX problem with this change (especially as it's only affecting editing an existing connector but not creating a new one), but if there is a strong reason it can't work this way we can introduce a boolean top level state to pass this information around.


airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/SecretConfirmationControl.tsx
```diff
@@ -0,0 +1,114 @@
...
+        error={error}
+      />
+    ) : (
+      <Input
+        {...field}
+        autoComplete="off"
+        value={field.value ?? ""}
+        type="password"
+        disabled={disabled}
+        error={error}
```
lmossman, line 55: Something I noticed while testing: the secret input border is not set to red when it has an error
Screen Shot 2022-12-12 at 11 58 57 AM

I think this is because the `error` prop is not being passed to this component from `Control.tsx`

flash1293: Good catch, seems like it's like that on master too. Should be fixed:
Screenshot 2022-12-12 at 21 24 38

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/SecretConfirmationControl.tsx
```diff
@@ -0,0 +1,114 @@
...
+      />
+    );
+
+  useEffect(() => {
+    if (!dirty && !touched && previousValue) {
+      setPreviousValue(undefined);
+    }
+  }, [dirty, helpers, previousValue, touched]);
```
lmossman, line 63: > Listen to the global formik state - if the dirty flag flips to false but there is a "previous value", reset that as well

Just to confirm my understanding: this code is basically ensuring that if the formik form is reset back to the initial state, then the `previousValue` state will be cleared out? An in practice, this would happen if a user clicks the `Cancel` button at the bottom of the ConnectorForm

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/SecretConfirmationControl.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const onCancel = () => {
+    if (previousValue) {
+      helpers.setValue(previousValue);
+    }
+    setPreviousValue(undefined);
+  };
+
+  return (
+    <ComponentContainer>
+      {React.cloneElement(component, {
```
lmossman, line 91: Why do we need `cloneElement` here? Could we not just set the ref and the other attributes here to those values directly on the`<SecretTextArea>` and `<Input>` above?

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/SecretConfirmationControl.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const [field, , helpers] = useField(name);
+  const [previousValue, setPreviousValue] = useState<unknown>(undefined);
+  const isEditInProgress = Boolean(previousValue);
+  const controlRef = useRef<HTMLElement>(null);
+
+  const { dirty, touched } = useFormikContext();
+
+  const component =
+    multiline && (isEditInProgress || !showButtons) ? (
+      <SecretTextArea
```
lmossman, line 40: I noticed a couple small styling differences with the text area component which I put up a small PR here to fix: https://github.com/airbytehq/airbyte/pull/20397

lmossman (approved):
Had a couple more small questions but overall this looks good to me. Tested locally and it seems to work as expected. The small UX change you called out also feels fine to me

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/SecretConfirmationControl.tsx
flash1293: Exactly, if the form-level "Cancel" button is hit, all open secret controls are "closed" as well.

airbyte-webapp/src/views/Connector/ConnectorForm/components/Property/SecretConfirmationControl.tsx
flash1293: good point, we don't - simplified this.

PR #20076, opened 2022-12-05
ci-credentials: update GSM secrets with updated configuration values

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/19441 (2/2)

https://github.com/airbytehq/airbyte/pull/19979 enabled SAT to write updated configurations by capturing CONTROL messages. 
When running in the CI, we need to upload these updated configurations to GSM. This is the scope of this PR.

## How
1. Refactor `ci-credentials` to expose two commands via `click` framework: 
  - `write-to-storage`: the already existing command that downloads GSM secrets locally for test runs
  - `update-secrets`: the new command to update GSM secrets according to the content of the `secrets/updated_configurations` folder that might contain SAT updated configs. This command finds the latest local updated configuration and updates the existing GSM secret by creating a new secret version and disabling the previous one.
  https://github.com/airbytehq/airbyte/blob/1e75110cadffcb25ea691363325a6b3f88b066ca/tools/ci_credentials/ci_credentials/main.py#L23

2. Create `Secret` and `RemoteSecret` models for convenient manipulation of secret structure.
https://github.com/airbytehq/airbyte/blob/1e75110cadffcb25ea691363325a6b3f88b066ca/tools/ci_credentials/ci_credentials/models.py#L15

3. Rename `SecretLoader` to `SecretManager` and add methods to update secrets and disable version. 

https://github.com/airbytehq/airbyte/blob/1e75110cadffcb25ea691363325a6b3f88b066ca/tools/ci_credentials/ci_credentials/secrets_manager.py#L246

4. Test `SecretManager` public methods and models: 

https://github.com/airbytehq/airbyte/blob/1e75110cadffcb25ea691363325a6b3f88b066ca/tools/ci_credentials/tests/test_secrets_manager.py#L57
https://github.com/airbytehq/airbyte/blob/1e75110cadffcb25ea691363325a6b3f88b066ca/tools/ci_credentials/tests/test_models.py#L44

5. Change Github workflows using `ci-credentials` to make them call the new `update-secrets` commands after test run.

## 🚨 User Impact 🚨
This PR changes the `/test` and `/publish` workflows, so we shall make sure this command are still functioning as expected.

PR #20138, opened 2022-12-06
Enable full SAT for the Snowflake source

### LINKED ISSUE #19896: Enable SAT for hosted sources - Snowflake
Enable the SAT for the Snowflake source. See epic description for more context airbytehq/airbyte#19708.

DoNotPanicUA:
## What
Enable SAT for the Snowflake source

## Configured test instances
- sat_test_dataset

## Found issues
SAT related issues:
- The source returns time data with milliseconds, but the SAT treats this as a type violation. - :warning: **Shall we trim the data in the source?**
- The SAT fails when face non-typical values like `Nan`, `+inf` or `-inf`


airbyte-integrations/connectors/source-snowflake/acceptance-test-config.yml
```diff
@@ -5,3 +5,19 @@ tests:
   spec:
     - spec_path: "src/test-integration/resources/expected_spec.json"
       config_path: "secrets/config.json"
+  connection:
```
alafanechere, line 8: Please use the latest acceptance-test-config format. [Example here.](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-zendesk-support/acceptance-test-config.yml#L1). I'd be curious to check the test output under `high` test strictness level, we might get this for free 😄

alafanechere (changes requested):

airbyte-integrations/connectors/source-snowflake/acceptance-test-config.yml
```diff
@@ -1,7 +1,29 @@
 # See [Source Acceptance Tests](https://docs.airbyte.com/connector-development/testing-connectors/source-acceptance-tests-reference)
 # for more information about how to configure these tests
 connector_image: airbyte/source-snowflake:dev
-tests:
+acceptance_tests:
```
alafanechere, line 4:
```suggestion
test_strictness_level: high
acceptance_tests:
```
I'd be curious to check the test output under high test strictness level, we might get this for free.

DoNotPanicUA:
@alafanechere 
The CI run fails with the new format, but the local pass is fine.
Screenshot from 2022-12-07 17-05-27
any ideas on how to fix it?

alafanechere:
> any ideas on how to fix it? 
I think you spotted a bug in my latest PR: https://github.com/airbytehq/airbyte/pull/20076
I'm going to take a look.

DoNotPanicUA:
@alafanechere 
Looks like we can't set the strictness level as High because it requires the future cursor check.
This check doesn't work for the java sources because they use the old State massage handler, and it requires more deep investigation with the significant refactoring of the JDBC sources.

alafanechere:
> Looks like we can't set the strictness level as High because it requires the future cursor check.
This check doesn't work for the java sources because they use the old State massage handler, and it requires more deep investigation with the significant refactoring of the JDBC sources.

It makes sense, thank you for trying!

alafanechere (changes requested):
Requesting changes for the reasons explained here: https://github.com/airbytehq/airbyte/pull/19915#discussion_r1042325573

alafanechere (approved):

PR #20141, opened 2022-12-06
New Streams Table Column Sizing Fix

### LINKED ISSUE #17937: New connection streams table design pass
Update the connection streams table design to match the updated design [seen in the prototype](https://www.figma.com/proto/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?page-id=857%3A24319&node-id=865%3A52984&viewport=124%2C-519%2C0.25&scaling=min-zoom&starting-point-node-id=865%3A52984&show-proto-sidebar=1)

All the foundational work should be done in #17823. This is to match the table row design to match what's in Figma.

krishnaglick:
## What
Resolves https://github.com/airbytehq/airbyte/issues/17937

## How
Explicitly sets the widths for the columns, and prevents growing and shrinking.
Cleans up Cell component properties


airbyte-webapp/src/components/SimpleTableComponents/SimpleTableComponents.tsx
```diff
@@ -22,18 +22,17 @@ export const Header = styled(Row)`
   padding: 0;
 `;
 
+const flexCalc = (flex?: string | number) => (typeof flex === "number" ? `${flex} 0 0 ` : flex) ?? "1 0 0";
```
krishnaglick, line 25: Previous flex functionality was `${flex || 1} 0 0`. Now it can be a string.

krishnaglick:
Did a timebox on repeating this on the old table. It's a touch more complex than the new table use-case and I don't feel it's worth more time.

airbyte-webapp/src/components/connection/CatalogTree/CatalogTreeSubheader.tsx
```diff
@@ -9,7 +9,7 @@ import { useConnectionFormService } from "hooks/services/ConnectionForm/Connecti
 
 import styles from "./CatalogTreeSubheader.module.scss";
 
-const SubtitleCell = styled(Cell).attrs(() => ({ lighter: true }))`
+const SubtitleCell = styled(Cell).attrs(() => ({ light: true }))`
```
krishnaglick, line 12: This was not caught by the rename! Can't wait to get rid of styled components.

airbyte-webapp/src/components/connection/CatalogTree/CatalogTreeBody.module.scss
```diff
@@ -1,9 +1,10 @@
 .container {
   margin-bottom: 29px;
   max-height: 600px;
-  overflow-y: auto;
+  min-width: fit-content;
```
krishnaglick, line 4: Absolutely core to this working.

airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableCell.module.scss
```diff
@@ -0,0 +1,21 @@
+$small: 60px;
+$medium: 120px;
+$large: 200px;
+
+.tableCell {
+  flex: 0 0 $medium;
+  min-width: $medium;
```
krishnaglick, line 7: Need to pair this with the flex-basis for it to work.

krishnaglick:
If you have multiple primary keys and you add or remove some, the implicit width (due to the use of `min-width: fit-content;`) of the table changes so you will see the scrollbar jump.
This feels like a minor issue compared with the table actually working like we expect that I don't think it's worth revisiting.

edmundito (commented):
Looks great! I tested with the new table and the old table as well as bulk edit for both.

Unfortunately, there's one more thing. While expanding the old table with the fields, I found this: 
image

krishnaglick:
> Looks great! I tested with the new table and the old table as well as bulk edit for both.
> 
> Unfortunately, there's one more thing. While expanding the old table with the fields, I found this: image

Looks like setting the CatalogTree to `overflow: auto;` in the old view fixes that so I've added it!

edmundito:
Tested it and found an issue with the old and new table:
When I disable the connection, it collapses the checkbox:
image

krishnaglick:
> Tested it and found an issue with the old and new table: When I disable the connection, it collapses the checkbox: image

This issue exists in prod so I made a [ticket](https://github.com/airbytehq/airbyte/issues/20974).

edmundito:
I did another review pass on the table fixes, and it looks good despite the checkbox issue. One thing I noticed is that the table headers were always visible, but they're not anymore. Was this a conscious decision?

krishnaglick:
> I tested again and it looks good despite the checkbox issue. One thing I noticed is that the table headers were always visible, but they're not anymore. Was this a conscious decision?

This is likely due to needed to move the header into the same containing div as the body so the flex-width matched. So it was not explicitly done but is likely not an easy revert.

airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableRow.module.scss
```diff
@@ -43,31 +42,19 @@
...
 }
-
+%streamRowCheckboxCell,
 .streamRowCheckboxCell {
-  margin-right: variables.$spacing-sm;
   max-width: 43px;
   text-align: center;
   font-size: 10px;
   line-height: 13px;
   display: flex;
```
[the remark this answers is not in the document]
krishnaglick: The intent of this was that fixing the header columns would be a quick fix, but that has not been the case. This table really should be a proper table, probably utilizing react-table. But that's concern beyond this ticket right now.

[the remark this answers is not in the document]
josephkmh: Ok, maybe I'll take that on as a part of adding the new column 👍

edmundito (approved):
Looks good. Tested both tables locally.

Discussed with @krishnaglick about not having sticky headers for now and making any other detail improvements separately.

PR #20146, opened 2022-12-06
🪟🚨 Refactor connector form code

### LINKED ISSUE #14250: (Resolve) Tech debt of the ConnectorForm

### LINKED ISSUE #15108: Don't change validationSchema based on input
Currently our code does change the `validationSchema` of the ServiceForm based on some values, e.g. if the user selects another oneOf value with a sub object the whole validationSchema will change to reflect that.

I think we should try to generate the `validationSchema` once for the connectors spec and not change it based on the form values. Yup supports [`when`](https://github.com/jquense/yup#schemawhenkeys-string--string-builder-object--values-any-schema--schema-schema) to actually annotate that specific parts of a schema are only in effect when specific criterias are met. I think we should rather use that functionality to generate one validationSchema, that uses `when` to validate the different choices of a `oneOf` JSONSchema.

Especially since Formik doesn't revalidate the form when the validationSchema changes, we need to have special hacks for that in place.

This would require https://github.com/airbytehq/airbyte/issues/13056 to be solved, so that the formik form actually only contain the individual connectors form, and thus the validationSchema could be kept unchanged while filling the form.

flash1293:
Closes https://github.com/airbytehq/airbyte/issues/14250
Closes https://github.com/airbytehq/airbyte/issues/15108

## What

This PR takes care of the rest of the connector form tech debt parts:
*  Remove uiWidgetsInfo state by finding the enum/const field defining the selected oneOf and adding it to the condition form block to check the selected value by looking at the formik state
* Only do default value calculation once before rendering the formik component the first time (remove PatchInitialValuesWithWidgetConfig hack)
* Build yup schema once by flattening the oneOf conditions into a single object and adding when conditions on the found condition key (remove RevalidateOnValidationSchemaChange hack)

## How

Selected conditions of a oneOf are not tracked via `uiWidgetsInfo` state anymore. Instead, the `FormBlock` data structure contains a new property "selectionPath" which is pointing to the shared const prop of all conditions. As this is part of the formik values, it can be looked up from there directly. A second new property "selectionConstValues" contains an array of possible values for the const field in the same order as the condition form blocks itself - checking which one is selected becomes a `selectionConstValues.indexOf(formikState.get(selectionPath))`.

The yup schema used for actual validation is not rebuilt based on the `uiWidgetsInfo` state anymore. Instead, the keys out of all conditions are flattened into a single yup object shape, with `when` conditions applying the correct sub-schema based on the value of the "selectionPath". If the selected condition does not specify a certain key at all, its schema is set to "strip" which removes it on casting.

Setting default values doesn't happen within a hack-component anymore but is part of the `useBuildForm` hook now which is also generating the form blocks data structure.

In a few places from within the new logic, errors are thrown if the rules of airbyte schema are violated. For these, there is a special error type which is caught by the error boundary and rendered with a nice message and a link to the documentation. This should not happen for any of the Airbyte-maintained connectors but might be a case for custom ones.

The uiWidgets terminology was used in inconsistent ways - this PR removes it all together - now this is the flow that's happening when rendering the form for the first time:
* json schema -> form blocks and initial default values
* json schema + form blocks -> yup schema

On update, only the formik values are updated.

As the initial form values on empty forms passed down from the connector card were re-built on every render, the init logic outlined above would sometimes happen multiple times (depending on how often react-query is re-rendering the component). By memoizing these initial values on the connector card level, it's only happening a single time.

## 🚨 User Impact 🚨

The snowflake destination in version `<=0.4.40` does not work together with the changes on this PR - existing connections will continue to work fine, but it's not possible to change the configuration. Please update the snowflake destination connector to `0.4.41`.

oneOf properties not following the rules described in [the documentation](https://docs.airbyte.com/connector-development/connector-specification-reference/#airbyte-modifications-to-jsonschema) will stop working in the UI - the form will crash and a meaningful error is shown which also links to the documentation:
Screenshot 2022-12-09 at 12 00 11


airbyte-webapp/src/views/Connector/ConnectorCard/ConnectorCard.tsx
```diff
@@ -165,7 +165,12 @@ export const ConnectorCard: React.FC<ConnectorCardCreateProps | ConnectorCardEdi
   const job = jobInfo || LogsRequestError.extractJobInfo(errorStatusRequest);
 
   // Fill form with existing connector values otherwise set the default service name
-  const formValues = isEditMode ? props.connector : { name: selectedConnectorDefinition?.name };
+  const formValues = useMemo(
+    () => (isEditMode ? props.connector : { name: selectedConnectorDefinition?.name }),
+    // linter doesn't understand that props.connector is checked
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [isEditMode, isEditMode && props.connector, selectedConnectorDefinition?.name]
```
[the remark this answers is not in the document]
flash1293: Because props.connector only exists if isEditMode is truthy (it's a union type)

airbyte-webapp/src/views/Connector/ConnectorForm/useBuildForm.tsx
```diff
@@ -41,16 +41,51 @@ export function useBuildForm(
...
+    const baseValues = {
       name: "",
       connectionConfiguration: {},
       ...initialValues,
-    }),
-    [initialValues]
-  );
+    };
+
+    function setDefaultValues(formGroup: FormGroupItem, values: Record<string, unknown>) {
```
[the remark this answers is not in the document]
flash1293: As it's only called in here recursively this seemed cleaner to me (keeps it closer to the place where it's used). No strong opinion on these things though, let me know if you would prefer to move it.

airbyte-webapp/src/views/Connector/ConnectorCard/ConnectorCard.tsx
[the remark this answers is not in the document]
krishnaglick: My only argument for doing the above method is if this hook changes the linter will pick up new deps.

airbyte-webapp/src/views/Connector/ConnectorForm/useBuildForm.tsx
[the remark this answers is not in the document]
krishnaglick: I'm partial to keeping unwrapped (non-useCallback, etc) methods at the top scope within the react render tree since every time this is called (which may be a small number of times) it's redefined.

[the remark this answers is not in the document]
flash1293: Wouldn't matter in this specific case, but I agree it's a good default. Improved

airbyte-webapp/src/views/Connector/ConnectorCard/ConnectorCard.tsx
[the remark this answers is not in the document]
flash1293: Found a way to structure this better, thanks for the suggestion.

airbyte-webapp/src/core/form/schemaToFormBlock.ts
```diff
@@ -37,20 +38,49 @@ export const jsonSchemaToUiWidget = (
...
 
   if (jsonSchema.oneOf?.length && jsonSchema.oneOf.length > 0) {
-    const conditions = Object.fromEntries(
-      jsonSchema.oneOf.map((condition) => {
-        if (typeof condition === "boolean") {
-          return [];
-        }
-        return [condition.title, jsonSchemaToUiWidget({ ...condition, type: jsonSchema.type }, key, path)];
-      })
+    let possibleConditionSelectionKeys: Set<string> | null = null as Set<string> | null;
```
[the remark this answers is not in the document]
flash1293: Yes, otherwise typescript tries to be smart and assumes it will always be `null`:
Screenshot 2022-12-13 at 16 19 14

However, we can omit the type of the variable. Improved.

flash1293:
Thanks for the review @krishnaglick , fixed the things you highlighted.

krishnaglick (approved):
Nothing else stands out to me but someone closer to these changes should definitely do a thorough review!

flash1293:
> but someone closer to these changes should definitely do a thorough review!

Definitely, this is a change that can break quite a few things in subtle ways, let's make sure to test it thoroughly. I will give it another round myself.

airbyte-webapp/src/core/form/schemaToYup.ts
```diff
@@ -0,0 +1,208 @@
...
+      Object.entries(condition.properties || {}).forEach(([key, prop], propertyIndex) => {
+        if (!flattenedKeys.has(key)) {
+          flattenedKeys.set(key, []);
+        }
+        flattenedKeys
+          .get(key)
+          ?.push([
+            selectionConstValue,
+            typeof prop === "boolean"
+              ? yup.mixed()
```
lmossman, line 59: Why `yup.mixed()` instead of `yup.boolean()`?

airbyte-webapp/src/core/form/schemaToYup.ts
```diff
@@ -0,0 +1,208 @@
...
+    | null = null;
+
+  if (jsonSchema.oneOf && propertyPath) {
+    const conditionFormField = formField as FormConditionItem;
+    // for all keys in all of the possible objects from the oneOf, collect the sub yup-schema in a map.
+    // the keys of the map are the keys of the property, the value is an array of the selection const value
+    // for the condition plus the sub schema for that property in that condition.
+    // As not all keys will show up in every condition, there can be a different number of possible sub schemas
+    // per key; at least one and at max the number of conditions (if a key is part of every oneOf)
+    const flattenedKeys: Map<string, Array<[unknown, yup.AnySchema]>> = new Map();
```
lmossman, line 43: Why is the value type of the map `Array<[unknown, yup.AnySchema]>`? Shouldn't it be `Array<[string, yup.AnySchema]>`, since all properties are strings?

airbyte-webapp/src/core/form/schemaToYup.ts
```diff
@@ -0,0 +1,208 @@
...
+      break;
+    case "object":
+      let objectSchema = yup.object();
+
+      const keyEntries = Object.entries(jsonSchema.properties || {}).map(([propertyKey, propertySchema]) => {
+        const correspondingFormField = (formField as FormGroupItem).properties.find(
+          (property) => property.fieldKey === propertyKey
+        );
+        if (!correspondingFormField) {
+          throw new FormBuildError("mistmatch between form fields and schema");
```
lmossman, line 153:
```suggestion
          throw new FormBuildError("mismatch between form fields and schema");
```

airbyte-webapp/src/core/form/schemaToYup.ts
```diff
@@ -0,0 +1,208 @@
...
+): yup.AnySchema => {
+  let schema:
+    | yup.NumberSchema
+    | yup.StringSchema
+    | yup.AnyObjectSchema
+    | yup.ArraySchema<yup.AnySchema>
+    | yup.BooleanSchema
+    | null = null;
+
+  if (jsonSchema.oneOf && propertyPath) {
```
lmossman, line 36: This whole if block took me a few tries to understand as it is fairly complex. It would be great if we could simplify it a bit but I'm not immediately seeing a way to do that, and I don't think that should necessarily block merging

airbyte-webapp/src/core/form/schemaToYup.ts
lmossman, line 153: This seems like more of an internal bug, so we should probably track this error in datadog so we can be aware of it when it happens. We have a service (app monitoring service) which can be used to track errors like this. Here is an example: https://github.com/airbytehq/airbyte/blob/2392acb845891f9c031b06feee9af2a2f975f1a7/airbyte-webapp/src/views/Connector/ConnectorForm/useAuthentication.tsx#L156-L159
Here is a slack post with more context about this service: https://airbytehq-team.slack.com/archives/C039SG3DQ8K/p1668506492916659

In fact, we may just want to track all FormBuildErrors?

lmossman (commented):
Had several comments. Overall I think the changes make sense. They get a bit complex in some areas but that is likely just a result of this being a complex problem to solve.

I did not get a chance to test this but I will do so before I give it a full approval

airbyte-webapp/src/core/form/schemaToFormBlock.ts
```diff
@@ -37,20 +38,49 @@ export const jsonSchemaToUiWidget = (
...
-      jsonSchema.oneOf.map((condition) => {
-        if (typeof condition === "boolean") {
-          return [];
-        }
-        return [condition.title, jsonSchemaToUiWidget({ ...condition, type: jsonSchema.type }, key, path)];
-      })
+    let possibleConditionSelectionKeys = null as Set<string> | null;
+    const conditions = jsonSchema.oneOf.flatMap((condition) => {
+      if (typeof condition === "boolean") {
+        throw new FormBuildError("Spec uses oneOf without using object types for all conditions");
```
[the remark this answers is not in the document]
flash1293: You are 100% right, fixed

airbyte-webapp/src/core/form/schemaToFormBlock.ts
```diff
@@ -37,20 +38,49 @@ export const jsonSchemaToUiWidget = (
...
 
   if (jsonSchema.oneOf?.length && jsonSchema.oneOf.length > 0) {
-    const conditions = Object.fromEntries(
-      jsonSchema.oneOf.map((condition) => {
-        if (typeof condition === "boolean") {
-          return [];
-        }
-        return [condition.title, jsonSchemaToUiWidget({ ...condition, type: jsonSchema.type }, key, path)];
-      })
+    let possibleConditionSelectionKeys = null as Set<string> | null;
```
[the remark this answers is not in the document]
flash1293: It would be, but if you do that, Typescript will auto-narrow down your type to just `null`:
Screenshot 2022-12-15 at 11 54 04

airbyte-webapp/src/core/form/schemaToFormBlock.ts
```diff
@@ -37,20 +38,49 @@ export const jsonSchemaToUiWidget = (
...
+        throw new FormBuildError("Spec uses oneOf without using object types for all conditions");
+      }
+      const formBlock = jsonSchemaToFormBlock({ ...condition, type: jsonSchema.type }, key, path);
+      if (formBlock._type !== "formGroup") {
+        return [];
+      }
+
+      const constProperties = new Set(
+        formBlock.properties.filter((property) => property.const).map((property) => property.fieldKey)
+      );
```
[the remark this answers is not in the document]
flash1293: The `Set` class is an ordered set in Javascript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

I know it's not important for the amount of data we are dealing with, but theoretically the intersect is slightly faster this way because the `has` method is constant time now instead of linear time for ``, reducing the algorithm from `n*n` to `n`.

That said, it's probably better to stick with arrays here to keep the logic simple as there's a lot of stuff going on already - rewrote it using arrays and the lodash intersection helper which I hope will do if fast (if not its easy to refactor later on)

airbyte-webapp/src/core/jsonSchema/index.ts
```diff
@@ -1,3 +1,27 @@
...
+  airbyte_secret?: boolean;
+  airbyte_hidden?: boolean;
+  multiline?: boolean;
+  order?: number;
+}
+
+/**
+ * Remaps all {@link JSONSchema7} to Airbyte Json schema
+ */
+export type AirbyteJSONSchema = {
```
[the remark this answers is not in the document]
flash1293: Good point, changed.

airbyte-webapp/src/core/form/schemaToYup.ts
flash1293: I inherited this from the old code. It didn't strike me as exceptionally important as we render a switch anyway for these so the user can't really do wrong on them in the first place.

Still, changed it to keep the code unsurprising.

airbyte-webapp/src/core/form/schemaToYup.ts
flash1293: The keys of the map correspond with the property keys in each of the possible `oneOf`s - those need to be strings. But the array of `Array<[unknown, yup.AnySchema]>` is a list of mappings from the value of the const property to the schema it implies - const properties can be anything, they are not limited to strings (in practice they are probably always strings, but numbers could be the case as well).

This is where the type for these comes from:
Screenshot 2022-12-15 at 12 13 15

`JSONSchema7Type` is pretty much everything you want:
Screenshot 2022-12-15 at 12 14 15

airbyte-webapp/src/core/form/schemaToYup.ts
flash1293: Yeah, that's pretty dense code. I added an example to the comment to explain what's going on.

airbyte-webapp/src/core/form/schemaToYup.ts
flash1293: Good idea, added that to the error boundary - if a form group error bubbles up, it is tracking the error using datadog.

flash1293:
Thanks for the review, addressed all your points.

airbyte-webapp/src/core/form/schemaToYup.test.ts
```diff
@@ -0,0 +1,269 @@
...
+              default: "oauth",
+            },
+          },
+        },
+      ],
+    },
+  },
+};
+
+it("should build  mixed schema for conditional case", () => {
```
lmossman, line 117:
```suggestion
it("should build mixed schema for conditional case", () => {
```

airbyte-webapp/src/core/form/schemaToYup.test.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  const expectedSchema = yup.object().shape({
+    start_date: yup.string().trim().required("form.empty.error").transform(String),
+    max_objects: yup.number().transform((x) => x),
+    credentials: yup.object().shape({
+      type: yup.mixed(),
+      api_key: yup
+        .mixed()
+        .when("type", { is: "", then: (x) => x, otherwise: (x) => x })
+        .when("type", { is: "", then: (x) => x, otherwise: (x) => x }),
```
lmossman, line 128: This doesn't make sense to me. Wouldn't we expect the `api_key` schema to be something like
```
api_key: yup
  .mixed()
  .when("type", { is: "api", then: (x) => x.string().matches(\\w{5}), otherwise: (x) => x })
  .when("type", { is: (val) => !["api"].includes(val), then: (x) => x.strip() })
```

same comment goes for the `redirect_uri` schema below, and the schemas at the end of this file

airbyte-webapp/src/core/form/schemaToYup.test.ts
```diff
@@ -0,0 +1,269 @@
...
+      }
+    );
+    expect(cleanedValues.credentials).toEqual({
+      type: "api",
+      api_key: "X",
+    });
+  });
+});
+
+it("should build schema for conditional case with inner schema and form blocks", () => {
```
lmossman, line 206: This test implementation doesn't seem to line up with this description to me. I'm not sure what `topKey` and `topKey.subKey` are doing on lines 250-251. It seems like this should be nesting `credentials` under a few levels or something?

airbyte-webapp/src/core/form/schemaToYup.ts
lmossman: The examples are super helpful, thanks!

airbyte-webapp/src/core/form/schemaToYup.ts
lmossman: Would it work to make the type of `FormConditionItem.selectionConstValues` be something like `Array<JSONSchema7Type | undefined>` then?

Just asking because it feels like we should try to avoid `unknown` where we can and this feels like a place where we may be able to eliminate it

lmossman (commented):
Had a few more comments, main ones being about how the tests look like they may be incorrect 

I also tested this locally with a few sources (postgres, linkedin ads, etc) and the behavior looked correct to me!

airbyte-webapp/src/core/form/schemaToFormBlock.ts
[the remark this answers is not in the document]
flash1293: I think there are two kinds of errors thrown from our code:
* Our fault - unexpected errors which are 100% on us because there is a bug in the code
* User fault - errors that can happen due to user input

I think the first kind doesn't need to be i18n-ified, it will most likely anyway not tell the user anything useful, it's just relevant for devs to figure out what's going on.

The second kind should be i18n-ified because ideally the user can fix it themselves.

I originally though these errors would be of the first kind, but on second thought they can also be seen as user errors, especially with the push to allow custom connectors. So I went ahead and translated them. `FormBuildError` is now only meant for errors which are caused by a json schema not following the documented rules. I changed one place where it was thrown back to `Error` because it can only be caused by a bug on our side.

airbyte-webapp/src/core/form/schemaToYup.test.ts
flash1293: This test is only testing the structure of the yup schema which does not capture the actual conditions of the whens:
```json
 "api_key": {
              "type": "mixed",
              "_whitelist": {
                "list": {},
                "refs": {}
              },
              "_blacklist": {
                "list": {},
                "refs": {}
              },
              "exclusiveTests": {},
              "deps": [
                "type",
                "type"
              ],
              "conditions": [
                {
                  "refs": [
                    {
                      "key": "type",
                      "isContext": false,
                      "isValue": false,
                      "isSibling": true,
                      "path": "type"
                    }
                  ]
                },
                {
                  "refs": [
                    {
                      "key": "type",
                      "isContext": false,
                      "isValue": false,
                      "isSibling": true,
                      "path": "type"
                    }
                  ]
                }
              ],
```

The tests below this one (the stuff in `should schema build conditional case that validates correctly based on selection key`) test whether the `when`s actually kick in. 

Added a comment to explain better.

airbyte-webapp/src/core/form/schemaToYup.test.ts
flash1293: I copied this test over and didn't adjust much about it. It is testing whether conditionals also work in nested cases, but I don't see how we need that test as the base schema is also nesting the oneOf in an object...

Removed the test as it's more confusing than helpful.

airbyte-webapp/src/core/form/schemaToYup.ts
flash1293: I'm not too worried about using `unknown` as it's a clear marker that you shouldn't assume anything about this value -  it basically allows you to do nothing with it except for passing it around and checking whether it's the same value. 

Switching to `JSONSchema7Type` is closer to the truth though - updated the code (we can even drop the "undefined" as this would mean in one condition no `const` is set which is something we throw on).

flash1293:
I clicked through all the sources and all the destinations and found one that doesn't work - the Snowflake destination.
It does not specify the "auth_type" property for the simple password case:
```json
[
    {
        "type": "object",
        "order": 0,
        "title": "OAuth2.0",
        "required": [
            "access_token",
            "refresh_token"
        ],
        "properties": {
            "auth_type": {
                "enum": [
                    "OAuth2.0"
                ],
                "type": "string",
                "const": "OAuth2.0",
                "order": 0,
                "default": "OAuth2.0"
            },
            "client_id": {
                "type": "string",
                "title": "Client ID",
                "description": "Enter your application's Client ID",
                "airbyte_secret": true
            },
            "access_token": {
                "type": "string",
                "title": "Access Token",
                "description": "Enter you application's Access Token",
                "airbyte_secret": true
            },
            "client_secret": {
                "type": "string",
                "title": "Client Secret",
                "description": "Enter your application's Client secret",
                "airbyte_secret": true
            },
            "refresh_token": {
                "type": "string",
                "title": "Refresh Token",
                "description": "Enter your application's Refresh Token",
                "airbyte_secret": true
            }
        }
    },
    {
        "type": "object",
        "order": 1,
        "title": "Key Pair Authentication",
        "required": [
            "private_key"
        ],
        "properties": {
            "auth_type": {
                "enum": [
                    "Key Pair Authentication"
                ],
                "type": "string",
                "const": "Key Pair Authentication",
                "order": 0,
                "default": "Key Pair Authentication"
            },
            "private_key": {
                "type": "string",
                "title": "Private Key",
                "multiline": true,
                "description": "RSA Private key to use for Snowflake connection. See the <a href=\"https://docs.airbyte.com/integrations/destinations/snowflake\">docs</a> for more information on how to obtain this key.",
                "airbyte_secret": true
            },
            "private_key_password": {
                "type": "string",
                "title": "Passphrase",
                "description": "Passphrase for private key",
                "airbyte_secret": true
            }
        }
    },
    {
        "type": "object",
        "order": 2,
        "title": "Username and Password",
        "required": [
            "password"
        ],
        "properties": {
            "password": {
                "type": "string",
                "order": 1,
                "title": "Password",
                "description": "Enter the password associated with the username.",
                "airbyte_secret": true
            }
        }
    }
]
```

I can see two ways of fixing this:
* Update the connector to include the auth_type property for the password case as well before merging this PR. However, without a migration it would preselect the wrong condition in the form for existing configured destinations (no data loss would occur) - not sure whether this would be OK
* Adjust the frontend logic to handle the case by allowing one condition to _not_ have a const key (by exclusion that would be enough to make it work) - however that seems hacky and ad hoc.

I'm going to prepare the PR for updating the connector spec because we should definitely do this no matter which direction we decide to take.

Interested in your thoughts @lmossman

airbyte-webapp/src/components/common/ApiErrorBoundary/ApiErrorBoundary.tsx
```diff
@@ -72,6 +80,12 @@ class ApiErrorBoundaryComponent extends React.Component<
     }
   }
 
+  componentDidCatch(error: { message: string; status?: number; __type?: string }) {
+    if (isFormBuildError(error)) {
+      this.props.trackError(error);
```
lmossman, line 85: It would be nice if this could include a second argument to add some metadata about these errors, so they are easier to discover in datadog. I'm thinking of something like how errors are tracked in the useAuthenticator hook: https://github.com/airbytehq/airbyte/blob/2392acb845891f9c031b06feee9af2a2f975f1a7/airbyte-webapp/src/views/Connector/ConnectorForm/useAuthentication.tsx#L156-L159

If this could include an `id` like "formBuildError",  and also attach the connector ID to it, then we could see which connector the error occurred for. This seems important because I don't know how we would address a generic form build error that we see in datadog unless we know what connector it occurred for.

lmossman (commented):
Just had one comment about improving error tracking a bit, but I don't think it is blocking.

I think the approach you laid out for dealing with the Snowflake destination issue makes sense, though it does raise one concern: 

If OSS users upgrade their airbyte version to pull in these webapp changes, but they don't update their snowflake destination connector version to the one with your fix, then they will see the `Spec uses oneOf without a shared const property` error message if they try to create or edit any snowflake connector.

This may not be a huge concern as they would just need to update that connector in order to fix it -- we just need to be sure to communicate that clearly in the release notes. But is it worth exploring a way for us to handle that case without completely failing the webapp?

flash1293:
> If OSS users upgrade their airbyte version to pull in these webapp changes, but they don't update their snowflake destination connector version to the one with your fix, then they will see the Spec uses oneOf without a shared const property error message if they try to create or edit any snowflake connector

agreed, this needs to be made obvious.

> But is it worth exploring a way for us to handle that case without completely failing the webapp?

I honestly can’t think of a good way to do so without building a whole separate system of tracking the state which kind of goes counter the idea of the refactoring. Do you have an idea?

lmossman:
@flash1293 no I think I agree that it would just add more complexity to track the state in a way that we don't even want to support, so I don't think it makes sense. I think your idea of adding something like 
> Please ensure your connector is updated to the latest version

is a good approach for making that clear.

flash1293:
> to the error message is a good approach for making that clear, so let's go with that

done

> It would be nice if this could include a second argument to add some metadata about these errors, so they are easier to discover in datadog

done, also refactored the hooks a bit as it made things easier - now the connector form only knows a single "useBuildForm" hook which does all the things.

airbyte-webapp/src/components/common/ApiErrorBoundary/ApiErrorBoundary.tsx
```diff
@@ -72,6 +80,15 @@ class ApiErrorBoundaryComponent extends React.Component<
     }
   }
 
+  componentDidCatch(error: { message: string; status?: number; __type?: string }) {
+    if (isFormBuildError(error)) {
+      this.props.trackError(error, {
+        id: "formBuildError",
+        connector: error.connectorId,
```
lmossman, line 87:
```suggestion
        connectorDefinitionId: error.connectorId,
```
nit suggestion to make it clear that this is a connector definition ID (since connector ID usually refers to the ID of an instantiated source or destination record). 

Could you also update the field on the FormBuildError class to be `connectorDefinitionId` instead of `connectorId`?

lmossman (approved):
Just had one more small suggestion, otherwise this LGTM and I didnt see any unexpected behavior during testing.

Reminder: this should not be merged until https://github.com/airbytehq/airbyte/pull/20566 is published and merged

Issue #15628, opened 2022-08-12
[low-code connectors] lookback window should be applied on the cursor instead of on the start date


girarda:
## Current Behavior
The DatetimeStreamSlicer's lookback window is only applied on the start date, which means it won't work for subsequent syncs.

context: https://github.com/airbytehq/airbyte/pull/15027#discussion_r944858564

## Expected Behavior
The lookback window should be based off the cursor date instead of the state date when performing a sync that uses stream slicing. If the look back window is 3 days, on the first time sync, records starting from 3 days before the start date should be retrieved. For the subsequent incremental sync, the we should attempt to get records starting from 3 days before the date specified in the incoming state json.

## Implementation Approach:
Within the `datetime_stream_slicer.py` `DatetimeStreamSlicer.stream_slices()` method, we derive what the starting time of the request should be. We apply the lookback to the start time specified and then choose the greater of the `start_datetime` and `cursor_datetime`. We should actually compare the two datetimes first and then apply the lookback accordingly.

## Acceptance Criteria

- [ ] When running an incremental sync with state, it should yield records starting after the cursor date subtracted from the look back window

Good context for validation:
A good integration to test this on is `source_wikipedia_pageviews` because it doesn't require authentication and has datetime stream slicing implemented already. It doesn't have a lookback window specified, but you can add that in for testing purposes.

When running the first time sync, you should see the lookback window reflected by the date of the records received when you run a read command. Now you can pass in a state json starting later than the start_datetime using `--state`. On this second, you should see see records with dates earlier than that of the date in `state.json` according to the lookback window defined.

brianjlai:
make the acceptance criteria one bullet point and keep the context

maxi297:
PR associated with this change: https://github.com/airbytehq/airbyte/pull/20156

PR #20156, opened 2022-12-06
[ISSUE #15628] apply lookback window on earliest datetime between sta…

maxi297:
…rt and cursor

## What
This change addresses issue #15628 . The lookback period should be applied to whatever is the most recent datetime between the start and the cursor defined by the state. 

## How
Applying the lookback window on `max(start_datetime, cursor_datetime)` instead of the start_datetime

## 🚨 User Impact 🚨
There should not be any breaking changes. However starting from this version, we will pull more data if there is a lookback window defined. So assuming a source with a state including the cursor field and a lookback window:
* Before: incremental sync would start from the cursor field value defined in the state
* After: incremental sync will start from the cursor field value defined in the state *minus the lookback window*


maxi297:
Do we release at every pull request?

If so, I know https://github.com/airbytehq/airbyte/pull/20019 bumps the version and should be merged soon so I'll wait for it to be merge before working on updating the version

brianjlai:
There isn't a strict rule and on occasion we'll package up multiple PRs into the same release for convenience, so if you want them grouped because the other PR is going out, thats fine. Just make sure the changelog reflects both changes whether you include the blurb in this PR or the other that is actually doing the release

airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py
```diff
@@ -145,28 +145,19 @@ def stream_slices(self, sync_mode: SyncMode, stream_state: Mapping[str, Any]) ->
...
-        start_datetime = self.start_datetime.get_datetime(self.config, **kwargs) - lookback_delta
-        start_datetime = min(start_datetime, end_datetime)
-        if self.cursor_field.eval(self.config, stream_state=stream_state) in stream_state:
-            cursor_datetime = self.parse_date(stream_state[self.cursor_field.eval(self.config)])
-        else:
-            cursor_datetime = start_datetime
-
-        start_datetime = max(cursor_datetime, start_datetime)
 
-        state_cursor_value = stream_state.get(self.cursor_field.eval(self.config, stream_state=stream_state))
```
brianjlai, line 157: not anything wrong with your code, but i do find this whole block of code very odd in that we get and parse the cursor date time, and then we do it again and compare the max multiple times.

brianjlai (approved):
changes look good! nice job cleaning up this method too, I don't quite remember why we had originally wrote this method in a confusing way with the multiple cursor field comparisons against the start date

airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py
```diff
@@ -145,28 +145,19 @@ def stream_slices(self, sync_mode: SyncMode, stream_state: Mapping[str, Any]) ->
...
-        if state_date:
-            # If the input_state's date is greater than start_datetime, the start of the time window is the state's next day
-            next_date = state_date + datetime.timedelta(days=1)
-            start_datetime = max(start_datetime, next_date)
         dates = self._partition_daterange(start_datetime, end_datetime, self._step)
         return dates
 
+    def _calculate_cursor_datetime_from_state(self, stream_state: Mapping[str, Any]) -> datetime.datetime:
+        if self.cursor_field.eval(self.config, stream_state=stream_state) in stream_state:
+            return self.parse_date(stream_state[self.cursor_field.eval(self.config)]) + datetime.timedelta(days=1)
```
maxi297, line 158: Why do we have this `+ datetime.timedelta(days=1)`here? What does it mean? This isn't so clear to me and I would like to extract this as a constant with a clear name as to what it does. I see [a comment](https://github.com/airbytehq/airbyte/pull/20156/files#diff-3a6ff2f6eca71b4e166b1c8c852becb143ad4e079f5308967ff52828e747f035L164) here describing what we do, but I'm not sure as to why we do it

brianjlai: Just for extra context about incremental syncs a little bit. For streams that are configured as incrementals. The first time we sync, we will sync all the data up to the current day. Then at the end we save a stream state checkpoint to remind us where we stopped. On the second iteration, we'll use that `stream_state` checkpoint to tell us to begin at that point.

So idea here is that if we have received a `stream_state` as a datetime input, that would indicate that we had successfully completed an incremental sync before the current sync we're processing. And now on this current one, we want to start our next sync on the next day because we had already synced the current day in the prior sync.

For ex:
If we sync up to 12/7. We persist 12/7 into the state database. So that way on 12/8 when we perform our next sync, we only start getting records beginning on 12/8. This is pretty common for our connectors where the cursor date like created is `12-07-2022`. But that does raise a good question about how accurate this for integrations whose granularity is at the hours, minutes or even seconds, we could be missing thing. I think we should follow up on this. But it might fall a little outside the scope of this fix since this was how our existing behavior was.

PR #20162, opened 2022-12-06
🪟 🐛 Focus Connection Name input on field selection

### LINKED ISSUE #19229: Connection name is still active after clicking outside and the name has not been changed
## Actual 

When clicking outside the connection name without editing, it stays in the edit mode.

## Expected

When clicking outside the connection name without editing, it should leave the edit mode.

## Steps to reproduce

1. Navigate to a connection
2. Click on the name
3. Click outside of the name

teallarson:
## What

closes #19229 

**BEFORE**: 
We would render the input, but not focus it.  
https://www.loom.com/share/815b76c0ef8c48e2a0887a62b087d881

**AFTER**:
I added `autoFocus` to the Input element.  

The problem was that the field was never actually focused, just rendered, so the `blur` never got triggered.

This is technically against our a11y linting rules _however_ I am breaking this rule intentionally.  `autoFocus` is not recommended because it can lead to unexpected behavior for users.  However, when clicking into an input (or triggering the input appearing by clicking that button), it does not seem unexpected that the input would be focused.

This resolves the original issue as seen here:
https://www.loom.com/share/029c011240dd40ee815a5ede1f1723e6

- Decided against using a ref here to keep the fix simple, but am open to reconsidering if that seems right


edmundito (approved):
Tested locally

PR #20163, opened 2022-12-06
lint CDK files

evantahler:
Fix the build!

This is the result of `SUB_BUILD=CONNECTORS_BASE ./gradlew build -x test`


brianjlai (approved):

airbyte-integrations/connector-templates/source-java-jdbc/acceptance-test-config.yml.hbs
```diff
@@ -5,5 +5,4 @@ acceptance_tests:
   spec:
     tests:
       - spec_path: "src/test-integration/resources/expected_spec.json"
-        config_path: "src/test-integration/resources/dummy_config.json"
```
evantahler, line 8: There's a trailing space here that was ruining things :(

PR #20186, opened 2022-12-07
🪟 🔧 Clear invalid primary key on schema refresh

### LINKED ISSUE #19387: Clear invalid primary key or cursor when field is removed after schema refresh
Update the connection streams table to invalidate the primary key or cursor mode after the schema is refreshed and the field has been removed.

Currently, the field that was removed will remain selected as primary key or cursor as it was before. This is because the refreshed schema response still returns the field as the selected value, even though it has been removed.

There are two options to make this possible:

1. Update the backend to remove the fields from the updated syncCatalog. This may be the ideal solution because the backend should manage this logic and the frontend should just render what it's given
2. Have the frontend cleanup the syncCatalog when there's a diff. This is the easier solution

YatsukBogdan1:
## What
Closes #19387 

## How
While schema refreshing is in the process there is a utility function that wipes out any primary key and cursor inside the connection config, that was removed from the database


dizel852 (changes requested):
Tested locally - seems like we need to validate the form after refreshing the schema source since  PK is required for chosen sync mode.

Current behavior: 
The form is valid, but PK is required, and only after we click "Save changes", the form become invalid:
Screenshot at Dec 09 15-29-48

Previous(PK is not selected so the form is invalid):
Screenshot at Dec 09 15-30-26

Need confirmation from @edmundito and @tealjulia:
It seems like this case relates to the "breaking changes" case, am I right?

сс:
@SofiiaZaitseva +1 test case 🙂

edmundito:
Replaced by #20203

PR #20188, opened 2022-12-07
🪟 🐛 Fix test assertion in datepicker tests

joeykmh:
## What
* Fixes a bad assertion that relied on test running on a machine set to UTC


airbyte-webapp/src/components/ui/DatePicker/DatePicker.test.tsx
```diff
@@ -26,12 +26,15 @@ describe(`${toEquivalentLocalTime.name}`, () => {
...
 
   it("outputs the same YYYY-MM-DDTHH:mm:ss", () => {
+    timezoneMock.register("Etc/GMT+10");
+    const TEST_TIMEZONE_UTC_OFFSET_IN_MINUTES = 600; // correspond to GMT+10
     const TEST_UTC_TIMESTAMP = "2000-01-01T12:00:00Z";
 
     const result = toEquivalentLocalTime(TEST_UTC_TIMESTAMP);
 
     // Regardless of the timezone, the local time should be the same
-    expect(result?.toISOString().substring(0, 19)).toEqual(TEST_UTC_TIMESTAMP.substring(0, 19));
```
josephkmh, line 34: `Date.toISOString()` is [always set to the UTC timezone](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString#:~:text=The%20timezone%20is%20always%20zero%20UTC%20offset%2C%20as%20denoted%20by%20the%20suffix%20Z), so this assertion would only work in a UTC timezone.

Instead, we can assert that `toEquivalentLocalTime()` has the expected timezone offset with `Date.getTimezoneOffset()` and that `dayjs.format()`, which outputs a local timezone, will have the same `YYYY-MM-DDTHH:mm:ss` value.

flash1293 (approved):
LGTM, code review only as we discussed it in person.

PR #20203, opened 2022-12-07
🪟 🎉 Clear user-defined primary key(s) and cursor field(s) if the field is removed

### LINKED ISSUE #19387: Clear invalid primary key or cursor when field is removed after schema refresh

teallarson:
## What
closes airbytehq/airbyte#19387

Clears the primary key and/or cursor properties on the `stream.config` object if any of the fields referenced there are removed during a refresh.

## How
- `useInitialValues` (in `formConfig`) identifies if there is a breaking change on the connection.  If there is, it will traverse the diff and pull out the transforms for the field(s) that were removed
- the transforms get sent along to `calculateInitialValues` which is already used to calculate and override config fields
- as each stream is evaluated, if it is included in the list of streams with broken changes, a cleanup function is called that will find a match on either a primary key (or a partial of a compound primary key) or cursor field.  If a match is found for the removed field, we clear that property on the config object.

## Recommended reading order
1. `calculateInitialCatalog.test.tsx` (for expected behaviors)
2. `formConfig.tsx` 
3. `calculateInitialCatalog.ts` 

## Testing
This is not behind the feature flag, but should not affect anything negatively.  Open to discussion if we'd rather keep it behind the flag.  Either way, useful to test both with `REACT_APP_AUTO_DETECT_SCHEMA_CHANGES` as true and as false

At bare minimum:
- removing fields for user-configured primary keys and cursors
- adding and removing fields for source-defined primary keys and cursors

These are also covered by unit tests.

## Next steps
These files are due for a [refactor](https://github.com/airbytehq/airbyte-internal-issues/issues/1177).


airbyte-webapp/src/views/Connection/ConnectionForm/calculateInitialCatalog.ts
```diff
@@ -13,6 +16,29 @@ const getDefaultCursorField = (streamNode: SyncSchemaStream): string[] => {
...
 
+const cleanBreakingFieldChanges = (streamNode: SyncSchemaStream, breakingChangesByStream: StreamTransform[]) => {
+  breakingChangesByStream.forEach((streamTransformation) => {
+    // get all of the removed field paths
+    const removedFieldPaths = streamTransformation.updateStream?.map((update) => update.fieldName);
+
+    if (streamNode.config?.primaryKey && removedFieldPaths?.length) {
+      // if any of the field paths in the primary key match any of the field paths that were removed, clear the entire primaryKey property
+      if (
+        streamNode.config.primaryKey.some((key) => removedFieldPaths.some((removedPath) => isEqual(key, removedPath)))
```
tealjulia, line 27: it's possible that [lodash intersection](https://lodash.com/docs/#intersection) could tidy this up, but i haven't figured out how

airbyte-webapp/src/views/Connection/ConnectionForm/calculateInitialCatalog.test.ts
```diff
@@ -654,4 +674,128 @@ describe("calculateInitialCatalog", () => {
...
+      ],
+      false
+    );
+    expect(values.streams[0].config?.cursorField).toEqual([]);
+    expect(values.streams[1].config?.primaryKey).toEqual([["id"]]);
+  });
+
+  it("should not populate the source default cursor if the user-specified cursor field was removed during refresh", () => {
+    expect(2 + 2).toEqual(4);
+    // tbd expected behavior...
```
tealjulia, line 799: https://airbytehq-team.slack.com/archives/C042ABK94FR/p1670451834601009?thread_ts=1670360753.045569&cid=C042ABK94FR

Waiting for feedback on what the desired behavior is.  As it stands, if there is a source default cursor (but it isn't a source-defined cursor), we will populate it since the old cursor was removed and the property is now empty...

tealjulia:
Cleaned up the pieces seen earlier.  

Also, added an issue to tidy up the tests in this file #20233

airbyte-webapp/src/views/Connection/ConnectionForm/calculateInitialCatalog.ts
```diff
@@ -133,13 +181,29 @@ const calculateInitialCatalog = (
...
+          ? streamsWithBreakingFieldChanges.filter((streamTransformFromDiff) => {
+              return (
+                streamTransformFromDiff.streamDescriptor.name === nodeStream?.stream?.name &&
+                streamTransformFromDiff.streamDescriptor.namespace === nodeStream.stream?.namespace
+              );
+            })
+          : [];
+
+      // if we're in edit or readonly mode and the stream is not new, check for breaking changes then return
+      if (isNotCreateMode && !isStreamNew) {
```
tealjulia, line 203: I don't love that this check happens on `readonly` connections.  Presumably `readonly` connections wouldn't ever have a breaking change so should return immediately?  However looking at where `isNotCreateMode` is passed around, refactoring the code to pass the _actual_ mode is pretty hairy unfortunately.

It's not the only place we touch a `readonly` connection in this method (see `verifySourceDefinedProperties` being called on line 182 here) so maybe it's ok for now if not ideal.

airbyte-webapp/src/views/Connection/ConnectionForm/calculateInitialCatalog.ts
```diff
@@ -13,6 +16,49 @@ const getDefaultCursorField = (streamNode: SyncSchemaStream): string[] => {
...
+  for (const streamTransformation of breakingChangesByStream) {
+    // get all of the removed field paths for this transformation
+    const removedFieldPaths = streamTransformation.updateStream?.map((update) => update.fieldName);
+
+    if (!removedFieldPaths?.length) {
+      return nodeStream;
+    }
+
+    // if there is a primary key in the config, and any of its field paths were removed, we'll be clearing it
+    if (
```
tealjulia, line 36: We _only_ want to be overriding this if true, hence why it makes sense to use an if statement here rather than setting `clearPrimaryKey` as a boolean of this statement.  Otherwise we end up with `clearPrimaryKey = !!(!!nodeStream.config?.primaryKey?.length && ...` and things get less readable.

airbyte-webapp/src/views/Connection/ConnectionForm/calculateInitialCatalog.ts
```diff
@@ -13,6 +16,49 @@ const getDefaultCursorField = (streamNode: SyncSchemaStream): string[] => {
...
+      !!nodeStream.config?.primaryKey?.length &&
+      nodeStream.config?.primaryKey?.some((key) => removedFieldPaths.some((removedPath) => isEqual(key, removedPath)))
+    ) {
+      clearPrimaryKey = true;
+    }
```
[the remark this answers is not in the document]
tealjulia: Yes.  Because we _only_ want to override for the `true` case, not for `false` (which could mistakenly overwrite a previous item setting the variable as true), it would end up reading something like `clearPrimaryKey=!!(!!nodeStream.config?.primaryKey?.length && ....` which did not seem very readable.

airbyte-webapp/src/views/Connection/ConnectionForm/calculateInitialCatalog.ts
```diff
@@ -13,6 +16,49 @@ const getDefaultCursorField = (streamNode: SyncSchemaStream): string[] => {
...
+
+    // if there is a cursor field, and any of its field path was removed, we'll be clearing it
+    if (
+      !!nodeStream.config?.cursorField?.length &&
+      removedFieldPaths.some((removedPath) => isEqual(removedPath, nodeStream?.config?.cursorField))
+    ) {
+      clearCursorField = true;
+    }
+  }
```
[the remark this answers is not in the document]
tealjulia: Same answer here.  It is an intentional choice I made for readability.

edmundito (approved):

airbyte-webapp/src/views/Connection/ConnectionForm/calculateInitialCatalog.ts
```diff
@@ -13,6 +16,49 @@ const getDefaultCursorField = (streamNode: SyncSchemaStream): string[] => {
   return streamNode.config?.cursorField || [];
 };
 
+const clearBreakingFieldChanges = (nodeStream: SyncSchemaStream, breakingChangesByStream: StreamTransform[]) => {
+  if (!breakingChangesByStream?.length || !nodeStream.config) {
```
krishnaglick, line 20:
```suggestion
  if (!breakingChangesByStream.length || !nodeStream.config) {
```

airbyte-webapp/src/views/Connection/ConnectionForm/calculateInitialCatalog.ts
```diff
@@ -13,6 +16,49 @@ const getDefaultCursorField = (streamNode: SyncSchemaStream): string[] => {
...
+
+  let clearPrimaryKey = false;
+  let clearCursorField = false;
+
+  for (const streamTransformation of breakingChangesByStream) {
+    // get all of the removed field paths for this transformation
+    const removedFieldPaths = streamTransformation.updateStream?.map((update) => update.fieldName);
+
+    if (!removedFieldPaths?.length) {
+      return nodeStream;
```
krishnaglick, line 32: It doesn't seem like this state is possible, but `continue;` is likely the better choice.
```suggestion
      continue;
```
Maybe even add a datadog event.

krishnaglick (approved):

airbyte-webapp/src/views/Connection/ConnectionForm/calculateInitialCatalog.ts
tealjulia: Agree.  Returning here prevents any additional checks on the stream.  It _shouldn't_ be a code path that is ever hit, but this is safer!

PR #20208, opened 2022-12-07
Disable all broken connections when source is refreshed

alovew:
When one connection's source is refreshed, we want to check each connection that uses that same source and disable them where necessary.

Changes are:
- new ConfigRepository and ConnectionHandler methods to fetch all connections for a particular source id
- updates in the SchedulerHandler to iterate through all connections for a particular source and check their diffs

### DIFF
--- a/airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
+++ b/airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
@@ -865,6 +865,20 @@ public class ConfigRepository {
     return getStandardSyncsFromResult(connectionAndOperationIdsResult);
   }
 
+  public List<StandardSync> listConnectionsBySource(final UUID sourceId, final boolean includeDeleted) throws IOException {
+    final Result<Record> connectionAndOperationIdsResult = database.query(ctx -> ctx
+        .select(
+            CONNECTION.asterisk(),
+            groupConcat(CONNECTION_OPERATION.OPERATION_ID).separator(OPERATION_IDS_AGG_DELIMITER).as(OPERATION_IDS_AGG_FIELD))
+        .from(CONNECTION)
+        .leftJoin(CONNECTION_OPERATION).on(CONNECTION_OPERATION.CONNECTION_ID.eq(CONNECTION.ID))
+        .where(CONNECTION.SOURCE_ID.eq(sourceId)
+            .and(includeDeleted ? noCondition() : CONNECTION.STATUS.notEqual(StatusType.deprecated)))
+        .groupBy(CONNECTION.ID)).fetch();
+
+    return getStandardSyncsFromResult(connectionAndOperationIdsResult);
+  }
+
   private List<StandardSync> getStandardSyncsFromResult(final Result<Record> connectionAndOperationIdsResult) {
     final List<StandardSync> standardSyncs = new ArrayList<>();
 
--- a/airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
+++ b/airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
@@ -16,6 +16,7 @@ import io.airbyte.api.model.generated.CheckConnectionRead;
 import io.airbyte.api.model.generated.CheckConnectionRead.StatusEnum;
 import io.airbyte.api.model.generated.ConnectionIdRequestBody;
 import io.airbyte.api.model.generated.ConnectionRead;
+import io.airbyte.api.model.generated.ConnectionReadList;
 import io.airbyte.api.model.generated.ConnectionStatus;
 import io.airbyte.api.model.generated.ConnectionUpdate;
 import io.airbyte.api.model.generated.DestinationCoreConfig;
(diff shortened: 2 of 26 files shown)


airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -865,6 +865,20 @@ public List<StandardSync> listWorkspaceStandardSyncs(final UUID workspaceId, fin
     return getStandardSyncsFromResult(connectionAndOperationIdsResult);
   }
 
+  public List<StandardSync> listConnectionsBySource(final UUID sourceId, final boolean includeDeleted) throws IOException {
```
gosusnp, line 868: I just put up a PR with a more complete version of this: https://github.com/airbytehq/airbyte/pull/20264/files#diff-2d29b293778e290157be96e22294c2df1fdfa7c72a48961d03396776446deeddR852
However, my PR might take longer to merge due to breaking UI changes.

airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
```diff
@@ -383,30 +384,32 @@ public JobInfoRead cancelJob(final JobIdRequestBody jobIdRequestBody) throws IOE
...
-            CatalogConverter.toProtocol(currentAirbyteCatalog));
-    final boolean containsBreakingChange = containsBreakingChange(diff);
-    final ConnectionUpdate updateObject =
-        new ConnectionUpdate().breakingChange(containsBreakingChange).connectionId(discoverSchemaRequestBody.getConnectionId());
-    final ConnectionStatus connectionStatus;
-    if (shouldDisableConnection(containsBreakingChange, connectionRead.getNonBreakingChangesPreference(), diff)) {
-      connectionStatus = ConnectionStatus.INACTIVE;
-    } else {
-      connectionStatus = connectionRead.getStatus();
+    final ConnectionReadList connectionsForSource = connectionsHandler.listConnectionsForSource(discoverSchemaRequestBody.getSourceId(), false);
```
gosusnp, line 390: I would advocate for avoiding having handlers depend on other handlers. At the handler level, I am currently leaning towards using our DB abstraction directly until we introduce a better layer.
It might be worth checking with the team to agree on a direction moving forward here.

airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
```diff
@@ -383,30 +384,32 @@ public JobInfoRead cancelJob(final JobIdRequestBody jobIdRequestBody) throws IOE
...
+      ConnectionUpdate updateObject =
+          new ConnectionUpdate().breakingChange(containsBreakingChange).connectionId(connectionRead.getConnectionId());
+      ConnectionStatus connectionStatus;
+      if (shouldDisableConnection(containsBreakingChange, connectionRead.getNonBreakingChangesPreference(), diff)) {
+        connectionStatus = ConnectionStatus.INACTIVE;
+      } else {
+        connectionStatus = connectionRead.getStatus();
+      }
+      updateObject.status(connectionStatus);
+      connectionsHandler.updateConnection(updateObject);
```
gosusnp, line 408: This feels very broad, from this function, I expect we only modify syncCatalog and the breaking change flag, but this function will actually override the whole object. Should we introduce a more scoped update here?

airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
```diff
@@ -383,30 +384,32 @@ public JobInfoRead cancelJob(final JobIdRequestBody jobIdRequestBody) throws IOE
...
+      if (connectionRead.getConnectionId().equals(discoverSchemaRequestBody.getConnectionId())) {
+        discoveredSchema.catalogDiff(diff).breakingChange(containsBreakingChange).connectionStatus(connectionStatus);
```
gosusnp, line 410: Can we add some description to the function to highlight that we are modifying the input?

airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
alovew: This is actually a patch update even though the object is named ConnectionUpdate, so it is only modifying the breaking change field and status

airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
alovew: oh interesting, you mean use configPersistence directly instead of going through the connection handler? I may have misunderstood how we wanted to migrate this

airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
gosusnp: Yes, my point is that if we add other fields to the `updateObject` we could end up overriding fields by mistake.
I don't think it is a big issue currently, but this could lead to some surprises.

airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
gosusnp: Yes and I don't think we agreed on a strategy here.

airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
alovew: you mean adding other fields within this method here? we would only be updating other fields here if we explicitly stated them on the ConnectionUpdate object, but I think that would be expected

airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java
alovew: just added a comment where we're calling the function - do you think that works?

gosusnp: I'd say it should be part of the documentation of this function.

gosusnp (approved):

PR #20211, opened 2022-12-07
Bmoric/convert connection micronaut

benmoriceau:
## What
- Convert the connection Controller to Micronaut
- Switch the server beans initialization to eager because it has been observe that the first request could be slow because of the bean initialization.


airbyte-server/src/main/java/io/airbyte/server/Application.java
```diff
@@ -9,7 +9,12 @@
 public class Application {
 
   public static void main(final String[] args) {
-    Micronaut.run(Application.class, args);
+    Micronaut
+        .build(args)
+        // Lazy initialization can make the first requests to be slow
+        .eagerInitSingletons(true)
```
jdpgrailsdev, line 15: This may be a bit dangerous, as it will initialize all singletons, including those in transitive dependencies.  It's fine for now, but if we start to see issues, we may want to use the [Context annotation](https://docs.micronaut.io/latest/api/io/micronaut/context/annotation/Context.html) to explicitly mark beans as eager.  I'm not sure if that would work on a `@Controller` annotated class, so we may have to switch back to the JAX-RS annotations + `@Context`.

airbyte-server/src/main/java/io/airbyte/server/apis/ConnectionApiController.java
```diff
@@ -16,9 +16,13 @@
 import io.airbyte.server.handlers.ConnectionsHandler;
 import io.airbyte.server.handlers.OperationsHandler;
 import io.airbyte.server.handlers.SchedulerHandler;
+import io.micronaut.http.MediaType;
+import io.micronaut.http.annotation.Body;
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Post;
 import javax.ws.rs.Path;
```
jdpgrailsdev, line 23: This import can be removed.

jdpgrailsdev:
@benmoriceau Looks like there are some acceptance test failures:

```
Replacing default Chart.yaml and values.yaml with a test one
mv: cannot stat 'charts/airbyte/Chart.yaml': No such file or directory
```

Seems unrelated to this PR.

airbyte-server/src/main/java/io/airbyte/server/Application.java
benmoriceau: Done

airbyte-server/src/main/java/io/airbyte/server/apis/ConnectionApiController.java
benmoriceau: Done, I think that I need to reconfigure the save action plugin

jdpgrailsdev:
@benmoriceau Looks like there is now a conflict with this PR.

jdpgrailsdev:
@benmoriceau Is this an expected failure in the acceptance tests due to the lack of a proxy in k8s:

```
E1209 00:29:43.922386   24391 portforward.go:400] an error occurred forwarding 8001 -> 8001: error forwarding port 8001 to pod 28fa8b99acf11fc76ab9515622561aa0e900c27a7ce66bfa6292db0e6216adbe, uid : failed to execute portforward in network namespace "/var/run/netns/cni-692b2da6-7a9c-e09e-f80f-4f09e3178c66": failed to connect to localhost:8001 inside namespace "28fa8b99acf11fc76ab9515622561aa0e900c27a7ce66bfa6292db0e6216adbe", IPv4: dial tcp4 127.0.0.1:8001: connect: connection refused IPv6 dial tcp6 [::1]:8001: connect: connection refused 
```

If not, is this failure in any way related to the changes or just the usual failing acceptance tests on master?

benmoriceau:
> 



> @benmoriceau Is this an expected failure in the acceptance tests due to the lack of a proxy in k8s:
> 
> ```
> E1209 00:29:43.922386   24391 portforward.go:400] an error occurred forwarding 8001 -> 8001: error forwarding port 8001 to pod 28fa8b99acf11fc76ab9515622561aa0e900c27a7ce66bfa6292db0e6216adbe, uid : failed to execute portforward in network namespace "/var/run/netns/cni-692b2da6-7a9c-e09e-f80f-4f09e3178c66": failed to connect to localhost:8001 inside namespace "28fa8b99acf11fc76ab9515622561aa0e900c27a7ce66bfa6292db0e6216adbe", IPv4: dial tcp4 127.0.0.1:8001: connect: connection refused IPv6 dial tcp6 [::1]:8001: connect: connection refused 
> ```
> 
> If not, is this failure in any way related to the changes or just the usual failing acceptance tests on master?

@jdpgrailsdev Yes, it is expected since there is no proxy in kube. For the other failures, I need to merge the health PR since the fix was done there.

jdpgrailsdev (approved):
:shipit:

PR #20202, opened 2022-12-07
database migration to add column for field selection info

mfsiega-airbyte:
## What
Config database support for field/column selection info.

## How
Adds a column in the `connection` table.

This will be a map from `StreamDescriptor` -> `true/false` indicating whether the given stream has requested column selection.

See [spec](https://docs.google.com/document/d/1mCY9pi3iwWEzx7V2_--80E7tkmeWs_ckcLfqin1k_rA).

PR #20212, opened 2022-12-07
Add field selection info to the StandardSync persistence

mfsiega-airbyte:
## What
Include field selection in the standard sync persistence.

This is step 2 of 3 in backend changes to support new APIs for column selection.

Step 1 was https://github.com/airbytehq/airbyte/pull/20202 (database migration to add the column for this info) and Step 3 in the future will use this field to make a map from `StreamDescriptor` -> `true/false` whether column selection is enabled for that stream.

## How
- Add it to the model
- Make sure it gets written/read to/from the database

### DIFF
--- a/airbyte-config/config-models/src/main/resources/types/StandardSync.yaml
+++ b/airbyte-config/config-models/src/main/resources/types/StandardSync.yaml
@@ -42,6 +42,11 @@ properties:
     type: string
   catalog:
     existingJavaType: io.airbyte.protocol.models.ConfiguredAirbyteCatalog
+  fieldSelectionData:
+    type: object
+    description: A map of stream description to an indicator of whether field selection is enabled for that stream.
+    additionalProperties:
+      type: boolean
   status:
     type: string
     enum:
--- a/airbyte-config/config-models/src/main/resources/types/StandardSync.yaml
+++ b/airbyte-config/config-models/src/main/resources/types/StandardSync.yaml
@@ -44,7 +44,7 @@ properties:
     existingJavaType: io.airbyte.protocol.models.ConfiguredAirbyteCatalog
   fieldSelectionData:
     type: object
-    description: A map of stream description to an indicator of whether field selection is enabled for that stream.
+    description: A map of StreamDescriptor to an indicator of whether field selection is enabled for that stream.
     additionalProperties:
       type: boolean
   status:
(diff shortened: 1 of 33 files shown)


airbyte-config/config-models/src/main/resources/types/StandardSync.yaml
```diff
@@ -42,6 +42,11 @@ properties:
     type: string
   catalog:
     existingJavaType: io.airbyte.protocol.models.ConfiguredAirbyteCatalog
+  fieldSelectionData:
+    type: object
+    description: A map of stream description to an indicator of whether field selection is enabled for that stream.
```
alovew, line 47: can you change stream description to StreamDescriptor? it's a special object for uniquely identifying streams.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/StandardSyncPersistence.java
```diff
@@ -187,6 +187,7 @@ private void writeStandardSync(final StandardSync standardSync, final DSLContext
           .set(CONNECTION.DESTINATION_ID, standardSync.getDestinationId())
           .set(CONNECTION.NAME, standardSync.getName())
           .set(CONNECTION.CATALOG, JSONB.valueOf(Jsons.serialize(standardSync.getCatalog())))
+          .set(CONNECTION.FIELD_SELECTION_DATA, JSONB.valueOf(Jsons.serialize(standardSync.getFieldSelectionData())))
```
alovew, line 190: will we need to add this for connection create as well? not sure of the UX here

airbyte-config/config-models/src/main/resources/types/StandardSync.yaml
mfsiega-airbyte: Done.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/StandardSyncPersistence.java
mfsiega-airbyte: Good call, not sure how I missed this - added.

alovew (approved):

pmossman (approved):
Nice, this looks good to me!

PR #20214, opened 2022-12-08
convert WorkspaceApiController to micronaut

### LINKED ISSUE #20148: convert WorkspaceApiController to Micronaut
part of #13274

colesnodgrass:
- fix #20148
- convert `WorkspaceApiController` and associated dependencies (where possible) to Micronaut
    - add `@Singleton` to `DestinationHandler` class
    - add `@Singleton` to `SourceHandler` class
    - add `@Singleton` to `WorkspaceHandler` class
    - add `workspace.root` to `application.yml` (necessary to use existing `@Singleton` classes)
    - add `temporal.*` to `application.yml` (necessary to use existing `@Singleton` temporal classes)
    - remove `WorkspaceApiBinder`
    - remove `WorkspaceApiFactory`
- add `/api/v1/workspaces` to nginx config
    - unsure if this is necessary or not
- add micronaut jaxrs dependencies in an attempt to utilize jaxrs annotations
    - I cannot get this to work, I still _must_ annotate the Controller class with the Micronaut annotations


jdpgrailsdev (approved):
:shipit:

airbyte-server/src/main/java/io/airbyte/server/config/WorkspaceApiFactory.java
```diff
@@ -0,0 +1,87 @@
...
+import io.micronaut.context.annotation.Factory;
+import jakarta.inject.Named;
+import jakarta.inject.Singleton;
+import org.jooq.DSLContext;
+
+@Factory
+public class WorkspaceApiFactory {
+
+  @Singleton
+  EnvConfigs envConfigs() {
```
benmoriceau, line 33: we shouldn't use the `EnvConfig` but the micronaut application values like in https://github.com/airbytehq/airbyte/pull/20211/files#diff-ac6719193135419f98d16e4e7ae96685864ee34a7302ac7ce428dc142736b17aR40

airbyte-server/src/main/java/io/airbyte/server/config/WorkspaceApiFactory.java
```diff
@@ -0,0 +1,87 @@
...
+import io.airbyte.server.scheduler.TemporalEventRunner;
+import io.airbyte.validation.json.JsonSchemaValidator;
+import io.airbyte.workers.helper.ConnectionHelper;
+import io.micronaut.context.annotation.Factory;
+import jakarta.inject.Named;
+import jakarta.inject.Singleton;
+import org.jooq.DSLContext;
+
+@Factory
+public class WorkspaceApiFactory {
```
benmoriceau, line 30: We shouldn't have factory per controller since those will be share across controller. I think that we should have branched out from the connection branch, sorry about that.

airbyte-server/src/main/java/io/airbyte/server/apis/WorkspaceApiController.java
```diff
@@ -15,63 +15,76 @@
...
+@Controller("/api/v1/workspaces")
 public class WorkspaceApiController implements WorkspaceApi {
 
   private final WorkspacesHandler workspacesHandler;
 
+  public WorkspaceApiController(final WorkspacesHandler workspacesHandler) {
+    this.workspacesHandler = workspacesHandler;
+  }
+
+  @Post("/create")
```
benmoriceau, line 31: Should we add the payload mediatype?

airbyte-proxy/nginx-auth.conf.template
```diff
@@ -37,6 +37,23 @@ http {
         auth_basic off;
       }
     }
+
+    location /api/v1/workspaces {
```
benmoriceau, line 41: Should this be `location ~ ^/api/v1/workspaces/.*` instead.

airbyte-server/src/main/java/io/airbyte/server/apis/WorkspaceApiController.java
colesnodgrass: The `@Controller` annotation defaults to `application/json`, so we only need to specify it if we're not using json.

airbyte-proxy/nginx-auth.conf.template
colesnodgrass: This doesn't need to be a regex, from their [docs](https://docs.nginx.com/nginx/admin-guide/web-server/web-server/#locations)
> To find the location that best matches a URI, NGINX Plus first compares the URI to the locations with a prefix string. It then searches the locations with a regular expression.

> Higher priority is given to regular expressions, unless the ^~ modifier is used. Among the prefix strings NGINX Plus selects the most specific one (that is, the longest and most complete string).

This non-regex version will still match `/api/v1/workspaces/list`, `/api/v1/workspaces/get`, `/api/v1/workspaces/*`.

airbyte-server/src/main/java/io/airbyte/server/config/WorkspaceApiFactory.java
colesnodgrass: I think I already handled this, double-checking...

colesnodgrass: ...and only partially handled.

jdpgrailsdev: Sorry...I missed this in my review.  We should definitely avoid using `EnvConfig` and convert any code that does use it to get the configuration via injection.

colesnodgrass:
@jdpgrailsdev @benmoriceau - updated and aligned with other work that Benoit had done, please re-review

benmoriceau (approved):

PR #20230, opened 2022-12-08
Source Jira: correct description and example for additional fields to avoid misleading users

### LINKED ISSUE #13961: Source Jira: custom fields not returned
## Environment
- **Airbyte version**:  All versions
- **OS Version / Instance**: AWS EC2
- **Deployment**:  Docker
- **Source Connector and version**: JIRA
- **Destination Connector and version**: Redshift
- **Step where error happened**: Sync Job
## Current Behavior
Custom Fields do not appear in the issue stream

## Expected Behavior
Custom fields appear in the issue stream

## Steps to Reproduce
1. Setup custom fields on source config page. customfield_00000, cf[00000], customfield
2. Execute sync
3. Custom fields are not returned in issue stream. I can see these custom fields in the changelog regardless of the custom field config on source. Expectation is that the current value is returned for the customfield in its own field

darynaishchenko:
correct description and example for additional fields to avoid misleading users. Now it's array of fields names, not their ids

### DIFF
--- a/airbyte-integrations/connectors/source-jira/source_jira/spec.json
+++ b/airbyte-integrations/connectors/source-jira/source_jira/spec.json
@@ -47,8 +47,8 @@
         "items": {
           "type": "string"
         },
-        "description": "List of additional fields to include in replicating issues.",
-        "examples": ["customfield_10096", "customfield_10071"]
+        "description": "List of additional fields names to include in replicating issues.",
+        "examples": ["Approvers", "Change completion date"]
       },
       "expand_issue_changelog": {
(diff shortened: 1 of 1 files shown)


davydov-d (changes requested):
Please bump version in Dockerfile and update the changelog

grubberr:
@darynaishchenko

I think we can just close this issue (not merge) because in this PR https://github.com/airbytehq/airbyte/pull/20128 I have removed `additional_fields` param from spec. In new implemenation source-jira just fetch all possible issue fields.

PR #20231, opened 2022-12-08
🪟 🔧 Allow overwriting features in dev by env variables

timroes:
## What

Allows any `FeatureItem` to be overwritten in development by a env variable with the name `REACT_APP_FEATURE_{ID}`, e.g. one can have the following in the `.env.development` file:

```
REACT_APP_FEATURE_ALLOW_SYNC=false
REACT_APP_FEATURE_ALLOW_CHANGE_DATA_GEOGRAPHIES=true
```

To disable syncs and allow geography changes.

This is only enabled during development, since this isn't meant to be misused accidentally in environments instead of the regular feature chain (default features, workspace, users).

### DIFF
--- a/airbyte-webapp/src/hooks/services/Feature/FeatureService.test.tsx
+++ b/airbyte-webapp/src/hooks/services/Feature/FeatureService.test.tsx
@@ -181,6 +181,13 @@ describe("Feature Service", () => {
         expect(getFeature(FeatureItem.AllowSync)).toBe(false);
         expect(getFeature(FeatureItem.AllowChangeDataGeographies)).toBe(true);
       });
+
+      it("should not overwrite in a non dev environment", () => {
+        (process.env.NODE_ENV as string) = "productions";
+        const getFeature = (feature: FeatureItem) => renderHook(() => useFeature(feature), { wrapper }).result.current;
+        expect(getFeature(FeatureItem.AllowSync)).toBe(true);
+        expect(getFeature(FeatureItem.AllowChangeDataGeographies)).toBe(false);
+      });
     });
   });
 
--- a/airbyte-webapp/src/hooks/services/Feature/FeatureService.test.tsx
+++ b/airbyte-webapp/src/hooks/services/Feature/FeatureService.test.tsx
@@ -162,6 +162,26 @@ describe("Feature Service", () => {
       rerender({ overwrite: undefined });
       expect(result.current.sort()).toEqual([FeatureItem.AllowDBTCloudIntegration, FeatureItem.AllowSync]);
     });
+
+    describe("env variable overwrites", () => {
+      beforeEach(() => {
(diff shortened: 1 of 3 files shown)


edmundito (commented):
Looking good, just a couple of questions and... should we add a test for this as well?

airbyte-webapp/src/hooks/services/Feature/types.tsx
```diff
@@ -14,4 +14,4 @@ export enum FeatureItem {
   AllowSyncSubOneHourCronExpressions = "ALLOW_SYNC_SUB_ONE_HOUR_CRON_EXPRESSIONS",
 }
 
-export type FeatureSet = Record<FeatureItem, boolean>;
+export type FeatureSet = Partial<Record<FeatureItem, boolean>>;
```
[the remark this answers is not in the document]
timroes: It would actually have been more correct already in the past that this should be a `Partial` in the other places as well. They actually just pass us partial objects here and not the full enum. But since in the other places we usually cast it from e.g. launchdarkly this was not noticed. This type reflects the actual type we expect better.

airbyte-webapp/src/hooks/services/Feature/FeatureService.test.tsx
```diff
@@ -162,6 +162,26 @@ describe("Feature Service", () => {
...
     });
+
+    describe("env variable overwrites", () => {
+      beforeEach(() => {
+        process.env.REACT_APP_FEATURE_ALLOW_SYNC = "false";
+        process.env.REACT_APP_FEATURE_ALLOW_CHANGE_DATA_GEOGRAPHIES = "true";
+      });
+
+      afterEach(() => {
+        (process.env.NODE_ENV as string) = "test";
```
timroes, line 173: :information_source: Need to cast this, since this is by default `readonly` so we can't overwrite it.

krishnaglick (commented):
Have not tested this locally yet.

airbyte-webapp/src/hooks/services/Feature/FeatureService.test.tsx
```diff
@@ -162,6 +162,33 @@ describe("Feature Service", () => {
...
+
+      it("should allow overwriting it in dev", () => {
+        (process.env.NODE_ENV as string) = "development";
+        const getFeature = (feature: FeatureItem) => renderHook(() => useFeature(feature), { wrapper }).result.current;
+        expect(getFeature(FeatureItem.AllowSync)).toBe(false);
+        expect(getFeature(FeatureItem.AllowChangeDataGeographies)).toBe(true);
+      });
+
+      it("should not overwrite in a non dev environment", () => {
+        (process.env.NODE_ENV as string) = "productions";
```
edmundito, line 186: Nit: It's probably `production` without the s, but it works for the test.

edmundito (approved):
🎉

PR #20235, opened 2022-12-08
Break inheritance dependency on java.lang.Process

jdpgrailsdev:
## What
* Remove the direct inheritance dependency on `java.lang.Process` to avoid issues caused by partial implementation

## How
* Remove extension from `java.lang.Process`
* Add ability to create a `java.lang.Process` implementation from existing "process" objects

This PR is inspired by a discussion [following an incident](https://github.com/airbytehq/airbyte/pull/20028) where the use of `.pid()` in `KubePodProcess` lead to failures, as the parent class (`java.lang.Process`) throws exceptions for certain non-abstract, public methods that are not implemented by the child class.  More broadly speaking, the classes that we have identified as "processes" are not really processes in the Java sense (e.g. they are not native processes provided by an operating system).  This change will break that dependency so that changes made to these classes will not be exposed to the problem outlined above (e.g. calling a method on the parent class that is not and cannot really be implemented by the child class due to the fact that it is not a true native process).  This is analogous to extending `java.lang.Thread` instead of implementing `Runnable` or `Callable`.

## Recommended reading order
1. `KubePod.java`
2. `KubePodProcess.java`
3. `AsyncOrchestratorPodProcess.java`
4. `KubeProcessFactory.java`

## Tests

* All unit tests pass
* All Kube acceptance tests pass locally
* Project builds locally

### DIFF
--- a/airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubePodProcess.java
+++ b/airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubePodProcess.java
@@ -101,7 +101,7 @@ import org.slf4j.MDC;
 // it is required for the connectors
 @SuppressWarnings("PMD.AvoidPrintStackTrace")
 // TODO(Davin): Better test for this. See https://github.com/airbytehq/airbyte/issues/3700.
-public class KubePodProcess extends Process implements KubePod {
+public class KubePodProcess implements KubePod {
 
   private static final Configs configs = new EnvConfigs();
 
@@ -623,21 +623,6 @@ public class KubePodProcess extends Process implements KubePod {
     });
   }
 
-  @Override
-  public OutputStream getOutputStream() {
-    return this.stdin;
-  }
-
-  @Override
-  public InputStream getInputStream() {
-    return this.stdout;
-  }
-
-  @Override
-  public InputStream getErrorStream() {
-    return this.stderr;
-  }
-
   /**
    * Waits for the Kube Pod backing this process and returns the exit value after closing resources.
    */
@@ -650,15 +635,6 @@ public class KubePodProcess extends Process implements KubePod {
     }
 
     return exitValue();
-  }
-
-  /**
-   * Waits for the Kube Pod backing this process and returns the exit value until a timeout. Closes
-   * resources if and only if the timeout is not reached.
-   */
-  @Override
-  public boolean waitFor(final long timeout, final TimeUnit unit) throws InterruptedException {
-    return super.waitFor(timeout, unit);
   }
 
   /**
@@ -677,11 +653,6 @@ public class KubePodProcess extends Process implements KubePod {
       close();
       LOGGER.info(prependPodInfo("Destroyed Kube process.", podNamespace, podName));
     }
-  }
-
-  @Override
-  public Info info() {
-    return new KubePodProcessInfo(podDefinition.getMetadata().getName());
   }
 
   private Container getMainContainerFromPodDefinition() {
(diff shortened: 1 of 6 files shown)


cgardens (approved):
lgtm!

We have a parallel antipattern in our docker stack (see: `DockerProcessFactory`). I was surprised that this change was possible without having to touch it. Maybe that says something about code reuse about K8s and docker impls. Don't know if it is worth doing this in `DockerProcessFactory` too. This PR is good though and should not be blocked on that.

davinchia (commented):
@jdpgrailsdev just to make sure I understand, this change makes it so that we continue to use the Process interface (which we want) while restricting consumer of Airbyte process implementation from accessing Process interface internal information (which we don't want).

Is that right?

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubePodProcess.java
```diff
@@ -652,15 +637,6 @@ public int waitFor() throws InterruptedException {
     return exitValue();
   }
 
-  /**
-   * Waits for the Kube Pod backing this process and returns the exit value until a timeout. Closes
-   * resources if and only if the timeout is not reached.
-   */
-  @Override
-  public boolean waitFor(final long timeout, final TimeUnit unit) throws InterruptedException {
```
davinchia, line 660: what happened to this?

jdpgrailsdev:
> @jdpgrailsdev just to make sure I understand, this change makes it so that we continue to use the Process interface (which we want) while restricting consumer of Airbyte process implementation from accessing Process interface internal information (which we don't want).
> 
> Is that right?

@davinchia Exactly correct.  We still want to represent it as a process, without the extra baggage of extending the `Process` class, which results in an incomplete implementation.

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubePodProcess.java
jdpgrailsdev: @davinchia Our implementation just called the super class's implementation, which is in the `java.lang.Process` class.  Therefore, once the inheritance was removed, we no longer need this as we can just allow the implementation in `java.lang.Process` to stand in the new object we return from the `toProcess()` method.

jdpgrailsdev:
> lgtm!
> 
> We have a parallel antipattern in our docker stack (see: `DockerProcessFactory`). I was surprised that this change was possible without having to touch it. Maybe that says something about code reuse about K8s and docker impls. Don't know if it is worth doing this in `DockerProcessFactory` too. This PR is good though and should not be blocked on that.

@cgardens I'll take a look at `DockerProcessFactory`.  I did a scan for classes that extend `java.lang.Process` and didn't see any other hits than `KubePodProcess`, so I'll take another pass to see what I missed and open a separate PR if I find anything related to this type of change.

jdpgrailsdev:
@cgardens I took a look at `DockerProcessFactory` and it is doing the correct thing already.  Instead of creating a new instance of a class that we created that extends `java.lang.Process`, it uses `java.lang.ProcessBuilder` to create a new `java.lang.Process` instance (the correct way to create processes in Java).  Therefore, I don't believe that we have the same risk with the `DockerProcessFactory` that we have with the `KubePodProcess`/`KubeProcessFactory` classes.

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubePodProcess.java
davinchia: got it!

davinchia (approved):

PR #20253, opened 2022-12-08
Use existing status when no breaking change

benmoriceau:
## What
use the existing schema instead of ACTIVE when there is no breaking change.

Address: https://github.com/airbytehq/oncall/issues/1172


gosusnp (approved):

jdpgrailsdev (approved):
:shipit:

alovew (approved):

PR #20259, opened 2022-12-08
API changes to support configuring column selection

mfsiega-airbyte:
## What
API changes to support configuring column selection.

See [spec](https://docs.google.com/document/d/1mCY9pi3iwWEzx7V2_--80E7tkmeWs_ckcLfqin1k_rA).

## How
- Introduce new `fieldSelectionEnabled` and `selectedFields` parameters.
- Consume these params to only persist the selected fields in the configured catalog.
- Use the persisted data to re-construct the selected fields for web backend endpoint.

TODO: expand unit testing, send e2e test in future PR.

## Recommended reading order
1. `airbyte-api/src/main/openapi/config.yaml` for API changes.
2. `airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java` for most of how the new fields are consumed.
3. Everything else.


airbyte-api/src/main/openapi/config.yaml
```diff
@@ -3744,7 +3744,24 @@ components:
...
+          description: Whether field selection should be enabled. If this is true, only the properties in `selectedFields` will be included.
+          type: boolean
+        selectedFields:
+          description: Paths to the fields that will be included. This must be set if `fieldSelectedEnabled` is set. An empty list indicates that no properties will be included.
+          type: array
+          items:
+            $ref: "#/components/schemas/SelectedFieldInfo"
+    SelectedFieldInfo:
+      type: object
+      description: Path to a field/column/property in a stream to be selected.
```
pmossman, line 3759: Maybe an example in this description could help explain what a "path" is, I think it's very unclear without context on nested fields

airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
```diff
@@ -111,20 +162,62 @@ public static io.airbyte.protocol.models.ConfiguredAirbyteCatalog toProtocolKeep
...
   }
 
+  /**
+   * Generate the map from StreamDescriptor to indicator of whether field selection is enabled for
+   * that stream.
+   *
+   * @param syncCatalog the catalog
+   * @return the map as a FieldSelectionData object
+   */
+  public static FieldSelectionData getFieldSelectionData(AirbyteCatalog syncCatalog) {
```
pmossman, line 201: nit: `final AirbyteCatalog syncCatalog`

airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
```diff
@@ -111,20 +162,62 @@ public static io.airbyte.protocol.models.ConfiguredAirbyteCatalog toProtocolKeep
...
+    if (syncCatalog == null) {
+      return null;
+    }
+    final var fieldSelectionData = new FieldSelectionData();
+    syncCatalog.getStreams().stream().forEach((streamAndConfig) -> {
```
pmossman, line 206: nit: I've started preferring a `for (final StreamAndConfig streamAndConfig : syncCatalog.getStreams()) {` style instead of Java Streams for this sort of thing. I tend to only use Java Streams if I'm using collectors or maps, but totally fine to leave it as-is if you prefer

airbyte-api/src/main/openapi/config.yaml
```diff
@@ -3744,7 +3744,24 @@ components:
           description: Alias name to the stream to be used in the destination
           type: string
         selected:
+          description: If this is true, the stream is selected with all of its properties.
           type: boolean
+        fieldSelectionEnabled:
+          description: Whether field selection should be enabled. If this is true, only the properties in `selectedFields` will be included.
```
alovew, line 3750: should we say 'included in the configured catalog'? maybe that's already clear, though - just a suggestion

airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
```diff
@@ -30,7 +42,30 @@ private static io.airbyte.api.model.generated.AirbyteStream toApi(final io.airby
...
+      final JsonNode properties = streamSchema.findValue("properties");
+      if (properties.isObject()) {
+        ((ObjectNode) properties).retain(selectedFieldNames);
+      } else {
+        throw new JsonValidationException("Requested field selection but no properties node found");
+      }
+    }
     return new io.airbyte.protocol.models.AirbyteStream()
         .withName(stream.getName())
         .withJsonSchema(stream.getJsonSchema())
```
pmossman, line 71: is this `stream.getJsonSchema()` a reference to the `streamSchema` variable that you've define and modified? Just making sure we're actually setting the updated schema value in the return value here.

pmossman (approved):
Looks good @mfsiega-airbyte! Left a few small comments but nothing blocking. Saw the TODOs around supporting nested fields, looks like the change there should be pretty small which is great

airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
```diff
@@ -30,7 +42,30 @@ private static io.airbyte.api.model.generated.AirbyteStream toApi(final io.airby
...
+        if (selectedFieldInfo.getFieldPath().size() > 1) {
+          // TODO(mfsiega-airbyte): support nested fields.
+          throw new UnsupportedOperationException("Nested field selection not supported");
+        }
+      }
+      // Only include the selected fields.
+      List<String> selectedFieldNames = config.getSelectedFields().stream().map((field) -> field.getFieldPath().get(0)).collect(Collectors.toList());
+      final JsonNode properties = streamSchema.findValue("properties");
+      if (properties.isObject()) {
+        ((ObjectNode) properties).retain(selectedFieldNames);
```
alovew, line 64: what if there's a selected field that doesn't appear in properties? i think i need to pull this down to run locally to fully understand all this logic and what order stuff happens, but wanted to comment just in case we're missing something

airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
```diff
@@ -111,20 +162,62 @@ public static io.airbyte.protocol.models.ConfiguredAirbyteCatalog toProtocolKeep
...
+      final boolean fieldSelectionEnabled =
+          streamAndConfig.getConfig().getFieldSelectionEnabled() == null ? false : streamAndConfig.getConfig().getFieldSelectionEnabled();
+      fieldSelectionData.setAdditionalProperty(streamDescriptorToStringForFieldSelection(streamDescriptor), fieldSelectionEnabled);
+    });
+    return fieldSelectionData;
+  }
+
+  // Return a string representation of a stream descriptor that's convenient to use as a key for the
+  // field selection data.
+  private static String streamDescriptorToStringForFieldSelection(final StreamDescriptor streamDescriptor) {
```
mfsiega-airbyte, line 219: @alovew wanted to get your input here - As far as I can tell the `additionalProperties` construct needs to map `String` -> `Something`. I initially used the toString method but it's a bit clunky and I figured this has the same uniqueness and would be easier to deal with in logs/db queries/etc.

Is there some existing other standard on using StreamDescription like this?

alovew: unfortunately I think this is implemented at least one other time (possibly more) in various places - probably something we should clean up at some point but I think it's ok if we go this way for now, but I'm not sure we have to:

I'm a little confused by the additional properties thing, I didn't notice this in the last PR. normally our objects look like this:

```
  schedule:
    type: object
    required:
      - timeUnit
      - units
    additionalProperties: false
    properties:
      timeUnit:
        type: string
        enum:
          - minutes
          - hours
          - days
          - weeks
          - months
      units:
        type: integer
```

but this one looks like this:
```
  fieldSelectionData:
    type: object
    description: A map of StreamDescriptor to an indicator of whether field selection is enabled for that stream.
    additionalProperties:
      type: boolean
```

I thought that to declare objects in json schema you had to have a properties field that nested the properties. and I think we can make a StreamDescriptor object one of the properties that is nested in the object

mfsiega-airbyte: This one is indeed a bit different, because it's mapping arbitrary keys to bool values. In other words if we declared a StreamDescriptor object as a property, that could just represent a single stream. We could do an array of (StreamDescriptor, bool)-pairs, but having proper key-value lookup felt worthwhile.

I find it a bit surprising that a map isn't a first-class feature the way arrays are (maybe there's a good reason) but this seems to be the recommended OpenAPI way to do this: https://swagger.io/docs/specification/data-models/dictionaries/

(Unless I've misinterpreted and there's something better? which is very possible)

alovew: Ooh I see now, thank you for explaining! ok yes, this makes sense, and I think it's ok to add that stream descriptor -> string method here for now

airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
mfsiega-airbyte: It's indeed modifying by reference (or by pointer, if you're a C++ user and a pedant) so it's all the same, but I think it isn't being used often enough to justify a local variable.

airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
mfsiega-airbyte: I'm not so attached to Java Streams :)

airbyte-api/src/main/openapi/config.yaml
mfsiega-airbyte: I tried to clarify this a bit; but I think what's called for is a proper snippet of documentation on nested fields in the API, since we use the same convention for cursors and primary keys without any explanation (and it's indeed confusing).

airbyte-api/src/main/openapi/config.yaml
mfsiega-airbyte: Done.

airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
mfsiega-airbyte: Good question. We could tolerate this but I think it's better to throw an error; adding in the latest revision.

alovew (approved):

PR #13400, opened 2022-06-02
Sweep old scheduler code

lmossman:
## What
Resolves https://github.com/airbytehq/airbyte/issues/8445

This PR removes the old scheduler app and all of its related now-obsolete code.

## How
Delete the scheduler app and related classes, remove the non-new-scheduler conditional branches from the code, and remove classes/methods that were only called in those branches.

An airbyte-cloud PR to update the usages of the constructors that were changed in this PR will be coming soon.

## Recommended reading order
Almost all of the changes in this PR are just deleting the SchedulerApp code, its related classes, and airbyte-scheduler deployment code. 
The only additions are changes to add a few test cases that weren't testing the new-scheduler side of the conditionals, and renaming of things like "scheduler v2" or "new scheduler"

PR #20261, opened 2022-12-08
fix: remove unused CONTAINER_ORCHESTRATOR_ENABLED var

c-p-b:
## What
The behavior that this environment variable affects was removed in [this pr](https://github.com/airbytehq/airbyte/pull/13400/files). Therefore, this environment variable is no longer used and can be removed entirely.

## How
Remove all references to the environment variable from code and charts


airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
```diff
@@ -1043,11 +1042,6 @@ public Set<Integer> getTemporalWorkerPorts() {
     return Arrays.stream(ports.split(",")).map(Integer::valueOf).collect(Collectors.toSet());
   }
 
-  @Override
```
benmoriceau, line 1046: I think that is is missing an update to the `Configs.java` file as well

benmoriceau (approved):

airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
cpdeethree: 👍

rushil-somwanshi:
@cpdeethree Faced an error due to this PR where installation failed with the following error:

Error: INSTALLATION FAILED: template: airbyte/templates/env-configmap.yaml:67:42: executing "airbyte/templates/env-configmap.yaml" at <.Values.worker.containerOrchestrator.image>: nil pointer evaluating interface {}.image

Fixed it by adding back the following code in values.yaml file
`
worker.containerOrchestrator.image: ""
`

git-phu:
this isn't right
container orchestrator is required for cloud and is used here: https://github.com/airbytehq/airbyte/blob/f76833e2e02c661141631fcab3e2eaabe5b19e88/airbyte-workers/src/main/java/io/airbyte/workers/config/ContainerOrchestratorConfigBeanFactory.java#L46-L47

Issue #19850, opened 2022-11-28
web_backend/v1/connections/list should allow filtering by source or destination


krishnaglick:
There are several places in the frontend where we do a full connections list call _just_ to filter connections against sources or destinations.

Ideally we can provide a `workspaceId` and a `sourceId` or a `destinationId` to get a list of connections associated with that source or destination.

gosusnp:
Is the response the same? If yes, could we have optional filters on `connections/list`?
For example, `connections/list?destination_ids=A&destination_ids=B&destination_ids=C`? (querystring param not necessarily the actual syntax)

krishnaglick:
Clarified the ticket, we'll add filtering by source or destination id to the connections list endpoint.

dizel852:
Need to create separate FE issue

krishnaglick:
Closing this as completed since the backend work was done.

PR #20264, opened 2022-12-08
Improve performance of list connection operations

### LINKED ISSUE #19849: Trim Connection List Endpoint
The connections/list endpoint has some fields that are unnecessary for the frontend. To improve performance we should slim down this endpoint to be closer to _just_ what the frontend needs.

1. Remove `sourceId` and `destinationId` from the response since they are not used for the primary use case of this endpoint, and can be found in the attached source and destination objects.
2. Scope down the source portion of the response to only `sourceId`, `sourceName`, `name`, and `icon`.
3. Scope down the destination portion of the response to only `destinationId`, `destinationName`, `name`, and `icon`.
4. **The connectionConfiguration should be removed from the source and destination objects.**
5. [LOW] Remove `latestSyncJobCreatedAt` from the response. It's only used on Source and Destination list pages. This should be done after [getEntityTableData](https://github.com/airbytehq/airbyte/issues/19852) is migrated to the backend.

Depends on #20264, need to create PR against the branch in this PR.

gosusnp:
## What

Backend:
* Enable source/destination when querying connections
* Trim the response to contain what is being used by the FE

Resolves https://github.com/airbytehq/airbyte/issues/19849, https://github.com/airbytehq/airbyte/issues/19850, and https://github.com/airbytehq/airbyte/issues/20667

## How

Backend:
* Update API payload and ConfigRepository querying to support source/destination filtering
* Add new object for trimmed Source/Destination for the API response
* Removed a redundant db query to fetch schema changes

## Recommended reading order
1. ConfigRepository and WebBackendConnectionHandler contain most of the backend changes

## 🚨 User Impact 🚨
Faster load times


gosusnp:
Note for the backend, we're running into some fairly coupled interfaces changes here, the idea is to add the frontend to this PR to avoid the extra API changes back and forth.

alovew (approved):

benmoriceau (approved):

krishnaglick:
Vlad is taking the FE work here. We may need to add sourceDefinitionId and destinationDefinitionId back to the source and destination objects respectively, but waiting for Vlad to confir,

krishnaglick:
I believe this PR should cover https://github.com/airbytehq/airbyte/issues/19850 too? Correct me if I'm wrong please!

gosusnp:
@krishnaglick, yes both https://github.com/airbytehq/airbyte/issues/19849 and https://github.com/airbytehq/airbyte/issues/19850 are mentioned in the description.

gosusnp:
@krishnaglick, @dizel852, I added `sourceDefinitionId` to `SourceSnippetRead` and `destinationDefinitionId` to `DestinationSnippetRead`.
Let me know if anything else is missing.

krishnaglick (approved):
Tested locally, LGTM!

PR #20267, opened 2022-12-08
🪟 🎉 Column selection UI

joeykmh:
## What
Closes https://github.com/airbytehq/airbyte-cloud/issues/3423

Adds the column selection UI to the stream table:

https://user-images.githubusercontent.com/7550957/206715696-8cb12c27-0559-48d9-ae9f-5b72dd1130cd.mov

## How
To support column selection, we are adding two fields to the `AirbyteStreamConfiguration` object:

```ts
export interface AirbyteStreamConfiguration {
  syncMode: SyncMode;
  cursorField?: string[];
  destinationSyncMode: DestinationSyncMode;
  primaryKey?: string[][];
  aliasName?: string;
  selected?: boolean;
  fieldSelectionEnabled?: boolean;        // <-- new field!
  selectedFields?: SelectedFieldInfo[];   // <-- new field!
}
```

In the stream table, there is a new column - "Sync" - that displays a `<Switch />` for each field in the stream:
Screenshot 2022-12-09 at 2 50 02 PM

Nested fields cannot be individually selected/deselected, so their `<Switch />` is synced with their parents' value.

## Recommended reading order
1. `StreamFieldTable.tsx`, `FieldRow.tsx`, `FieldHeader.tsx` - where the new UI is actually rendered
2. `CatalogSection.tsx` - where the callbacks manipulating the new config object fields are located
3. The rest

## 🚨 User Impact 🚨
These changes are currently behind an experiment flag (`"connection.columnSelection"`) that is off by default (and therefore off in OSS). The web API is already accepting the new field selection configuration (https://github.com/airbytehq/airbyte/pull/20259) but not yet filtering the fields. Once we test the end to end column feature selection is working properly, we will roll out the feature out to all users.


airbyte-webapp/src/core/domain/catalog/traverseSchemaToField.test.ts
```diff
@@ -0,0 +1,106 @@
+import { AirbyteJSONSchema } from "core/jsonSchema";
+
+import { traverseSchemaToField } from "./traverseSchemaToField";
```
josephkmh, line 3: These are just some basic tests to cover this method, since there were none before.

airbyte-webapp/src/hooks/services/Feature/constants.ts
```diff
@@ -1,10 +1,16 @@
...
   FeatureItem.AllowUploadCustomImage,
   FeatureItem.AllowSyncSubOneHourCronExpressions,
 ];
+
+export const defaultCloudFeatures = [
+  FeatureItem.AllowOAuthConnector,
+  FeatureItem.AllowSync,
+  FeatureItem.AllowChangeDataGeographies,
+  FeatureItem.AllowDBTCloudIntegration,
+];
```
josephkmh, line 16: Perhaps unrelated to this PR, but I think this is more intuitive for our defaults for `FeatureService`.

airbyte-webapp/src/components/connection/CatalogTree/styles.tsx
```diff
@@ -20,3 +20,13 @@ export const ArrowCell = styled(HeaderCell)`
...
+
+export const SyncHeaderContainer = styled.span`
+  padding-left: 34px;
+  display: inline-block;
+  min-width: 62px;
+`;
+
+export const SyncCheckboxContainer = styled.div`
+  padding-left: 24px;
+`;
```
josephkmh, line 32: Kind of nasty hard-coded values 😕 but since this table won't be around long anyway, I did not spend much time here.

airbyte-webapp/src/components/connection/DestinationNamespaceModal/ExampleSettingsTable.module.scss
```diff
@@ -1,50 +1,52 @@
...
+@use "scss/colors";
+@use "scss/variables";
+
+.exampleSettingsTable {
+  border-spacing: 0;
+  border-collapse: separate;
+
+  tr {
+    th,
+    td {
```
josephkmh, line 10: So I merged this in from master, and apparently my prettier settings conflict with whatever was there (`th, td` got moved to two lines). But I'm not sure why this whole file is showing up in the diff... Is this a GitHub bug or am I missing something here...?

josephkmh: Update: it's because these files were commited with CRLF terminators (probably from a windows machine?) and I assume prettier converted them to just LF on my unix machine (macbook).

Screenshot 2022-12-09 at 9 27 32 PM

josephkmh: I opened a separate PR to fix these changes here: https://github.com/airbytehq/airbyte/pull/20370

Hopefully we can get that merged first, and this will disappear from the diff in this PR 😄

ambirdsall:
Visual nit: when the checkbox is not being rendered, a tiny thin `<label>` is still showing a 1px border where it seems like there should be nothing. Also: if you deselect a table, should all of its columns be deselected in the dropdown, too? I can see both sides, aside from complexity, a user might 
column-selection-questions

airbyte-webapp/src/hooks/services/Feature/types.tsx
```diff
@@ -13,6 +13,7 @@ export enum FeatureItem {
   AllowSync = "ALLOW_SYNC",
   AllowChangeDataGeographies = "ALLOW_CHANGE_DATA_GEOGRAPHIES",
   AllowSyncSubOneHourCronExpressions = "ALLOW_SYNC_SUB_ONE_HOUR_CRON_EXPRESSIONS",
+  AllowFieldSelection = "ALLOW_FIELD_SELECTION",
```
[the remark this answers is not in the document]
josephkmh: Oops, `FeatureItem.AllowFieldSelection` needs to be removed. I started with a `FeatureItem` but realized an experiment is more appropriate for a slow rollout. Thanks 🙂

josephkmh:
> Also: if you deselect a table, should all of its columns be deselected in the dropdown, too? I can see both sides, aside from complexity, a user might

Good question - I also thought about this and decided to leave the column selection state alone when enabling/disabling a stream entirely. I can imagine that someone might want to temporarily disable a stream, but keep their column selection configuration in place. Also since you can edit other properties of a disabled stream (e.g. sync mode) it seems to follow that you should also be able to change the column selection independent of enabling the stream.

josephkmh:
> Visual nit: when the checkbox is not being rendered, a tiny thin <label> is still showing a 1px border where it seems like there should be nothing. 

I think this is unrelated to the column selection changes? The bulk changes checkbox is a separate interaction that already exists. Or am I misunderstanding?

airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
```diff
@@ -96,6 +98,32 @@ const CatalogSectionInner: React.FC<CatalogSectionInnerProps> = ({
     [updateStreamWithConfig]
   );
 
+  const numberOfFieldsInStream = Object.keys(streamNode?.stream?.jsonSchema?.properties).length ?? 0;
+
+  const onSelectedFieldsUpdate = (selectedFields: SelectedFieldInfo[]) => {
```
ambirdsall, line 103: A thought: if `onSelectedFieldsUpdate` took `previouslySelectedFields` as a second param, instead of just `selectedFields`, then the special-case props `onFirstFieldDeselected` and `onAllFieldsSelected` could go away and and the fact that those are special cases would be more cleanly scoped to this one file.

ambirdsall:
> I think this is unrelated to the column selection changes? The bulk changes checkbox is a separate interaction that already exists.

Double-checked on `master` and the same visual bug exists, so definitely not part of the required scope for this change. Would you want to bundle that in with your a11y changes to `CheckBox`, or should I cut a new issue?

airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
josephkmh: Good point, I think that would indeed be cleaner! As discussed offline, I'll defer this to a follow-up PR

josephkmh:
@ambirdsall the bug is within the `<Checkbox />` component, right? Could definitely bundle it with the a11y issue then IMO!

ambirdsall (approved):
Per offline discussion, we'll defer the `onSelectedFieldsUpdate` refactoring to a follow-up PR. Everything else looks good to me, and the UI acted as expected in (admittedly cursory) manual testing. Good work :+1:

PR #20272, opened 2022-12-08
Add secrets for pulling custom connector images

xiaohansong:
## What
Add secrets for pulling custom connector images

## How
Adding an optional config field and pass it along to the pod creation process. Pod can accept them as a list and can pull images from both repositories if configured.


airbyte-config/config-models/src/main/java/io/airbyte/config/Configs.java
```diff
@@ -466,6 +466,8 @@ public interface Configs {
    */
   String getJobKubeMainContainerImagePullSecret();
 
+  Optional<String> getCustomImagePullSecret();
```
pmossman, line 469: I think instead of adding a brand new config, it would make more sense for our OSS users if we instead updated the existing config to support a list.

For an OSS user who wants to use custom images, they're supposed to use the [existing config,](https://github.com/airbytehq/airbyte/blob/master/airbyte-config/config-models/src/main/java/io/airbyte/config/Configs.java#L464-L467) so having a second config will add confusion.

Basically, we just need to add support for defining multiple image pull secrets, so that Cloud can use two private image repositories instead of being limited to one.

Does that make sense?

pmossman (changes requested):

airbyte-config/config-models/src/main/java/io/airbyte/config/Configs.java
xiaohansong: Makes sense, good call! I'll fix this.

pmossman: Thanks! Sorry to make you rework it all 😅

xiaohansong: Fixed - and retested on cloud, this should work! (though some other bugs were detected, PR coming in...)

airbyte-commons-worker/src/main/java/io/airbyte/workers/WorkerConfigs.java
```diff
@@ -21,7 +21,7 @@ public class WorkerConfigs {
   private final Map<String, String> workerKubeNodeSelectors;
   private final Optional<Map<String, String>> workerIsolatedKubeNodeSelectors;
   private final Map<String, String> workerKubeAnnotations;
-  private final String jobImagePullSecret;
+  private final List<String> jobImagePullSecret;
```
pmossman, line 24: nit: would call this `jobImagePullSecrets` with the added 's'

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/KubePodProcess.java
```diff
@@ -515,8 +515,13 @@ public KubePodProcess(final boolean isOrchestrator,
...
     }
 
+    List<LocalObjectReference> pullSecrets = imagePullSecrets
+        .stream()
+        .map(imagePullSecret -> new LocalObjectReference(imagePullSecret))
+        .collect(Collectors.toList());
+
     final Pod pod = podBuilder.withTolerations(buildPodTolerations(tolerations))
-        .withImagePullSecrets(new LocalObjectReference(imagePullSecret)) // An empty string turns this into a no-op setting.
+        .withImagePullSecrets(pullSecrets) // An empty string turns this into a no-op setting.
```
pmossman, line 524: nit: comment should probably now say "An empty list turns this into a no-op setting."

airbyte-config/config-models/src/main/java/io/airbyte/config/Configs.java
```diff
@@ -464,7 +464,7 @@ public interface Configs {
   /**
    * Define the Job pod connector image pull secret. Useful when hosting private images.
    */
-  String getJobKubeMainContainerImagePullSecret();
+  List<String> getJobKubeMainContainerImagePullSecret();
```
pmossman, line 467: nit: would add an 's' and make it `getJobKubeMainContainerImagePullSecrets()`

airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
```diff
@@ -108,6 +108,7 @@ public class EnvConfigs implements Configs {
   public static final String JOB_DEFAULT_ENV_PREFIX = "JOB_DEFAULT_ENV_";
   private static final String SECRET_PERSISTENCE = "SECRET_PERSISTENCE";
   public static final String JOB_KUBE_MAIN_CONTAINER_IMAGE_PULL_SECRET = "JOB_KUBE_MAIN_CONTAINER_IMAGE_PULL_SECRET";
+  public static final String JOB_KUBE_CUSTOM_IMAGE_PULL_SECRET = "JOB_KUBE_CUSTOM_IMAGE_PULL_SECRET";
```
pmossman, line 111: I think this can be deleted

pmossman (approved):
Looks good @xiaohansong! I left a few nits about making variable names plural. I was wondering if we should change the environment variable from `JOB_KUBE_MAIN_CONTAINER_IMAGE_PULL_SECRET` to `JOB_KUBE_MAIN_CONTAINER_IMAGE_PULL_SECRETS` but it might not be worth it since it'd be a breaking change for OSS deployments that use it (and we'd need to update the environment variables in Cloud too, though I'm more concerned with breaking changes in OSS)

I think renaming our code variables and method to be plural while leaving the environment variable alone is a good middle ground, at least in my opinion.

PR #20276, opened 2022-12-08
🪟 🎉 Prevent users from editing existing connection if they have a schema change

### LINKED ISSUE #20201: Prevent the user from editing stream if there is a breaking schema change
When the user has a breaking schema change, they should not be allowed to break the schema ([Figma](https://www.figma.com/file/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?node-id=972%3A37260&t=FCfVs1rpFzu3Yk62-1))

image

* Needs to be supported in both the old table and new table design

teallarson:
## What
closes #20201 

Screenshot 2022-12-08 at 7 33 31 PM

## How
We would like this to only cover the sync catalog, not the schedule and non-breaking changes preferences dropdown.  Because that requires more refactoring of this form, for now we are just blocking the entire form.

## Recommended reading order
1. `<SchemaChangeBackdrop/>` and its scss module
2. `<ConnectionReplicationTab />`
3. others

## Testing
Have `REACT_APP_AUTO_DETECT_SCHEMA_CHANGES=true` when you start the frontend.  
Make a breaking change to a connection, then refresh the schema and navigate away to trigger the new flow.
If you navigate back to `Replication` tab, you should see the overlay!

When you hit `View Changes` it should disappear.

When you resolve the breaking change and submit, you should not see the overlay.


airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionReplicationTab.tsx
```diff
@@ -193,28 +197,33 @@ export const ConnectionReplicationTab: React.FC = () => {
...
-            <Form>
-              <ValidateFormOnSchemaRefresh />
-              <ConnectionFormFields
-                values={values}
-                isSubmitting={isSubmitting}
-                dirty={dirty || schemaHasBeenRefreshed}
-              />
-              {status.editControlsVisible && (
-                <EditControls
+            <SchemaChangeBackdrop
```
tealjulia, line 200: only change here is wrapping this in the backdrop

airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionPageTitle.module.scss
```diff
@@ -1,9 +1,11 @@
-@use "../../../../scss/colors";
-@use "../../../../scss/variables";
+@use "scss/colors";
+@use "scss/variables";
+@use "scss/_z-indices";
 
 .container {
   text-align: center;
   padding: variables.$spacing-lg variables.$spacing-xl;
+  z-index: z-indices.$connectionPageTitle;
```
tealjulia, line 8: this resolves the bug we saw in the bug bash where users were unable to click various tabs or click the review button with the backdrop present.  it was because we're using absolute positioning to keep octavia from scrolling.

airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionReplicationTab.tsx
```diff
@@ -193,28 +194,30 @@ export const ConnectionReplicationTab: React.FC = () => {
...
-            <Form>
-              <ValidateFormOnSchemaRefresh />
-              <ConnectionFormFields
-                values={values}
-                isSubmitting={isSubmitting}
-                dirty={dirty || schemaHasBeenRefreshed}
-              />
-              {status.editControlsVisible && (
-                <EditControls
+            <SchemaChangeBackdrop>
```
tealjulia, line 197: only change here is wrapping the form in the backdrop

airbyte-webapp/src/scss/_z-indices.scss
```diff
@@ -5,3 +5,6 @@ $sidebar: 9999;
 $panelSplitter: 0;
 $dropdownMenu: 2;
 $notification: 20;
+$schemaChangesBackdrop: 3;
```
tealjulia, line 8: explicitly set a higher z-index than the dropdown menu and the switch (2)

airbyte-webapp/src/views/Connection/ConnectionForm/components/SchemaChangesBackdrop.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+      connection: { mockConnection, schemaChange: SchemaChange.breaking },
+      schemaHasBeenRefreshed: false,
+      schemaRefreshing: false,
+    });
+
+    const { getByTestId } = renderComponent();
+
+    expect(getByTestId("schemaChangesBackdrop")).toMatchSnapshot();
+    userEvent.click(getByTestId("bg-button"));
+    expect(buttonSpy).not.toHaveBeenCalled();
```
tealjulia, line 44: testing that background elements aren't clickable... though in reality, the implementation ends up dependent on the z-index of elements within the connection form.  i didn't feel that implementing a whole integration test to include the actual form elements in this was a good use of time for this PR.

airbyte-webapp/src/locales/en.json
```diff
@@ -154,6 +154,8 @@
   "connectionForm.nonBreakingChangesPreference.message": "Set how Airbyte handles syncs when it detects a non-breaking schema change in the source",
   "connectionForm.nonBreakingChangesPreference.ignore": "Ignore",
   "connectionForm.nonBreakingChangesPreference.disable": "Pause sync",
+  "connectionForm.breakingChanges.backdrop.message": "We detected some breaking schema updates, please review the changes in your connection header",
+  "connectionForm.nonBreakingChanges.backdrop.message": "We detected some schema updates, please review the changes in your connection header",
```
tealjulia, line 158: We could put these into one and pass in a variable to decide whether or not to add in the word "breaking"... I didn't feel too strongly on it either way.

airbyte-webapp/src/components/ui/Switch/Switch.module.scss
```diff
@@ -41,7 +42,7 @@
 
     &::before {
       position: absolute;
-      z-index: 1;
+      z-index: z-indices.$switchSliderBefore;
```
tealjulia, line 45: Moving this to a variable.  It was causing a bug where users could interact with the switches behind the backdrop.  We really want any z-index in the global file to prevent collisions.

airbyte-webapp/src/scss/_z-indices.scss
tealjulia: Also, unsure if there is a reason not to, but it might be nice to have these ordered by index?  

Happy to make a quick commit that does that...

airbyte-webapp/src/views/Connection/ConnectionForm/components/SchemaChangeBackdrop.module.scss
```diff
@@ -0,0 +1,46 @@
...
+.schemaChangeBackdropContainer {
+  @extend %cover;
+
+  position: relative;
+
+  .backdrop {
+    @extend %cover;
+
+    position: absolute;
+    background: linear-gradient(180deg, rgba(colors.$grey-50, 0.5) 0%, colors.$white 92.54%);
```
tealjulia, line 19: Figma shows the old app bg color fading into white.  I've kept the gradient the same, and subbed in our new app bg color.

airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionPageTitle.module.scss
edmundito: Intead of adding more complexity to the z-index. One alternative I experimented with was not to have the backdrop be fixed but instead I chagned these rules to the backdrop css:

* remove the justify-content: center
* remove the bottom rule
* add 120 px margin top
* let it scroll with the page

This works well with what we have available with the layout. We can revisit this later when we introduce a layout that is maxed to the height of the page.

tealjulia: Agree.  That was close to what I had in place before the chat with Nico earlier this week.  I will switch that over.

airbyte-webapp/src/locales/en.json
tealjulia: Yep.  Fine with whatever Andy proposes there.  I just copy/pasted from the designs.

airbyte-webapp/src/views/Connection/ConnectionForm/components/SchemaChangeBackdrop.tsx
```diff
@@ -0,0 +1,49 @@
...
+    return <>{children}</>;
+  }
+
+  return (
+    <div className={styles.schemaChangeBackdropContainer} data-testid="schemaChangesBackdrop">
+      <div className={styles.backdrop}>
+        <div className={styles.contentContainer}>
+          <div>{schemaChangeImage}</div>
+          <Text className={styles.text}>
+            <FormattedMessage id="connectionForm.breakingChangesBackdrop.message" />
```
edmundito, line 42: The id in the en file is `connectionForm.schemaChangesBackdrop.message`

edmundito (approved):

edmundito:
Tested with and without feature enabled, breaking and non-breaking change. Canceled and applied changes.

PR #20277, opened 2022-12-09
Bmoric/convert destination definition controller micronaut

benmoriceau:
## What
Convert the destination definition controller to micronaut


jdpgrailsdev (approved):
:shipit:

airbyte-server/src/main/java/io/airbyte/server/apis/DestinationDefinitionApiController.java
```diff
@@ -15,20 +15,27 @@
...
-import javax.ws.rs.Path;
-import lombok.AllArgsConstructor;
+import io.micronaut.context.annotation.Context;
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Post;
 
-@Path("/v1/destination_definitions")
-@AllArgsConstructor
+@Controller("/api/v1/destination_definitions")
+@Context
```
colesnodgrass, line 23: Same question as on the Source PR, is this annotation necessary?

colesnodgrass (approved):
one question about the `@Context` annotation

airbyte-server/src/main/java/io/airbyte/server/apis/DestinationDefinitionApiController.java
jdpgrailsdev: @colesnodgrass I believe the intent is to force this singleton to be created on startup for performance reasons.  Otherwise, the first call to this endpoint has to wait for all of the related beans to be initialized.  We should look into why it is slow (maybe we are doing something in a constructor of a singleton that should be deferred, etc).  I think that the use of `@Context` here is okay in this context because we are using it on endpoints/controllers to avoid a first request impact.

benmoriceau: I answered in the other PR

PR #20280, opened 2022-12-09
[Connector Builder] Config UI error handling

lmossman:
## What
This PR improves the form error handling in the connector builder config ui.

## How
This PR utilizes formik hooks to disable the test and download buttons if there are any errors in formik. If a user clicks one of these buttons before some fields that actually do have errors have been touched, form validation is performed and any erroring fields are touched, and the actions of those buttons are prevented in that case.

It also adds a red dot error indicator to the view buttons to indicate when the view contains a field with an error.

## Recommended reading order
1. Main changes are in the new StreamTestButton component and the new useBuilderErrors custom hook

## 🚨 User Impact 🚨
Errors in the configuration ui of the connector builder are now clearer to the user.


airbyte-webapp/src/components/connectorBuilder/Builder/BuilderField.tsx
```diff
@@ -53,25 +52,7 @@ const ArrayField: React.FC<ArrayFieldProps> = ({ name, value, setValue, error })
...
-        return undefined;
-      } catch (err) {
-        if (err instanceof yup.ValidationError) {
-          return err.errors.join(", ");
-        }
-        throw err;
-      }
-    },
-  };
-  const [field, meta, helpers] = useField(fieldConfig);
```
lmossman, line 74: I could no longer have BuilderFields dynamically set their own yup validation schemas, because this meant that switching away from some stream/global view would cause those validation schemas to no longer be applied, so errors across the form could not be tracked.

I instead had to define a single global validation schema that is used for the whole builder form, which I added to types.ts. This will make it a bit more annoying to add new fields, as that validation schema will also need to be updated, but I feel it is worth it for the better error tracking functionality.

(this is one more point in favor of generating these things from the JSON schema)

lmossman:
There's still one issue with this that I want to fix before merging - when starting from a clean form (e.g. if you click the `Reset Builder` button and refresh), as soon as you start typing into the connector name field, it already disables the download button and shows an error indicator on the Global Configuration view button.

I'd like this instead to only show those error indicators and disable the download button in cases like these:
- The user has clicked into a required field and then clicked away
- The user clicks the download / test button and some required field is empty

Open to suggestions here but I will continue to look into it tomorrow

flash1293:
This works pretty well!

Not related to this PR but I noticed that the stream test panel gets confused if the name is not set consistently in the yaml:
Screenshot 2022-12-09 at 16 14 03

One more reason to align and simplify the yaml here. Is this a bug we should report separately?

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.module.scss
```diff
@@ -100,3 +100,17 @@
...
+  flex: 1;
+  overflow: hidden;
+  display: flex;
+  gap: variables.$spacing-sm;
+  align-items: center;
+}
+
+.errorIndicator {
+  flex: 0 0 auto;
+  margin-left: auto;
```
flash1293, line 114: should not be necessary

flash1293:
> I'd like this instead to only show those error indicators and disable the download button in cases like these:
> The user has clicked into a required field and then clicked away
> The user clicks the download / test button and some required field is empty

I think this should work by changing `hasErrors` to ignore errors on fields that are not touched yet when looking to disable the buttons / showing the error indicator (you still need to take them into account in the validateAndTouch).

flash1293:
> I'd also like to get this to focus a view if it has an error and is not currently focused when the user clicks test

on the other PR I recommended to move the selected view state into the main connector builder state manager - this could come in handy here as you can check where the first error is coming from in validateAndTouch and set the selected view to that one:
```
if (!hasErrors(limitToStream, errors)) {
        callback();
} else {
 const firstErrorView = workoutFirstErrorView(errors) // if there are global errors, take those, otherwise look for the first stream
 setSelectedView(firstErrorView);
}
```

lmossman:
> One more reason to align and simplify the yaml here. Is this a bug we should report separately?

Alex said that they are planning to eliminate the need to specify `name` at multiple levels this quarter (and instead just specify it once at the stream level), so that issue should be resolved by that change

flash1293:
>  I decided to not disable the Test button when there is an error in the form, because now that it auto-selects the erroring view, I think its a nicer iteration loop to fix all errors in your current view, click Test again, and it brings you to the next relevant view with errors, so you fix those and click Test again and then it works.

I think that's a good idea, should we do the same for the "Download Config" button? It seems like a similar situation. Also, to make it more obvious what's going on I think we should change the icon to the warning triangle.

So for both Test and Download:
* Make it clickable and focus the next error view
* Show a warning triangle as icon to tell the user it won't actually submit in the current state

airbyte-webapp/src/components/connectorBuilder/useBuilderErrors.ts
```diff
@@ -1,45 +1,72 @@
...
+    return limitToViews === undefined ? invalidViews : intersection(invalidViews, limitToViews);
   };
 
-  const validateAndTouch = (callback: () => void, limitToStream?: number) => {
+  // Returns true if the global config fields or any stream config fields have errors in the provided formik errors, and false otherwise.
+  // If limitToViews is provided, the error check is limited to only those views.
+  const hasErrors = (
+    ignoreUntouched: boolean,
+    limitToViews?: BuilderView[],
+    inputErrors?: FormikErrors<BuilderFormValues>
```
flash1293, line 47: Seems like inputErrors is never used?

flash1293 (commented):
Left two small notes but it seems mostly good to me.

Somehow the tests have been failing a lot on this PR, not sure whether that's a real problem or some infra issue.

lmossman:
@flash1293 I agree with your suggestions, and I've pushed those changes up.

I'll look into the CI errors but on first glance it does seem unrelated to my changes

flash1293 (approved):
LGTM, thanks for these adjustments

PR #20286, opened 2022-12-09
airbyte-metric: implement `recordErrorOnRootSpan` in ApmUtils

alafanechere:
## What
We want to leverage the `Error Tracking` panel of Datadog.
To do so, we need to bubble up exceptions to the root span, so datadog considers this a new error.

## How
Add `recordErrorOnRootSpan` method to `ApmTraceUtils`

## Context
I'm using it in https://github.com/airbytehq/airbyte-cloud/pull/3799


airbyte-server/build.gradle
```diff
@@ -37,9 +39,10 @@ dependencies {
     implementation 'org.glassfish.jersey.inject:jersey-hk2'
     implementation 'org.glassfish.jersey.media:jersey-media-json-jackson'
     implementation 'org.glassfish.jersey.ext:jersey-bean-validation'
+    implementation 'io.opentelemetry:opentelemetry-api'
+    implementation 'io.opentelemetry:opentelemetry-extension-annotations'
```
jdpgrailsdev, line 43: Per our discussion, we don't need the OpenTelemetry dependencies in airbyte-server.

jdpgrailsdev (approved):
:shipit:

PR #20287, opened 2022-12-09
🪟🔧 Fix node version on CI

timroes:
## What

This aligns our node version handling we use across the product. This is caused by us having had `package-lock.json` changes due to a missmatch in `npm` versions run during development.

## How

* Use fixed versions for webapp development instead of the `lts` shorthands, so we are set to an explicit node version, since the `lts` version might change over time while new minor releases are incoming.
* Load the version Gradle downloads for building the webapp and e2e tests from said `.nvmrc` files, to avoid duplication in where we store this version.
* Add the version to `engines.node` in the `package.json` as an explicit version. This will now cause `npm install` to fail if someone tries to execute it locally with a version that's not exactly the one we're expecting.
* Since gradle takes care of downloading the node version for webapp and e2e tests, we don't need to use github actions to load them. I removed the corresponding `setup-node` actions from the workflow files, to not cause confusion there with missmatching versions.
* Adjust the `package-lock.json` files to actually match the now explicitly set npm version... not sure how many times they had a format change during one major version :weary: 

I tried to use the latest LTS support for this, but unfortunately there are still issues with CRA app, that won't work on the latest LTS release, so we're still stuck on LTS Gallium till they fix this: https://github.com/facebook/create-react-app/issues/11708

You'll see there are some `setup-node` actions left in the workflow files. We use Spotless, which uses prettier in the top level `build.gradle` file to format JSON and YAML across the repository (except the webapp). Thus the spotless tasks actually needs a locally installed `npm` version, why the workflows that trigger a full build still require the `npm` to be installed locally. The version here isn't that important, since it's just using npm once to install prettier, so I set it to the latest lts release. The webapp is not using that npm version, and always will download it's own version via Gradle!

I tried to solve this by downloading that Gradle version for spotless also via the gradle node plugin, but that unfortunately doesn't work, since the spotless plugin required it to be there during Gradle's Configuration phase (while parsing the build.gradle files), but the download task cannot download it before the Execution phase (when tasks are actually executed). So we can't leverage gradle to download npm for Spotless.

## Follow up

I want to create a separate PR still that adds a Gradle task that makes the FE build (and e2e test build) fail if the `package-lock.json` file is changed after running `npm  install`. Since this requires adding custom tasks I prefer having that in a separate PR.


krishnaglick (approved):
Approving pending build

josephkmh:
> I want to create a separate PR still that adds a Gradle task that makes the FE build (and e2e test build) fail if the package-lock.json file is changed after running npm install. Since this requires adding custom tasks I prefer having that in a separate PR.
Do we need gradle for this? Couldn't we just use `npm ci` instead of `npm install`? AFAIK that will install from `package-lock.json` and also throw if there is a mismatch with `package.json` https://docs.npmjs.com/cli/v9/commands/npm-ci

timroes:
I can investigate that next week. My understanding was that npm ci will only install from package-lock.json but wouldn't fail on those minor inconsistencies. That's why I decided to go with the change detection mechanism, but I will give this another look.

.github/workflows/gke-kube-test-command.yml
```diff
@@ -72,7 +72,7 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
-          node-version: "lts/gallium"
+          node-version: "lts/*"
```
josephkmh, line 75: Why is `lts/*` ok here? I'm assuming node is just required here (and in other workflows) for some scripting, which is unrelated to any of the frontend build/test steps?

joeykmh (approved):
LGTM! I pulled and tested locally to confirm:
- `npm install` failed unless I was using node `16.18.1`
- `npm build` does not fail when using other versions (although has build errors with node 18 😄 )
- `npm run start` does not fail when using other versions

flash1293 (commented):
Looks good to me, should we add a small section to the readme how to keep your local node version in sync using nvm?

timroes:
> Looks good to me, should we add a small section to the readme how to keep your local node version in sync using nvm?

I'll put something about this into our development setup guide.

PR #20303, opened 2022-12-09
ci-connector-ops: declare review requirements on source connectors changes

### LINKED ISSUE #20304: ci-connector-ops: require reviews from connector team
We want to automatically block a PR from being merged without the approval of the Connector Extensibility /Operations team when the backward compatibility checks are disabled.

* Add a new command to `ci_connector_ops` to check if the current branch disabled backward compatibility tests in `acceptance-test-config.yml`
* Create the following logic in a GitHub action:
```
if script returns 1: 
  if not any([reviewers in connectors_team for reviewers in approving_reviewers])
    fail job
```

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/20304

This [GitHub action](https://github.com/Automattic/action-required-review) can help us check if a set of teams gave a review on a PR by creating a "review requirements" file.
I made a new command in `ci-connector-ops` to write this file according to the current changes in the branch.

I changed the `connector_ops_ci.yml` workflow to:
* Call the new `write-review-requirements-file` command to write the requirements file. This file structure is defined by the `Automattic/action-required-review@v3` action
* Run the `Automattic/action-required-review@v3` if this file was written. 

When the requirements are not met, a status check as this one will appear:
Screen Shot 2022-12-12 at 13 48 59

## The new following review requirements rules:
* If a PR touches a GA connector: a review from `gl-python` team is required
* If a PR changes the test strictness level in `acceptance-test-config.yml`: a review from `connector-operations `teams is required
* If a PR changes backward compatibility test in `acceptance-test-config.yml`: a review from `connector-operations` or `connector-extensibility` is required.

### `ci_connector_ops` python package
* The new command to write the review requirements file: https://github.com/airbytehq/airbyte/blob/9da488761dae17b709d7057f51bba8a85dd2a0de/tools/ci_connector_ops/ci_connector_ops/sat_config_checks.py#L76
* Implement `get_changed_acceptance_test_config` to track changes on `acceptance-test-config.yml`: 
https://github.com/airbytehq/airbyte/blob/9da488761dae17b709d7057f51bba8a85dd2a0de/tools/ci_connector_ops/ci_connector_ops/utils.py#L55
* "Integration" tests:
https://github.com/airbytehq/airbyte/blob/9da488761dae17b709d7057f51bba8a85dd2a0de/tools/ci_connector_ops/tests/test_sat_config_checks.py#L90

### `connector_ops_ci.yml` GitHub action workflow 
* Add a new `check-review-requirements` job
https://github.com/airbytehq/airbyte/blob/45d990a063a5bc2dddee9306318be3a96fd00655/.github/workflows/connector_ops_ci.yml#L25


tools/ci_connector_ops/setup.py
```diff
@@ -19,7 +19,8 @@
     python_requires=">=3.9",
     entry_points={
         "console_scripts": [
-            "check-test-strictness-level = ci_connector_ops.check_test_strictness_level:main",
+            "check-test-strictness-level = ci_connector_ops.sat_config_checks:check_test_strictness_level",
+            "requires-connector-team-review = ci_connector_ops.sat_config_checks:requires_connector_team_review"
```
erohmensing, line 23: Even if it has been reviewed by a team member, won't it still return 1 and make the build not green? 

I'm not sure exactly how it would look, but maybe in the github action we can do something like (pseudocode in python-ish that has to be bash)
```
if script returns 1: 
  if not any([reviewers in connectors_team for reviewers in approving_reviewers])
    fail job
```

alafanechere: Yes the outer logic of when to call this command should be built in a GitHub action... 
@pedroslopez was suggesting to at least request for review to our team if the script returns 1. Getting the list of approving reviewers looks a bit trickier. Is is something you've done in GHA already?

erohmensing: You're right, I assumed there'd be some sort of readily available `github.event.pull.request.reviewers` variable, but there doesn't seem to be anything regarding reviewers [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows). 

Another option that may be easier would be instead of checking the review itself, the action could be the one that requests the team for review if the script returns 1. There seem to be a decent amount of actions available that do that:
* https://github.com/marketplace/actions/auto-request-review seems maintained
  * `Auto-assign the default reviewers if no reviewers are matched to your rules` or `Request review only in certain conditions` is probably the relevant behavior here 
  * https://github.com/marketplace/actions/add-pull-request-reviewer is a bit older but seems simple

erohmensing (approved):
Testing, testing, 1,2,3...

.github/workflows/connector_ops_ci.yml
```diff
@@ -21,3 +22,35 @@ jobs:
         run: pip install --quiet -e ./tools/ci_connector_ops
       - name: Check test strictness level
         run: check-test-strictness-level
+  requires-connector-team-review:
+    name: "Check if a review is required from Connector teams"
+    runs-on: ubuntu-latest
+
+    # GitHub should provide a "pull_request_review_target", but they don't and
+    # the action will fail if run on a forked PR.
+    if: github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name
```
evantahler, line 31: I've done this kind of thing in the past:

```
if: ${{ github.repository == 'airbytehq/airbyte' }}
```

.github/workflows/connector_ops_ci.yml
```diff
@@ -21,3 +22,35 @@ jobs:
...
+        if: steps.check-requires-connector-team-review.outputs.REQUIRES_CONNECTOR_TEAM_REVIEW == 'True'
+        uses: Automattic/action-required-review@v3
+        with:
+          status: Required review
+          fail: true
+          token: ${{ secrets.OCTAVIA_4_ROOT_ACCESS }}
+          requirements: |
+            - paths: unmatched
+              teams:
+                - connector-operations
```
evantahler, line 56: [non-blocking] Do you think that we are /always/ the proper people to take the review?  On one hand, we know the most about the spec and possibly the CDK (+connector-extensibility?), but perhaps the question is connector specific and we will need GL...

I guess we can re-assign the review as needed.

.github/workflows/connector_ops_ci.yml
```diff
@@ -21,3 +22,35 @@ jobs:
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.9"
+      - name: Install ci-connector-ops package
+        run: pip install --quiet -e ./tools/ci_connector_ops
+      - name: Check if the PR requires a connector team review
+        id: check-requires-connector-team-review
+        run: check-requires-connector-team-review >> $GITHUB_OUTPUT
+      - name: Check if the required teams reviewed
+        if: steps.check-requires-connector-team-review.outputs.REQUIRES_CONNECTOR_TEAM_REVIEW == 'True'
```
evantahler, line 47: [not-blocking] The way python stringifies "true" is so strange.  Maybe the output should be "true" or "1"?

tools/ci_connector_ops/ci_connector_ops/utils.py
```diff
@@ -48,11 +50,21 @@ def get_changed_connector_names() -> Set[str]:
...
+def get_changed_acceptance_test_config() -> Set[str]:
+    """Retrieve a list of connector names for which the acceptance_test_config file was changed in the current branch (compared to master).
+
+    Returns:
+        Set[str]: Set of connector names e.g {"source-pokeapi"}
+    """
+    changed_acceptance_test_config_paths = {
+        file_path
+        for file_path in AIRBYTE_REPO.git.diff("--name-only", "origin/master...").split("\n")
+        if file_path.startswith(SOURCE_CONNECTOR_PATH_PREFIX) and file_path.endswith(ACCEPTANCE_TEST_CONFIG_FILE_NAME)
```
evantahler, line 64: cool use of git to see if the file changed!

evantahler (approved):
This is great!

I have the feeling we'll be iterating on /who/ needs to do the required review, but this is an excellent first step.

.github/workflows/connector_ops_ci.yml
alafanechere: A cool evolution would be to make the python script return the name of the teams that should review. To keep things simple at the moment, I'll make both `connector-operations` and `connector-extensibility` required reviewers (in an `any-of` manner).

.github/workflows/connector_ops_ci.yml
alafanechere: 👌 I'll make it return `true` or `false`

tools/ci_connector_ops/ci_connector_ops/utils.py
alafanechere: The current limitation is that it checks if the `acceptance-test-config.yml` was modified, not specifically that `disable-for-version` was added. We'd be required a review even if the `disable-for-version` key was added in previous PRs and the change to `acceptance-test-config.yml` are not related to backward compatibility. I'll try to check the diff content itself.

.github/workflows/connector_ops_ci.yml
alafanechere: 👍 Thanks!

tools/ci_connector_ops/ci_connector_ops/sat_config_checks.py
```diff
@@ -0,0 +1,99 @@
...
+def find_mandatory_reviewers() -> List[Union[str, Dict[str, List]]]:
+    teams = []
+    ga_connector_changes = find_changed_ga_connectors()
+    backward_compatibility_changes = utils.get_changed_acceptance_test_config(diff_regex="disable_for_version")
+    test_strictness_level_changes = utils.get_changed_acceptance_test_config(diff_regex="test_strictness_level")
+    if ga_connector_changes:
+        teams += GA_CONNECTOR_REVIEWERS
+    if backward_compatibility_changes:
+        teams += [{"any-of": list(BACKWARD_COMPATIBILITY_REVIEWERS)}]
+    if  test_strictness_level_changes:
```
erohmensing, line 62:
```suggestion
    if test_strictness_level_changes:
```

tools/ci_connector_ops/ci_connector_ops/sat_config_checks.py
```diff
@@ -0,0 +1,99 @@
...
+
+def write_review_requirements_file():
+    mandatory_reviewers = find_mandatory_reviewers()
+
+    if mandatory_reviewers:
+        requirements_file_content = [{
+            "name": "Required reviewers from the connector org teams",
+            "paths": "unmatched",
+            "teams": mandatory_reviewers
+        }]
```
erohmensing, line 84: Cool idea to generate this on each run!

tools/ci_connector_ops/ci_connector_ops/sat_config_checks.py
```diff
@@ -0,0 +1,99 @@
...
+        print('REVIEW_REQUIREMENTS_FILE=""')
+
+def print_mandatory_reviewers():
+    teams = []
+    mandatory_reviewers = find_mandatory_reviewers()
+    for mandatory_reviewer in mandatory_reviewers:
+        if isinstance(mandatory_reviewer, dict):
+            teams += mandatory_reviewer["any-of"]
+        else:
+            teams.append(mandatory_reviewer)
```
erohmensing, line 98: Doing the `any-of` check here feels a little un-DRY, since you also manually add `any-of` in `find_mandatory_reviewers` -- maybe you could put this logic of adding the any-of into a helper method you call in `find_mandatory_reviewers`?

tools/ci_connector_ops/ci_connector_ops/sat_config_checks.py
```diff
@@ -0,0 +1,99 @@
...
+    print(f"MANDATORY_REVIEWERS=A review is required from these teams: {','.join(teams)}")
```
erohmensing, line 99:
```suggestion
    print(f"MANDATORY_REVIEWERS=A review is required from these teams: {','.join(teams)}")

```

.github/workflows/connector_ops_ci.yml
```diff
@@ -21,3 +22,33 @@ jobs:
...
+        id: write-review-requirements-file
+        run: write-review-requirements-file >> $GITHUB_OUTPUT
+      - name: Get mandatory reviewers
+        id: get-mandatory-reviewers
+        run: print-mandatory-reviewers >> $GITHUB_OUTPUT
+      - name: Check if the review requirements are met
+        if: steps.write-review-requirements-file.outputs.REVIEW_REQUIREMENTS_FILE != ''
+        uses: Automattic/action-required-review@v3
+        with:
+          status: ${{ steps.get-mandatory-reviewers.outputs.MANDATORY_REVIEWERS  }}
```
erohmensing, line 51: To be clear, generating and outputting this separately is a requirement of the github action, right? We can't generate that into `connector_org_review_requirements.yaml`?

erohmensing (approved):
This looks like it will be super helpful! Might be worth sharing this action and how you implemented it (Loom?) to other teams who may be unsatisfied with how static codeowners is

.github/workflows/connector_ops_ci.yml
alafanechere: Exactly, I added the  `MANDATORY_REVIEWERS` to conveniently display a detailed status like `A review is required from teams: A, B, C` instead of the default `Review required` status. But all the essential information is captured in the `REVIEW_REQUIREMENTS_FILE` whose format is defined by the external action.

erohmensing: Cool, I figured. Strange that you can't configure different status messages for the different sets of requirements in the action 😄

YowanR:
This makes sense at a high-level to me. Adding @lazebnyi as a reviewer to make sure that GL is plugged into this conversation.

lazebnyi (approved):

evantahler (approved):

evantahler (changes requested):
I left a comment and that triggered a failing check!

https://github.com/airbytehq/airbyte/actions/runs/3680412195

```
[Invalid workflow file: .github/workflows/connector_ops_ci.yml#L49](https://github.com/airbytehq/airbyte/actions/runs/3680412195/workflow)
The workflow is not valid. .github/workflows/connector_ops_ci.yml (Line: 49, Col: 13): Unexpected symbol: '""'. Located at position 74 within expression: steps.write-review-requirements-file.outputs.REVIEW_REQUIREMENTS_FILE != ""
```

alafanechere:
@evantahler, there was a glitch in the boolean check to conditionally run the final action. I fixed it in https://github.com/airbytehq/airbyte/pull/20303/commits/e7202ff10dff46b8128f1e82c6f60e8c33a29e31

evantahler (approved):
👍 assuming I don't get another failing comment :D

PR #20311, opened 2022-12-09
remove the requirement on the activity

benmoriceau:
## What
For the auto detect schema change, we need to check the status of a connection after the refresh schema call because it can be disabled by the refresh schema. This means that the activity need to be accessed in the DataPlane. This PR remove the limitation on the activity bean.


colesnodgrass (approved):

jdpgrailsdev (changes requested):
The import of @Requires most likely also needs to be removed or PMD will fail.

alovew (approved):

jdpgrailsdev (approved):
:shipit:

PR #20315, opened 2022-12-09
Set breaking change to false when connection is fixed

alovew:
When users update their connection's sync catalog after there is a breaking schema change, the breaking_change field should be set to false

### DIFF
--- a/airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
+++ b/airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
@@ -489,13 +489,13 @@ public class WebBackendConnectionsHandler {
       // Get the most recent actor catalog fetched for this connection's source and the newly updated sync
       // catalog
       Optional<ActorCatalog> mostRecentActorCatalog = configRepository.getMostRecentActorCatalogForSource(originalConnectionRead.getSourceId());
-      AirbyteCatalog newActorCatalog = webBackendConnectionPatch.getSyncCatalog();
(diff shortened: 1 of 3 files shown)


jdpgrailsdev (approved):
:shipit:

airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
```diff
@@ -654,6 +654,11 @@ protected static ConnectionUpdate toConnectionPatch(final WebBackendConnectionUp
                                                       final List<UUID> finalOperationIds) {
     final ConnectionUpdate connectionPatch = new ConnectionUpdate();
 
+    if (webBackendConnectionPatch.getSyncCatalog() != null) {
+      // the user has updated their sync catalog, so there is no breaking change
```
benmoriceau, line 658: Should we check that all the change are now not breaking?

alovew: this was considered in the tech design but it requires getting the source schema and calculating a diff again. what do you think?

alovew: this is a simpler way of doing it based on FE validations (they can't update an invalid sync catalog), but maybe it's better for us to re-validate here too

alovew: just realized the env variable won't work if we do this, since we're marking connections as broken even in the absence of the env variable. so I'll have to update to fetch the most recent actor catalog and do a diff to check whether it's broken

benmoriceau (approved):

PR #20318, opened 2022-12-09
Move dest oauth to micronaut

benmoriceau:
## What
*Convert the Destiantion OAuth controller


airbyte-server/src/main/java/io/airbyte/server/apis/DestinationOauthApiController.java
```diff
@@ -10,26 +10,33 @@
...
+import io.micronaut.context.annotation.Context;
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Post;
 import java.util.Map;
 import javax.ws.rs.Path;
 import lombok.AllArgsConstructor;
 
-@Path("/v1/destination_oauths")
+@Controller("/api/v1/destination_oauths")
 @AllArgsConstructor
```
jdpgrailsdev, line 21: Nit:  we should remove this in favor of an explicit constructor that takes in the `OAuthHandler` as a parameter.  It's not a big deal, but the Lombok generated constructor doesn't allow us to annotate the parameters to the constructor, should we need them for Micronaut dependency injection.

jdpgrailsdev (approved):
:shipit:

airbyte-server/src/main/java/io/airbyte/server/apis/DestinationOauthApiController.java
benmoriceau: Done, do you know how Micronaut works with records?

jdpgrailsdev: I don't, though I would doubt that we would want to make a record a singleton.  Records are generally meant to encapsulate data and do not typically provide any sort of business logic.

airbyte-server/src/main/java/io/airbyte/server/apis/DestinationOauthApiController.java
```diff
@@ -10,26 +10,34 @@
...
+import io.micronaut.http.annotation.Controller;
+import io.micronaut.http.annotation.Post;
 import java.util.Map;
-import javax.ws.rs.Path;
-import lombok.AllArgsConstructor;
 
-@Path("/v1/destination_oauths")
-@AllArgsConstructor
+@Controller("/api/v1/destination_oauths")
+@Context
```
colesnodgrass, line 19: Is `@Context` necessary here?  Per the [docs](https://docs.micronaut.io/latest/api/io/micronaut/context/annotation/Context.html) `WARNING: This annotation should be used sparingly as Micronaut is designed in such a way as to encourage minimal bean creation during startup.`

colesnodgrass (commented):
Does the `AirbyteApiClient` also need to be updated?

airbyte-server/src/main/java/io/airbyte/server/apis/DestinationOauthApiController.java
benmoriceau: @colesnodgrass While testing, I have observed that the first call to the API are longer than the others because of bean initialization. I made the bean to have an eager initialization.

benmoriceau:
> Does the `AirbyteApiClient` also need to be updated?

No, the destiantion OAuth don't have a client

PR #20327, opened 2022-12-09
🪟 🎉 Re-validate connection edit form after schema refresh to highlight any missing keys or cursors

edmundito:
## What
This change ensures that the form highlights any areas that may have gone missing after the schema was refreshed. For example, the primary key and the cursor.

image

## How
This uses a hook component that sets the form as touch and revalidates after the schema has been refreshed. Sets the catalog to touched so that the changes are immediately cancelable.


teallarson (approved):
Looks good.  I suppose at this point the form is "touched".  We could also choose to just [run validation directly](https://formik.org/docs/guides/validation#when-does-validation-run).  I support this way of doing things though!

PR #20328, opened 2022-12-09
build: fix ensure no file change check

c-p-b:
## What
```bash

SUB_BUILD=PLATFORM ./gradlew format --scan --info --stacktrace 

```

needs to be ran on certain db connector pr's.  A PR went in recently that was missing the results of this.

## How
Run above command and commit results

### DIFF
--- a/airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumPropertiesManager.java
+++ b/airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumPropertiesManager.java
@@ -20,6 +20,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class DebeziumPropertiesManager {
+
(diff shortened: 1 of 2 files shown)


edgao (approved):

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumPropertiesManager.java
```diff
@@ -95,14 +96,14 @@ protected Properties getDebeziumProperties() {
...
   public static String getTableIncludelist(final ConfiguredAirbyteCatalog catalog) {
-    // Turn  "stream": {
-    //          "namespace": "schema1"
-    //          "name": "table1
-    //        },
-    //        "stream": {
-    //          "namespace": "schema2"
-    //          "name": "table2
-    //        } -------> info "schema1.table1, schema2.table2"
+    // Turn "stream": {
```
rodireich, line 99: The original is actually what I intended the comment to be, for readability reasons.
I manually reverted it back, which probably contributed to the build breaking.

rodireich (approved):

PR #20335, opened 2022-12-09
🪟 🎨 Add style for disabled state to `<CheckBox />` + replace `styled-components` + add `small` size +  add storybook 

### LINKED ISSUE #20246: Add style for `disabled` state to `<CheckBox />` component
There is no style for the `disabled` state in the `<CheckBox />` component

Need to extend the `<CheckBox />` component to support style for the `disabled` state

Figma:
https://www.figma.com/file/gMNRYVJav25pdsoxAyW7Qp/00_01_WEB-APP-LIBRARIES?node-id=2%3A397&t=wfCZoskkk2Hsnxhn-0

image

dizel852:
## What
Resolves https://github.com/airbytehq/airbyte/issues/20246

## How
Since we don't extend styled components, additional work was also done:
- replace `styled-components` with `css-modules`
- add `small` size variation
- add storybook

Figma:
https://www.figma.com/file/gMNRYVJav25pdsoxAyW7Qp/00_01_WEB-APP-LIBRARIES?node-id=2%3A397&t=2cfAe93e6xY7Gk4L-0

image


airbyte-webapp/src/components/ui/CheckBox/CheckBox.tsx
```diff
@@ -1,49 +1,38 @@
...
-  border-radius: 4px;
-  font-size: 13px;
-  line-height: 13px;
-  display: inline-block;
-  padding: 1px 0;
-  cursor: pointer;
-  vertical-align: top;
-  position: relative;
-`;
+  small?: boolean;
```
YatsukBogdan1, line 10: I think we should consider adding the type `CheckBoxSize` (or something similar; the same we have for `Button` component) in case we will have more sizes in the future, so it will be easier to just additional sizes, instead of refactoring all the places where this property was used before

dizel852: Agree, also was thinking about that. But tried to be in sync with Figma... 
Ok, will change it 👍

airbyte-webapp/src/components/ui/CheckBox/CheckBox.tsx
```diff
@@ -1,49 +1,38 @@
...
+        {
+          [styles.checked]: checked,
+          [styles.indeterminate]: indeterminate,
+          [styles.disabled]: disabled,
+          [styles.small]: small,
+        },
+        className
+      )}
+    >
+      <input type="checkbox" aria-checked={checked} {...inputProps} />
```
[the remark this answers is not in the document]
dizel852: Good catch! Yep, we indicate the `indeterminate` only on UI, but semantically it remains unchanged. Will fix that 👍 Thanks!

airbyte-webapp/src/components/ui/CheckBox/CheckBox.tsx
```diff
@@ -1,49 +1,40 @@
...
-  border-radius: 4px;
-  font-size: 13px;
-  line-height: 13px;
-  display: inline-block;
-  padding: 1px 0;
-  cursor: pointer;
-  vertical-align: top;
-  position: relative;
-`;
+  elSize?: CheckBoxSize;
```
krishnaglick, line 12: Nit:
```suggestion
  checkboxSize?: CheckBoxSize;
```

krishnaglick (approved):
Tested in storybook, LGTM

airbyte-webapp/src/components/ui/CheckBox/CheckBox.tsx
timroes: Could we potentially go with `size` here just? It feels given this is the checkbox component whose size we control, we should not try to prefix it with any other prefix, but just use `size` plainly. This would help us also getting to a bit more aligned property names across components.

dizel852: That was the first I've tried 😄 But unfortunately the native `<input />` has its own `size` property.
We have two options:
1. Omit the native `size` prop and use our one for sizing:
`export interface CheckBoxProps extends Omit<InputHTMLAttributes<HTMLInputElement>, "size"> {
  indeterminate?: boolean;
  size?: CheckBoxSize;
}`
3. Use another prop name for sizing(elSize, checkboxSize, etc.)

@timroes IMO the first option doesn't look dangerous 😬 WDYT?

m19berger (changes requested):
Just a minor thing :wink:

airbyte-webapp/src/components/ui/CheckBox/CheckBox.tsx
timroes: Given that the `size` attribute on an input type="checkbox" has anyway no effect (since it's only for text, search, tel, url, email, and password), I think we could safely omit it, so we can nicely reuse it.

dizel852: Great! Will omit it 👍

dizel852: Did a try to omit the `size` prop, and here is what I've discovered:
On some higher levels, we through the props via `{...props}`
Here is an example:
image

And since we try to change the internal `size` prop, it argues that.

Options we have:
1. Make unavailable to pass props like this way - `{...props}`
2. Use `checkboxSize` as a prop name for sizing

@timroes

dizel852: Don't wanna block the PR because of the prop name, will change it to `checkboxSize`

dizel852:
Updated the PR:
1. Changed prop name from `elSize` to `checkboxSize` @krishnaglick @timroes 
2. Added `aria-checked="mixed"` state for  indeterminate state @YatsukBogdan1 
3. Removed unneeded styled checkbox from Sign Up page, [Figma](https://www.figma.com/file/qDwWxPe8NqUMJlvDMikXCr/01_04_SIGN_UP?t=M005x0GPTWrEDtxE-0) for verification

krishnaglick (approved):
LGTM!

m19berger (approved):
LGTM!

PR #20338, opened 2022-12-09
Bmoric/convert source definition micronaut

benmoriceau:
## What
Convert the source definiton to micronaut.

- Usual files
- Change the `AirbyteProtocolVersionRange` in order to support injection


airbyte-server/src/main/resources/application.yml
```diff
@@ -45,6 +45,9 @@ airbyte:
     url: ${WEBAPP_URL:}
   workspace:
     root: ${WORKSPACE_ROOT}
+  protocol:
+    min-version: ${AIRBYTE_PROTOCOL_VERSION_MIN:`0.0.0`}
```
jdpgrailsdev, line 49: I don't think that we want the ticks in here.  Micronaut will create a string with those in it.  I suspect we just want `0.0.0`, etc.

jdpgrailsdev (approved):
:shipit:

PR #20339, opened 2022-12-10
Migrate to micronaut

benmoriceau:
## What
Convert the destination definition specification controller to micronaut.


jdpgrailsdev (approved):
:shipit:

PR #20349, opened 2022-12-11
emit failure type in attempt_failure_by_origin

cgardens:
## What
* We emit attempt failure by origin (e.g. where in the system it was caused), but we do not include failure type. This PR adds it.
* Adding it because I want to be able to see it in our DD dashboards and be able to cut by failure type.

### DIFF
--- a/airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/JobCreationAndStatusUpdateActivityImpl.java
+++ b/airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/JobCreationAndStatusUpdateActivityImpl.java
@@ -521,11 +521,13 @@ public class JobCreationAndStatusUpdateActivityImpl implements JobCreationAndSta
     if (failureSummary != null) {
       for (final FailureReason reason : failureSummary.getFailures()) {
         MetricClientFactory.getMetricClient().count(OssMetricsRegistry.ATTEMPT_FAILED_BY_FAILURE_ORIGIN, 1,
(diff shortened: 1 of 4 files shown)


airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/JobCreationAndStatusUpdateActivityImpl.java
```diff
@@ -521,11 +521,13 @@ private void trackFailures(final AttemptFailureSummary failureSummary) {
     if (failureSummary != null) {
       for (final FailureReason reason : failureSummary.getFailures()) {
         MetricClientFactory.getMetricClient().count(OssMetricsRegistry.ATTEMPT_FAILED_BY_FAILURE_ORIGIN, 1,
-            new MetricAttribute(MetricTags.FAILURE_ORIGIN, MetricTags.getFailureOrigin(reason.getFailureOrigin())));
+            new MetricAttribute(MetricTags.FAILURE_ORIGIN, MetricTags.getFailureOrigin(reason.getFailureOrigin())),
+            new MetricAttribute(MetricTags.FAILURE_TYPE, MetricTags.getFailureType(reason.getFailureType())));
```
jdpgrailsdev, line 525: Could we also add this to the `traceFailures` method right about the `trackFailures` one?  That would add the information as a facet on the APM trace in addition to the metric.

pmossman (approved):
Looks good to me! (leaving an approval assuming you'll decide whether it's worth it to add the same thing to the `traceFailures` method as per Jonathan's suggestion

PR #20354, opened 2022-12-12
🪟🐛 Connector form: Show success message right away

flash1293:
## What
This PR is fixing a little visual bug when submitting an unsaved connector form. When doing so, the loading bar would be shown while the connection test is running. If the test succeeds, the UI reverts back to the submit button for a short amount of time before changing again to the "connection test succeeded" message:

https://user-images.githubusercontent.com/1508364/207023325-92432931-9bc9-483d-a7bb-a798d4ad9e15.mp4

After this change, it behaves like a user would expect: loading -> success -> navigate away:

https://user-images.githubusercontent.com/1508364/207023480-cd38a757-0827-4e59-86fe-af51f798444c.mp4

## How

This behavior was caused by the fact that the connection success message is only shown if the "hasSuccess" flag  passed down from the page component rendering the form is truthy. However, this flag is not about the connection test, but the success of the request actually saving the connection.

As this is not really related to each other (and doesn't matter for the user), I removed this prop being passed down and only check whether the connection test succeeded.

As a side note: We don't catch errors happening when creating a new source or destination which means they are caught by the default error boundary. This is a bad behavior because it means that the data provided by the user gets lost. I will open a separate issue for this.


lmossman (approved):
Nice refactor, a lot of duplicate code removed and the solution ended up being super simple!

Tested locally and it seems to resolve the issue correctly for me

PR #20356, opened 2022-12-12
🪟🐛 Do not show modal if connector type is changed

### LINKED ISSUE #20332: Confirmation modal is shown upon connector setup if the connector type was changed
## The issue
A confirmation modal is currently being shown after clicking `Set up source` if the source type was changed at any point. Here is how to reproduce:
1. Go to Sources > New Source
2. Choose a source type from the dropdown (e.g. Google Ads)
3. Change the source type to a different one (e.g. Poke API)
4. Type in a Pokemon name (e.g. `ditto`)
5. Click `Set up source`
6. After connection tests pass, a couple seconds later the `Discard changes` confirmation modal will show up

It looks like the source is still being created in this case, so this issue is really just that the confirmation modal is appearing when it shouldn't.

## Expected behavior

Confirmation modal should not appear when saving a connector.

flash1293:
## What
Fixes https://github.com/airbytehq/airbyte/issues/20332

## How

The issue is that form dirty state is tracked like this:
```
{
  "<form id>": true, // true means dirty, false means not dirty
}
```

When changing the connector type, we re-mount the connector form, this means there is a new form id and the old state of the form id stays around. On submit, this is the state:
```
{
  "<old form id>": true,
  "<new form id>": false,
}
```

This PR fixes this by clearing the dirty flag of the current form in the `FormChangeTracker` if the id changes or the component gets unmounted. The other option would be to solve it explicitly in the source and destination view components, but this seems general enough to make it part of this helper.

To actually make this work, I needed to optimize the `useFormChangeTrackerService` a little bit. It would write to a single global state object which would cause the mutation functions `trackFormChange`, `clearFormChange` and so on to be re-created every time a dirty state flag is actually changed. This PR keeps the function references stable by using a ref on the global state. This is a best practice anyways - function references that change frequently are often the cause of unintended cascading updates and re-renders.


airbyte-webapp/src/hooks/services/FormChangeTracker/hooks.ts
```diff
@@ -22,18 +24,18 @@ export const useFormChangeTrackerService = (): FormChangeTrackerServiceApi => {
...
   const trackFormChange = useCallback(
     (id: string, changed: boolean) => {
-      if (Boolean(changedFormsById?.[id]) !== changed) {
-        setChangedFormsById({ ...changedFormsById, [id]: changed });
+      if (Boolean(changedFormsByIdRef.current?.[id]) !== changed) {
+        setChangedFormsById({ ...changedFormsByIdRef.current, [id]: changed });
       }
     },
-    [changedFormsById, setChangedFormsById]
+    [changedFormsByIdRef, setChangedFormsById]
```
krishnaglick, line 38: I feel like this change might be excessive? But not sure. If we need to store the value in a ref maybe the `useChangedFormsById` should be a ref instead?

flash1293: It's important because otherwise it's easy to run into endless update cycles (you have an effect that needs the reference of the latest trackFormChange, which updates the changedFormsById, which updates the reference of trackFormChange, which re-triggers the effect and so on)

changedFormsById could probably be a ref, but it seems like it's important that it's global state and there is no createGlobalRef, so that seems like the cleanest solution to me.

Happy to implement another path if you want though.

krishnaglick: I don't want to hold up this PR and if this fixes it I'm fine with it.
I'd need to dig into the code again to see what I can come up with. Perhaps all this could be streamlined but would require investigation.

flash1293: It would be nice if we would never actually need multi-form tracking, that would allow us to throw away most of this code. it felt out of scope for this PR though.

airbyte-webapp/src/hooks/services/FormChangeTracker/hooks.ts
```diff
@@ -10,6 +10,8 @@ export const useUniqueFormId = (formId?: string) => useMemo(() => formId ?? uniq
 
 export const useFormChangeTrackerService = (): FormChangeTrackerServiceApi => {
   const [changedFormsById, setChangedFormsById] = useChangedFormsById();
+  const changedFormsByIdRef = useRef(changedFormsById);
+  changedFormsByIdRef.current = changedFormsById;
```
lmossman, line 14: Is this line actually necessary? The [docs](https://reactjs.org/docs/hooks-reference.html#useref) say:
> useRef returns a mutable ref object whose .current property is initialized to the passed argument (initialValue)

lmossman: Oh, is this necessary because the `changedFormsById` object will be modified from one render to the next, so this line updates the `current` pointer to the new object whenever it changes?

lmossman (approved):
Had a question about the refs, but otherwise I think the solution makes sense.

Tested locally and it seems to have solved the issue for connector form. I also tested out the connection form and it doesn't seem to have regressed.

airbyte-webapp/src/hooks/services/FormChangeTracker/hooks.ts
flash1293: > so this line updates the current pointer to the new object whenever it changes

Exactly, when the dirty state changes, the whole object gets replaced here:
https://github.com/airbytehq/airbyte/blob/b72cd7150ddce3caa5290953f90aec3822bf53aa/airbyte-webapp/src/hooks/services/FormChangeTracker/hooks.ts#L25

So in order to not bust the `useCallbacks` we need to ref it.

PR #20360, opened 2022-12-12
🪟 🎉 Add search capability for existing sources and destinations

### LINKED ISSUE #18276: Add search capability for existing sources
On the /new-connection page, the user is prompted to select an existing source or create a new one.

When selecting an existing source, the user is unable to search to find the relevant existing source. They should be able to search similar to the connector type selection. Same applies for the destination. 

Screen Shot 2022-10-20 at 4 09 59 PM

cc @andyjih  @edmundito

YatsukBogdan1:
## What
Closes #18276 

## How
Adds is `isSearchable=true` property to `DropDown` inside `ExistingEntityForm` component. Tested and seems all logic is already handled inside `DropDown` component

## Example
https://user-images.githubusercontent.com/25852567/207052850-0ac84319-5fb9-4729-b971-6cf60a72ae97.mov


krishnaglick (approved):
Works locally!

dizel852 (approved):

PR #20363, opened 2022-12-12
Document IntelliJ exclusion manual workaround

maxi297:
## What
The steps for excluding files sometimes does not work in IntelliJ. The reason is not clear but @erohmensing found a manual workaround that users can leverage to alleviate the issue.

## How
Improving documentation to share the information on https://docs.airbyte.com/contributing-to-airbyte/monorepo-python-development/#excluding-files-from-venv


erohmensing (approved):
Helpful in any case!

Maybe @girarda has an idea what changed? Looks like he is the one who wrote those instructions

maxi297:
> Helpful in any case!
> 
> Maybe @girarda has an idea what changed? Looks like he is the one who wrote those instructions

Alexandre doesn't have this insights unfortunately. Let's merge and if we get more information on this, let's update the documentation.

PR #20364, opened 2022-12-12
fix: remove retry attempt from "publish connector image" step

Mykyta-Serbynevskyi:
This PR removes retry attempts for "Publish connector" step in [Publish connector image workflow](https://github.com/airbytehq/airbyte/blob/master/.github/workflows/publish-command.yml) due to confrontation with [fail-fast = false](https://github.com/airbytehq/airbyte/blob/21e7290eb6cd0a328e4f11b26bb5e29c6168080d/.github/workflows/publish-command.yml#L227) github actions logic.

Example - https://github.com/airbytehq/airbyte/pull/19610#issuecomment-1343269603

### DIFF
--- a/.github/workflows/publish-command.yml
+++ b/.github/workflows/publish-command.yml
@@ -307,19 +307,15 @@ jobs:
           attempt_delay: 5000 # in ms
           command: curl -sL https://sentry.io/get-cli/ | bash || echo "sentry cli already installed"
       - name: Publish ${{ matrix.connector }}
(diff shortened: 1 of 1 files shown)


supertopher (approved):
let's try this, but I'm not super convinced this is an ideal state.  if we see places where retries would have saved us we might revisit it

Mykyta-Serbynevskyi:
Closing w/o merging as not actual anymore.

PR #20369, opened 2022-12-12
implement column filtering in the replication workflow

mfsiega-airbyte:
## What
Implement column selection in the replication workflow.

It is behind a feature flag, default disabled.

## How
If field selection is enabled, only retain the fields in the message from the source that are explicitly included in the catalog.

## Recommended reading order
1. `airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java`
2. Everything else.

### DIFF
--- a/airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
+++ b/airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
@@ -10,7 +10,9 @@ import static io.airbyte.metrics.lib.ApmTraceConstants.Tags.JOB_ROOT_KEY;
 import static io.airbyte.metrics.lib.ApmTraceConstants.WORKER_OPERATION_NAME;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import datadog.trace.api.Trace;
 import io.airbyte.commons.io.LineGobbler;
 import io.airbyte.config.FailureReason;
@@ -28,6 +30,7 @@ import io.airbyte.protocol.models.AirbyteMessage;
 import io.airbyte.protocol.models.AirbyteMessage.Type;
 import io.airbyte.protocol.models.AirbyteRecordMessage;
 import io.airbyte.protocol.models.AirbyteStreamNameNamespacePair;
+import io.airbyte.protocol.models.ConfiguredAirbyteCatalog;
 import io.airbyte.workers.RecordSchemaValidator;
 import io.airbyte.workers.WorkerMetricReporter;
 import io.airbyte.workers.WorkerUtils;
@@ -41,6 +44,7 @@ import io.airbyte.workers.internal.AirbyteSource;
 import io.airbyte.workers.internal.book_keeping.MessageTracker;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -94,6 +98,7 @@ public class DefaultReplicationWorker implements ReplicationWorker {
   private final AtomicBoolean hasFailed;
   private final RecordSchemaValidator recordSchemaValidator;
   private final WorkerMetricReporter metricReporter;
+  private final boolean fieldSelectionEnabled;
 
   public DefaultReplicationWorker(final String jobId,
                                   final int attempt,
@@ -102,7 +107,8 @@ public class DefaultReplicationWorker implements ReplicationWorker {
                                   final AirbyteDestination destination,
                                   final MessageTracker messageTracker,
                                   final RecordSchemaValidator recordSchemaValidator,
-                                  final WorkerMetricReporter metricReporter) {
+                                  final WorkerMetricReporter metricReporter,
+                                  final boolean fieldSelectionEnabled) {
     this.jobId = jobId;
     this.attempt = attempt;
     this.source = source;
@@ -112,6 +118,7 @@ public class DefaultReplicationWorker implements ReplicationWorker {
     this.executors = Executors.newFixedThreadPool(2);
     this.recordSchemaValidator = recordSchemaValidator;
     this.metricReporter = metricReporter;
+    this.fieldSelectionEnabled = fieldSelectionEnabled;
 
     this.cancelled = new AtomicBoolean(false);
     this.hasFailed = new AtomicBoolean(false);
@@ -198,8 +205,9 @@ public class DefaultReplicationWorker implements ReplicationWorker {
           });
 
       final CompletableFuture<?> readSrcAndWriteDstThread = CompletableFuture.runAsync(
-          readFromSrcAndWriteToDstRunnable(source, destination, cancelled, mapper, messageTracker, mdc, recordSchemaValidator, metricReporter,
-              timeTracker),
+          readFromSrcAndWriteToDstRunnable(source, destination, sourceConfig.getCatalog(), cancelled, mapper, messageTracker, mdc,
+              recordSchemaValidator, metricReporter,
+              timeTracker, fieldSelectionEnabled),
           executors)
           .whenComplete((msg, ex) -> {
             if (ex != null) {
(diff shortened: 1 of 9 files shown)


airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -549,6 +566,30 @@ private static void validateSchema(final RecordSchemaValidator recordSchemaValid
     }
   }
 
+  private static void populatedStreamToSelectedFields(ConfiguredAirbyteCatalog catalog,
```
pmossman, line 569: probably worth some javadoc with a brief explanation of why this achieves column selection, ie "The configured catalog only contains fields that were selected, so by iterating through each field for each stream, we identify just the fields that were selected"

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -549,6 +566,30 @@ private static void validateSchema(final RecordSchemaValidator recordSchemaValid
...
+      if (propertiesNode.isObject()) {
+        propertiesNode.fieldNames().forEachRemaining((fieldName) -> selectedFields.add(fieldName));
+      } else {
+        throw new RuntimeException("No properties node in stream schema");
+      }
+      streamToSelectedFields.put(AirbyteStreamNameNamespacePair.fromConfiguredAirbyteSteam(s), selectedFields);
+    }
+  }
+
+  private static void filterSelectedFields(Map<AirbyteStreamNameNamespacePair, List<String>> streamToSelectedFields, AirbyteMessage airbyteMessage) {
```
pmossman, line 583: nit: args can be `final`

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -549,6 +566,30 @@ private static void validateSchema(final RecordSchemaValidator recordSchemaValid
...
+    final AirbyteRecordMessage record = airbyteMessage.getRecord();
```
pmossman, line 584: Similar to what happens in the private method `validateSchema` in this class, probably worth an early return if `message.getRecord() == null`. In fact I wonder if this current implementation would break if there were `airbyteMessages` that weren't `records` but something else like `state`?

pmossman (commented):
@mfsiega-airbyte high level approach looks solid to me! I left some comments. Let me know when you've added tests and I can do another review with an explicit approval. Also, were you planning on adding an Acceptance Test in this PR as well?

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -549,6 +566,30 @@ private static void validateSchema(final RecordSchemaValidator recordSchemaValid
...
+    final AirbyteStreamNameNamespacePair messageStream = AirbyteStreamNameNamespacePair.fromRecordMessage(record);
+    final List<String> selectedFields = streamToSelectedFields.getOrDefault(messageStream, Collections.emptyList());
+    final JsonNode data = record.getData();
+    if (data.isObject()) {
```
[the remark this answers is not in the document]
mfsiega-airbyte: This could be false in a case where record looks like `{"data": "a-string-and-not-an-object"}`. Any simple type would meet the requirement that it's a non-null JsonNode.

However, the docs say: `The data contains the record data and must always be populated. It is a JSON blob.` I think it's fair to interpret "JSON blob" as object here; so I'm comfortable throwing a RuntimeException if it happens.

airbyte-commons/src/main/java/io/airbyte/commons/features/EnvVariableFeatureFlags.java
```diff
@@ -47,6 +48,11 @@ public boolean needStateValidation() {
     return getEnvOrDefault(NEED_STATE_VALIDATION, true, Boolean::parseBoolean);
   }
 
+  @Override
+  public boolean applyFieldSelection() {
+    return getEnvOrDefault(APPLY_FIELD_SELECTION, true, Boolean::parseBoolean);
```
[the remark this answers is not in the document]
mfsiega-airbyte: I think I unthinkingly imitated one of the other flags, this should definitely start off false.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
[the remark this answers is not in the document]
pmossman: Ah hm, I forgot that a plain String could be a valid JsonNode. I wonder if that's a valid use case? Maybe we can ask someone who is more familiar with the Airbyte protocol if it seems like a good idea to throw a RuntimeException if we encounter a string in a record's `data` field. @gosusnp perhaps?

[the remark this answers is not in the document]
mfsiega-airbyte: The source acceptance tests seem to confirm that records must always have some key/value pairs: see https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py#L475 and https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/utils/json_schema_helper.py#L147 (though I could well be reading the tests wrong).

mfsiega-airbyte:
@pmossman I think this should be ready for review, marked it as such! The acceptance test passes locally; if I have issues getting it to pass in CI I'd say go ahead without it in this PR (since the feature is disabled by default) and I can work it out after.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -549,6 +575,47 @@ private static void validateSchema(final RecordSchemaValidator recordSchemaValid
...
+                                           final AirbyteMessage airbyteMessage) {
+    final AirbyteRecordMessage record = airbyteMessage.getRecord();
+
+    if (record == null) {
+      // This isn't a record message, so we don't need to do any filtering.
+      return;
+    }
+
+    final AirbyteStreamNameNamespacePair messageStream = AirbyteStreamNameNamespacePair.fromRecordMessage(record);
+    final List<String> selectedFields = streamToSelectedFields.getOrDefault(messageStream, Collections.emptyList());
```
mfsiega-airbyte, line 610: One small note here - the spike implementation I used for my loose performance testing was re-creating this list of fields by iterating through the catalog for every single record. This should be faster.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -549,6 +575,47 @@ private static void validateSchema(final RecordSchemaValidator recordSchemaValid
...
+    final JsonNode data = record.getData();
+    if (data.isObject()) {
+      ((ObjectNode) data).retain(selectedFields);
+    } else {
+      throw new RuntimeException("Unexpected data in record");
```
pmossman, line 615: Probably a good idea to include the record's body in this error message because if it ever happens, I think we'll want to know what the record looked like!

airbyte-commons-worker/src/test/java/io/airbyte/workers/general/DefaultReplicationWorkerTest.java
```diff
@@ -334,6 +336,68 @@ void testOnlyStateAndRecordMessagesDeliveredToDestination() throws Exception {
     verify(destination, never()).accept(TRACE_MESSAGE);
   }
 
+  @Test
```
pmossman, line 339: Nice tests!

tools/bin/acceptance_test.sh
```diff
@@ -21,7 +21,7 @@ check_success() {
 echo "Starting app..."
 
 # Detach so we can run subsequent commands
-VERSION=dev TRACKING_STRATEGY=logging USE_STREAM_CAPABLE_STATE=true BASIC_AUTH_USERNAME="" BASIC_AUTH_PASSWORD="" docker-compose -f docker-compose.yaml -f docker-compose.acceptance-test.yaml up -d
+VERSION=dev TRACKING_STRATEGY=logging USE_STREAM_CAPABLE_STATE=true BASIC_AUTH_USERNAME="" BASIC_AUTH_PASSWORD="" APPLY_FIELD_SELECTION=true docker-compose -f docker-compose.yaml -f docker-compose.acceptance-test.yaml up -d
```
pmossman, line 24: I'd add a quick comment about `APPLY_FIELD_SELECTION` here, explaining that it's currently behind a feature flag that defaults to false

pmossman (approved):
Looks great! Left a couple tiny suggestions for comments/error message but looks mergeable to me!

benmoriceau (commented):
LGTM for adding the feature flag.

PR #20370, opened 2022-12-12
🪟 🎨 Fix endOfLine formatting with prettier

joeykmh:
## What
There are a few files that do not appear to have been run through prettier, or at least were formatted with different rules. I noticed this when they showed up in a diff in an [unrelated PR](https://github.com/airbytehq/airbyte/pull/20267#discussion_r1044556330).

image

This PR fixes the formatting of these three files and adds an explicit [endOfLine](https://prettier.io/docs/en/options.html#end-of-line) rule to our `.prettierrc.js` config file.

## 🚨 User Impact 🚨
None!


m19berger (approved):
Thanks for fix! LGTM

krishnaglick (approved):

PR #20373, opened 2022-12-12
PR builds run diff check after formatting

edgao:
Make formatting changes fail faster in the `build-connectors-base` pr check action. `platform-build` already does this.


supertopher (approved):
Much love for more clear action names.  thanks for the add Ed

fail fast!

PR #20377, opened 2022-12-12
🪟 🐛 Connection Name styles match other fields on Connection create

### LINKED ISSUE #20194: Connection name label in connection create page is larger than other labels
## Current Behavior
Screen Shot 2022-12-07 at 11 14 52

## Expected Behavior
The size of the connection name label should be the same as the rest of the labels in the form

## Steps to Reproduce
1. Go to Create a new connection
2. Select a source and a destination
3. Look at the name field at the top

teallarson:
## What
closes #20194 

BEFORE:
Screenshot 2022-12-12 at 12 23 03 PM

AFTER:
Screenshot 2022-12-12 at 12 24 45 PM

## How
Uses our built in headers rather than custom styles.  Gives this section a header so it has parity with the others.

## Recommended reading order
any


krishnaglick (approved):
LGTM, did not test locally

edmundito:
Test snapshots need to be updated.

airbyte-webapp/src/components/CreateConnection/CreateConnectionNameField.tsx
```diff
@@ -13,20 +12,15 @@ export const CreateConnectionNameField = () => {
   const { formatMessage } = useIntl();
 
   return (
-    <Section>
+    <Section title={<FormattedMessage id="connection.title" />}>
```
edmundito, line 15: This needs to be reviewed with @Upmitt. The reason why this was set up this way in the first place was to match a design. However, I'm not sure where the latest design is located.

edmundito (approved):
The code looks good. I did not test locally.

PR #20387, opened 2022-12-12
🪟 🧹 Match sync catalog items by name and namespace, not their index

teallarson:
## What
Match by stream name and namespace instead of by syncCatalog index.

Fixes a bug when editing a connection where adding or removing a stream from the source, refreshing the source schema, then canceling the changes in the form would throw an error in `<CatalogTree/>`

## How
Name and namespace are more stable matchers than index.  The array may change based on refreshes, making it not a stable reference.

### DIFF
--- a/airbyte-webapp/src/components/connection/CatalogTree/CatalogTree.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogTree/CatalogTree.tsx
@@ -57,19 +57,12 @@ const CatalogTreeComponent: React.FC<React.PropsWithChildren<CatalogTreeProps>> 
   const changedStreams = useMemo(
     () =>
       streams.filter((stream) => {
-        if (!stream.config) {
-          return false;
-        }
-        const matchingInitialValue = initialValues.syncCatalog.streams.filter(
+        const matchingInitialValue = initialValues.syncCatalog.streams.find(
           (initialStream) =>
             initialStream.stream?.name === stream.stream?.name &&
             initialStream.stream?.namespace === stream.stream?.namespace
-        )[0];
-        if (!matchingInitialValue) {
-          return false;
-        }
-
-        return stream.config?.selected !== matchingInitialValue.config?.selected;
+        );
+        return stream.config?.selected !== matchingInitialValue?.config?.selected;
       }),
     [initialValues.syncCatalog.streams, streams]
   );
--- a/airbyte-webapp/src/components/connection/CatalogTree/CatalogTree.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogTree/CatalogTree.tsx
@@ -56,8 +56,20 @@ const CatalogTreeComponent: React.FC<React.PropsWithChildren<CatalogTreeProps>> 
 
   const changedStreams = useMemo(
     () =>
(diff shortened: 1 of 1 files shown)


airbyte-webapp/src/components/connection/CatalogTree/CatalogTree.tsx
```diff
@@ -56,8 +56,20 @@ const CatalogTreeComponent: React.FC<React.PropsWithChildren<CatalogTreeProps>>
...
+        const matchingInitialValue = initialValues.syncCatalog.streams.filter(
+          (initialStream) =>
+            initialStream.stream?.name === stream.stream?.name &&
+            initialStream.stream?.namespace === stream.stream?.namespace
+        )[0];
+        if (!matchingInitialValue) {
+          return false;
+        }
+
+        return stream.config?.selected !== matchingInitialValue.config?.selected;
```
edmundito, line 72: Can be checked in the same return statement: `matchingInitialValue?.config?.selected`

edmundito (commented):
The solution looks correct, but the code can be streamlined

airbyte-webapp/src/components/connection/CatalogTree/CatalogTree.tsx
```diff
@@ -56,8 +56,13 @@ const CatalogTreeComponent: React.FC<React.PropsWithChildren<CatalogTreeProps>>
 
   const changedStreams = useMemo(
     () =>
-      streams.filter((stream, idx) => {
-        return stream.config?.selected !== initialValues.syncCatalog.streams[idx].config?.selected;
+      streams.filter((stream) => {
+        const matchingInitialValue = initialValues.syncCatalog.streams.find(
+          (initialStream) =>
```
tealjulia, line 61: We also match by id (which is the stringified index of the item in the array) a little higher up, and in the `BulkEditService`.  TBD whether those should also be changed to use name & namespace as matchers.  I am unsure if either would get called in a case where using the index would be unreliable.

edmundito: I had the same thought as well when I was looking at this. I wonder if we can trust the ID, but in this solution we know we can trust name + namespace as a matching criteria so this is less risky.

edmundito (approved):
So fresh and so clean

airbyte-webapp/src/components/connection/CatalogTree/CatalogTree.tsx
tealjulia: Do you think it's worth making a ticket to audit other uses of index/stream id?  I'm not _sure_ they're a problem, but I'm not sure they're _not_ a problem.

edmundito: I think it's good to track this in a ticket. If the id proves to be trustworthy, we can simplify a lot of checks in our code.

tealjulia: Ope.  I hit merge.  The id is the item's index in the array converted to a string, so it would presumably have the same problems.

PR #20404, opened 2022-12-12
Java connectors use protocol v0 objects

### LINKED ISSUE #20037: Have Java Connectors depend on v0 of the protocol.
## What

Decouple the Java Connectors and the Platform on protocol version to allow them to effectively use different versions.
We need to introduce an explicit dependency of the Java Connectors on the v0 objects of the protocol rather than use the objects without an explicit version in the namespace.

## How

* Move Java connectors dependency on `io.airbyte.protocol.models.AirbyteMessage` to `io.airbyte.protocol.models.v0.AirbyteMessage`.
* Some shared common utils are currently using the versionless objects, we need to add a version with an explicit version.

gosusnp:
## What

Migrate the Java connectors to use the protocol objects in the v0 namespace (`io.airbyte.protocol.models.v0`).

Closes #20037 

## How

* Duplicate some helper objects into `io.airbyte.protocol.models.v0`
* Update imports for the Java connectors
* Add object conversions in some shared test objects of the Java connectors to keep the build happy

The object conversions is a short term mitigation for the build, a better solution should be implemented when we start migrating some connectors to the v1 of the protocol.

## Recommended reading order
1. `airbyte-protocol/protocol-models/src/main/java/io/airbyte/protocol/models/v0` contain the helper objects that have been duplicated to be version specific
2. `airbyte-integrations/bases/standard-source-test/src/main/java/io/airbyte/integrations/standardtest/source/AbstractSourceConnectorTest.java`, `airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java` and `airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/LocalAirbyteDestination.java` contain the tests that have been updated with the object conversion
3. the rest should be namespacing updates to reference `v0` objects rather than version-less objects.

## 🚨 User Impact 🚨

There should be no behavior change, the `v0` and `v1` of the protocol are structurally identical, what differs is the interpretation of some fields.


airbyte-protocol/protocol-models/src/main/java/io/airbyte/protocol/models/v0/CatalogHelpers.java
```diff
@@ -0,0 +1,478 @@
...
+  /**
+   * Maps a list of fields into a JsonSchema object with names and types. This method will throw if it
+   * receives multiple fields with the same name.
+   *
+   * @param fields fields to map to JsonSchema
+   * @return JsonSchema representation of the fields.
+   */
+  public static JsonNode fieldsToJsonSchema(final List<Field> fields) {
+    return Jsons.jsonNode(ImmutableMap.builder()
+        .put("type", "object")
```
benmoriceau, line 209: Nit: those are defined in the file `JsonSchemas.java`. If the dependency doesn't exist I don't think it worth adding it for this.

benmoriceau (approved):

airbyte-protocol/protocol-models/src/main/java/io/airbyte/protocol/models/v0/CatalogHelpers.java
gosusnp: So the dependencies was already there, given the pretty broad scope of this PR already, I avoided any refactoring to alter dependencies.
I have seen some parts that could be simplified or some objects that are redundant. If we want to improve those areas, they should be tackled separately.

edgao (approved):
lgtm. I kicked off integration tests just in case, but expecting them to Just Work ™️  (feel free to cancel them if you already ran these locally)

akashkulk (approved):

gosusnp:
@edgao, @akashkulk, while looking into the connector failure, found extra references to the v-null objects.
Most of the fixes should be purely namespace changes, one to be aware of is [AbstractDbSource.java](https://github.com/airbytehq/airbyte/pull/20404/files/caac0ef5eb5ec552c8b508cfc26c406109c8eb70..0d33ee8293168c5648511d9e4ae3b8d3729b7bb1#diff-20db3421275ed851d9374c4f7d9e24c8c168025c0fe1d467bc8e8e12e2a924b6) where I added a conversion for the state messages. I think it is the first conversion that is happening in the main code path (as opposed to tests only for the others)

edgao (approved):
thanks for the heads up - not expecting anything to break here 🤞hopefully the tests pass

sherifnada:
@gosusnp review is requested from the extensibility team but this doesn't seem to touch any extensibility code. Was that review request intentional?

akashkulk:
Looks good - I assume that was why the Postgres test was failing? Just kicked off another run there

gosusnp:
@sherifnada, that was not intentional, I suspect some change detection script ([Check if a review is required from Connector teams](https://github.com/airbytehq/airbyte/actions/runs/3699689622/jobs/6267312413#logs) maybe?) automatically tagged extensibility

gosusnp:
@akashkulk, I think https://github.com/airbytehq/airbyte/pull/20404/commits/b16f07da893c6bc5044bf9be563b024b04c39209 should fix the issues in the source-postgres. Just kicked off another run to confirm this.

PR #20375, opened 2022-12-12
Enable low-code CDK users to specify schema in the manifest

clnoll:
## What
Allows users of the low-code CDK to define stream schemas inline in the manifest.

## How
Adds a new `StaticSchemaLoader` to the list of schema loader options. When a user specifies that the `StaticSchemaLoader` should be used with a stream, they must also supply the `schema` property.
```
  applications_stream:
    $ref: "*ref(definitions.base_incremental_stream)"
    $options:
      name: "applications"
      path: "applications"
    stream_cursor_field: "applied_at"
    schema_loader:
      type: "StaticSchemaLoader"
      schema: {...}
```

## Recommended reading order
1. [airbyte-integrations/connectors/source-greenhouse/source_greenhouse/greenhouse.yaml](https://github.com/airbytehq/airbyte/pull/20405)
2. `airbyte-cdk/python/airbyte_cdk/sources/declarative/schema/static_schema_loader.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/declarative/config_component_schema.json`
    `airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/class_types_registry.py`
    `airbyte-cdk/python/airbyte_cdk/sources/declarative/schema/__init__.py`
    `airbyte-cdk/python/unit_tests/sources/declarative/schema/test_static_schema_loader.py`

## 🚨 User Impact 🚨
No breaking changes; the output of `discovery` is exactly the same with inline schema versus on-disk schema.

## Pre-merge Checklist

- [X] Unit & integration tests added and passing
- [X] Code reviews completed
- [x] Documentation updated
- [X] PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests)
- [X] Issue acceptance criteria met

## Tests

<details><summary><strong>Unit</strong></summary>

*Put your unit tests output here.*

</details>

<details><summary><strong>Integration</strong></summary>

*Put your integration tests output here.*

</details>

<details><summary><strong>Acceptance</strong></summary>

*Put your acceptance tests output here.*

</details>

Issue #19499, opened 2022-11-16
[connector builder]: Support declaring & exporting stream schemas in the UI


sherifnada:
## Tell us about the problem you're trying to solve
a user of the connector builder UI needs the ability to specify stream schemas in the UI such that when they click "export connector" the schemas they specified in the UI are the same ones that actually run in the connector itself outside of the connector builder. 

## Implementation options

**Bundle into manifest**
The most obvious way to achieve this is to bundle schema definitions into the manifest itself. This way the user maintains only one file they download from the UI, which they copy-paste into their connector directory to build the connector. 

The pro of this approach is that it is the most portable/lightweight
The con is that it makes the YAML file very large/maybe unreadable

Implementation steps:
1. Create `StaticSchemaLoader` class which takes a schema mapping object as parameter
2. Add `StaticSchemaLoader` to the class type registry
3. Update one of the connectors to use the new schema definition:
```
  applications_stream:
    $ref: "*ref(definitions.base_incremental_stream)"
    $options:
      name: "applications"
      path: "applications"
    stream_cursor_field: "applied_at"
    schema_loader:
      type: "StaticSchema"
      schema:
        status:
          type: string
        source:
          type: object
          properties:
            public_name:
              type: string
            id:
              type: integer
    <...>
    retriever:
      $ref: "*ref(definitions.retriever)"
      requester: "*ref(definitions.requester)"
      stream_slicer:
        request_cursor_field: "created_after"
        cursor_field: "applied_at"
        class_name: source_greenhouse.components.GreenHouseSlicer
```

**Continue using `schemas/` folders pattern**
Instead of bundling, we could also change the behavior of the "export YAML" button to download instead a zip folder which contains the YAML manifest in addition to a `schemas/` folder, containing one `.json` schema file per stream. This would mirror exactly the way we typically specify schemas today. 

The pro of this approach is that it keeps the YAML file clean
The con of this approach is that it complicates how a user consumes a downloaded connector

### Acceptance Criteria
- [x] Follow implementation steps for "bundle" option 1 above
- [x] At least 1 existing connector defines its stream schemas inline
- [x] The output of the discover method must be unchanged for connectors we migrated

Out of scope:
- Schema auto-detection 

### Open questions
- do we want to rename `SchemaLoader` to `Schema` since we're not always loading it from a file?

girarda:
notes:
- UX feels worse using a zip file
- moving the spec to the manifest was well received by early users

clnoll:
[...]

PR #20405, opened 2022-12-12
Update source-gnews manifest to use inline stream schemas

clnoll:
## What
Updates the `source-gnews` low-code connector to include the stream schema inline in the manifest.
See https://github.com/airbytehq/airbyte/issues/19499.

This is blocked by https://github.com/airbytehq/airbyte/pull/20375.

TODO: update the `airbyte-cdk` version used by `source-gnews` once https://github.com/airbytehq/airbyte/pull/20375 has been released.

## How
- Updates `gnews.yaml` to include a yaml version of the schema previously stored in the `schemas` directory.
- Deletes the existing schema files.

## Recommended reading order
1. `gnews.yaml`
2. https://github.com/airbytehq/airbyte/pull/20375

## 🚨 User Impact 🚨
No breaking changes; the output of the `discover` method is the same with the inline stream schemas as it was with the on-disk schema files.

## Pre-merge Checklist

- [X] Unit & integration tests passing.
- [x] Code reviews completed
- [X] Documentation updated
    - [X] Changelog updated in `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
- [x] New Connector version released on Dockerhub and connector version bumped by running the `/publish` command described [here](https://docs.airbyte.io/connector-development#updating-an-existing-connector)

## Tests

</details>

</details>

</details>


airbyte-integrations/connectors/source-gnews/source_gnews/gnews.yaml
```diff
@@ -1,5 +1,109 @@
 version: "0.1.0"
 
+schemas:
+  search_stream_schema:
+    "$schema": http://json-schema.org/draft-07/schema#
```
girarda, line 5: do we need this line? I think it makes sense at the top level of a yaml file to specify the version of jsonschema to use to validate, but I doubt it does anything here

clnoll: Ah, that can be removed. Just kept it in so I could diff the output of `discover` before and after the schema was included inline in the manifest.

clnoll: @girarda just pushed the update removing these lines. Did you want to review the change?

airbyte-integrations/connectors/source-gnews/source_gnews/gnews.yaml
```diff
@@ -44,7 +146,12 @@ definitions:
     nullable: "{{ ','.join(config['nullable']) }}"
     from: "{{ stream_slice['start_time'] }}"
     to: "{{ stream_slice['end_time'] }}"
+  schema_loader:
+    type: StaticSchemaLoader
```
brianjlai, line 150: This should be `InlineSchemaLoader` correct?

airbyte-integrations/connectors/source-gnews/setup.py
```diff
@@ -6,7 +6,7 @@
 from setuptools import find_packages, setup
 
 MAIN_REQUIREMENTS = [
-    "airbyte-cdk~=0.1",
+    "airbyte-cdk~=0.13.2",
```
brianjlai, line 9: We should remove the patch version from here. Right now, this will cause us to only pull the `airbyte-cdk` from `0.13.x` onward as per [PEP 440](https://peps.python.org/pep-0440/#compatible-release). Whereas ideally we probably want to continue to get minor bumps as the CDK continues to be changed in a non-breaking way.

airbyte-integrations/connectors/source-gnews/source_gnews/gnews.yaml
```diff
@@ -1,5 +1,107 @@
 version: "0.1.0"
 
+schemas:
```
brianjlai, line 3: nit, but can we move the schemas to the bottom of the file? i think the definitions and streams are likely to be the parts of the manifest that change most, so it's more intuitive to have them at the top.

airbyte-integrations/connectors/source-gnews/source_gnews/gnews.yaml
clnoll: 😬 yep! Fixed.

airbyte-integrations/connectors/source-gnews/setup.py
clnoll: @brianjlai this change does rely on CDK version 0.13.2 or greater. WDYT about changing it to `airbyte-cdk>=0.13.2`?

airbyte-integrations/connectors/source-gnews/source_gnews/gnews.yaml
clnoll: @brianjlai I'm not super familiar with yaml but am not seeing a way to do forward referencing. From what I've found the anchor needs to come before the reference. Or are you aware of a syntax that I can use to do this?

airbyte-integrations/connectors/source-gnews/setup.py
brianjlai: I agree that we should definitely pin it to use `0.13` or above since it relies on a new feature, but the problem with explicitly marking it with the patch version `2` is that next time we build, we will only pull `airbyte-cdk` where minor version is `13`. So in the next month if we progress to CDK `0.20.0`, this connector will continue to be pinned at `0.13.x` since it will only look for the latest patch. And I think based on PEP 440 `~=` and `>=` are interchangeable. Basically I think we want to continue to get latest minors and specifying a patch in setup.py will inhibit that.

And actually given that this is a new non-breaking feature, we should probably also be bumping this to `0.14.0`.

clnoll: Got it, thanks for clarifying. Sounds like I should have bumped the CDK change to `0.14.0`. Looks like that does exist now so I'll change this to use `airbyte-cdk~=0.14`.

brianjlai: ah sorry my bad, I was getting some PRs mixed. Up this was more that in hindsight that we should've minor version bumped the cdk schema change to 14 since it was a feature add. But regardless `0.14.0` is still fine since that was released earlier today.

airbyte-integrations/connectors/source-gnews/source_gnews/gnews.yaml
brianjlai: ah sorry you are correct and my mistake. I thought we did some preparsing to allow for forward references even if defined after, but doesn't seem like we do, so it's fine to have this at the top.

clnoll: No worries (sorry, deleted my comment because I thought maybe I'd spoken prematurely and was looking more closely at the issue). Going to write up an issue for updating the parser to handle forward references, since I do think that will make the yaml more readable.

airbyte-integrations/connectors/source-gnews/source_gnews/gnews.yaml
```diff
@@ -44,7 +146,12 @@ definitions:
     nullable: "{{ ','.join(config['nullable']) }}"
     from: "{{ stream_slice['start_time'] }}"
     to: "{{ stream_slice['end_time'] }}"
+  schema_loader:
+    type: InlineSchemaLoader
   search_stream:
+    schema_loader:
+      $ref: "*ref(definitions.schema_loader)"
```
sherifnada, line 153: fwiw I would find it more understandable if this said

```
  schema_loader:
    type: InlineSchemaLoader
```

instead of the ref

clnoll: Updated.

clnoll:
@girarda @brianjlai @sherifnada did you want to rereview this? It needs an approval to merge, since we don't want to wait on the potential forward reference change.

sherifnada:
@clnoll no review needed from me! I just happened to see that one line where I made a comment :)

girarda (approved):

PR #20422, opened 2022-12-13
[Low-Code CDK] Write the component schema and use it during manifest validation

brianjlai:
Closes  https://github.com/airbytehq/airbyte/issues/19759

## What
Together Catherine, Maxime, and I handwrote json schema that documents all the available components in the low-code language and their respective relationships. This handwritten schema will then be used as the source of truth for the Pydantic models generated, a reference document for low-code developers, and hopefully the UI.

It is also incorporated into the manifest validation flow.

## How
A majority of the work was writing the schema by hand. We went through this [checklist](https://docs.google.com/document/d/1eX6mD36zCVcS5RDGG1P9FMrHcfz_s62tg44eeu6KbCU/edit#heading=h.oxggnkz8ju0d) to avoid duplicates and ensure that we had consistent formatting and styling for the YAML.

The other aspect was incorporating this into the validation flow. We read in the new schema YAML file and perform the same validator check against the incoming schema. The one nuance here is that because we want to deprecate the factory once we have finished the refactor, the new validation flow doesn't use any part of the existing `factory.py` to modify the incoming manifest. That means that we're not doing some of the behind the scenes preprocessing in order to adhere to the component schema.

This is handled in the form of the `manifest_component_transformer.py`, which adds certain fields to components so that it can be validate against our ideal schema:

- Applies default types to components that are missing `type` and have a defined default value
- Propagates `$options` dictionary to all subcomponents

## Recommended reading order
1. `low_code_component_schema.yaml`
2. `manifest_component_transformer.py`
3. `manifest_declarative_source.py`

PR #20431, opened 2022-12-13
🪟🎉 Connector builder: Allow defining inputs

### LINKED ISSUE #20414: Add user-configurable fields
## Context

In order to allow users to make their connectors built with the Connector Builder reusable, they need to have a place to specify user input fields, which they can then reference in other inputs in the UI.

The designs for this can be found in Figma [here](https://www.figma.com/proto/F1ljvrJerLoVbNvtejBWid/LOW-CODE-CONNECTOR-BUILDER?node-id=2643%3A109882&scaling=min-zoom&page-id=2634%3A108623&starting-point-node-id=2634%3A108627&show-proto-sidebar=1).

This UI will be used to build out the `spec` block of the connector manifest, which is then consumed by the Test inputs menu to render fields for users to fill out.

## Acceptance Criteria
- User Input view select button exists in the sidebar
- User Input view has a button to add a new input, which opens a modal allowing them to configure the type and other metadata about the field, along with a preview of what the rendered input will look like
- The fields in the User Input view can be reordered to determine the order in the Test Inputs screen
- All other input fields have an embedded button which lets them select a user input field to use, adding the `{{config['<field-name>']}}` interpolated string value to the input upon selection

flash1293:
## What

Built on top of https://github.com/airbytehq/airbyte/pull/20385

Closes https://github.com/airbytehq/airbyte/issues/20414

## How

Add inputs management and write to underlying manifest:
Screenshot 2022-12-16 at 14 47 26
Screenshot 2022-12-16 at 14 47 37
Screenshot 2022-12-16 at 14 47 45

### The InputsView

Extended the BuilderView by a new view (handled like the others). The InputsView renders all the inputs using the formik context. When editing an input, a new form context is opened for the modal to do the editing in a separate state (called `InputInEditing`) to allow to cancel the action.

When mapping to the spec `Array<{ key: string, definition: JSON7Schema, required: boolean }>` is mapped to a json schema object conforming to the `spec` property of the low code manifest. This can later be done in reverse as well.

The `InputInEditing` shape holds a few more things than the raw json schema to handle some form-specific things:
* There is a type "enum" which translates into type: string and enum: < the list of possible values > in the json schema - for this reason the "type" needs to be handled outside of the actual json schema definition and mapped back once the user submits the modal form
* Additional state to show or hide the default value field

The field id (the key of the property in the final spec) is derived from the name. It also has a validation that blocks duplicated field ids. I had to touch the input styling a bit, but IMHO it's for the better in general (read-only looks like disabled now, before that it would look like a regular input)

There is a list of explicitly supported types for properties - if another one is used, the form is only partially rendered, with a warning that you need to switch to the yaml to edit this. It won't show right now but with our current plan for state management if will become important in the near future.

### The ConfigMenu

If you can edit the spec while there is a config, you can run into weird situations. Added a reset button to reset the form if that happens to a user.

### The BuilderField

To simplify a bit the inputs view code I aligned the types of a jsonschema property with the types the BuilderField component takes (this included renaming text to string). Seems like being consistent here doesn't hurt.

### The state service

Patched the connector manifest so the spec is part of it (can be removed once https://github.com/airbytehq/airbyte/pull/20422 is merged)

Also as there can't be invalid json anymore as the user can't edit the test input directly, we don't need to separate states to track the string and the json - simplified the code a bit.

### The connector form

Added support for the reset button

I noticed that when adding default values, it's weird if they are not reflected in the config menu. However, the current logic does not allow to patch existing configs with new default values. I added an option for that which becomes active if the connector form is passed an unsaved draft of a connector definition spec (like in case of the connector builder). Also, for draft definition specs it will check whether the existing config can be cast into the new schema - if not it will reset automatically.

## Recommended reading order
See the sections above

## 🚨 User Impact 🚨

User can configure inputs for their connector in the connector builder.


lmossman:
Haven't looked very closely at this PR yet, but I tried running it locally and it just showed the "Unexpected error" screen for me when I tried to open the connector builder. This is because there are a bunch of places where `values.inputs` are being accessed and chained with a function call, but for me `values.inputs` was undefined so all of those calls failed. Example screenshot:
Screen Shot 2022-12-19 at 1 56 47 PM
I'm guessing you may not have run into this because your local state contained `values.inputs` somehow during development?

flash1293:
@lmossman yeah, the state in local storage isn’t compatible with the new code anymore. For development you can simply run “localStorage.clear()” in the console and reload to get it working.

I assumed that keeping backwards compatibility is not important at this stage but I guess you are right, as this is released on OSS already we should try to not break users configuration. Your change looks like a simple way to achieve this for now.

airbyte-webapp/src/components/connectorBuilder/Builder/InputsView.tsx
```diff
@@ -0,0 +1,303 @@
...
+          />
+          <BuilderField
+            path="key"
+            type="string"
+            readOnly
+            label={formatMessage({ id: "connectorBuilder.inputModal.fieldId" })}
+            tooltip={formatMessage(
+              { id: "connectorBuilder.inputModal.fieldIdTooltip" },
+              {
+                syntaxExample: "{{my_input}}",
```
lmossman, line 212: I tried interpolating a test input (`pokemon_name`) that I added by setting the Path URL of a stream to `/pokemon/{{pokemon_name}}` , but it seems like the value I type into the Test Input menu is not being used.

Originally I thought this was because doing `{{field_id}}` was not actually the right way to interpolate it, as all of the examples in the low-code docs do `{{config['field_id']}}` to interpolate user input values. But even when I tried that, it still didn't work. (Though, this `syntaxExample` should still be updated to say something like `{{config['my_input']}}`, and we may want to file a separate issue for allowing users to just enter `{{my_input}}` instead -- thoughts?)

I just double checked and this is working on the `feature/connector-builder` branch: Open that branch, switch to the YAML view, and paste in this YAML:
```
version: "0.1.0"

streams:
  - type: DeclarativeStream
    $options:
      name: "pokemon"
      primary_key: "id"
      path: "/pokemon/{{config['pokemon_name']}}"
    primary_key: "id"
    retriever:
      requester:
        url_base: "https://pokeapi.co/api/v2"
        http_method: "GET"
      record_selector:
        extractor:
            field_pointer: []

check:
  stream_names:
    - "customers"

spec:
  documentation_url: https://docsurl.com
  connectionSpecification:
    title: Poke API Spec
    type: object
    required:
      - pokemon_name
    additionalProperties: true
    properties:
      pokemon_name:
        type: string
        description: Pokemon name
```

Then open the test inputs menu and enter `ditto` into the pokemon_name field. You can see that the URL is correctly pulling in the `ditto` value, and clicking Test properly fetches data about ditto.

But if you try the same thing on this branch, `ditto` isn't used in the URL. So this seems to be a regression specific to this PR.

airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/ConfigMenu.tsx
```diff
@@ -88,14 +84,17 @@ export const ConfigMenu: React.FC<ConfigMenuProps> = ({ className }) => {
...
                   onSubmit={async (values) => {
-                    setConfigString(JSON.stringify(values.connectionConfiguration, null, 2) ?? "");
+                    setConfigJson(values);
                     setIsOpen(false);
                   }}
                   onCancel={() => {
                     setIsOpen(false);
                   }}
+                  onReset={() => {
+                    setConfigJson({});
```
lmossman, line 96: Slightly weird UX here: If I open the config menu and all the inputs are blank, then I type a value into one of them, then click Reset, it doesn't clear the value. 
It seems like it only clears values if they had been saved to the form before, likely because values that have not yet been saved arent set on ConfigJson so setting that to an empty object doesn't affect unsaved values.

Ideally this would always clear all values in all inputs when clicked.

lmossman (changes requested):
Overall this is looking pretty great. Left some comments on issues I found when testing locally, so those should be addressed before merging

airbyte-webapp/src/components/connectorBuilder/Builder/InputsView.tsx
```diff
@@ -0,0 +1,303 @@
...
+    <BuilderConfigView heading={formatMessage({ id: "connectorBuilder.inputsTitle" })}>
+      <Text centered className={styles.inputsDescription}>
+        <FormattedMessage id="connectorBuilder.inputsDescription" />
+      </Text>
+      {inputs.value.length > 0 && (
+        <Card withPadding className={styles.inputsCard}>
+          <ol className={styles.list}>
+            {inputs.value.map((input) => (
+              <li className={styles.listItem} key={input.key}>
+                <div className={styles.itemLabel}>{input.definition.title || input.key}</div>
```
lmossman, line 108: nit: these labels currently look different from the labels in the other views, so we should make them consistent.

This should be achievable by changing this `<div>` to be a `<Label>` instead, so that the same styles are applied.

lmossman, line 108: Also, not a must have but it could be cool if we showed the type of the field in smaller grey (maybe italicized?) text next to the title

airbyte-webapp/src/locales/en.json
```diff
@@ -677,6 +678,31 @@
   "connectorBuilder.uiYamlToggle.yaml": "YAML",
   "connectorBuilder.resetAll": "Reset all",
   "connectorBuilder.emptyName": "(empty)",
+  "connectorBuilder.inputsTitle": "User inputs",
```
lmossman, line 681: To be consistent with Global Configuration:
```suggestion
  "connectorBuilder.inputsTitle": "User Inputs",
```

airbyte-webapp/src/components/connectorBuilder/Builder/InputsView.tsx
```diff
@@ -0,0 +1,303 @@
...
+          setInputInEditing(newInputInEditing());
+        }}
+        icon={<FontAwesomeIcon icon={faPlus} />}
+        iconPosition="left"
+        variant="secondary"
+      >
+        <FormattedMessage id="connectorBuilder.addInputButton" />
+      </Button>
+      {inputInEditing && (
+        <Formik
```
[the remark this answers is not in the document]
flash1293: What I did to fix these 
* Clear default on type change as you suggested. I added a `onChange` prop to the builder field to do so as an effect hook will do it in another render cycle so the form would still render in the broken state
* For the dropdown with `...` the issue was that the BuilderField would automatically select the first option if no value is set yet which is the wrong behavior IMHO (it looks like the first option is selected but this is not actually the case). So when you enable default values for an enum user input, it is not picking the first value, but instead you have to choose. The enum builder field is only used in one other place which is the http method picker and it always has a value by the default state of a stream.
* Errors about duplicate field IDs weren't shown right away because the field wasn't touched. I worked around this by auto-touching the key field and setting the title-derived key right away in the newly introduced onChange instead of waiting for a hook to trigger

airbyte-webapp/src/components/connectorBuilder/Builder/InputsView.tsx
flash1293: Yeah, I noticed this as well and fixed it in the authentication PR: https://github.com/airbytehq/airbyte/pull/20645/files#diff-05522b7f5df1df909f26423e40703073e6ab13f90f10ceeef5ca00a65561b351R88

Pulled it into this branch as well.

airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/ConfigMenu.tsx
flash1293: Fixed by also calling resetForm on the formik context to purge local formik state.

airbyte-webapp/src/components/connectorBuilder/Builder/InputsView.tsx
flash1293: done

flash1293: Let me split that out into a separate issue

airbyte-webapp/src/components/connectorBuilder/Builder/InputsView.tsx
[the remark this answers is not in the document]
lmossman: - Looks good
- That makes sense, I agree that it makes sense to make the user actually select a value if no default is set
- Nice, that seems to work well to me!

airbyte-webapp/src/components/connectorBuilder/Builder/InputsView.tsx
```diff
@@ -0,0 +1,312 @@
...
+    <BuilderConfigView heading={formatMessage({ id: "connectorBuilder.inputsTitle" })}>
+      <Text centered className={styles.inputsDescription}>
+        <FormattedMessage id="connectorBuilder.inputsDescription" />
+      </Text>
+      {inputs.value.length > 0 && (
+        <Card withPadding className={styles.inputsCard}>
+          <ol className={styles.list}>
+            {inputs.value.map((input) => (
+              <li className={styles.listItem} key={input.key}>
+                <Label className={styles.itemLabel}>{input.definition.title || input.key}</Label>
```
lmossman, line 112: One small issue with using a <Label> here that I forgot about when I suggested it: Label has a small bottom padding in its styles, which means the label is now not lined up with the edit button in the UI here:
Screen Shot 2022-12-21 at 12 04 38 PM
maybe you can override that bottom padding here or refactor those styles to remove it?

lmossman (approved):
One more small nit about the label padding, but otherwise this LGTM. Tested locally and it works great

PR #20439, opened 2022-12-13
Enable record schema validation using v1 type system; CI uses MSG to start EC2 runners

edgao:
Unreverts https://github.com/airbytehq/airbyte/pull/19625, and hardcodes `build` instead of using `${buildDir}`. Relevant changes are in the two `build.gradle` files, plus the Dockerfile.

https://github.com/airbytehq/airbyte/pull/19625 was causing the OSS release action to fail (https://github.com/airbytehq/airbyte/actions/runs/3625448595/jobs/6113524965) because the docker build couldn't find `WellKnownTypes.json`. This is working properly now.

Tested in https://github.com/airbytehq/airbyte/actions/runs/3699371623/jobs/6266680684#step:6:9612 - this action run failed because of an intentional `COPY nonexistent_file.txt`, rather than not being able to find `WellKnownTypes.json`. This action was running against https://github.com/airbytehq/airbyte/pull/20439/files/ba107138acde16cce98a659ecd70fa5ce6705c9b; the subsequent commits are just reverting the CI hacks.

---

Also switches the `start ec2 runner` action to use `MSGv0.0.2` (which supports Multiple Security Groups). This is useful for being able to SSH onto CI runner instances.


git-phu (approved):
still unclear why the original release workflow run failed, but that specific error no longer happens

gosusnp (approved):

PR #20443, opened 2022-12-13
🪟 🧹 check that stream is defined when matching

teallarson:
## What
After https://github.com/airbytehq/airbyte/pull/20387 was merged, @krishnaglick noticed that if `stream` is undefined in both the initial catalog and the current catalog, the streams could have a false match (since `undefined` === `undefined`)

This should not be a possible code path to get into, but we can clean the code up just to be safe!


krishnaglick (approved):
LGTM assuming tests green. Did not check locally.

PR #20459, opened 2022-12-13
Cleanup

benmoriceau:
## What
Remove the old glassfish server:
- Remove the proxy pass through
- fix one path
- remove not needed constructors
- remove micronaut api client
- remove old server files
- update the banner for the serve location


airbyte-server/src/main/resources/micronaut-banner.txt
```diff
@@ -6,3 +6,6 @@
 /_/  |_/_/_/  /_.___/\__, /\__/\___/
                     /____/
         : airbyte-server :
+--------------------------------------------------
+ Will soon be available at http://localhost:8000/
+--------------------------------------------------
```
colesnodgrass, line 11: Should we leave this message?

colesnodgrass (approved):

airbyte-server/src/main/resources/micronaut-banner.txt
benmoriceau: I have added it on purpose because the old server is available is not displayed anymore

gosusnp (approved):
clean up all the code

jdpgrailsdev (approved):
:shipit:

PR #20463, opened 2022-12-13
:window: :bug: [Connector Builder] correctly ignore individually touched fields in useBuilderErrors

lmossman:
## What
I found a bug in which if you touched a field in some view of the connector builder which did not have an error, and a different field had an error but was _untouched_, then the error indicator would still show for that view even though it should only be showing if there is a _touched_ field with an error in the view.

## How
This issue was caused by the useBuilderErrors object just checking if the entire view had been touched and had some errors, rather than checking individual fields in the view.

This PR fixes the issue by looking into the touched and error objects for each view and finding if there are any individual fields that are both touched _and_ have an error, which makes the error indicator correctly only show when there are touched and erroring fields in the corresponding view.

## 🚨 User Impact 🚨
Connector builder view error indicators are shown correctly.

### DIFF
--- a/airbyte-webapp/src/components/connectorBuilder/useBuilderErrors.ts
+++ b/airbyte-webapp/src/components/connectorBuilder/useBuilderErrors.ts
@@ -1,10 +1,10 @@
 import { flatten } from "flat";
-import { FormikErrors, useFormikContext } from "formik";
+import { FormikErrors, FormikTouched, useFormikContext } from "formik";
 import intersection from "lodash/intersection";
 
 import { BuilderView, useConnectorBuilderState } from "services/connectorBuilder/ConnectorBuilderStateService";
 
-import { BuilderFormValues } from "./types";
+import { BuilderFormValues, BuilderStream } from "./types";
 
 export const useBuilderErrors = () => {
   const { touched, errors, validateForm, setFieldTouched } = useFormikContext<BuilderFormValues>();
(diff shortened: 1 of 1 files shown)


airbyte-webapp/src/components/connectorBuilder/useBuilderErrors.ts
```diff
@@ -16,24 +16,46 @@ export const useBuilderErrors = () => {
...
-      ? intersection(Object.keys(errorsToCheck), Object.keys(touched))
-      : Object.keys(errorsToCheck);
+    const errorKeys = Object.keys(errorsToCheck);
 
     const invalidViews: BuilderView[] = [];
 
     if (errorKeys.includes("global")) {
-      invalidViews.push("global");
+      if (ignoreUntouched) {
+        if (errorsToCheck.global && touched.global) {
```
lmossman, line 25: Without this if statement (and the similar one for streams on line 47), this code would crash the webapp because flatten() can't handle undefineds

flash1293 (approved):
Tested, understood and LGTM.

A general note - this code starts to become pretty complex, but it's only handling a small detail in the UI which is rarely a good sign.

We don't need to do it now, but what do you think about refactoring this like this:
* Flatten the list of all errors and all touched
* Intersect them to get a list of touched errors
* Strip out prefix (global, streams.0, ...) from the keys in the list
* Dedupe
* Filter through `limitToViews` if set
* Result is the list of invalid views

I didn't build this but I think it would be a shorter implementation as it's mostly chaining standard operations.

Small performance consideration - right now all these functions are created directly in the hook which means they are re-constructed and re-run on every render. We can avoid this by memoizing (especially since the hook is used in multiple places).

Still approving as none of these are blockers.

lmossman:
Captured the above refactoring in a ticket here: https://github.com/airbytehq/airbyte/issues/20488

PR #20468, opened 2022-12-14
airbyte-metrics/server: add tags to  root span from OAuth handler

alafanechere:
## What
For better error tracking, I want to tag the root span with workspace id and definition id when OAuth handler public methods are called.

## How
* Create a `ApmTraceUtils.addTagsToRootSpan` function to add tags to the root span
* Tag the root span from OAuthHandler

### DIFF
--- a/airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/ApmTraceUtils.java
+++ b/airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/ApmTraceUtils.java
@@ -109,8 +109,10 @@ public class ApmTraceUtils {
    */
   public static void recordErrorOnRootSpan(final Throwable t) {
     final Span activeSpan = GlobalTracer.get().activeSpan();
-    activeSpan.setTag(Tags.ERROR, true);
-    activeSpan.log(Map.of(Fields.ERROR_OBJECT, t));
+    if (activeSpan != null) {
+      activeSpan.setTag(Tags.ERROR, true);
+      activeSpan.log(Map.of(Fields.ERROR_OBJECT, t));
+    }
     if (activeSpan instanceof MutableSpan) {
       final MutableSpan localRootSpan = ((MutableSpan) activeSpan).getLocalRootSpan();
       localRootSpan.setError(true);
--- a/airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/ApmTraceUtils.java
+++ b/airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/ApmTraceUtils.java
@@ -88,6 +88,21 @@ public class ApmTraceUtils {
   }
 
   /**
+   * Adds all the provided tags to the root span.
+   *
+   * @param tags A map of tags to be added to the root span.
+   */
+  public static void addTagsToRootSpan(final Map<String, Object> tags) {
+    final Span activeSpan = GlobalTracer.get().activeSpan();
+    if (activeSpan instanceof MutableSpan) {
+      final MutableSpan localRootSpan = ((MutableSpan) activeSpan).getLocalRootSpan();
+      tags.entrySet().forEach(entry -> {
+        localRootSpan.setTag(formatTag(entry.getKey(), TAG_PREFIX), entry.getValue().toString());
+      });
+    }
+  }
+
+  /**
    * Adds an exception to the root span, if an active one exists.
    *
    * @param t The {@link Throwable} to be added to the provided span.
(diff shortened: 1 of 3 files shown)


alafanechere:
I confirm that this changed propagates tags to the root span: 
Screen Shot 2022-12-14 at 10 14 38

jdpgrailsdev (approved):
:shipit:  @alafanechere Would you mind adding tests for this new method (and the one that adds an error to the root span) to the `ApmTraceUtilsTest` class?

airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/ApmTraceUtils.java
```diff
@@ -87,6 +87,21 @@ public static void addExceptionToTrace(final Span span, final Throwable t) {
...
   }
 
+  /**
+   * Adds all the provided tags to the root span.
+   *
+   * @param tags A map of tags to be added to the root span.
+   */
+  public static void addTagsToRootSpan(final Map<String, Object> tags) {
+    final Span activeSpan = GlobalTracer.get().activeSpan();
+    if (activeSpan instanceof MutableSpan) {
```
evantahler, line 97: [question] If the framework is wrapping all requests in a span, how would we _not_ have a span?  Perhaps if Data dog is disabled?

airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
```diff
@@ -70,8 +70,10 @@ public OAuthHandler(final ConfigRepository configRepository,
 
   public OAuthConsentRead getSourceOAuthConsent(final SourceOauthConsentRequest sourceOauthConsentRequest)
       throws JsonValidationException, ConfigNotFoundException, IOException {
-    ApmTraceUtils.addTagsToTrace(Map.of(WORKSPACE_ID_KEY, sourceOauthConsentRequest.getWorkspaceId(), SOURCE_DEFINITION_ID_KEY,
-        sourceOauthConsentRequest.getSourceDefinitionId()));
+    final Map<String, Object> traceTags = Map.of(WORKSPACE_ID_KEY, sourceOauthConsentRequest.getWorkspaceId(), SOURCE_DEFINITION_ID_KEY,
+        sourceOauthConsentRequest.getSourceDefinitionId());
+    ApmTraceUtils.addTagsToTrace(traceTags);
+    ApmTraceUtils.addTagsToRootSpan(traceTags);
```
evantahler, line 76: [non-blocking nit] when I've done instrumentation like this in the past, the strategy was to append as much information to the trace / root-span at the top level before getting into any individual controller.  This looks like a global rootSpanHelper which intercepts /every/ request (e.g. "request middleware") and:
* if there's a user, addd the user_id
* if there's a workspace, add the workspace_id
* if there are any LaunchDarkly feature flags, add them to the span

Then, in the controller itself, you only have to add new information (the source_id in this case).

evantahler (approved):

airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/ApmTraceUtils.java
jdpgrailsdev: @evantahler Good catch.  This method should check if `activeSpan` is not `null`.  It will be `null` when the DataDog Java agent is not present.  cc:  @alafanechere

alafanechere: Thanks for the suggestion. Isn't `if activeSpan instanceof MutableSpan` already checking a `not null`  by itself?

alafanechere: @evantahler `MutableSpan` is a datadog interface. And `Span` and  `GlobalTracer.get().activeSpan()` are from `opentelemetry`. So when datadog is disabled we might not received Span object implementing the MutableSpan interface, hence the importance of this check. Am I right @jdpgrailsdev ?

alafanechere:
> :shipit: @alafanechere Would you mind adding tests for this new method (and the one that adds an error to the root span) to the `ApmTraceUtilsTest` class?

Test added 👌

airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
alafanechere: I like this idea of helper/request middleware. This abstraction could also help us switch tracking backend (sentry/datadog) with a single interface. @evantahler could it fit a bigger project we'd have in Q1?

airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/ApmTraceUtils.java
jdpgrailsdev: @alafanechere Correct on both accounts.  The `instanceof` check does protect us from `null`, as it will resolve to `false`.  `MutableSpan` is a DataDog only class, so if a customer uses a different APM product that supports OpenTracing, that check protects us from calling methods that don't exist on the object.

PR #20469, opened 2022-12-14
source-postgres-cdc: allow LSN to be acknowledged after sync

### LINKED ISSUE #20177: Build custom Debezium Postgres Connector Jar
We merged the change in Debezium postgres connector (see issue https://github.com/airbytehq/airbyte/issues/19376) but this change would be released as part of later postgres debezium conector version. We are currently using 1.9.6 version. We could add that change to the 1.9.6 version and build a custom jar and add the custom jar in our code base to have that feature available immediately.

### LINKED ISSUE #20178: Introduce logic to flush LSN on source database based on user choice
As part of the epic https://github.com/airbytehq/airbyte/issues/19373 we need to introduce a new option in the setup form where users can select if they want the LSN to be flushed on the source postgres after data has been loaded into the destination successfully. We will have to write logic to flush the LSN and change the setup form as well

subodh1810:
Issue : https://github.com/airbytehq/airbyte/issues/20177 and https://github.com/airbytehq/airbyte/issues/20178


airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -345,19 +347,29 @@ public List<AutoCloseableIterator<AirbyteMessage>> getIncrementalIterators(final
...
           // Postgres CDC,
           // we run all the check operations and one of the check validates that the replication slot exists
           // and has only 1 entry
           getReplicationSlot(database, sourceConfig).get(0),
-          sourceConfig);
+          savedOffset);
 
       if (!savedOffsetAfterReplicationSlotLSN) {
         LOGGER.warn("Saved offset is before Replication slot's confirmed_flush_lsn, Airbyte will trigger sync from scratch");
+      } else if (PostgresUtils.shouldFlushAfterSync(sourceConfig)) {
```
akashkulk, line 367: if the user is in flush after sync mode, can the saved offset ever be after the replication slot LSN?

airbyte-integrations/connectors/source-postgres/src/main/resources/spec.json
```diff
@@ -270,6 +271,14 @@
                 "order": 5,
                 "min": 120,
                 "max": 1200
+              },
+              "lsn_commit_behaviour": {
```
akashkulk, line 275: Could you post a quick screenshot of how this looks like?

airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/CdcPostgresSourceTest.java
```diff
@@ -438,4 +449,102 @@ public SourceRecord sourceRecord() {
...
+
+    final AutoCloseableIterator<AirbyteMessage> firstBatchIterator = getSource()
+        .read(config, CONFIGURED_CATALOG, null);
+    final List<AirbyteMessage> dataFromFirstBatch = AutoCloseableIterators
+        .toListAndClose(firstBatchIterator);
+    final List<AirbyteStateMessage> stateAfterFirstBatch = extractStateMessages(dataFromFirstBatch);
+
+    final Long replicationSlotAfterFirstSync = PgLsn.fromPgString(
+        source.getReplicationSlot(defaultJdbcDatabase, config).get(0).get("confirmed_flush_lsn").asText()).asLong();
+    assertEquals(0, replicationSlotAfterFirstSync.compareTo(replicationSlotAtTheBeginning));
```
akashkulk, line 478: nit : can you just to 
assertTrue(replicationSlotAfterFirstSync == replicationSlotAtTheBeginning) 

just find it easier to read. Same with assertions below.

airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/CdcPostgresSourceTest.java
```diff
@@ -438,4 +449,102 @@ public SourceRecord sourceRecord() {
     assertNull(pctp.getHeartbeatPosition(null));
   }
 
+  @Test
+  protected void syncShouldIncrementLSN() throws Exception {
```
akashkulk, line 453: nit : Can you add comments on the different scenarios being tested?
e.g. 

1. Assert LSNs don't change before/after sync in ack after sync mode. 
2. Assert LSNs change after a subsequent with new state
...etc

airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/CdcPostgresSourceTest.java
edgao: drive-by comment - `assertEquals(replicationSlotAtTheBeginning, replicationSlotAfterFirstSync)` ? (so if the assert fails, junit will also print the expected+actual values for easier debugging)

airbyte-integrations/connectors/source-postgres/src/main/resources/spec.json
subodh1810: Screenshot 2023-01-03 at 1 11 47 PM

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
subodh1810: In case another process consumes the WAL records and acks the replication slot then yes its possible

akashkulk (approved):

rodireich (approved):

PR #20474, opened 2022-12-14
🪟 🎨 Refining BulkEditPanel component

### LINKED ISSUE #20372: New Bulk Edit Panel changes
List of items related to the bulk edit panel. ([Figma](https://www.figma.com/file/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?node-id=2732%3A60474&t=fMqpOXfEj3pHRloD-1))

Sync switch needs to be updated to match the design:
image

Should have a border and whit text color for source-defined cursors or primary keys:
image

Should say "not available" when the cursor or primary key is not available for the mode:
image

Other: 
* When Selecting sync modes that don't need the cursor field or the primary key, it should clear
* When canceling, it should reset all the fields to empty instead of saving the previous ones
* Add unit tests to the component to ensure the functionality works as expected

### LINKED ISSUE #20731: Update source defined cursor and primary key to match selected color
Figma:
https://www.figma.com/file/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?node-id=972%3A36208&t=gUgiRjka8M752UGy-1

Add color to the source-defined primary key or cursor that is not editable. These match the text colors of the pill button text:

image

These colors appear when the stream row is:

* Standard
* Selected
* Adding
* Removing

YatsukBogdan1:
## What
Closes #20372 
Closes #20731 

## How
- Updates rendering logic, for the cases while Primary Key or Cursor is not available or defined by source
- We also had discussion with @edmundito that `Switch` styling will be done in separate PR

## Screenshots
Screenshot 2022-12-14 at 16 24 30
Screenshot 2022-12-14 at 16 24 37

Screenshot 2022-12-22 at 15 18 27
Screenshot 2022-12-22 at 15 18 31
Screenshot 2022-12-22 at 15 18 48


tealjulia:
> the cases while Primary Key or Cursor is not available

Would that be in cases where all of the selected streams do not have at least one column with the same name?  Are there any other cases?

krishnaglick:
Failing snapshot tests :)

airbyte-webapp/src/components/connection/CatalogTree/next/StreamPathSelect.tsx
```diff
@@ -37,19 +35,18 @@ type PathPopoutProps = StreamPathSelectBaseProps & (StreamPathSelectMultiProps |
...
-      return (
-        <Text className={styles.text}>
-          <Tooltip placement="bottom-start" control={text}>
-            {text}
-          </Tooltip>
-        </Text>
-      );
+      node = props.isMulti ? props.path.map(pathDisplayName).join(", ") : pathDisplayName(props.path);
+    } else {
+      node = <FormattedMessage id="connection.catalogTree.sourceDefined" />;
```
krishnaglick, line 42: Nit: This could be in a `useMemo`

krishnaglick (commented):
Should this PR have more tests added?

YatsukBogdan1:
> > the cases while Primary Key or Cursor is not available
> 
> Would that be in cases where all of the selected streams do not have at least one column with the same name? Are there any other cases?
@tealjulia for now we decided to put only design updates in this PR, and keep any logic updates in separate one

> Should this PR have more tests added?
@krishnaglick as we discussed on the stand up yesterday, the component itself is dumb, and just renders what is passed from `useBulkEditService` hook, so the component itself is tested, and lets keep any tests related to the logic, into another PR

edmundito (approved):

edmundito (changes requested):
I re-tested again and did notice a discrepancy in the table itself. It looks like the non-selectable values have a background, but that only applies to the bulk edit

Here you can see that id is not selectable but it has the pill background around it
Screen Shot 2022-12-21 at 16 23 58

But this is really similar to what's needed in #20731

YatsukBogdan1:
@edmundito I updated table row cursor and primary key as you suggested

Screenshot 2022-12-22 at 15 18 27
Screenshot 2022-12-22 at 15 18 31
Screenshot 2022-12-22 at 15 18 48

YatsukBogdan1:
Added a few more updates:
- removed animation for `<BulkEditPanel />`
- added `<StreamPathSelect />` new property `pillIfChangeUnavailable` for cases when we need or not to render pill under unavailable to change path text
image

edmundito:
@YatsukBogdan1 Just checked the latest changes and I noticed that what I see doesn't match the screenshot int he previous comment.
Screen Shot 2023-01-03 at 14 39 05

YatsukBogdan1:
updated PR with next changes:
- disables pointer events for `PillButton` component while `disabled: true`

airbyte-webapp/src/components/connection/CatalogTree/next/StreamPathSelect.tsx
```diff
@@ -19,6 +18,10 @@ interface StreamPathSelectBaseProps {
   placeholder?: React.ReactNode;
   variant?: PillButtonVariant;
   disabled?: boolean;
+  // This property is used for cases when the path is defined by source, therefore
+  // in some cases we need this path to render with pill background (BulkEditPanel) and
+  // in some cases it should be only text (StreamsTable)
+  pillIfChangeUnavailable?: boolean;
```
edmundito, line 24: Maybe `withSourceDefinedPill` is more clear?

airbyte-webapp/src/components/ui/PillSelect/PillSelect.tsx
```diff
@@ -34,7 +36,7 @@ export const PillSelect: React.FC<PillSelectProps> = ({ className, ...props }) =
                 active={isOpen}
                 className={className}
               >
-                {label}
+                {disabled ? <FormattedMessage id="connectionForm.bulkEdit.pillButtonLabel.notAvailable" /> : label}
```
edmundito, line 39: This formatted message is part of the connection form, but this component is part of the UI, so it crosses different domains. We're not sure if the default behavior for disabled is always to say "not available"

edmundito (changes requested):
Tested locally and the functionality looks good but had some comments about the implmentation.

airbyte-webapp/src/components/ui/PillSelect/PillSelect.tsx
YatsukBogdan1: @edmundito I added `renderDisabledState: () => React.ReactNode` property for `PillSelect`, so now, when `disabled: true`, the content can be controlled from different application parts. please let me know if this looks good to you

edmundito (commented):
LGTM. Tested locally.

edmundito (approved):

PR #20481, opened 2022-12-14
prevent the cursor or primary key fields from being deselected

mfsiega-airbyte:
## What
Ensure that the cursor and primary key fields can't be deselected.

We might want to allow this in some cases, but for simplicity we disallow it entirely per the PRD.

## How
If but `cursor`/`primary key` and `selectedFields` are provided, check that the key field is included in the list of selected fields.

### DIFF
--- a/airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
+++ b/airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
```diff
@@ -66,7 +67,18 @@ private static io.airbyte.protocol.models.AirbyteStream toProtocol(final Airbyte
         }
       }
       // Only include the selected fields.
-      List<String> selectedFieldNames = config.getSelectedFields().stream().map((field) -> field.getFieldPath().get(0)).collect(Collectors.toList());
+      final Set<String> selectedFieldNames =
+          config.getSelectedFields().stream().map((field) -> field.getFieldPath().get(0)).collect(Collectors.toSet());
+      // TODO(mfsiega-airbyte): we only check the top level of the cursor/primary key fields because we
+      // don't support filtering nested fields yet.
+      if (!selectedFieldNames.contains(config.getCursorField().get(0))) {
+        throw new JsonValidationException("Cursor field cannot be de-selected");
```
pmossman, line 75: I'm trying to think if there could be a case where a non-cursor field shares a top-level key with a cursor field, and therefore can't be de-selected. Even if that is the case, probably better to merge this as-is for now because we're planning on adding support for nested fields relatively soon anyways

pmossman (approved):
Looks reasonable to me!

airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
mfsiega-airbyte: Yeah as it is today we only allow selecting top-level fields (which implies all of the children) - so the only way _to_ deselect that non-cursor-field-which-shares-a-top-level-key-with-a-cursor is to deselect the top-level field, which in turn deselects the cursor...

But indeed we should support nested fields soon, at least on the backend.

pmossman: Cool, makes sense to me, thanks for confirming!

PR #20491, opened 2022-12-14
add ReplicationWorkerFactory to dry creation logic

cgardens:
## What
* We duplicate most of the logic for creating a DefaultReplicationWorker in the orchestrator and not orchestrator versions of the platform. This PR creates a factory to de-duplicate that logic.
* Is it helpful? We can close it if it doesn't feel like an improvement.


jdpgrailsdev (commented):
@cgardens This seems fine to me.  Might make sense to add a test for the new factory class or is this code already covered by tests for the the ReplicationJob/ReplicationActivity classes?

cgardens:
agreed. a test is in order.

colesnodgrass:
Closing as this PR as it's old and most of this code now lives in https://github.com/airbytehq/airbyte-platform/.

PR #20494, opened 2022-12-14
Fix incorrect naming in application.yml

xiaohansong:
Incorrect naming in application.yml so the variable was not in effect.


pmossman (approved):

xiaohansong:
Updated another misused name in the configuration as well. We should have a check to make sure cloud env config and configs in oss are aligned!

PR #20464, opened 2022-12-14
[Connector Builder] Update view styles and add stream delete button

lmossman:
## What
This PR updates the styles of the view titles/names in the Connector Builder UI (e.g. Global Configuration, Stream) to match the [current design](https://www.figma.com/proto/F1ljvrJerLoVbNvtejBWid/LOW-CODE-CONNECTOR-BUILDER?node-id=2643%3A110377&scaling=min-zoom&page-id=2634%3A108623&starting-point-node-id=2634%3A108627&show-proto-sidebar=1), as well as adds a delete button to the stream views.

Screenshots:
Screen Shot 2022-12-14 at 2 30 48 PM
Screen Shot 2022-12-14 at 2 30 58 PM

## How
Add a new BuilderTitle component, which is basically just a re-styled Input, and use that for the connector name and stream name inputs.

Add a button to the stream view which removes the current stream from the formik values and sets the view to the next one that makes sense.

## 🚨 User Impact 🚨
Titles in the connector builder are more prominent, and streams can now be deleted

PR #20497, opened 2022-12-14
:window: :tada: [Connector Builder] Add request options component

### LINKED ISSUE #19104: [Connector Builder UI] Configurable Request Options
Add UI field to the global configuration view to allow configuration of authentication.

Acceptance Criteria:
- An input exists in the global view allowing users to configure the authentication method
- Each input field exists as described in the manifest json schema
- The values entered into this element is propagated to the underlying JSON connector manifest source of truth

lmossman:
## What
Resolves https://github.com/airbytehq/airbyte/issues/19104

Add a request options component to the stream view of the connector builder.

This PR is stacked on top of https://github.com/airbytehq/airbyte/pull/20464, so should only be merged after that is merged

## How
This required adding a new component type (KeyValueListField), which allows users to add or remove key-value entries to the grouped component in the UI. 

Here is a screenshot of how this looks:
Screen Shot 2022-12-14 at 2 52 51 PM
As you can see, the header and parameter key/values were added to the request as shown in the testing panel on the right.

## Recommended reading order
1. KeyValueListField.tsx
2. types.ts
3. StreamConfigView.tsx

## 🚨 User Impact 🚨
Users can now specify request options for streams in the connector builder


airbyte-webapp/src/components/connectorBuilder/types.ts
```diff
@@ -42,6 +52,11 @@ export const convertToManifest = (values: BuilderFormValues): ConnectorManifest
           name: stream.name,
           url_base: values.global?.urlBase,
           path: stream.urlPath,
+          request_options_provider: {
+            request_parameters: Object.fromEntries(stream.requestOptions.requestParameters),
+            request_headers: Object.fromEntries(stream.requestOptions.requestHeaders),
+            request_body_data: Object.fromEntries(stream.requestOptions.requestBody),
```
flash1293, line 58: I liked the idea with the toggle between json and url encoded, are you planning to do this in a separate PR?

airbyte-webapp/src/components/connectorBuilder/Builder/KeyValueListField.module.scss
```diff
@@ -0,0 +1,39 @@
...
+.labeledInput {
+  flex: 1;
+  max-width: 400px;
+  display: flex;
+  gap: variables.$spacing-sm;
+  align-items: center;
+}
+
+.kvLabel {
+  color: colors.$grey;
```
flash1293, line 22: Could we use a darker grey here? This will be hard to read for the vision-impaired and bad-screen-havers:
Screenshot 2022-12-15 at 15 40 18

flash1293 (approved):
LGTM as well except for a nit

airbyte-webapp/src/components/connectorBuilder/types.ts
lmossman: Yeah I am planning to do that separately. I want to check with the rest of the team on that though, and also ask them about if we want to have a generic JSON blob input for the request body since it seems like that is supported by the framework as well. So I will create a ticket and bring it to backlog grooming

lmossman: Created an issue here: https://github.com/airbytehq/airbyte/issues/20548

PR #20519, opened 2022-12-15
Ignore sigterm when closing source/destination

jdpgrailsdev:
## What
* Reduce error noise when closing sources/destinations

## How
* Ignore the SIGTERM exit code (143) in addition to exit code 0 when closing source/destination processes.

**N.B.:** I decided to put the set of ignored exit codes in both classes, as over time we may end up with a different set that we want to ignore for one but not both.  I don't hold this belief strongly, so if we'd rather extract it to a common utility to avoid duplication, let me know.

## Recommended reading order
1. `DefaultAirbyteSource.java`
2. `DefaultAirbyteDestination.java`

## Tests

* All tests pass
* Kube acceptance tests pass locally
* Project builds successfully locally


davinchia (approved):
Great idea!

Can we pull the exit codes out into a separate class so they don't diverge?

Feel free to merge after.

Note: Worth throwing up a quick test to make sure we don't remove this functionality?

jdpgrailsdev:
> 

@davinchia I debated pulling the codes into a common place.  On one hand, reducing repeated code is always a good thing.  On the other hand, we may end up with different codes that we allow for a source vs destination.  If we don't think that is a concern, then I am happy to move it to shared place.

I will see about adding a test to cover the additional exit code.

PR #20520, opened 2022-12-15
🎉 New Source Braze: new source implemented in low-code

roman-yermilov-gl:
## What
*New Braze connector*

## How
*Low-code connector*

## Recommended reading order
1. `braze.yaml`
2. `y.python`


lazebnyi (approved):

airbyte-integrations/connectors/source-braze/source_braze/datetime_stream_slicer.py
```diff
@@ -0,0 +1,49 @@
...
+
+from airbyte_cdk.sources.declarative.interpolation import InterpolatedString
+from airbyte_cdk.sources.declarative.requesters import RequestOption
+from airbyte_cdk.sources.declarative.requesters.request_option import RequestOptionType
+from airbyte_cdk.sources.declarative.stream_slicers import DatetimeStreamSlicer
+from airbyte_cdk.sources.declarative.types import StreamSlice
+
+
+@dataclass
+class DatetimeStreamSlicerComponent(DatetimeStreamSlicer):
```
sherifnada, line 14: can you create an issue requesting that this is added to the datetimestreamslicer [like described here](https://airbytehq-team.slack.com/archives/C02U9R3AF37/p1670428588243999) and clarifying what the new component should do, and add the issue to the docstring?

sherifnada: I don't understand why this is different than the datetimestreamslicer. Can you clarify in the docstring? doesn't datetime stream slicer already have step size?

sherifnada (changes requested):

airbyte-integrations/connectors/source-braze/source_braze/datetime_stream_slicer.py
roman-yermilov-gl: Step exists in basic `DatetimeStreamSlicer` but I need it to be an option which will be passed as a request parameter and to be calculated based on difference between start/end dates. I need the step amount to be more than actual size of the difference by 1, but not for fist slice. I also do not need last slice if start/end dates a equal. And it is important thing that step amount is dynamic and because last slice must not be equal to the step amount if the difference lesser than in step. All of this is not a standard thing and I am not sure we need it in `DatetimeStreamSlicer`. 
Why I need it:
Braze API provides us with `ending_at` and `length` query params, so we can to specify latest point and then move back by `length` until reach `start_date`(initially specified by user). Date specified in `ending_at` is an exclusive property which means we are getting `length` records strictly less than `ending_at`, but `DatetimeStreamSlicer` adds 1 day to previous end date to calculate next start date for next slice, that is why we loose 1 record between each slice. Having `step` query parameter amount one higher than actual difference lets us avoid the problem with gaps in slice windows

airbyte-integrations/connectors/source-braze/source_braze/transformations.py
```diff
@@ -0,0 +1,26 @@
+from dataclasses import dataclass
+from typing import Optional
+
+import dpath
+from airbyte_cdk.sources.declarative.transformations import AddFields
+from airbyte_cdk.sources.declarative.types import Record, Config, StreamState, StreamSlice
+
+
+@dataclass
+class TransformToRecordComponent(AddFields):
```
sherifnada, line 10: please create a Github issue for this custom component as well so we can keep track of which components we need to add to the CDK

sherifnada (approved):
LGTM for this PR but please create two github issues to track the custom components in this PR

bazarnov (approved):

ni-aschuler:
Thank you for this connector, very much appreciated! 
Any plans on supporting additional "details" endpoints? especially:
https://www.braze.com/docs/api/endpoints/export/canvas/get_canvas_details/
https://www.braze.com/docs/api/endpoints/export/campaigns/get_campaign_details/
https://www.braze.com/docs/api/endpoints/export/segments/get_segment_details/
Thank you, Cheers

PR #20523, opened 2022-12-15
CDK 0.15.0 and source-github 0.3.10 -- revert AvailabillityStrategy changes

erohmensing:
## What
* New 0.15.0 release of the CDK that reverts changes added in 0.13.0 and 0.13.3
* New version of github 0.3.10 that reverts changes from 0.3.9 (should functionally be the same as 0.3.8)

## Why
We ran into issues especially with substreams interacting with these changes, maybe also API errors that are not strictly HTTPErrors. (Not quite sure, will update this when we've tracked down the issue). Connectors that have CDK versions 0.13.x and 0.14.x installed might run into these issues. We publish a new CDK version so we can build those connectors with 0.15.0. 

We publish a new github version with reverted changes so that we can unpin 0.3.8 in the source definition mask on cloud.


airbyte-integrations/connectors/source-github/Dockerfile
```diff
@@ -12,5 +12,5 @@ RUN pip install .
 ENV AIRBYTE_ENTRYPOINT "python /airbyte/integration_code/main.py"
 ENTRYPOINT ["python", "/airbyte/integration_code/main.py"]
 
-LABEL io.airbyte.version=0.3.9
```
erohmensing, line 15: @girarda what should be here? I guess maybe 0.3.10 and we release a new version of this as well?

girarda: yes, let's bump to 0.3.10

erohmensing: Done, putting a reminder here that we can't publish til we confirm cdk 0.15 is available on pypi

girarda (approved):

airbyte-cdk/python/setup.py
```diff
@@ -58,12 +58,10 @@
         "requests_cache",
         "Deprecated~=1.2",
         "Jinja2~=3.1.2",
-        "cachetools",
     ],
     python_requires=">=3.9",
     extras_require={
         "dev": [
-            "freezegun",
```
erohmensing, line 66: These were removed accidentally - they were part of the commits that shouldn't have been reverted. Need to be added back in like the changelog. This was a merge conflict resolution issue. Adding them back in!

PR #20527, opened 2022-12-15
🪟 🐛 Fix JobService for jobs without attempts

timroes:
## What

There are sometimes jobs that don't have any attempts (yet). When we load them this code broke, since we tried to access `attempt.endedAt` on `undefined` now. Fixed this to make sure we have any attempts actually before trying to get the last one.

Since we can't fetch any logs in this case, we're actually just showing now a simple error message when the job is expanded:

screenshot-20221215-202147

The code below that line already handled the case when `lastAttemptEndTimestamp` is undefined.

### DIFF
--- a/airbyte-webapp/src/components/JobItem/components/JobLogs.module.scss
+++ b/airbyte-webapp/src/components/JobItem/components/JobLogs.module.scss
@@ -1,0 +1,5 @@
+@use "scss/variables";
+
+.jobStartFailure {
+  padding: variables.$spacing-lg;
+}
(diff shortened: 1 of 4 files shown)


joeykmh (approved):

krishnaglick (approved):
Is the log just empty in this scenario?

timroes:
We can't even load the log in this case, thus I think this implementation won't solve it alone. I'm waiting for feedback from platform workflow to understand how we can end up in scenarios without attempts at all.

timroes:
This is what you'd be seeing atm:

screenshot-20221215-195544

lmossman (approved):
LGTM, did not test locally

krishnaglick (approved):
👍

PR #20547, opened 2022-12-15
Adjust connector icons

timroes:
## What

Adjust all our connector icons with ones Nico adjusted (to square them and add correct padding). Also I minimized all SVGs (using svgo) of connector icons to make sure we're not transferring more bytes than needed for those icons.

This also adds a new gradle task to `airbyte-config` that will validate, that the icons that are used inside the seed files actually exist as well as making sure that no orphaned icons are laying around in the icons folder that aren't used anywhere.


lmossman (commented):
Two comments:
- It looks like some of these icons which are not square-shaped have been made a bit larger vertically or horizontally than the square icons (maybe to make up for the fact that they are skinnier in one of those directions?), but this makes them look noticeably larger than the squarish icons. For example, TikTok and Zendesk Chat in this screenshot, compared to Stripe and Twilio. Do we want to try to make all of these look more similar in size, or is this acceptable?
  Screen Shot 2022-12-15 at 2 47 21 PM
- The placeholder "wrench/screwdriver" icon looks off-center when compared with the other icons. It seems to be positioned more to the left than everything else:
  Screen Shot 2022-12-15 at 2 51 52 PM

timroes:
For the first point, @Upmitt can you have a look at that.

For the second: yes that problem though is unrelated to this change and already existing on master. Did not to plan changing this with this PR.

timroes:
Failure to validate that the new gradle check is working: https://github.com/airbytehq/airbyte/actions/runs/3743133413/jobs/6354963833#step:18:8067

lmossman (approved):
The S3 icon was the only one that surprised me when I was testing locally. I haven't seen that icon used for S3 before -- the old S3 icon that we currently have is the one that I recognize as S3. But maybe I'm just not aware of an icon switch that amazon has done? 

@Upmitt might be worth double checking that the s3 icon is correct.

Otherwise these icons look good to me. Tested locally by rebuilding the airbyte-server on this branch and it looks good

PR #20549, opened 2022-12-15
source-postgres: enable SAT high test strictness level

### LINKED ISSUE #19899: Enable SAT for local Java sources - Postgres
Enable the SAT for the Postgres source. See epic description for more context https://github.com/airbytehq/airbyte/issues/19258.

Because this is a local source, this story includes:
- Setting up a hosted instance of postgres to use in tests
- Enabling SAT to run for the postgres source

rodireich:
## What
Closes #19899
As `source-postgres` is a GA connector we want it to pass Source Acceptance Tests (SAT) in `high` strictness level mode.


alafanechere:
@rodireich : 
Here's what I've done to make SAT pass:
* I’ve set up all tests in `acceptance-test-config.yml` with the minimum requirements to pass `high` test strictness level.
* SAT connects to a remote PG instance I’ve set up on CloudSQL. The configuration is stored on GSM in the following secret `SECRET_SOURCE-POSTGRES__CREDS`
* I manually seeded the Postgres database with an `id_and_name` table.


**Here are my remaining questions for you**:
* Can we leverage the local test database that `setupEnvironment` is spinning? We’d need it to make it write a valid config file that we could use instead of the GSM secret. (We’d need to change the `config_path` field in `acceptance-test-config.yml` if you can’t write it to `secrets/config.json`) ?
* I bypassed the `future_state` [test](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_incremental.py#L250) because it’s not passing: when we run a read with a state with a nonexisting cursor value, an NPE is thrown. This test expects the read to:
  * succeed
  * output 0 records, 
  * output a STATE message. 
  I suggest we log an issue to ensure the PG connector adopts this behavior.   
* It’d be great to increase the coverage of this test by creating CDC configurations and seeding the test db accordingly. Could you share a CDC configuration and approach to seed the test DB with an interesting table for CDC?

evantahler (approved):
Don't forget to publish a new CDK version out of this PR

rodireich:
Thanks for the great work Augustine.
I'm OK to merge.

Whether or not this is merged, What I would like to first understand and fix is:
Why we see an invalid cursor error. The `ID` column cursor you point to is an integer which is a valid type.

evantahler:
Is this ready to merge?

alafanechere:

alafanechere:
> Why we see an invalid cursor error. The ID column cursor you point to is an integer which is a valid type.

The cursor with value `4` I've set in the `abnormal_state.json` is indeed a valid type but this value does not exist in the table I've seeded. When a state is passed to a `READ` command with a cursor value greater than what exists on the source we expect the source connector to:
- not output any record
- emit a state message 

This is what the `future_state` test is testing. But the postgres source is failing when we pass this cursor value with an NPE.

alafanechere:

PR #20550, opened 2022-12-15
Platform uses protocol v1 objects

gosusnp:
## What

Platform is now using the version 1 of the protocol objects.
https://github.com/airbytehq/airbyte/pull/20036 should have been merged beforehand and added the support of protocol v1 in the platform. This change is about using the correct version of the objects and prepare the removal of the un-versioned protocol objects.

## How

* Add protocol v1 specific helper classes
* Switch the platform code to use v1 objects by default
* Fix some java connectors test code that rely on the platform code to convert the objects as expected

## Recommended reading order
1. `airbyte-protocol/protocol-models/src/main/java/io/airbyte/protocol/models/v1` contains the v1 specific helpers
2. Rest of the code should be mostly namespace swaps

## 🚨 User Impact 🚨

No visible impact, https://github.com/airbytehq/airbyte/pull/20036 should already have enabled v1 of the protocol.


gosusnp:
No longer relevant.

PR #3663, opened 2021-05-27
Issue 3662: HTTP request source support 'csv type' response

dmateusp:
## What
Resolves #3662

## How
It implements the second solution described in the issue, if the response received by the HTTP Request Source is a list of lists, then it wraps the lists into `{"data": <list>}` records

## Pre-merge Checklist
- [ ] *Run integration tests*
- [ ] *Publish Docker images*

## Recommended reading order
1. `test.java`
1. `component.ts`
1. the rest

PR #20563, opened 2022-12-16
Fix OSS integration tests for docs (#3663)

tkorenko:
## What
+ Changes `deploy-docs-site` workflow and `deploy_docusaurus` script so these two things are used for doc PR verification as well.

### DIFF
--- a/.github/workflows/deploy-docs-site.yml
+++ b/.github/workflows/deploy-docs-site.yml
@@ -41,9 +41,9 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |-
-          export SKIP_DEPLOY="no"
-          if [ "${{github.event_name}}" = "pull_request" ]; then
-              export SKIP_DEPLOY="yes"
+          export SKIP_DEPLOY="yes"
+          if [ "${{github.event_name}}" = 'push' -o "${{github.event_name}}" = 'workflow_dispatch' ]; then
+              export SKIP_DEPLOY="no"
           fi
 
           ./tools/bin/deploy_docusaurus
@@ -55,8 +55,8 @@ jobs:
         env:
           SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN_AIRBYTE_TEAM }}
         with:
-          # 'C02TYDSUM8F' => "#dev-deploys'
-          # 'C03BEADRPNY' => presumably '#docs'
+          # 'C02TYDSUM8F' => '#dev-deploys'
+          # 'C03BEADRPNY' => '#docs'
           args: >-
             {\"channel\":\"C02TYDSUM8F\",\"attachments\":[
             {\"color\":\"#ff0000\",\"blocks\":[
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-docs-site.yml
```diff
@@ -7,13 +7,20 @@ on:
     paths:
       - "docs/**"
 
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+    paths:
+      - "docs/**"
```
supertopher, line 16: @sherifnada can you give us a path for the connector docs that we import into docs so we can add it here and on line no. 8?

.github/workflows/deploy-docs-site.yml
```diff
@@ -29,22 +36,47 @@ jobs:
...
+
       - name: Run Docusaurus
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: ./tools/bin/deploy_docusaurus
+        run: |-
+          export SKIP_DEPLOY="no"
+          if [ "${{github.event_name}}" = "pull_request" ]; then
+              export SKIP_DEPLOY="yes"
+          fi
```
supertopher, line 47: can we invert this logic?  I don't expect this to fail, but I think it would be safer to have `SKIP_DEPLOY=yes` as our default.  
The way this is currently written if we were to find an unintended input for `${{github.event_name}}` we would deploy and that feels more risky

.github/workflows/deploy-docs-site.yml
```diff
@@ -29,22 +36,47 @@ jobs:
...
+
+
+      - name: Notify Slack -- deploy failed
+        if: always() && ( github.event_name == 'push' || github.event_name == 'workflow_dispatch' ) && job.status != 'success'
+        uses: abinoda/slack-action@master
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN_AIRBYTE_TEAM }}
+        with:
+          # 'C02TYDSUM8F' => "#dev-deploys'
+          # 'C03BEADRPNY' => presumably '#docs'
```
supertopher, line 59: `C032Y32T065` is #docs

Screenshot 2022-12-16 at 7 40 39 AM

supertopher (commented):
Looks great, I have a few minor nits.  Please ping me on slack if you want me to take another look.  I get low signal on second requests.

.github/workflows/deploy-docs-site.yml
supertopher: Taras for context, I believe we import connector docs programmatically, but they live outside of the docs folder

.github/workflows/deploy-docs-site.yml
tkorenko: _i cannot check this because this channel is absent in GL slack workspace_

tkorenko: _corrected in_ [a4300bb](https://github.com/airbytehq/airbyte/pull/20563/commits/a4300bbbe0af81e3d12f91039c49ace3a2d33bb7)

.github/workflows/deploy-docs-site.yml
tkorenko: ok, _corrected in_ [a4300bb](https://github.com/airbytehq/airbyte/pull/20563/commits/a4300bbbe0af81e3d12f91039c49ace3a2d33bb7)

supertopher (approved):
thanks taras.  I think this looks great

Issue #20559, opened 2022-12-16
`Removal of NormalizationRunnerFactory` introduced a bug - "this.normalizationIntegrationType" is null


evantahler:
Releasing https://github.com/airbytehq/airbyte/pull/20245 to Airbyte Cloud caused a number of syncs to fail. The relevant stack trace is:

```
2022-12-16 00:42:30 normalization-orchestrator > Normalization failed for job 967706.
java.lang.NullPointerException: Cannot invoke "String.toLowerCase()" because "this.normalizationIntegrationType" is null
	at io.airbyte.workers.normalization.DefaultNormalizationRunner.normalize(DefaultNormalizationRunner.java:107) ~[io.airbyte-airbyte-commons-worker-dev-6b86f5024c.jar:?]
	at io.airbyte.workers.general.DefaultNormalizationWorker.run(DefaultNormalizationWorker.java:77) ~[io.airbyte-airbyte-commons-worker-dev-6b86f5024c.jar:?]
	at io.airbyte.workers.general.DefaultNormalizationWorker.run(DefaultNormalizationWorker.java:34) ~[io.airbyte-airbyte-commons-worker-dev-6b86f5024c.jar:?]
	at io.airbyte.container_orchestrator.orchestrator.NormalizationJobOrchestrator.runJob(NormalizationJobOrchestrator.java:82) ~[io.airbyte-airbyte-container-orchestrator-dev-6b86f5024c.jar:?]
	at io.airbyte.container_orchestrator.Application.run(Application.java:79) ~[io.airbyte-airbyte-container-orchestrator-dev-6b86f5024c.jar:?]
	at io.airbyte.container_orchestrator.Application.main(Application.java:38) ~[io.airbyte-airbyte-container-orchestrator-dev-6b86f5024c.jar:?]
Log4j2Appender says: Normalization executed in 0 seconds for job 967706.
2022-12-16 00:42:30 normalization-orchestrator > Normalization executed in 0 seconds for job 967706.
Log4j2Appender says: Killing orchestrator because of an Exception
2022-12-16 00:42:30 ERROR i.a.c.Application(run):81 - Killing orchestrator because of an Exception
io.airbyte.workers.exception.WorkerException: Normalization Failed.
	at io.airbyte.workers.general.DefaultNormalizationWorker.run(DefaultNormalizationWorker.java:103) ~[io.airbyte-airbyte-commons-worker-dev-6b86f5024c.jar:?]
[...]

PR #20566, opened 2022-12-16
Fix snowflake destination spec

flash1293:
## What
According to https://docs.airbyte.com/connector-development/connector-specification-reference/#using-oneof there needs to be a shared `const` property if `oneOf` is used in the spec. The snowflake destination does not follow this rule - the simple password case is not featuring the `auth_type` prop.

This PR fixes the spec.


flash1293:
Currently failing due to https://github.com/airbytehq/airbyte/issues/20559

lmossman (approved):
LGTM

edgao:
(also submitted https://github.com/airbytehq/airbyte/issues/20660 for us to consume this new property)

Issue #19500, opened 2022-11-16
support limiting number of pages / records retrieved from connector builder server


sherifnada:
## Tell us about the problem you're trying to solve
As a connector builder UI developer, if my stream contains a large number of records and I press the test read button, I might have to wait a very long time for all the records to be retrieved. 

## Describe the solution you’d like
I want a way to limit the number of records retrieved from the connector builder. 

## Acceptance Criteria
- [ ] #19937
- [ ] The API read endpoints exposes a `record_limit:` parameter which limits the number of records returned
- [ ] Wire up the two points above

lmossman:
@sherifnada I think the second acceptance criteria may have not been fully written

I think my main open question here is: should this allow users to limit the number of pages or number of records, or both? Some thoughts here:
- Limiting the number of pages only makes sense if pagination is configured, so if it is not configured we would either need to only allow users to limit records, or not let them limit anything
- My guess is that most of the turnaround time for a `read` operation is taken up by submitting an HTTP request to the source and waiting for the result (but correct me if I'm wrong here). Therefore, limiting pages may make more sense than limiting records, because one HTTP request is submitted per page

sherifnada:
@lmossman woops, brain timed out while writing the issue. Your reasoning makes a lot of sense. Changed to talk about pages.

girarda:
grooming notes:
* we only want to limit the number of records so we don't need to modify the CDK

PR #20575, opened 2022-12-16
[Connector-builder server] Allow client to specify record limit and enforce max of 1000

clnoll:
## What
In order to improve performance of test read requests in the connector builder UI, we want to avoid fetching more data from the source than is necessary to test the connection.

This update allows connector builder user to specify a `record_limit` that will limit the number of records returned to the client when they are testing the connector. Also enforces a maximum limit of 1000 records.

This is the server portion of https://github.com/airbytehq/airbyte/issues/19500.

## How
- Updates the `stream/read` endpoint's to accept an optional `record_limit` key in the request body.
- Adds a `MAX_RECORD_LIMIT` whose values is 1000; if the limit input by the user is not set or exceeds 1000, `MAX_RECORD_LIMIT` will be enforced.
- Stops iteration over the `messages` generator (which fetches records from the source) once `<limit>` records have been seen.

## Recommended reading order
1. `airbyte-connector-builder-server/connector_builder/impl/default_api.py`
2. `airbyte-connector-builder-server/src/main/openapi/openapi.yaml`

## 🚨 User Impact 🚨
No breaking changes. The user will see `<limit>` records when they click `Test` to test the connector, rather than all messages that would have been fetched before this change.

## Tests

</details>

</details>

</details>


airbyte-connector-builder-server/src/main/openapi/openapi.yaml
```diff
@@ -144,6 +144,9 @@ components:
           type: object
           description: The AirbyteStateMessage object to use as the starting state for this read
           # $ref: "#/components/schemas/AirbyteProtocol/definitions/AirbyteStateMessage"
+        record_limit:
```
sherifnada, line 147: @lmossman do we need to regenerate any typescript code? (if yes we should make it more obvious/automatic for anyone touching this code)

airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
```diff
@@ -267,6 +271,85 @@ def test_read_stream_with_logs():
...
+        record_message("hashiras", {"name": "Mitsuri Kanroji"}),
+        response_log_message(response),
+    ])
+    n_records = 2
+    if request_record_limit is None:
+        record_limit = max_record_limit
+    else:
+        record_limit = min(request_record_limit, max_record_limit)
+
+    with patch.object(DefaultApiImpl, "MAX_RECORD_LIMIT", max_record_limit):
```
sherifnada, line 307: feels like having to patch in this way breaks encapsulation. Does it make sense to make this an optional param on `DefaultApiImpl`?

airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
```diff
@@ -267,6 +271,85 @@ def test_read_stream_with_logs():
...
+        record_message("hashiras", {"name": "Shinobu Kocho"}),
+        record_message("hashiras", {"name": "Muichiro Tokito"}),
+        request_log_message(request),
+        response_log_message(response),
+        record_message("hashiras", {"name": "Mitsuri Kanroji"}),
+        response_log_message(response),
+    ])
+    request_record_limit = 0
+
+    with patch.object(DefaultApiImpl, "_create_low_code_adapter", return_value=mock_source_adapter):
```
sherifnada, line 340: same question, is this telling us that our code structure needs to change?

sherifnada (commented):
Overall LGTM but just a question on test/code structure

airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
clnoll: I agree with you, doesn't quite feel right to have `MAX_RECORD_LIMIT` on `DefaultApiImpl`. I'll move it to a constant outside of the class.

sherifnada: I just meant it would be better to patch it using depenency injection i.e: make this an init variable for `DefaultApiImpl`

airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
clnoll: Is there a particular structural change you had in mind? I imagine this function was patched out of convenience (it's used throughout the `test_default_api.py` file) - looks to me like there are multiple places downstream that we could patch instead if we wanted but that might be a little trickier. As a simple alternative, `_create_low_code_adapter` is a `staticmethod` and doesn't need to be a method on the class - it could be moved to its own function if we don't want to surface the details of `DefaultApiImpl` in tests.

airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
clnoll: Okay cool, I think that could be nice and tidy. Will make that change.

airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
clnoll: I think one restructuring that could make sense would be to move some implementation out of `DefaultApiImpl`, and have a separate class doing the business logic of fetching records. That separate class could take an `adapter` object, which could be injected during tests. Thoughts?

airbyte-connector-builder-server/src/main/openapi/openapi.yaml
lmossman: The generated typescript code is not committed to github, and since this is just adding a new optional parameter it won't break any of our code that consumes that generated typescript code. So this shouldn't require any frontend changes

airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
clnoll: I made a lighter-weight change than moving all of the record-fetching out of `DefaultApiImpl`, injecting the adapter class into `DefaultApiImpl.__init__` instead. This cleans up the patches.

airbyte-connector-builder-server/connector_builder/impl/default_api.py
```diff
@@ -23,11 +20,18 @@
...
 from connector_builder.generated.models.streams_list_request_body import StreamsListRequestBody
 from connector_builder.impl.low_code_cdk_adapter import LowCodeSourceAdapter
+from fastapi import Body, HTTPException
+from jsonschema import ValidationError
 
 
 class DefaultApiImpl(DefaultApi):
     logger = logging.getLogger("airbyte.connector-builder")
 
+    def __init__(self, adapter_cls: Type, max_record_limit: int = 1000):
```
girarda, line 30: why are we passing the class instead of the object?

airbyte-connector-builder-server/connector_builder/impl/default_api.py
```diff
@@ -105,15 +109,22 @@ async def read_stream(self, stream_read_request_body: StreamReadRequestBody = Bo
         Using the provided manifest and config, invokes a sync for the specified stream and returns groups of Airbyte messages
         that are produced during the read operation
         :param stream_read_request_body: Input parameters to trigger the read operation for a stream
+        :param limit: The maximum number of records requested by the client
```
girarda, line 112: should we explicitly define the range as [1, max_record_limit]?

airbyte-connector-builder-server/src/main/openapi/openapi.yaml
```diff
@@ -144,6 +144,9 @@ components:
...
+          type: integer
+          description: Number of records that will be returned to the client from the connector builder (max of 1000)
```
girarda, line 149: the minimum and maximum could be defined here so the FE doesn't have to redefine it https://swagger.io/docs/specification/data-models/data-types/#numbers

airbyte-connector-builder-server/connector_builder/impl/default_api.py
clnoll: The object requires a manifest, which we don't have when `DefaultApiImpl` is instantiated.

airbyte-connector-builder-server/connector_builder/impl/default_api.py
clnoll: Yes, good call. Added.

airbyte-connector-builder-server/src/main/openapi/openapi.yaml
clnoll: Done.

clnoll:
@girarda I've addressed your comments if you want to take another look.

girarda (approved):

airbyte-connector-builder-server/connector_builder/impl/default_api.py
```diff
@@ -27,7 +27,8 @@
 class DefaultApiImpl(DefaultApi):
     logger = logging.getLogger("airbyte.connector-builder")
 
-    def __init__(self, max_record_limit: int = 1000):
+    def __init__(self, adapter_cls: Type, max_record_limit: int = 1000):
```
sherifnada, line 30: I think there are two problems with this type signature: 
1. `Type` is not very informative (is it the correct type? it seems imported from airbyte_cdk.models which doesn't seem right?) 
2. as a result it's not clear what methods can be called on that type and with what params (e.g: how does the reader know the `adapter_cls` can be called with a `manifest` argument?) 

Example of a signature which solves these might be: 

```
class CdkAdapter(abc):
  <..describe methods we expect this class to have..>

class DefaultApiImpl(..):
  def __init__(self, adapter_cls: Callable[[], CdkAdapter], ...
```

sherifnada (changes requested):

airbyte-connector-builder-server/connector_builder/impl/default_api.py
clnoll: Ah, good call @sherifnada - that was supposed to be lowercase `type`. I like your suggestion to use an interface to be clearer about the type and went ahead and made that change.

sherifnada (approved):

PR #20577, opened 2022-12-16
🪟🎉 Add copy stream button

### LINKED ISSUE #19108: [Connector Builder UI] Duplicate full stream configuration
Add the ability for users to duplicate a full stream configuration into a new stream.

Acceptance Criteria:
- A button exists in the stream view to copy/duplicate that stream into a new one
- Users can decide the name and path of the new stream in a modal that pops up before the first stream is duplicated

flash1293:
## What

Closes #19108

Adds a copy stream button.

https://user-images.githubusercontent.com/1508364/208134495-eeb79523-2bce-4ee8-b4ec-75da807bbc78.mp4

## How

Most of the logic was already present in the `AddStreamButton` component. Extended that slightly to take a custom button element and initial values (which correspond with the values of the currently selected stream in this case) and added it next to the delete button.

## 🚨 User Impact 🚨

Ability to copy streams with all settings.


airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
```diff
@@ -49,7 +51,19 @@ export const StreamConfigView: React.FC<StreamConfigViewProps> = ({ streamNum })
       {/* Not using intl for the labels and tooltips in this component in order to keep maintainence simple */}
       <BuilderTitle path={streamFieldPath("name")} label="Stream Name" size="md" />
       <div className={styles.controls}>
-        <button className={styles.deleteButton} type="button" onClick={handleDelete}>
+        <AddStreamButton
```
lmossman, line 54: nit: could you add a small gap between this button and the delete button?

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
```diff
@@ -49,7 +51,19 @@ export const StreamConfigView: React.FC<StreamConfigViewProps> = ({ streamNum })
...
+          onAddStream={(addedStreamNum) => {
+            setSelectedView(addedStreamNum);
+            setTestStreamIndex(addedStreamNum);
+          }}
+          initialValues={field.value[streamNum]}
+          button={
+            <button className={styles.controlButton} type="button" onClick={handleDelete}>
```
lmossman, line 61: This button should not have `onClick={handleDelete}`. Looks like it is being overwritten in AddStreamButton which I suppose is why this didn't cause the button to delete the stream, but it still shouldn't be here

lmossman (approved):
Left a couple small comments but I don't think they need another look from me once addressed.

Tested locally and it works as expected

PR #20585, opened 2022-12-16
Skip validation when previous state is empty due to reset

jdpgrailsdev:
## What

Fix an edge case related to resets and syncs where at least one stream doesn't produce any data.  The specific scenario plays out something like this:

* New connection created
* Connection is reset before any state is ever saved
    * This saves the state as `LEGACY` with a value of `{"state":{}}`
* Sync is attempted with more than one configured stream
* State save fails due to the validation on migration from legacy to per-stream state because not all streams produced data    

## How
* Ignore state migration validation if previous state is `LEGACY` **AND** is empty.  In this case, we don't care if we overwrite the existing state, as it has nothing in it and we don't risk losing state.

## Recommended reading order
1. `PersistStateActivityImpl.java`

## Tests

* Added unit test to verify expected behavior
* All other tests pass
* Project builds locally


airbyte-workers/src/test/java/io/airbyte/workers/temporal/sync/PersistStateActivityTest.java
```diff
@@ -156,12 +162,50 @@ void testPersistWithGlobalStateDuringMigration() throws ApiException {
...
+        .withType(AirbyteStateType.STREAM)
+        .withStream(
+            new AirbyteStreamState().withStreamDescriptor(CatalogHelpers.extractDescriptor(stream2)));
+    final JsonNode jsonState = Jsons.jsonNode(List.of(stateMessage1, stateMessage2));
+    final State state = new State().withState(jsonState);
+
+    final AirbyteApiClient airbyteApiClient1 = mock(AirbyteApiClient.class);
+    final StateApi stateApi1 = mock(StateApi.class);
+    final ConnectionState connectionState = mock(ConnectionState.class);
+    Mockito.lenient().when(connectionState.getStateType()).thenReturn(ConnectionStateType.LEGACY);
```
benmoriceau, line 195: Why do we bypass strict stubbing? Do we need those mock?

benmoriceau (approved):

jdpgrailsdev:
@benmoriceau I discovered a failing acceptance test which lead me to add an additional `null` check.  There is a scenario where the `ConnectionState` object may be non-null, but the state object stored within it is `null`.  The code has been refactored a bit to make it easier to read.  Specifically, take a look at the `isStateEmpty` method in `PersistStateActivityImpl`.  I also added tests to cover the three different null/empty state scenarios.  Please re-review at your convenience.

gosusnp (approved):

PR #20588, opened 2022-12-16
🪟🎉 Connector form: Respect `default` value for oneOf properties

### LINKED ISSUE #17352: Consider default values for oneOf fields in connector specs

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/17352

This PR makes sure the specified default of a oneOf property is respected if it matches one possible value of the shared const property of all conditions.

E.g. if there is a property like this in the spec:
```
      replication_method: {
        type: "object",
        title: "Replication Method",
        default: "CDC",
        oneOf: [
          {
            title: "Standard",
            required: ["method"],
            properties: {
              method: {
                type: "string",
                const: "STANDARD",
                order: 0,
              },
            },
          },
          {
            title: "Logical Replication (CDC)",
            required: ["method"],
            properties: {
              method: {
                type: "string",
                const: "CDC",
                order: 0,
              },
              data_to_sync: {
                title: "Data to Sync",
                type: "string",
                default: "Existing and New",
                enum: ["Existing and New", "New Changes Only"],
                order: 1,
              },
            },
          },
        ],
      },
```

the "CDC" option will be picked initially when configuring a fresh source, even though the "Standard" option is listed first.

## How

When setting default values, try to find the default property of the form node in the list of const selection keys. If a match is found, pick that condition. Otherwise fall back to the first one.

There is no regular connector where this can currently be observed, so you need to fake a spec like this in the frontend (e.g. by changing the default property in the ConnectorForm before it's passed down to the hook).


lmossman (approved):
Changes LGTM. Tested locally by updating my local airbyte database to set the `default` property on the source-postgres connector, and it correctly used that value when creating a new connector on local dev. 

@flash1293 one note: I think this should be merged into `flash1293/connector-form-remove-ui-widget-state` rather than `feature/connector-builder`, since this isn't really related to the connector builder

PR #20589, opened 2022-12-16
pass workspace id to sync workflow and use it to selectively enable field selection

mfsiega-airbyte:
## What
Support feature-flagging field selection by workspace id.

Also ensures that the flags get passed properly through.

## How
- Pass the workspace id in to the sync workflow.
- Parse a list of workspace ids from the environment.
- If the workspace running the sync is in the list, enable field selection.


airbyte-persistence/job-persistence/src/main/java/io/airbyte/persistence/job/DefaultJobCreator.java
```diff
@@ -60,7 +60,8 @@ public Optional<Long> createSyncJob(final SourceConnection source,
                                       final List<StandardSyncOperation> standardSyncOperations,
                                       @Nullable final JsonNode webhookOperationConfigs,
                                       final StandardSourceDefinition sourceDefinition,
-                                      final StandardDestinationDefinition destinationDefinition)
+                                      final StandardDestinationDefinition destinationDefinition,
+                                      UUID workspaceId)
```
pmossman, line 64: nit: `final UUID workspaceId`

pmossman (approved):
Looks straightforward to me!

mfsiega-airbyte:
@pmossman @benmoriceau I had a bit of a journey to figure out how to get these flags set properly. Not 100% sure whether everything here is necessary but it seems to work as expected?

benmoriceau (approved):

PR #20590, opened 2022-12-16
Fix OSS integration tests for docs (#3663)

tkorenko:
## What
+ In "Docs PR check mode" stop before "yarn deploy"

### DIFF
--- a/tools/bin/deploy_docusaurus
+++ b/tools/bin/deploy_docusaurus
@@ -79,6 +79,12 @@ if ! yarn build; then
   exit 1
 fi
 
+# "check PR mode" -- we run "yarn build" and stop before "yarn deploy"
+if [ "${SKIP_DEPLOY}" = "yes" ]; then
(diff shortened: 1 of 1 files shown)


supertopher (approved):
taras can we make this run locally without a failure?  

otherwise looks great

tools/bin/deploy_docusaurus
```diff
@@ -79,6 +79,12 @@ if ! yarn build; then
   exit 1
 fi
 
+# "check PR mode" -- we run "yarn build" and stop before "yarn deploy"
+if [ "${SKIP_DEPLOY}" = "yes" ]; then
```
supertopher, line 83: SKIP_DEPLOY doesn't exist for local runs

tkorenko: which is perfectly fine for local deploys

tkorenko: we get then `if [ "" = "yes" ]; then ... "

PR #20595, opened 2022-12-16
🪟 🎉 Display "Credits" in sidebar and add info when user is running out of credits

### LINKED ISSUE #19571: Display "Credits" in the left hand sidebar
We've received questions from users about how to convert to a paid account, and we want to add clarity to users where they can buy credits. The "coin" icon in the left hand sidebar shows the current balance of the workspace:
Screen Shot 2022-11-17 at 1 54 02 PM

Instead of showing the balance, replace the balance with the word "Credits".  (Alternative options considered but not chosen: Billing, Usage as they are less indicative of the content on that page and the intent of the change)

if a workspace falls below 20 credits, add a small yellow/orange/red notification dot to the icon in the menu, so that the user is drawn to take action by clicking on the icon.

From there, the user would then be directed to /credits, where we show an info box near the top of the page (where we currently show the email verification info box) letting the user know they are low on credits and should buy more credits to ensure their syncs continue to run successfully. 

If the workspace falls below the threshold where we stop all syncs, we make the info box with a red background color and modify the messaging to indicate their syncs have been automatically stopped and they need to buy more credits to enable the connections again.

letiescanciano:
Closes #19571

Demo: https://www.loom.com/share/ca43d6b72dcc42208add0bd91e1b1efc

## What
- Add "Credits" in credits sidebar menu button
- Add Info boxes in `<CreditsPage />` to inform users about low/zero balance

### DIFF
--- a/airbyte-webapp/src/components/ui/InfoBox/InfoBox.tsx
+++ b/airbyte-webapp/src/components/ui/InfoBox/InfoBox.tsx
@@ -21,7 +21,7 @@ export const InfoBox: React.FC<React.PropsWithChildren<Props>> = ({ children, cl
   return (
     <Box className={className}>
       {icon && <FontAwesomeIcon size="lg" icon={icon} />}
-      <div>{children}</div>
+      {children}
     </Box>
   );
 };
--- a/airbyte-webapp/src/components/ui/InfoBox/InfoBox.tsx
+++ b/airbyte-webapp/src/components/ui/InfoBox/InfoBox.tsx
@@ -5,10 +5,11 @@ import styled from "styled-components";
 interface Props {
   className?: string;
   icon?: IconDefinition;
+  error?: boolean;
 }
 
-const Box = styled.div`
-  background: ${({ theme }) => theme.yellow100};
+const Box = styled.div<{ error?: boolean }>`
+  background: ${({ theme, error }) => (error ? theme.red50 : theme.yellow100)};
   border-radius: 8px;
   padding: 18px 25px 22px;
   font-size: 14px;
@@ -17,9 +18,9 @@ const Box = styled.div`
   align-items: center;
 `;
 
-export const InfoBox: React.FC<React.PropsWithChildren<Props>> = ({ children, className, icon }) => {
+export const InfoBox: React.FC<React.PropsWithChildren<Props>> = ({ children, className, icon, error }) => {
   return (
-    <Box className={className}>
+    <Box className={className} error={error}>
       {icon && <FontAwesomeIcon size="lg" icon={icon} />}
       {children}
     </Box>
--- a/airbyte-webapp/src/components/ui/InfoBox/InfoBox.tsx
+++ b/airbyte-webapp/src/components/ui/InfoBox/InfoBox.tsx
@@ -5,11 +5,11 @@ import styled from "styled-components";
 interface Props {
   className?: string;
   icon?: IconDefinition;
-  error?: boolean;
+  variant?: "default" | "error";
 }
 
-const Box = styled.div<{ error?: boolean }>`
-  background: ${({ theme, error }) => (error ? theme.red50 : theme.yellow100)};
+const Box = styled.div<{ variant: "default" | "error" }>`
+  background: ${({ theme, variant }) => (variant === "error" ? theme.red50 : theme.yellow100)};
   border-radius: 8px;
   padding: 18px 25px 22px;
   font-size: 14px;
@@ -18,9 +18,14 @@ const Box = styled.div<{ error?: boolean }>`
   align-items: center;
 `;
 
-export const InfoBox: React.FC<React.PropsWithChildren<Props>> = ({ children, className, icon, error }) => {
+export const InfoBox: React.FC<React.PropsWithChildren<Props>> = ({
+  children,
+  className,
+  icon,
+  variant = "default",
+}) => {
   return (
-    <Box className={className} error={error}>
+    <Box className={className} variant={variant}>
       {icon && <FontAwesomeIcon size="lg" icon={icon} />}
       {children}
     </Box>
--- a/airbyte-webapp/src/packages/cloud/locales/en.json
+++ b/airbyte-webapp/src/packages/cloud/locales/en.json
@@ -153,6 +153,8 @@
   "credits.emailVerificationRequired": "You need to verify your email address before you can buy credits.",
   "credits.emailVerification.resendConfirmation": "We send you a new verification link.",
   "credits.emailVerification.resend": "Send verification link again",
+  "credits.lowBalance": "Your credit balance is low. You need to add more credits to prevent your syncs from being stopped",
+  "credits.zeroBalance": "All your syncs have been stopped because your credit balance is 0. Add more credits to enable your syncs again",
 
   "firebase.auth.error.invalidPassword": "Incorrect password",
   "firebase.auth.error.networkRequestFailed": "There appears to be a network issue. Please try again later.",
@@ -174,5 +176,7 @@
   "webapp.cannotReachServer": "Cannot reach server.",
 
   "inviteUsersHint.message": "Need help from a teammate to set up the {connector}?",
-  "inviteUsersHint.cta": "Invite users"
+  "inviteUsersHint.cta": "Invite users",
+
+  "sidebar.credits": "Credits"
 }
--- a/airbyte-webapp/src/packages/cloud/locales/en.json
+++ b/airbyte-webapp/src/packages/cloud/locales/en.json
@@ -153,8 +153,8 @@
   "credits.emailVerificationRequired": "You need to verify your email address before you can buy credits.",
   "credits.emailVerification.resendConfirmation": "We send you a new verification link.",
   "credits.emailVerification.resend": "Send verification link again",
-  "credits.lowBalance": "Your credit balance is low. You need to add more credits to prevent your syncs from being stopped",
-  "credits.zeroBalance": "All your syncs have been stopped because your credit balance is 0. Add more credits to enable your syncs again",
+  "credits.lowBalance": "Your credit balance is low. You need to add more credits to prevent your syncs from being stopped.",
+  "credits.zeroBalance": "All your syncs have been stopped because your credit balance is 0. Add more credits to enable your syncs again.",
 
   "firebase.auth.error.invalidPassword": "Incorrect password",
   "firebase.auth.error.networkRequestFailed": "There appears to be a network issue. Please try again later.",
--- a/airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.module.scss
+++ b/airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.module.scss
@@ -1,0 +1,28 @@
+@use "../../../../../../../scss/colors";
+
+.container {
+  width: 100%;
+
+  &.zeroBalance {
+    background-color: colors.$red-50;
+
+    & > svg {
+      color: colors.$red-300;
+    }
+  }
+
+  &.lowBalance {
+    background-color: colors.$yellow-50;
+
+    & > svg {
+      color: colors.$yellow-500;
+    }
+  }
+}
+
+.wrapper {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  width: 100%;
+}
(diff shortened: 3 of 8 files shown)


airbyte-webapp/src/components/ui/InfoBox/InfoBox.tsx
```diff
@@ -21,7 +21,7 @@ export const InfoBox: React.FC<React.PropsWithChildren<Props>> = ({ children, cl
   return (
     <Box className={className}>
       {icon && <FontAwesomeIcon size="lg" icon={icon} />}
-      <div>{children}</div>
```
letiescanciano, line 24: Shouldn't this be the responsibility of the children? 
I remove it in order to be able to have full control on the children node

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.tsx
```diff
@@ -0,0 +1,37 @@
...
+import classNames from "classnames";
+import { FormattedMessage } from "react-intl";
+
+import { InfoBox } from "components/ui/InfoBox";
+
+import { useGetCloudWorkspace } from "packages/cloud/services/workspaces/CloudWorkspacesService";
+import { useCurrentWorkspace } from "services/workspaces/WorkspacesService";
+
+// eslint-disable-next-line css-modules/no-unused-class
+import styles from "./LowCreditBalanceHint.module.scss";
```
letiescanciano, line 11: Since I'm accessing classes dynamically, the linter can't determine if classes on `LowCreditBalanceHint.module.scss` are read, therefore I need to add ignore it.

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/RemainingCredits.tsx
```diff
@@ -116,29 +118,44 @@ const RemainingCredits: React.FC<Props> = ({ selfServiceCheckoutEnabled }) => {
...
+            isLoading={isLoading || isWaitingForCredits}
+            icon={<FontAwesomeIcon icon={faPlus} />}
+          >
+            <FormattedMessage id="credits.buyCredits" />
+          </Button>
+          <Button size="xs" onClick={() => window.open(links.contactSales, "_blank")}>
+            <FormattedMessage id="credits.talkToSales" />
+          </Button>
+        </Actions>
+      </Block>
```
letiescanciano, line 157: This snippet remains the same.

airbyte-webapp/src/packages/cloud/views/layout/SideBar/SideBar.tsx
```diff
@@ -82,11 +84,14 @@ const SideBar: React.FC = () => {
         </ul>
       </div>
       <ul className={styles.menu}>
-        <li>
+        <li className={styles.creditsButton}>
           <NavLink className={navLinkClassName} to={CloudRoutes.Credits}>
+            {cloudWorkspace.remainingCredits <= LOW_BALANCE_CREDIT_TRESHOLD && (
+              <Indicator className={styles.lowBalanceIndicator} />
+            )}
```
letiescanciano, line 91: Show indicator only when credits <= 20

airbyte-webapp/src/components/ui/InfoBox/InfoBox.tsx
timroes: I guess the only reason might have been vertical alignment issues why this was here. But if it works in all places without it, no reason to keep it around.

airbyte-webapp/src/packages/cloud/locales/en.json
```diff
@@ -153,6 +153,8 @@
   "credits.emailVerificationRequired": "You need to verify your email address before you can buy credits.",
   "credits.emailVerification.resendConfirmation": "We send you a new verification link.",
   "credits.emailVerification.resend": "Send verification link again",
+  "credits.lowBalance": "Your credit balance is low. You need to add more credits to prevent your syncs from being stopped",
```
timroes, line 156:
```suggestion
  "credits.lowBalance": "Your credit balance is low. You need to add more credits to prevent your syncs from being stopped.",
```

airbyte-webapp/src/packages/cloud/locales/en.json
```diff
@@ -153,6 +153,8 @@
...
+  "credits.zeroBalance": "All your syncs have been stopped because your credit balance is 0. Add more credits to enable your syncs again",
```
timroes, line 157:
```suggestion
  "credits.zeroBalance": "All your syncs have been stopped because your credit balance is 0. Add more credits to enable your syncs again.",
```

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.tsx
timroes: We usually try to create a map to status to classname, so we don't need to disable linting here, e.g.

```js
const statusStyles = {
  "zeroBalance": styles.zeroBalance,
  "lowBalance": styles.lowBalance,
};
```

Since this is only two options, I feel we can inline this more nicely into the `className` call in line 30 directly:

```js
classNames(styles.container, {
  [styles.zeroBalance]: status === "zeroBalance",
  [styles.lowBalance]: status === "lowBalance",
});
```

This way we get proper checking on the used and unused classnames still from the linter.

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export const LOW_BALANCE_CREDIT_TRESHOLD = 20;
+
+export const LowCreditBalanceHint: React.FC = ({ children }) => {
```
timroes, line 15:
```ts
export const LowCreditBalanceHint: React.FC<React.PropsWithChildren<unknown>> = ({ children }) => {
```

Adding the `PropsWithChildren` will ease the migration to React 18 later on. So we should always wrap the actual props in it when accessing `children` (or `unknown` in case we don't have any other props).

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.module.scss
```diff
@@ -0,0 +1,28 @@
...
+  &.zeroBalance {
+    background-color: colors.$red-50;
+
+    & > svg {
+      color: colors.$red-300;
+    }
+  }
+
+  &.lowBalance {
+    background-color: colors.$yellow-50;
```
timroes, line 15: I'm not sure we should really diverge from our regular background color that `InfoBox` has here. I think it would be better if we stay aligned with the color and not trying to overwrite it. I'd also recommend not overwriting the icon color for the same reason (not to end up with too many unaligned styles around the app).

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.module.scss
```diff
@@ -0,0 +1,28 @@
+@use "../../../../../../../scss/colors";
+
+.container {
+  width: 100%;
+
+  &.zeroBalance {
+    background-color: colors.$red-50;
```
timroes, line 7: Could we maybe make this a property on `InfoBox` instead, like `style="error"` so that we can use this across the app, and make sure it's not a one-off styling here.

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.module.scss
letiescanciano: I've followed the colors that we have in [Figma](https://www.figma.com/file/gMNRYVJav25pdsoxAyW7Qp/00_01_WEB-APP-LIBRARIES?node-id=1833%3A3458&t=HDnTXz1PvdIMoXap-0) for this component (both bg and icon). But happy to consolidate styling.
Should I follow the styles on the figma or stick with the existing ones in code?

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.module.scss
letiescanciano: Sounds great!

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.module.scss
josephkmh: @letiescanciano good catch, IMO that looks like a design inconsistency that we should address with Nico. But I think the component you linked is rather a toast message than an `InfoBox`, right? For now, can we keep the existing colors we have in code (i.e. `yellow-100`)?

airbyte-webapp/src/components/ui/InfoBox/InfoBox.tsx
```diff
@@ -5,10 +5,11 @@ import styled from "styled-components";
 interface Props {
   className?: string;
   icon?: IconDefinition;
+  error?: boolean;
 }
 
-const Box = styled.div`
-  background: ${({ theme }) => theme.yellow100};
+const Box = styled.div<{ error?: boolean }>`
+  background: ${({ theme, error }) => (error ? theme.red50 : theme.yellow100)};
```
josephkmh, line 12: Kind of a nitpick here: could we do something like `variant: 'default' | 'error'` instead of `error?: boolean`? That would make it easier later to add other variants of the `InfoBox`. It also avoids a potential situation like `<InfoBox error={true} someOtherStyleVariant={true} />` where the styling becomes unclear.

The prop signature for `InfoBox` could then set a default value like this: 
```
React.FC<React.PropsWithChildren<Props>> = ({ children, className, icon, variant = 'default' })
```

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const LowCreditBalanceHint: React.FC<React.PropsWithChildren<unknown>> = ({ children }) => {
+  const workspace = useCurrentWorkspace();
+  const cloudWorkspace = useGetCloudWorkspace(workspace.workspaceId);
+
+  if (cloudWorkspace.remainingCredits > LOW_BALANCE_CREDIT_TRESHOLD) {
+    return null;
+  }
+
+  const status = cloudWorkspace.remainingCredits === 0 ? "zeroBalance" : "lowBalance";
```
josephkmh, line 21: I think in some cases we have workspaces with negative `remainingCredits` (see my cloud workspace [here](https://cloud.airbyte.io/workspaces/c508c8ed-38d2-4645-b8c6-dd9140316279) for an example), so I think we should do a `<= 0` comparison here.

joeykmh (changes requested):
Just a couple minor comments!

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint.module.scss
letiescanciano: There's two actually Toast and Message. It was the closest to an `Infobox` I could find but agree, let's keep the ones in code for consistency 👍🏼

airbyte-webapp/src/components/ui/InfoBox/InfoBox.tsx
letiescanciano: I like it!

joeykmh (approved):
Looks good!

nataliekwong:
@timroes It looks like the notification button (red or yellow) is no longer showing up in the lef-thand sidebar. Do you know if that was intentional?

timroes:
Yes that is indeed a bug in the current code, it's somehow mixed up checking for connector updates with the indicator for low credits. Could you maybe fill a new issue for this and cc Joey and Teal on it?

PR #20597, opened 2022-12-16
[source-tiktok] Remove audience reports hourly streams

brianjlai:
## What
Resolves an issue for certain TikTok streams where we were receiving back:
```
{'code': 40002, 'message': 'Conflict occurs. The input params including metrics and dimensions have no relation.', 'request_id': '20221215073730DB34302A4354C213D7D7', 'data': {}}
```

## How
For some of our Audience Report streams, TikTok would be throwing an error when we included the `stat_time_hour` dimension. After going through the TikTok API documentation [here](https://ads.tiktok.com/marketing_api/docs?id=1707957217727489). It appears that they have deprecated that dimension. Other time dimensions like Daily (and Lifetime for Advertiser Audience Reports) still work as expected. Given that we no longer can make requests using these dimensions we should remove these streams entirely to prevent future customers from configuring them during setup because they are effectively unusable. 

Note: This would affect existing customers, but most likely all customers using these streams would have failing syncs because of it. Customers that did not enable these streams should be unaffected.


brianjlai:
my bad i thought i was on master when I wrote this. I'm gonna close and reopen this for simplicity

PR #20598, opened 2022-12-16
[source-tiktok] Remove audience reports hourly streams

brianjlai:
## What
Resolves an issue for certain TikTok streams where we were receiving back:
```
{'code': 40002, 'message': 'Conflict occurs. The input params including metrics and dimensions have no relation.', 'request_id': '20221215073730DB34302A4354C213D7D7', 'data': {}}
```

## How
For some of our Audience Report streams, TikTok would be throwing an error when we included the `stat_time_hour` dimension. After going through the TikTok API documentation [here](https://ads.tiktok.com/marketing_api/docs?id=1707957217727489). It appears that they have deprecated that dimension. Other time dimensions like Daily (and Lifetime for Advertiser Audience Reports) still work as expected. Given that we no longer can make requests using these dimensions we should remove these streams entirely to prevent future customers from configuring them during setup because they are effectively unusable. 

Note: This would affect existing customers, but most likely all customers using these streams would have failing syncs because of it. Customers that did not enable these streams should be unaffected.


airbyte-integrations/connectors/source-tiktok-marketing/integration_tests/streams.json
```diff
@@ -108,13 +108,6 @@
         "json_schema": {},
         "supported_sync_modes": ["full_refresh"]
       },
-      {
```
brianjlai, line 111: This streams.json I don't think is actually used in our acceptance test YAML but just keeping this up to date regardless

airbyte-integrations/connectors/source-tiktok-marketing/unit_tests/unit_test.py
```diff
@@ -134,8 +134,8 @@ def test_random_items(prepared_prod_args):
 @pytest.mark.parametrize(
     "config, stream_len",
     [
-        (PROD_CONFIG_FILE, 26),
```
brianjlai, line 137: We have 4 less hourly streams (1 for each audience report) in prod accounts
We have 3 less hourly streams (advertiser audience report is prod only) for sandbox accounts.

sherifnada (approved):

airbyte-integrations/connectors/source-tiktok-marketing/acceptance-test-config.yml
```diff
@@ -28,22 +28,22 @@ tests:
   discovery:
     - config_path: "secrets/prod_config.json"
       backward_compatibility_tests_config:
-        disable_for_version: "0.1.17"
+        disable_for_version: "1.0.0"
```
brianjlai, line 31: Removing streams are considered a breaking change, however we need to do this because the streams aren't usable any more. We will communicate this to the customers attempting to use these non-working streams to reset or disable them.

girarda (approved):

girarda:
thank you for taking care of this @brianjlai!

Issue #20322, opened 2022-12-09
Incremental stream more granular than a day using DatetimeStreamSlicer without lookback might lose records


maxi297:
## Tell us about the problem you're trying to solve
Following this discussion https://github.com/airbytehq/airbyte/pull/20156#discussion_r1042214726, it seems like we might be losing some records. For example, let's use source-gnews' `search` stream:

* Date format: `"%Y-%m-%d %H:%M:%S"`. Therefore, the most granular unit of time is the second. 
* Config
```
{
  "api_key": <an api key>,
  "query": "bogart",
  "start_date": "2022-11-20 00:00:00",
  "end_date": "2022-11-30 00:00:00"
}
```
* State
```
{
  "search": {
    "publishedAt": "2022-11-25T09:00:26Z"
  }
}
```
* Catalog
```
{
  "streams": [
    {
      "stream": {
        "name": "search",
        "json_schema": {},
        "supported_sync_modes": ["full_refresh", "incremental"]
      },
      "sync_mode": "incremental",
      "cursor_field": ["timestamp"],
      "destination_sync_mode": "overwrite"
    }
  ]
}
```

When running `python main.py read --config config.json --catalog catalog.json --state state.json`, DatetimeStreamSlicer will use the partition `[{'start_time': '2022-11-26T09:00:26Z', 'end_time': '2022-11-30T00:00:00Z'}]`. This means that if there was an entry between the `2022-11-26 00:00:00` (datetime of `config.end_date`) and `2022-11-26T09:00:26Z` (start_time calculated by the DatetimeStreamSlicer), there would be missing records that Airbyte does not see. Note that if `config.end_time` was not provided, the missing records would only be between the time of the last execution instead of `config.end_date`.

This is because of [this hardcoded timedelta](https://github.com/airbytehq/airbyte/blob/abee21e51e391055b4f073890b2312270dc81f39/airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py#L157) which increate the cursor on the state from one day in every case. This makes sense when the granularity of the cursor was in days. However, it can cause issues when:
* The granularity is smaller than a day (see example above)
* The granularity is bigger than a day: in that case, we would consume records multiple times

## Describe the solution you’d like
[...]

PR #20599, opened 2022-12-16
[ISSUE #20322] use datetime_format to calculate datetime stream slices

maxi297:
## What
An issue was identified in the datetime stream slicer where some records are not fetched (see [this ticket](https://github.com/airbytehq/airbyte/issues/20322)). We later got report that [this issue was occurring for some users](https://github.com/airbytehq/airbyte/pull/20500). The lost of records can occur in two circumstances:
1. When we start fetching from "cursor + 1 day" (see [this](https://github.com/airbytehq/airbyte/blob/7ae05b16c6dc9975d0ce8c855212e5a487d2cac1/airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py#L157))
2. When we generate the end of a slice without considering the granularity (see [this](https://github.com/airbytehq/airbyte/blob/7ae05b16c6dc9975d0ce8c855212e5a487d2cac1/airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py#L168))

## How
To solve case #1, we now start fetching from the from the cursor itself. This will generate duplication but is something needed to ensure we avoid duplication. The other alternative is to maybe lose records. The case is the following: let’s assume that we fetch hourly an API with date format %Y-%m-%d. In such cases, the end_datetime is often an interpolation of `today_utc()`. What will probably happen is:
* At 12:00 AM, the date switch from 2022-12-12 to 2022-12-13
* Let’s assume we fetch at 12:30 AM. The fetch will have the boundaries:
    * start_datetime: 2022-12-12
    * end_datetime: 2022-12-13
* Stream state will have cursor on the 2022-12-13
* At 01:30 AM we fetch again. Which start_datetime/end_datetime should we use?

To solve case #2 above, where before we would always remove one day, now we remove the most granular unit of time based on the datetime_format. Therefore:
* Given a datetime_format of `"%Y-%m-%dT%H:%M:%S.%f%z"`, the slice will now be `"start_time": "2021-01-05T00:00:00.000000+0000", "end_time": "2021-01-05T23:59:59.999999+0000"` instead of `"start_time": "2021-01-05T00:00:00.000000+0000", "end_time": "2021-01-05T00:00:00.000000+0000"`
* Given a datetime_format of `"%Y-%m-%d"`, the result will be the same i.e. `"start_time": "2021-01-05", "end_time": "2021-01-05"`

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py`: refers to the business logic and therefore should be read first
2. `airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py`: contains the implementation to find the most granular unit of time

## 🚨 User Impact 🚨
As mentioned above, we might have more duplication how since we won't stop querying a cursor until we have new records in a more recent slice. I'll put up a document soon describing what we could do to reduce such duplication.

Although there will be more duplication, this will ensure we don't lose record.


airbyte-cdk/python/unit_tests/sources/declarative/stream_slicers/test_datetime_stream_slicer.py
```diff
@@ -45,15 +45,15 @@ def mock_datetime_now(monkeypatch):
...
+                {"start_time": "2021-01-01T00:00:00.000000+0000", "end_time": "2021-01-01T23:59:59.999999+0000"},
+                {"start_time": "2021-01-02T00:00:00.000000+0000", "end_time": "2021-01-02T23:59:59.999999+0000"},
+                {"start_time": "2021-01-03T00:00:00.000000+0000", "end_time": "2021-01-03T23:59:59.999999+0000"},
+                {"start_time": "2021-01-04T00:00:00.000000+0000", "end_time": "2021-01-04T23:59:59.999999+0000"},
+                {"start_time": "2021-01-05T00:00:00.000000+0000", "end_time": "2021-01-05T23:59:59.999999+0000"},
+                {"start_time": "2021-01-06T00:00:00.000000+0000", "end_time": "2021-01-06T23:59:59.999999+0000"},
+                {"start_time": "2021-01-07T00:00:00.000000+0000", "end_time": "2021-01-07T23:59:59.999999+0000"},
+                {"start_time": "2021-01-08T00:00:00.000000+0000", "end_time": "2021-01-08T23:59:59.999999+0000"},
+                {"start_time": "2021-01-09T00:00:00.000000+0000", "end_time": "2021-01-09T23:59:59.999999+0000"},
                 {"start_time": "2021-01-10T00:00:00.000000+0000", "end_time": "2021-01-10T00:00:00.000000+0000"},
```
maxi297, line 57: Those last slices are a little bit weird but should work if we keep the assumption that boundaries are always inclusive

airbyte-cdk/python/unit_tests/sources/declarative/stream_slicers/test_datetime_stream_slicer.py
```diff
@@ -358,7 +360,7 @@ def test_stream_slices(
     )
     stream_slices = slicer.stream_slices(SyncMode.incremental, stream_state)
 
-    assert expected_slices == stream_slices
+    assert stream_slices == expected_slices
```
maxi297, line 363: Every test does `assert expected == actual` but PyTest assertions should be `assert actual == expected`. I have only changed this one but it could be interesting to write new tests using this convention

airbyte-cdk/python/unit_tests/sources/declarative/datetime/test_datetime_parser.py
```diff
@@ -44,3 +45,65 @@ def test_format_datetime(test_name, input_dt, datetimeformat, expected_output):
...
+    output_delta = DatetimeParser().find_most_granular_timedelta("%Y-%m-%dT%H:00:00")
+    _relativedelta_is_bigger_than(output_delta, datetime.timedelta(seconds=1))
+
+
+def test_given_no_known_directives_when_find_most_granular_timedelta_then_raise_error():
+    with pytest.raises(ValueError):
+        DatetimeParser().find_most_granular_timedelta("there is no known directive in this string %q")
+
+
+def _relativedelta_is_bigger_than(relative_delta: relativedelta, time_delta: datetime.timedelta):
```
maxi297, line 107: `relativedelta` does not support `__lt__` because there are not ways to know if `relativedelta(days=29) > relativedelta(months=1)`. Therefore, I found a way to evaluate this and hid it in a method

airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py
```diff
@@ -36,3 +52,17 @@ def format(self, dt: datetime.datetime, format: str) -> str:
             return str(int(dt.timestamp()))
         else:
             return dt.strftime(format)
+
+    def find_most_granular_timedelta(self, datetime_format: str) -> relativedelta:
+        """
+        The granularity is based on the 1989 C standard directives. If the datetime_format matches `%Y-%m-%dT%H:00:00` for example, this
```
maxi297, line 58: What is the expectation regarding a format like `%Y-%m-%dT%H:00:00`?

airbyte-cdk/python/unit_tests/sources/declarative/stream_slicers/test_datetime_stream_slicer.py
```diff
@@ -45,15 +45,15 @@ def mock_datetime_now(monkeypatch):
             None,
```
maxi297, line 45: TODO: add a test with a different datetime_format granularity

airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py
maxi297: [A new PR](https://github.com/airbytehq/airbyte/pull/20717) has been opened with a solution that requires user input but supports this case

maxi297:
By checking edge cases, we've found out that this solution is not optimal (see https://github.com/airbytehq/airbyte/pull/20599#discussion_r1051102598) 

A new PR has been opened with the new solution: https://github.com/airbytehq/airbyte/pull/20717

PR #20605, opened 2022-12-16
Utilize React-Lazy for Code Splitting

### LINKED ISSUE #20604: Use React Lazy to code-split
React gives us code splitting through [React Lazy](https://reactjs.org/docs/code-splitting.html).
We should utilize the output of the performance work (lighthouse reports notably) and attempt some bundle splitting to see where performance can be improved.

Baseline (Prod build, served statically locally):
Screenshot 2022-12-16 at 2 51 56 PM

krishnaglick:
## What
Resolves #20604

## How
Implements some top-level react-lazy usage, as well as a little deeper in the tree. Still room to improve.
Here's the lighthouse result.

Screenshot 2022-12-16 at 2 57 29 PM


airbyte-webapp/src/pages/routes.tsx
```diff
@@ -13,18 +13,20 @@ import { CompleteOauthRequest } from "views/CompleteOauthRequest";
...
-import CreateDestinationPage from "./destination/CreateDestinationPage";
-import { DestinationItemPage } from "./destination/DestinationItemPage";
-import { DestinationOverviewPage } from "./destination/DestinationOverviewPage";
-import { DestinationSettingsPage } from "./destination/DestinationSettingsPage";
-import PreferencesPage from "./PreferencesPage";
 import { RoutePaths, DestinationPaths } from "./routePaths";
-import SettingsPage from "./SettingsPage";
-import SourcesPage from "./SourcesPage";
+
+const ConnectionPage = React.lazy(() => import("./ConnectionPage"));
```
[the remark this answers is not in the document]
krishnaglick: No! I've gone ahead and done a Cloud pass too :)

airbyte-webapp/src/pages/destination/AllDestinationsPage/index.tsx
```diff
@@ -1 +1,4 @@
+import { AllDestinationsPage } from "./AllDestinationsPage";
+
 export { AllDestinationsPage } from "./AllDestinationsPage";
+export default AllDestinationsPage;
```
josephkmh, line 4: Could we just do the default export here? I don't think we need the named one at all.

airbyte-webapp/src/pages/destination/DestinationItemPage/index.tsx
```diff
@@ -1 +1,4 @@
+import { DestinationItemPage } from "./DestinationItemPage";
+
 export { DestinationItemPage } from "./DestinationItemPage";
+export default DestinationItemPage;
```
josephkmh, line 4: Same here, I think we only need the default export

airbyte-webapp/src/pages/destination/DestinationSettingsPage/index.ts
```diff
@@ -1 +1,4 @@
+import { DestinationSettingsPage } from "./DestinationSettingsPage";
+
 export { DestinationSettingsPage } from "./DestinationSettingsPage";
+export default DestinationSettingsPage;
```
josephkmh, line 4: Same here, I think we only need the default export

airbyte-webapp/src/packages/cloud/views/auth/LoginPage/index.tsx
```diff
@@ -1,3 +1,3 @@
 import LoginPage from "./LoginPage";
 
-export { LoginPage };
+export default LoginPage;
```
josephkmh, line 3: nitpick: these default exports could all be one-liners:

```ts
export { LoginPage as default } from "./LoginPage";
```

airbyte-webapp/src/components/ui/DatePicker/index.ts
```diff
@@ -1 +1,3 @@
-export { DatePicker } from "./DatePicker";
+import React from "react";
+
+export default React.lazy(() => import("./DatePicker"));
```
josephkmh, line 3: Could we just export this normally here, and defer the lazy import to the consumer?

joeykmh (changes requested):
The code changes look good, I just made a few small comments mostly about syntax.

I am wondering whether this change will have the desired impact. Code splitting strictly along routes means smaller JS bundle chunks at the cost of more network requests. This means route changes that currently load "instantly" (because the JS chunk was loaded on the initial page load) will now potentially show a loading spinner until all their JS dependencies are loaded over the network. This type of optimization is generally good when you want to optimize for quick individual page loads (e.g. for SEO/inbound traffic from search), but I'm not sure if this is what we really want to optimize for in Airbyte. I don't feel too strongly either way currently, but I wanted to call it out. Is this tradeoff something you considered?

Could you maybe add a comparison of the lighthouse scores before and after so we can understand the net change that code splitting is having on initial page loads? This is probably most relevant for both the login page and the first page load (`AllConnectionsPage.tsx`)

krishnaglick:
> The code changes look good, I just made a few small comments mostly about syntax.
> 
> I am wondering whether this change will have the desired impact. Code splitting strictly along routes means smaller JS bundle chunks at the cost of more network requests. This means route changes that currently load "instantly" (because the JS chunk was loaded on the initial page load) will now potentially show a loading spinner until all their JS dependencies are loaded over the network. This type of optimization is generally good when you want to optimize for quick individual page loads (e.g. for SEO/inbound traffic from search), but I'm not sure if this is what we really want to optimize for in Airbyte. I don't feel too strongly either way currently, but I wanted to call it out. Is this tradeoff something you considered?
> 
> Could you maybe add a comparison of the lighthouse scores before and after so we can understand the net change that code splitting is having on initial page loads? This is probably most relevant for both the login page and the first page load (`AllConnectionsPage.tsx`)

Screenshot 2023-01-04 at 3 35 31 PM

Screenshot 2023-01-04 at 3 35 05 PM

Current master vs this branch. A 7 point increase is pretty nice imo.
I'm aware this will increase load times when navigating around but the goal of this batch of work is to improve first-load times. Splitting out the routing like this does accomplish that.

josephkmh:
Ok, I'm fine with these changes as long as we're aware of the tradeoffs 🙂 we can always look into preloading other chunks in the background or chunking some routes together later if that makes sense.

joeykmh (approved):

PR #20606, opened 2022-12-16
fix `LOG_CONNECTOR_MESSAGES` when source has no state

pedroslopez:
## What

When using `LOG_CONNECTOR_MESSAGES=true` to have syncs log all the connector messages, if you're running from an empty state this would trigger a null pointer exception when trying to log the initial state.

## How

- Fix the issue by checking for null


evantahler (approved):
Oops!

PR #20610, opened 2022-12-16
airbyte-server: tag span of additional OAuth routes

alafanechere:
## What
I forgot to add tags from additional OAuth routes after this initial [PR](https://github.com/airbytehq/airbyte/pull/20468/files#diff-a51e8edc400f0098044b097b2548f045beef41ad99958dcbe3be3dd6d19e2ca8).

## How
Tag the root span with definition and workspace id on `complete_oauth` + `get_consent_url` for sources and destinations.

### DIFF
--- a/airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
+++ b/airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
@@ -116,8 +116,10 @@ public class OAuthHandler {
 
   public OAuthConsentRead getDestinationOAuthConsent(final DestinationOauthConsentRequest destinationOauthConsentRequest)
       throws JsonValidationException, ConfigNotFoundException, IOException {
-    ApmTraceUtils.addTagsToTrace(Map.of(WORKSPACE_ID_KEY, destinationOauthConsentRequest.getWorkspaceId(), DESTINATION_DEFINITION_ID_KEY,
-        destinationOauthConsentRequest.getDestinationDefinitionId()));
+    final Map<String, Object> traceTags = Map.of(WORKSPACE_ID_KEY, destinationOauthConsentRequest.getWorkspaceId(), DESTINATION_DEFINITION_ID_KEY,
+        destinationOauthConsentRequest.getDestinationDefinitionId());
+    ApmTraceUtils.addTagsToTrace(traceTags);
+    ApmTraceUtils.addTagsToRootSpan(traceTags);
 
     final StandardDestinationDefinition destinationDefinition =
         configRepository.getStandardDestinationDefinition(destinationOauthConsentRequest.getDestinationDefinitionId());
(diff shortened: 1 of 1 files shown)


airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
```diff
@@ -210,8 +214,10 @@ public Map<String, Object> completeSourceOAuth(final CompleteSourceOauthRequest
 
   public Map<String, Object> completeDestinationOAuth(final CompleteDestinationOAuthRequest completeDestinationOAuthRequest)
       throws JsonValidationException, ConfigNotFoundException, IOException {
-    ApmTraceUtils.addTagsToTrace(Map.of(WORKSPACE_ID_KEY, completeDestinationOAuthRequest.getWorkspaceId(), DESTINATION_DEFINITION_ID_KEY,
-        completeDestinationOAuthRequest.getDestinationDefinitionId()));
+    final Map<String, Object> traceTags = Map.of(WORKSPACE_ID_KEY, completeDestinationOAuthRequest.getWorkspaceId(), SOURCE_DEFINITION_ID_KEY,
```
pedroslopez, line 217:
```suggestion
    final Map<String, Object> traceTags = Map.of(WORKSPACE_ID_KEY, completeDestinationOAuthRequest.getWorkspaceId(), DESTINATION_DEFINITION_ID_KEY,
```

pedroslopez (changes requested):

airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
alafanechere: Always the same typo 😮‍💨

airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
```diff
@@ -162,8 +164,10 @@ public OAuthConsentRead getDestinationOAuthConsent(final DestinationOauthConsent
 
   public Map<String, Object> completeSourceOAuth(final CompleteSourceOauthRequest completeSourceOauthRequest)
       throws JsonValidationException, ConfigNotFoundException, IOException {
-    ApmTraceUtils.addTagsToTrace(Map.of(WORKSPACE_ID_KEY, completeSourceOauthRequest.getWorkspaceId(), SOURCE_DEFINITION_ID_KEY,
-        completeSourceOauthRequest.getSourceDefinitionId()));
+    final Map<String, Object> traceTags = Map.of(WORKSPACE_ID_KEY, completeSourceOauthRequest.getWorkspaceId(), SOURCE_DEFINITION_ID_KEY,
+        completeSourceOauthRequest.getSourceDefinitionId());
```
pedroslopez, line 168: should we have some sort of helper like getOAuthTraceTags or getDestinationTraceTags to make sure they're consistent?

alafanechere: I'll definitely consider more helper if we trace more things in the future 😄

pedroslopez (approved):

PR #20616, opened 2022-12-17
:window: :bug: Add DbtCloudErrorBoundary

### LINKED ISSUE #20441: Error statuses from dbt Cloud requests should not crash transformation tab in cloud build
The backend forwards any not-specifically-caught error responses from dbt Cloud to the frontend; at present, these error statuses get propagated right up to the default page-wide error boundary. Since we don't control the source of these errors, we should wrap the dbt cloud transformation card component in its own error boundary, so a dbt Cloud outage or misconfigured integration doesn't prevent users from doing other transformation management.

ambirdsall:
## What
Adds an error boundary for the dbt cloud integration. As a follow up, I would like to better test out the error response coming from the server; while I did use the actual JSON structure of a known dbt Cloud error response and code the error parsing defensively (so if it's incorrect it should gracefully fall back to a generic error UI instead of erroring) , all my attempts to actually trigger an error response from dbt Cloud caused downstream errors in the cloud server and thus a `500` response with no forwarding of dbt Cloud's `4XX` response payload.

Closes https://github.com/airbytehq/airbyte/issues/20441, but not https://github.com/airbytehq/airbyte-cloud/issues/4181


ambirdsall:
@mfsiega-airbyte, I tried loading the transformation tab for an account using a deleted access token, but got a 500 response from the list dbt jobs endpoint. This error shouldn't propagate to the point that it crashes a server thread, it should rather be a 4XX status, ideally including the dbt Cloud API payload so the frontend can display something useful to the user. Not urgent: in the short-term, I'll just clean up this code in order to make the generic message ready to merge and in the medium term I'm going to be pretty focused on free connectors.

mfsiega-airbyte:
@ambirdsall just wanted to check - do we want to try to merge this? We got another similar report today.

TBernstein4:
@mfsiega-airbyte @ambirdsall Just bumping this to see if we can proceed with the related OC issue?

ambirdsall:
@TBernstein4 @mfsiega-airbyte I don't think there's any reason to hold off on getting this merged: it would be nice to iterate a little further on the code to parse out dbt-sourced error messages, but even if it shows the generic fallback message 100% of the time that's a clear improvement over the app's current behavior. I'll resolve all conflicts and pull this PR out of draft status so we can ship it soon.

mfsiega-airbyte:
Just to check - what did the validation look like here? (Looks reasonable to me, as long as it Works.)

ambirdsall:
> Just to check - what did the validation look like here? (Looks reasonable to me, as long as it Works.)

Steps I :
- generated a new service token with job admin privileges
- registered it to an airbyte workspace
- verified everything worked as expected
- deleted the new token in dbt cloud
- saw the generic error message in the transformation tab

It certainly could look nicer (ignore that black border, it's a bug in my screenshot program), but it seems like the first priority should be unbreaking the UX, followed by a quick visual iteration with Nico's help:
2023-01-24-16:04:59-screenshot

airbyte-webapp/src/pages/connections/ConnectionTransformationPage/DbtCloudTransformationsCard.tsx
```diff
@@ -25,6 +25,50 @@ interface DbtJobListValues {
   jobs: DbtCloudJob[];
 }
 
+const DbtCloudErrorBoundary = class extends React.Component {
```
josephkmh, line 28: This file is getting really full of components - could we split this into a new file to ease readability & file navigation?

airbyte-webapp/src/pages/connections/ConnectionTransformationPage/DbtCloudTransformationsCard.tsx
```diff
@@ -25,6 +25,50 @@ interface DbtJobListValues {
...
+  static getDerivedStateFromError(error: Error) {
+    // TODO I'm pretty sure I did not correctly mock the exact error response format.
+    // eslint-disable-next-line
+    const displayMessage = (error?.message as any)?.status?.user_message;
+    return { error, displayMessage };
+  }
+
+  componentDidCatch(error: Error) {
+    console.log(error);
+  }
```
josephkmh, line 44: How about using our `AppMonitoringService` to track when errors catch here? Right now we would not have any idea about how often this is happening, since the error is just handled in the UI. With `trackError` we can log occurrences in datadog. There are examples of how to do this with an error boundary [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/pages/SpeakeasyRedirectPage/SpeakeasyRedirectPage.tsx) and [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/components/common/ApiErrorBoundary/ApiErrorBoundary.tsx#L83)

airbyte-webapp/src/pages/connections/ConnectionTransformationPage/DbtCloudTransformationsCard.tsx
```diff
@@ -39,7 +83,9 @@ export const DbtCloudTransformationsCard = ({ connection }: { connection: WebBac
   const { hasDbtIntegration, isSaving, saveJobs, dbtCloudJobs } = useDbtIntegration(connection);
 
   return hasDbtIntegration ? (
-    <DbtJobsForm saveJobs={saveJobs} isSaving={isSaving} dbtCloudJobs={dbtCloudJobs} />
+    <DbtCloudErrorBoundary>
+      <DbtJobsForm saveJobs={saveJobs} isSaving={isSaving} dbtCloudJobs={dbtCloudJobs} />
+    </DbtCloudErrorBoundary>
```
josephkmh, line 88: One of the downsides of using a suspense query is that we need an error boundary! Maybe we could use a generic `ComponentErrorBoundary` or something that we can wrap individual components. That's probably better than crashing the whole page and showing `<ErrorOccurredView>`. Anyway, not for this PR, just food for thought 🙂

joeykmh (changes requested):

airbyte-webapp/src/pages/connections/ConnectionTransformationPage/DbtCloudTransformationsCard.tsx
josephkmh: @ambirdsall not sure if you intend to address this comment still?

joeykmh (approved):
Code LGTM - just one of my comments is still open: https://github.com/airbytehq/airbyte/pull/20616#discussion_r1088816381

Not sure if you want to address that before merging? OK for me either way 👍

mfsiega-airbyte (approved):

airbyte-webapp/src/pages/connections/ConnectionTransformationPage/DbtCloudTransformationsCard.tsx
ambirdsall: @josephkmh this is what happens when I make a mental note but not a physical one. Great point, totally slipped my mind 🤦 👍

ambirdsall:
@josephkmh I split out helper components. The main component has a lot less code to wade through, now.

The scss import lint rule meant I had to restructure the CSS, too, so I recorded a screen capture (using a chrome extension to trigger artificial API errors) to show the actual rendered styles in the different contexts (a 500 response to `/get_available_dbt_jobs`, an empty jobs list, a nonempty obs list):
dbt-cloud-error-bound-transformation-card

joeykmh (approved):
Thanks for breaking out those components into separate files! Looks good

PR #20620, opened 2022-12-17
🪟🔧 Connector builder: Performance improvements

flash1293:
Some experimental performance work:

* Replace state+effect with useMemo where possible
* Split up context
* Memoize places that seem fitting (especially around formik as it's very trigger happy)

There is some positive effect - here the react commits for this PR (scenario is adding one character to the base path).
The tooltip is showing the render time of the highest bar to make things more comparable (see cumulative sums down below).

Screenshot 2022-12-17 at 17 01 15

 vs master:
Screenshot 2022-12-17 at 17 01 22

There's one commit less and you can see how in the latter one each commit takes the same amount of time because it's always the whole app re-rendering, while this PR there's more variance because some commits are only updating parts of the app (e.g. once the response from the server arrives only the testing panel gets updated, not everything)

However, the positive effect is not huge, probably because now there is more work done to check the memo boundaries  - around 25% faster (**100ms to 130 ms cumulative render time which 6x CPU slowdown**).

I didn't try, but I have the suspicion that adding more fine-grained memoization (like discussed with a `memoWithBuilderState` higher-order-component) won't necessarily reduce the cumulative render time by much

My summary:
* No huge low hanging performance fruits in there at the moment
* Might become more relevant in the future though because the app will get more complex and always re-rendering everything multiple times will get more and more expensive
* Formik is re-rendering a lot - a render of the base component will definitely render every place that's tapping into the context in some form (something to keep in mind to make things easier later on). I think we can shave off some time here but it would be harder to achieve and make the code more complex.
* No pressure on doing any refactorings at the moment, but we should keep an eye on this - the current structure might also get hard to grok at some point when we keep adding features


flash1293:
@lmossman do you think it's worth making this PR mergeable? The split into two contexts (one for the basic state, one for the testing part) seems to make sense as it's kind of separate concerns which have little overlap in the UI.

airbyte-webapp/src/components/connectorBuilder/useBuilderErrors.ts
```diff
@@ -9,6 +9,7 @@ import { BuilderFormValues } from "./types";
 
 export const useBuilderErrors = () => {
   const { touched, errors, validateForm, setFieldTouched } = useFormikContext<BuilderFormValues>();
+  console.log(touched);
```
lmossman, line 12: todo: remove before merging

airbyte-webapp/src/pages/ConnectorBuilderPage/ConnectorBuilderPage.tsx
```diff
@@ -1,61 +1,93 @@
...
 export const ConnectorBuilderPage: React.FC = () => (
   <ConnectorBuilderStateProvider>
-    <ConnectorBuilderPageInner />
+    <ConnectorBuilderAPIProvider>
+      <ConnectorBuilderPageInner />
+    </ConnectorBuilderAPIProvider>
   </ConnectorBuilderStateProvider>
 );
+
+const Panels = React.memo(
```
lmossman, line 50: The motivation behind moving this into it's own component is so that we can wrap it in a React.memo(), correct?

airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderStateService.tsx
```diff
@@ -148,7 +125,68 @@ export const ConnectorBuilderStateProvider: React.FC<React.PropsWithChildren<unk
   return <ConnectorBuilderStateContext.Provider value={ctx}>{children}</ConnectorBuilderStateContext.Provider>;
 };
 
-export const useConnectorBuilderState = (): Context => {
+export const ConnectorBuilderAPIProvider: React.FC<React.PropsWithChildren<unknown>> = ({ children }) => {
```
lmossman, line 128: I like this split of the contexts, I think it makes sense conceptually  👍 

Only thing I'm not sure of is the naming - maybe we could call them something like `ConnectorBuilderFormState` and `ConnectorBuilderTestState`? I want to make sure this doesn't get confused with https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderApiService.ts

airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderStateService.tsx
```diff
@@ -148,7 +125,68 @@ export const ConnectorBuilderStateProvider: React.FC<React.PropsWithChildren<unk
...
+      setTestStreamIndex(streams.length - 1);
+    }
+  }, [streams, testStreamIndex, setTestStreamIndex]);
+
+  const ctx = {
+    streams,
+    streamListErrorMessage,
+    configJson,
+    configString,
+    setConfigString,
```
lmossman, line 174: nit: I feel like we should rename this to something like `testInputJson`, since I think `config` may start to get too overloaded. Not super related to the other changes here though

lmossman (commented):
@flash1293 yes, I think we should productionize this. It gives some performance benefits as well as adding a nice conceptual split to the state, which I imagine will only be harder to do in the future if we wait to merge this

flash1293:
Good suggestions, let’s merge all of the features first and then come back to this as I did a few changes that will conflict.

airbyte-webapp/src/pages/ConnectorBuilderPage/ConnectorBuilderPage.tsx
flash1293: Exactly, as validation is async in formik this re-renders multiple times which I try to contain as much as possible

flash1293:
Took care of the review comments.

Also, one more change I did because I stumbled over it a few time - turn the editor view state into a local storage state as well because it's super weird it's switching back to the UI after a reload.

airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderStateService.tsx
```diff
@@ -101,22 +116,54 @@ export const ConnectorBuilderStateProvider: React.FC<React.PropsWithChildren<unk
...
+  const [selectedView, setSelectedView] = useState<BuilderView>("global");
+
+  const ctx = {
+    builderFormValues,
+    jsonManifest: derivedJsonManifest,
+    lastValidJsonManifest,
+    yamlManifest,
+    yamlEditorIsMounted,
+    yamlIsValid,
+    testStreamIndex,
```
lmossman, line 137: What is the reasoning behind putting `testStreamIndex` in FormState instead of TestState?

airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderStateService.tsx
```diff
@@ -127,42 +174,33 @@ export const ConnectorBuilderStateProvider: React.FC<React.PropsWithChildren<unk
...
   };
 
-  return <ConnectorBuilderStateContext.Provider value={ctx}>{children}</ConnectorBuilderStateContext.Provider>;
+  return <ConnectorBuilderTestStateContext.Provider value={ctx}>{children}</ConnectorBuilderTestStateContext.Provider>;
+};
+
+export const useConnectorBuilderTestState = (): TestStateContext => {
+  const connectorBuilderState = useContext(ConnectorBuilderTestStateContext);
+  if (!connectorBuilderState) {
+    throw new Error("useConnectorBuilderAPI must be used within a ConnectorBuilderAPIProvider.");
```
lmossman, line 196:
```suggestion
    throw new Error("useConnectorBuilderTestState must be used within a ConnectorBuilderTestStateProvider.");
```

airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderStateService.tsx
```diff
@@ -127,42 +174,33 @@ export const ConnectorBuilderStateProvider: React.FC<React.PropsWithChildren<unk
...
+
+  return connectorBuilderState;
 };
 
-export const useConnectorBuilderState = (): Context => {
-  const connectorBuilderState = useContext(ConnectorBuilderStateContext);
+export const useConnectorBuilderFormState = (): FormStateContext => {
+  const connectorBuilderState = useContext(ConnectorBuilderFormStateContext);
   if (!connectorBuilderState) {
     throw new Error("useConnectorBuilderState must be used within a ConnectorBuilderStateProvider.");
```
lmossman, line 205:
```suggestion
    throw new Error("useConnectorBuilderFormState must be used within a ConnectorBuilderFormStateProvider.");
```

lmossman (changes requested):
I tried testing this locally and I noticed an issue: when you add a stream, that stream is not available in the test panel stream selector dropdown until you switch the stream view manually (either in the sidebar or in the stream selector). 

I spent a few minutes looking into this and I found that the `lastValidBuilderFormValuesRef.current` is properly being updated when a new stream is added, but I think the issue is that this doesn't cause the FormState provider to re-render, so is doesn't get propagated to `lastValidBuilderFormValues` and therefore doesn't get propagated to `lastValidJsonManifest` or to the TestState so a streams/list call is not made. 

Only once the focused stream is manually changed does all of the above happen since the state provider gets re-rendered in that case.

airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderStateService.tsx
flash1293: The test stream index is set in a bunch of places (from the side bar, from the stream config view, from the builder errors hook) - those places do not subscribe to the test state right now. If they would, an update to the test state would re-render most of the UI again.

flash1293: It's a little annoying so many places need to subscribe to the test state, solely because they want to set the test stream index. But actually I found a better way to handle that - instead of explicitly setting testing stream index and view index to the same value, I added a hook to the new test state service to handle that. Makes things much simpler and is a better separation of concerns.

flash1293:
> I tried testing this locally and I noticed an issue: when you add a stream, that stream is not available in the test panel stream selector dropdown until you switch the stream view manually (either in the sidebar or in the stream selector).

Took me a while to figure it out, but it should work fine now. The problem was that it would update the ref _after_ `setStoredBuilderFormValues` is called - however that call directly triggers a new react render, so the ref was outdated. Switching the order made it work.

lmossman (approved):
Changes LGTM, tested locally and it is working as expected

PR #20625, opened 2022-12-18
🪟 🐛 Fix non-breaking schema changes preference issues

edmundito:
## What
Fixes three issues:

* Adds missing i18n string for when sync is disabled
* Ensures that the options rendered are the ones we can support
* Custom styling on the label that is not necessary breaks when the resizing small

### DIFF
--- a/airbyte-webapp/src/locales/en.json
+++ b/airbyte-webapp/src/locales/en.json
@@ -153,6 +153,7 @@
   "connectionForm.nonBreakingChangesPreference.label": "Non-breaking schema updates detected*",
   "connectionForm.nonBreakingChangesPreference.message": "Set how Airbyte handles syncs when it detects a non-breaking schema change in the source",
   "connectionForm.nonBreakingChangesPreference.ignore": "Ignore",
+  "connectionForm.nonBreakingChangesPreference.disable": "Disable connection",
   "connectionForm.schemaChangesBackdrop.message": "Please review the schema updates before making changes to the connection",
 
   "connectionForm.modal.destinationNamespace.title": "Destination namespace",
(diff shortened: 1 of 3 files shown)


airbyte-webapp/src/locales/en.json
```diff
@@ -153,6 +153,7 @@
   "connectionForm.nonBreakingChangesPreference.label": "Non-breaking schema updates detected*",
   "connectionForm.nonBreakingChangesPreference.message": "Set how Airbyte handles syncs when it detects a non-breaking schema change in the source",
   "connectionForm.nonBreakingChangesPreference.ignore": "Ignore",
+  "connectionForm.nonBreakingChangesPreference.disable": "Disable connection",
```
edmundito, line 156: I made this "Disable connection" instead of "Pause sync" as shown in the design because we don't really have a concept of pausing in the UI and this is what will actually happen - @andyjih thoughts?

airbyte-webapp/src/views/Connection/ConnectionForm/components/NonBreakingChangesPreferenceField.tsx
```diff
@@ -11,12 +11,13 @@ import { useConnectionFormService } from "hooks/services/ConnectionForm/Connecti
 
 import styles from "./NonBreakingChangesPreferenceField.module.scss";
 
+const SUPPORTED_PREFERENCES = [NonBreakingChangesPreference.ignore, NonBreakingChangesPreference.disable];
```
edmundito, line 14: We may add more options in the future that may not have an i18n string or may need to wait on the backend, so they need to be supported manually.

airbyte-webapp/src/views/Connection/ConnectionForm/components/NonBreakingChangesPreferenceField.module.scss
```diff
@@ -9,12 +9,6 @@
   gap: variables.$spacing-md;
 }
 
-.connectorLabel {
-  max-width: 328px;
-  margin-right: variables.$spacing-xl;
-  vertical-align: top;
-}
```
edmundito, line 16: Probably copied from the Schedule field. That also needs to be cleaned up but I omitted it here because of the scope of the PR.

teallarson (approved):
Approved pending a thumbs up on wording!

Issue #19991, opened 2022-12-01
Destination Snowflake: Emit STATE message after each commit operation


ryankfu:
The ticket here is for Destination Snowflake to emit STATE message after a commit operation. The action items for this ticket are to:
* update the functionality of `stateManager` to mark the STATE message as `committed`
* validate that a STATE message was received by platform

grishick:
Notes from grooming:
once this is implemented, verify that platform starts subsequent attempts from the last  `STATE` message emitted by the destination. If this is not happening - file an issue for platform and make sure it gets prioritized. The desired effect of implementing this is that incremental sync jobs can resume from the last `STATE` message emitted by the destination.

PR #20631, opened 2022-12-19
🚨 Destination Snowflake: checkpointing flush/commit and emit STATE message 🚨 

### LINKED ISSUE #19955: Destination Snowflake: Bundle `flush` and `commit` operation for destination staging
Currently the `flush` and `commit` operation for connectors using staging operations executes at different cadences. The `commit` operation is a marker of when a destination has successfully written data to the destination's tables

With the current logic for destination, the commit operation _only_ runs at the end of a fully successful sync/replication and a state can only be emitted at the end of this commit phase. Since that occurs once it does not allow for Airbyte to receive more periodic STATE messages

By bundling/grouping these two operations together, it will have a definitive point where information was flushed from the staging area buffer and directly into user's destination tables. This will be followed with work on having the destination send back the received STATE message to the platform to use as a checkpoint

ryankfu:
## What
Closes #19955 and #19991 which bundles flush the buffer and commit operations and emits STATE messages back to the platform through the usage of `DestStateLifecycleManager` 

Also includes a reviewed PR for BigQuery staging https://github.com/airbytehq/airbyte/issues/20823

## How
These changes will introduce the logic of removing `airbyte_tmp` tables from the destination lifecycle. Currently, destination connectors that use staging as an option to load data go through the lifecycle of 
```
buffer -(flush_1)-> staging area
buffer -(flush_2)-> staging area
...
buffer -(flush_n)-> staging area

-----AFTER ALL DATA HAS BEEN READ-----

staging area -(copy into {staging_files})-> destination tmp table
destination tmp table -(insert into)-> destination raw table
```

What this change formally allows is the ability to "checkpoint" or emit `AirbyteStateMessages` more frequently through writing to the destination raw table without going through the destination tmp table so logic will look like 
```
buffer -(flush_1)-> staging area -(copy into {staging_file})-> destination raw table
buffer -(flush_2)-> staging area -(copy into {staging_file})-> destination raw table
...
buffer -(flush_n)-> staging area -(copy into {staging_file})-> destination raw table
```

Notable trade-offs is:
* `airbyte_raw` will be a `hot` table
* more frequent SQL loading operations

Benefits
* frequent updates to the destination tables (data is more readily available)
* checkpointing can exist because there is a known point where data is loaded into the destination's raw table
* estimated 30% performance increase since `tmp` table is no longer utilized

Drawbacks
* cost will increase with more SQL load from staging
* _FULL_REFRESH_ will see partial data if fails part way through, this is being addressed in phase 2 with a table in destination that tracks when a sync starts and ends

### STATE LIFECYCLE MANAGER 
(how STATE is marked between whether it is still in PENDING (received but not processed), FLUSHED (when the buffer has flushed the records), COMMITTED (records have been successfully written to the destination's output/final/target table, all of which are synonymous), and EMITTED (when a STATE message has been sent back to the platform for record-keeping to checkpoint
1. `DestStateLifecycleManager.java`
2. `DefaultDestStateLifecycleManager.java`
3. `DestStreamStateLifecycleManager.java` (PER-STREAM)
4. `DestSingleStateLifecycleManager.java` (GLOBAL/LEGACY)

### Buffering Strategy
How we add AirbyteRecordMessages to the buffer and when a buffer has been filled in two conditions do we flush the buffer and return a value indicating _which_ flush type occurred

1. `BufferingStrategy.java`
2. `SerializedBufferingStrategy.java`
3. `InMemoryRecordBufferingStrategy.java`

### Write Config
The Plain Old Java Object (POJO) that had information related to staging files, sync mode and respective mapping (in the case with BigQuery), which tables to write to, etc

1. `WriteConfig.java` (removed staged files since after each upload we will clear the staging area)

### Staging Operations
Collection of operations (either API in the case with BigQuery or SQL operations with Snowflake) that revolve around table, stage creation and delection

1. `StagingOperations.java`
2. `SnowflakeGcsStagingSqlOperations.java`
3. `SnowflakeInternatlStagingSqlOperations.java`
4. `SnowflakeS3StagingSqlOperations.java`

### Buffered Stream Consumer
Acts as the top-level manager of tracking when a buffer flush has occurred to update the STATE message to be marked as committed and emitted back to the platform. Also, manages the way tearing down a failed sync is handled

1. `BufferedStreamConsumer.java`
2. `StagingConsumerFactory.java`

## 🚨 User Impact 🚨
Are there any breaking changes? What is the end result perceived by the user? If yes, please merge this PR with the 🚨🚨 emoji so changelog authors can further highlight this if needed

🚨 This will mostly affect users that expect complete data for _FULL_REFRESH_ mode, cost / performance is a trade-off that is acceptable since this will prevent redundant work from occurring when a sync fails part-way with the advent of checkpointing 🚨 

- [x] Test for Incremental Mode
- [x] Test for Full Refresh Mode
- [x] Test for GLOBAL/LEGACY state
- [x] Test for PER-STREAM state
- [x] Test for Destination Snowflake
    - [x] Verify STATE messages are emitted through the usage of `LOG_CONNECTOR_MESSAGE` env variable
- [x] Test for Destination BigQuery (relevant since `BufferedStreamConsumer` is utilized as the stream consumer for other destination connectors
    - [ ] Verify STATE messages are stores in `airbyte-db`
- [ ] Long running sync (2 hours, check how many commits, logger)

## Pre-merge Checklist

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>

</details>

## Tests

</details>

</details>

</details>


airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumer.java
```diff
@@ -125,20 +128,46 @@ protected void acceptTracked(final AirbyteMessage message) throws Exception {
...
       }
 
-      // if the buffer flushes, update the states appropriately.
-      // TODO: ryankfu (if true, this is where bundling up logic to also commit to airbyte_raw table)
-      if (bufferingStrategy.addRecord(stream, message)) {
-        markStatesAsFlushedToTmpDestination();
+      final Optional flushType = bufferingStrategy.addRecord(stream, message);
+      // if present means that a flush occurred
+      if (flushType.isPresent()) {
+        // TODO: (ryankfu) separate the logic for handling flush states in a separate
```
grishick, line 146: "in a separate..."  - was there a continuation to the sentence?

ryankfu: This comment was resolved with the changes within BigQuery. It was an internal note for myself that I didn't push to remote but was part of the BigQuery branch

grishick:
Is there a recommended reading order of the changes?

ryankfu:
> Is there a recommended reading order for the changes?

Give me a couple mins, I'm going to update the top-level description with reading order for:
* `DestLifecycleStateManager` aka how STATE messages are managed
* `{BigQuery}StagingConsumerFactory` for how the setup/flushing/tear down of message consumer is done
* `BufferedStreamConsumer` for how we process Airbyte Messages and differentiate between RECORD/STATE and update STATE to be emitted back to the platform through `outputRecordCollector`

grishick (commented):
Will this PR will make Redshift destination write directly to `_raw` table and emit state on every flush?

grishick (commented):
How will Snowflake, Redshift, and Bigquery destinations behave when a sync is taking an initial snapshot of a Postgres table in CDC mode? Will they emit state only at the very end of the sync or will they emit states before they write the entire snapshot to the destination? What if the source is taking snapshots of multiple tables?

ryankfu:
> Will this PR will make Redshift destination write directly to _raw table and emit state on every flush?

Yes, this PR will have Redshift write directly to `airbyte_raw` table and emit state since it calls the same `StagingConsumerFactory` and uses the same `BufferedStreamConsumer`. That said, I have yet to set up a connection that tests so a freeze on that connection publishing _may_ be necessary until full testing can be completed

> How will Snowflake, Redshift, and Bigquery destinations behave when a sync is taking an initial snapshot of a Postgres table in CDC mode? Will they emit state only at the very end of the sync or will they emit states before they write the entire snapshot to the destination? What if the source is taking snapshots of multiple tables?

I'm not familiar with the way the initial snapshot works for Postgres CDC mode, from what I recall the initial snapshot is first marked as `FULL REFRESH | OVERWRITE` and so for that first snapshot it will periodically emit STATE messages while writing parts of the snapshot to destination. This means that we can enter an edge case where the snapshotting is partially successful meaning, not the entire snapshot was written before switching to CDC mode. From the vantage point of the destination connector, we're only given information about `DestinationSyncMode` but that doesn't include information about `CDC` so will need to clarify the downstream impacts

If the source is taking snapshots of multiple tables then the destination will still do the same which is to emit STATE throughout the "snapshotting phase". This was brought up in the [Delayed LSN commit on Postgres Source # Outcome](https://docs.google.com/document/d/1IMEXPpweKln3gEMV1o346LFvwZEpdyJtXvWdpBwy3mw/edit#) although I would need to confirm with @subodh1810 on how CDC mode works fully to ensure this change doesn't have unintended side-effects

Thanks for bringing these two points up though

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumer.java
```diff
@@ -134,12 +138,32 @@ protected void acceptTracked(final AirbyteMessage message) throws Exception {
         return;
       }
 
-      // if the buffer flushes, update the states appropriately.
-      // TODO: ryankfu (if true, this is where bundling up logic to also commit to airbyte_raw table)
-      if (bufferingStrategy.addRecord(stream, message)) {
-        markStatesAsFlushedToTmpDestination();
+      final Optional flushType = bufferingStrategy.addRecord(stream, message);
```
grishick, line 141: may need to change this to `Optional<BufferFlushType>` to appease SpotBugs

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/record_buffer/BufferingStrategy.java
```diff
@@ -20,14 +21,16 @@
...
-   * @param message - {@link AirbyteMessage} to buffer
-   * @return true if this record cause ALL records in the buffer to flush, otherwise false.
+   * @param stream stream associated with record
+   * @param message {@link AirbyteMessage} to be added to the buffer
+   * @return an optional value if a flushed occur with the respective flush type, otherwise an empty
+   *         value means only a record was added
    * @throws Exception throw on failure
    */
-  boolean addRecord(AirbyteStreamNameNamespacePair stream, AirbyteMessage message) throws Exception;
+  Optional addRecord(AirbyteStreamNameNamespacePair stream, AirbyteMessage message) throws Exception;
```
grishick, line 33: Also, may have to change this to `Optional<BufferFlushType>` to appease SpotBugs

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumer.java
```diff
@@ -134,12 +138,32 @@ protected void acceptTracked(final AirbyteMessage message) throws Exception {
...
+      // if present means that a flush occurred
+      if (flushType.isPresent()) {
+        if (flushType.equals(BufferFlushType.FLUSH_ALL)) {
```
grishick, line 144: This should make SpotBugs happy:
```
final Optional<BufferFlushType> flushType = bufferingStrategy.addRecord(stream, message);
...
if (flushType.get().equals(BufferFlushType.FLUSH_ALL)) {
...
} else if (flushType.get().equals(BufferFlushType.FLUSH_SINGLE_STREAM)) {

```

ryankfu:
@grishick Following up on your questions with CDC and snapshot mode, from the vantage point of destination there are no downstream effects. As discussed earlier and with Subodh, in essence, Postgres CDC mode will create a `FULL REFRESH` when no STATE message has been stored. Throughout the snapshotting process, Postgres will _only_ emit a STATE message at the *end* of the snapshot, similar to how normal `FULL REFRESH` occurs. When Postgres is in `CDC` mode *and* a STATE message has been stored then it will switch to the typical CDC mode (which works similar to `INCREMENTAL` mode and periodically emits STATE messages)

This means that upon failure of snapshot-ing there will be no discernable changes from the current workflow (even with destination checkpointing changes). As far as the delayed LSN commit changes, this has already been implemented and will be benefitted immediately once these destination checkpointing changes are published since upon each _retry_ within the same sync (2 retries, 3 total attempts) CDC will commit the LSN value (helping clear out any records that have already been committed to the destination)

So to conclude, there will be no perceived changes with CDC snapshotting with single or multiple tables even after these changes are merged

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumer.java
jbfbell: Actually I think `.get()` [throws an error](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#get--) if the value is not present

My stylistic opinion here would be wrap the conditionals into a helper method `shouldMarkStateAsFlushedToDestination`

This would then become:

```
flushType.filter(shouldMarkStateAsFlushedToDestination).ifPresent(markStateAsFlushedToDestination);
```

jbfbell: Also `shouldMarkStateAsFlushedToDestination` could look like this:

```
private boolean shouldMarkStateAsFlushedToDestination(BufferFlushType flushType) {
  switch(flushType) {
    case BufferFlushType.SINGLE_STREAM:
      return stateManager instanceOf DestStreamLifecycleManager;
    case BufferFlushType.FLUSH_ALL:
      return True;
    default:
      return False;
  }
}
```

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/dest_state_lifecycle_manager/DestStateLifecycleManager.java
```diff
@@ -16,6 +16,9 @@
  * been committed</li>
  * <li>committed - associated records have been committed</li>
  * </ol>
+ *
+ * Since the methods for flushing and committing are not atomic, adding in a method
+ * `markPendingAsCommitted` while deprecating `
```
edgao, line 21: while deprecating what? :O

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/dest_state_lifecycle_manager/DestStateLifecycleManager.java
```diff
@@ -28,6 +31,8 @@ public interface DestStateLifecycleManager {
 
   /**
    * Moves any tracked state messages that are currently pending to flushed.
+   *
+   * @Deprecated
```
edgao, line 35: I generally like to add a message for why this method is deprecated, but totally reasonable to say that you never want to look at this file again

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/dest_state_lifecycle_manager/DestSingleStateLifecycleManager.java
```diff
@@ -58,6 +59,22 @@ public void markFlushedAsCommitted() {
     }
   }
 
+  @Override
+  public void markCommittedAsEmitted() {
+    if (lastCommittedState != null) {
+      lastEmittedState = lastCommittedState;
```
edgao, line 65: where do we actually use `lastEmittedState`? afaict this field is write-only, i.e. we're holding onto this variable but never actually doing anything with it

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/dest_state_lifecycle_manager/DestSingleStateLifecycleManager.java
```diff
@@ -58,6 +59,22 @@ public void markFlushedAsCommitted() {
     }
   }
 
+  @Override
+  public void markCommittedAsEmitted() {
+    if (lastCommittedState != null) {
```
edgao, line 64: This null check seems redundant; if lastCommitted is null, then this is a noop anyway

airbyte-integrations/bases/base-java/src/test/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumerTest.java
```diff
@@ -185,6 +185,7 @@ void test1StreamWithStateAndThenMoreRecordsSmallerThanBuffer() throws Exception
     verify(outputRecordCollector).accept(STATE_MESSAGE1);
   }
 
+  /* TODO (ryankfu) */
```
edgao, line 188: ?

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumer.java
edgao: I'd suggest enabling this intellij inspection, since spotbugs is kind of annoying to work with 
image

it's (correctly) pointing out that this if condition is never true, since you're comparing an enum to an optional

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumer.java
```diff
@@ -134,12 +139,32 @@ protected void acceptTracked(final AirbyteMessage message) throws Exception {
...
-      // TODO: ryankfu (if true, this is where bundling up logic to also commit to airbyte_raw table)
-      if (bufferingStrategy.addRecord(stream, message)) {
-        markStatesAsFlushedToTmpDestination();
+      final Optional<BufferFlushType> flushType = bufferingStrategy.addRecord(stream, message);
+      // if present means that a flush occurred
+      if (flushType.isPresent()) {
+        if (flushType.equals(BufferFlushType.FLUSH_ALL)) {
+          // when all buffers have been flushed then we can update all states as flushed
+          markStatesAsFlushedToDestination();
+        } else if (flushType.equals(BufferFlushType.FLUSH_SINGLE_STREAM)) {
```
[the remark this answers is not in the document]
edgao: strong +1 for inverting these equals checks, it's a good habit to get into even when technically unnecessary.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumer.java
```diff
@@ -134,12 +139,32 @@ protected void acceptTracked(final AirbyteMessage message) throws Exception {
...
+          if (stateManager instanceof DestStreamStateLifecycleManager) {
```
edgao, line 149: mild preference to add some sort of "supports per-stream flush" method to state lifecycle manager, rather than explicity doing an instanceof check

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumer.java
```diff
@@ -148,8 +173,14 @@ protected void acceptTracked(final AirbyteMessage message) throws Exception {
 
   }
 
-  private void markStatesAsFlushedToTmpDestination() {
-    stateManager.markPendingAsFlushed();
+  /**
+   * After marking states as committed, emit the state message to platform then mark state messages as
+   * emitted to avoid resending the same state message to the platform
+   */
+  private void markStatesAsFlushedToDestination() {
```
edgao, line 180: (nonblocking) one of the ideas brought up in the initial spec was that we'd do the flushes in a separate thread. This method is (probably?) not thread-safe, i.e. if there's a thread ingesting new record/state messages at the same that we're flushing stuff, weird things might happen

have you had a chance to sketch out what would need to change to support that sort of concurrency?

ryankfu: I have not sketched out the changes needed to migrate this function to a thread-safe (atomic) operation. For the time being, that is something that would be a future consideration. There has been [discussion](https://airbytehq-team.slack.com/archives/C048E3BK1KM/p1674062757692439) on this topic, however, the goal was to first get a working checkpointing solution to not introduce any more scope creep work

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/dest_state_lifecycle_manager/DestSingleStateLifecycleManager.java
ryankfu: `lastEmittedState` is not used, it was more of design decision initially to create a category that the STATE message(s) have been committed and returned back to the platform. However, while speaking with Joe, it makes more practical sense to simply mark this value as `null` and add a comment how come it marks it as `null`
```java
// This is synonymous with saying the STATE message has been returned to the platform so we clear out the committed state
```

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/dest_state_lifecycle_manager/DestSingleStateLifecycleManager.java
ryankfu: Yes, I agree, the main rationale behind keeping the `null` check was that it follows the same pattern as the rest of `DestSingleStateLifecycleManager`

airbyte-integrations/connectors/destination-snowflake/src/test/java/io/airbyte/integrations/destination/snowflake/SnowflakeDestinationTest.java
```diff
@@ -92,35 +75,6 @@ public void useInsertStrategyTest() {
     assertFalse(SnowflakeDestinationResolver.isS3Copy(stubConfig));
   }
 
-  @Test
-  public void testCleanupStageOnFailure() throws Exception {
```
ryankfu, line 96: This test has been removed because we no longer attempt to `copyIntoTmpTableFromStage` within the `onCloseFunction` when no failure occurs while flushing buffer, this is done within the `flushBufferFunction` where if a failure occurs, the failure is tracked within the `hasFailed` variable and regardless of failure or success the logic will execute `dropStageIfExists` if the user specifies they want to `purgeStagingData`

`dropStageIfExists` will clear out the entire bucket which will supersede cleaning up individual staging files

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/dest_state_lifecycle_manager/DestStateLifecycleManager.java
```diff
@@ -16,6 +16,9 @@
  * been committed</li>
  * <li>committed - associated records have been committed</li>
  * </ol>
+ *
+ * Since the methods for flushing and committing are not atomic, adding in a method
```
grishick, line 20: I don't understand this comment. How are method for flushing and committing not atomic and why that means that instead of `markPendingAsFlushed` is replaced with `markPendingAsCommitted`

ryankfu: Good catch, this was mostly a quick comment that I forgot to update, I'll update this but the main point was that since we're bundling flushing and committing that `markPendingAsFlushed` will be superseded by `markPendingAsCommitted`

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/FailureTrackingAirbyteMessageConsumer.java
```diff
@@ -49,7 +49,10 @@ public void start() throws Exception {
   }
 
   /**
-   * Wraps actual processing of each {@link AirbyteMessage}
+   * Business logic for handling {@link io.airbyte.protocol.models.AirbyteRecordMessage} and
```
jbfbell, line 52: I know this is an abstract method, but is there anything more specific we could say here? `acceptTracked` makes it sound like it might accept some tracking?

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumer.java
```diff
@@ -125,6 +127,7 @@ protected void acceptTracked(final AirbyteMessage message) throws Exception {
       final AirbyteRecordMessage recordMessage = message.getRecord();
       final AirbyteStreamNameNamespacePair stream = AirbyteStreamNameNamespacePair.fromRecordMessage(recordMessage);
 
+      // if stream is not part of list of streams to sync to then throw invalid stream exception
```
jbfbell, line 130: you can remove this comment, the code is very clear and readable

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumer.java
```diff
@@ -134,22 +137,47 @@ protected void acceptTracked(final AirbyteMessage message) throws Exception {
         return;
       }
 
-      // if the buffer flushes, update the states appropriately.
-      // TODO: ryankfu (if true, this is where bundling up logic to also commit to airbyte_raw table)
-      if (bufferingStrategy.addRecord(stream, message)) {
-        markStatesAsFlushedToTmpDestination();
+      final Optional<BufferFlushType> flushType = bufferingStrategy.addRecord(stream, message);
```
jbfbell, line 140: Just stating that I still prefer separating the logic of whether or not to `markStateAsFlushedToDestination` to a helper method and using the Optional oneliner rather than the nested `if` statements

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/BufferedStreamConsumer.java
```diff
@@ -166,21 +203,28 @@ protected void close(final boolean hasFailed) throws Exception {
 
     streamToIgnoredRecordCount
         .forEach((pair, count) -> LOGGER.warn("A total of {} record(s) of data from stream {} were invalid and were ignored.", count, pair));
+    // If the replication has failed then log the issue as failed otherwise wrap up any remaining data
```
jbfbell, line 206: I think you can remove this comment

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/dest_state_lifecycle_manager/DestStreamStateLifecycleManager.java
```diff
@@ -52,6 +52,10 @@ Queue<AirbyteMessage> listPending() {
     return listStatesInOrder(streamToLastPendingState);
   }
 
+  /*
```
jbfbell, line 55: Does this comment imply this method is deprecated?

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/dest_state_lifecycle_manager/DestStreamStateLifecycleManager.java
```diff
@@ -62,16 +66,36 @@ public Queue<AirbyteMessage> listFlushed() {
     return listStatesInOrder(streamToLastFlushedState);
   }
 
+  /*
+   * During the process of migration to destination checkpointing, this method should no longer be in
```
jbfbell, line 70: Does this comment imply this method is deprecated?

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/record_buffer/SerializedBufferingStrategy.java
```diff
@@ -77,18 +89,18 @@ public boolean addRecord(final AirbyteStreamNameNamespacePair stream, final Airb
     }
     final long actualMessageSizeInBytes = streamBuffer.accept(message.getRecord());
     totalBufferSizeInBytes += actualMessageSizeInBytes;
+    // Flushes buffer when either the buffer was completely filled or only a single stream was filled
```
jbfbell, line 92: Nonblocking: comment above conditional could be changed to well named helper method with docs, i.e. 
```
if (shouldFlushAll()) {
...
```

airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/BigQueryStagingOperations.java
```diff
@@ -7,42 +7,77 @@
...
 import io.airbyte.integrations.destination.record_buffer.SerializableBuffer;
-import io.airbyte.protocol.models.v0.DestinationSyncMode;
 import java.util.List;
 
 /**
  * This interface is similar to
  * {@link io.airbyte.integrations.destination.s3.BlobStorageOperations}.
+ *
+ * <p>
+ * Similar interface to {@link io.airbyte.integrations.destination.jdbc.SqlOperations}
```
jbfbell, line 17: could maybe be combined into the comment above if it's similar to both interfaces?

jbfbell: meta question: if the interfaces are similar should they be the same?

airbyte-integrations/connectors/destination-jdbc/src/main/java/io/airbyte/integrations/destination/staging/StagingConsumerFactory.java
```diff
@@ -234,52 +263,19 @@ private OnCloseFunction onCloseFunction(final JdbcDatabase database,
...
-          stagingOperations.createTableIfNotExists(database, schemaName, dstTableName);
-          switch (writeConfig.getSyncMode()) {
-            case OVERWRITE -> queryList.add(stagingOperations.truncateTableQuery(database, schemaName, dstTableName));
-            case APPEND, APPEND_DEDUP -> {}
-            default -> throw new IllegalStateException("Unrecognized sync mode: " + writeConfig.getSyncMode());
-          }
-          queryList.add(stagingOperations.insertTableQuery(database, schemaName, srcTableName, dstTableName));
-        }
+        /*
+         * This section is mostly a no-op since the #flushBufferFunction will flush and commit within
```
jbfbell, line 267: I think you can delete this comment

grishick:
Do you think we need `DefaultDestStateLifecycleManager` at all? IIUC `DefaultDestStateLifecycleManager` delegates everything it does to `internalStateManagerSupplier` which is either `DestSingleStateLifecycleManager` or `DestStreamStateLifecycleManager. Once `DefaultDestStateLifecycleManager` is instantiated `internalStateManagerSupplier` is also instantiated. This makes `DefaultDestStateLifecycleManager` look more like a factory than a default implementation of `DestStateLifecycleManager`. What's more confusing is that `DefaultDestStateLifecycleManager` implements the same interface as `internalStateManagerSupplier` and essentially encapsulates the implementation of the interface. Could we simplify the code by skipping `DefaultDestStateLifecycleManager` altogether? I am not suggesting to do this in this PR, but if that makes sense, then I am suggesting to file a followup issue to simplify this inheritance/delegation pattern.

grishick (approved):
Other than several nitpick comments from myself and Joe, the PR looks good to go!

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/dest_state_lifecycle_manager/DestStreamStateLifecycleManager.java
ryankfu: Yes, this should be marked as `Deprecated`, thanks for the catch. It appears during the rebase I _mistakenly_ deleted the `@Deprecated` tag

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/FailureTrackingAirbyteMessageConsumer.java
ryankfu: Thanks for pointing this out, I've updated the comment to better match that this functionality covers:
* storing of STATE messages
* serializing RECORD messages and storage within a buffer

with a note that this doesn't _always_ apply but is generally true

jbfbell (approved):

PR #20632, opened 2022-12-19
Source Record Heartbeat Monitor

cgardens:
## What
* Adds a heartbeat mechanism that tracks when the last record that came from the source was.
* If it too much time passes in between records, the platform ends the replication.
* Risk: It is set to 5 min right now. If 5 min is too short, some syncs will fail. To mitigate the risk we could make this a warning for now if we wanted.

## How
* It tracks the time between records by adding a "chaperone" thread that run the existing thread that listens for records from the source and a separate heartbeat thread. If the heartbeat thread ever says too long has passed, the chaperone shuts down the heartbeat thread and the replication.
* Note: There maybe a better pattern handling the threads. Would love suggestions. This was the only way I could figure how to do it, that kept the control over the threads easy to follow.

## Recommended reading order
1. `HeartbeatTimeoutChaperone.java`
2. `DefaultReplicationWorker.java`
3. the rest...


airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorkerFactory.java
```diff
@@ -92,6 +94,10 @@ public static DefaultReplicationWorker create(final String jobId,
         getStreamFactory(destProtocolVersion, DefaultAirbyteDestination.CONTAINER_LOG_MDC_BUILDER, serdeProvider, migratorFactory),
         new VersionedAirbyteMessageBufferedWriterFactory(serdeProvider, migratorFactory, destProtocolVersion));
 
+    final var srcHeartbeatMonitor = new HeartbeatMonitor(HeartbeatMonitor.DEFAULT_HEARTBEAT_FRESHNESS_THRESHOLD);
```
cgardens, line 97: In the future, if we have some sources that we know need a longer heartbeat, we could add a field the actor definition with an override for what the threshold should be.

benmoriceau:
Going to try the performances of this approach since I don't manage to run the other approach in dev

cgardens:
@benmoriceau picked up this project in a different PR. closing.

PR #20633, opened 2022-12-19
remove old (unused) usages of heartbeat monitor

cgardens:
## What
* At some point we had a working heartbeat monitor that tracked that the source was regularly sending records to the platform. 
* At some point, it got partially disconnected. In other words, it doesn't work at all, but it still exists in the code.
* This PR removes the old, unused usages.
* Next PR will plug in a new heartbeat monitor

@davinchia a sanity check here would be appreciated.

### DIFF
--- a/airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteSource.java
+++ b/airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteSource.java
@@ -49,6 +49,10 @@ public class DefaultAirbyteSource implements AirbyteSource {
   private Integer exitValue = null;
   private final boolean logConnectorMessages = new EnvVariableFeatureFlags().logConnectorMessages();
 
+  public DefaultAirbyteSource(final IntegrationLauncher integrationLauncher) {
+    this(integrationLauncher, new DefaultAirbyteStreamFactory(CONTAINER_LOG_MDC_BUILDER));
+  }
+
   public DefaultAirbyteSource(final IntegrationLauncher integrationLauncher, final AirbyteStreamFactory streamFactory) {
     this.integrationLauncher = integrationLauncher;
     this.streamFactory = streamFactory;
--- a/airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteSource.java
+++ b/airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteSource.java
@@ -6,7 +6,6 @@ package io.airbyte.workers.internal;
 
 import static io.airbyte.metrics.lib.ApmTraceConstants.WORKER_OPERATION_NAME;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import datadog.trace.api.Trace;
 import io.airbyte.commons.features.EnvVariableFeatureFlags;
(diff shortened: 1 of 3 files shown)


airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteSource.java
```diff
@@ -87,7 +72,6 @@ public void start(final WorkerSourceConfig sourceConfig, final Path jobRoot) thr
     logInitialStateAsJSON(sourceConfig);
 
     messageIterator = streamFactory.create(IOs.newBufferedReader(sourceProcess.getInputStream()))
-        .peek(message -> heartbeatMonitor.beat())
```
cgardens, line 90: we are triggering the heartbeat here, but never checking it anywhere else.

airbyte-commons-worker/src/test/java/io/airbyte/workers/WorkerUtilsTest.java
```diff
@@ -5,125 +5,22 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 @SuppressWarnings("PMD.JUnitTestsShouldIncludeAssert")
 class WorkerUtilsTest {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(GentleCloseWithHeartbeat.class);
-
-  @Nested
-  class GentleCloseWithHeartbeat {
```
cgardens, line 42: i don't think this is testing anything that is still used. but @davinchia  this is where i feel least sure. it looks like in the original orchestrator implementation some of this code got copied over. but i think it's not used.

airbyte-commons-worker/src/test/java/io/airbyte/workers/internal/DefaultAirbyteSourceTest.java
```diff
@@ -165,21 +160,16 @@ void testSuccessfulLifecycle() throws Exception {
...
     when(process.getErrorStream()).thenReturn(new ByteArrayInputStream(("rewq").getBytes(StandardCharsets.UTF_8)));
 
-    when(heartbeatMonitor.isBeating()).thenReturn(true).thenReturn(false);
-
-    final AirbyteSource source = new DefaultAirbyteSource(integrationLauncher, streamFactory,
-        heartbeatMonitor);
+    final AirbyteSource source = new DefaultAirbyteSource(integrationLauncher, streamFactory);
     source.start(SOURCE_CONFIG, jobRoot);
 
-    final List<AirbyteMessage> messages = Lists.newArrayList();
```
cgardens, line 179: this array isn't being used.

davinchia:
looking

airbyte-commons-worker/src/test/java/io/airbyte/workers/WorkerUtilsTest.java
davinchia: your comment looks right to me. 

best way to confirm this is to make sure the acceptance tests pass. I've fixed the PMD error so we can run the full build.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteSource.java
```diff
@@ -57,16 +54,8 @@ public DefaultAirbyteSource(final IntegrationLauncher integrationLauncher) {
   }
```
davinchia, line 54: @cgardens this is needed for the abstract source tests - did not look closely why. mainly wanted to get the build working.

davinchia:
Build is working and I'm confident this change is good.

cgardens:
Thanks, Davin! Can you approve please then?

davinchia (approved):

PR #20634, opened 2022-12-19
Fix failing test

### LINKED ISSUE #19776: Snowflake source encodes integers as floats
(this issue came up in OC and I reproduced it with test data)

I have a table in Snowflake created as follows:

```
create or replace TABLE INTEGRATION_TEST_DESTINATION.GREGS_SCHEMA.MY_TEST_TABLE (
	ID VARCHAR(16777216),
	IMPRESSIONS NUMBER(38,0)
);
```
Attached is the sample data that I inserted into this table

[oc-966.csv](https://github.com/airbytehq/airbyte/files/10079367/oc-966.csv)

when I run a sync from this table to BigQuery, the records in raw table show up as following (in column `_airbyte_data_`):
```
{"ID":"19c9e8f81c9b9d44c17148ab45a4eb25","IMPRESSIONS":1790.0}
```

values like `1790.0` break normalization SQL, because the schema reported by Snowflake source says that `IMPRESSIONS` is an `Integer`.

rodireich:
## What
Fix failing tests in source-oracle and source-db2 due to previous change in #19776


rodireich:
/test connector=connectors/source-oracle
> :clock2: connectors/source-oracle https://github.com/airbytehq/airbyte/actions/runs/3736637551

akashkulk (approved):

airbyte-integrations/connectors/source-oracle/src/test-integration/java/io/airbyte/integrations/source/oracle/OracleSourceDatatypeTest.java
```diff
@@ -146,7 +147,8 @@ protected void initTests() {
             .sourceType("NUMBER")
             .airbyteType(JsonSchemaType.NUMBER)
             .addInsertValues("null", "1", "123.45", "power(10, -130)", "9.99999999999999999999 * power(10, 125)")
-            .addExpectedValues(null, "1", "123.45", String.valueOf(Math.pow(10, -130)), String.valueOf(9.99999999999999999999 * Math.pow(10, 125)))
+            .addExpectedValues(null, "1", "123.45", String.valueOf(Math.pow(10, -130)),
+                "999999999999999999999000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000")
```
ryankfu, line 151: Is this not possible to have the last expected value be 
```java
String.valueOf(9.99999999999999999999 * Math.pow(10, 125)
```

This would be difficult to understand, but if the usage of `Math.pow(10,125)` is not an option can we add a comment to this test case explaining how come this string is the way it is?

rodireich: The `String.valueOf` we used to have was printed as "1.0E126" - the scientific notation.
I couldn't find a way to compel BigDecimal to accept a value using `pow` that is not printed as a `1` with 126 `0`s, so had to resort to direct string. 
I'll add a comment.

airbyte-integrations/connectors/source-oracle/src/test-integration/java/io/airbyte/integrations/source/oracle/OracleSourceDatatypeTest.java
```diff
@@ -14,6 +14,7 @@
 import io.airbyte.integrations.standardtest.source.TestDataHolder;
 import io.airbyte.integrations.standardtest.source.TestDestinationEnv;
 import io.airbyte.protocol.models.JsonSchemaType;
+import java.math.BigDecimal;
```
ryankfu, line 17: Is this being used anywhere? Unless the diff doesn't show the changes that use this import

airbyte-integrations/connectors/source-db2/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/Db2SourceDatatypeTest.java
```diff
@@ -126,7 +127,7 @@ protected void initTests() {
             .airbyteType(JsonSchemaType.NUMBER)
             .fullSourceDataType("DECIMAL(31, 0)")
             .addInsertValues("null", "1", "DECIMAL((-1 + 10E+29), 31, 0)", "DECIMAL((1 - 10E+29), 31, 0)")
-            .addExpectedValues(null, "1", "1.0E30", "-1.0E30")
+            .addExpectedValues(null, "1", "%.0f".formatted(Double.valueOf("1.0E30")), "%.0f".formatted(Double.valueOf("-1.0E30")))
```
ryankfu, line 130: If I'm understanding this correctly, this will be showing integer values _only_, right? I haven't seen the usage of `%.0f` before, normally it would have one decimal precision value.

Also this seems to work but curious if this should be 
```java
Float.valueOf("1.0E30")
```
since the string formatting is indicating a Float `f` vs Double `d`

airbyte-integrations/connectors/source-oracle/src/test-integration/java/io/airbyte/integrations/source/oracle/OracleSourceDatatypeTest.java
ryankfu: I might add in the comment the reason how come it's a plain string representation, like something along the lines of 
```java
"because normalization expects values in integer strings whereas `Math.pow(10, 125)` returns a scientific notation
```

ryankfu (approved):
Overall looks good, minor nit and comments to make sure business context is clear

airbyte-integrations/connectors/source-db2/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/Db2SourceDatatypeTest.java
rodireich: The weeds of floating point string format. 
How I missed thee… :-)

`f` signifies a floating point number - a `d` modifier is a decimal **integer** I think, but it won't accept a double or float value.

the `%.0f` formats a floating point number to print with 0 decimal digits (0 precision) - which is the expected string.

And a `Float.valueOf("1.0E30")` is theoretically ok, but depending on actual hardware and implementation won't print as a clean `1` with zeroes but some weird floating point "1000000015047466200000000000000" and so on- it's the whole single/double precision implementation of respecting float/double types.

airbyte-integrations/connectors/source-oracle/src/test-integration/java/io/airbyte/integrations/source/oracle/OracleSourceDatatypeTest.java
rodireich: Added. thank you

octavia-approvington:

octavia-approvington (approved):

PR #20635, opened 2022-12-19
🎉 Destination-gcs handling v1 data protocol

### LINKED ISSUE #17646: Update destination-gcs to handle the new data types protocol
See https://github.com/airbytehq/airbyte/issues/17637.  This destination _should_ just be using all the same code as destination-s3, so we can maybe get away with doing https://github.com/airbytehq/airbyte/issues/17645 first and then just bump version + publish here.

suhomud:
## What
Destination gcs protocol V1 support

## How
The main change is done in Destination s3 protocol update (https://github.com/airbytehq/airbyte/pull/20088) plus json-avro-converter (https://github.com/airbytehq/json-avro-converter/pull/17)
This PR has next change: 
- integration tests are using V1 dataset
- minor changes for gcs test's comparator
    - base64 decoding for test comparison. 
    - time comparison.

## Recommended reading order
1. `GcsAvroTestDataComparator.java`
2. `GcsAvroParquetDestinationAcceptanceTest.java` - V1 protocol dataset `$ref` type
3.  `GcsCsvDestinationAcceptanceTest.java`  - V1 protocol dataset plus `$ref` type
4.  for others - V1 protocol dataset 

## 🚨 User Impact 🚨
No impact


edgao (approved):
glad to see this didn't require any changes to runtime code - code reuse :100:

airbyte-integrations/connectors/destination-gcs/src/test-integration/java/io/airbyte/integrations/destination/gcs/GcsAvroParquetDestinationAcceptanceTest.java
```diff
@@ -85,7 +91,7 @@ private JsonNode getJsonNode(final AirbyteStream stream, final String name) {
   }
 
   private Set<Type> getExpectedSchemaType(final JsonNode fieldDefinition) {
-    final JsonNode typeProperty = fieldDefinition.get("type");
+    final JsonNode typeProperty = fieldDefinition.get("type") == null ? fieldDefinition.get("$ref") : fieldDefinition.get("type");
```
ryankfu, line 94: Nice work, suggestion to maybe considering a comment here that notes that `$ref` is utilized as a result of the version migration and that `type` is the original standard since tests are a great way to maintain knowledge of business context

airbyte-integrations/connectors/destination-gcs/src/test-integration/java/io/airbyte/integrations/destination/gcs/GcsAvroTestDataComparator.java
```diff
@@ -33,4 +36,23 @@ protected boolean compareDateTimeValues(String airbyteMessageValue, String desti
     return super.compareDateTimeValues(airbyteMessageValue, format.format(dateTime));
   }
 
+  @Override
+  protected boolean compareTime(final String airbyteMessageValue, final String destinationValue) {
+    var destinationDate = LocalTime.ofInstant(getInstantFromEpoch(destinationValue), ZoneOffset.UTC);
```
ryankfu, line 41: Minor nit: instead of `var` have this be `LocalTime` so it's clear that the difference between this `super.compareTime` and this child method is that this is doing a `LocalTime` comparison vs `String` comparison

ryankfu (approved):
LGTM, minor comments but nothing blocking, great usage of code reuse

grishick (approved):

airbyte-integrations/connectors/destination-gcs/src/test-integration/java/io/airbyte/integrations/destination/gcs/GcsAvroParquetDestinationAcceptanceTest.java
suhomud: The `type` is still valid since we have object and arrays (for example: `"type": ["object"]`) in schemas but it will be worth to mentioned the `$ref` in the comments since it's not obvious. Thanks for suggesting!

suhomud:
The fixed tests for bigquery-denormalized https://github.com/airbytehq/airbyte/pull/20640
Will re-run after code freeze

PR #20637, opened 2022-12-19
🪟🐛 Connector form: Render input as number for type number

flash1293:
## What

When `type: number` is used in the connector form, the value is cast correctly to a number, but the input it still rendered as type text which makes it possible for the user to type string values as well.

This PR fixes this problem by switching the input type to number for this property type.


lmossman (approved):
LGTM, tested locally and works as expected

PR #20639, opened 2022-12-19
Source notion: fix schema

### LINKED ISSUE #19777: Stream `pages` in Notion source has incorrect schema for field `properties`
Stream `pages` has a field `properties` which is an array of objects. The objects are defined as `oneOf` and describes several types of objects that can appear inside `properties` array. One of these object variations can have 3 fields: `id`, `type`, and `title` defined as following:
```
{
									"type": "object",
									"properties": {
										"id": {
											"type": "string"
										},
										"type": {
											"enum": ["title"]
										},
										"title": {
											"properties": {
												"text": {
													"type": ["null", "array"],
													"items": {
														"type": ["null", "object"],
														"properties": {
															"href": {
																"type": ["null", "string"]
															},
															"text": {
																"type": ["null", "object"],
																"properties": {
																	"link": {
																		"type": ["null", "object"],
																		"properties": {
																			"url": {
																				"type": ["null", "string"]
																			},
																			"type": {
																				"enum": ["url"]
																			}
																		},
																		"additionalProperties": true
																	},
																	"content": {
																		"type": ["null", "string"]
																	}
																}
															},
															"type": {
																"type": ["null", "string"]
															},
															"plain_text": {
																"type": ["null", "string"]
															},
															"annotations": {
																"type": ["null", "object"],
																"properties": {
																	"bold": {
																		"type": ["null", "boolean"]
																	},
																	"code": {
																		"type": ["null", "boolean"]
																	},
																	"color": {
																		"type": ["null", "string"]
																	},
																	"italic": {
																		"type": ["null", "boolean"]
																	},
																	"underline": {
																		"type": ["null", "boolean"]
																	},
																	"strikethrough": {
																		"type": ["null", "boolean"]
																	}
																},
																"additionalProperties": true
															}
														},
														"additionalProperties": true
													}
												},
												"type": {
													"type": ["null", "string"]
												}
											}
										}
									}
								}
```

the bug in this schema definition is that the actual data looks like this, where `title` is an array of objects:

```
"properties": [{
			"name": "title",
			"value": {
				"id": "title",
				"type": "title",
				"title": [{
					"type": "text",
					"text": {
						"content": "blahblah",
						"link": null
					},
					"annotations": {
						"bold": false,
						"italic": false,
						"strikethrough": false,
						"underline": false,
						"code": false,
						"color": "default"
					},
					"plain_text": "blahblah",
					"href": null
				}]
			}
		}],
```

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1047
https://github.com/airbytehq/airbyte/issues/19777

## How
 - Fix stream schema
 - Set test strictness level to `high`


airbyte-integrations/connectors/source-notion/acceptance-test-config.yml
```diff
@@ -1,28 +1,36 @@
...
   full_refresh:
-    - config_path: "secrets/config.json"
-      configured_catalog_path: "integration_tests/configured_catalog.json"
-      ignored_fields:
-        "pages": ["icon"]
+    tests:
+      - config_path: "secrets/config.json"
+        configured_catalog_path: "integration_tests/configured_catalog.json"
+        ignored_fields:
+          "pages": ["icon"]
```
evantahler, line 36: Can you add a comment to why this field is skipped?

airbyte-integrations/connectors/source-notion/acceptance-test-config.yml
```diff
@@ -1,28 +1,36 @@
 # See [Source Acceptance Tests](https://docs.airbyte.com/connector-development/testing-connectors/source-acceptance-tests-reference)
 # for more information about how to configure these tests
 connector_image: airbyte/source-notion:dev
-tests:
+test_strictness_level: high
```
evantahler, line 4: 🥳

docs/integrations/sources/notion.md
```diff
@@ -83,7 +83,8 @@ The connector is restricted by Notion [request limits](https://developers.notion
 ## Changelog
 
 | Version | Date       | Pull Request                                             | Subject                                                         |
-| :------ | :--------- | :------------------------------------------------------- | :-------------------------------------------------------------- |
+|:--------|:-----------|:---------------------------------------------------------|:----------------------------------------------------------------|
+| 1.0.0   | 2022-12-19 | [20639](https://github.com/airbytehq/airbyte/pull/20639) | Fix `Pages` stream schema                                       |
```
evantahler, line 87: You noted that this was a breaking change (1.0) - what should users expect to see that's changed from previous versions in their records?

evantahler (changes requested):
The changes look good, and I've got high-confidence with `test_strictness_level: high`.  Please add some more information into the change log as to what the breaking change is for users of the connector.

airbyte-integrations/connectors/source-notion/acceptance-test-config.yml
davydov-d: I don't know why it was here - removed it and the test passes successfully

docs/integrations/sources/notion.md
alafanechere: @davydov-d I think your intuition is correct. Even if SAT did not detect the backward incompatibility, this is a breaking change on the `page` stream. Could you please sync with @girarda and @jnr0790 to assess the impact of this breaking change and check if customer outreach is needed?
I'm under the impression you're fixing a nonworking stream, but it'd be worth double-checking that the stream is failing for all the users. It would help to spot the differences if you could share the output of `discover` in `0.1.10` vs `1.0.0`.

davydov-d: Sure! I'm attaching the output for the `Pages` stream as files as it's pretty big

[0.1.10.txt](https://github.com/airbytehq/airbyte/files/10268166/0.1.10.txt)
[1.0.0.txt](https://github.com/airbytehq/airbyte/files/10268167/1.0.0.txt)

alafanechere: The change you introduced is removing and adding properties.
The backward compatibility checks pass because:
- Removal of properties is considered backward compatible because downstream normalization is not destructive:  the previously created column for this property won't be deleted
- The addition of properties will not break an existing catalog but will trigger schema validation errors that are not failing syncs.

@evantahler @sherifnada  wdyt? Shall we make the backward compatibility test fail in this situation? 

I think that the change you introduced are not functionally breaking changes but will lead to very messy destination schema after normalization. 
I'd suggest you locally try a sync with `0.1.10` and then upgrade to 1.0.0:
- check if the sync succeeds
- check how the destination schema looks like

evantahler: I would think that any property removal should be a breaking change.  Regardless of if normalization removes the column or not, new records will be missing data that was previously a part of the record.

alafanechere: I tracked this evolution in a issue: https://github.com/airbytehq/airbyte/issues/20754

airbyte-integrations/connectors/source-notion/integration_tests/expected_records.txt
```diff
@@ -1,37 +1,23 @@
...
+{"stream": "users", "data": {"object": "user", "id": "f5ac1fcb-a06b-4dcc-80e5-403c40dfb38a", "name": "Airyte", "avatar_url": null, "type": "person", "person": {"email": "integration-test@airbyte.io"}}, "emitted_at": 1671456048322}
+{"stream": "users", "data": {"object": "user", "id": "6f1be537-9a62-4686-bb6a-02918894529f", "name": "karina.kuznietsova@zazmic.com", "avatar_url": null, "type": "person", "person": {"email": "karina.kuznietsova@zazmic.com"}}, "emitted_at": 1671456048322}
+{"stream": "users", "data": {"object": "user", "id": "b14821de-846b-4719-8a94-2453eeafa609", "name": "Iryna Grankova", "avatar_url": null, "type": "person", "person": {"email": "iryna.grankova@airbyte.io"}}, "emitted_at": 1671456048322}
+{"stream": "users", "data": {"object": "user", "id": "1dec424e-8fc2-4265-b82f-731a90dffcd0", "name": "Airbyte", "avatar_url": "https://s3-us-west-2.amazonaws.com/public.notion-static.com/daf02971-a282-49e6-810d-860fcb98ad1d/Screenshot_2021-10-19_at_16.42.29.png", "type": "bot", "bot": {"owner": {"type": "workspace", "workspace": true}, "workspace_name": "Airbyte Testing"}}, "emitted_at": 1671456048322}
```
evantahler, line 4: There's PII here - is Karina ok with this?

evantahler (approved):

docs/integrations/sources/notion.md
davydov-d: just to clarify - any action item here to be done? does it still make sense to run a sync locally to look at the destination schema regarding we have decided to treat it as an incompatible change or may I resolve this thread?

airbyte-integrations/connectors/source-notion/integration_tests/expected_records.txt
lazebnyi: Hi @evantahler
Karina, is this someone from Airbyte?

evantahler: zazmic was a previous company we've worked with - it is OK

docs/integrations/sources/notion.md
jnr0790: This will be affecting all cloud users with a Notion connection correct?

davydov-d: @jnr0790 this will affect those users with `Pages` stream enabled. I will post a list of them in the issue

jnr0790: @davydov-d can you tag me and Erica on that comment cc @erica-airbyte

lazebnyi (approved):

ryankfu:
@davydov-d Are there still any blockers for this PR? We recently saw this issue pop up within our on-call and are curious if you can give an update on whether the code freeze is still holding this back

cc: @etsybaev

davydov-d:
@ryankfu sorry for the long delay - i have been on a sick leave. I'll sync with the TCS and deploy this change asap

erica-airbyte:
@davydov-d this is not been merged yet correct? I think we are waiting on a list of customers that this will affect that we need to reach out to before the change go to production.

davydov-d:
@erica-airbyte that's right, it is not merged yet. The list is [here](https://github.com/airbytehq/oncall/issues/1047#issuecomment-1363787693) as Juan asked (I will take a look if something changed for the last three weeks and let you know in the issue)

PR #20645, opened 2022-12-19
🪟🎉 Connector builder authentication

### LINKED ISSUE #19105: [Connector Builder UI] Configurable Authentication
Add UI field to the global configuration view to allow configuration of authentication.

Acceptance Criteria:
- An input exists in the global view allowing users to configure the authentication method
- The input's fields are auto-generated from the json schema of the Authenticator component
- The values entered into this element is propagated to the underlying JSON connector manifest source of truth

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/19105

Based on https://github.com/airbytehq/airbyte/pull/20431

This PR adds API key, bearer and basic auth to the connector builder UI

Screenshot 2022-12-19 at 17 12 33

## How

The following things have been added:

* Store a single authenticator configuration in the root of the form builder form values
* Have a UI to choose the authenticator and configure fields
* For fields which should really really be user inputs and not be hardcoded in the connector (api key, username, password), do not offer a way to fill them as part of the manifest but instead hardcode them to read a user input value via interpolated string. To make the connection obvious, the fields which are hardcoded to use config values are shown with a little user icon in the authenticator form
* This inferred user input is added to the list of user inputs and can't be removed. Some things like title and description can be edited (called inferred user input overrides), but not the field id or the type.
* Inferred user inputs and regular user inputs are merged together into the spec so the user can edit them easily in the input form

## Split out

The following things are split out into separate PRs

* Validate form and show indicator next to input button
* Add oauth and session authentication


airbyte-webapp/src/components/connectorBuilder/Builder/InputsView.tsx
```diff
@@ -95,14 +106,31 @@ export const InputsView: React.FC = () => {
...
+                  size="sm"
+                  variant="secondary"
+                  aria-label="Edit"
+                  onClick={() => {
+                    setInputInEditing(formInputToInputInEditing(input, true));
+                  }}
+                >
+                  <FontAwesomeIcon className={styles.icon} icon={faGear} />
+                </Button>
+              </li>
```
lmossman, line 132: This is almost the same code as the code below it (besides the `true/false` passed to formInputToInputInEditing) - can we extract it into its own component to reduce code duplication?

airbyte-webapp/src/components/connectorBuilder/Builder/InputsView.tsx
```diff
@@ -222,21 +255,25 @@ const InputModal = ({
...
+                      type="array"
+                      optional
+                      label={formatMessage({ id: "connectorBuilder.inputModal.enum" })}
+                      tooltip={formatMessage({ id: "connectorBuilder.inputModal.enumTooltip" })}
+                    />
+                  )}
+                </>
               )}
               <BuilderField
                 path="definition.airbyte_secret"
```
lmossman, line 279: Should this be modifiable for inferred input overrides? It seems somewhat unsafe to let users make one of these fields non-secret, right?

Similar question for the default value - if we let users enter a default value, will that open them up to potentially saving credentials into their connector manifest?

I suppose the argument for doing these things this way are that they let users have more flexibility in these fields, in case they have a use case like always using the same username in their connector. So as long as we default to the most secure options (which we are doing), I think I'm okay with this.

airbyte-webapp/src/components/connectorBuilder/types.ts
```diff
@@ -31,11 +34,95 @@ export interface BuilderStream {
...
     connectorName: yup.string().required("form.empty.error"),
     urlBase: yup.string().required("form.empty.error"),
+    authenticator: yup.object({
+      header: yup.mixed().when("type", {
+        is: "ApiKeyAuthenticator",
+        then: yup.string().required("form.empty.error"),
+        otherwise: (schema) => schema.strip(),
+      }),
+      api_token: yup.mixed().when("type", {
+        is: "ApiKeyAuthenticator",
```
lmossman, line 111: I think this should be:
```suggestion
        is: (val: string) => ["ApiKeyAuthenticator", "BearerAuthenticator"].includes(val),
```

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderOneOf.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export const BuilderOneOf: React.FC<BuilderOneOfProps> = ({ options, path, label, tooltip }) => {
+  const [, , oneOfPathHelpers] = useField(path);
+  const typePath = `${path}.type`;
+  const [typePathField, , typePathHelpers] = useField(typePath);
+  const value = typePathField.value ?? options[0].typeValue;
+
+  useEffect(() => {
+    typePathHelpers.setValue(value);
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
lmossman, line 36: I added this at one point but later removed it from BuilderFields since I didn't think it was actually needed anymore. Could you double check if this is necessary to keep?

lmossman (approved):
Had a few small comments but nothing truly blocking.

These changes look good to me and they work well from my local testing. Nice!

lmossman:
Actually I did discover one issue with this branch: if you don't fill out all of the required fields in the chosen oneOf option, the inferred user input will not be displayed in the test inputs menu. 

To reproduce:
1. Change Authentication to No Auth, then refresh the page
2. Change Authentication to API Key. Notice that `User inputs` on the left now shows a (2)
3. Open the test input menu in the testing panel. See that the API Key input is not rendered. 

I think this is because there is technically an error present for the Header field when the API Key option is chosen, which prevents the underlying builder form from being updated (?)

Expected behavior here is either that I see the API Key input in the testing panel when opened, or the test panel cannot be opened while there are any form errors (though this latter option feels slightly strange)

airbyte-webapp/src/components/connectorBuilder/Builder/GlobalConfigView.tsx
```diff
@@ -16,6 +18,70 @@ export const GlobalConfigView: React.FC = () => {
       <BuilderCard className={styles.content}>
         <BuilderField type="string" path="global.urlBase" label="API URL" tooltip="Base URL of the source API" />
       </BuilderCard>
+      <BuilderCard>
+        <BuilderOneOf
+          path="global.authenticator"
+          label="Authentication"
+          tooltip="Authentication method to use for requests send to the API"
```
lmossman, line 25:
```suggestion
          tooltip="Authentication method to use for requests sent to the API"
```

airbyte-webapp/src/components/connectorBuilder/Builder/InputsView.tsx
flash1293: Good call, feels cleaner anyway.

airbyte-webapp/src/components/connectorBuilder/Builder/InputsView.tsx
flash1293: I went through the same thought process - however as you aren't 100% sure either let's go with minimal options for now. We can always allow more things later, but it's hard to take things away.

This is how it looks now:
Screenshot 2022-12-21 at 12 01 05

airbyte-webapp/src/components/connectorBuilder/types.ts
flash1293: You are correct that this is wrong, but as we moved these into user inputs we don't need to validate them here at all as the user can't set them. Removed along with the username/password stuff.

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderOneOf.tsx
flash1293: Good catch, we don't and I think it's better to truthfully show no selected option instead of defaulting to the first one as that wouldn't be reflected in the actual form state.

flash1293:
> Actually I did discover one issue with this branch: if you don't fill out all of the required fields in the chosen oneOf option, the inferred user input will not be displayed in the test inputs menu.

This happens because I changed the logic to only push updates from formik to the builderFormValues if the current state is valid. I had to do this because otherwise the `/list` request would fail in some situations if the user needs to make choices before they can send the form (e.g. the header name for API Key authentication). I didn't anticipate this side effect, I agree it's a little weird.

> Expected behavior here is either that I see the API Key input in the testing panel when opened, or the test panel cannot be opened while there are any form errors (though this latter option feels slightly strange)

Agree, I built this by maintaining a ref with the last valid manifest derived from the builder form values so everything but the stream list updates even with an invalid form.

lmossman (approved):
Last round of changes look good to me and work well in my local testing 👍

PR #20670, opened 2022-12-19
Use ConnectionApi to fetch source ID

alovew:
Remove dependence on configRepository for fetching source id


jdpgrailsdev (approved):
:shipit: Follow on PR's will remove the database objects.

Issue #20581, opened 2022-12-16
Connector builder: Dropdown with available inputs next to each text field


flash1293:
## Context

Split out of https://github.com/airbytehq/airbyte/issues/20414

To make it easier to use user inputs in various components of the builder, a dropdown with available inputs should be shown next to each applicable form element. Selecting an entry form the list fills in the placeholder with correct syntax 

## Acceptance criteria

* Ability to select usable inputs next to each form element
* Selecting an input will append the placeholder to the value of the form element

flash1293:
Pending for review (just finalized)

PR #20698, opened 2022-12-20
[Connector Builder] Add paginator

lmossman:
## What
Add the paginator to the connector builder, as well as the primaryKey and cursorField inputs to the top-level stream view

## How
Adds a paginator to the connector builder, with a toggleable builder card. Screenshot:
Screen Shot 2022-12-20 at 9 53 10 PM

## Recommended reading order
1. BuilderCard.tsx - adds ability to add a toggle configuration to the card
2. PaginationSection.tsx - paginator component (extracted to its own class since it is large)
3. types.ts - adds the types for primaryKey, cursorField, and paginator


airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderStateService.tsx
```diff
@@ -65,7 +55,7 @@ export const ConnectorBuilderStateProvider: React.FC<React.PropsWithChildren<unk
     DEFAULT_BUILDER_FORM_VALUES
   );
   const builderFormValues = useMemo(() => {
-    return { ...DEFAULT_BUILDER_FORM_VALUES, ...(storedBuilderFormValues ?? {}) };
+    return merge(DEFAULT_BUILDER_FORM_VALUES, storedBuilderFormValues);
```
lmossman, line 58: I added this change because I realized that using the spread operator causes the top-level properties to be completely overwritten. This was a problem for me because my locally stored values did not have any `global.authenticator` set, so I loaded up this branch, the entire locally-stored `global` value overrwrote the default `global` value, meaning that `global.authenticator` was no longer set and some code that was trying to access that failed, so the webapp crashed.

I therefore switched to using lodash `merge` here, so that sub-properties in the default values would still remain if they didnt exist in the locally store values. However, when I added this it caused this entire state provider to seemingly stop re-rendering as a result of the builderFormValues changing, even though there is a useEffect below that consumes it.

This meant that whenever I changed a value in the forms in the UI, that change was not being reflected in the side panel, because the `jsonManifest` values wasn't actually changing anymore. 

@flash1293 do you have any idea why switching to lodash `merge()` here would cause that behavior? I tried logging out the `builderFormValues` and I confirmed that they were indeed changing as expected, but for some reason the useEffect hook that is dependent on it stopped firing.

flash1293: React is doing change detection purely based on the reference of the object you pass in. So if you mutate an object, its reference will stay the same and React will not run effect hooks that depend on it or re-render memoized components if its passed in as a prop.

As you can see here https://lodash.com/docs/4.17.15#merge , lodash merge is mutating the object you pass in as the first argument, merging in all the props from the tail of arguments. This is very much not what you want as it will mutate `DEFAULT_BUILDER_FORM_VALUES`.

You can solve this by creating a new object to merge everything into as part of the useMemo hook:
```
merge({}, DEFAULT_BUILDER_FORM_VALUES, storedBuilderFormValues);
```

As `{}` will create a new object, `builderFormValues` will change and effect hooks and memoized components will re-evaluate.

airbyte-webapp/src/components/connectorBuilder/types.ts
```diff
@@ -284,6 +367,21 @@ export const convertToManifest = (values: BuilderFormValues): PatchedConnectorMa
...
         },
+        paginator: stream.paginator
+          ? {
+              type: "DefaultPaginator",
+              page_token_option: {
+                ...stream.paginator.pageTokenOption,
+                // ensures that empty field_name is not set, as connector builder server cannot accept a field_name if inject_into is set to 'path'
+                field_name: stream.paginator.pageTokenOption.field_name
+                  ? stream.paginator.pageTokenOption.field_name
+                  : undefined,
```
lmossman, line 378: This could also be solved by modifying the BuilderField to not set this value on the formik values when it is an empty string, but this was a more straightforward change. Currently, when it is an empty string, it is still setting it on the formik forms. Thoughts?

airbyte-webapp/src/components/connectorBuilder/types.ts
```diff
@@ -231,12 +276,47 @@ export const builderFormValidationSchema = yup.object().shape({
...
+                otherwise: (schema) => schema.strip(),
+              }),
+              start_from_page: yup.mixed().when("type", {
+                is: "PageIncrement",
+                then: yup.string(),
+                otherwise: (schema) => schema.strip(),
+              }),
+            })
+            .notRequired()
+            .default(undefined),
```
lmossman, line 316: Without this, the UI was failing when toggling on the paginator component I think because the values are not all initialized

airbyte-webapp/src/components/connectorBuilder/Builder/PaginationSection.tsx
```diff
@@ -0,0 +1,180 @@
...
+        options={injectIntoValues}
+        label="Inject into"
+        tooltip="Configures where the page token should be set on the HTTP requests"
+      />
+      <BuilderField
+        type="string"
+        path={streamFieldPath("paginator.pageTokenOption.field_name")}
+        label="Field name"
+        tooltip="Configures which key should be used in the location that the page token is being injected into"
+        optional
```
flash1293, line 58: Is this actually optional? The list endpoint is giving me an error:
Screenshot 2022-12-21 at 14 34 38

airbyte-webapp/src/components/connectorBuilder/types.ts
flash1293: I think it's fine to have this lightweight translation logic in here.

airbyte-webapp/src/components/connectorBuilder/Builder/PaginationSection.tsx
```diff
@@ -0,0 +1,180 @@
...
+      }
+    >
+      <BuilderField
+        type="enum"
+        path={streamFieldPath("paginator.pageTokenOption.inject_into")}
+        options={injectIntoValues}
+        label="Inject into"
+        tooltip="Configures where the page token should be set on the HTTP requests"
+      />
+      <BuilderField
```
flash1293, line 53: This seems to be too simplistic - it depends on the inject_into whether or not you are allowed to provide a field name:
Screenshot 2022-12-21 at 14 34 38
Screenshot 2022-12-21 at 14 43 12

airbyte-webapp/src/components/connectorBuilder/Builder/PaginationSection.tsx
```diff
@@ -0,0 +1,180 @@
...
+        type="string"
+        path={streamFieldPath("paginator.pageTokenOption.field_name")}
+        label="Field name"
+        tooltip="Configures which key should be used in the location that the page token is being injected into"
+        optional
+      />
+    </GroupControls>
+  );
+
+  const pageSizeOption = (
```
flash1293, line 63: A similar situation here - you need to set the page size if yopu want to use the page size option:
Screenshot 2022-12-21 at 14 44 47

airbyte-webapp/src/components/connectorBuilder/Builder/PaginationSection.tsx
```diff
@@ -0,0 +1,180 @@
...
+              <>
+                <BuilderField
+                  type="number"
+                  path={streamFieldPath("paginator.strategy.page_size")}
+                  label="Page size"
+                  tooltip="Size of pages to request from API"
+                />
+                <BuilderField
+                  type="number"
+                  path={streamFieldPath("paginator.strategy.start_from_page")}
```
flash1293, line 136: Side note: I just realized that for these it's very common to interpolate a value from the state, but it's not possible right now as this is a number input. 

Seems like https://github.com/airbytehq/airbyte/issues/20581 is not just a UX improvement but necessary for adding certain functionality in cases like this.

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
```diff
@@ -81,13 +82,26 @@ export const StreamConfigView: React.FC<StreamConfigViewProps> = ({ streamNum })
           label="HTTP Method"
           tooltip="HTTP method to use for requests sent to the API"
         />
+        <BuilderField
```
flash1293, line 85: Seems like these are not required, should they be marked as optional in the UI?

airbyte-webapp/src/components/connectorBuilder/types.ts
flash1293: yup is getting more annoying by the day as well.... I think it's fine for now, but I would really like to just leverage the json schema of the manifest itself.

flash1293:
Biggest takeaway is that there are some constraints to what should be configured which are not expressed in the types itself. We can also split this kind of validation out and merge as is, but it would be nice to capture as much as possible in the UI because it clearly marks where something needs to be changed instead of requiring the user to parse a textual error message and map it back to the UI.

flash1293:
> here is currently an issue where if you set the Field name on the stream slicer to something without changing the value of Inject into, it gives me the error
> I might just be something that we need to add a useEffect with an empty dependency array to set it on the values on creation like in the oneOf.

As it's a very similar case you could also use the `onChange` I introduced to the BuilderField for the inputs form and set inject_to this way (an effect hook will only run after the next render, so it will cause the error, then fix it a split second later). But in general I like the idea with the checkbox better for these request options (this way we can bring them into a valid default state.

airbyte-webapp/src/components/connectorBuilder/Builder/PaginationSection.tsx
```diff
@@ -0,0 +1,180 @@
...
+                  path={streamFieldPath("paginator.strategy.cursor_value")}
+                  label="Cursor value"
+                  tooltip="Value of the cursor to send in requests to the API"
+                />
+                <BuilderField
+                  type="number"
+                  path={streamFieldPath("paginator.strategy.page_size")}
+                  label="Page size"
+                  tooltip="Size of pages to request from API"
+                  optional
```
flash1293, line 162: It seems like we have a problem with optional number fields - clearing them out will turn them into an empty string which will fail the validation.

airbyte-webapp/src/components/connectorBuilder/Builder/PaginationSection.tsx
flash1293: I think putting the request options behind a checkbox like you suggested and looking for those to make page size optional or required would work.

airbyte-webapp/src/components/connectorBuilder/types.ts
lmossman: 100% agree. It was a pretty unfun and tedious process to go through and recreate this manually, and it is not very readable

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
lmossman: Yes they should, good catch

airbyte-webapp/src/components/connectorBuilder/Builder/PaginationSection.tsx
```diff
@@ -0,0 +1,164 @@
...
+                  path={streamFieldPath("paginator.strategy.page_size")}
+                  onChange={(newValue) => {
+                    if (newValue === undefined || newValue === "") {
+                      pageSizeOptionHelpers.setValue(undefined);
+                    }
+                  }}
+                  label="Page size"
+                  tooltip="Set the size of each page"
+                  optional
+                />
```
lmossman, line 154: I decided to keep the page size field here outside of the pageSizeOption group, because it is technically allowed by the low code framework to set a page size here, but not set a page_size_option

flash1293: It's slightly weird that starting to type into the page size input suddenly pops in the page size request options, but I think we can start like this.

flash1293 (approved):
Tested and didn't notice any invariants that aren't covered in an elegant way. LGTM

PR #20699, opened 2022-12-20
:window: :art: Move `Add` button into the line of Connector Builder key value list fields

lmossman:
## What
I felt that the request options components of the connector builder were taking up too much space when no entries had been added:
Screen Shot 2022-12-19 at 6 35 11 PM

## How
In order to address this, I wanted to try out moving the `Add` button into the top line of this component. To accomplish this, I modified the GroupControls component to take in a generic "control" rather than specifically a "dropdown", and moved the add button into there, achieving this look:
Screen Shot 2022-12-19 at 6 35 22 PM

This also has the nice benefit of keeping the Add button in the same place when clicked, rather than it jumping down below the newly added entry.

### DIFF
--- a/airbyte-webapp/src/components/GroupControls/GroupControls.module.scss
+++ b/airbyte-webapp/src/components/GroupControls/GroupControls.module.scss
@@ -36,11 +36,10 @@ $border-width: variables.$border-thick;
   white-space: nowrap;
 }
 
-.dropdown {
+.control {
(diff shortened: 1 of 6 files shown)


flash1293 (approved):
Love it! Only nit - could you add a story for this to `airbyte-webapp/src/components/GroupControls/index.stories.tsx` ?

PR #20711, opened 2022-12-20
🪟 🔧 Fixed passed props to `<CheckBoxControl />` which cause error in console

dizel852:
## What
Fixes passed prop to `<CheckBoxControl />` which cause error in console

## How
Remove unnecessary passed props to `<CheckBoxControl />`:
- `checkbox` - redundant prop
- `message={meta.touched && meta.error && formatMessage({ id: meta.error })}` - there is no error message for this type of control 

image


YatsukBogdan1 (approved):
LGTM

edmundito:
Would you mind cleaning up the props in CheckBoxControl as well? Looks like the `checkbox: boolean` prop is there but is it needed?

dizel852:
> Would you mind cleaning up the props type/interface in CheckBoxControl as well? Looks like the `checkbox: boolean` prop is there but is it needed?

Sure, I wanted to make a "tiny" fix, but okay will fix it. 👍

edmundito (approved):

PR #20715, opened 2022-12-20
🪟🔧  Connector builder: use new lowcode manifest

flash1293:
## What

Makes connector builder compatible with latest CDK version

## How

Pulls in newly created yaml for the low code manifest via orval and fixes all the problems.

Note: After this PR is merged in, all PRs built on top of `feature/connector-builder` need to be updated accordingly.


airbyte-webapp/src/components/connectorBuilder/types.ts
```diff
@@ -45,36 +49,40 @@ export const builderFormValidationSchema = yup.object().shape({
...
       retriever: {
+        type: "SimpleRetriever",
         name: stream.name,
         requester: {
+          type: "HttpRequester",
           name: stream.name,
           url_base: values.global?.urlBase,
           path: stream.urlPath,
           request_options_provider: {
+            // TODO can't declar type here because the server will error out, but the types dictate it is needed. Fix here once server is fixed.
```
lmossman, line 63:
```suggestion
            // TODO can't declare type here because the server will error out, but the types dictate it is needed. Fix here once server is fixed.
```

lmossman (approved):
LGTM, thanks for putting up the changes! Tested locally and it works as expected

PR #20717, opened 2022-12-20
[ISSUE #20322] add datetime_granularity logic to DatetimeStreamSlicer…

maxi297:
… and migrate duration to ISO8601

## What
An issue was identified in the datetime stream slicer where some records are not fetched (see [this ticket](https://github.com/airbytehq/airbyte/issues/20322)). We later got report that [this issue was occurring for some users](https://github.com/airbytehq/airbyte/pull/20500). The lost of records can occur in two circumstances:
1. When we start fetching from "cursor + 1 day" (see [this](https://github.com/airbytehq/airbyte/blob/7ae05b16c6dc9975d0ce8c855212e5a487d2cac1/airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py#L157))
2. When we generate the end of a slice without considering the granularity (see [this](https://github.com/airbytehq/airbyte/blob/7ae05b16c6dc9975d0ce8c855212e5a487d2cac1/airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py#L168))

## How
To solve case #1, we now start fetching from the from the cursor itself. This will generate duplication but is something needed to ensure we avoid duplication. The other alternative is to maybe lose records. The case is the following: let’s assume that we fetch hourly an API with date format %Y-%m-%d. In such cases, the end_datetime is often an interpolation of `today_utc()`. What will probably happen is:
* At 12:00 AM, the date switch from 2022-12-12 to 2022-12-13
* Let’s assume we fetch at 12:30 AM. The fetch will have the boundaries:
    * start_datetime: 2022-12-12
    * end_datetime: 2022-12-13
* Stream state will have cursor on the 2022-12-13
* At 01:30 AM we fetch again. Which start_datetime/end_datetime should we use?

To solve case #2 above, where before we would always remove one day, now we remove the most granular unit of time based on the datetime_format. Therefore:
* Given a datetime_format of `"%Y-%m-%dT%H:%M:%S.%f%z"`, the slice will now be `"start_time": "2021-01-05T00:00:00.000000+0000", "end_time": "2021-01-05T23:59:59.999999+0000"` instead of `"start_time": "2021-01-05T00:00:00.000000+0000", "end_time": "2021-01-05T00:00:00.000000+0000"`
* Given a datetime_format of `"%Y-%m-%d"`, the result will be the same i.e. `"start_time": "2021-01-05", "end_time": "2021-01-05"`

A first attempt was done [here](https://github.com/airbytehq/airbyte/pull/20599) to autodetect the granularity of the datetime format directives. In the end, we decided it was not a good idea since a user could user could use a date format `%Y-%m-%dT%H:00:00` and therefore, the premise that we could use the datetimeformat to detect granularity was void. Hence here, we ask the user to provide the information.

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py`: refers to the business logic and therefore should be read first
2. `airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py`: contains the implementation to find the most granular unit of time

## 🚨 User Impact 🚨
As mentioned above, we might have more duplication how since we won't stop querying a cursor until we have new records in a more recent slice. I'll put up a document soon describing what we could do to reduce such duplication.

Although there will be more duplication, this will ensure we don't lose record.


airbyte-integrations/connectors/source-posthog/source_posthog/posthog.yaml
```diff
@@ -120,9 +119,9 @@ definitions:
         stream_slicers:
           - type: SubstreamSlicer
             parent_stream_configs:
-            - stream: "*ref(definitions.projects_stream)"
-              parent_key: "id"
-              stream_slice_field: "project_id"
+              - stream: "*ref(definitions.projects_stream)"
```
maxi297, line 122: This was formatted by pre-commit

airbyte-cdk/python/unit_tests/sources/declarative/stream_slicers/test_datetime_stream_slicer.py
```diff
@@ -33,27 +34,28 @@ def mock_datetime_now(monkeypatch):
...
+                {"start_time": "2021-01-01T00:00:00.000000+0000", "end_time": "2021-01-01T23:59:59.999999+0000"},
+                {"start_time": "2021-01-02T00:00:00.000000+0000", "end_time": "2021-01-02T23:59:59.999999+0000"},
+                {"start_time": "2021-01-03T00:00:00.000000+0000", "end_time": "2021-01-03T23:59:59.999999+0000"},
+                {"start_time": "2021-01-04T00:00:00.000000+0000", "end_time": "2021-01-04T23:59:59.999999+0000"},
+                {"start_time": "2021-01-05T00:00:00.000000+0000", "end_time": "2021-01-05T23:59:59.999999+0000"},
+                {"start_time": "2021-01-06T00:00:00.000000+0000", "end_time": "2021-01-06T23:59:59.999999+0000"},
+                {"start_time": "2021-01-07T00:00:00.000000+0000", "end_time": "2021-01-07T23:59:59.999999+0000"},
+                {"start_time": "2021-01-08T00:00:00.000000+0000", "end_time": "2021-01-08T23:59:59.999999+0000"},
+                {"start_time": "2021-01-09T00:00:00.000000+0000", "end_time": "2021-01-09T23:59:59.999999+0000"},
                 {"start_time": "2021-01-10T00:00:00.000000+0000", "end_time": "2021-01-10T00:00:00.000000+0000"},
```
maxi297, line 59: Those last slices are a little bit weird but should work if we keep the assumption that boundaries are always inclusive

airbyte-integrations/connectors/source-close-com/source_close_com/close_com.yaml
```diff
@@ -55,9 +55,10 @@ definitions:
     end_datetime:
       datetime: "{{ now_utc().strftime('%Y-%m-%dT%H:%M:%S.%f%z') }}"
       datetime_format: "%Y-%m-%dT%H:%M:%S.%f%z" # ISO8601
-    step: "365d"
+    step: "P365D"
     cursor_field: "date_updated"
     datetime_format: "%Y-%m-%d"
```
maxi297, line 60: Is there a reason why we allow different datetimeformat for start_datetime/end_datetime/datetime_format? Shouldn't we always allow the dateformat of start and end to be as granular as the one used in the API? Here, a user could set up:
* start_datetime: 2021-01-01T05:30:00.000000+0000
* end_datetime: 2021-02-01T12:30:00.000000+0000

... but since the datetime format is "%Y-%m-%d", it would in fact use:

* start_datetime: 2021-01-01
* end_datetime: 2021-02-01

This seems confusing to me since there will be 5 hours 30 minutes more data at the beginning than expected and 11 hours 30 minutes more data at the end than expected.

maxi297:
I'll be waiting for your review to update the version number from 0.16.1 to 0.17.0, update the changelog, release the airbyte-cdk and run some source test. 

That being said, this is somewhat a big change. Would there be a way to run the source acceptance tests before releasing the new version?

girarda:
@maxi297 you can follow these instructions to build the connectors with the local CDK files instead of the published ones https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/README.md?plain=1#L100.

You can then run the SATs with that custom docker image

airbyte-integrations/connectors/source-posthog/unit_tests/test_components.py
```diff
@@ -85,42 +86,48 @@ def test_update_cursor(test_name, initial_state, stream_slice, last_record, expe
...
-         [{'end_time': '2021-01-10T00:00:00.000000+0000',
+         [{'end_time': '2021-01-10T23:59:59.999999+0000',
            'project_id': '2331',
            'start_time': '2021-01-01T00:00:00.000000+0000'},
-          {'end_time': '2021-01-20T00:00:00.000000+0000',
+          {'end_time': '2021-01-20T23:59:59.999999+0000',
            'project_id': '2331',
-           'start_time': '2021-01-10T00:00:00.000000+0000'},
-          {'end_time': '2021-01-30T00:00:00.000000+0000',
+           'start_time': '2021-01-10T23:59:59.999999+0000'},
```
maxi297, line 94: posthog has code to align end_time from previous slide to start_time from current slice: https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-posthog/source_posthog/components.py#L114

airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py
```diff
@@ -45,22 +35,24 @@ class DatetimeStreamSlicer(StreamSlicer, JsonSchemaMixin):
     Attributes:
         start_datetime (Union[MinMaxDatetime, str]): the datetime that determines the earliest record that should be synced
         end_datetime (Union[MinMaxDatetime, str]): the datetime that determines the last record that should be synced
-        step (str): size of the timewindow
+        step (str): size of the timewindow (ISO8601 duration)
         cursor_field (Union[InterpolatedString, str]): record's cursor field
         datetime_format (str): format of the datetime
+        datetime_format_granularity (str): smallest increment the datetime_format has (ISO8601 duration)
```
clnoll, line 41: Similar to my comment above - this is quite literal but I wonder if we want the comment to reflect how it will be used (as we do for `start_datetime` and `end_datetime`.

airbyte-integrations/connectors/source-posthog/unit_tests/test_components.py
```diff
@@ -85,42 +86,48 @@ def test_update_cursor(test_name, initial_state, stream_slice, last_record, expe
     [
         ("test_empty_state",
          {},
-         [{'end_time': '2021-01-10T00:00:00.000000+0000',
+         [{'end_time': '2021-01-10T23:59:59.999999+0000',
            'project_id': '2331',
            'start_time': '2021-01-01T00:00:00.000000+0000'},
```
maxi297, line 91: The start time here would have normally changed since we don't do `state_start_time + 1 day` when we start from the state but the posthog source fix the start time from the first slice: https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-posthog/source_posthog/components.py#L118

airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py
```diff
@@ -178,19 +170,11 @@ def parse_date(self, date: str) -> datetime.datetime:
...
-        :return relativedelta: A relativedelta object
+        :return Parses an ISO 8601 durations into datetime.timedelta or Duration objects.
         """
-        parts = cls.timedelta_regex.match(time_str)
-
-        assert parts is not None
-
-        time_params = {name: float(param) for name, param in parts.groupdict().items() if param}
-        return relativedelta(**time_params)
+        return parse_duration(time_str)
```
clnoll, line 177: Can we find something with higher adoption that does the same sort of thing? Feels like we want to be careful introducing dependencies that aren't widely used. 

ETA: doing a quick search I didn't really see an alternative for parsing ISO duration. I wonder at what point it's better to use our own rather than introducing a dependency.

airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py
```diff
@@ -144,7 +133,10 @@ def stream_slices(self, sync_mode: SyncMode, stream_state: Mapping[str, Any]) ->
         stream_state = stream_state or {}
         kwargs = {"stream_state": stream_state}
         end_datetime = min(self.end_datetime.get_datetime(self.config, **kwargs), datetime.datetime.now(tz=self._timezone))
-        lookback_delta = self._parse_timedelta(self.lookback_window.eval(self.config, **kwargs) if self.lookback_window else "0d")
+        if self.lookback_window and (evaluated_lookback_window := self.lookback_window.eval(self.config, **kwargs)):
+            lookback_delta = self._parse_timedelta(evaluated_lookback_window)
+        else:
+            lookback_delta = self._parse_timedelta("P0D")
```
clnoll, line 139: Is this still what we want if user input for granularity is > 1 day? Wondering if that could result in scenario 1 (from your PR description) issues again.

clnoll (commented):
Nice @maxi297! Just left a few comments & questions for you.

maxi297:
> @maxi297 you can follow these instructions to build the connectors with the local CDK files instead of the published ones https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/README.md?plain=1#L100.
> 
> You can then run the SATs with that custom docker image

It did not seem to work for me. I added exceptions to make sure none of the tests passed and it still did. After checking, it seems like the SATs were failing to build the new Docker image but still run the tests which was confusing. Here was the error message:

```
 => ERROR [builder 5/6] COPY airbyte-cdk airbyte-cdk                                                                                                                                                                                            0.0s
------
 > [builder 5/6] COPY airbyte-cdk airbyte-cdk:
------
failed to compute cache key: too many links
latest: Pulling from airbyte/source-acceptance-test
```

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -330,6 +330,7 @@ definitions:
       - cursor_field
       - end_datetime
       - datetime_format
+      - datetime_format_granularity
```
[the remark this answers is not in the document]
maxi297: I agree with you. I was a little bit confused that there is a `end_datetime.datetime_format`, a `start_datetime.datetime_format` and a `datetime_format` where only the `datetime_format` applies to the actual slicing. Right now, the granularity applies to the `datetime_format` which is the cursor one hence why I tried to align these. @girarda, any preference here?

airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py
maxi297: I had the same question. I don't understand how parsing ISO8601 duration is not part of the standard Python library (my guess is that the standard library does not support timedelta for months and years but eh..)

Using `relativedelta` to support years and months, we would only have the parsing of the string to do. I don't have information to confirm it but my guts feeling was that it's less work to use this library than to write the parsing and test it. My reasoning was that there are 29,9k users for this lib which seems like a good enough community to battle test the code. If you have information that would suggest otherwise, please share them with me and I'll change that.

airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py
maxi297: I changed this code because the previous `_parse_timedelta` was taking an empty string as a parameter which makes me think I should have keep the behavior it had before in case someone else calls it with an empty string so I'll fix that.

I think this is still what we want. If a user specifies a lookback, I think we should use it whatever it is. Maybe we could have an error if the lookback_window is smaller than the granularity since the impact would be that the lookback_window applied would be the granularity and not the lookback per se. For example:
* lookback_window = 1 second
* datetimeformat = "%y-%m-%d"
* granularity = 1 day
* start_datetime = 2022-12-15T00:00:00Z

In this case, we would `2022-12-15T00:00:00Z - 1 second` which would give `2022-12-14T23:59:00Z`. However, when formatting this, it ends up being `2022-12-14` and therefore, the lookback window is 1 day, not 1 second. @girarda can you confirm a breaking exception here with a nice error message would make sense? Are there side effects that we would not want and maybe just logging is better?

girarda:
@maxi297 after builder the docker image with the dev tag, you can use the following command from the connector module to run the SATs:
`./acceptance-test-docker.sh`

Alternative command in case that file is not in all the connector module:
`python -m pytest integration_tests -p integration_tests.acceptance`

maxi297:
Ok! So I've updated the documentation regarding how we can test locally. If you are interested, please see [this PR](https://github.com/airbytehq/airbyte/pull/20821)

maxi297:
Test performed locally using this version of airbyte-cdk:
- source-close-com: N/A (tests are broken)
- source-coin-api: 404 OK
- source-coingecko-coins: N/A (tests are broken)
- source-delighted: OK
- source-posthog: OK
- source-punk-api: 404 OK
- source-square: OK
- source-tempo: OK
- source-tvmaze-schedule: 404 OK
- source-twitter: 404 OK
- source-wikipedia-pageviews: 404 OK
- source-woocommerce: OK

Note: "404 OK" means that I didn't have previous test results to base my changes on but the tests are still passing locally

airbyte-integrations/connectors/source-close-com/source_close_com/close_com.yaml
girarda: Yup this seems fishy. It also looks like the format of the cursor field is ISO8601 so it would make sense to use that same format for the 3 field (https://developer.close.com/resources/leads/). It would also simplify the component configuration.

airbyte-integrations/connectors/source-woocommerce/source_woocommerce/components.py
```diff
@@ -1,22 +0,0 @@
-#
-# Copyright (c) 2022 Airbyte, Inc., all rights reserved.
-#
-
-import datetime
-
-from airbyte_cdk.sources.declarative.stream_slicers import DatetimeStreamSlicer
-
-
-class DatetimeStreamSlicerWoocommerce(DatetimeStreamSlicer):
```
girarda, line 10: 🔪

docs/connector-development/config-based/understanding-the-yaml-file/stream-slicers.md
```diff
@@ -103,24 +106,28 @@ Example:
 stream_slicer:
   start_datetime: "2021-02-01T00:00:00.000000+0000",
   end_datetime: "2021-03-01T00:00:00.000000+0000",
-  step: "1d"
+  datetime_format: "%Y-%m-%dT%H:%M:%S.%f%z",
+  datetime_format_granularity: "PT0.000001S"
+  step: "P1D"
```
girarda, line 111: probably worth linking to docs describing the format

girarda (approved):
lgtm. I think this is the right tradeoff, especially since we have a plan to reduce the number of duplicates. 

left a couple of small comments, but nothing worth blocking the PR on

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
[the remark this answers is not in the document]
clnoll: @maxi297 were you going to make the variable name change here?

airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py
clnoll: Makes sense to me, and I agree with you that 29.9k users seems legit - I'd overlooked that (was looking at number of stars).

clnoll (commented):
Just one last q for you @maxi297!

Issue #20462, opened 2022-12-13
Tech Spec for "Improve Check Error Reporting"


evantahler:
We are going to add CHECK and DISCOVER failure reasons to the API and UI.

We need a shared [tech spec](https://docs.google.com/document/d/1GycFYbt572BrX8szKSkrAhsp_7KeoPyLlaj0lspcgxY/edit) which describes the API schema which will be returned.  The Connectors Extensibility team will work on the Front End, and Connector Ops will work on the API.

Some interesting topics to consider:
* What does the API look like?
* What are the cases where there might be a failure but no FailureReason (e.g. no AirbyteTraceErrorMessage) object.  How will those be represented.
 * Should the UI present all the data we have, or be "expandable" in some way?
* consider the heiercy of messages which could be shown:
  * status message, then failure.external_message, then failure.internal_message, then failure.stack_trace

evantahler:
Grooming:

* There's a similar API for showing failure reasons in jobs/syncs.  Should we copy this API? FailureReason objects are being returned here too

alafanechere:
@erohmensing provided valuable interrogations around the `check` flow here: https://github.com/airbytehq/airbyte/pull/20721#pullrequestreview-1246677467

erohmensing:
Link to the ratified tech spec [here](https://docs.google.com/document/d/16N5NAs9ahPTUEEebbij9fkny4vnjE4QnjkTuCqJKXbk/edit#)!

PR #20721, opened 2022-12-20
airbyte-common-workers: Collect trace message on failed connection_status

### LINKED ISSUE #18865: Collect trace message on failed connection_status
When there is a failed connection status (CHECK), sometimes the connector also emits an AirbyteTraceMessage with a failure reason.  Right now, we don't use the TraceMessage to build the failure reason like we do in a running sync.  We should!  

These errors are not presisted to the DB, but are sent to Sentry in https://github.com/airbytehq/airbyte/issues/19898

To reproduce this error:
* exit 0 (container didn't crash)
* The connector returned success=false + a TraceMessage.

We *do* properly capture trace messages with failures if the connector exited with a non-0 code.

alafanechere:
## What
Closes airbytehq/airbyte#18865 

Synchronous commands like SPEC, DISCOVER and CHECK can output trace messages that are not currently captured as failure reasons when the process exits with a `0` status code.

This PR aims at normalizing the behavior of these three commands:
- Do not consider the process exit code to characterize an error
- When a trace message is received: always add it to the job output
- When no trace message or catalog/connection status/spec is received -> throw a WorkerException

## How
Implement an explicit logic in:
https://github.com/airbytehq/airbyte/blob/1f6c31044fc0a092f7034f36bbd1c65f1467867b/airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultGetSpecWorker.java#L54

https://github.com/airbytehq/airbyte/blob/1f6c31044fc0a092f7034f36bbd1c65f1467867b/airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultCheckConnectionWorker.java#L59

https://github.com/airbytehq/airbyte/blob/1f6c31044fc0a092f7034f36bbd1c65f1467867b/airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultDiscoverCatalogWorker.java#L66

+ Update the unit tests


pedroslopez (commented):
We talked about this on a call, but leaving a comment for posterity:

Looks good overall, but we should:
- Add testing for the changes on both WorkerUtils and DefaultCheckConnectionWorker
- Look into following the same approach of always looking for trace messages in the other synchronous commands (spec, discover)

alafanechere:
@pedroslopez I tried to increase the shared logic between SPEC, DISCOVER and CHECK. I'd be interested in getting your opinion about the `throwOnNoFailureReason` approach.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultCheckConnectionWorker.java
```diff
@@ -78,28 +79,29 @@ public ConnectorJobOutput run(final StandardCheckConnectionInput input, final Pa
...
           .getOrDefault(Type.CONNECTION_STATUS, new ArrayList<>()).stream()
           .map(AirbyteMessage::getConnectionStatus)
           .findFirst();
 
-      if (status.isPresent() && exitCode == 0) {
+      final String defaultErrorMessage = String.format("Error checking connection, status: %s, exit code: %d", status, exitCode);
+
+      if (status.isEmpty()) {
+        LOGGER.error(defaultErrorMessage);
+        return WorkerUtils.getJobOutput(OutputType.CHECK_CONNECTION, failureReason, defaultErrorMessage, true);
```
pedroslopez, line 95: In reading this method being used it's not clear that this might actually throw - I would suggest either throwing here or having a different method that throws rather than having a boolean param.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultCheckConnectionWorker.java
```diff
@@ -78,28 +79,29 @@ public ConnectorJobOutput run(final StandardCheckConnectionInput input, final Pa
...
         LOGGER.debug("Check connection job received output: {}", output);
         LineGobbler.endSection("CHECK");
-        return new ConnectorJobOutput().withOutputType(OutputType.CHECK_CONNECTION).withCheckConnection(output);
-      } else {
-        final String message = String.format("Error checking connection, status: %s, exit code: %d", status, exitCode);
-        LOGGER.error(message);
-
-        return WorkerUtils.getJobFailureOutputOrThrow(OutputType.CHECK_CONNECTION, messagesByType, message);
+        ConnectorJobOutput jobOutput = WorkerUtils.getJobOutput(OutputType.CHECK_CONNECTION, failureReason, defaultErrorMessage, false);
+        return jobOutput.withCheckConnection(output);
```
pedroslopez, line 103: I think it's a little confusing having a `getJobOutput` method that doesn't fully get the job output (you still have to set the actual output on the JobOutput). Perhaps naming it something like `prepareJobOutput` or `buildJobOutput` can be clearer.

pedroslopez: Actually, we already have the FailureReason at this point (it's ok if it's null) so maybe we don't need that method at all and it can be 

```
return new ConnectorJobOutput()
    .withOutputType(CHECK_CONNECTION)
    .withFailureReason(failureReason)
    .withCheckConnection(output);
```

which doesn't seem too bad. Including things like error messages in that method call creates a bit of indirection imo since it's never going to be used

pedroslopez: (the same comment goes for discover and spec)

pedroslopez (commented):
Looks good logically, most of my comments are around making things a bit clearer but non-blocking.

The only thing I'd like to see before approving is some additional tests for WorkerUtils / Default***Workers.

alafanechere:
@pedroslopez I tried to make things a bit more readable and explicit: 
We do not check the exit code anymore: we throw a `WorkerException` if: 
- we can't find the object we're trying to read from the output (connection status, spec, catalog) 
- **AND** we don't have a failure reason

This logic is kept within the worker classes (Default*Worker.java), the Worker utils class only has utils to extract useful objects from messages. 

I'll add the tests if the overall logic looks good to you.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultCheckConnectionWorker.java
```diff
@@ -67,38 +64,35 @@ public ConnectorJobOutput run(final StandardCheckConnectionInput input, final Pa
...
           .map(AirbyteMessage::getConnectionStatus)
           .findFirst();
 
-      if (status.isPresent() && exitCode == 0) {
-        final StandardCheckConnectionOutput output = new StandardCheckConnectionOutput()
-            .withStatus(Enums.convertTo(status.get().getStatus(), Status.class))
-            .withMessage(status.get().getMessage());
+      final Optional<FailureReason> failureReason = WorkerUtils.getJobFailureReasonFromMessages(OutputType.CHECK_CONNECTION, messagesByType);
+      if (failureReason.isPresent()) {
+        jobOutput = jobOutput.withFailureReason(failureReason.get());
```
pedroslopez, line 81: nit: can use `.setFailureReason`  / `.setCheckConnection` so we don't have to reassign jobOutput.

pedroslopez (commented):
Nice! I think this makes it much clearer. Only one comment about reassigning the job output that is non-blocking - just waiting on the tests 😄

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultDiscoverCatalogWorker.java
```diff
@@ -73,43 +70,40 @@ public ConnectorJobOutput run(final StandardDiscoverCatalogInput discoverSchemaI
...
-        return new ConnectorJobOutput().withOutputType(OutputType.DISCOVER_CATALOG_ID).withDiscoverCatalogId(catalogId);
+        jobOutput = jobOutput.withDiscoverCatalogId(catalogId);
       } else {
-        return WorkerUtils.getJobFailureOutputOrThrow(
-            OutputType.DISCOVER_CATALOG_ID,
-            messagesByType,
-            String.format("Discover job subprocess finished with exit code %s", exitCode));
+        if (failureReason.isEmpty()) {
+          throw new WorkerException("Integration failed to output a catalog struct.");
+        }
```
pedroslopez, line 104: This means that if a connector:
- exited with a non-zero exit code
-  _did_ output a catalog
- _did not_ output a failure reason

We would consider this a success. Should these conditions throw an exception as well? A non-zero exit code probably means _something_ went wrong but we just don't know what (since there isn't a failure reason)

alafanechere: Yes, you are right. 
But should we throw an exception if we can parse a catalog?  
I'd like to characterize an error only when we can't retrieve what we are looking for (spec, connection status, catalog).

pedroslopez: I don't feel super strongly about this, but I think having a non-zero exit code returned by the source means something went wrong so the job should fail. My worry would be that even though we have a catalog, a non-zero exit code may mean we shouldn't trust that catalog to be correct. Ideally we get a failure reason that tells us more, but it's kind of ambiguous if there isn't one present.

I can see the other side of the argument though. The protocol doesn't really say anything about the exit code, so just the presence of the right type of message might be enough on that side. I can also see it being ok to process the output if the exit code indicates something like the container was killed because it ran out of memory.

alafanechere:
@pedroslopez I updated the tests.
@evantahler could you suggest a tie break 😄 : Shall we throw an exception when the process exit code of this command is not `0`. The logic I implemented is the following, ignoring the status code: 
- If no trace message is captured: throw an exception
- If a trace message is captured: set the failure reason in the job output
- If a check status / spec / catalog AND a trace message is captured: set status / spec / catalog + a failure reason on the output.

pedroslopez (approved):

pedroslopez (approved):
Looks good, though some tests are failing - ok to merge once those are green. Will leave the final decision on whether to fail on non-zero to you 😄

evantahler:
Regarding `If no trace message is captured: throw an exception`, there's a case I'd like to learn more about:
* The connector exists with a non-0 exit code
* No trace messages were emitted
* The stack trace the connector emitted is meaningful.

Maybe the container doesn't have enough disk or memory to boot up.  The OS or python stack would say something like `could not allocate more RAM, crashing...`, and that message would be important to get back to the user.  Can the exception you throw take the last line of stderr or stdout and append that?

alafanechere:
@evantahler would something like https://github.com/airbytehq/airbyte/pull/20721/commits/51d8da17722b0b802d2b2eb1ea812a11b1ff6836 match your suggestion?

alafanechere:
I think I discovered a side effect:
When setting failure reason on a process that exit with a `0` status code the server response (using temporal metadata) is not considered a `success` (code [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-server/src/main/java/io/airbyte/server/handlers/SchedulerHandler.java#L435)) and leads to a generic `Check Connection Failed!` error message in the front-end.

On my branch while setting a bad PG connection (I double-check the job output has a failure and the exit code is 0):
Screen Shot 2023-01-03 at 18 42 24


On `0.40.26`:
Screen Shot 2023-01-03 at 18 47 36

@pedroslopez and @evantahler it looks like a regression to me so I might tweak the `SchedulerHandler` code to get a similar error output as we have on 0.40.26.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultDiscoverCatalogWorker.java
erohmensing: FWIW, I'd definitely be confused if I saw a log that said a job exited with a non-zero status, and everything continued as planned. If we think we can generally trust the catalog outputted in this case, at least indicating a non-zero exit code as a warning might be helpful for hunting it down if things get messed up due to a strange catalog?

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultGetSpecWorker.java
```diff
@@ -59,39 +55,37 @@ public ConnectorJobOutput run(final JobGetSpecConfig config, final Path jobRoot)
...
-            messagesByType,
-            String.format("Spec job subprocess finished with exit code %s", exitCode));
+        if (failureReason.isEmpty()) {
+          final String stderr = WorkerUtils.getStdErrFromErrorStream(process.getErrorStream());
+          if (stderr.isEmpty()) {
+            throw new WorkerException("Integration failed to output a spec struct and did not output a failure reason.");
+          } else {
+            throw new WorkerException("Integration failed to output a spec struct and did not output a failure reason:\n" + stderr);
+          }
+        }
```
erohmensing, line 85: Nit: here you indicate that there wasn't a failure reason for both cases, which seems correct here - but in the other two jobs, it's left out for both, or only included where there is no stderr output. 

Also wondering if it's worth combining these back into a function like before (but not returning) - something like `throwOnEmptyFailureReason("Integration failed to output a spec struct and did not output a failure reason.")`, since they all use the same code except for the included error message

pedroslopez:
@alafanechere I've updated the branch with the latest changes from master and fixed the conflicts that came up because of that.

 I took a stab at addressing the issue you raised about swallowing the check connection messages when a failure reason is also returned by the connector in [1f2138e](https://github.com/airbytehq/airbyte/pull/20721/commits/1f2138e51f6f0bf2ffb056c7dc768f1150646ae8). #20462 should provide more guidance as to what we should show in the case we receive both of these, but to keep behavior the same as it was prior I've changed the `succeeded` check to just verify that we have an output. I think saying the attempt failed if there was a failure reason is still correct, we just want to return the right thing to the api if we have it.

Before

image

After

image

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultDiscoverCatalogWorker.java
alafanechere: @erohmensing I've changed the logic to log a warning if status code != 0.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultGetSpecWorker.java
alafanechere: I combined this logic in a utility function ✅ .

> here you indicate that there wasn't a failure reason for both cases, which seems correct here - but in the other two jobs, it's left out for both, or only included where there is no stderr output.

I'm not sure I understand this comment. I believe the logic is similar across the three classes because I implemented the utility function.

alafanechere: I figured you mean the error message :D

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultCheckConnectionWorker.java
```diff
@@ -73,18 +70,12 @@ public ConnectorJobOutput run(final StandardCheckConnectionInput input, final Pa
...
-      }
+      LineGobbler.gobble(process.getErrorStream(), LOGGER::error);
 
-      final int exitCode = process.exitValue();
-      final Optional<AirbyteConnectionStatus> status = messagesByType
+      final Map<Type, List<AirbyteMessage>> messagesByType = WorkerUtils.getMessagesByType(process, streamFactory, 30);
+      final Optional<AirbyteConnectionStatus> connectionStatus = messagesByType
           .getOrDefault(Type.CONNECTION_STATUS, new ArrayList<>()).stream()
           .map(AirbyteMessage::getConnectionStatus)
           .findFirst();
```
erohmensing, line 81: Want to make sure my understanding is correct - here we look at all of the messages getting sent though, filter for connection status messages and grab the connection status from those, but only use the first one as the `connectionStatus`. Do we expect to have multiple of these messages and if so, why does the first one accurately represent the connection status?

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultCheckConnectionWorker.java
```diff
@@ -104,25 +95,30 @@ public ConnectorJobOutput run(final StandardCheckConnectionInput input, final Pa
...
-        return new ConnectorJobOutput().withOutputType(OutputType.CHECK_CONNECTION).withCheckConnection(output);
-      } else {
-        final String message = String.format("Error checking connection, status: %s, exit code: %d", status, exitCode);
-        LOGGER.error(message);
-
-        return WorkerUtils.getJobFailureOutputOrThrow(OutputType.CHECK_CONNECTION, messagesByType, message);
+            .withStatus(Enums.convertTo(connectionStatus.get().getStatus(), Status.class))
+            .withMessage(connectionStatus.get().getMessage());
+        LOGGER.info("Check connection job received output: {}", output);
+        jobOutput.setCheckConnection(output);
```
erohmensing, line 111: So we determine success by checking the `jobOutput` for a `checkConnection`, which will only be set if there is a connectionStatus.

We return the jobOutput either way, but if it's successful we can `getCheckConnection`, otherwise we can't and we need to check the FailureReason to figure out why

erohmensing: Do we need to worry at all about any places where `worker.run()` might interpret receiving a JobOutput with no checkConnection but with a FailureReason as a success, since we used to always return a jobOutput with a checkConnection, or throw ?

erohmensing (commented):
I think this looks pretty good -- I added a few clarifying questions mostly for my own understanding! Following the logic, it looks like this 
* looks for a connection status message
* looks for failure reasons in check connection messages
* sets the failure reason of the job if there's failure reasons
* if we got a connection status message, set the checkConnection output of the jobOutput
* if no connection status message and no failure reason, throw
* return the jobOutput

I followed the logic for checkConnection, and glanced over the other 2 jobs.

A few things that might have helped understand this better at first glance of a new code base:
* the existence CONNECTION_STATUS messages are being used to determine success, I think. could one consider a `connectionStatus` to be something like a connectionSuccess or connectionConfirmation? 
* I wonder if it would be worth adding a method `isSuccess` or `isSuccessful` that returns `output.getCheckConnection != null`? (and similar for the other methods)? On spec, it's relatively intuitive that if `getSpec` is null, the job failed, but for `getCheckConnection`, as someone new to the code base, it feels like a method named `getCheckConnection` could contain information about a failed job or a successful job - I have to look at the code to see that it only gets set in the case of success.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultCheckConnectionWorker.java
alafanechere: I think the `first` makes sense because the connector should output a single status message. But this is not something the platform should be aware of. I think picking the last status instead of the first one would be safer. I will create an issue.

alafanechere: https://github.com/airbytehq/airbyte/issues/21400

erohmensing: Yeah, that seems like a very implicit expectation! Agree that the last probably is more likely to represent the actual status (e.g. if we for some reason output status messages for multiple retries while trying to connect). Thanks for making an issue

PR #20725, opened 2022-12-20
🪟🎉 Connector form: Use proper validation in array section

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/20234

Screenshot 2022-12-20 at 17 03 30

## How

Instead of holding the state of the modal form in a separate place, edit in in-place so formik validation just works.
In order to keep the form behind the UI stable and to be able to roll back on the user hitting the Cancel button, the ArraySection pulls the item in edit into a local state and renders the list of items based on that.

On submitting the modal form, that local state is thrown away.

Based on top of the refactoring PR because I was running into weird issues with the unit tests which do not occur after the refactoring.


lmossman:
This PR title probably shouldn't start with `Connector builder:`, since it is modifying ArraySection which the connector builder does not use. Did you mean ConnectorForm?

airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/ArraySection.tsx
```diff
@@ -79,10 +103,13 @@ export const ArraySection: React.FC<ArraySectionProps> = ({ formField, path, dis
           render={(arrayHelpers) => (
             <ArrayOfObjectsEditor
               editableItemIndex={editIndex}
-              onStartEdit={setEditIndex}
+              onStartEdit={(n) => {
+                setEditIndex(n);
+                setItemInEdit(items[n]);
```
lmossman, line 108: Do we not need to create a copy of this?

airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/ArraySection.tsx
```diff
@@ -42,15 +42,26 @@ const getItemDescription = (item: Record<string, string>, properties: FormBlock[
 };
 
 export const ArraySection: React.FC<ArraySectionProps> = ({ formField, path, disabled }) => {
-  const [field, , fieldHelper] = useField(path);
-  const [editIndex, setEditIndex] = useState<number>();
+  const [field, , fieldHelper] = useField<Array<Record<string, string>>>(path);
+  const [editIndex, setEditIndex] = useState<number | undefined>();
+  const [itemInEdit, setItemInEdit] = useState<Record<string, string> | undefined>();
```
lmossman, line 47: Initially I thought `itemInEdit` contained the new value of the item, and it took me a bit to understand that this is the old value of the item from before it was edited.
Maybe a name like `uneditedItem`, `oldItem`, `prevItem` would make this a bit more clear?

lmossman (approved):
A couple small comments but neither blocking, LGTM. Tested locally and seems to work as expected

airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/ArraySection.tsx
flash1293: yes, to avoid the live changes in the modal form to be reflected behind the backdrop as well as to restore the old state in case the user cancels.

airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/ArraySection.tsx
flash1293: Good point, changed

flash1293:
> Did you mean ConnectorForm

I did, but I typed connector builder too many times in the last two weeks.

PR #20737, opened 2022-12-20
🪟 🐛 🧹 Fix issues with Button component

edmundito:
## What
* Ensures that the Button element is set to disabled when in the loading state
* Ensures that disable loading state does not apply the disabled state to the button
* Cleans up use of `defaultProps`

### DIFF
--- a/airbyte-webapp/src/components/ui/Button/Button.tsx
+++ b/airbyte-webapp/src/components/ui/Button/Button.tsx
@@ -8,19 +8,21 @@ import { ButtonProps } from "./types";
 
 export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>((props, ref) => {
   const {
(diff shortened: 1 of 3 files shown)


airbyte-webapp/src/components/ui/Button/Button.tsx
```diff
@@ -8,19 +8,21 @@ import { ButtonProps } from "./types";
 
 export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>((props, ref) => {
   const {
+    full = false,
+    size = "xs",
+    iconPosition = "left",
+    variant = "primary",
```
krishnaglick, line 14: Nit: Could convert these to a `const enum`

krishnaglick (approved):
LGTM

PR #20558, opened 2022-12-16
Create realistic datasets of 10GB, 100GB, and 1TB in size

evantahler:
Closes https://github.com/airbytehq/airbyte/issues/20556

---

We need big data to test with (specifically, @akashkulk does)!

## Idea 1: Sync data using Faker

We made 3 `source-faker` connections which would produce the desired amount of data

```sql
SELECT 
    pg_size_pretty(pg_total_relation_size('"public"."users_10M"')) -- 2909MB
	, pg_size_pretty(pg_total_relation_size('"public"."purchases_10M"')) -- 1024MB
	
-- (2909+2237)/1024 = 5.02GB
```

```
So... we are going to need 2 Billion faker users for 1TB
10,000,000*(1024/5.02) = 2,039,840,637
200 Million faker users for 100GB
10,000,000*(100/5.02) = 199,203,187
20 Million faker users for 10GB
10,000,000*(10/5.02) = 19,920,318
```

While the first small sync completed (20M faker users, 10GB of data), the remaining 2 keep hitting platform errors and failing after a few days.  Due to the current instability of the platform, this is unlikely to work.  Also of note, each attempt leaves lingering `airbyte_tmp` tables around, ever increasing the cost of the server's storage...

## Idea 2: Generate CSV files

Borrowing from the [Performance Research](https://docs.google.com/presentation/d/1SpzxbElxZV_m9-i3tR-94-CkEjRaf0cW2RJ8EO88Yqs/edit#slide=id.g1918d255a5d_0_162), the next plan is to produce a number of CSV files which will contain the proper data.

To do this, we will:
1. Modify the faker source to emit data over stdout in CSV format
2. Pipe that data to a csv file on disk in 10GB chunks.  We can manually manipulate faker's `/secrets/state.json` to move our cursor forward each chunk
3. Upload that data to Google Cloud Storage
4. Load that data into postgres via the "import from cloud storage" option

A bonus of this approach is that we will be able to persist these CSV files for repeated use on other databases

Ideas about how to take a stream of JSON and turn it into CSV -> https://stackoverflow.com/questions/32960857/how-to-convert-arbitrary-simple-json-to-csv-using-jq

PR #20741, opened 2022-12-20
Faker gets 250% faster

evantahler:
After the platform performance [research](https://docs.google.com/presentation/d/1SpzxbElxZV_m9-i3tR-94-CkEjRaf0cW2RJ8EO88Yqs/edit#slide=id.g1918d255a5d_0_98) and the failure to [use Faker within the Airbyte platform to generate 1TB](https://github.com/airbytehq/airbyte/pull/20558) of data, I spent some time looking into the bottlenecks of our Python CDK sources.  [Just like our Java sources](https://docs.google.com/presentation/d/1u1E8e1QYev6bnI-ZyfGK6lz9IS4L4ReWnlOQIzy96Nc/edit#slide=id.g1b4feb94f64_0_172), the limiting factors are:
* Parallelization of CPU-intensive tasks
* JSON parsing is slow

This PR explores solutions to these problems:
1. Use a [WorkerPool](https://docs.python.org/3/library/multiprocessing.html) to generate the fake records in parallel.
    * The code had to be rearchitected to become thread-safe
    * The number of worker "threads" is now available within the connector's configuration (default 4)
2. In those parallel workers, also generate the AirbyteRecord JSON in parallel, ofloading that work from the main thread. 
    * A stub `AirbyteMessageWithCachedJSON` message was created that renders its own JSON string upon intilization

How much faster did we get?  **The speedups in this PR add up to a 250% speedup for a 100,000 user sync (20s to 7s)!**

Notes:
* Testing (SAT) now needs to run with only one thread or else you can't predict the order of the records returned


sherifnada (commented):
> I spent some time looking into the bottlenecks of our Python CDK sources

Do you have some graphs or numbers you can share about CPU hotspots?

do you have a good understanding of why this speed up happened?
My understanding of Python is that due to the GIL, multi-threading doesn’t help on CPU-bound operations like Faker (so threads are mostly useful to speed up blocking I/O bound operations). This is just my mental model which is subject to change when brought into contact with real world experience.

evantahler:
@sherifnada this code uses multi-process code, and your mental model is right!  I initially started with mutli-threading, and indeed, there's wasn't a speed up. WorkerPool actually makes a few different python process and shares data with IPC, and that's how we win.

airbyte-integrations/connectors/source-faker/source_faker/streams.py
```diff
@@ -4,21 +4,37 @@
...
 
-from .utils import format_airbyte_time, read_json
+from .airbyte_message_with_cached_json import AirbyteMessageWithCachedJSON
+from .utils import format_airbyte_time, generate_estimate, now_millis, read_json
+
+
+class FakerMultithreaded:
+    def worker_init(self):
+        """For the workers, we want a unique instance of the faker packages with their own seeds to differentiate the generated responses"""
+        global person
```
sherifnada, line 22: why are these global? wouldn't every instance of `FakerMultithreaded` children (e.g: `Users`) have their own instance anyways? or is the goal to have a single static instance of person/datetime/numeric shared across all ?

airbyte-integrations/connectors/source-faker/source_faker/streams.py
```diff
@@ -89,73 +104,79 @@ def state(self, value: Mapping[str, Any]):
...
         median_record_byte_size = 450
         yield generate_estimate(self.name, self.count - total_records, median_record_byte_size)
 
-        for i in range(total_records, self.count):
-            user = self.generate_user(i)
-            yield user
-            total_records += 1
-            records_in_sync += 1
-            records_in_slice += 1
+        running = True
```
sherifnada, line 149: is there value in the `running` var? seems like wherever we set it to false we also `break`

airbyte-integrations/connectors/source-faker/source_faker/spec.json
```diff
@@ -35,8 +35,16 @@
         "description": "How many fake records will be in each page (stream slice), before a state message is emitted?",
         "type": "integer",
         "minimum": 1,
-        "default": 100,
+        "default": 1000,
         "order": 3
+      },
+      "threads": {
+        "title": "Worker Threads",
+        "description": "How many parallel worker threads should be used to generate records?",
```
sherifnada, line 43: is the user really in a position to tell us what this number should be? can't we make a best effort guess internally? even as an engineer I might be confused at what this number should be.

airbyte-integrations/connectors/source-faker/source_faker/spec.json
```diff
@@ -35,8 +35,16 @@
         "description": "How many fake records will be in each page (stream slice), before a state message is emitted?",
         "type": "integer",
         "minimum": 1,
-        "default": 100,
+        "default": 1000,
         "order": 3
+      },
+      "threads": {
+        "title": "Worker Threads",
```
sherifnada, line 42: can we call this `Number of parallel data generation processes` or something that: 
1) doesn't use the word `Worker` (my guess is the user doesn't know what a worker is) 
2) doesn't contain the word `threads`

airbyte-integrations/connectors/source-faker/source_faker/spec.json
sherifnada, line 43: if we must keep this optino, could you add some guidance on how one should choose this number? (or at least tell me if I don't know what this is doing, just ignore it)

airbyte-integrations/connectors/source-faker/source_faker/streams.py
```diff
@@ -60,18 +76,17 @@ def read_records(self, **kwargs) -> Iterable[Mapping[str, Any]]:
...
-    def __init__(self, count: int, seed: int, records_per_sync: int, records_per_slice: int, **kwargs):
+    def __init__(self, count: int, seed: int, threads: int, records_per_sync: int, records_per_slice: int, **kwargs):
         super().__init__(**kwargs)
         self.count = count
         self.seed = seed
         self.records_per_sync = records_per_sync
         self.records_per_slice = records_per_slice
-        self.person = Person(locale=Locale.EN, seed=self.seed)
-        self.dt = Datetime(seed=self.seed)
+        self.threads = threads
```
sherifnada, line 89: can we call this something other than `threads` in the code? 😬

airbyte-integrations/connectors/source-faker/source_faker/streams.py
```diff
@@ -183,77 +208,89 @@ def random_date_in_range(
...
 
-        for i in range(total_user_records, self.count):
-            purchases = self.generate_purchases(i + 1, total_purchase_records)
-            for purchase in purchases:
-                total_purchase_records += 1
-                yield purchase
-            total_user_records += 1
-            user_records_in_sync += 1
-            user_records_in_slice += 1
+        running = True
```
sherifnada, line 274: same question about `running`

airbyte-integrations/connectors/source-faker/source_faker/airbyte_message_with_cached_json.py
```diff
@@ -0,0 +1,17 @@
+#
+# Copyright (c) 2022 Airbyte, Inc., all rights reserved.
+#
+
+from airbyte_cdk.models import AirbyteMessage
+
+
+class AirbyteMessageWithCachedJSON(AirbyteMessage):
+    """I a monkeypatch to AirbyteMessage which pre-renders the JSON-representation of the object upon initialization.  This allows the JSON to be calculated in the thread/process that builds the object rather than the main thread."""
```
sherifnada, line 9:
```suggestion
    """I a monkeypatch to AirbyteMessage which pre-renders the JSON-representation of the object upon initialization.  This allows the JSON to be calculated in the process that builds the object rather than the main thread."""
```

airbyte-integrations/connectors/source-faker/source_faker/streams.py
```diff
@@ -4,21 +4,37 @@
...
 from airbyte_cdk.sources.streams import IncrementalMixin, Stream
 from mimesis import Datetime, Numeric, Person
 from mimesis.locales import Locale
 
-from .utils import format_airbyte_time, read_json
+from .airbyte_message_with_cached_json import AirbyteMessageWithCachedJSON
+from .utils import format_airbyte_time, generate_estimate, now_millis, read_json
+
+
+class FakerMultithreaded:
```
sherifnada, line 19: nitpick but nothing about this class (I think?) does any threading, so it was a little confusing to see it had `multithreaded` in the name

airbyte-integrations/connectors/source-faker/source_faker/streams.py
```diff
@@ -89,73 +104,83 @@ def state(self, value: Mapping[str, Any]):
...
 
-        for i in range(total_records, self.count):
-            user = self.generate_user(i)
-            yield user
-            total_records += 1
-            records_in_sync += 1
-            records_in_slice += 1
+        running = True
+        with Pool(initializer=self.worker_init, processes=self.threads) as pool:
+            while running and records_in_sync < self.count:
```
sherifnada, line 151: to help the next person working on this code, it would be super useful to describe the parallelism model in words in a docstring

sherifnada (changes requested):

evantahler:
Docs for how to allocate more resources to a connector when deployed: https://docs.airbyte.com/operator-guides/configuring-connector-resources#configuring-connector-specific-requirements

airbyte-integrations/connectors/source-faker/source_faker/airbyte_message_with_cached_json.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._json = self.json(exclude_unset=True)
+        self.json = self.get_json
+
+    def get_json(self, **kwargs):
```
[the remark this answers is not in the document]
evantahler: We can't use @cache here because the LRU cache is not serializable when passed to child workers.  

Comment added in code

airbyte-integrations/connectors/source-faker/source_faker/streams.py
evantahler: Comment added to code after much experimentation:

>         Note: the instances of the mimesis generators need to be global.
>         Yes, they *should* be able to be instance variables on this class, which should only instantiated once-per-worker, but that's not quite the case:
>         * relying only on prepare as a pool initializer fails because we are calling the parent process's method, not the fork
>         * Calling prepare() as part of generate() (perhaps checking if self.person is set) and then `print(self, current_process()._identity, current_process().pid)` reveals multiple object IDs in the same thread, resetting the internal random counters

airbyte-integrations/connectors/source-faker/source_faker/purchase_generator.py
```diff
@@ -0,0 +1,103 @@
...
+        * relying only on prepare as a pool initializer fails because we are calling the parent process's method, not the fork
+        * Calling prepare() as part of generate() (perhaps checking if self.person is set) and then `print(self, current_process()._identity, current_process().pid)` reveals multiple object IDs in the same process, resetting the internal random counters
+        """
+
+        seed_with_offset = self.seed
+        if self.seed is not None and len(current_process()._identity) > 0:
+            seed_with_offset = self.seed + current_process()._identity[0]
+
+        global dt
+        global numeric
```
evantahler, line 34: @sherifnada I was able to address about 1/2 of your comments.  For the rest, I added docs :D

sherifnada (approved):

PR #20745, opened 2022-12-21
Update airbyte protocol migration

### LINKED ISSUE #20695: Update airbyte protocol migration
## What

We have two types of objects at the top level of the Airbyte protocol: `AirbyteMessage` and `ConfiguredAirbyteCatalog`.
In order to support migrations for all objects, we should add the migration layer for `ConfiguredAirbyteCatalog`

Some migrations of AirbyteMessage need the Configured catalog for a better upgrade/downgrade. `AirbyteMessageMigration::{downgrade,upgrade}` should take an extra `ConfiguredAirbyteCatalog`

## How

* Add `ConfiguredAirbyteCatalogMigration` similar to `AirbyteMessageMigration`.
* `AirbyteMessageMigration::{downgrade,upgrade}` should take an extra `ConfiguredAirbyteCatalog`

gosusnp:
## What

Addresses a few gaps in the current protocol migration:
* `AirbyteMessage` migration need the `ConfiguredAirbyteCatalog` to be able to downgrade messages
* `ConfiguredAirbyteCatalog` is a top level object that has a separate lifecycle from the `AirbyteMessage`, add the necessary classes to support its migration
* `ConfiguredAirbyteCatalog` is passed as part of the connector inputs that has a separate serialization, we need to add versioning support in that code path.

Closes #20695 

## How

* Add ConfiguredAirbyteCatalog migrations
* Pass ConfiguredAirbyteCatalog to AirbyteMessage migrations
* Add `ProtocolSerializer` to wrap migrations that are not directly no AirbyteMessages
* Some refactoring to share the migration selection logic across  `AirbyteMessageMIgrator` and `ConfiguredAirbyteCatologMigrator`. As a result, `MigrationContainer` has been extracted

## Recommended reading order
1.  airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/{MigrationContainer.java,Migration.java}: refactoring to share migration selection logic
2. airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/ConfiguredAirbyteCatalogMigrator.java: Entry point for the ConfiguredAirbyteCatalog migrations
3. airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/ProtocolSerializer.java: Entry point for the abstraction to serialize protocol objects that are not directly AirbyteMessages
4. Rest is mostly wiring to get the new objects where they are needed

## 🚨 User Impact 🚨

There should be no user impact.
This will enable the migration of the ConfiguredAirbyteCatalog.


edgao (approved):
lgtm, I only really reviewed the migration interface stuff (i.e. skipped over all the files that actually call into migrations). Did a quick merge into my upgrade/downgrade branches locally and they seem happy, so :shipit: and I'll do the merges correctly + extract the catalog migration

edgao:
@gosusnp is this good to merge (into https://github.com/airbytehq/airbyte/pull/20036)?

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/AirbyteProtocolVersionedMigratorFactory.java
```diff
@@ -0,0 +1,37 @@
...
+package io.airbyte.commons.protocol;
+
+import io.airbyte.commons.version.Version;
+import jakarta.inject.Singleton;
+
+/**
+ * Factory to build AirbyteMessageVersionedMigrator
+ */
+@Singleton
+public class AirbyteProtocolVersionedMigratorFactory {
```
benmoriceau, line 14: Should this be an actual micronaut factory (the `@Factory` annotation)?

benmoriceau (commented):
LGTM with a small comment about creation a micronaut factory

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/AirbyteProtocolVersionedMigratorFactory.java
```diff
@@ -0,0 +1,37 @@
...
+  private final ConfiguredAirbyteCatalogMigrator configuredAirbyteCatalogMigrator;
+
+  public AirbyteProtocolVersionedMigratorFactory(final AirbyteMessageMigrator airbyteMessageMigrator,
+                                                 final ConfiguredAirbyteCatalogMigrator configuredAirbyteCatalogMigrator) {
+    this.airbyteMessageMigrator = airbyteMessageMigrator;
+    this.configuredAirbyteCatalogMigrator = configuredAirbyteCatalogMigrator;
+  }
+
+  public <T> AirbyteMessageVersionedMigrator<T> getAirbyteMessageMigrator(final Version version) {
+    return new AirbyteMessageVersionedMigrator<>(this.airbyteMessageMigrator, version);
```
colesnodgrass, line 26: is `this` qualifier necessary here?

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/ConfiguredAirbyteCatalogMigrator.java
```diff
@@ -0,0 +1,68 @@
...
+    migrationContainer.initialize();
+  }
+
+  /**
+   * Downgrade a message from the most recent version to the target version by chaining all the
+   * required migrations
+   */
+  public <PreviousVersion, CurrentVersion> PreviousVersion downgrade(final CurrentVersion message, final Version target) {
+    return migrationContainer.downgrade(message, target, ConfiguredAirbyteCatalogMigrator::applyDowngrade);
+  }
```
colesnodgrass, line 36: Do we need to validate that this version is indeed a downgrade?

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/ConfiguredAirbyteCatalogMigrator.java
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * Upgrade a message from the source version to the most recent version by chaining all the required
+   * migrations
+   */
+  public <PreviousVersion, CurrentVersion> CurrentVersion upgrade(final PreviousVersion message, final Version source) {
+    return migrationContainer.upgrade(message, source, ConfiguredAirbyteCatalogMigrator::applyUpgrade);
+  }
```
colesnodgrass, line 44: Do we need to validate that this version is indeed an upgrade?

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/ConfiguredAirbyteCatalogMigrator.java
```diff
@@ -0,0 +1,68 @@
...
+
+  public Version getMostRecentVersion() {
+    return migrationContainer.getMostRecentVersion();
+  }
+
+  // Helper function to work around type casting
+  private static <PreviousVersion, CurrentVersion> PreviousVersion applyDowngrade(final ConfiguredAirbyteCatalogMigration<PreviousVersion, CurrentVersion> migration,
+                                                                                  final Object message) {
```
colesnodgrass, line 52: As this method is generic, can you have `final CurrentVersion message` instead of `final Object message`?

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/ConfiguredAirbyteCatalogMigrator.java
```diff
@@ -0,0 +1,68 @@
...
+    return migration.downgrade((CurrentVersion) message);
+  }
+
+  // Helper function to work around type casting
+  private static <PreviousVersion, CurrentVersion> CurrentVersion applyUpgrade(final ConfiguredAirbyteCatalogMigration<PreviousVersion, CurrentVersion> migration,
```
colesnodgrass, line 57: As this method is generic, can you have `final PreviousVersion message` instead of `final Object message`?

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/VersionedProtocolSerializer.java
```diff
@@ -0,0 +1,26 @@
...
+  private final Version protocolVersion;
+
+  public VersionedProtocolSerializer(final ConfiguredAirbyteCatalogMigrator configuredAirbyteCatalogMigrator, final Version protocolVersion) {
+    this.configuredAirbyteCatalogMigrator = configuredAirbyteCatalogMigrator;
+    this.protocolVersion = protocolVersion;
+  }
+
+  @Override
+  public String serialize(final ConfiguredAirbyteCatalog configuredAirbyteCatalog) {
+    return Jsons.serialize(configuredAirbyteCatalogMigrator.downgrade(configuredAirbyteCatalog, protocolVersion));
```
colesnodgrass, line 23: Why does this call downgrade?  Probably worth throwing a comment into the code here explaining the logic.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/ConfiguredAirbyteCatalogMigration.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) 2022 Airbyte, Inc., all rights reserved.
+ */
+
+package io.airbyte.commons.protocol.migrations;
+
+public interface ConfiguredAirbyteCatalogMigration<PreviousVersion, CurrentVersion> extends Migration {
+
+  /**
+   * Downgrades a ConfiguredAirbyteCatalog to from the new version to the old version
```
colesnodgrass, line 10: typo in this javadoc, the first `to` doesn't apply

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/AirbyteProtocolVersionedMigratorFactory.java
gosusnp: Given the current structure, we inject the factory so that the worker can get a version dependent instance. Because this is per request, I'd expect a bigger refactoring to be needed for this.
Trying to limit complexity of the changes at the moment.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/MigrationContainer.java
```diff
@@ -0,0 +1,101 @@
...
+    return new Version(mostRecentMajorVersion, "0", "0");
+  }
+
+  /**
+   * Downgrade a message from the most recent version to the target version by chaining all the
+   * required migrations
+   */
+  public <PreviousVersion, CurrentVersion> PreviousVersion downgrade(final CurrentVersion message,
+                                                                     final Version target,
+                                                                     final BiFunction<T, Object, Object> applyDowngrade) {
```
colesnodgrass, line 39: Could this be `BiFunction<T, PreviousVersion, CurrentVersion> applyDowngrade`?

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/MigrationContainer.java
```diff
@@ -0,0 +1,101 @@
...
+    if (target.getMajorVersion().equals(mostRecentMajorVersion)) {
+      return (PreviousVersion) message;
+    }
+
+    Object result = message;
```
colesnodgrass, line 44: Could this be defined as `PreviousVersion results`?

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/MigrationContainer.java
colesnodgrass: Same question applies to the next method as well.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/MigrationContainer.java
```diff
@@ -0,0 +1,101 @@
...
+      migrations.put(key, migration);
+      if (migration.getCurrentVersion().getMajorVersion().compareTo(mostRecentMajorVersion) > 0) {
+        mostRecentMajorVersion = migration.getCurrentVersion().getMajorVersion();
+      }
+    } else {
+      throw new RuntimeException("Trying to register a duplicated migration " + migration.getClass().getName());
+    }
+  }
+
+  public Set<String> getMigrationKeys() {
```
colesnodgrass, line 97: Should this be `public` or `@VisibleForTesting`?

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/AirbyteMessageMigrator.java
```diff
@@ -25,104 +24,59 @@
...
+  private final MigrationContainer<AirbyteMessageMigration<?, ?>> migrationContainer;
 
-  public AirbyteMessageMigrator(List<AirbyteMessageMigration<?, ?>> migrations) {
-    migrationsToRegister = migrations;
-  }
-
-  public AirbyteMessageMigrator() {
-    this(Collections.emptyList());
+  public AirbyteMessageMigrator(final List<AirbyteMessageMigration<?, ?>> migrations) {
+    migrationContainer = new MigrationContainer<>(migrations);
```
colesnodgrass, line 30: Should the `MigrationContainer` be injected here instead of the `List<>`?

gosusnp: I don't think so, the intent is to inject the list of migrations. We have different migrations intended for the `AirbyteMessageMigrator` or the `ConfiguredAirbyteCatalogMigrator`.
The `MigrationContainer<>` is a shared container for abstracting the logic of how we fetch the migrations. Injecting the MigrationContainer would involve an extra layer of subclassing which looks more like overhead than solving a problem at the moment.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/AirbyteProtocolVersionedMigratorFactory.java
gosusnp: removed.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/ConfiguredAirbyteCatalogMigrator.java
gosusnp: I don't think so, current design enforces a one-way look up from/to the current version. If the version provided isn't a valid downgrade or upgrade, we either fail because the version isn't supported (in the list of known migrations) or a mismatch in the type conversion.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/ConfiguredAirbyteCatalogMigrator.java
gosusnp: answered above.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/ConfiguredAirbyteCatalogMigrator.java
gosusnp: I was not able to make it work because of the `MigrationContainer`.
More specifically, eventually, we apply a chain of transforms `f(A -> B -> C)` which is falling appart when passing in `final BiFunction<T, Object, Object> applyDowngrade`. Type end up being casted dynamically from object because of this.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/VersionedProtocolSerializer.java
gosusnp: Indeed, added some docs to the `VersionedProtocolSerializer`

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/ConfiguredAirbyteCatalogMigration.java
gosusnp: fixed

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/MigrationContainer.java
gosusnp: Similar answer as a previous question, reposting here for context:

We apply a chain of transforms f(A -> B -> C) which is falling appart when passing in final BiFunction<T, Object, Object> applyDowngrade. Type end up being casted dynamically from object because of this.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/MigrationContainer.java
gosusnp: No because of the chaining of transforms, if we end up having more than one migration, the type of `results` will change.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/MigrationContainer.java
gosusnp: This should be public, once the bootloader is migrated to micronaut, we should use this to check that we actually have all the migrations required to support the range of protocol versions from the config.

colesnodgrass (approved):

benmoriceau (approved):

PR #20747, opened 2022-12-21
[Low-Code CDK] Parse incoming manifest objects into Pydantic models

### LINKED ISSUE #20045: [Low-Code CDK] Parse incoming connector manifest as Pydantic objects
Low Code Schema Refactor Phase 3

Now that the Pydantic model classes for the low-code language are generated from the component schema, incoming YAML manifests should be parsable into those models.

After parsing a manifest into Mapping and dereferencing components, we should then be able to generate the Pydantic models from the incoming schema. This would also double as manifest schema validation because objects would not be parsable if fields were the incorrect type or required fields were missing.

## Implementation Details
At the very simplest, we'll want to use `ComponentModel.parse_obj()` and leverage how it will recursively build pydantic objects. We don't want to integrate this into the parsing flow just yet, but we can define temporary method `create_models()` which will take in the referenced components and emit Pydantic models for testing.

A few things to note during implementation
- Before performing the Pydantic parse, we'll need to pre-parse the object to propagate `$options` downward to child components.
- Pydantic uses a form of recursive duck typing to determine how to convert an object into it's respective model. The `parse_obj()` does the heavy lifting, but we should be careful to verify the correct objects are being made especially with unit tests
- Something to fleshed out a little is whether we should rely on Pydantic's duck typing or if we want to continue to use the default implementation registry and append that to options wherever relevant.

## Acceptance Criteria

- [ ] Valid YAML configs result in a set of Pydantic models where the top most parent is the Declarative Source
- [ ] Invalid YAML configs fail while trying to translate the connector manifest into Pydantic models

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/20045

## What
Now that we have Pydantic model classes generated from the declarative_component_schema.yaml, we need to translate the incoming manifest object into those models. This is a precursor to constructing declarative components from these models (Phase 4).

Note: This is intentionally gated off and won't impact anything since we'll need to have he full object creation complete before we can release this

## How
We can leverage Pydantic's existing parser to transform a Mapping into it's respective Pydantic model and subcomponent models. The method to do that is `BaseModel.parse_obj()`.

Without turning these models into concrete declarative components, this code path is not usable so I've also gated it behind a field on the ManifestDeclarativeSource called `construct_using_pydantic_models`, which will remain off at all times until we are ready to release the changes.

Lastly, I added some boilerplate as to what the new factory will be structured and added construction of `CheckStream` and `Spec` components because they are very simple. I didn't include `DeclarativeStream` and just created a stub because that would require building out the entire set of constructors which will be in Phase 4.

## Recommended reading order


airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -46,18 +50,28 @@ class ManifestDeclarativeSource(DeclarativeSource):
...
+
         evaluated_manifest = {}
-        resolved_source_config = ManifestReferenceResolver().preprocess_manifest(source_config, evaluated_manifest, "")
-        self._source_config = resolved_source_config
+        resolved_source_config = ManifestReferenceResolver().preprocess_manifest(manifest, evaluated_manifest, "")
+        propagated_source_config = ManifestComponentTransformer().propagate_types_and_options("", resolved_source_config, {})
+        self._source_config = propagated_source_config
         self._debug = debug
-        self._factory = DeclarativeComponentFactory()
+        self._factory = DeclarativeComponentFactory()  # Legacy factory used to instantiate declarative components from the manifest
```
girarda, line 73: nit: rename to `_legacy_factory` if this will get merged in with the constructor  to avoid confusion

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -123,38 +150,33 @@ def _configure_logger_level(self, logger: logging.Logger):
...
+        if self.construct_using_pydantic_models:
+            # Validates the connector manifest against the low-code component schema defined in declarative_component_schema.yaml
+            try:
+                raw_component_schema = pkgutil.get_data("airbyte_cdk", "sources/declarative/declarative_component_schema.yaml")
+                declarative_component_schema = yaml.load(raw_component_schema, Loader=yaml.SafeLoader)
+            except FileNotFoundError as e:
+                raise FileNotFoundError(f"Failed to read manifest component json schema required for validation: {e}")
+
+            try:
+                validate(self._source_config, declarative_component_schema)
```
girarda, line 162: nit: we should validate using the legacy validation until we delete the code (unless we're 110% confident they're equivalent)

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -0,0 +1,58 @@
...
+
+def create_spec(model: SpecModel, config: Config):
+    return Spec(connection_specification=model.connection_specification, documentation_url=model.documentation_url, options={})
+
+
+class ModelToComponentFactory:
+    def create_component(self, model_type: Type[BaseModel], component_definition: ComponentDefinition, config: Config) -> type:
+        component_type = component_definition.get("type")
+        if component_definition.get("type") != model_type.__name__:
+            raise ValueError(f"Expected manifest component of type {model_type}, but received {component_type} instead")
```
girarda, line 37: want to make sure I understand what could cause this error... is this for the case where the parameters match a different type?
eg
```
authenticator:
  type: "BasicHttpAuthenticator"
  api_token: "{{ config['api_key'] }}"
```
where BasicHttpAuthenticator expects a username, not an api_key

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -123,38 +150,33 @@ def _configure_logger_level(self, logger: logging.Logger):
...
+            except ValidationError as e:
+                raise ValidationError("Validation against json schema defined in low_code_component_schema.yaml schema failed") from e
+        else:
+            # Validates the connector manifest against the one schema auto-generated from the component dataclasses
+            full_config = {}
+            if "version" in self._source_config:
+                full_config["version"] = self._source_config["version"]
+            full_config["check"] = self._source_config["check"]
+            streams = [self._factory.create_component(stream_config, {}, False)() for stream_config in self._stream_configs()]
+            if len(streams) > 0:
```
girarda, line 172: question: what should we do / how should we fail if there are no streams?

girarda (commented):
me gusta!

thanks for sharing an early draft. it really helps understanding the full picture.

This seems like a good approach assuming we solve the problem of the fields that need to be duplicated. Maybe worth validating early to make sure we don't need to revisit this solution

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -46,18 +50,28 @@ class ManifestDeclarativeSource(DeclarativeSource):
 
     VALID_TOP_LEVEL_FIELDS = {"check", "definitions", "schemas", "spec", "streams", "type", "version"}
 
-    def __init__(self, source_config: ConnectionDefinition, debug: bool = False):
+    def __init__(self, source_config: ConnectionDefinition, debug: bool = False, construct_using_pydantic_models: bool = False):
```
maxi297, line 53: I'm wondering if it would have made sense to use a strategy pattern. My reasoning is that we see doing a migration generates a lot of change in this file. This is code that we will need to clean eventually and that will probably reappear in some for for every migration we will do. Having a strategy pattern allows us to not care about how the objects are created and therefore this class would stay the same whatever the migration is. 

So right now, one difference I see is that `ManifestDeclarativeSource` now pass as an argument the object it wants created. I think we could deduce it using the `type` in the pydantic model

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
maxi297: Why is that? I'm wondering because I feel like the earlier we enable this validation, the more feedback we will get to know if it behaves as we expect?

maxi297 (commented):
I like the way you structured this change. Good job!

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
brianjlai: We did validate it in part 1, but its trivial to leave it in and doesn't hurt to so I'll remove the conditional here, but keep the rest to gate construction.

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
brianjlai: Good question. In the context of this code, we're just appending the constructed declarative streams back to the object for validation. We'll end up deleting this code in the long run though so I don't think we need to change anything here.

That being said, we should be able to update the json schema to enforce a minimum of 1 element and avoid having a specific conditional check in our code.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: Not quite. So this is a pretty niche case because the `create_component()` will only serve as the entrypoint for a few different invocations by ManifestDeclarativeSource (CheckStream, DeclarativeStream, and Spec components). But if somehow a schema had:
```
check:
  type: SimpleRetriever
  stream_names: ["applications"]
```

then we would want to fail this. This should be caught by any form of schema validation we do, so I'm also okay w/ just getting rid of this extra check too

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
brianjlai: Annoyingly, it looks like there's a bug in pydantic that causes an error when we try to use field constraints on generics. 
```
E   ValueError: On field "streams" the following field constraints are set but not enforced: min_items. 
E   For more details see https://pydantic-docs.helpmanual.io/usage/schema/#unenforced-field-constraints
```

There's a bug filed here w/ the exact use case, in the meantime I'll have to get rid of the field constraint because the way the model generator works, it will always emit classes in the format that causes the error until it is fixed:
https://github.com/pydantic/pydantic/issues/3358

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -128,7 +157,7 @@ def _validate_source(self):
         if "version" in self._source_config:
             full_config["version"] = self._source_config["version"]
         full_config["check"] = self._source_config["check"]
-        streams = [self._factory.create_component(stream_config, {}, False)() for stream_config in self._stream_configs()]
+        streams = [self._legacy_factory.create_component(stream_config, {}, False)() for stream_config in self._stream_configs()]
```
clnoll, line 160: Should we be flagging this on `self.construct_using_pydantic_models` so that we're validating the streams constructed by `self._constructor.create_component`? Or is blocked on implementing `create_declarative_stream` (in which case maybe a TODO here would be warranted?)?

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
clnoll: If we keep it in there should probably be a test for it. But looks like `ValidationError` will be thrown before we end up hitting the codepath that would give `ValueError`, in which case I'd be in favor of removing this check.

clnoll (commented):
Awesome @brianjlai! Just a couple questions/comments.

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
brianjlai: This is just a no-op for renaming the variable. I don't think we need to gate the validation behind anything because in the first part of the refactor we decided to validate the schema against both the auto-generated one (which is this block of code) as well as against the YAML handwritten one.  This should already be working as is, and the pydantic parsing we're doing in this PR is done after validations as we actually instantiate components.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: Yep good suggestion, I'll add some basic tests for this flow, but a majority of the tests will be construction related in the follow on PR.

brianjlai:
> This seems like a good approach assuming we solve the problem of the fields that need to be duplicated. Maybe worth validating early to make sure we don't need to revisit this solution

This is a good idea. Catherine and I have been working on the follow up to this PR where we actually create the declarative components and I validated that we can solve the duplicative fields as we want. For example, if we want to avoid duplicating base_urls or primary keys, between layers, at a parent layer we have access to the current model and sub-components and we can get specific fields and pass them to children via kwargs. I verified this for base_url

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
brianjlai: Sorry I might not be understanding exactly what you mean in how we might use the strategy pattern. This file should not undergo very many changes during the code its basically just the boolean to say use the old or new factory. Do you have a bit more of a tangible example to help me understand what you're suggesting better.

Once we finish this refactor in the next PR we would want to just remove the legacy factory in favor of the `model_to_component_factory.py` entirely so I'm not sure if we need to necessarily go the route of inserting this pattern for what should be short lived code.

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -46,37 +50,62 @@ class ManifestDeclarativeSource(DeclarativeSource):
...
+        if "type" not in manifest:
+            manifest["type"] = "DeclarativeSource"
+
         evaluated_manifest = {}
-        resolved_source_config = ManifestReferenceResolver().preprocess_manifest(source_config, evaluated_manifest, "")
-        self._source_config = resolved_source_config
+        resolved_source_config = ManifestReferenceResolver().preprocess_manifest(manifest, evaluated_manifest, "")
+        propagated_source_config = ManifestComponentTransformer().propagate_types_and_options("", resolved_source_config, {})
+        self._new_source_config = propagated_source_config
+        self._legacy_source_config = resolved_source_config
```
brianjlai, line 72: In the immediate, it's a little ugly to retain both the legacy manifest and the new manifest (the key difference being the new manifest goes through the transformer to propagate options and assign types). But we will be getting rid of `_legacy_source_config` in the final part of the refactor anyway. Doing it this way just helps make it very explicit what lines of code will get deleted in that PR.

The final flow after we finish will look like this:
```
evaluated_manifest = {}
resolved_source_config = ManifestReferenceResolver().preprocess_manifest(manifest, evaluated_manifest, "")
propagated_source_config = ManifestComponentTransformer().propagate_types_and_options("", resolved_source_config, {})
self._source_config = propagated_source_config
self._debug = debug
self._constructor = ModelToComponentFactory() # New factory which converts the manifest to Pydantic models to construct components
        
self._validate_source()
```

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -46,37 +50,62 @@ class ManifestDeclarativeSource(DeclarativeSource):
...
         self._debug = debug
-        self._factory = DeclarativeComponentFactory()
+        self._legacy_factory = DeclarativeComponentFactory()  # Legacy factory used to instantiate declarative components from the manifest
+        self._constructor = ModelToComponentFactory()  # New factory which converts the manifest to Pydantic models to construct components
 
         self._validate_source()
 
         # Stopgap to protect the top-level namespace until it's validated through the schema
-        unknown_fields = [key for key in self._source_config.keys() if key not in self.VALID_TOP_LEVEL_FIELDS]
+        unknown_fields = [key for key in self._legacy_source_config.keys() if key not in self.VALID_TOP_LEVEL_FIELDS]
```
girarda, line 80: is this check needed or would it be enough to validate the `ConnectorManifest` with additionalProperties = false? would be nice to only define the top level fields once

brianjlai: Yep I will add it to the declarative handwritten schema. But while we are keeping the old validation flow, we should leave this block. But during cleanup we will remove this. Will update this w/ a comment

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
maxi297: We would pass the strategy at the construction of the object like this:
```
class ManifestDeclarativeSource(DeclarativeSource):
    def __init__(self, source_config: ConnectionDefinition, debug: bool = False, models_construction_strategy: ModelConstructionStrategy):
        self.models_construction_strategy = models_construction_strategy
```

Now, everywhere where we need a model objects, we don't need the `if self.construct_using_pydantic_models:`. For example, `connection_checker` becomes:
```
    @property
    def connection_checker(self) -> ConnectionChecker:
        check = self._config["check"]
        if "type" not in check:
            check["type"] = "CheckStream"
        return self.models_construction_strategy.create_check_stream(CheckStreamModel, check, dict())(source=self)
```

This is interesting because:
* it allows `ManifestDeclarativeSource` to be decoupled from the implementation details. It doesn't have to care if it's pydantic, Django or even custom code. So when we have a new way of constructing objects, we can just pass the right strategy to `ManifestDeclarativeSource` and no need to add `if self.construct_using_pydantic_models` everywhere we construct objects. This means that the unit tests do not change. Sure we will probably add a couple of integration tests but those should be less
* it simplifies the migration. Instead of having to go back to all the `if self.construct_using_pydantic_models`, we can simply pass the other strategy to the constructor and that's it

Maybe this is overkill for here, I'm not so sure. I don't fully understand the whole context so don't hesitate to continue with the approach you were taking.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: i added the test that covers this codepath

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
brianjlai: Ah okay that makes sense, so each strategy adheres to a common interface, but depending on the strategy's method implementation that dictates which flow gets invoked. Thanks for clarifying.

Thinking about it a bit, if we were planning to support both flows or long term expect more component creation strategies I think this would be a good fit. But the plan is to roll over to the new flow in the other in flight PR so we'll end up with a single strategy very soon. Since just one strategy for one type might not be quite as beneficial, I think I'll leave this as is. But I'm gonna remember this idea in case we can use it in the future

girarda (approved):
🚢

brianjlai:
note: I won't be cutting a release of this because the changes are gated anyway, I will release this as part of https://github.com/airbytehq/airbyte/pull/21050

PR #20748, opened 2022-12-21
[Connector Builder] Add stream slicer

lmossman:
## What
Adds stream slicer to the connector builder, in a similar approach as to what was taken to add paginator in: https://github.com/airbytehq/airbyte/pull/20698


flash1293:
Screenshot 2022-12-21 at 15 01 14

Similar situation as with the pagination constraints - slices can't be injected into the path

flash1293:
Do we still need a cursor field on the top level with the cursor field on the slicer? My understanding based on the slack discussion was that the top level cursor does not really make sense (and especially it doesn't make sense to have different cursor fields on the top level and the slicer):
Screenshot 2022-12-21 at 15 05 21

flash1293:
Seems like `step` is failing if it's not passed something that can be parsed correctly by this method: https://github.com/airbytehq/airbyte/blob/50f22cd5d702ad7fe8abf1ea8a1077417e5e8254/airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py#L188

Screenshot 2022-12-21 at 15 10 06

We could enforce the pattern defined here in the UI as well (as it's anyway not possible to use interpolated values in the step size): https://github.com/airbytehq/airbyte/blob/50f22cd5d702ad7fe8abf1ea8a1077417e5e8254/airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/datetime_stream_slicer.py#L75

flash1293:
In general, it seems like we can't rule out that the list request will error out because something isn't configured quite right. We should show the inputs form even if that happens in case that's related to the problem (can definitely happen on a separate PR)

flash1293:
> here is currently an issue where if you set the Field name on the stream slicer to something without changing the value of Inject into, it gives me the error
> I might just be something that we need to add a useEffect with an empty dependency array to set it on the values on creation like in the oneOf.

As it's a very similar case you could also use the `onChange` I introduced to the BuilderField for the inputs form and set inject_to this way (an effect hook will only run after the next render, so it will cause the error, then fix it a split second later). But in general I like the idea with the checkbox better for these request options (this way we can bring them into a valid default state.

lmossman:
> In general, it seems like we can't rule out that the list request will error out because something isn't configured quite right. We should show the inputs form even if that happens in case that's related to the problem (can definitely happen on a separate PR)

@flash1293 I didn't have time to make this change. Do you mind adding that to our feature/connector-builder branch if you have a chance while I am out?

flash1293 (approved):
Tested and works well, LGTM

PR #20756, opened 2022-12-21
Highlight invalid cursor or primary key in new connection streams table

m19berger:
## What
Closes [#19918](https://github.com/airbytehq/airbyte/issues/19918)

## How
Added error state for required fields for new connection streams table

_For test purposes, use the environment variable:_ REACT_APP_NEW_STREAMS_TABLE=true

## Loom
https://www.loom.com/share/4c8c6dac261d48728bbd2803315dcbcb

### DIFF
--- a/airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
@@ -166,7 +166,12 @@ const CatalogSectionInner: React.FC<CatalogSectionInnerProps> = ({
   );
 
   const destName = prefix + (streamNode.stream?.name ?? "");
-  const configErrors = getIn(errors, `schema.streams[${streamNode.id}].config`);
+  const configErrors = getIn(
+    errors,
+    isNewStreamsTableEnabled
+      ? `syncCatalog.streams[${streamNode.id}].config`
+      : `schema.streams[${streamNode.id}].config`
+  );
   const hasError = configErrors && Object.keys(configErrors).length > 0;
   const pkType = getPathType(pkRequired, shouldDefinePk);
   const cursorType = getPathType(cursorRequired, shouldDefineCursor);
@@ -194,6 +199,7 @@ const CatalogSectionInner: React.FC<CatalogSectionInnerProps> = ({
         onExpand={onExpand}
         changedSelected={changedSelected}
         hasError={hasError}
+        configErrors={configErrors}
         disabled={disabled}
       />
       {isRowExpanded &&
--- a/airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
@@ -194,7 +194,6 @@ const CatalogSectionInner: React.FC<CatalogSectionInnerProps> = ({
         onExpand={onExpand}
         changedSelected={changedSelected}
         hasError={hasError}
-        configErrors={configErrors}
         disabled={disabled}
       />
       {isRowExpanded &&
--- a/airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
@@ -194,6 +194,7 @@ const CatalogSectionInner: React.FC<CatalogSectionInnerProps> = ({
         onExpand={onExpand}
         changedSelected={changedSelected}
         hasError={hasError}
+        configErrors={configErrors}
         disabled={disabled}
       />
       {isRowExpanded &&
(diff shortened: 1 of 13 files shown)


edmundito:
@matter-q The storybook failed to build

edmundito:
@matter-q While testing, I noticed an edge case that I didn't realize when creating the original ticket. You can use a hubspot source to reproduce this:

1. Use bulk edit to select all the streams
2. Toggle the streams to be enabled, select "Incremental | Deduped + History"


Notice that the sync mode is empty in one of the rows and should also be highlighted red:

image

Does it make sense to fix it in this PR?

airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
```diff
@@ -166,7 +166,12 @@ const CatalogSectionInner: React.FC<CatalogSectionInnerProps> = ({
   );
 
   const destName = prefix + (streamNode.stream?.name ?? "");
-  const configErrors = getIn(errors, `schema.streams[${streamNode.id}].config`);
+  const configErrors = getIn(
+    errors,
+    isNewStreamsTableEnabled
+      ? `syncCatalog.streams[${streamNode.id}].config`
+      : `schema.streams[${streamNode.id}].config`
```
edmundito, line 173: I wonder why these are different. They should be the same 🤔

airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionReplicationTab.tsx
```diff
@@ -203,18 +203,20 @@ export const ConnectionReplicationTab: React.FC = () => {
...
-                    dirty={dirty}
-                    resetForm={async () => {
-                      resetForm();
-                      discardRefreshedSchema();
-                    }}
-                    successMessage={saved && !dirty && <FormattedMessage id="form.changesSaved" />}
-                    errorMessage={getErrorMessage(isValid, dirty)}
-                    enableControls={schemaHasBeenRefreshed || dirty}
-                  />
+                  <div>
```
edmundito, line 206: Why does this need to be inside a div? should the div be part of EditControls?

airbyte-webapp/src/views/Connection/ConnectionForm/formConfig.tsx
```diff
@@ -90,8 +90,109 @@ export const createConnectionValidationSchema = ({
   mode,
   allowSubOneHourCronExpressions,
   allowAutoDetectSchema,
-}: CreateConnectionValidationSchemaArgs) =>
-  yup
+}: CreateConnectionValidationSchemaArgs) => {
+  const isNewStreamsTableEnabled = process.env.REACT_APP_NEW_STREAMS_TABLE ?? false;
+
+  if (isNewStreamsTableEnabled) {
```
edmundito, line 96: How come this is duplicated, what is the different between the two?

edmundito (commented):
I had a few questions about the code. Tested locally and the functionality is good.

matter-q:
> @matter-q While testing, I noticed an edge case that I didn't realize when creating the original ticket. You can use a hubspot source to reproduce this:
> 
> 1. Use bulk edit to select all the streams
> 2. Toggle the streams to be enabled, select "Incremental | Deduped + History"
> 
> Notice that the sync mode is empty in one of the rows and should also be highlighted red:
> 
> image
> 
> Does it make sense to fix it in this PR?

I think it's better to fix in this PR

airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
matter-q: As I understand correctly documentation, we used a custom format for errors.

image

In my case, I have an empty error object with this path `schema.streams[${streamNode.id}].config` due to different logic for error handling (yup way). Regarding yup schema, I have the correct error object with this path `syncCatalog.streams[${streamNode.id}].config`

image

airbyte-webapp/src/views/Connection/ConnectionForm/formConfig.tsx
matter-q: There are different approaches to validating the form 

The new implementation (we have independent rules for each config value):
image

Old implementation (with this implementation, we can't understand where is an error in the config):
image

edmundito: Got it. One way to avoid this duplication is by declaring config first based on which schema we are using, then assigning to the object that is returned

airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionReplicationTab.tsx
matter-q: Sorry! Forgot to remove it after some local tests

airbyte-webapp/src/views/Connection/ConnectionForm/formConfig.tsx
matter-q: Fixed by a conditional operator inside the yup object

edmundito:
Followed up over standup - The sync mode issue will be skipped for now and resolved separately as part of the larger bulk edit mode behavior

edmundito (approved):
Tested locally with new and old table. Code looks good. 🚢

PR #20759, opened 2022-12-21
Use explicit configuration for Micronaut endpoints

jdpgrailsdev:
## What
* Use explicit configuration for built-in Micronaut endpoints

## How
* Remove `intercept-url-map`
* Explicitly list built-in endpoints
* Add anonymous security annotation to health/heartbeat endpoints

## Recommended reading order
1. `application.yml`

## Tests

* Ran locally and verified that the correct endpoints are reachable


colesnodgrass (approved):

airbyte-container-orchestrator/src/main/resources/application.yml
```diff
@@ -3,4 +3,30 @@ micronaut:
...
\ No newline at end of file
+  config-dir: /config
+
+endpoints:
+  beans:
+    enabled: true
+    sensitive: false
+  env:
+    enabled: true
+    sensitive: false
```
gosusnp, line 14: What is sensitive in this context?

jdpgrailsdev: @gosusnp Sorry, I missed this context.  The `sensitive` keyword controls whether or not you need to authenticate in order to view the endpoint.  That does not impact whether or not Micronaut obfuscates values for properties, which it does regardless of whether or not you require authentication.

gosusnp: In this case, shouldn't `env` and `beans` be declared as sensitive?

benmoriceau (approved):

airbyte-container-orchestrator/src/main/resources/application.yml
jdpgrailsdev: @gosusnp Maybe.  While the `/env` endpoint will show all of the environment variables and properties, it will mask those that are sensitive.  If we don't want the names of the properties being seen unless authenticated, then we can mark it as sensitive.  I think I marked them as not sensitive because this is an open source project and you can see all of the environment variables and properties already -- there is nothing hidden.  As long as the assigned value is obfuscated, there is little risk.  Happy to change it though if anyone feels strongly.

gosusnp: I am not feeling super strongly about this, in practice, none of orchestrator/cron/workers should be exposed to the outside directly. However, for the same reason we enabled auth by default on the frontend, I'd say we may want to restrict access to endpoints that exposes config by default.
How does the authentication work in this context?

jdpgrailsdev: @gosusnp We would need to do the same thing:  expose the port, enable auth for any HTTP calls to that port and then enable Micronaut to verify that the expected auth has been provided.  Not a lot of work to do and something we can definitely add.  I agree that since the port is not exposed explicitly and because any sensitive information is already obfuscated, the risk is low here.

gosusnp (approved):

PR #20762, opened 2022-12-21
Remove "Transfer" title from card in replication settings when new table is enabled

### LINKED ISSUE #20452: Remove "Transfer" card heading on ConnectionFormFields
From new table design review.

(We discussed this change with regards to the Replication tab, confirm whether it should also be removed on the Create form as it shares this component)

This should be hidden behind the new table feature flag.

edmundito:
## What
Resolves #20452 

Removes the "Transfer" Title when the new table is enabled so that it matches the design:
image


krishnaglick (approved):
LGTM

PR #20765, opened 2022-12-21
Source Mailchimp - fix the way request params are built

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1058

## How
 - Do not use state as a source for request parameters. Use stream slices instead
 - Set test strictness level to high


bazarnov (approved):

airbyte-integrations/connectors/source-mailchimp/source_mailchimp/streams.py
```diff
@@ -110,13 +126,30 @@ def path(self, **kwargs) -> str:
...
+    ) -> Iterable[Optional[Mapping[str, Any]]]:
+        stream_state = stream_state or {}
+        if self.campaign_id:
+            # this is a workaround to speed up SATs and enable incremental tests
+            campaigns = [{"id": self.campaign_id}]
+        else:
+            campaigns = Campaigns(authenticator=self.authenticator).read_records(sync_mode=SyncMode.full_refresh)
+        for campaign in campaigns:
+            slice_ = {"campaign_id": campaign["id"]}
+            cursor_value = stream_state.get(campaign["id"], {}).get(self.cursor_field)
```
murphpdx, line 149: It looks like `cursor_field` is set to timestamp. I know you have a sort_field of create_time but Mailchimp does not allow you to change the sorting. That means that you're going to lose a lot of records.

davydov-d: @murphpdx as for now there's just a single slice per campaign, this means we request all the data after the `since` date from the API and dont actually care if there is sorting or not. Can you confirm the big still persists with you after upgrading to the latest version?

murphpdx: You're not paginating?

murphpdx: We decided not to use airbyte last summer due in part to this bug. So I don't have the time to test it out for you.

davydov-d: Yes we are paginating and the since param remains the same throughout all the pages within one campaign. The thing is you are pointing to the `stream_slices` method which yields one slice per campaign and the pagination is made within one slice.

PR #20774, opened 2022-12-21
Add cancelled status icon

### LINKED ISSUE #19630: Cancellation should not be red
Screen Shot 2022-11-18 at 4 18 12 PM
* When an attempt is cancelled, the UI shows a red X and a failure reason. This implies that the sync is broken in some way. This is misleading.

* When an attempt is cancelled the symbol should be gray with a more neutral symbol (perhaps a dash), and we should not show the failure reason line. There should be no red.

@timroes is the workflows team the right place to send this?

edmundito:
## What
Resolves: #19630 

Add cancelled status icon and shows the icon in the connection list and jobs page:
image

image

## How
Adds new icon to StatusIcon
Updates conditions to render cancelled separated from failure
Extra: Updates the fail icon to use the cross icon instead of font awesome

## Recommended reading order
Start with `airbyte-webapp/src/components/ui/StatusIcon/StatusIcon.tsx`


airbyte-webapp/src/components/EntityTable/types.ts
```diff
@@ -32,10 +32,11 @@ interface ITableDataItem {
   entityIcon?: string;
 }
 
-enum Status {
+const enum Status {
```
krishnaglick, line 35: Today I learned! Cool!

krishnaglick (commented):
Need to test locally, code LGTM

krishnaglick (approved):

cgardens:
Screen Shot 2023-01-03 at 12 51 26 PM

@edmundito do we still need the red "Failure Origin" and all that text. It was cancelled, not a failure.

PR #20775, opened 2022-12-21
Randomize namespace to avoid false successes

### LINKED ISSUE #17105: DATs should verify non-default namespace behavior
from https://docs.google.com/document/d/15xLW97ZqPnttLluP80ZCqXxCHHzB-vn8bhBBpUmjGMI/edit#

Apparently it's possible for a destination to only write data into its default namespace and still pass our DATs. We should add a test case that verifies it can write to a non-default namespace. I.e. if the catalog looks like:
```json
{
  "streams": [
    {
      "stream": {
        "name": "whatever",
        "namespace": "some_non_default_namespace",
        "json_schema": {...}
      }
    }
  ]
}
```

Then the DATs should verify that after a sync, a table called `some_non_default_namespace.whatever` was created.

## Details

It looks like this test case already exists - https://github.com/airbytehq/airbyte/blob/09154f3351fadcaa799edb29169d09e3daed8f0a/airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java#L950

Let's look into why it wasn't failing prior to https://github.com/airbytehq/airbyte/pull/17054, and fix that. (A likely avenue of investigation: should this [`sourcenamespace`](https://github.com/airbytehq/airbyte/blob/09154f3351fadcaa799edb29169d09e3daed8f0a/airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java#L960) be randomized per test run?)

In order to run `BigQueryGcsDestinationAcceptanceTest` locally, you'll need to manually create `airbyte-integrations/connectors/destination-bigquery/secrets/credentials.json`, using the contents [here](https://console.cloud.google.com/security/secret-manager/secret/SECRET_DESTINATION-BIGQUERY_CREDENTIALS__CREDS/versions?project=dataline-integration-testing).

And you can explore our integration test BigQuery instance [here](https://console.cloud.google.com/bigquery?project=dataline-integration-testing&ws=!1m4!1m3!3m2!1sdataline-integration-testing!2ssourcenamespace).

jbfbell:
## What
In #17105 a previously created test was passing but should have been catching a bug related to overriding the default namespace. Subsequent runs of this test would always succeed because the data had already been created by previous runs and was not cleaned up.

## How
Adding a random suffix to the `datasetId` for bigquery should ensure that the data loaded and retrieved from bigquery is the same.

### DIFF
--- a/airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java
+++ b/airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java
@@ -987,7 +987,8 @@ public abstract class DestinationAcceptanceTest {
         Jsons.deserialize(
             MoreResources.readResource(DataArgumentsProvider.EXCHANGE_RATE_CONFIG.getCatalogFileVersion(getProtocolVersion())),
             AirbyteCatalog.class);
-    // A randomized namespace is required otherwise you can generate a "false success" with data from a previous run.
+    // A randomized namespace is required otherwise you can generate a "false success" with data from a
+    // previous run.
     final String namespace = Strings.addRandomSuffix("airbyte_source_namespace", "_", 8);
 
     catalog.getStreams().forEach(stream -> stream.setNamespace(namespace));
--- a/airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java
+++ b/airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java
@@ -23,6 +23,7 @@ import io.airbyte.commons.jackson.MoreMappers;
 import io.airbyte.commons.json.Jsons;
 import io.airbyte.commons.lang.Exceptions;
 import io.airbyte.commons.resources.MoreResources;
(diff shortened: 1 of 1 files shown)


airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java
```diff
@@ -986,7 +987,10 @@ void testSyncUsesAirbyteStreamNamespaceIfNotNull() throws Exception {
         Jsons.deserialize(
             MoreResources.readResource(DataArgumentsProvider.EXCHANGE_RATE_CONFIG.getCatalogFileVersion(getProtocolVersion())),
             AirbyteCatalog.class);
-    final String namespace = "sourcenamespace";
+    // A randomized namespace is required otherwise you can generate a "false success" with data from a
+    // previous run.
+    final String namespace = Strings.addRandomSuffix("airbyte_source_namespace", "_", 8);
```
edgao, line 992: do we have cleanup logic for these randomized datasets? Or is the plan to just add a TTL on the tables (or some other way to clean them out eventually)

(it looks like `tearDownBigQuery` only deletes the default dataset, so these `airbyte_source_namespace_*` won't get deleted)

edgao (approved):
looks good! One question about the table cleanup, but otherwise I think we're good to ship (post code freeze)

:shipit: assuming PR checks turn green

PR #20779, opened 2022-12-21
Avoid prematurely ending a destination

### LINKED ISSUE #17926: Destination should not be SIGTERM-ed if source fails
## Current Behavior
If the source connector container fails (e.g. due to OOMKill), the destination connector container is SIGTERM-ed 60 seconds later. Example in https://cloud.airbyte.io/workspaces/94f992c9-be44-4e0f-8194-dfcac73b88a2/connections/6e1f6af6-cceb-4dd6-a2e7-c10f5529e983/status#696389::0:
* `2022-10-12 15:44:12 source > Killed`
* `2022-10-12 15:45:13 replication-orchestrator > (pod: jobs / destination-bigquery-write-696389-0-uhdzm) - Destroying Kube process.` (followed by a stacktrace `Suppressed: io.airbyte.workers.exception.WorkerException: Destination process exit with code 143. This warning is normal if the job was cancelled.`)

This SIGTERM prevented the destination from committing data to the `_raw` tables. This then caused normalization to also fail (because this was the first sync, and so the `_raw` tables didn't exist at all).

This also means we're not committing a new state message, so we're wasting effort. (it does mean that we're not losing data at least, since the next sync will just restart those streams from scratch / previous state)

## Expected Behavior
The destination connector should be allowed to run long enough to commit all the data the source produced, even if the source crashes.

gosusnp:
## What

Closes airbytehq/airbyte-internal-issues#1038

### DIFF
--- a/airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
+++ b/airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
@@ -194,6 +194,7 @@ public class DefaultReplicationWorker implements ReplicationWorker {
           readFromDstRunnable(destination, cancelled, messageTracker, mdc, timeTracker),
           executors)
           .whenComplete((msg, ex) -> {
+            LOGGER.info("Read from destination thread complete.");
             if (ex != null) {
               ApmTraceUtils.addExceptionToTrace(ex);
               if (ex.getCause() instanceof DestinationException) {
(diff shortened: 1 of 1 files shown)


airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -233,10 +235,7 @@ private void replicate(final Path jobRoot,
 
       LOGGER.info("Waiting for source and destination threads to complete.");
       // CompletableFuture#allOf waits until all futures finish before returning, even if one throws an
-      // exception. So in order to handle exceptions from a future immediately without needing to wait for
-      // the other future to finish, we first call CompletableFuture#anyOf.
-      CompletableFuture.anyOf(readSrcAndWriteDstThread, readFromDstThread).get();
-      LOGGER.info("One of source or destination thread complete. Waiting on the other.");
+      // exception.
       CompletableFuture.allOf(readSrcAndWriteDstThread, readFromDstThread).get();
```
gosusnp, line 239: I think the only future we really care about here is readFromDstThread, it might make sense to just wait for this one instead.

PR #20783, opened 2022-12-21
Source Postgres : Emit estimate trace messages for non-CDC mode

### LINKED ISSUE #19199: Database Sources (Postgres) implements Progress Bar protocol - `AirbyteEstimateTraceMessage`
Implement the `AirbyteEstimateTraceMessage` messages described in https://github.com/airbytehq/airbyte/pull/18875

Here is an example PR for another source: https://github.com/airbytehq/airbyte/pull/19197

The goal is to emit the predicated total number of rows/bytes in the sync (per stream).

akashkulk:
## What
Closes #19199

Postgres sources implement Progress Bar protocol - `AirbyteEstimateTraceMessage`
An `AirbyteEstimateTraceMessage` is emitted at the beginning of each sync with: 
- Estimated row count
- Estimated byte count

## How
Estimated byte count is calculated by asking postgres for the fast table silze `select pg_relation_size('table_name')` and then scaling by the amount of bytes that correspond to the actual number of rows being transmitted (considering the incremental offset). To better represent the amount of data Airbyte is moving (due to serialization), we are multiplying this by a factor of 2. Each row is estimated to be the same size. 

Estimates are sent once at the beginning of each sync. Progress bar for CDC is NOT supported.

## Recommended reading order
1. `AbstractDbSource.java` : define abstract methods for `estimateFullRefreshSyncSize` and `estimateIncrementalSyncSize` & call methods to estimate sync size while creating read iterators. Default behavior is a no-op for all non-Postgres connectors.
2. `PostgresSource.java` and `PostgresQueryUtils.java` : Logic to query Postgres to estimate row count + estimated bytes. 
3. Everything else

## 🚨 User Impact 🚨
Currently, there is no user impact. Still waiting on platform changes to implement stats persistence during a sync for user to actually see the progress bar


airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -538,4 +536,117 @@ protected boolean verifyCursorColumnValues(final JdbcDatabase database, final St
...
+
+      syncRowCount = getIncrementalTableRowCount(database, fullTableName, cursorInfo, cursorFieldType);
+      syncByteCount = (tableByteCount / tableRowCount) * syncRowCount;
+
+      // Here, we double the bytes estimate to account for serialization. Perhaps a better way to do this
+      // is to
+      // read a row and Stringify it to better understand the accurate volume of data sent over the wire.
+      // However, this approach doesn't account for different row sizes
+      AirbyteTraceMessageUtility.emitEstimateTrace(2 * syncByteCount, Type.STREAM, syncRowCount, tableName, schemaName);
+      LOGGER.info("estimate for table: " + fullTableName + " : {row_count:" + syncRowCount + ", total_bytes : " + syncByteCount);
```
ryankfu, line 591: It could be useful to also have within this log message the stream name since multiple streams can be written to similarly named table names

Also, minor nit but using a `String.format()` would make this cleaner to read/extend parameters

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -538,4 +536,117 @@ protected boolean verifyCursorColumnValues(final JdbcDatabase database, final St
...
+      final long syncByteCount;
+
+      syncRowCount = getIncrementalTableRowCount(database, fullTableName, cursorInfo, cursorFieldType);
+      syncByteCount = (tableByteCount / tableRowCount) * syncRowCount;
+
+      // Here, we double the bytes estimate to account for serialization. Perhaps a better way to do this
+      // is to
+      // read a row and Stringify it to better understand the accurate volume of data sent over the wire.
+      // However, this approach doesn't account for different row sizes
+      AirbyteTraceMessageUtility.emitEstimateTrace(2 * syncByteCount, Type.STREAM, syncRowCount, tableName, schemaName);
```
ryankfu, line 590: Would consider putting the value `2` as a constant since this seems like a value that will change over time

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -538,4 +536,117 @@ protected boolean verifyCursorColumnValues(final JdbcDatabase database, final St
...
+    try {
+      final String schemaName = configuredAirbyteStream.getStream().getNamespace();
+      final String tableName = configuredAirbyteStream.getStream().getName();
+      final String fullTableName =
+          getFullyQualifiedTableNameWithQuoting(getQuoteString(), schemaName, tableName);
+
+      final List<JsonNode> tableEstimateResult = getFullTableEstimate(database, fullTableName);
+
+      final long tableRowCount = tableEstimateResult.get(0).get(ROW_COUNT_RESULT_COL).asLong();
+      final long tableByteCount = tableEstimateResult.get(0).get(TOTAL_BYTES_RESULT_COL).asLong();
```
ryankfu, line 578: Unless I'm mistaken, these two values are only created once. Is it valuable to have these values logged so that they are persisted within the logs? 

```java
Logger.log(String.format("Total table row count: %s, Total table byte count: %s", tableRowCount, tableByteCount));
```

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresQueryUtils.java
```diff
@@ -0,0 +1,51 @@
...
+package io.airbyte.integrations.source.postgres;
+
+import java.sql.SQLException;
+
+/**
+ * Utility class to define constants related to querying postgres
+ */
+public class PostgresQueryUtils {
+
+  public static final String NULL_CURSOR_VALUE_WITH_SCHEMA =
```
ryankfu, line 14: Minor nit: since both `NULL_CURSOR_VALUE_WITH_SCHEMA` and `NULL_CURSOR_VALUE_NO_SCHEMA` are table queries, can we add `_QUERY` to the end to keep it in alignment with `TABLE_ESTIMATE_QUERY`? 

It also helps to quickly know this is a database query string

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
akashkulk: I've logged them below

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
akashkulk: Done

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresQueryUtils.java
akashkulk: Done

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
akashkulk: Done

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresUtils.java
```diff
@@ -42,6 +42,7 @@ public class PostgresUtils {
   public static final Duration MAX_FIRST_RECORD_WAIT_TIME = Duration.ofMinutes(20);
   public static final Duration DEFAULT_FIRST_RECORD_WAIT_TIME = Duration.ofMinutes(5);
 
+
```
rodireich, line 45: nit

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -550,4 +549,119 @@ protected boolean verifyCursorColumnValues(final JdbcDatabase database, final St
...
+                                             final ConfiguredAirbyteStream configuredAirbyteStream) {
+    try {
+      final String schemaName = configuredAirbyteStream.getStream().getNamespace();
+      final String tableName = configuredAirbyteStream.getStream().getName();
+      final String fullTableName =
+          getFullyQualifiedTableNameWithQuoting(schemaName, tableName, getQuoteString());
+
+      final List<JsonNode> tableEstimateResult = getFullTableEstimate(database, fullTableName);
+
+      final long syncRowCount = tableEstimateResult.get(0).get(ROW_COUNT_RESULT_COL).asLong();
```
rodireich, line 563: Need to check if key exists first?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -550,4 +549,119 @@ protected boolean verifyCursorColumnValues(final JdbcDatabase database, final St
...
+
+      // Here, we double the bytes estimate to account for serialization. Perhaps a better way to do this
+      // is to
+      // read a row and Stringify it to better understand the accurate volume of data sent over the wire.
+      // However, this approach doesn't account for different row sizes.
+      AirbyteTraceMessageUtility.emitEstimateTrace(PLATFORM_DATA_INCREASE_FACTOR * syncByteCount, Type.STREAM, syncRowCount, tableName, schemaName);
+      LOGGER.info(String.format("Estimate for table: %s : {sync_row_count: %s, sync_bytes: %s, total_table_row_count: %s, total_table_bytes: %s}",
+          fullTableName, syncRowCount, syncByteCount, syncRowCount, syncByteCount));
+    } catch (final SQLException e) {
+      throw new ConfigErrorException("Error occurred while attempting to estimate sync size", e);
```
rodireich, line 574: Can we warn rather than fail a sync altogether?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -550,4 +549,119 @@ protected boolean verifyCursorColumnValues(final JdbcDatabase database, final St
...
+      final List<JsonNode> tableEstimateResult = getFullTableEstimate(database, fullTableName);
+
+      final long tableRowCount = tableEstimateResult.get(0).get(ROW_COUNT_RESULT_COL).asLong();
+      final long tableByteCount = tableEstimateResult.get(0).get(TOTAL_BYTES_RESULT_COL).asLong();
+
+      final long syncRowCount;
+      final long syncByteCount;
+
+      syncRowCount = getIncrementalTableRowCount(database, fullTableName, cursorInfo, cursorFieldType);
+      syncByteCount = (tableByteCount / tableRowCount) * syncRowCount;
```
rodireich, line 598: edge case `tableRowCount == 0`

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -550,4 +549,119 @@ protected boolean verifyCursorColumnValues(final JdbcDatabase database, final St
...
+
+      // Here, we double the bytes estimate to account for serialization. Perhaps a better way to do this
+      // is to
+      // read a row and Stringify it to better understand the accurate volume of data sent over the wire.
+      // However, this approach doesn't account for different row sizes
+      AirbyteTraceMessageUtility.emitEstimateTrace(PLATFORM_DATA_INCREASE_FACTOR * syncByteCount, Type.STREAM, syncRowCount, tableName, schemaName);
+      LOGGER.info(String.format("Estimate for table: %s : {sync_row_count: %s, sync_bytes: %s, total_table_row_count: %s, total_table_bytes: %s}",
+          fullTableName, syncRowCount, syncByteCount, tableRowCount, tableRowCount));
+    } catch (final SQLException e) {
+      throw new ConfigErrorException("Error occurred while attempting to estimate sync size", e);
```
rodireich, line 608: likewise - maybe just warn?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -550,4 +549,119 @@ protected boolean verifyCursorColumnValues(final JdbcDatabase database, final St
...
+                  connection, fullTableName, quotedCursorField, cursorFieldType, cursorInfo.getCursor());
+              LOGGER.info("Table {} cursor count: expected {}, actual {}", fullTableName, cursorInfo.getCursorRecordCount(), actualRecordCount);
+              if (actualRecordCount == cursorInfo.getCursorRecordCount()) {
+                operator = ">";
+              } else {
+                operator = ">=";
+              }
+            }
+
+            final StringBuilder sql = new StringBuilder(String.format("SELECT COUNT(*) FROM %s WHERE %s %s ?",
```
rodireich, line 648: nit: move query to PostgresQueryUtils as well

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -550,4 +549,119 @@ protected boolean verifyCursorColumnValues(final JdbcDatabase database, final St
...
+                fullTableName,
+                quotedCursorField,
+                operator));
+
+            final PreparedStatement preparedStatement = connection.prepareStatement(sql.toString());
```
rodireich, line 653: nit: perhaps unify the way we query with a `prepareStatement()`

subodh1810 (approved):
Approved once Rodi's comments are addressed

rodireich (approved):

evantahler (commented):
At Grouparoo, we learned that if your database is sufficiently large the `COUNT(*)` (and `RANDOM()`) functions in postgres can take a really long time and be blocking because they might require a full table scan.

However, there's a way faster way to get a /rough/ estimate.  I learned about it from [TablePlus](https://tableplus.com/). Using the `performance` database you can compare:

```sql
SELECT COUNT(*) FROM "20m_users"."users"; -- 5s
select reltuples::int8 as count from pg_class c JOIN pg_catalog.pg_namespace n ON n.oid=c.relnamespace where nspname='20m_users' AND relname='users'; --30 ms
```

rodireich:
@akashkulk I merged a small change in the null cursor query that will cause a conflict. It's very trivial.
Let me know if you have any question.
#21051

akashkulk:
> At Grouparoo, we learned that if your database is sufficiently large the `COUNT(*)` (and `RANDOM()`) functions in postgres can take a really long time and be blocking because they might require a full table scan.
> 
> However, there's a way faster way to get a /rough/ estimate. I learned about it from [TablePlus](https://tableplus.com/). Using the `performance` database you can compare:
> 
> ```sql
> SELECT COUNT(*) FROM "20m_users"."users"; -- 5s
> select reltuples::int8 as count from pg_class c JOIN pg_catalog.pg_namespace n ON n.oid=c.relnamespace where nspname='20m_users' AND relname='users'; --30 ms
> ```

Thanks, didn't know about the fast estimate! I tried this out, and it is actually quite fast for some of the larger tables. However, it is an estimate and for small tables, this returns -1 as a rowcount. 

Thinking through the two sync cases : 
1) Full Refresh Syncs : All of the data will need to be read so the latency from the `COUNT(*)` operation will be small in comparison
2) Incremental Syncs : We need to perform another `SELECT COUNT(*) WHERE CURSOR > VALUE` operation which has a similar latency. 

Another solution is to do the rough estimate mentioned and fallback to `SELECT COUNT(*)` if the rowcount is negative, but I'd want to avoid that for now due to complexity. Even in the 1TB case, the `SELECT COUNT(*)` takes ~15m to finish, while a sync given our current throughput would take a couple of days, so I think the tradeoff in latency here _should_ be ok when compared with the complexity. Let me know if you have any concerns!

ryankfu (approved):
LGTM

EDIT: I would say that your conclusion of utilizing the `COUNT(*)` for now makes sense but if you believe there's value to using the fast estimate then it would be useful to maybe first go with the simple approach and then follow through with a ticket that captures the discussion with Evan. It sounds like the change to do the fast estimate _and then_ the default estimate isn't trivial

Given that, it makes sense to ship out a working solution first and then circle back when the fast estimation provides better performance

evantahler:
👍 on doing the fast count later.  
Just a reminder - that `count()` operation will likely lock the table too, so you can't do the slow count while you are reading, which is a real shame

akashkulk:
Thanks, I do think the fast select should be revisited. Have created this issue to track : https://github.com/airbytehq/airbyte/issues/21499

visnup:
I believe something about this PR causes our Postgres sync to hang. For what it's worth, our Postgres database is on Heroku.

Notice below in logs from 1.0.36 that the `Queueing query for table: events` happens immediately:
```
2023-01-20 21:16:46 source > Found table: public.pg_stat_statements
2023-01-20 21:16:46 source > Found table: public.item_cache
2023-01-20 21:16:46 source > Found table: public.events_of_interest
2023-01-20 21:16:46 source > using CDC: false
2023-01-20 21:16:46 source > using CDC: false
2023-01-20 21:16:46 source > Queueing query for table: events
2023-01-20 21:16:46 source > Preparing query for table: events
2023-01-20 21:16:46 source > Table events cursor count: expected 1, actual 1
```

In logs from 1.0.37, the last source message happens a minute or more before I canceled it:
```
2023-01-20 21:34:08 source > Found table: public.events_of_interest
2023-01-20 21:34:08 source > using CDC: false
2023-01-20 21:34:08 source > using CDC: false
2023-01-20 21:34:09 destination > HikariPool-1 - Added connection net.snowflake.client.jdbc.SnowflakeConnectionV1@7cca01a8
2023-01-20 21:34:09 destination > HikariPool-1 - Start completed.
2023-01-20 21:34:09 destination > closing connection
2023-01-20 21:34:10 destination > Preparing staging area in destination completed for schema AIRBYTE stream analytics_events
2023-01-20 21:34:10 destination > Preparing tmp tables in destination completed.
2023-01-20 21:35:42 INFO i.a.w.t.TemporalAttemptExecution(lambda$getCancellationChecker$5):228 - Running sync worker cancellation...
```

I've only waited up to 6 minutes before canceling as this is on a sync that normally takes end-to-end 1 minute to run.

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresQueryUtils.java
```diff
@@ -0,0 +1,37 @@
...
+      SELECT
+        (EXISTS (SELECT FROM information_schema.columns WHERE table_name = '%s' AND is_nullable = 'YES' AND column_name = '%s'))
+      AND
+        (EXISTS (SELECT from \"%s\" where \"%s\" IS NULL LIMIT 1)) AS %s
+      """;
+
+  public static final String TABLE_ESTIMATE_QUERY =
+      """
+            SELECT (SELECT COUNT(*) FROM %s) AS %s,
+            pg_relation_size('%s') AS %s;
```
visnup, line 30: The PR description says the intention is to only estimate table sizes using `pg_relation_size`, but won't this query end up doing a full count also? I believe this is what's causing my sync to hang for minutes (and maybe more) as attempting a count on my 186M row table takes quite a while.

visnup: If the number of rows is needed, I'd recommend using an estimate via `SELECT reltuples AS estimate FROM pg_class WHERE relname = 'table_name'`. [[1](https://wiki.postgresql.org/wiki/Count_estimate)]

visnup: aha, you know about this as part of #21499.

akashkulk:
> I believe something about this PR causes our Postgres sync to hang. For what it's worth, our Postgres database is on Heroku.
> 
> Notice below in logs from 1.0.36 that the `Queueing query for table: events` happens immediately:
> 
> ```
> 2023-01-20 21:16:46 source > Found table: public.pg_stat_statements
> 2023-01-20 21:16:46 source > Found table: public.item_cache
> 2023-01-20 21:16:46 source > Found table: public.events_of_interest
> 2023-01-20 21:16:46 source > using CDC: false
> 2023-01-20 21:16:46 source > using CDC: false
> 2023-01-20 21:16:46 source > Queueing query for table: events
> 2023-01-20 21:16:46 source > Preparing query for table: events
> 2023-01-20 21:16:46 source > Table events cursor count: expected 1, actual 1
> ```
> 
> In logs from 1.0.37, the last source message happens a minute or more before I canceled it:
> 
> ```
> 2023-01-20 21:34:08 source > Found table: public.events_of_interest
> 2023-01-20 21:34:08 source > using CDC: false
> 2023-01-20 21:34:08 source > using CDC: false
> 2023-01-20 21:34:09 destination > HikariPool-1 - Added connection net.snowflake.client.jdbc.SnowflakeConnectionV1@7cca01a8
> 2023-01-20 21:34:09 destination > HikariPool-1 - Start completed.
> 2023-01-20 21:34:09 destination > closing connection
> 2023-01-20 21:34:10 destination > Preparing staging area in destination completed for schema AIRBYTE stream analytics_events
> 2023-01-20 21:34:10 destination > Preparing tmp tables in destination completed.
> 2023-01-20 21:35:42 INFO i.a.w.t.TemporalAttemptExecution(lambda$getCancellationChecker$5):228 - Running sync worker cancellation...
> ```
> 
> I've only waited up to 6 minutes before canceling as this is on a sync that normally takes end-to-end 1 minute to run.

Thanks for pointing this out. I have done some analysis and testing above to determine how much latency is added due to the select count(*)  : https://github.com/airbytehq/airbyte/pull/20783#issuecomment-1384698480

But it seems like there are still some cases where the latency is a problem if especially if your sync is really fast without the estimation.

I also opened issue
https://github.com/airbytehq/airbyte/issues/21499 to track the optimization you mentioned above, although I was experimenting with a different query : select reltuples::int8 as count from pg_class c JOIN pg_catalog.pg_namespace n ON n.oid=c.relnamespace where nspname='schema' AND relname='table_name';

The issue I had encountered with that query is that the estimates were really off for smaller tables. However, I don't think this is too much of a problem : With smaller tables we can omit emitting the trace message itself. So, will start fixing this issue. In the meantime, I suggest using 1.0.36.

visnup:
Thanks! Probably obvious, but works for me now!

PR #20816, opened 2022-12-22
🪟🎉 Connector builder: Add copy to/from stream buttons

### LINKED ISSUE #20514: [Connector Builder UI] Duplicate configuration sections across streams
# Tell us about the problem you're trying to solve
Add the ability for users to duplicate individual configuration sections (pagination, stream slicer, ...) from stream to stream

# Describe the solution you’d like

Acceptance Criteria:

* A button exists for each section in the stream config page, allowing to import the settings from another stream as well as copying it to another stream or all streams

flash1293:
Fixes https://github.com/airbytehq/airbyte/issues/20514

## What

Adds buttons to copy pagination and slicing settings between streams:

https://user-images.githubusercontent.com/1508364/209148433-dcfaf932-1263-43a8-8181-c3ff79da33c4.mp4

## How

The BuilderCard gets extended with a copy config which is rendering the copy to/from buttons (copy to is hidden if a section is disabled, both are hidden if there is only one stream).

Actual copying is done using `setFieldValue` with a path.

One thing to note: The target is not automatically touched, so on copying a section with errors, it won't directly flag the target streams with the error indicator (the user needs to click on the test button to trigger that). Two reasons why I left it like this:
* Touching the fields in the section to make the error show up is more complex than I would like
* I'm not sure whether it's the right behavior or not


airbyte-webapp/src/components/connectorBuilder/Builder/BuilderCard.tsx
```diff
@@ -33,7 +52,139 @@ export const BuilderCard: React.FC<React.PropsWithChildren<BuilderCardProps>> =
...
+          );
+        }}
+      >
+        <ModalBody className={styles.modalStreamListContainer}>
+          {streams.value.map((stream, index) => (
+            <label htmlFor={`copy-to-stream-${index}`} key={index} className={styles.toggleContainer}>
+              <CheckBox
+                id={`copy-to-stream-${index}`}
+                checked={selectMap[index] || false}
+                disabled={index === currentStreamIndex}
```
lmossman, line 142: I was surprised to see the current stream in this modal. Maybe we just shouldn't show it at all?

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderCard.tsx
```diff
@@ -33,7 +52,139 @@ export const BuilderCard: React.FC<React.PropsWithChildren<BuilderCardProps>> =
...
+                onChange={() => {
+                  setSelectMap({ ...selectMap, [index]: !selectMap[index] });
+                }}
+              />
+              {stream.name}
```
lmossman, line 147: small nit: if a stream does not have a name, then the text is just blank in both of these modals. Maybe we should show something like `(empty)` in this case, similar to how we do in the stream sidebar?

airbyte-webapp/src/components/connectorBuilder/Builder/Builder.module.scss
```diff
@@ -18,5 +18,6 @@
 .form {
   flex: 1;
   padding: variables.$spacing-xl;
+  padding-right: 80px;
```
lmossman, line 21: nit: apply this padding to the left as well, so that the form is centered in its containing div

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderCard.tsx
```diff
@@ -33,7 +52,139 @@ export const BuilderCard: React.FC<React.PropsWithChildren<BuilderCardProps>> =
           {toggleConfig.label}
         </div>
       )}
+      {copyConfig && streams.length > 1 && (
+        <div className={styles.copyButtonContainer}>
+          {get(streams[copyConfig.currentStreamIndex], copyConfig.path) && (
```
lmossman, line 57: nit: swap the order of the copy to/from buttons, so that the copy from button stays in the same place regardless of if the card is toggled on or off

lmossman (approved):
Left some suggestions, nothing is a major blocker. Tested locally and it LGTM

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderCard.tsx
flash1293: Valid point, removed

airbyte-webapp/src/components/connectorBuilder/Builder/Builder.module.scss
flash1293: 80px is a lot of horizontal screen space to lose, instead I made this styling conditional based on whether there are multiple streams in the first place. Also solves the issue of not having an unnecessary gap on user input and global views.

lmossman: Nice, I like that approach 👍

PR #20821, opened 2022-12-22
[local airbyte cdk tests] update README.md to fix gradle command and …

maxi297:
## What
Update documentation to allow to test a broader scope of images

### DIFF
--- a/airbyte-cdk/python/README.md
+++ b/airbyte-cdk/python/README.md
@@ -112,7 +112,13 @@ Create a symlink that connects `<connector-directory>/airbyte-cdk` to your local
 ln -s ../../../airbyte-cdk/python airbyte-cdk
 ```
 
-Add the following lines to your connector's `Dockerfile`, before the line that installs dependencies via `pip install -e .`:
+Now that you have a symbolic link in your source, you will nedd to install it in the docker image. Depending on how the docker image is build you will have two possibilities:
+1. If image building process uses a temporary image called `builder`, add the following just after the copy of the installed packages to the folder `/usr/local`: 
+```Dockerfile
+COPY airbyte-cdk airbyte-cdk
+RUN PYTHONUSERBASE=/usr/local pip install -e ./airbyte-cdk
+```
+2. Add the following lines to your connector's `Dockerfile`, before the line that installs dependencies via `pip install -e .`:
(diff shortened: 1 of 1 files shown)


airbyte-cdk/python/README.md
```diff
@@ -112,7 +112,13 @@ Create a symlink that connects `<connector-directory>/airbyte-cdk` to your local
 ln -s ../../../airbyte-cdk/python airbyte-cdk
 ```
 
-Add the following lines to your connector's `Dockerfile`, before the line that installs dependencies via `pip install -e .`:
+Now that you have a symbolic link in your source, you will nedd to install it in the docker image. Depending on how the docker image is build you will have two possibilities:
+1. If image building process uses a temporary image called `builder`, add the following just after the copy of the installed packages to the folder `/usr/local`:
```
girarda, line 116: can you link to an example of an image use a builder tmp image for reference?

girarda: don't we also need to run `pip install` when the the image does not use the builder tmp image?

maxi297: > can you link to an example of an image use a builder tmp image for reference?

This Dockerfile [pip install in a "builder" image](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-square/Dockerfile#L15) and [copy the files on the "real" image](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-square/Dockerfile#L22). This didn't seem to work for me. I assume it is because of the symlink. Therefore instead, I did add the lines:
```
COPY airbyte-cdk airbyte-cdk
RUN PYTHONUSERBASE=/usr/local pip install -e ./airbyte-cdk
```
... after https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-square/Dockerfile#L22 

> don't we also need to run pip install when the the image does not use the builder tmp image?

Yes! This is why I kept point #2 [here](https://github.com/airbytehq/airbyte/pull/20821/files#diff-0d47fcf235c76ca4ac08dad91ea80a845882bf5e5e862b6c729826303a15c0f7R121)

girarda (approved):

PR #20838, opened 2022-12-22
Only add definitions updater cron when enabled

gosusnp:
## What

I keep seeing the definitions updater cron spinning up for no reason. We should not have the task scheduled if it is disabled.

## How

Use `@requires` to restrict the bean to when it's enabled.


pedroslopez (approved):
Nice!

pedroslopez (commented):
We probably don't need to track `shouldUpdateDefinitions` anymore if this is the case

gosusnp:
@pedroslopez, yep, I was wondering about that, removed the flag.

pedroslopez (approved):

jdpgrailsdev (approved):
:shipit:

PR #20843, opened 2022-12-22
🪟 🔧 Allow overwriting experiments and features during development

timroes:
## What

This allows easily overwriting the state of experiments or features during development with a custom value. Those overwrites work whether the app is run in OSS or in Cloud mode.

Those overwrite files are ignored in git so they can't accidentally be commited. Also the overwrites will only work during development, so they can't be missued in production systems.

Documentation about that feature will be added to Notion after merging.

### Experiments

You can overwrite the state of experiments (`useExperiment`) via a file called `.experiments.json` in the `airbyte-webapp` folder, by their experiment id:

```json
{
  "authPage.signup.hideName": true,
  "connector.orderOverwrite": { "connectorId": 5 }
}
```


airbyte-webapp/scripts/dev-overwrites.js
```diff
@@ -0,0 +1,46 @@
...
+const FEATURES_FILE = path.resolve(__dirname, "../.features.dev");
+
+if (fs.existsSync(FEATURES_FILE)) {
+  console.log("\nOverwriting feature states from .features.dev ...");
+  const overwrites = dotenv.parse(fs.readFileSync(FEATURES_FILE));
+  Object.entries(overwrites).forEach(([key, value]) => {
+    console.log(`Overwrite feature ${key} with value ${value}`);
+    process.env[`REACT_APP_FEATURE_${key}`] = value;
+  });
+}
```
josephkmh, line 46: This feels like kind of a roundabout way of setting environment variables. Couldn't we just set `REACT_APP_FEATURE_FOO=bar` in `.env.development` instead?

airbyte-webapp/.gitignore
```diff
@@ -24,8 +24,12 @@ yarn-error.log*
 /.idea
 
 .npmrc
+
+# Environment overwrites
 .env.development
 .env.production
+.experiments.dev
+.features.dev
```
josephkmh, line 32: I don't feel super strongly about this, but: I would prefer keeping all environment variable overrides in one place, e.g. `.env.development`. My thinking is that the more gitignored environment override files we have, the more complex it is to know which values are coming from where, and the more potential for a confusing local dev experience.

Is there a reason why we don't just do something like `REACT_APP_FEATURE_FOO=bar` and `REACT_APP_EXPERIMENT_FOO=bar` in `.env.development`? The only blocker I see is that experiments are defined as a TS interface, so we would potentially need a different structure there so we can loop over it and look for overrides (instead of defining `REACT_APP_EXPERIMENT_OVERWRITES` as a JSON object).

airbyte-webapp/scripts/dev-overwrites.js
timroes: We can already do this, this was the purpose of this PR to extract that into a separate file.

airbyte-webapp/.gitignore
timroes: There are 2 main issues just using env files for this only:

* The main reason against that was, that ids of experiments are dot notated camel cased, which might lead to the situation that we might need to have properties either called `REACT_APP_EXPERIMENT_connector.SomeThing=true`, which looks just super weird, or we'd have some magic logic converting that into upper snake case, but then you'd need to implicitally understand how that would be looking alike, would it be `REACT_APP_EXPERIMENT_CONNECTOR_SOME_THING` or `REACT_APP_EXPERIMENT_CONNECTOR_SOMETHING`?
* Experiments (in contrast to features, who're just on/off) can support JSON values. We need some logic to handle that when overwriting. Saying we always do a `JSON.parse` on them, causes us to also need to quote plain string values, which might be non intuitive either e.g. `REACT_APP_EXPERIMENT_STRING_VALUE="foo"`, since all other string values in the env file won't need to be quoted.

I don't have strong opinions around keeping everything in one file together (as long as they following the same logic for all keys inside it), but know that this came up as a problem with the rest of the chapter, so maybe you could raise that question in the chapter slack to get more feedback on whether we want to go this route or not?

josephkmh: Right, the JSON values make sense why we would want to have a separate file. From this perspective keeping them in one file does not seem like the best solution to me. I think I'd either be in favor of your original syntax in `.experiments.dev` or something like `.experiments.json`:

```json
{
  "authPage.signup.hideName": "true",
  "authPage.oauth.position": "top",
  "connector.orderOverwrite": { "connectorId": 5 }
}
```

Which is maybe a little easier to read/write because you get native JSON syntax highlighting in your IDE.

airbyte-webapp/scripts/dev-overwrites.js
```diff
@@ -0,0 +1,14 @@
+const fs = require("fs");
+const path = require("path");
+
+const EXPERIMENTS_FILE = path.resolve(__dirname, "../.experiments.json");
+
+if (fs.existsSync(EXPERIMENTS_FILE)) {
+  console.log("\nOverwriting experiments from .experiments.json ...");
```
josephkmh, line 7: Just a note: these `console.log` calls get cleared by CRA in the console very quickly, so not _super_ useful. With vite we could disable the `clearScreen` option to keep this debug info in the console.

joeykmh (approved):
Looks good! I tested it locally overwriting a few experiment values, worked as expected 👍 This will testing experiments in OSS much easier 🙂

PR #20844, opened 2022-12-22
🪟 🎉 Disable deselection of cursor field/primary key in the UI

### LINKED ISSUE #20586: Disable deselection of cursor field/primary key
1. The toggle for a primary key should be disabled, with a tooltip reading `The primary key cannot be deselected`
2. The toggle for a cursor field should be disabled, with a tooltip reading `The cursor field cannot be deselected`
3. If a user changes the primary key or cursor field, that field is automatically selected (if it was deselected before) and then disabled

joeykmh:
## What
Closes https://github.com/airbytehq/airbyte/issues/20586

This PR automatically adds special handling of cursor fields and primary keys with regards to field selection in the airbyte UI.

## How
- Disables the deselection of the primary key(s) or cursor field, instead displaying a tooltip with an explanation
- In case a user selects a cursor or primary key that was previously deselected, the UI automatically selects this field
- Extracts logic from `useCallback()` functions into pure functions, adds unit tests to cover these

## Recommended reading order
1. `FieldRow.tsx` - where the pk & cursor field deselection is disabled
2. `CatalogSection.tsx` - where `useCallback()` logic has been extracted to pure functions
3. `streamConfigHelpers` - the pure functions that return a new stream config


teallarson (changes requested):
Haven't found why in the code yet, but it doesn't look like this takes into account source-defined cursors and primary keys.

Screenshot 2023-01-05 at 3 43 42 PM

(^ this example is from a Mailchimp source.  The disabled rows are part of my composite primary key that I configured, but my cursor (timestamp) can still be toggled.)

airbyte-webapp/src/components/connection/CatalogTree/FieldRow.tsx
```diff
@@ -54,18 +54,32 @@ const FieldRowInner: React.FC<FieldRowProps> = ({
   const isCursor = equal(config?.cursorField, field.path);
   const isPrimaryKey = !!config?.primaryKey?.some((p) => equal(p, field.path));
   const isNestedField = SyncSchemaFieldObject.isNestedField(field);
+  const isDisabled = (isCursorEnabled && isCursor) || (isPrimaryKeyEnabled && isPrimaryKey) || isNestedField;
```
tealjulia, line 57: Here's the culprit -- `isCursorEnabled` appears to actually be whether it's configurable, same with `isPrimaryKeyEnabled`.  Source-defined ones wouldn't be configurable, but are definitely required.

josephkmh: Oh, good call! I did not have that down as a requirement, but of course it makes sense.

josephkmh: Fixed this [here](https://github.com/airbytehq/airbyte/pull/20844/commits/b53c2a806cdb44f998600ca5d095cc3664702ca5). I updated the variable names to hopefully better express what they're used for. Thanks for catching that!

airbyte-webapp/src/components/connection/CatalogTree/streamConfigHelpers/streamConfigHelpers.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  // If field selection is enabled, we need to be sure the new fieldPath is also selected
+  if (
+    fieldIsSelected &&
+    config?.fieldSelectionEnabled &&
+    !previouslySelectedFields.find((field) => isEqual(field, fieldPath))
+  ) {
+    if (previouslySelectedFields.length === numberOfFieldsInStream - 1) {
+      return { primaryKey: newPrimaryKey, selectedFields: [], fieldSelectionEnabled: false };
```
tealjulia, line 84: Can you help me understand why we set `fieldSelectionEnabled: false` in this case?

josephkmh: In this case the only unselected field is the field being added to the primary key. So changing `fieldSelectionEnabled: false, selectedFields: []` effectively turns off column selection (because now all fields are selected).

Your comment did make me realize there is a different bug in `updatePrimaryKey()` in this same logic, so thanks for that 🙂

tealjulia: Cool!  One more question here -- it doesn't look like `fieldSelectionEnabled` is checked at (or above) the `FieldRow` level unless there was a piece I missed.  Could that help us avoid some potentially redundant calculation there?

josephkmh: I don't think there's any added calculation overhead related to field selection until the callbacks that change `primaryKey` and `cursorField` which are contained in `streamConfigHelpers.ts`. Did you spot something in particular?

tealjulia: `FieldRow` is evaluating if a field is a cursor or a primary key every time.  Couldn't we avoid that if we know `fieldSelectionEnabled = false` for that entire stream?

josephkmh: `isCursor` and `isPrimaryKey` aren't just for field selection, though. They're also needed to know whether the cursor field radio button or primary key checkbox is checked.

tealjulia: Ah, yep.  Thanks.

teallarson (approved):
Code lgtm and tested locally.

I think it would be prudent to add a test case specifically for source-defined cursors and source-defined primary keys not being able to be deselected, but I don't necessarily think it's blocking.

airbyte-webapp/src/components/connection/CatalogTree/streamConfigHelpers/streamConfigHelpers.ts
```diff
@@ -6,17 +6,18 @@ import { AirbyteStreamConfiguration, SelectedFieldInfo } from "core/request/Airb
...
 export function mergeFieldPathArrays(...args: SelectedFieldInfo[][]): SelectedFieldInfo[] {
-  const set = new Set<string[]>();
+  const set = new Set<string>();
 
   args.forEach((array) =>
     array.forEach((selectedFieldInfo) => {
       if (selectedFieldInfo.fieldPath) {
-        set.add(selectedFieldInfo.fieldPath);
+        const key = JSON.stringify(selectedFieldInfo.fieldPath);
+        set.add(key);
```
josephkmh, line 15: This converts the `fieldPath` (e.g. `['path', 'to', 'field']`) into a string (e.g. `"[\"path\",\"to\",\"field\"]"`) so that we can add it as a unique value to the set. It gets converted back into an array before returning.

This conversion between array > string > array might seem expensive, but it allows us use a `Set` to avoid duplicates when merging, which avoids a lot of nested comparisons of arrays.

josephkmh:
Thanks @tealjulia! I made one more minor adjustment here FYI https://github.com/airbytehq/airbyte/pull/20844/commits/d1938199b88db9560ba449607003442468963cd2#

airbyte-webapp/src/components/connection/CatalogTree/streamConfigHelpers/streamConfigHelpers.ts
tealjulia: I think that's a reasonable tradeoff.  It sounds like it would still be more efficient for a large sync catalog or a sync catalog with a lot of changes than what this was doing before, correct?

josephkmh: Yes, at least as efficient, and in the best case more efficient 👍

PR #20846, opened 2022-12-22
🐛 🎉 Source Airtable: migrate to the `Metadata API` for dynamic schema generation

### LINKED ISSUE #12491: 🐛 Airtable source: remove base & tables from spec.json, and use the API to list available bases & tables, using bases as the namespaces for streams
Airtable bases are equivalent to database schemas. We should use them as such. 

1. Remove the base & tables parameters from spec.json (we'll need to figure out how to handle backwards compatibility here, this is a tricky one)
2. Instead, list the bases inside the connector (when running `discover`) by calling the endpoint `https://api.airtable.com/v0/meta/bases` (see [the docs](https://airtable.com/api/meta)) and set them as the namespace property of each available table (stream). 
3. Then list all tables in a base , those are the streams and the base is the namespace 
4. Return all this in discovery output

### LINKED ISSUE #12955: 🐛 Source Airtable: fix `unit_tests`
Need to fix all unit_tests according to the code changes.

### LINKED ISSUE #17794: source-airtable: `discover` generates incomplete schema
## Environment
- Source-connector: `source-airtable:0.1.2`

## Current Behavior
The generated catalog misses some fields on Airtable stream if these fields are `null` in the first row of a table.
It breaks downstream normization

## Expected Behavior
The discovered catalog is complete

bazarnov:
## What
Resolves: 
- https://github.com/airbytehq/airbyte/issues/17794
- https://github.com/airbytehq/alpha-beta-issues/issues/74
- https://github.com/airbytehq/airbyte/issues/12491
- https://github.com/airbytehq/alpha-beta-issues/issues/19
- https://github.com/airbytehq/alpha-beta-issues/issues/168
- https://github.com/airbytehq/alpha-beta-issues/issues/95
- https://github.com/airbytehq/airbyte/issues/12955

## How
- implemented `Metadata API` approach to providing dynamic JSON schema generation
- removed `base_id` and `tables` properties from input config, now these properties are dynamically fetched and passed to each stream
- general refactor of the code + best practices + HTTP error handling

## 🚨 User Impact 🚨
Users should re-setup the connector and redo the sync all over again, since it's a major change.

## Pre-merge Checklist

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>


lazebnyi (approved):

docs/integrations/sources/airtable.md
```diff
@@ -35,6 +33,7 @@ See information about rate limits [here](https://support.airtable.com/hc/en-us/a
 
 | Version | Date       | Pull Request                                             | Subject                                  |
 | :------ | :--------- | :------------------------------------------------------- | :--------------------------------------- |
+| 0.1.4   | 2022-12-22 | [20846](https://github.com/airbytehq/airbyte/pull/20846) | Migrated to Metadata API for dynamic schema generation           |
```
evantahler, line 36: Can you share more about the migration plan for users?  I think this is at least a minor Semver change as a config value was removed... but might this be a major version change?

For users who previously provided the `base_id` and `tables` configuration values, will they need to re-configure the connector, or will things work as they used to as the new base+table streams will be unselected?

bazarnov: @evantahler 

The Customers should re-do the sync and ideally re-set up the connector with all the data. Here are a couple of reasons why:
1) the schemas were declared incorrectly, therefore even if present customers are using the connector, they might have data issues because the discovery  mechanism was not working correctly
2) It could be the major version, but I don't see a big difference, if you insist  - let's go with it.
3) the previous configuration should also work, BTW, but there could the issues with normalization, if applicable. Therefore:
The users should re-configure their connection, for only 1 valid reason  - to have the schemas and data fulfilled correctly.

bazarnov: @evantahler 
the version was updated to `0.2.0`  to reflect the major version updates.

airbyte-integrations/connectors/source-airtable/source_airtable/source.py
```diff
@@ -3,99 +3,77 @@
...
+            for base in AirtableBases(authenticator=auth).read_records(sync_mode=None):
+                base_id = base.get("id")
+                base_name = base.get("name")
+                self.logger.info(f"Reading first table info for base: {base_name}")
+                next(AirtableTables(base_id=base_id, authenticator=auth).read_records(sync_mode=None))
+            return True, None
+        except Exception as e:
+            return False, str(e)
+
+    def prepare_catalog_for_base(self, base_id: str, base_name: str, base_tables: list) -> Iterable[Mapping[str, Any]]:
```
alafanechere, line 37: The type hint of this function mentions we can expect it to return an iterable, but it actually does not return anything and mutate `self.streams`. I'd prefer it to produce a list of Airbyte Streams, and you make discover call this function to get the list.
```suggestion
    def get_streams_from_base(self, base_id: str, base_name: str, base_tables: list) -> Iterable[Mapping[str, Any]]:
```

airbyte-integrations/connectors/source-airtable/acceptance-test-config.yml
```diff
@@ -1,20 +1,28 @@
...
+      - config_path: "secrets/config.json"
+        status: "succeed"
+      - config_path: "integration_tests/invalid_config.json"
+        status: "failed"
   discovery:
-    - config_path: "secrets/config.json"
+    tests:
+      - config_path: "secrets/config.json"
+      # bypassed this check, because discovery mechanism was changed
+        backward_compatibility_tests_config:
```
alafanechere, line 18: Could you please share the differences between the previous catalog structure and the new one?
Can we find a way to make the new catalog backward compatible? This change might require all our users to reset their data and refresh their schema.

alafanechere (changes requested):
Thanks, this is a significant improvement!
It will require a customer outreach as to benefit from this evolution users would need to reset and refresh their schema.

A backward-compatible approach would be to keep the spec untouched. We lose the auto-discovery of bases and tables but we'd still benefit from more accurate schema discovery.

I'll let @erica-airbyte and @YowanR decide if the breaking change is tolerable or not.

bazarnov:
@alafanechere @erica-airbyte 
I confirm, we cannot make this backward capable with older versions, since the whole connectors logic was refactored to provide the correct discover functionality.

1) The biggest difference is what the schema will contain in the end. Previously it was built using a single complete row of data, omitting empty columns and reflecting wrong data structure
Now we fetch the correct schema and keep the empty columns in place, providing correct data structure.
2)the names of streams are dynamic now, based on where the table is located (base name/table name), which is also a part of dynamic schema processing and it's now backwards capable.
3) the old configs will work as expected but the base_id/tables fields are not considered anymore, so the sync after the upgrade will just be the full refresh using new streams and new approach it should not "break" the existing connections, but will create new tables in the destination I believe. (I'll make some tests for this asap and reply)

airbyte-integrations/connectors/source-airtable/Dockerfile
```diff
@@ -34,5 +34,5 @@ COPY source_airtable ./source_airtable
 ENV AIRBYTE_ENTRYPOINT "python /airbyte/integration_code/main.py"
 ENTRYPOINT ["python", "/airbyte/integration_code/main.py"]
 
-LABEL io.airbyte.version=0.1.3
+LABEL io.airbyte.version=0.2.0
```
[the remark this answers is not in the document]
bazarnov: Done.

airbyte-integrations/connectors/source-airtable/source_airtable/source.py
bazarnov: Now the logic populates ready-streams into the list, and processes that list, I've moved this logic into SchemaHelpers to clear the code, but would rather avoid returning the list from `get_streams_from_base` method and leave it as is.
the TypeHint was removed as well.

airbyte-integrations/connectors/source-airtable/acceptance-test-config.yml
bazarnov: The biggest difference is what the schema will contain in the end. Previously it was built using a single complete row of data, omitting empty columns and reflecting wrong data structure
Now we fetch the correct schema and keep the empty columns in place, providing correct data structure.

bazarnov:
@alafanechere 
UPDATE:
The customers need just to `Refresh Source Schema`, which brings a new format of tables and streams, and uncheck the `Reset affected streams (recommended)` while saving the replication settings to keep the old data in place and produce new streams and catalog, which will replace the old one, with old streams format. But that's because  I'm using a new Personal API Token with selected scopes, which was not available previously for all users, now it is.

The actual connection should also be refreshed with the new Personal API TOKEN acquired using Airtable UI because the whole mechanism of how Airtable does the authentication was changed. 
So the general recommendation is to set another new connection from Airtable to the destination,  to have a new way of data replication, and migrate gracefully in the future.

alafanechere:
Thanks @bazarnov for all these details.
@erica-airbyte / @YowanR I think we can merge this because:
- It's increasing the stability: more accurate schema, less normalization errors
- It's providing a more complete set of features: auto-discovery of available bases / tables
- This is an alpha connector

@erica-airbyte shall we hold this while you prepare a customer outreach?

bazarnov:
> * It's increasing the stability: more accurate schema, less normalization errors

About this one:
> It's increasing the stability: more accurate schema, less normalization errors

I've also tested sync with normalization after this change, works as expected.

YowanR:
I would agree with that statement @alafanechere! This helps makes Airtable better and might open more opportunities in the future as well. @bazarnov Let's wait for @erica-airbyte to comment though to make sure that comms are handled before merging!

erica-airbyte:
@alafanechere I want to make sure I am understanding correctly what I am telling customers:

Once this update goes out the customer connection will break but they will need to first update their API token due to how Airtable updated their authentication then `Refresh Source Schema` to bring in the new format of tables and streams, and uncheck the `Reset affected streams (recommended)` while saving the replication settings? 

Is that right? 

If so I will reach out today. When do we want to deploy this?

alafanechere:
@bazarnov could you confirm Erica's statement? ☝️ You're more aware than I do 😄

bazarnov:
I confirm, correct. Please follow the public docs https://airtable.com/developers/web/guides/personal-access-tokens to create new tokens before the docs are updated.

airbyte-integrations/connectors/source-airtable/source_airtable/schema_helpers.py
```diff
@@ -0,0 +1,60 @@
...
+
+from airbyte_cdk.models import AirbyteStream
+from airbyte_cdk.models.airbyte_protocol import DestinationSyncMode, SyncMode
+
+
+class SchemaHelpers:
+
+    streams_catalog: Iterable[Mapping[str, Any]] = []
+
+    def get_streams_from_base(self, base_id: str, base_name: str, base_tables: list):
```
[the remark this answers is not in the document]
bazarnov: I've actually removed this method and made it's logic a part of the `discover`  method, to have everything in 1 place.

erica-airbyte:
Ok thanks! Can we plan on having this go out next Tuesday at 9AM PST so customer know when they will need to be ready?

erica-airbyte:
All comms have been sent out to customers and I gave them Tuesday the 10th around 9AM PST that this will go out.

alafanechere:
@erica-airbyte @bazarnov I'll merge it on Tuesday 9 am PST and it will get automatically deployed to cloud.

alafanechere (approved):

erica-airbyte:
This is still going out tomorrow right at 9AM PST right?

PR #20788, opened 2022-12-22
Destination-bigquery: add ability to impersonate another GCP account

grishick:
## What
This PR is based on [Impersonation PR](https://github.com/airbytehq/airbyte/pull/15820) by @marcelopio and adds testing for credential impersonation. In order to add more test cases, I had to refactor `BigQueryDestinationTest`

PR #20851, opened 2022-12-23
Refactor BigQuery Destination Integration tests

grishick:
This refactoring reduces code duplication and moves configuration logic from code into config files. 
The purpose of this change is to make it easier to add more test cases for configuration variations such as
data location, accounts with various permission combinations, and account impersonation. This change was inspired by the impersonation [PR](https://github.com/airbytehq/airbyte/pull/20788). In order to run all the tests with impersonated credentials, we would have had to write a bunch more code. Instead, after this refactoring, we can just add two more credential files (1 - positive test, 2 - negative test) and about ~10 lines of code.

Part of the change is removing `BigQueryGcsDestinationTest`, by replacing it with a gcs-enabled config file (I added new config files to GSM before submitting this PR).


airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTest.java
```diff
@@ -108,81 +111,149 @@ class BigQueryDestinationTest {
...
+    if (!Files.exists(CREDENTIALS_STANDARD_INSERT_PATH)) {
       throw new IllegalStateException(
-          "Must provide path to a big query credentials file. By default {module-root}/config/credentials.json. Override by setting setting path with the CREDENTIALS_PATH constant.");
+          "Must provide path to a big query credentials file. By default destination-bigquery/secrets/credentials-standard.json");
+    }
+    if (!Files.exists(CREDENTIALS_WITH_MISSED_CREATE_DATASET_ROLE_PATH)) {
+      throw new IllegalStateException("""
+                                      Json config not found. Must provide path to a big query credentials file,
+                                       please add file with creds to
+                                      ../destination-bigquery/secrets/credentials-with-missed-dataset-creation-role.json.""");
```
edgao, line 165:
```suggestion
                                      <...>/destination-bigquery/secrets/credentials-with-missed-dataset-creation-role.json.""");
```
since `..` might get misinterpreted

edgao (approved):
checking my understanding:
* BigQueryDAT - runs the full DAT suite against a standard inserts config
* BigQueryGcsDAT - runs the full DAT suite against a GCS staging config. Extends BigQueryDAT for convenience
    * Maybe we should have a root-level `BigQueryDAT` and then inherit `BigQueryStandardInsertsDAT` + `BigQueryGcsDAT` from it? It might reduce some code duplication between them in the setup method
* BigQueryDestinationTest - verifies a couple different expected success/failure configs with basic behaviors (`check`, `write`) but doesn't run the full DAT process against all of them

?

---

left a couple small comments+questions but overall this makes sense.

grishick:
Looks like we broke DATs for destination-bigquery-denormalized back in November. I opened a separate PR to fix it: https://github.com/airbytehq/airbyte/pull/20871

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTestUtils.java
```diff
@@ -0,0 +1,121 @@
...
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.LinkedList;
+import java.util.List;
+import org.slf4j.Logger;
+
+public class BigQueryDestinationTestUtils {
+
+  public static JsonNode createConfig(Path configFile, String datasetId) throws IOException {
```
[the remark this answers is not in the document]
grishick: good point. We should make the secrets stored in GMS have the structure that the tests expect and avoid all this extra shuffling

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTestUtils.java
```diff
@@ -0,0 +1,121 @@
...
+   * Remove all the GCS output from the tests.
+   */
+  public static boolean tearDownGcs(AmazonS3 s3Client, JsonNode config, Logger LOGGER) {
+    if(s3Client == null) {
+      return false;
+    }
+
+    final JsonNode properties = config.get(BigQueryConsts.LOADING_METHOD);
+    final String gcsBucketName = properties.get(BigQueryConsts.GCS_BUCKET_NAME).asText();
+    final String gcs_bucket_path = properties.get(BigQueryConsts.GCS_BUCKET_PATH).asText();
```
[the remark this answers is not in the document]
grishick: is `gcs_bucket_path = ""` a valid config for staging?

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationAcceptanceTest.java
```diff
@@ -187,71 +177,37 @@ private List<JsonNode> retrieveRecordsFromTable(final String tableName, final St
...
+  }
 
-    setupBigQuery(credentialsJson);
+  protected void setUpBigQuery(JsonNode config, String datasetId) throws IOException {
+    String projectId = config.get(BigQueryConsts.CONFIG_PROJECT_ID).asText();
+    bigquery = BigQueryDestinationTestUtils.initBigQuery(config, projectId);
+    try {
+      dataset = BigQueryDestinationTestUtils.initDataSet(config, bigquery, datasetId);
+    } catch(Exception ex) {
+      //ignore
```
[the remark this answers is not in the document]
grishick: good point. I got lost in my own refactoring and left this try-catch behind

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationAcceptanceTest.java
```diff
@@ -187,71 +177,37 @@ private List<JsonNode> retrieveRecordsFromTable(final String tableName, final St
...
-    Runtime.getRuntime()
-        .addShutdownHook(
-            new Thread(
-                () -> {
-                  if (!tornDown) {
-                    tearDownBigQuery();
-                  }
-                }));
+  protected void addShutdownHook() {
+    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
```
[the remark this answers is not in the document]
grishick: Hm.. I thought it was already called after each test, because it overwrites `tearDown`. I'll add `@AfterEach` annotation. The shutdown hooks were there before refactoring and AFAIK, the reason is to execute cleanup even if the test crashes.

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTestUtils.java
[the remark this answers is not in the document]
edgao: I think so? Not aware of any reason why not

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationAcceptanceTest.java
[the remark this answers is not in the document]
edgao: I'm pretty sure junit takes care of running AfterEach methods, even if tests blow up (except maybe in cases where the entire runtime crashes?)

but yeah, don't worry about this if it looks super messy. Sounds like this has been working correctly as a shutdown hook anyway

airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/BigQueryUtils.java
```diff
@@ -70,7 +70,7 @@ public class BigQueryUtils {
       DateTimeFormatter.ofPattern("[yyyy][yy]['-']['/']['.'][' '][MMM][MM][M]['-']['/']['.'][' '][dd][d]" +
           "[[' ']['T']HH:mm[':'ss[.][SSSSSS][SSSSS][SSSS][SSS][' '][z][zzz][Z][O][x][XXX][XX][X]]]");
   private static final String USER_AGENT_FORMAT = "%s (GPN: Airbyte)";
-  private static final String CHECK_TEST_DATASET_SUFFIX = "_airbyte_check_stage_tmp_" + System.currentTimeMillis();
```
etsybaev, line 73: Just curious, wouldn't this timestamp remove cause issues during simultaneous tests execution? Ex. same tests for different  PRs, local run, etc?

etsybaev (approved):

airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/BigQueryUtils.java
ryankfu: This shouldn't affect anything since it's implemented back below in `checkHashCreateAndDeleteDatasetRole` and `CHECK_TEST_DATASET_SUFFIX` is only called within that method

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryGcsDestinationAcceptanceTest.java
```diff
@@ -4,59 +4,54 @@
...
-        .put(BigQueryConsts.LOADING_METHOD, loadingMethod)
-        .build());
-
-    setupBigQuery(bigqueryConfigFromSecretFile);
+  protected void tearDown(TestDestinationEnv testEnv) {
+    tearDownBigQuery();
+    tarDownGcs();
   }
 
+  protected void tarDownGcs() {
```
ryankfu, line 54: this probably should be `tearDownGcs`

airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/BigQueryUtils.java
grishick: The reason I moved the timestamp from class instantiation time to test execution time is that it was causing name collision when tests run too fast.

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTestUtils.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /**
+   * Remove all the GCS output from the tests.
+   */
+  public static void tearDownGcs(AmazonS3 s3Client, JsonNode config, Logger LOGGER) {
+    if(s3Client == null) {
+      return;
+    }
+    if(BigQueryUtils.getLoadingMethod(config) != UploadingMethod.GCS) {
```
ryankfu, line 106: This `if` check can probably be grouped together with the preceding `if` statement. Also just to make sure this is clear, is the reason for these empty `return` basically to say we shouldn't be logging any warnings? From what I can understand, it seems that if these values don't exist then it will log a warning in the `try-catch` block

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTest.java
```diff
@@ -108,106 +108,142 @@ class BigQueryDestinationTest {
...
+    if (!Files.exists(CREDENTIALS_STANDARD_INSERT_PATH)) {
       throw new IllegalStateException(
-          "Must provide path to a big query credentials file. By default {module-root}/config/credentials.json. Override by setting setting path with the CREDENTIALS_PATH constant.");
+          "Must provide path to a big query credentials file. By default destination-bigquery/secrets/credentials-standard.json");
+    }
+    if (!Files.exists(CREDENTIALS_WITH_MISSED_CREATE_DATASET_ROLE_PATH)) {
+      throw new IllegalStateException("""
+                                      Json config not found. Must provide path to a big query credentials file,
+                                       please add file with creds to
+                                      <...>/destination-bigquery/secrets/credentials-with-missed-dataset-creation-role.json.""");
```
ryankfu, line 159: Might be a good place to use the constants in a string format instead of having the values be re-type. Something like
```java
throw new IllegalStateException(String.format("""
                                      Json config not found. Must provide path to a big query credentials file,
                                       please add file with creds to
                                      <...>/destination-bigquery/%s""", 
                                      CREDENTIALS_WITH_MISSED_CREATE_DATASET_ROLE_PATH);
```

That said, it appears this already exists in tests so it's a minor nit

ryankfu (approved):
Overall, looks good and the usage of config files cleans up a lot of the overhead from before too. Thanks for taking on this change

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTest.java
grishick: Good catch. I should actually remove all of this copy pasta and just iterate over the array of paths.

octavia-approvington:

octavia-approvington (approved):

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTest.java
```diff
@@ -108,106 +110,128 @@ class BigQueryDestinationTest {
...
+        Arguments.of("configWithBadProjectId", "User does not have bigquery.datasets.create permission in project"),
+        Arguments.of("insufficientRoleConfig", "Permission bigquery.tables.create denied")
+    );
+  }
+
+  @BeforeAll
+  public static void beforeAll() throws IOException {
+    for(Path path : ALL_PATHS) {
+      if (!Files.exists(path)) {
+        throw new IllegalStateException(
```
jbfbell, line 155: Small thing - I believe this exception will be thrown on the first file to not exist, then exit the loop. It might be more convenient if all missing files are shown in the error message.

```
final List<Path> missingPaths = ALL_PATHS.stream().filter(path -> !Files.exist(path)).collect(Collectors.toList());
if (!missingPaths.isEmpty()) {
  throw new IllegalStateException(
  // ...
  )
}

```

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTest.java
```diff
@@ -108,106 +110,128 @@ class BigQueryDestinationTest {
...
-        .put(BigQueryConsts.CONFIG_PROJECT_ID, projectId)
-        .put(BigQueryConsts.CONFIG_CREDS, credentialsJson.toString())
-        .put(BigQueryConsts.CONFIG_DATASET_ID, datasetId)
-        .put(BigQueryConsts.CONFIG_DATASET_LOCATION, DATASET_LOCATION)
-        .put(BIG_QUERY_CLIENT_CHUNK_SIZE, 10)
-        .build());
-
-    tornDown = false;
-    addShutdownHook();
+    configs  = new HashMap<String, JsonNode>() {{
```
jbfbell, line 197: Just curious if there was motivation to remove the `ImmutableMap.builder()` flow here?

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTest.java
```diff
@@ -108,106 +110,128 @@ class BigQueryDestinationTest {
...
-      if (!tornDown) {
-        tearDownBigQuery();
-      }
-    }));
+  protected void initBigQuery(JsonNode config) throws IOException {
+    bigquery = BigQueryDestinationTestUtils.initBigQuery(config, projectId);
+    try {
+      dataset = BigQueryDestinationTestUtils.initDataSet(config, bigquery, datasetId);
+    } catch(Exception ex) {
+      //ignore
```
jbfbell, line 212: I think we should at least log a warning here, otherwise if an exception occurs at this point it would be difficult to debug.

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTest.java
```diff
@@ -374,11 +322,17 @@ void testWriteFailure(final DatasetIdResetter resetDatasetId) throws Exception {
   }
 
   private Set<String> fetchNamesOfTablesInDb() throws InterruptedException {
+    if(dataset == null || bigquery == null) {
```
jbfbell, line 325: I think the two conditionals on lines 325-327 and 333-335 can be combined. The `QueryJobConfiguration` AFAIK is just building a query rather than executing it. Still a little rusty on my Java but I think an `Optional` would be good here

```
Optional<Dataset> potentialDataset = Optional.ofNullable(dataset);
if (potentialDataset.filter(d -> !d.exists()).orElse(false) || bigquery == null) { 
  return Collections.emptySet();
}
  
```

jbfbell (commented):
Left a few comments but mostly LGTM

PR #20852, opened 2022-12-23
Exclude Era date from common v1 dataset

### LINKED ISSUE #20824: Exclude unsupported value types from V1 test cases
- NaN
- Infinity
- (- Infinity)
- Era dates (BC/AD)

suhomud:
## What
Era dates is not supported by all destinations. The era dates should be checked by connectors which supports it
## How
Exclude Era date from common dataset. Add separate tests for era date with https://github.com/airbytehq/airbyte-internal-issues/issues/1253

## Recommended reading order
1. `x.java`
2. `y.python`

## 🚨 User Impact 🚨
No Impact.


suhomud:
Checked destination-s3 tests since it used v1 datasets

grishick:
If a connector does not support eras in dates and a source emits a data with an era - what is the expected behavior?

suhomud:
> If a connector does not support eras in dates and a source emits a data with an era - what is the expected behavior?

@grishick Currently it leads normalization failure. I expect that we need to solve the issue in a scope of handling "bad data" tickets.

grishick (approved):

PR #20853, opened 2022-12-23
OSS changes to support acceptance test on data plane

xiaohansong:
## What
To support data plane acceptance test environment, so PR https://github.com/airbytehq/airbyte-cloud/pull/3913 can build

https://github.com/airbytehq/airbyte-cloud/issues/3014

## How
- added related functions to test harness
- bypass auth logic when data plane trying to call control plane;
- add noop secret hydrator because we haven't setup secret system on data plane local test environment


airbyte-workers/src/main/java/io/airbyte/workers/config/SecretPersistenceBeanFactory.java
```diff
@@ -68,6 +69,15 @@ public SecretPersistence vaultSecretPersistence(@Value("${airbyte.secret.store.v
   }
 
   @Singleton
+  @Requires(property = "airbyte.acceptance.test.enabled",
+            value = "true")
+  public SecretsHydrator noOpSecretsHydrator() {
```
davinchia, line 74: can you explain more how we plan to use this? do we want to use this instead of vault?

xiaohansong: For now yes. It enables us to test on basic e2e sync on data plane, while on control plane it still has postgres connection to test. (maybe i should rename the property name to be data-plane.test) 

I think it might take some time for me to figure out how to setup vault on k8s - once enabled I'll change this bean to vault secret hydrator.

davinchia: Got it.

I'm trying to understand how the data plane would hydrate secrets then. In this setup, would it reach out directly to the DB till we set up Vault?

xiaohansong: It won’t. Noop secret hydrator did not have any external dependency thus it won’t call DB nor vault. Thus, if the connection requires hydrator to fill in secrets it would fail - but for using e2e source and destination as setup for connection it won’t need this at all - all configuration will be passed in as JSON node and hydrator could simply return the input as “no-op”.

davinchia: Oh ok, got it. I missed that we are using the E2E tests. We should add a comment in the test cases in Cloud to make it clear. Thanks for explaning.

davinchia (approved):

PR #20862, opened 2022-12-23
🪟🎉 Connector builder: Add manual schema

### LINKED ISSUE #20834: Add user-editable schema in the Connector Builder UI

flash1293:
Fixes #20834

## What

Adds the ability to specify in-schemas for streams
Screenshot 2022-12-23 at 16 07 37
Screenshot 2022-12-23 at 16 07 46
Screenshot 2022-12-23 at 16 07 52

## How

* Add a new tab to the stream view
* Stores the entered string in the form values
* Validates whether the string can be parsed as json
* add the schema to the manifest if it is parseable
* Show the error indicator in the tab navigation if the schema is invalid (the idea is that the schema tab is like a separate input field so it's showing errors as well. The configuration tab does not have a separate error indicator as it's the "main view")


airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
```diff
@@ -69,53 +89,101 @@ export const StreamConfigView: React.FC<StreamConfigViewProps> = ({ streamNum })
...
+  </button>
+);
+
+const SchemaEditor = ({ streamFieldPath }: { streamFieldPath: (fieldPath: string) => string }) => {
+  const [field, meta, helpers] = useField<string | undefined>(streamFieldPath("schema"));
+
+  return (
+    <>
+      <CodeEditor
+        height="500px"
```
lmossman, line 174: I don't love that this always has a static height of 500px, so I tried making some changes to make it take up the full height instead.

I got it working and put up the changes in a separate PR that is branched off of this one: https://github.com/airbytehq/airbyte/pull/21000/files
feel free to merge that into here if you agree with the approach

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
```diff
@@ -48,11 +54,25 @@ export const StreamConfigView: React.FC<StreamConfigViewProps> = ({ streamNum })
...
+        <StreamTab
+          label={formatMessage({ id: "connectorBuilder.streamConfiguration" })}
+          selected={selectedTab === "configuration"}
+          onSelect={() => setSelectedTab("configuration")}
+        />
+        <StreamTab
+          label={formatMessage({ id: "connectorBuilder.streamSchema" })}
+          selected={selectedTab === "schema"}
+          onSelect={() => setSelectedTab("schema")}
+          showErrorIndicator={hasSchemaErrors}
```
lmossman, line 74: nit: should we also show this error indicator on the configuration tab if there are any errors in there? Maybe not necessary since we already show it on the sidebar

lmossman (approved):
See my few comments above. Neither fully blocking. Tested locally and looks good to me

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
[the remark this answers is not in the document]
flash1293: Restructured this a bit and I think I found an acceptable way to handle this and the comment about showing an error indicator for the configuration tab. It's always selecting the view that has errors if there is one by default.

flash1293: Really nice improvement!

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
flash1293: Really nice improvement!

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
```diff
@@ -23,14 +28,22 @@ interface StreamConfigViewProps {
...
+  const streamFieldPath = (fieldPath: string) => `${streamPath}.${fieldPath}`;
+
   const { formatMessage } = useIntl();
-  const [field, , helpers] = useField<BuilderStream[]>("streams");
+  const [field, meta, helpers] = useField<BuilderStream[]>("streams");
+  const currentStreamErrors = meta.error?.[streamNum] as FormikErrors<BuilderStream>;
+  const hasSchemaErrors = Boolean(currentStreamErrors?.schema);
+  const hasConfigErrors = Boolean(
+    Object.keys(currentStreamErrors || {}).filter((errorKey) => errorKey !== "schema").length > 0
+  );
```
lmossman, line 40: One small issue with just looking at `meta.error` directly is that fields which have errors but are untouched cause the error indicator to show on the tab, but the error is. not shown on the field because it is untouched.

Could we use the `useBuilderErrors` custom hook instead to ignore untouched fields, so that the error indicator is only ever shown if the error is also being shown on the field?

lmossman:
FYI I pushed up a commit here (https://github.com/airbytehq/airbyte/pull/20862/commits/7088b2c29d7cd84c1042baff0b8238c3c08c32bd) - this prevents the builder card containing the schema editor from being resized when the error message is shown or hidden. 

I did this because with the other change I made to make the editor take up the full height of the view, this was causing the builder card to grow past the bottom of the screen when the error message was shown. Feel free to change if you dont agree with the approach

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
flash1293: Ah, right, forgot about that. We can't use useBuilderErrors directly because it can't split out the `schema` bit so we can have separate error indicators for the configuration and the schema tabs. I could try to get that into the existing logic, but we already have the issue to refactor this to make it easier to work, I would rather like to do that first.

I reverted my attempt from yesterday and going to merge like that for now and open a follow-up PR.

PR #20877, opened 2022-12-23
Source Okta: OAuth2.0 support - disabled

lazebnyi:
## What
OAuth 2.0 temporarily disabled until full we get public account. - https://airbytehq.slack.com/archives/C02U9R3AF37/p1671731021620059

## How
Deleted part of spec implementation.


davydov-d (approved):

bazarnov (approved):

alafanechere:
@lazebnyi could you please share more details about the motivation of this PR (I can't access the slack conversation). What will be the impact on existing cloud connections?

lazebnyi:
@alafanechere During Okta certification to Beta we implement OAuth, but now we has problem with getting public application. So will be decided hide Okta OAuth flow from cloud (leave only credentials manual input) to avoid issue during authentication with common OAuth flow. 

I also refactored the acceptance tests to high strictness for speed up future certification.

alafanechere:
Thanks for the context. @lazebnyi does it mean that Okta OAuth is not working on cloud (because we don't have a public application) and its safe to disable it?

lazebnyi:
@alafanechere 
Okta OAuth has never worked on the cloud 😅
Yes, it's safe. With this PR, the user will be authenticated with client_id, client_secret and refresh_token or with an API token.

alafanechere (approved):
Thanks for the explanation and for setting high test strictness level!

evantahler (commented):
Do we need to remove any oAuth secrets from Cloud?  If users have previously used oAuth + Okta will the connector still try to use that method of connection from the config, even though the SEPC option is removed?

(edit) Looks like we never got the app authorized, so this is unlikely.  Carry on!

lazebnyi:
@evantahler we remove only part what get user authenticate with common OAuth flow (enter creds to provider and approve scopes). So,  if user already authenticate using OAuth flow he has refresh_token and can refresh it with connector authentication logic (but it's not possible because we never has public application for our account, so OAuth authentication with UI never work) 

We never had an public account for Okta, so we no need delete any oAuth secrets from Cloud.

evantahler (approved):

pedroslopez:
@evantahler / @lazebnyi - I do see we had configured global OAuth parameters in cloud prod for Okta. In the past we've seen leaving the OAuth parameters in place cause issues when removing OAuth support (e.g. PayPal Transactions pulling in the same data for all users), so whenever we decide to remove OAuth support we should also make sure to remove the global OAuth parameters that may have been set.

Since this has now been released to cloud, I have removed the global OAuth parameters for Okta from the Cloud prod db.

PR #13975, opened 2022-06-21
🎉 Destination BigQuery, Redshift, Snowflake: File buffer increase limit

adam-bloom:
## What
This addresses airbytehq/airbyte-internal-issues#496. Currently, the number of file buffers is limited to 10. If you have more than 10 streams on a source that performs an incremental sync (i.e. does not send messages in stream order, but in CDC order), the destination performance suffers immensely as the destination must repeatedly perform small flushes as it receives incremental updates.

## How
Rather than hard limiting the number of files to 10, limit the number of files to the number of expected streams. The total disk space used by all buffers remains capped at 1 GB - more than that should trigger a flush in the SerializedBufferStrategy. There doesn't seem to be a good reason to limit the number of files to 10 when we have a total size limit as well.

Note: I only performed acceptance testing of Redshift, as our setup doesn't use bigquery or snowflake. I'll update all 3 changelogs and version numbers though.

## Recommended reading order
1. `x.java`
2. `y.python`

## 🚨 User Impact 🚨
No breaking changes, but sync times for incremental sources with more than 10 streams should drop dramatically due to the improved buffering behavior. As an example, the below image shows 3 syncs on the same source. The top one used a connector with this change. Way more throughput in about the same time.

image

## Pre-merge Checklist
Expand the relevant checklist and delete the others.

<details><summary><strong>Updating a connector</strong></summary>

### Community member or Airbyter

- [x] Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
- [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [ ] Documentation updated
    - [x] Connector's `README.md`
    - [ ] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
[...]

PR #15707, opened 2022-08-16
🎉 Destination Redshift: File buffer increase limit

adam-bloom:
## What
See https://github.com/airbytehq/airbyte/pull/13975

## How
Increases the number of buffers that redshift destinations are allowed to use on disk.

## Recommended reading order
1. `x.java`
2. `y.python`

## 🚨 User Impact 🚨
Are there any breaking changes? What is the end result perceived by the user? If yes, please merge this PR with the 🚨🚨 emoji so changelog authors can further highlight this if needed.

## Pre-merge Checklist
Expand the relevant checklist and delete the others.

<details><summary><strong>Updating a connector</strong></summary>

### Community member or Airbyter

- [x] Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
- [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`
    - [ ] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - [ ] Changelog updated in `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
- [x] PR name follows [PR naming conventions](https://docs.airbyte.io/contributing-to-airbyte/updating-documentation#issues-and-pull-requests)

### Airbyter

If this is a community PR, the Airbyte engineer reviewing this PR is responsible for the below items.

- [ ] Create a non-forked branch based on this PR and test the below items on it
- [ ] Build is successful
- [ ] If new credentials are required for use in CI, add them to GSM. [Instructions](https://docs.airbyte.io/connector-development#using-credentials-in-ci).
- [ ] [`/test connector=connectors/<name>` command](https://docs.airbyte.io/connector-development#updating-an-existing-connector) is passing
[...]

PR #20879, opened 2022-12-26
Destination Redshift: Introduces configurable value for file buffer count

ryankfu:
## What
Follows discussion in https://github.com/airbytehq/airbyte/pull/13975 and partially addresses https://github.com/airbytehq/airbyte-internal-issues/issues/496 which asserts that interleaved messages (e.g. Change Data Capture aka CDC) trigger frequent `flush all` commands which reduces performance since buffers will thrash due to the constant need to switch buffers for different streams

This change doesn't _fully_ address those tickets but introduces work that a community member has introduced which is a user configurable field for concurrent buffers to be created when a `MessageConsumer` is created for destinations. This change follows a discussion that mentions that each buffer requires a minimum of 31 MB and having an unbounded number of buffers is not feasible primarily because the amount of memory would need to scale and since Airbyte aims to not break existing customer's workflow this number is increased with that consideration in mind

## How
Creates the ability for a user to exceed the default within the hard cap limit. If the value exceeds the soft cap then the user will be prompted with a warning indicating the reason to exceed grants additional performance so as long as the increase does not exceed the number of streams to be synced

## Recommended reading order
1. `RedshiftStagingS3Destination.java`
2. `FileBuffer.java`
3. `RedshiftStagingS3DestinationTest.java`

## 🚨 User Impact 🚨

This should not have any breaking changes as the increase in `FileBuffer` is within the number of memory that is allocated generally across all users (1 GB) and the soft cap is within that amount even with "wiggle" room

## Pre-merge Checklist

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>

</details>

## Tests

</details>

</details>

</details>


airbyte-integrations/connectors/destination-redshift/src/main/resources/spec.json
```diff
@@ -206,6 +206,15 @@
                     }
                   }
                 ]
+              },
+              "file_buffer_size": {
+                "title": "File Buffer Count",
+                "type": "integer",
+                "minimum": 15,
```
grishick, line 213: the new test checks that this value can be set as low as `10`, but the spec says that minimum us `15`. Should the two minimums be the same?

ryankfu: Good catch. Yes, the minimums should be the same. I'll add on a test and change that ensures at minimum there are `15` file buffers

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/record_buffer/FileBuffer.java
```diff
@@ -25,14 +25,33 @@ public class FileBuffer implements BufferStorage {
...
+   *
+   * Note: This value can be tuned to increase performance with the tradeoff of increased memory usage
+   * (~31 MB per buffer). See {@link StreamTransferManager}
+   *
+   * For connections with interleaved data (e.g. Change Data Capture), having less buffers than the
+   * number of streams being synced will cause buffer thrashing where buffers will need to be flushed
+   * before another stream's buffer can be created. Increasing the default max will reduce likelihood
+   * of thrashing but not entirely eliminate unless number of buffers equals streams to be synced
+   */
+  public static final int DEFAULT_MAX_CONCURRENT_STREAM_IN_BUFFER = 15;
```
grishick, line 46: We are changing the default and minimum from 10 to 15. This raises two questions:
 - How this will affect existing connections? 
 - How this will affect non-cloud users, who may be running this connector in containers with less than 1GB RAM?

ryankfu: * The likelihood of this affecting existing connections will be low since the minimum memory needed for all the default buffers will be 465 MB (15 * 31 MB) which is strictly less than the 1 GB `MAX_TOTAL_BUFFER_SIZE_BYTES` 
* Furthermore, if a customer had low memory availability then they would have run into issues earlier since our `MAX_TOTAL_BUFFER_SIZE_BYTES` was 1 GB before a flush. To not run into issues prior, users would have had to tune the original `DEFAULT_MAX_CONCURRENT_STREAM_IN_BUFFER` or lower the `MAX_TOTAL_BUFFER_SIZE_BYTES` to the amount of memory available for the JVM

It is reasonable to have within the change-log to mark this change as potentially breaking to users with memory less than 500 MB

grishick: AFAIK, we used to run containers with 500MB by default and recently switched to 1GB. Why not leave minimum as `10` instead?

ryankfu: The `MAX_TOTAL_BUFFER_SIZE_BYTES` has been 1 GB since March so most of this decision is based on that value. Didn't realize that you already had a comment about this [here](https://github.com/airbytehq/airbyte/pull/15707#discussion_r1040178169) in the community PR but it makes sense to reduce plausibilities of breakages. It begs the question of how we communicate better with our community about these increases. It seems like allowing it to be configurable is one option with proper warnings

Based on your comment, it seems better to just have the default maintain `10` with the option to configure for the user

grishick (approved):

ryankfu:
/publish connector=connectors/destination-redshift run-tests=false

ryankfu:
/publish connector=connectors/destination-redshift run-tests=false

adam-bloom:
Hi @ryankfu - I found a typo in this that prevent it from working correctly - buffers are still capped at 10.

https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/destination-redshift/src/main/java/io/airbyte/integrations/destination/redshift/RedshiftStagingS3Destination.java#L178 - this max is to the default, which is still 10. It should be a max with `FileBuffer.MAX_CONCURRENT_STREAM_IN_BUFFER`

Additionally, the default of 1 [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/destination-redshift/src/main/java/io/airbyte/integrations/destination/redshift/RedshiftStagingS3Destination.java#L173) doesn't seem correct - that should likely be `FileBuffer.DEFAULT_MAX_CONCURRENT_STREAM_IN_BUFFER`

adam-bloom:
Also - there's a discrepancy between the config keys used in the [spec](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/destination-redshift/src/main/resources/spec.json#L210) and the [application](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/record_buffer/FileBuffer.java#L47), so I don't think its even possible to configure this correctly.

ryankfu:
@adam-bloom thank you for pointing out the mismatched `spec` and `application` values, I will spin up a PR to address this

As for the buffers being capped at `10`, that should not be the case since 
```java
return Math.max(numOfFileBuffers, FileBuffer.DEFAULT_MAX_CONCURRENT_STREAM_IN_BUFFER);
```

will return the larger of `numOfFileBuffers` or `DEFAULT_MAX_CONCURRENT_STREAM_IN_BUFFER` (`10`)

In fact there's a test that captures this: https://github.com/airbytehq/airbyte/blob/0e3d2ba43101ab00df4d294ad1a6c06b6c2f95d2/airbyte-integrations/connectors/destination-redshift/src/test-integration/java/io/airbyte/integrations/destination/redshift/RedshiftStagingS3DestinationAcceptanceTest.java#L126-L130

adam-bloom:
I think it was really the config mismatch that I was seeing the impacts of. The 10 would have been coming from max(1, 10). Just needed a second pass!

PR #20883, opened 2022-12-27
Add feature to specify own selector and regular labels in deployment

xpuska513:
## What
* Add 2 new values `extraLabels` and `extraSelectorLabels` so users can specify their own labels for better resource management or monitoring


supertopher (approved):
Sanity check approval.  

I think this is a good change for us and OSS. users

PR #20893, opened 2022-12-27
[Low-Code CDK] Handle forward references in manifest

### LINKED ISSUE #20503: [Low-Code CDK] Support forward references in manifest
When a low-code CDK user creates their manifest file, they have limited flexibility with the layout because the parser does not support forward references - i.e. each reference must be defined above the reference's location in the file, or else an `UndefinedReferenceException` exception is raised. From the developer's point of view this means that you may not be able to organize the file in a way that is readable.

The specific example that highlights this issue is the new `schemas` key. When schemas are defined inline in the manifest, they're bulky and long - it's ideal to have them at the end of the file, but that's not currently possible.

Updating the manifest preprocessing code to support forward references.

The current code stores a temporary data structure, `evaluated_mapping` that collects manifest information that may be referenced later in the manifest. Instead, we can look up references by reading the value(s) at the referenced path in the raw manifest.

We could leave this as-is and assume that most people will have limited interaction with the yaml file. However, we expect that users are going to be toggling between the form view and the yaml even after the form is available, so it seems that the fact that a human will be looking at the yaml file isn't going to go away.

clnoll:
## What
Allow users to organize yaml manifests with more flexibility, by handling forward references.

Closes https://github.com/airbytehq/airbyte/issues/20503

## How
Updates the `ManifestReferenceResolver` class to look up references by reading the value(s) at the referenced path in the manifest (rather than looking for values in previously evaluated keys).

## Recommended reading order
1. `manifest_reference_resolver.py`
2. `manifest_declarative_source.py`
3. `custom_exceptions.py`
4. Tests

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
@@ -2,10 +2,10 @@
 # Copyright (c) 2022 Airbyte, Inc., all rights reserved.
 #
 
-from copy import deepcopy
+import re
 from typing import Any, Mapping, Tuple, Union
 
-from airbyte_cdk.sources.declarative.parsers.undefined_reference_exception import UndefinedReferenceException
+from airbyte_cdk.sources.declarative.parsers.custom_exceptions import CircularReferenceException, UndefinedReferenceException
 
 
 class ManifestReferenceResolver:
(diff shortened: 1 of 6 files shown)


airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
```diff
@@ -96,91 +96,102 @@ class ManifestReferenceResolver:
...
+    def _evaluate_node(self, node: Any):
+        if isinstance(node, dict):
+            evaluated_dict = {k: self._evaluate_node(v) for k, v in node.items() if not self._is_ref_key(k)}
+            if self.ref_tag in node:
+                # The node includes a $ref key, so we splat the referenced value(s) into the evaluated dict
+                evaluated_ref = self._evaluate_node(node[self.ref_tag])
+                if not isinstance(evaluated_ref, dict):
+                    return evaluated_ref
+                else:
+                    return evaluated_ref | evaluated_dict
```
brianjlai, line 119: If I'm understanding this right, we get the references values, and then update with the values defined on this component. So the values defined on this component take precedence. If so, can we add a small comment saying that values defined on this node take precedence over the referenced values.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
```diff
@@ -96,91 +96,102 @@ class ManifestReferenceResolver:
 
     ref_tag = "$ref"
 
-    def preprocess_manifest(self, manifest: Mapping[str, Any], evaluated_mapping: Mapping[str, Any], path: Union[str, Tuple[str]]):
+    def __init__(self, manifest: Mapping[str, Any]):
```
brianjlai, line 99: I can see why we're going with the approach to have the set and the manifest defined on the factory itself, but it steel feels a little weird for them to be inserted into the factory during instantiation. When I think of a factory I see it as generic and the preprocess method would take input and emit output.

Granted we need to reference the manifest as we traverse the manifest so its probably easier to define it on the factory class itself. 

The other thing that feels weird is that when the `set()` is defined at the start, we're very reliant on correctly popping off visited elements after traversal and that at the end of preprocess the visited set is cleared.

As part of the `preprocess_manifest()` either when starting or finishing a run, we could clear the set so we know on each attempt we start clean. And the same could be said for assigning `self.manifest =` in that method (Although we would probably still instantiate empty values during init?). I'm not married to this approach either, but I think its worth a discussion because it stuck out to me that we're coupling the manifest to the factory pretty tightly in the refactor.

brianjlai (commented):
I'm liking what you have so far and I think the code to parse the layers is more intuitive to understand so nice work! I had a few follow ups related to testing and how we invoke the factory a bit. Just some discussion points to flesh out rather than strict changes.

Also it might be worth it to run the validate Gradle command that Maxime wrote to verify. We should expect them to all pass (although one heads up about source-monday which has been acting a bit weird this week).

airbyte-cdk/python/unit_tests/sources/declarative/test_factory.py
```diff
@@ -65,7 +65,7 @@ def test_factory():
       request_body_json:
         body_offset: "{{ next_page_token['offset'] }}"
     """
-    config = resolver.preprocess_manifest(YamlDeclarativeSource._parse(content), {}, "")
+    config = resolver(YamlDeclarativeSource._parse(content)).preprocess_manifest()
```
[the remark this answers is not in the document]
clnoll: Good call! Added.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
```diff
@@ -96,91 +96,102 @@ class ManifestReferenceResolver:
...
+    def _lookup_reference_value(self, reference: str) -> Any:
+        path = re.match("\\*ref\\(([^)]+)\\)", reference).groups()[0]
+        if not path:
+            raise UndefinedReferenceException(path, reference)
+        try:
+            return self._read_reference_value(path, self.manifest)
+        except (KeyError, IndexError):
+            raise UndefinedReferenceException(path, reference)
+
+    def _read_reference_value(self, ref: str, manifest_node: Mapping[str, Any]) -> Any:
```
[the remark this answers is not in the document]
clnoll: 👍 done.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
clnoll: Yep that's right. Added a comment.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
clnoll: So if you grep for `ManifestReferenceResolver` we actually only instantiate it and do the preprocessing once - when `ManifestDeclarativeSource` is initialized, and before the component factory is invoked at all. I'm wondering if that changes the discomfort with initializing it with the manifest + `visited` set. From there on out, the fully resolved config is accessed via `ManifestDeclarativeSource` and no additional calls are made to `preprocess_manifest`. That said if it feels more natural we can certainly consider clearing out the `visited` set after processing, and go back to the original design where we just passed the manifest into `preprocess_manifest`.

clnoll (commented):
Thanks for the review @brianjlai! I've addressed your comments and added some info around the instantiation of `ManifestReferenceResolver`.

clnoll:
@brianjlai forgot to ask - what is the gradle command you're suggesting?

brianjlai:
The command is here:
https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/build.gradle#L22-L25

It basically just attempts to run the `spec` command against all of the existing low code connectors to see if they can be parsed and validated successfully

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
brianjlai: Right so it's not a hard blocker by any means since like you said it's only used once. It's just a little bit of an awkward design. I don't think this part of flow would change, but one of the risks would be if we ever moved to using the factory or resolver multiple times, if we don't pop every element off, we might end up with some unexpected results accidentally using a set w/ stale elements.

How much lift is it to pass the manifest to the resolver like we used to be doing? The part that I noticed in your new implementation was that we use the `self.manifest` field in `_lookup_reference_value()`. If the lift isn't too big to continue the pass in the manifest and start w/ a clean execution of the function every time then I think that feels more intuitive. That being said, if trying to accommodate this makes the function too confusing, I think it is fine to leave as you already had it.

clnoll: It's no trouble at all - the code was in that state in one iteration, but I'd modified it thinking that it would be a little cleaner this way since I was having to pass `manifest` around as an argument in so many places. But to protect future-us from issues if we use the resolver multiple times I'm happy to change it back!

clnoll:
Thanks @brianjlai, the `gradle` command passed.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
clnoll: Pushed this change.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
```diff
@@ -96,91 +96,102 @@ class ManifestReferenceResolver:
 
     ref_tag = "$ref"
 
-    def preprocess_manifest(self, manifest: Mapping[str, Any], evaluated_mapping: Mapping[str, Any], path: Union[str, Tuple[str]]):
-
+    def preprocess_manifest(self, manifest):
```
brianjlai, line 99: nit: can you add type hints and outputs to the method signature

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
```diff
@@ -96,91 +96,102 @@ class ManifestReferenceResolver:
...
-            ]
-            # Add the list's element to the evaluated config so they can be referenced
-            for index, elem in enumerate(evaluated_list):
-                evaluated_config[self._get_path_for_list_item(path, index)] = elem
-            return evaluated_list
+                return evaluated_dict
+        elif isinstance(node, list):
+            return [self._evaluate_node(v, manifest) for v in node]
+        elif isinstance(node, str) and node.startswith("*ref("):
+            if visited is None:
```
brianjlai, line 122: I think its a bit more clear to instantiate the empty set and pass it as a parameter which when we first invoke the first `_evaluate_node()` rather than rely on this conditional block to set it up. Also given how its used, in the flow it feels more like a required parameter instead of defaulting to `None`:

```
return self._evaluate_node(manifest, manifest, set())
```

What do you think?

brianjlai (approved):
Glad to see it running through the validation script. Just a couple of small suggestions and nits. Nice work!

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
clnoll: @brianjlai sorry I overlooked these before merging! Good suggestions, I'll make these changes in a separate PR.

clnoll: (https://github.com/airbytehq/airbyte/pull/21268)

PR #20894, opened 2022-12-28
process config control messages during `check` and `discover`

### LINKED ISSUE #20145: Update configuration from control messages in synchronous commands
Following https://github.com/airbytehq/airbyte/issues/17909, which introduced the capability for the platform to update connector configuration during syncs, we also want to update configuration during the other commands (e.g. check, discover)

We can use the UpdateConnectorConfigHelper introduced in the previous PR to perform the updating, which already handles things like masking secrets.

This will involve initializing the AirbyteApiClient in airbyte-server, as well as updating inputs to the jobs as required to have the right information for calling the endpoints (e.g. source or destination id)

pedroslopez:
## What

Process config control messages in `check` and `discover` calls to support connectors with short-lived/single-use refresh tokens.

closes airbytehq/airbyte#20145

## How

- Call ConnectorConfigUpdater introduced in airbytehq/airbyte#19811 from `DefaultCheckConnectionWorker` / `DefaultDiscoverCatalogWorker` to update the connector config if a config control message has been found. 
    - For `discover`, use the existing `sourceId` from the input.
    - For `check`, add `actorType` and `actorId` to the input so we can call updateSource/updateDestination accordingly. 

- In order to support refreshing tokens during `check_connection_for_update`, `sourceId` and `destinationId` are now being passed through. (see caviat in "User Impact" section)

image

## 🚨 User Impact / Known Issues 🚨

Whenever a connector emits a config control message during `check` or `discover`, the airbyte platform will now persist the updated configuration.

### Configuration updates via UI

There's an interesting side effect here that can happen when users are updating configuration via the UI. Before saving the new configuration, we perform a `check` on the configuration. If the `check` fails, we do not persist the user's changes. After this PR, if the connector issues a token refresh during that preliminary `check`, we will persist the new configuration. This means that a failed check may now cause your config to be updated with invalid values.

The following diagram illustrates the happy path of CHECK succeeding:

image

The above scenario works as expected and the updated config + new tokens emitted during `CHECK` are saved. Note that we are assuming that the connector updated `secret` fields here that have not been modified by the user in the form. Non-secret field values would be overwritten by the user's provided values in the later `update` call (see airbytehq/airbyte#20913)

Here's what happens in the case of CHECK failing or throwing an error:

image

Because we want to save the new token values even on a failed check (since old tokens are invalidated), this creates a side effect of updating the configuration with the values that led to the failed check. This means that if you had a working connector before but started making some changes to the configuration that rendered it invalid, now your connector isn't properly configured. Before this PR, changes leading to a failed check would just not be persisted and your previously working configuration would still be saved. However, if we didn't do this then every failed check would need a manual re-authentication. 

### Connector creation with eager token refreshes

When first setting up a connector, the `check` call happens before the connector has been saved in the database. This means that we _can't_ update the configuration at this phase. If the connector performs a token refresh and invalidates the old tokens, we have no way of saving these currently. See https://github.com/airbytehq/airbyte-internal-issues/issues/1260 for more details.


airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/TemporalClient.java
```diff
@@ -349,7 +349,10 @@ public TemporalResponse<ConnectorJobOutput> submitCheckConnection(final UUID job
         .withDockerImage(config.getDockerImage())
         .withProtocolVersion(config.getProtocolVersion())
         .withIsCustomConnector(config.getIsCustomConnector());
-    final StandardCheckConnectionInput input = new StandardCheckConnectionInput().withConnectionConfiguration(config.getConnectionConfiguration());
+    final StandardCheckConnectionInput input = new StandardCheckConnectionInput()
+        .withActorType(config.getActorType())
+        .withActorId(config.getActorId())
```
evantahler, line 354: I'm amazed we got this far without sending this information!

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultCheckConnectionWorker.java
```diff
@@ -84,6 +89,26 @@ public ConnectorJobOutput run(final StandardCheckConnectionInput input, final Pa
           .map(AirbyteMessage::getConnectionStatus)
           .findFirst();
 
+      if (input.getActorId() != null && input.getActorType() != null) {
+        LOGGER.info("has actor and type");
```
evantahler, line 93: can this log be removed?

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultCheckConnectionWorker.java
```diff
@@ -84,6 +89,26 @@ public ConnectorJobOutput run(final StandardCheckConnectionInput input, final Pa
...
+        final Optional<AirbyteControlConnectorConfigMessage> optionalConfigMsg = WorkerUtils.getLastConfigControlMessage(messagesByType);
```
evantahler, line 94: It looks like `messagesByType` will store _all_ the AirbyteMessages we get from the container, arranged by type.  Not that you added this in this PR, but that _could_ be huge.  I guess since this is the CHECK method, it's not likely to grow that big...

airbyte-commons-worker/src/main/java/io/airbyte/workers/WorkerUtils.java
```diff
@@ -109,6 +111,16 @@ public static WorkerDestinationConfig syncToWorkerDestinationConfig(final Standa
         .withState(sync.getState());
   }
 
+  public static Optional<AirbyteControlConnectorConfigMessage> getLastConfigControlMessage(final Map<Type, List<AirbyteMessage>> messagesByType) {
```
evantahler, line 114: nit: `last` could be ambiguous.  Perhaps `getMostRecentConfigControlMessage` could be clearer?

airbyte-workers/src/main/java/io/airbyte/workers/temporal/check/connection/CheckConnectionActivityImpl.java
```diff
@@ -88,9 +89,12 @@ public ConnectorJobOutput runWithJobOutput(final CheckConnectionInput args) {
     ApmTraceUtils
         .addTagsToTrace(Map.of(ATTEMPT_NUMBER_KEY, args.getJobRunConfig().getAttemptId(), JOB_ID_KEY, args.getJobRunConfig().getJobId(),
             DOCKER_IMAGE_KEY, args.getLauncherConfig().getDockerImage()));
-    final JsonNode fullConfig = secretsHydrator.hydrate(args.getConnectionConfiguration().getConnectionConfiguration());
+    final StandardCheckConnectionInput rawInput = args.getConnectionConfiguration();
+    final JsonNode fullConfig = secretsHydrator.hydrate(rawInput.getConnectionConfiguration());
```
evantahler, line 93: 👍 using the secrets tooling

airbyte-workers/src/test/java/io/airbyte/workers/general/DefaultCheckConnectionWorkerTest.java
```diff
@@ -88,9 +105,65 @@ void testEnums() {
...
+  }
+
+  @Test
+  void testCheckConnectionWithConfigUpdateSource() throws WorkerException {
+    final DefaultCheckConnectionWorker worker =
+        new DefaultCheckConnectionWorker(integrationLauncher, connectorConfigUpdater, configMessageStreamFactory);
     final ConnectorJobOutput output = worker.run(input, jobRoot);
 
+    verify(connectorConfigUpdater).updateSource(ACTOR_ID, CONNECTOR_CONFIG);
+    verifyNoMoreInteractions(connectorConfigUpdater);
```
evantahler, line 128: I'm having trouble reading this... does this mean that you are asserting that attempting to update the config again does nothing because the values already have been updated within the check?

evantahler (approved):
👍 from me with some minor questions and suggestions

airbyte-commons-worker/src/main/java/io/airbyte/workers/WorkerUtils.java
pedroslopez: 👍🏻 done in [d94feb1](https://github.com/airbytehq/airbyte/pull/20894/commits/d94feb1d439f6aef4e876428965b1534cf5f3d65)

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultCheckConnectionWorker.java
pedroslopez: Oops! Done in [89622f6](https://github.com/airbytehq/airbyte/pull/20894/commits/89622f620e04088c24e9df7735fc74b2a2ab3442)

airbyte-workers/src/test/java/io/airbyte/workers/general/DefaultCheckConnectionWorkerTest.java
pedroslopez: `verifyNoMoreInteractions` just checks that we're not making any other calls to the connectorConfigUpdater. Since the worker can choose to use updateSource or updateDestination based on the type of actor I just wanted to verify that the other method isn't getting called.

pedroslopez:
Hey @gosusnp - tagging you in form the workflows team since there was a change to the inputs for the check command. These new parameters are optional so I don't for-see any issues with introducing these, but just want to make sure 😄

gosusnp (approved):
Looks good to me.

I have learned recently that modifying workers can impact connectors test, to be sure, you can trigger integration tests for one source and destination.
Left a couple of questions/suggestions.

airbyte-api/src/main/openapi/config.yaml
```diff
@@ -2667,6 +2667,8 @@ components:
         - connectionConfiguration
         - workspaceId
       properties:
+        sourceId:
```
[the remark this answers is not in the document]
pedroslopez: This isn't marked required because checks that happen on initial connector setup don't have these IDs at that point in the flow (they haven't been created yet) - so it's totally valid that checks can happen without a set sourceId / destinationId

airbyte-commons-worker/src/main/java/io/airbyte/workers/WorkerUtils.java
```diff
@@ -109,6 +111,16 @@ public static WorkerDestinationConfig syncToWorkerDestinationConfig(final Standa
         .withState(sync.getState());
   }
 
+  public static Optional<AirbyteControlConnectorConfigMessage> getMostRecentConfigControlMessage(final Map<Type, List<AirbyteMessage>> messagesByType) {
```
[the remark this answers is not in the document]
pedroslopez: Yep! control messages contain the full config object, so the final message should have the latest version we need to update

airbyte-commons-worker/src/main/java/io/airbyte/workers/WorkerUtils.java
```diff
@@ -109,6 +111,16 @@ public static WorkerDestinationConfig syncToWorkerDestinationConfig(final Standa
...
+    final List<AirbyteControlConnectorConfigMessage> configMessages = messagesByType.getOrDefault(Type.CONTROL, new ArrayList<>()).stream()
```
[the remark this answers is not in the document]
pedroslopez: 👍🏻 done in [60987e2](https://github.com/airbytehq/airbyte/pull/20894/commits/60987e21f072b9f1afd67305aeb9816009012e13)

PR #20899, opened 2022-12-28
SAT: check for nullity of docker runner in `previous_discovered_catalog`

alafanechere:
## What
When we create the `previous_discovered_catalog_fixture` we call discover on the `previous_connector_docker_runner` object. This object is `None` if we cannot find a previous connector version. 
The existing fixture code did not check for nullity of the object and the `AttributeError: 'NoneType' object has no attribute 'call_discover` can be raised.

This bug is noticeable now because of the recent change of https://github.com/airbytehq/airbyte/pull/20435 which accesses `previous_discovered_catalog` before checking the nullity of `previous_connector_docker_runner`.

## How
Make the fixture return None if the previous docker runner is None (like we do for `previous_connector_spec_fixture`)

### DIFF
--- a/airbyte-integrations/bases/source-acceptance-test/CHANGELOG.md
+++ b/airbyte-integrations/bases/source-acceptance-test/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.2.24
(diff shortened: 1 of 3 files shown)


evantahler (approved):

evantahler:
There are few pending CDK PRs - just a reminder to make sure the version numbers are right :D

alafanechere:

PR #20903, opened 2022-12-28
Improve Gradle task caching

jdpgrailsdev:
## What
* Improve Gradle builds by ensuring tasks are only executed if changes are detected.

## How
* Remove deprecated task to ensure that there is an unique output directory per task for the API code generation.

## Recommended reading order
1. `build.gradle`

**N.B**:  There are a couple of things to consider with this PR.  The main issue is that there currently exists two tasks that target the same output directory:  one that generates the single API class that contains all endpoints and another that breaks that up into a class per domain.  The former is what the code currently uses and the latter is what we will use once we convert airbyte-server to Micronaut.  This PR should not be merged or re-attempted until we have removed all usages of the single, generated server API class across both code bases.  I am going to mark this as a draft PR for now.


jdpgrailsdev:
Closing in favor of https://github.com/airbytehq/airbyte/pull/23213

PR #20905, opened 2022-12-28
Add some abstraction to avoid direct generated object use

gosusnp:
## What

With the introduction of Airbyte Protocol versioning, direct use of the generated objects comes with some overhead:
* Moving to a new version means patching all the `import` statement of classes that are relying on protocol objects
* Java Connectors are sharing some classes from the platform which led to some non-straight forward dependencies. We have a disconnect since there will be a transition period during which we will have different versions of the protocol for different connectors at the same time.

## How

In order to reduce the coupling:
* we introduce interfaces that describes the different concepts used in the protocol
* we leverage adapters to map the most current version of the protocol to the interface

The key benefits:
* This layer of abstraction gives us more flexibility to decouple our internal logic from the network representation of the objects
* Bumping a version means update the adapter rather all the imports
* This removes the explicit dependencies on the protocol version for the java connectors and allows us to shift the focus towards the information read rather than the container

The tradeoff is that we need to explicitly update the interfaces/adapters to channel new information.

## Recommended reading order
1. `x.java`
2. `y.python`

## 🚨 User Impact 🚨

No behavior change expected, this is targeting dev efficiency.

### DIFF
--- a/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/objects/ConnectorSpecification.java
+++ b/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/objects/ConnectorSpecification.java
@@ -19,11 +19,11 @@ public interface ConnectorSpecification {
 
   JsonNode getConnectorSpecification();
 
-  boolean getSupportsIncremental();
+  Boolean isSupportingIncremental();
 
-  boolean getSupportsNormalization();
+  Boolean isSupportingNormalization();
 
-  boolean getSupportsDBT();
+  Boolean isSupportingDBT();
 
   // TODO should be a new Enum
   List<DestinationSyncMode> getSupportedDestinationSyncModes();
--- a/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/objects/ConnectorSpecification.java
+++ b/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/objects/ConnectorSpecification.java
@@ -7,7 +7,6 @@ package io.airbyte.commons.protocol.objects;
 import com.fasterxml.jackson.databind.JsonNode;
 import io.airbyte.protocol.models.AdvancedAuth;
 import io.airbyte.protocol.models.AuthSpecification;
-import io.airbyte.protocol.models.DestinationSyncMode;
 import java.net.URI;
 import java.util.List;
 
@@ -25,15 +24,16 @@ public interface ConnectorSpecification {
 
   Boolean isSupportingDBT();
 
-  // TODO should be a new Enum
   List<DestinationSyncMode> getSupportedDestinationSyncModes();
 
-  // TODO we should introduce specific interfaces
+  // TODO introduce specific interfaces
   AuthSpecification getAuthSpecification();
 
+  // TODO introduce specific interfaces
   AdvancedAuth getAdvancedAuth();
 
-  // TODO Temp hack to avoid having to migrate all the code to this interface at once
+  // Clients should use this interface rather than the underlying object
+  @Deprecated
   io.airbyte.protocol.models.ConnectorSpecification getRaw();
 
 }
--- a/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/objects/ConnectorSpecification.java
+++ b/airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/objects/ConnectorSpecification.java
@@ -5,12 +5,19 @@
 package io.airbyte.commons.protocol.objects;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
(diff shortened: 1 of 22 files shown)


gosusnp:
@edgao, @akashkulk, in our previous discussions, I mentioned we should look into using interfaces rather than generated objects directly. This PR is trying to illustrate what this could look like.
The goal would be to have an interface to abstract some details of the generated objects. I took the `ConnectorSpecification` case here because it has many touch points currently. Let me know what you guys think, it could be worth taking a stab at abstracting the Records to see how it plays with the data types changes.

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/objects/impl/ConnectorSpecificationAdapter.java
```diff
@@ -0,0 +1,91 @@
...
+import io.airbyte.commons.protocol.objects.DestinationSyncMode;
+import io.airbyte.protocol.models.AdvancedAuth;
+import io.airbyte.protocol.models.AuthSpecification;
+import java.net.URI;
+import java.util.List;
+import java.util.Map;
+import lombok.EqualsAndHashCode;
+
+@EqualsAndHashCode
+public class ConnectorSpecificationAdapter implements ConnectorSpecification {
```
edgao, line 19: checking my understanding: is the idea that we would have an adapter for every protocol version?

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/objects/ConnectorSpecification.java
```diff
@@ -0,0 +1,46 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import io.airbyte.commons.protocol.objects.serde.ConnectionSpecificationDeserializer;
+import io.airbyte.commons.protocol.objects.serde.ConnectionSpecificationSerializer;
+import io.airbyte.commons.protocol.objects.serde.JsonSerializable;
+import io.airbyte.protocol.models.AdvancedAuth;
+import io.airbyte.protocol.models.AuthSpecification;
+import java.net.URI;
+import java.util.List;
+
+@JsonDeserialize(using = ConnectionSpecificationDeserializer.class)
```
edgao, line 18: should these be declared on the adapters? E.g. if the platform is on protocol v42, but an older connector still uses protocol v41, and it has a call to `Jsons.serialize(new ConnectorSpecificationAdapterV41(...))`

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/objects/impl/ConnectorSpecificationAdapter.java
edgao, line 19: also, how sketchy is it to have the generated protocol models directly implement the interfaces? I.e.
* interface declares `default Whatever getWhatever() { return null; }`
* and the generated classes look like `public class AirbyteMessage implements io.airbyte.commons.protocol.objects.AirbyteMessage`

then:
* adding new fields is still easy (just add a new `default` method to the interface)
* deleting fields is still easy (mark the interface method as deprecated, or just delete it entirely)
* renaming fields is... probably fine? (`default Whatever getWhatever() { return getOldWhatever(); }`)

(I forget if the issue was that there's a technical problem, or that the code generator just doesn't support this - if the latter, we could maaaaybe do something similar to https://github.com/LiveRamp/extravagance/ and edit the generated code)

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/objects/serde/ConnectionSpecificationSerializer.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  protected ConnectionSpecificationSerializer(Class<ConnectorSpecification> t) {
+    super(t);
+  }
+
+  @Override
+  public void serialize(ConnectorSpecification value, JsonGenerator gen, SerializerProvider provider) throws IOException {
+    gen.writeStartObject();
+    gen.writeStringField("object", value.toJson());
```
edgao, line 26: which components would receive these `{object: {spec...}}` messages? I.e. do connectors need to produce this format? Or are connectors still expected to produce protocol messages, and then the worker would translate them into this format? (or something else?)

edgao (commented):
took a super quick first pass on this - added a few questions before I take a deeper look

airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/objects/serde/ConnectionSpecificationSerializer.java
gosusnp: I think those should remain within the platform, those objects should be mostly used for temporal payloads which is generated by the clients we use and the activity. It currently leaks to the connector's tests because some tests rely on the workers.
Connectors having to generate those should be a red flag.

davinchia:
Jimmy, going to close this for now. Feel free to move this on to airbyte-platform if it's relevant!

PR #20907, opened 2022-12-28
Remove config repo from refresh schema activity

alovew:
Remove config repository dependency from RefreshSchemaActivity

### DIFF
--- a/airbyte-api/src/main/openapi/config.yaml
+++ b/airbyte-api/src/main/openapi/config.yaml
@@ -617,7 +617,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/ActorCatalog"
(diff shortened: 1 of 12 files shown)


airbyte-api/src/main/openapi/config.yaml
```diff
@@ -599,6 +599,29 @@ paths:
           $ref: "#/components/responses/NotFoundResponse"
         "422":
           $ref: "#/components/responses/InvalidInputResponse"
+  /v1/sources/most_recent_source_actor_catalog:
```
jdpgrailsdev, line 602: @benmoriceau FYI for when you are back.  We will need to ensure this gets ported into the Micronaut branch of airbyte-server

jdpgrailsdev (approved):
:shipit:

PR #20908, opened 2022-12-28
remove config repository from config fetch activity impl scheduling

alovew:
Remove config repo from config fetch activity impl


jdpgrailsdev:
@alovew Looks like there are failing tests in airbyte-workers -- possibly related to all the other merging to master?

jdpgrailsdev (approved):
:shipit:

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/ConfigFetchActivityImpl.java
```diff
@@ -261,4 +264,46 @@ public Optional<ConnectionStatus> getStatus(final UUID connectionId) {
...
 
+  private Long getIntervalInSecond(final ConnectionScheduleDataBasicSchedule schedule) {
+    return getSecondsInUnit(schedule.getTimeUnit()) * schedule.getUnits();
+  }
+
+  private Long getIntervalInSecond(final ConnectionSchedule schedule) {
+    return getSecondsInUnit(schedule.getTimeUnit()) * schedule.getUnits();
+  }
+
+  private Long getSecondsInUnit(final TimeUnitEnum timeUnitEnum) {
```
colesnodgrass, line 275: I wonder if this (and the similar methods surrounding it) should be defined somewhere else.  Don't know where that would be currently, but they seem a little out of play in this class.

colesnodgrass (approved):

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/ConfigFetchActivityImpl.java
alovew: I worked on this with Parker, we did discuss this. the issue is that these methods are currently defined in ScheduleHelpers, but if we wanted to add new methods there that took ConnectionRead instead of StandardSync we'd have to add a new dependency for airbyte-config to depend on airbyte-api. Not sure if possibly a better solution would be to move ScheduleHelpers elsewhere?

PR #7506, opened 2021-10-31
🎉 Source Gitlab: add `OAuth2.0` authentication support

bazarnov:
## What
#7480 - Source GitLab: support oauth

## How
- added `GitlabAuthenticator` class inside of `source.py`
- extended `spec.json` with `oauthFlowSpecification`
- added `GitlabOauthFlow.java`
- added `GitlabOauthFlowTest.java` unit_test
- added `GitlabOauthFlowIntegrationTest.java` integration_test

## Pre-merge Checklist
Expand the relevant checklist and delete the others. 

<details><summary> <strong> Updating a connector </strong></summary>
<p>
   
#### Community member or Airbyter
   
- [X] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [X] Documentation updated 
    - [X] Changelog updated in `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
- [X] PR name follows [PR naming conventions](https://docs.airbyte.io/contributing-to-airbyte/updating-documentation#issues-and-pull-requests)
- [X] Connector version bumped like described [here](https://docs.airbyte.io/connector-development#publishing-a-connector)
   
#### Airbyter

If this is a community PR, the Airbyte engineer reviewing this PR is responsible for the below items. 
   
- [ ] Create a non-forked branch based on this PR and test the below items on it
- [ ] Build is successful
- [ ] Credentials added to Github CI. [Instructions](https://docs.airbyte.io/connector-development#using-credentials-in-ci). 
- [ ] [`/test connector=connectors/<name>` command](https://docs.airbyte.io/connector-development#updating-an-existing-connector) is passing. 
- [ ] New Connector version released on Dockerhub by running the `/publish` command described [here](https://docs.airbyte.io/connector-development#updating-an-existing-connector)

</p>
</details>

---
This change is [Reviewable](https://reviewable.io/reviews/airbytehq/airbyte/7506)

PR #20923, opened 2022-12-29
CDK: `SingleUseRefreshTokenOauth2Authenticator` update config with access tokens and expiration date

### LINKED ISSUE #20914: CDK: `SingleUseRefreshTokenOauth2Authenticator` refreshes tokens when necessary

Currently, the `SingleUseRefreshTokenOauth2Authenticator` is set up to always perform a refresh when it's initialized. This is because the expiration and initial access tokens are not set on initialization.

The GitLab connector's implementation in https://github.com/airbytehq/airbyte/pull/7506 tries to implement a smarter refresh by checking the token's validity and setting the expiration accordingly:

https://github.com/airbytehq/airbyte/blob/5000dfb2f1b0eee3c2bd7a9d0b6c2a438a183461/airbyte-integrations/connectors/source-gitlab/source_gitlab/source.py#L54-L59

However, this only works the first time around and when the authenticator itself performs a refresh, the new access token is not persisted in the config.

This is an issue because we can't fully support eagerly refreshing tokens at the moment due to platform limitations. Connectors that always perform a refresh and invalidate the old tokens during check are not able to be set up (see https://github.com/airbytehq/airbyte/issues/20910) so we should make it easy for connectors to implement "smarter" refresh flows that only refresh when necessary.

The authenticator should have a way of getting the expiration time and only refresh when necessary. It should also persist the access token in the config (in addition to the refresh token) so that it can be re-used.

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/20914

Before this PR the `SingleUseRefreshTokenOauth2Authenticator` eagerly performed access token refresh:
- In each operation (`check`, `discover`, `read`) a new OAuth flow is started even if the previous access token is still valid.

As synchronous operation, like `check` on source creation, can't process control message, we can end up with invalid refresh token as the latest one are not persisted to the DB. (more details in https://github.com/airbytehq/airbyte-internal-issues/issues/1260)
To mitigate this problem we try to store the expiration date of the access token in the config and only perform refresh when the expiration date is met. 

## How
* Read `access_token_expiration_datetime` from connector configuration
* On access token refresh:
  * Update the `access_token_value`
  * Update the `access_token_expiration_datetime`

## Recommended reading order

https://github.com/airbytehq/airbyte/blob/833d978fe1b2cff2779397bc747eeba1f8c6954f/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py#L93

https://github.com/airbytehq/airbyte/blob/833d978fe1b2cff2779397bc747eeba1f8c6954f/airbyte-cdk/python/unit_tests/sources/streams/http/requests_native_auth/test_requests_native_auth.py#L183

## 🚨 User Impact 🚨
The connector using `SingleUseRefreshTokenOauth2Authenticator` should be reworked:
https://github.com/airbytehq/airbyte/pull/7506

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
@@ -113,6 +113,7 @@ class SingleUseRefreshTokenOauth2Authenticator(Oauth2Authenticator):
         grant_type: str = "refresh_token",
         client_id_config_path: Sequence[str] = ("credentials", "client_id"),
         client_secret_config_path: Sequence[str] = ("credentials", "client_secret"),
+        access_token_config_path: Sequence[str] = ("credentials", "access_token"),
         refresh_token_config_path: Sequence[str] = ("credentials", "refresh_token"),
     ):
         """
@@ -129,10 +130,12 @@ class SingleUseRefreshTokenOauth2Authenticator(Oauth2Authenticator):
             grant_type (str, optional): OAuth grant type. Defaults to "refresh_token".
             client_id_config_path (Sequence[str]): Dpath to the client_id field in the connector configuration. Defaults to ("credentials", "client_id").
             client_secret_config_path (Sequence[str]): Dpath to the client_secret field in the connector configuration. Defaults to ("credentials", "client_secret").
+            access_token_config_path (Sequence[str]): Dpath to the access_token field in the connector configuration. Defaults to ("credentials", "access_token").
             refresh_token_config_path (Sequence[str]): Dpath to the refresh_token field in the connector configuration. Defaults to ("credentials", "refresh_token").
         """
         self._client_id_config_path = client_id_config_path
         self._client_secret_config_path = client_secret_config_path
+        self._access_token_config_path = access_token_config_path
         self._refresh_token_config_path = refresh_token_config_path
         self._refresh_token_name = refresh_token_name
         self._connector_config = observe_connector_config(connector_config)
@@ -181,13 +184,20 @@ class SingleUseRefreshTokenOauth2Authenticator(Oauth2Authenticator):
     def get_refresh_token(self) -> str:
         return dpath.util.get(self._connector_config, self._refresh_token_config_path)
 
-    def set_refresh_token(self, new_refresh_token: str):
-        """Set the new refresh token value. The mutation of the connector_config object will emit an Airbyte control message.
+
+    def _update_config_with_access_and_refresh_tokens(self, new_access_token: str, new_refresh_token: str):
+        """Update the connector configuration with new access and refresh token values. 
+        The mutation of the connector_config object will emit Airbyte control messages.
 
         Args:
+            new_access_token (str): The new access token value.
             new_refresh_token (str): The new refresh token value.
         """
+        # TODO alafanechere this will sequentially emit two control messages. 
+        # We should rework the observer/config mutation logic if we want to have atomic config updates in a single control message.
+        dpath.util.set(self._connector_config, self._access_token_config_path, new_access_token)
         dpath.util.set(self._connector_config, self._refresh_token_config_path, new_refresh_token)
+
 
     def get_access_token(self) -> str:
         """Retrieve new access and refresh token if the access token has expired.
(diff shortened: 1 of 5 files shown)


airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
```diff
@@ -104,36 +104,39 @@ def __init__(
...
             client_secret_config_path (Sequence[str]): Dpath to the client_secret field in the connector configuration. Defaults to ("credentials", "client_secret").
+            access_token_config_path (Sequence[str]): Dpath to the access_token field in the connector configuration. Defaults to ("credentials", "access_token").
             refresh_token_config_path (Sequence[str]): Dpath to the refresh_token field in the connector configuration. Defaults to ("credentials", "refresh_token").
+            access_token_expiration_datetime_config_path (Sequence[str]): Dpath to the access_token_expiration_datetime field in the connector configuration. Defaults to ("credentials", "access_token_expiration_datetime").
         """
         self._client_id_config_path = client_id_config_path
         self._client_secret_config_path = client_secret_config_path
+        self._access_token_config_path = access_token_config_path
         self._refresh_token_config_path = refresh_token_config_path
+        self._access_token_expiration_datetime_config_path = access_token_expiration_datetime_config_path
```
pedroslopez, line 139: We should also set the initial access_token from the config as part of the initialization - otherwise I believe we wouldn't have an access token and require it to refresh, right?

GitLab was doing this manually:

https://github.com/airbytehq/airbyte/blob/5000dfb2f1b0eee3c2bd7a9d0b6c2a438a183461/airbyte-integrations/connectors/source-gitlab/source_gitlab/source.py#L57-L59

pedroslopez (changes requested):
Nice- main thing I think should be addressed is making sure that the initial access token is loaded from the config.

Should we always expect the expiration date to be part of the config or should we allow connectors to decide to use an an endpoint to retrieve the expiration date of the token?

Also, did you check the APIs for whether they had these kinds of endpoints like you mentioned? If so, I'm curious what the results of that investigation were.

airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
```diff
@@ -181,13 +184,25 @@ def get_client_secret(self) -> str:
...
         Args:
+            new_access_token (str): The new access token value.
             new_refresh_token (str): The new refresh token value.
+            new_access_token_expiration_datetime (pendulum.DateTime): The new access token expiration date.
         """
+        # TODO alafanechere this will sequentially emit three control messages. 
+        # We should rework the observer/config mutation logic if we want to have atomic config updates in a single control message.
+        dpath.util.set(self._connector_config, self._access_token_config_path, new_access_token)
         dpath.util.set(self._connector_config, self._refresh_token_config_path, new_refresh_token)
+        dpath.util.set(self._connector_config, self._access_token_expiration_datetime_config_path, new_access_token_expiration_datetime)
```
pedroslopez, line 204: note: this means as currently implemented during the sync the platform will perform 3 separate config updates (api calls) for each of these pieces

alafanechere: I discarded the use of `ObservedDict` and directly call a function from this method to emit the control message.

airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
alafanechere: I made the change and added a test to ensure `access_token` attribute is set on init.

alafanechere:
> Should we always expect the expiration date to be part of the config or should we allow connectors to decide to use an an endpoint to retrieve the expiration date of the token?

Yes, we should expect the expiration date to be part of the config if the `SingleUseRefreshTokenOauth2Authenticator` is directly used. 
I think that sources having a dedicated endpoint to fetch expiration can override the `token_has_expired` method.

> Also, did you check the APIs for whether they had these kinds of endpoints like you mentioned? If so, I'm curious what the results of that investigation were.

I did but did not find these endpoints for Quickbook, TypeForm, Jira and Pipedrive

evantahler:
This is good work!  Do you happen to know how common it is for oAuth endpoints to return an `expires_in` / `expires_at`?  If this is common, than storing this timestamp along with the token in the config is the way to go.  Otherwise, perhaps we need to continue eagerly refreshing, but not in CHECK and DISCOVER modes (only READ and WRITE)

alafanechere:
> Do you happen to know how common it is for oAuth endpoints to return an expires_in / expires_at?

I'm pretty sure returning an `expires_in` field after an access token refresh is part of the OAuth standard.
I double-checked that it's the case for the new connectors we want to support and they do all return this field in their responses: 

✅ Quickbook
✅ Jira
✅ Typeform
✅ Pipedrive

pedroslopez (approved):

PR #20928, opened 2022-12-29
tolerate missing source catalog used to make the configured catalog

mfsiega-airbyte:
## What
Tolerate a missing source catalog used to make the configured catalog.

## How
Use the configured catalog itself. This is a bandaid -- instead we should have clear guarantees about whether that source catalog should exist -- but it's consistent with the existing approach.

### DIFF
--- a/airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
+++ b/airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
@@ -348,10 +348,7 @@ public class WebBackendConnectionsHandler {
       /*
        * constructs a full picture of all existing configured + all new / updated streams in the newest
        * catalog.
-       */
-      syncCatalog = updateSchemaWithRefreshedDiscoveredCatalog(configuredCatalog, catalogUsedToMakeConfiguredCatalog.get(),
-          refreshedCatalog.get().getCatalog());
-      /*
+       *
        * Diffing the catalog used to make the configured catalog gives us the clearest diff between the
        * schema when the configured catalog was made and now. In the case where we do not have the
        * original catalog used to make the configured catalog, we make due, but using the configured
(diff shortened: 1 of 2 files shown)


airbyte-server/src/main/java/io/airbyte/server/handlers/WebBackendConnectionsHandler.java
```diff
@@ -348,17 +348,17 @@ public WebBackendConnectionRead webBackendGetConnection(final WebBackendConnecti
...
-      /*
+       *
        * Diffing the catalog used to make the configured catalog gives us the clearest diff between the
        * schema when the configured catalog was made and now. In the case where we do not have the
        * original catalog used to make the configured catalog, we make due, but using the configured
        * catalog itself. The drawback is that any stream that was not selected in the configured catalog
        * but was present at time of configuration will appear in the diff as an added stream which is
        * confusing. We need to figure out why source_catalog_id is not always populated in the db.
        */
+      syncCatalog = updateSchemaWithRefreshedDiscoveredCatalog(configuredCatalog, catalogUsedToMakeConfiguredCatalog.orElse(configuredCatalog),
```
alovew, line 359: in the update endpoint, we just check to make sure that catalogUsedToMakeConfiguredCatalog is present before calling this. Maybe we should do the same thing here?

```
    final Optional<AirbyteCatalog> catalogUsedToMakeConfiguredCatalog = connectionsHandler
        .getConnectionAirbyteCatalog(connectionId);
    if (catalogUsedToMakeConfiguredCatalog.isPresent()) {
      // Update the Catalog returned to include all streams, including disabled ones
      final AirbyteCatalog syncCatalog =
          updateSchemaWithRefreshedDiscoveredCatalog(updatedConnectionRead.getSyncCatalog(), catalogUsedToMakeConfiguredCatalog.get(),
              catalogUsedToMakeConfiguredCatalog.get());
      updatedConnectionRead.setSyncCatalog(syncCatalog);
    }
```

mfsiega-airbyte: The behaviour here is that if the `catalogUsedToMakeConfiguredCatalog` is missing, then we'll show any previously de-selected fields as "newly added" rather than "already existed but de-selected". This is indeed not ideal, but not returning a diff is also a problem since the frontend needs the diff to decide what to display to the user.

The previous decision was that if we don't have the catalogUsedToMakeConfiguredCatalog then we can wrongly display streams as newly-added, so this approach is consistent with that.

I don't actually understand the behaviour in the update endpoint - it may be that the frontend isn't actually using the value returned by the update endpoint? Otherwise, I can't figure out how it's okay to just sometimes not populate the sync catalog field?

alovew: ah I see what you're saying here. I'd have to look at the update endpoint more closely to figure out the logic there, I'm not 100% sure why it would be ok to not populate sync catalog. but I think for this endpoint this is fine

alovew (approved):

octavia-approvington:

octavia-approvington (approved):

PR #20932, opened 2022-12-30
fix(oauth): dont convert config values to string when building consent url

pedroslopez:
## What

Fixes OC issue https://github.com/airbytehq/oncall/issues/1250

If the configuration passed to generate the consent URL includes non-string values, they would be implicitly converted to strings when replacing masked values with secrets.

We noticed this happening specifically for Salesforce when re-authenticating the source, since it uses the `is_sandbox` field, which is a boolean. Issuing the `get_consent_url` call for an already created connector would return a 422 status code with the following error:

```
io.airbyte.validation.json.JsonValidationException: json schema validation failed when comparing the data to the json schema. 
Errors: $.is_sandbox: string found, boolean expected 
Schema: 
{
  "type" : "object",
  "properties" : {
    "is_sandbox" : {
      "type" : "boolean",
      "path_in_connector_config" : [ "is_sandbox" ]
    }
  },
  "additionalProperties" : false
}
```

## How

The `getOAuthFromDBIfNeeded` method was using a string map and calling `Jsons.deserializeToStringMap`, which basically converts all values to strings. In order to keep the original types, I reworked the implementation to use JsonNodes directly.

### DIFF
--- a/airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
+++ b/airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
@@ -327,6 +327,7 @@ public class OAuthHandler {
       final String k = entry.getKey();
       final JsonNode v = entry.getValue();
 
+      // Note: This does not currently handle replacing masked secrets within nested objects.
       if (AirbyteSecretConstants.SECRETS_MASK.equals(v.textValue())) {
         if (oAuthInputConfigurationFromDB.has(k)) {
           result.set(k, oAuthInputConfigurationFromDB.get(k));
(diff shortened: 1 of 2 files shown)


airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
```diff
@@ -321,23 +321,25 @@ Map<String, String> buildJsonPathFromOAuthFlowInitParameters(final Map<String, L
...
-        });
-
-    return Jsons.jsonNode(result);
+    final ObjectNode result = (ObjectNode) Jsons.emptyObject();
+
+    oAuthInputConfigurationFromInput.fields().forEachRemaining(entry -> {
+      final String k = entry.getKey();
+      final JsonNode v = entry.getValue();
+
+      // Note: This does not currently handle replacing masked secrets within nested objects.
```
pedroslopez, line 330: I noticed this, but it's not an existing issue. I verified with the previous behavior and if there was a nested object in the input configuration, `deserializeToStringMap` actually threw an error.

alafanechere (approved):

evantahler (approved):

PR #20933, opened 2022-12-30
fix(oauth): handle fields missing in the stored config when building consent url

pedroslopez:
## What

Fixes OC issue https://github.com/airbytehq/oncall/issues/1251

When an input used for the oauth flow is missing from the stored configuration, a `java.util.NoSuchElementException: No value present` exception was being thrown. 

The value not being stored in the database is expected in some scenarios, like is the case for Monday. It has a field located within the oauth credentials block (credentials.subdomain) that would only be set if the connector was previously authenticated with OAuth. Having setting up the connector with an api key, this field is not part of the saved configuration. When the OAuthHandler tries to look up the value for that field, it doesn't find anything and throws the above exception. 

## How

Update the implementation to only set fields in the returned config that it are present. Fields that are not present in the stored configuration will be left out from the map.

### DIFF
--- a/airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
+++ b/airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
@@ -43,6 +43,7 @@ import java.net.http.HttpClient;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
```diff
@@ -342,10 +343,17 @@ JsonNode getOauthFromDBIfNeeded(final JsonNode oAuthInputConfigurationFromDB, fi
 
   @VisibleForTesting
   JsonNode getOAuthInputConfiguration(final JsonNode hydratedSourceConnectionConfiguration, final Map<String, String> pathsToGet) {
-    return Jsons.jsonNode(pathsToGet.entrySet().stream()
-        .collect(Collectors.toMap(
-            Map.Entry::getKey,
-            entry -> JsonPaths.getSingleValue(hydratedSourceConnectionConfiguration, entry.getValue()).get())));
+    final Map<String, JsonNode> result = new HashMap<>();
+    pathsToGet.forEach((k, v) -> {
+      final Optional<JsonNode> configValue = JsonPaths.getSingleValue(hydratedSourceConnectionConfiguration, v);
```
alafanechere, line 348: Could we instead make sure the `hydratedSourceConnectionConfiguration` has all the expected paths?

alafanechere (approved):

airbyte-server/src/main/java/io/airbyte/server/handlers/OAuthHandler.java
pedroslopez: The expected behavior is actually that `hydratedSourceConnectionConfiguration` _doesn't_ have the path we're trying to get in this case.

Because of the way the spec is defined, `credentials.subdomain` is only present in the config if the OAuth authentication type has been selected (since it's only used in the oauth case)

So, if you're coming from an API key auth, your config will look something like `{"credentials": {"api_token": "sometoken", "auth_type": "api_token"}}`. When this method runs after switching for oauth, we try to get the value of `credentials.subdomain` from the existing configuration but realize it's not there. Previously since we always expected it to be present an error would be thrown. Now, we just log a warning and don't retrieve the value.

evantahler (approved):

PR #20934, opened 2022-12-30
🎉 Source Airtable: implement OAuth 

### LINKED ISSUE #20708: 🎉 Source Airtable: Implement OAuth2.0 authentication
Based on the official docs: https://airtable.com/developers/web/api/oauth-reference
the Oauth flow is also supported and could be implemented to access Airtable resources.

bazarnov:
## What
Resolving: https://github.com/airbytehq/airbyte/issues/20708

## How
- updated the connector to work with `SingleUseRefreshTokens`
- added Java part for server
- tested on Local Dev UI
- added  the `prod` OAuth2.0 credentials to the `prod-db`

## 🚨 User Impact 🚨
No impact is expected if OLD credentials are used already.

## Pre-merge Checklist

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>


bazarnov:
Blocked by: https://github.com/airbytehq/airbyte/issues/20145

@pedroslopez Please consider this one as dependency, we need to cover `check` and `discover` stages with `ConfigUpdate` to have the oAuth for Airtable working.

airbyte-integrations/connectors/source-airtable/source_airtable/auth.py
```diff
@@ -0,0 +1,56 @@
...
+            url=self.get_token_refresh_endpoint(),
+            data=self.build_refresh_request_body(),
+            headers=self.build_refresh_request_headers(),
+        )
+        response.raise_for_status()
+        return response.json()
+
+
+class AirtableAuth:
+    def __new__(self, config: dict) -> Union[TokenAuthenticator, AirtableOAuth]:
```
antixar, line 51: Hi, 
Please pay your attention to the next points:
1. Any `__new__` magic function accepts an instance as the first argument. And it will be better if it is "cls".
2. `config["api_key"]`  can be empty or `config` can be `None`. I would have check by this if had written this condition: `(config or {}).get("api_key")`.

airbyte-integrations/connectors/source-airtable/source_airtable/source.py
```diff
@@ -43,7 +42,7 @@ def discover(self, logger: AirbyteLogger, config) -> AirbyteCatalog:
 
         Retrieve: Bases, Tables from each Base, generate JSON Schema for each table.
         """
-        auth = TokenAuthenticator(token=config["api_key"])
+        auth = self._auth if self._auth else AirtableAuth(config)
```
antixar, line 45: You can use more simple and readable variant:
`auth = self._auth or AirtableAuth(config)`

airbyte-integrations/connectors/source-airtable/source_airtable/auth.py
```diff
@@ -0,0 +1,56 @@
...
+    https://airtable.com/developers/web/api/oauth-reference#token-expiry-refresh-tokens
+    """
+
+    def build_refresh_request_headers(self) -> Mapping[str, Any]:
+        """
+        https://airtable.com/developers/web/api/oauth-reference#token-refresh-request-headers
+        """
+        config_creds = self._connector_config["credentials"]
+        return {
+            "Authorization": BasicHttpAuthenticator(config_creds["client_id"], config_creds["client_secret"]).token,
```
antixar, line 26: Try to use ready functions: [get_client_id](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py#L184) and [get_client_secret](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py#L187).

airbyte-integrations/connectors/source-airtable/source_airtable/source.py
```diff
@@ -10,21 +10,20 @@
...
 from .streams import AirtableBases, AirtableStream, AirtableTables
 
 
 class SourceAirtable(AbstractSource):
 
     logger: logging.Logger = logging.getLogger("airbyte")
-
-    # prepared streams catalog
     streams_catalog: Iterable[Mapping[str, Any]] = []
+    _auth: AirtableAuth = None
```
antixar, line 23: I see to the class variable `_auth` is used by 2 diff functions: `streams` and `check_connection`. I am not sure to these ones can be called by same python script call and this code optimisation will have some benefit because AirByte engine calls them by diff commands. Please correct me if I am not right.

airbyte-integrations/connectors/source-airtable/source_airtable/spec.json
```diff
@@ -4,15 +4,69 @@
...
+          {
+            "title": "API Key",
+            "type": "object",
+            "required": ["api_key"],
+            "properties": {
+              "auth_method": {
+                "type": "string",
+                "const": "api_key"
+              },
+              "api_key": {
```
antixar, line 51: if we know its permanent length or some RegExp, we can set this additional validation here.

bazarnov: No need to RegExp of alphanumeric chars, especially if this could be changed on the API side, this would require the spec tunings later on, which is not necessary I think.

airbyte-integrations/connectors/source-airtable/source_airtable/source.py
bazarnov: You're correct, therefore in order to have a local test and development more manageable, we need to save the auth for later use, if possible. The Platform calls the check, discover, and streams using different commands; therefore, while using `SingleUseRefreshToken` we need to be sure that the same `access_token` acquired from OAuth2.0 will be passed to all the stream class instances and will not be triggered for each stream generated.

airbyte-integrations/connectors/source-airtable/source_airtable/auth.py
bazarnov: Perfectly fitted)

lazebnyi (approved):

airbyte-integrations/connectors/source-airtable/acceptance-test-config.yml
```diff
@@ -11,18 +12,27 @@ acceptance_tests:
...
+        status: "succeed"
+      - config_path: "integration_tests/invalid_config_oauth.json"
+        status: "failed" 
   discovery:
     tests:
       - config_path: "secrets/config.json"
       # bypassed this check, because discovery mechanism was changed
         backward_compatibility_tests_config:
-          disable_for_version: "0.1.3"
+          disable_for_version: "1.0.1"
```
alafanechere, line 24: Is this the same bypass reason as for 0.1.3?

airbyte-integrations/connectors/source-airtable/integration_tests/invalid_config_oauth.json
```diff
@@ -0,0 +1,10 @@
+{
+    "credentials": {
+        "auth_method": "oauth2.0",
+        "client_id": "client_id",
+        "client_secret": "client_secret",
+        "access_token": "access_token",
+        "refresh_token": "refresh_token",
+        "token_expiry_date": "2023-01-01T12:12:12.000000+00:00"
+    }
+}
```
alafanechere, line 10: missing end line

airbyte-integrations/connectors/source-airtable/source_airtable/source.py
```diff
@@ -43,7 +42,7 @@ def discover(self, logger: AirbyteLogger, config) -> AirbyteCatalog:
 
         Retrieve: Bases, Tables from each Base, generate JSON Schema for each table.
         """
-        auth = TokenAuthenticator(token=config["api_key"])
+        auth = self._auth or AirtableAuth(config)
```
alafanechere, line 45: Could you please explain this logic?

airbyte-integrations/connectors/source-airtable/source_airtable/spec.json
```diff
@@ -4,14 +4,119 @@
     "$schema": "http://json-schema.org/draft-07/schema#",
     "title": "Airtable Source Spec",
     "type": "object",
-    "required": ["api_key"],
+    "additionalProperties": true,
```
alafanechere, line 7:
```suggestion
```
True is the default.

alafanechere (approved):
Code LGTM but I think we should pin the airtable version in `airbyte-cloud` to make sure the java code is deployed before we update the connector version.

Please also sync with @airbytehq/cloud-support before merging / deploying the new connector version on cloud as there is a breaking change related to configuration. Could you explain what actions the users need to take to make their sync working again?

cc. @pedroslopez are you comfortable having a second "single use refresh token" connector on cloud or shall we hold until your current platform improvement are live?

airbyte-integrations/connectors/source-airtable/integration_tests/expected_records.txt
```diff
@@ -0,0 +1,42 @@
+{"stream": "users/table_1", "data": {"_airtable_id": "recgRMRMHxgcgJeDe", "_airtable_created_time": "2021-11-16T13:30:17.000Z", "name": "test2", "status": "test", "notes": "test_note2"}, "emitted_at": 1674657815126}
```
alafanechere, line 1: please use the `.jsonl` extension instead of `.txt`

bazarnov: Tricky part) It used to be `txt`  as far as I remember [here](https://docs.airbyte.com/connector-development/testing-connectors/source-acceptance-tests-reference/#example-of-expected_recordsjsonl). 

Sure, thanks.

airbyte-integrations/connectors/source-airtable/acceptance-test-config.yml
bazarnov: This part could be removed for this particular version release, there are no changes with the previous version, so it should be backward capable with the previous version.

erica-airbyte:
@alafanechere or @bazarnov looks like this has been deployed?

From @alafanechere note it sounds like this is a breaking change? 

If so please provide the below information
- Workspaces affected by the change
- Time when the change will be applied
- Reason for the breaking change
- What is being updated/changed
- What will the customer see that is different? 
- What actions the customers need to take once the changes goes live? (Refresh source schema, Select to reset affected streams, Reset, Update permissions (if this why are permissions changing))

airbyte-integrations/connectors/source-airtable/source_airtable/source.py
bazarnov: We need to trigger the catalog generation to produce streams from the auto-generated catalog and not generate it 2-ce, take a look at the `streams` method, it reveals this mystery)

airbyte-integrations/connectors/source-airtable/source_airtable/spec.json
bazarnov: It should be `True` to provide the backward capability.

bazarnov:
> Please also sync with @airbytehq/cloud-support before merging / deploying the new connector version on cloud as there is a breaking change related to configuration. Could you explain what actions the users need to take to make their sync working again?

@alafanechere 
No breaking changes are expected with this PR, it doesn't touch the Customers that are already has their connections. The old connections will still work - tested by setting up the old version > updated to this one > run the sync again >  looks good.


> @alafanechere or @bazarnov looks like this has been deployed?
> 
> From @alafanechere note it sounds like this is a breaking change?
> 
> If so please provide the below information
> 
> * Workspaces affected by the change
> * Time when the change will be applied
> * Reason for the breaking change
> * What is being updated/changed
> * What will the customer see that is different?
> * What actions the customers need to take once the changes goes live? (Refresh source schema, Select to reset affected streams, Reset, Update permissions (if this why are permissions changing))

@erica-airbyte No at all, these changes are not  deployed yet) we mention your team here to notify that this is going to be deployed soon)

pedroslopez:
> cc. @pedroslopez are you comfortable having a second "single use refresh token" connector on cloud or shall we hold until your current platform improvement are live?

All good from my end!

bazarnov:
> Agree we should do the following ahead of time though:

Sure.

bazarnov:
@erica-airbyte 

> * Workspaces affected by the change

Hard to say close enough but maybe 2-5

> * Time when the change will be applied

After [This Cloud PR](https://github.com/airbytehq/airbyte-cloud/pull/4212) is reverted back and version is >=1.0.1

> * Reason for the breaking change (is it breaking?)

Users must have OAuth for `source-airtable`, it simplifies the setting right scopes and selecting the other user-related sections for authorization purposes, which are important for the schema auto-generation mechanisms to provide the dynamic schema discovery based on the authenticated user in this PR https://github.com/airbytehq/airbyte/pull/21851 
However it's not directly implied the possibility to generate schemas automatically, OAuth gives it a better shine.

> * What will the customer see that is different?

The customer will see 2 options instead of 1 on the set-up source page, for `source-airtable`. 

> * What actions the customers need to take once the changes goes live?
No actions are required.

bazarnov:
Version `1.0.1` is pinned for Cloud, after the server components are updated and released to Cloud, the PR: https://github.com/airbytehq/airbyte-cloud/pull/4212 should be reverted and the version should be unpinned to catch the latest image.

Issue #20417, opened 2022-12-13
Investigation : Improve LIMIT size calculation to increase DB source syncs


akashkulk:
## Tell us about the problem you're trying to solve
Currently, we optimize our batch sizes/limits based on current row size. As this [PR](https://github.com/airbytehq/airbyte/pull/19514) showed, there are some major gains in performance from increasing batch size while executing queries. 

## Describe the solution you’d like
Investigate if it is possible to make additional gains in performance by further modifying this strategy while not OOMing

bleonard:
The previous PR changed the parameters in the current model fairly conservatively.

We are totally opn to changing the model drsatically or incrementally is that's what makes sense.

When testing out other data loading tools (BEAM), we were seeing it query 100x what we are doing on a skinny table on a test that @grishick did.

bleonard:
1. Let's update the logs to say what we are setting as the fetch size.

2. Let's fix the bug we found about the ratio.

PR #20939, opened 2022-12-30
Increase memory allocation for JDBC buffers in source DB connectors

akashkulk:
Work is part of https://github.com/airbytehq/airbyte/issues/20417. 

- Follow-up to https://github.com/airbytehq/airbyte/pull/19514 where allocation of source's total memory for JDBC buffers was bumped from 50% -> 60%
- Increases `MAX_FETCH_SIZE` frpm 1,000,000 -> 2,000,000
- Removes concept of `MAX_BUFFER_BYTE_SIZE`, which was hard coded to `1GB`. The memory allocation now has an upper bound of  `TARGET_BUFFER_SIZE_RATIO * MAX_HEAP_SPACE`. This is so we can take advantage of the increased memory for source containers. 

Tested locally, and hasn't caused any additional OOMs


akashkulk:
@evantahler / @davinchia I was trying to follow the comments from the previous PR. What's a good way to test this increase?

evantahler:
> @evantahler / @davinchia I was trying to follow the comments from the previous PR. What's a good way to test this increase?

... I didn't come up with a good way to test this kind of change 😢 
I think for now, the best we can do is to publish the change (OSS) and run a few syncs with it on a local or staging K8s cluster and compare new vs old memory consumption for a sync. In the future, we'll be looking into blue/green connector deployments to try this connector out on a few cloud workspaces before rolling it out to everyone.

davinchia:
@akashkulk you should be able to build the image locally, load it into a local airbyte deployment (either docker or K8s) and inspect memory usage as the sync runs. This all can be done before we merge this change into OSS.

Alternatively, you can also build the connector locally and run the connector's read command and inspect local memory usage. I'm going to find some time over the next few days to draft our a local test harness for this. Till then, I would recommend either of these two options.

evantahler (approved):
Approving.... but you probably want to test it first

akashkulk:
> @akashkulk you should be able to build the image locally, load it into a local airbyte deployment (either docker or K8s) and inspect memory usage as the sync runs. This all can be done before we merge this change into OSS.
> 
> Alternatively, you can also build the connector locally and run the connector's read command and inspect local memory usage. I'm going to find some time over the next few days to draft our a local test harness for this. Till then, I would recommend either of these two options.

@davinchia What do you recommend to inspect docker memory usage in a local airbyte deployment?

akashkulk:
Actually, I've realized that increasing this ration doesn't actually increase the memory allocation for JDBC buffers in source DB connectors. The algorithm is : 

`Math.min(1GB, Math.max(0.7*JVM_max_memory, 250MB))`

Where 0.7 is the defined target buffer ration. For our source container pods with ~2GB of memory, the JDBC buffer will always be at 1GB, so we'd have to remove that.

evantahler:
Remove it!

akashkulk:

akashkulk:

akashkulk:
/publish connector=connectors/source-mysql-strict-encrypt run-tests=false

PR #20941, opened 2023-01-02
CDK: Add schema inferrer class

### LINKED ISSUE #20819: [Low-Code CDK] Implement schema inferrer
As specified in the tech spec, there should be a helper class in the CDK which can automatically create schemas based on sample records.

## Acceptance criteria

* There is a util class exposed as part of the CDK which can be passed records for multiple streams and will return the derived schemas

flash1293:
Fixes #20819 

## What

This is the first part of adding automatic schema inferring to the connector builder UI. It adds a utility class `SchemaInferrer` to the CDK which can be used by the connector builder server and later on directly through the `read` command of a declarative source to infer the schema from some records of test data.

## How

Use the same lib powering https://github.com/airbytehq/airbyte/tree/master/tools/schema_generator to implement this generic class (do not use it anywhere yet). In a subsequent PR this class is used in the connector builder server to pipe all encountered records through and send the derived schema along with the response to the connector builder UI.


airbyte-cdk/python/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.16.3
```
brianjlai, line 3: nit: but given this is additive non-breaking functionality (and not a bug fix), we should bump the minor version so `0.17.0`

airbyte-cdk/python/airbyte_cdk/utils/__init__.py
```diff
@@ -1,6 +1,7 @@
 #
 # Copyright (c) 2022 Airbyte, Inc., all rights reserved.
 #
+from .schema_inferrer import SchemaInferrer
```
brianjlai, line 4: need an extra space between the licence and import or our CI formatting checks will fail

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        Returns the JSON schemas for all encountered streams inferred by inspecting all records
+        passed via the accumulate method
+        """
+        schemas = {}
+        for stream_name, builder in self.builders.items():
+            schemas[stream_name] = builder.to_schema()
+        return schemas
+
+    def get_stream_schema(self, stream_name: str) -> Optional[InferredSchema]:
```
brianjlai, line 71: The original tech spec outlined the `accumulate` and `get_inferred_schemas` functions, but for context what's the use case for a single stream or how are we using this `get_stream_schema()` function
Also why do we no longer need the `reset()`?

brianjlai (approved):
no blocking issues but just a few nits about formatting and version. Also wanted to get some context on interface changes from the original tech spec [here](https://docs.google.com/document/d/1fhAIHYBoqUk5vtTqEqcNrQMoXrUwrZ9HtAB5asgy74k/edit#).

I think the CI checks should pass once you fix formatting, but double check there aren't any other issues before the publish + merge.

airbyte-cdk/python/unit_tests/utils/test_schema_inferrer.py
```diff
@@ -0,0 +1,40 @@
...
+
+import json
+
+from airbyte_cdk.models.airbyte_protocol import AirbyteRecordMessage
+from airbyte_cdk.utils.schema_inferrer import SchemaInferrer
+
+NOW = 1234567
+
+
+def test_deriving_schema():
```
sherifnada, line 13: could we also test for arrays and objects?

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
sherifnada: fwiw when writing the spec i didn't account for this object needing to infer schema on multiple streams (that iface assumed a single stream). I think this iface makes more sense.

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Instances of this class are stateful, meaning they build their inferred schemas
+    from every record passed into the accumulate method.
+
+    """
+
+    builders: Dict[str, SchemaBuilder]
+
+    def __init__(self):
+        self.builders = {}
```
sherifnada, line 48: you can simplify the `if/else` block of `accumulate` by using `defaultdict` e.g: 

```
def __init__():
  self.builders = defaultdict(NoRequiredSchemaBuilder)

...

def accumulate(self, record): 
  self.builders[record.stream].add_object(record.data)

```

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
```diff
@@ -0,0 +1,78 @@
...
+class NoRequiredObj(Object):
+    """
+    This class has Object behaviour, but it does not generate "required[]" fields
+    every time it parses object. So we dont add unnecessary extra field.
+    """
+
+    def to_schema(self):
+        schema = super(Object, self).to_schema()
+        schema["type"] = "object"
+        if self._properties:
```
sherifnada, line 21: are these two if blocks what prevent generating `required` fields?

sherifnada (changes requested):

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
flash1293: Yeah, when integrating with the API I noticed that for that use case we only need a specific one and this API seemed nicer. The `get_inferred_schemas` will still be nice for the planned integration with the `read` command / implementing a separate command for schema inferrence

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
flash1293: Neat, didn't know about defaultdict

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
flash1293: In a sense, I copied this over from https://github.com/airbytehq/airbyte/blob/master/tools/schema_generator/schema_generator/infer_schemas.py#L41

However on checking the genson code I changed the logic to just call the existing `to_schema` logic of the object strategy and then dropping the `required` prop. It's doing a very small amount of additional work but it seems much cleaner, LMK what you think.

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
flash1293: I dropped the `reset` because in the places where we plan to use this class it's much easier to just lose the reference to the inferrer instance and have the runtime clean it up, then create a new instance for the next usage. Happy to change that, but it doesn't seem heavy enough to explicitly keep an instance or a pool of them around.

airbyte-cdk/python/unit_tests/utils/test_schema_inferrer.py
```diff
@@ -0,0 +1,99 @@
...
+        AirbyteRecordMessage(stream="my_stream", data={"field_A": 1.0}, emitted_at=NOW)
+    )
+    inferrer.accumulate(
+        AirbyteRecordMessage(
+            stream="my_stream", data={"field_A": "abc"}, emitted_at=NOW
+        )
+    )
+    assert (
+        json.dumps(inferrer.get_inferred_schemas()["my_stream"])
+        == '{"$schema": "http://json-schema.org/schema#", "type": "object", "properties": {"field_A": {"type": ["number", "string"]}}}'
```
sherifnada, line 47: wouldn't it be better to compare without casting to string to avoid mismatches around things like ordering of properties, which doesn't matter in objects? e.g would it be better to do 
```
assert (
inferrer.get_inferred_schemas()["my_stream"] 
== {"$schema": "http://json-schema.org/schema#", "type": "object", "properties": {"field_A": {"type": ["number", "string"]}}}
```

airbyte-cdk/python/unit_tests/utils/test_schema_inferrer.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+
+
+def test_deriving_schema_nested_structures():
```
sherifnada, line 51: would recommend considering the use of `pytest.mark.fixture` here to parametrize tests

sherifnada (approved):
LGTM once the comment is addressed, no need for another review

PR #20942, opened 2023-01-02
Connector builder server: Add inferred schema to read API response

### LINKED ISSUE #20831: [Connector Builder Server] Expose inferred schema in API

flash1293:
Fixes #20831 

## What

Adds the inferred schema for all encountered records to the response of the read API. Will be consumed by the UI in a subsequent PR to allow to add the schema for a stream with a single click

## How

Uses the utility class introduced in https://github.com/airbytehq/airbyte/pull/20941 to run all records through the inferrer and attach the final response to the response.


flash1293:
Unit tests are not working right currently due to unrelated changes, this is fixed on https://github.com/airbytehq/airbyte/pull/21107

airbyte-connector-builder-server/connector_builder/impl/default_api.py
```diff
@@ -108,12 +111,14 @@ async def read_stream(self, stream_read_request_body: StreamReadRequestBody = Bo
         :return: Airbyte record messages produced by the sync grouped by slice and page
         """
         adapter = self._create_low_code_adapter(manifest=stream_read_request_body.manifest)
+        schema_inferrer = SchemaInferrer()
```
girarda, line 114: this class is stateful right? we'll need to reset it or create a new SchemaInferrer for every read to avoid leaking the type of records produced from a different configuration

girarda (changes requested):

airbyte-connector-builder-server/connector_builder/impl/default_api.py
flash1293: isn't this a local variable in the scope of the current `read_stream` invocation? If that's the case there should be a new schema inferrer for every call of `read_stream` which is exactly what we want, right?

girarda (approved):
:shipit:

PR #20944, opened 2023-01-02
🪟🧹 Introduce flex component

flash1293:
## What

This PR introduces new ui components `FlexContainer` and `FlexItem` which can be used for basic flex layouting instead of introducing new classes for this purpose over and over again.

## How

The new components introduce props to control some of the flex properties. I kept it as simple as possible for now, if there is a strong reason we can extend these later on.

To demo the usage I replaced a bunch of components with these components (there was a `flexRow` class duplicated a bunch of times across the code base, this is exactly the kind of things this component is solving)

## Recommended reading order
1. `airbyte-webapp/src/components/ui/*`
2. The rest


airbyte-webapp/src/components/ui/Flex/FlexItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+interface FlexItemProps {
+  className?: string;
+  /**
+   * Sets `flex-grow` to 1 if truthy
+   */
+  grow?: boolean;
+  /**
+   * The `self-align` css property
+   */
+  selfAlign?: "flex-start" | "flex-end" | "center" | "baseline" | "stretch";
```
josephkmh, line 15: Could we use `alignSelf` to reflect the CSS `align-self` property?

airbyte-webapp/src/components/ui/Flex/FlexContainer.module.scss
```diff
@@ -0,0 +1,69 @@
+@use "../../../scss/variables";
```
josephkmh, line 1: nitpick: 
```suggestion
@use "scss/variables";
```

airbyte-webapp/src/components/ui/Flex/FlexContainer.tsx
```diff
@@ -0,0 +1,67 @@
...
+import React, { HTMLAttributes } from "react";
+
+import styles from "./FlexContainer.module.scss";
+
+interface FlexContainerProps {
+  className?: string;
+  /**
+   * The flex-direction css property
+   */
+  direction?: "row" | "column";
```
josephkmh, line 11: Can we add `row-reverse` and `column-reverse`?

airbyte-webapp/src/components/ui/Flex/FlexItem.module.scss
```diff
@@ -0,0 +1,25 @@
+@use "../../../scss/variables";
```
josephkmh, line 1: nitpick:

```suggestion
@use "scss/variables";
```

airbyte-webapp/src/components/ui/Flex/FlexContainer.module.scss
```diff
@@ -0,0 +1,69 @@
...
+  gap: variables.$spacing-md;
+}
+
+.gapLg {
+  gap: variables.$spacing-lg;
+}
+
+.gapXl {
+  gap: variables.$spacing-xl;
+}
```
josephkmh, line 25: Any reason to exclude `$spacing-xs` and `$spacing-2xl`?

joeykmh (changes requested):
Nice, I like the syntax! I think you captured the most essential flex properties IMO. Just left a few small comments.

One thought I had while reading through this PR was whether we could use [csstype](https://www.npmjs.com/package/csstype) to avoid needing to define CSS types ourselves. But after looking through the typings there, they are actually probably more permissive than we want. For example, [flex-grow](https://github.com/frenic/csstype/blob/master/index.d.ts#L18438) could be any number or string. Thought I'd mention it anyway!

flash1293:
Thanks for the review, added the things you commented on.

> But after looking through the typings there, they are actually probably more permissive than we want. For example, [flex-grow](https://github.com/frenic/csstype/blob/master/index.d.ts#L18438) could be any number or string

Yeah, I did that on purpose because I wanted to keep it focused on what we actually need. In this case 95% of the cases you're doing a `flex-grow: 1`, so this API allows you to simply do `<FlexItem grow>`.

Theres no clear boundary for what should be possible with the component vs. what requires you to include custom css - I mostly guessed based on what I think is "essentials".

airbyte-webapp/src/components/ui/Flex/FlexContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const FlexContainer: React.FC<React.PropsWithChildren<FlexContainerProps & HTMLAttributes<HTMLDivElement>>> = ({
+  className,
+  direction = "row",
+  gap = "md",
+  alignItems = "stretch",
+  justifyContent = "flex-start",
+  children,
+  ...otherProps
+}) => {
+  const fullClassName = classNames(
```
edmundito, line 39: Is there a benefit to memoizing the className generation?

edmundito (commented):
Just one question, but I like what I see. I will defer to @josephkmh for ✅

airbyte-webapp/src/components/ui/Flex/FlexItem.tsx
```diff
@@ -0,0 +1,47 @@
...
+interface FlexItemProps {
+  className?: string;
+  /**
+   * Sets `flex-grow` to 1 if truthy
+   */
+  grow?: boolean;
+  /**
+   * The `align-self` css property
+   */
+  alignSelf?: "flex-start" | "flex-end" | "center" | "baseline" | "stretch";
```
krishnaglick, line 15: See previous comment about `React.CSSProperties`

joeykmh (approved):
Thanks for making those changes! LGTM apart from the trivial merge conflicts

airbyte-webapp/src/components/ui/Flex/FlexContainer.tsx
flash1293: As it depends on so many props my feeling was that the dirty-check would take longer than re-evaluating the className.

airbyte-webapp/src/components/ui/Flex/FlexContainer.tsx
```diff
@@ -0,0 +1,72 @@
...
+import React, { HTMLAttributes } from "react";
+
+import styles from "./FlexContainer.module.scss";
+
+interface FlexContainerProps {
+  className?: string;
+  /**
+   * The flex-direction css property
+   */
+  direction?: "row" | "column" | "row-reverse" | "column-reverse";
```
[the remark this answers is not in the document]
flash1293: That would allow the prop to be set to ` "-moz-initial" | "inherit" | "initial" | "revert" | "unset"` as well, but these are not actually applied by the current logic. Besides that I liked the explicit nature of the interface. Let me know what you think.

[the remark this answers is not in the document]
krishnaglick: Sounds reasonable enough to me!
Doing a `Pick` on it to just get those props would be pointless I think :)

krishnaglick (approved):

PR #20957, opened 2023-01-03
🪟🧹 Connector builder: Fast fields performance improvements

### LINKED ISSUE #20947: Connector builder UI: Use formik fast fields and memoization to reduce full rerenders
While working on https://github.com/airbytehq/airbyte/pull/20809 I noticed how the UI performance inches close to becoming a problem because the whole tree re-renders very often. I mitigated this for the time being by debouncing the validation and upstream state setting which is the biggest issue, but there are still a lot of re-renders going on and it's bound to become a problem again.

A more sustainable approach fixing the problem at the root would be the following:
* Memoize the root components for the form views so they only re-render on actual changes
* Do not access the formik context directly anywhere - instead calculate the relevant bits (like number of streams) and pass them down through the memoized component
* Switch to [formik fast field](https://formik.org/docs/api/fastfield) in the builderfield component

This way the biggest issue (re-render of the large and complex form on every keystroke) is avoided.

flash1293:
Fixes https://github.com/airbytehq/airbyte/issues/20947

This PR is introducing fast field and memoization in some key places to speed up the perceived performance of the UI.

I left some comments in the code to explain a bit in greater detail what the individual improvements are about. However, these are the ones I added:
* Using FastField for oneOf and BuilderField
* Get formik to render half as often
* Memoize UserInputHelper (change a bit the helper functions for building the inputs list so they don't take the entire state)
* Memoize heaviest part of keyvalue list (we can improve this further by using fast field for each individual item, but there's a separate issue for that)
* Move stream config controls to a separate component so the whole component is re-rendered less

Measurements below are using a complex form with slicer, paginator and multiple request params, 4x CPU slowdown and heavy instrumentation - production build is much faster:

Before when typing quickly (with the other performance PR merged already):
There are two commits per keystroke, and both take 60ms - this causes heavy frame dropping
Screenshot 2023-01-06 at 12 26 33

Now:
One commit per keystroke, roughly 16ms - very little frame dropping 
Screenshot 2023-01-06 at 12 28 39

While I'm not super happy with this solution, I think it's good enough for the moment until we can switch away from formik which will re-render much less with idiomatic usage.


airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderStateService.tsx
```diff
@@ -61,21 +60,23 @@ export const ConnectorBuilderFormStateProvider: React.FC<React.PropsWithChildren
...
     },
     [setStoredBuilderFormValues]
   );
 
-  const builderFormValues = useMemo(() => {
-    return merge({}, DEFAULT_BUILDER_FORM_VALUES, storedBuilderFormValues);
-  }, [storedBuilderFormValues]);
+  // use the ref for the current builder form values because useLocalStorage will always serialize and deserialize the whole object,
+  // changing all the references which re-triggers all memoizations
+  const builderFormValues = currentBuilderFormValuesRef.current || DEFAULT_BUILDER_FORM_VALUES;
```
flash1293, line 79: Doing the merge on render is problematic because it will create a lot of new references, busting memoize barriers.

airbyte-webapp/src/components/connectorBuilder/types.ts
```diff
@@ -115,8 +115,8 @@ export const DEFAULT_BUILDER_STREAM_VALUES: Omit<BuilderStream, "id"> = {
   },
 };
 
-function getInferredInputList(values: BuilderFormValues): BuilderFormInput[] {
-  if (values.global.authenticator.type === "ApiKeyAuthenticator") {
+function getInferredInputList(global: BuilderFormValues["global"]): BuilderFormInput[] {
```
flash1293, line 118: Not taking the full `BuilderFormValues`, but just the required parts allows for memoization (otherwise every change will require to re-run this function)

airbyte-webapp/src/components/connectorBuilder/Builder/StreamConfigView.tsx
```diff
@@ -151,6 +109,70 @@ export const StreamConfigView: React.FC<StreamConfigViewProps> = ({ streamNum, h
...
+  setSelectedTab,
+  streamFieldPath,
+}: {
+  streamNum: number;
+  streamFieldPath: (path: string) => string;
+  setSelectedTab: (tab: "configuration" | "schema") => void;
+  selectedTab: "configuration" | "schema";
+}) => {
+  const { formatMessage } = useIntl();
+  const [field, , helpers] = useField<BuilderStream[]>("streams");
```
flash1293, line 124: Splitting out the controls to isolate the re-render to this component instead of re-rendering the full stream config view.

airbyte-webapp/src/components/connectorBuilder/Builder/KeyValueListField.tsx
```diff
@@ -60,29 +61,53 @@ interface KeyValueListFieldProps {
 export const KeyValueListField: React.FC<KeyValueListFieldProps> = ({ path, label, tooltip }) => {
   const [{ value: keyValueList }, , { setValue: setKeyValueList }] = useField<Array<[string, string]>>(path);
```
flash1293, line 62: Can't get rid of the "useField" here, but we can memoize everything happening _within_ the component if the key value list didn't actually change

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderFieldWithInputs.tsx
```diff
@@ -26,20 +27,18 @@ export const BuilderFieldWithInputs: React.FC<BuilderFieldProps> = (props) => {
...
+interface UserInputHelperProps {
   setValue: (value: string) => void;
   currentValue: string;
-}) => {
+}
+
+export const UserInputHelper = (props: UserInputHelperProps) => {
   const { formatMessage } = useIntl();
-  const [modalOpen, setModalOpen] = useState(false);
   const { builderFormValues } = useConnectorBuilderFormState();
```
flash1293, line 37: We can't get rid of subscribing to the builder form state, but we can memoize building the actual options list and memoize the actual listbox component based on that.

airbyte-webapp/src/pages/ConnectorBuilderPage/ConnectorBuilderPage.tsx
```diff
@@ -29,16 +29,25 @@ const ConnectorBuilderPageInner: React.FC = React.memo(() => {
...
-            editorView={editorView}
-            validateForm={validateForm}
-            switchToUI={switchToUI}
-            values={values}
-            switchToYaml={switchToYaml}
-          />
-        )}
+      <Formik
+        initialValues={initialFormValues.current}
+        validateOnBlur={false}
```
flash1293, line 34: Noticed that formik would still re-trigger renders if all validateOn* are explicitly disabled - this is a big saving.

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderOneOf.tsx
```diff
@@ -60,3 +64,10 @@ export const BuilderOneOf: React.FC<BuilderOneOfProps> = ({ options, path, label
     </GroupControls>
   );
 };
+export const BuilderOneOf: React.FC<BuilderOneOfProps> = (props) => {
+  return (
+    <FastField name={`${props.path}.type`}>
+      {(fastFieldProps: FastFieldProps<string>) => <InnerBuilderOneOf {...props} {...fastFieldProps} />}
```
flash1293, line 70: Wrap the oneOf block in its own fastfield

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderField.tsx
```diff
@@ -133,3 +140,13 @@ export const BuilderField: React.FC<BuilderFieldProps> = ({
     </ControlLabels>
   );
 };
+
+export const BuilderField: React.FC<BuilderFieldProps> = (props) => {
+  return (
+    <FastField name={props.path}>
+      {({ field, form, meta }: FastFieldProps<unknown>) => (
```
flash1293, line 147: Wrap each builder field in a fast field instead of using `useField`

airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderStateService.tsx
```diff
@@ -119,9 +120,9 @@ export const ConnectorBuilderFormStateProvider: React.FC<React.PropsWithChildren
       editorView !== "ui"
         ? jsonManifest
         : builderFormValues === lastValidBuilderFormValues
-        ? jsonManifest
+        ? derivedJsonManifest
```
lmossman, line 123: FYI I pushed this change because I noticed when testing this branch that the stream URL in the testing panel was not being updated when I would change the URL path of the stream.

After some digging, I figured out that this should be using the `derivedJsonManifest` in this case, because `jsonManifest` just contains the local storage JSON value of the yaml editor, which means it contained outdated info.

The only reason this was working on master is because we were doing the `merge()` call above (which was removed in this PR), which caused `builderFormValues === lastValidBuilderFormValues` to always evaluate to `false`. Now that the merge is removed, that evaluates to true when there are no errors, so it was returning the `jsonManifest` value which was outdated, resulting in weird behavior.

lmossman (approved):
@flash1293 I pushed one change here and left a few other comments as well. I'm going to merge this as is once the build is green in order to unblock the release, so if there are any other follow-up changes you want to make based on the comments I left then we can just do that in a follow-up PR.

After pushing my fix, it seems to be working as expected from my local testing, and it seems to be nice and snappy. Nice work!

PR #20968, opened 2023-01-03
fix build: OAuthHandler missing import

pedroslopez:
## What

When #20932 and #20933 went in, one removed an import that the other was using, causing a failed build on master.

## How

Fix the build by adding the import again


jdpgrailsdev (approved):
:shipit:

PR #20978, opened 2023-01-03
Remove definitions from getEntityTableData method

### LINKED ISSUE #20667: Remove calls to "list_by_workspace" and "list_latest" from Sources and Destination list pages
[getEntityTableData](https://github.com/airbytehq/airbyte/blob/a8c72121f8fd50407425250d3a6ddc3756db8b55/airbyte-webapp/src/components/EntityTable/utils.tsx#L17) currently makes calls to `list_by_workspace` and `list_latest` solely to get the connector icon. The connector icon is available on the source or destination object directly so the calls should be removed and the icon retrieved from the source or destination object.

krishnaglick:
## What
Resolves #20667

## How
The `icon` property is available on the top level source or destination so there was no need to load the source or destination definition.

### DIFF
--- a/airbyte-webapp/src/components/EntityTable/utils.tsx
+++ b/airbyte-webapp/src/components/EntityTable/utils.tsx
@@ -1,9 +1,7 @@
 import {
   ConnectionStatus,
-  DestinationDefinitionRead,
   DestinationRead,
   JobStatus,
-  SourceDefinitionRead,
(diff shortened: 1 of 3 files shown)


krishnaglick:
Holidays made me forget that https://github.com/airbytehq/airbyte/pull/20264 does this.

PR #20986, opened 2023-01-03
🪟 🐛 Show correct page header for creating connection from connector page

### LINKED ISSUE #19796: Incorrect page title displayed on the Create Connection page
## Environment
- **Deployment**: OSS
- **Step where error happened**: Setup new connection 

## STR
Go to Destination
Select existing Destination
Click "add source" button
Select existing source
Check page title

## Current Behaviour
Page title - New Source
Screenshot 2022-11-24 at 19 08 54

## Expected Behaviour
Page title - New Connection

teallarson:
## What
closes #19796 

Previously, creating a connection from a Source or Destination page (by clicking "add a Destination" or "add a Source" respectively) would display an incorrect title in the PageHeader component when a user was configuring their Connection.

This new flow follows the same page header pattern as for the "Create Connection" workflow when initiated from the Connections page.

Before:
Screenshot 2023-01-03 at 1 42 50 PM

After:
Screenshot 2023-01-03 at 1 36 42 PM

Note:
When creating a new Source and/or Destination during the Connection creation workflow, the "Create connection" page header always shows.  This is the same as what exists currently on master and matches the [current design](https://www.figma.com/file/wihAJg5mkp61I7lLHs0IpV/FREE-CONNECTOR-PROGRAM?node-id=1080%3A15370&t=u1eTzN0UCETkplHn-0) as well (layout is different, but header text is the same)

### DIFF
--- a/airbyte-webapp/src/pages/ConnectionPage/pages/CreationFormPage/CreationFormPage.tsx
+++ b/airbyte-webapp/src/pages/ConnectionPage/pages/CreationFormPage/CreationFormPage.tsx
@@ -222,7 +222,6 @@ export const CreationFormPage: React.FC = () => {
           } as Record<EntityStepsTypes, string>
         )[type];
 
-  console.log({ currentStep }, { type });
   return (
     <>
       <HeadTitle titles={[{ id: "connection.newConnectionTitle" }]} />
--- a/airbyte-webapp/src/pages/ConnectionPage/pages/CreationFormPage/CreationFormPage.tsx
+++ b/airbyte-webapp/src/pages/ConnectionPage/pages/CreationFormPage/CreationFormPage.tsx
@@ -211,14 +211,18 @@ export const CreationFormPage: React.FC = () => {
           },
         ];
 
-  const titleId: string = (
-    {
-      [EntityStepsTypes.CONNECTION]: "connection.newConnectionTitle",
-      [EntityStepsTypes.DESTINATION]: "destinations.newDestinationTitle",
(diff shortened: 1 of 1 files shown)


airbyte-webapp/src/pages/ConnectionPage/pages/CreationFormPage/CreationFormPage.tsx
```diff
@@ -211,13 +211,16 @@ export const CreationFormPage: React.FC = () => {
...
+  const titleId: string =
+    currentStep === "createConnection"
+      ? "connection.newConnectionTitle"
+      : (
+          {
+            [EntityStepsTypes.CONNECTION]: "connection.newConnectionTitle",
+            [EntityStepsTypes.DESTINATION]: "destinations.newDestinationTitle",
+            [EntityStepsTypes.SOURCE]: "sources.newSourceTitle",
+          } as Record<EntityStepsTypes, string>
+        )[type];
```
krishnaglick, line 223: This works but I hate how this is handled across this component.

krishnaglick (approved):

airbyte-webapp/src/pages/ConnectionPage/pages/CreationFormPage/CreationFormPage.tsx
tealjulia: Agree.  I started looking at cleaning up some of the types and names here, but it snowballed and got pretty out of scope pretty quickly.

PR #20998, opened 2023-01-03
:window: :art: Connector Builder UI: Rename offset pagination fields

lmossman:
## What
When testing out pagination in the connector builder, I felt that using the terms "Page size" and "Page token" for every pagination strategy was somewhat confusing, because those terms only really makes sense when using the Page Increment strategy.

## How
I tried going through and renaming the options to what I felt made more sense for each pagination strategy. 

Before:
Screen Shot 2023-01-03 at 3 46 50 PM

After:
Screen Shot 2023-01-03 at 3 48 25 PM

Since the labels/descriptions now differ for each strategy, I had to encapsulate those shared elements into separate components.


lmossman:
@flash1293 @brianjlai lmk what you think here. My opinion is that these labels are better, but open to suggestions

flash1293 (approved):
LGTM - I wasn't 100% sure about calling the page size "limit" for offset increment - it's weird that it's different from the names of the same concept for other pagination strategies, but OTOH offset+limit is a standard terminology. Happy either way.

I think the names for the page token property are way better like this.

lmossman:
Yeah the "limit" term was what I was least sure about here - I went with it because of what you said about offset+limit being standard

PR #21004, opened 2023-01-04
📝 Add warning for working with local files and Docker Desktop (`/tmp` and `/private`)

sspaeti:
## What
I had lots of trouble with not sharing the `/private` folder with the docker desktop on my MacBook. After debugging for a while we found that not only the `/tmp` need to be shared, but also /private folder.

If others do not run into that error, this PR should add a warning wherever necessary, but mostly for local destinations such as File CSV/JSON/Parquet, sqlite.

*It helps to add screenshots if it affects the frontend.*
image

### DIFF
--- a/docs/integrations/sources/file.md
+++ b/docs/integrations/sources/file.md
@@ -133,12 +133,14 @@ Please see (or add) more at `airbyte-integrations/connectors/source-file/integra
 In order to read large files from a remote location, this connector uses the [smart_open](https://pypi.org/project/smart-open/) library. However, it is possible to switch to either [GCSFS](https://gcsfs.readthedocs.io/en/latest/) or [S3FS](https://s3fs.readthedocs.io/en/latest/) implementations as it is natively supported by the `pandas` library. This choice is made possible through the optional `reader_impl` parameter.
 
 - Note that for local filesystem, the file probably have to be stored somewhere in the `/tmp/airbyte_local` folder with the same limitations as the [CSV Destination](../destinations/local-csv.md) so the `URL` should also starts with `/local/`.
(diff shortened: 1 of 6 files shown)


docs/integrations/sources/file.md
```diff
@@ -133,12 +133,14 @@ Please see (or add) more at `airbyte-integrations/connectors/source-file/integra
...
+- Please make sure that Docker Desktop has access to `/tmp` (and `/private` on a MacOS, as /tmp has a symlink that points to /private. It will not work otherwise). You allow it with "File sharing" in `Settings -> Resources -> File sharing -> add the one or two above folder` and hit the "Apply & restart" button.
 - The JSON implementation needs to be tweaked in order to produce more complex catalog and is still in an experimental state: Simple JSON schemas should work at this point but may not be well handled when there are multiple layers of nesting.
 
 ## Changelog
 
 | Version | Date       | Pull Request                                             | Subject                                                  |
 |:--------|:-----------| :------------------------------------------------------- |:---------------------------------------------------------|
+| 0.2.33  | 2023-01-04 | [21004](https://github.com/airbytehq/airbyte/pull/21004) | Source File: Adding Docker Desktop file share warning                |
```
evantahler, line 143: A change in docs isn't a version bump of the connector (and you didn't change the Dockerfile version or publish the connector)

evantahler (changes requested):

docs/integrations/sources/file.md
sspaeti: Ohh yes, I removed that 💯.

evantahler (approved):

PR #21021, opened 2023-01-04
Re-authenticating using the modern flow correcly sets form as dirty

krishnaglick:
## What
Resolves https://github.com/airbytehq/oncall/issues/1232

## How
Properly triggers a change to the form values object so Formik knows it's dirty by merging the values into a fresh object.


lmossman (approved):
Change looks good to me, did not test locally

PR #21027, opened 2023-01-04
Filter connections list on source or destination item pages

### LINKED ISSUE #21025: Utilize connection list filtering on source and destination item pages
#19850 has given the frontend the ability to filter connections based on a source or destination ID.
We should implement this work on the `DestinationOverviewPage` and the `SourceItemPage`.

As this is part of the ui/performance epic if there are additional optimizations to be made this ticket is a good place for it.

krishnaglick:
Resolves #21025

## What
Adds filtering to the source and destination item pages, as well as cleans up some excess API calls.

## How
The source or destination ID is now passed into the connections list call on the source and destination item pages.


teallarson (approved):
LGTM.

Had one suggested about comments but not blocking.

PR #21034, opened 2023-01-04
🪟 🧹 Custom connectors in Cloud UI updates

teallarson:
## What
closes airbytehq/airbyte-cloud#3753

- [x] Connector doc URL should not be required for custom connectors on cloud
- [x] More useful placeholder text
- [x]  Update field label for docker repository name --> docker full image name
- [x] Check if docs URL is a valid URL for custom connectors 
- [x] Denote which fields are required in the form
- [x] Docs side panel shouldn't show for custom connectors

Screenshot 2023-01-04 at 4 42 26 PM

## How
- Conditional yup schema, label, and placeholder in the modal form based on whether we are in Cloud mode.
- Also added some better visual feedback for validation errors (unblocking submit button as soon as the form is dirty, adding input highlighting and messaging for errors)

## Recommended reading order
1. `<CreateConnectorModal />`
2. everything else


airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
```diff
@@ -117,7 +122,7 @@ const CreateConnectorModal: React.FC<IProps> = ({ onClose, onSubmit, errorMessag
           }}
           validateOnBlur
           validateOnChange
-          validationSchema={validationSchema}
+          validationSchema={isCloudApp() === false ? standardValidationSchema : customConnectorValidationSchema}
```
tealjulia, line 125: In general, I would use the `FeatureService` over `isCloudApp()`, but the `AllowCustomImageUpload` doesn't tell us if a user is on Cloud specifically.  Open to a better check if someone has one!

airbyte-webapp/src/locales/en.json
```diff
@@ -507,10 +507,12 @@
   "admin.connectorName": "Connector display name",
   "admin.connectorName.placeholder": "Connector name",
   "admin.dockerRepository": "Docker repository name",
+  "admin.dockerFullImageName": "Docker full image name",
   "admin.dockerRepository.placeholder": "airbytehq/postgres",
+  "admin.dockerFullImageName.placeholder": "customer-name/customer-image",
   "admin.dockerImageTag": "Docker image tag",
   "admin.dockerImageTag.placeholder": "12.3",
-  "admin.documentationUrl": "Connector Documentation URL",
+  "admin.documentationUrl": "Connector documentation URL",
```
tealjulia, line 515: casing was inconsistent with other labels

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
```diff
@@ -139,32 +147,40 @@ const CreateConnectorModal: React.FC<IProps> = ({ onClose, onSubmit, errorMessag
                           <FormattedMessage id="admin.connectorName" />
                         </Label>
                       }
+                      error={meta.touched && !!meta.error}
+                      message={meta.touched && meta.error && <FormattedMessage id={meta.error} />}
```
tealjulia, line 151: let's help users out a little here :)

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
```diff
@@ -219,7 +239,7 @@ const CreateConnectorModal: React.FC<IProps> = ({ onClose, onSubmit, errorMessag
                   >
                     <FormattedMessage id="form.cancel" />
                   </Button>
-                  <Button type="submit" disabled={isSubmitting || !dirty || !isValid} isLoading={isSubmitting}>
+                  <Button type="submit" disabled={isSubmitting || !dirty} isLoading={isSubmitting}>
```
tealjulia, line 242: I'm not a fan of making this button disabled, and also not giving users any visual hints what they need to do to enable it.  Willing to be persuaded otherwise, but I think this is better.

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/ImageCell.tsx
```diff
@@ -18,6 +18,10 @@ const Link = styled.a`
 `;
 
 const ImageCell: React.FC<ImageCellProps> = ({ imageName, link }) => {
+  if (!link || !link.length) {
```
tealjulia, line 21: This is the cell in the tables on /settings/sources and /settings/destinations with the docker repository image name and the docs link.

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/ConnectorDocumentationLayout.tsx
```diff
@@ -14,9 +14,10 @@ const LazyDocumentationPanel = lazy(() =>
 
 export const ConnectorDocumentationLayout: React.FC<React.PropsWithChildren<unknown>> = ({ children }) => {
   const { formatMessage } = useIntl();
-  const { documentationPanelOpen } = useDocumentationPanelContext();
+  const { documentationPanelOpen, documentationUrl } = useDocumentationPanelContext();
   const screenWidth = useWindowSize().width;
-  const showDocumentationPanel = screenWidth > 500 && documentationPanelOpen;
+  const showDocumentationPanel =
+    screenWidth > 500 && documentationPanelOpen && documentationUrl.includes("docs.airbyte.com");
```
tealjulia, line 20: These documentation URLs come from the source/destination_definition seed yaml files, so the internal docs _always_ include `docs.airbyte.com` at this level.  (the panel itself does the stripping of the URL and pulling from the build)

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
```diff
@@ -84,13 +80,21 @@ const ErrorText = styled.div`
   max-width: 400px;
 `;
 const validationSchema = yup.object().shape({
-  name: yup.string().required("form.empty.error"),
-  documentationUrl: yup.string().required("form.empty.error"),
-  dockerImageTag: yup.string().required("form.empty.error"),
-  dockerRepository: yup.string().required("form.empty.error"),
+  name: yup.string().trim().required("form.empty.error"),
+  documentationUrl: yup.string().trim().url("form.url.error").notRequired(),
```
tealjulia, line 84: Note: this makes this field optional for OSS as well, which also does not appear to do anything with it beyond the link on the connectors list in settings + the docs panel.

tealjulia:
Note: I have not been able to manually test the documentation panel behavior for a custom connector on cloud, however I have verified using OSS, including the behavior for a connector with no documentation URL and an external documentation URL.

Custom connectors on cloud dev environments is blocked by airbytehq/airbyte-cloud/pull/3952

There is a workaround but I feel confident having seen the behavior in OSS.

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
```diff
@@ -84,13 +80,21 @@ const ErrorText = styled.div`
...
+  dockerImageTag: yup.string().trim().required("form.empty.error"),
+  dockerRepository: yup.string().trim().required("form.empty.error"),
 });
 
-const CreateConnectorModal: React.FC<IProps> = ({ onClose, onSubmit, errorMessage }) => {
+const Label: React.FC<React.PropsWithChildren<unknown>> = ({ children }) => {
+  return (
+    <Text as="span" bold size="lg" className={styles.label}>
```
tealjulia, line 91: `as="span"` allows us to use the built in `message` prop on the input to display any validation errors

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
```diff
@@ -116,18 +120,18 @@ const CreateConnectorModal: React.FC<IProps> = ({ onClose, onSubmit, errorMessag
             dockerRepository: "",
           }}
           validateOnBlur
-          validateOnChange
+          validateOnChange={false}
```
josephkmh, line 123: What's the reason for disabling this? Seems like the default `true` should be fine, since the error message is only shown if the field is `touched` (which happens on blur). Actually both `validateOnBlur` and `validateOnChange` default to `true`, so they could both be removed unless they are explicitly being set to `false`.

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
```diff
@@ -139,32 +143,40 @@ const CreateConnectorModal: React.FC<IProps> = ({ onClose, onSubmit, errorMessag
...
-                  {({ field }: FieldProps<string>) => (
+                  {({ field, meta }: FieldProps<string>) => (
                     <LabeledInput
                       {...field}
                       type="text"
                       autoComplete="off"
                       placeholder={formatMessage({
-                        id: "admin.dockerRepository.placeholder",
+                        id: `${
+                          isCloudApp() ? "admin.dockerFullImageName.placeholder" : "admin.dockerRepository.placeholder"
```
josephkmh, line 161: I thought more about this, and I couldn't think of a better solution using `FeatureService`. Similar to the [OSS/cloud specific markdown rendering](https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/DocumentationPanel.tsx#L70), I think it's ok to use `isCloudApp()` here.

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
```diff
@@ -177,13 +189,15 @@ const CreateConnectorModal: React.FC<IProps> = ({ onClose, onSubmit, errorMessag
                           <FormattedMessage id="admin.dockerImageTag" />
                         </Label>
                       }
+                      error={!!meta.error && meta.touched}
+                      message={meta.touched && meta.error && formatMessage({ id: meta.error })}
```
josephkmh, line 193: nit: any reason to use `formatMessage()` here and `<FormattedMessage>` on lines 147 and 172?

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
```diff
@@ -219,7 +235,7 @@ const CreateConnectorModal: React.FC<IProps> = ({ onClose, onSubmit, errorMessag
                   >
                     <FormattedMessage id="form.cancel" />
                   </Button>
-                  <Button type="submit" disabled={isSubmitting || !dirty || !isValid} isLoading={isSubmitting}>
+                  <Button type="submit" disabled={isSubmitting || !dirty} isLoading={isSubmitting}>
```
josephkmh, line 238: Any specific reason to remove `!isValid` as a reason to disable submission? AFAIK we usually disable submission buttons if the form is not valid (but I know this is not 100% consistent).

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/ImageCell.tsx
```diff
@@ -18,6 +18,10 @@ const Link = styled.a`
...
+    return <>{imageName}</>;
```
josephkmh, line 22: Should we wrap the `imageName` in `<Text>`? Having issues uploading a custom connector locally to test this, but I believe we don't want raw text inside the `<td>`

```suggestion
    return <Text>{imageName}</Text>;
```

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/ConnectorDocumentationLayout.tsx
josephkmh: Seems like a fine way to detect whether to show the panel! Just a minor suggestion to hint at why we're looking for this substring, take it or leave it:

```suggestion
    const isOfficialDocumentation = documentationUrl.includes("docs.airbyte.com");
    screenWidth > 500 && documentationPanelOpen && isOfficialDocumentation;
```

joeykmh (changes requested):
Looking good! Just left a few minor comments. I tested it locally and the form UX is much nicer ✨

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
tealjulia: Nope, I missed this when I was cleaning the others from a copy/paste!

tealjulia: Nope!  I missed some cleanup from a copy/paste.  Thanks.

airbyte-webapp/src/locales/en.json
```diff
@@ -507,10 +507,12 @@
   "admin.connectorName": "Connector display name",
   "admin.connectorName.placeholder": "Connector name",
   "admin.dockerRepository": "Docker repository name",
+  "admin.dockerFullImageName": "Docker full image name",
   "admin.dockerRepository.placeholder": "airbytehq/postgres",
+  "admin.dockerFullImageName.placeholder": "customer-name/customer-image",
```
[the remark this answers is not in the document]
tealjulia: Agree, but I'll wait for those tagged to chime in before I make the change!

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/ImageCell.tsx
tealjulia: I rewrote this component to use the `<Text/>` component and get the styled components out of there in https://github.com/airbytehq/airbyte/pull/21034/commits/321b68f80f32a1cf8690f72abce29371314b07d6

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
tealjulia: Yep neither of those lines are needed.  Removed.

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
tealjulia: It's not entirely consistent and I _think_ we were moving away from disabling the button and towards having more consistent and useful validation?  But I've turned it back on for now as that wasn't part of this ticket and instead added "Required" labels to the required fields.

Screenshot 2023-01-09 at 9 47 59 AM

There's a funky state when a user tabs into the URL space and starts typing (the button will re-disable until the URL is valid and a user won't see the error message until they tab or click out of that field) but I don't feel too strongly on it.

tealjulia:
Ready for re-review pending input on what to put for placeholder text.

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
josephkmh: How about displaying the error message for the docs URL regardless of `meta.touched`? I think that's slightly nicer UX, since it's an optional field that can only become invalid once you start typing.

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/ImageCell.tsx
josephkmh: Nice!

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
tealjulia: 👍 that'll work.  Added in https://github.com/airbytehq/airbyte/pull/21034/commits/f9fb974b513548a54dd3b7e3153ca1ac0a0e79bc

airbyte-webapp/src/locales/en.json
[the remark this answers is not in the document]
sophia-wiley: I agree that using `Docker image name` is the best way to go!

joeykmh (approved):
Looks good!

PR #21036, opened 2023-01-04
Update form field labels in connection form to match design

### LINKED ISSUE #19900: Update form field labels in connection form to match design
Update label design to match the latest in Figma:

* Description should be shown in tooltip
* Required fields no longer marked with *

Current:
image

Proposed:
image

Figma:
https://www.figma.com/proto/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?page-id=857%3A24319&node-id=865%3A52984&viewport=124%2C-519%2C0.25&scaling=min-zoom&starting-point-node-id=865%3A52984&show-proto-sidebar=1

m19berger:
## What
Closes [#19900](https://github.com/airbytehq/airbyte/issues/19900)

## How
Replaced ControlLabels message property with infoTooltipContent

_For test purposes, use the environment variable:_

- REACT_APP_NEW_STREAMS_TABLE=true
- REACT_APP_NEW_STREAMS_TABLE=false

[Figma design ](https://www.figma.com/proto/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?node-id=874%3A35099&scaling=min-zoom&page-id=857%3A24319&starting-point-node-id=857%3A25233&show-proto-sidebar=1)

## Loom
https://www.loom.com/share/ea52bb672849427e91a2d72b603f40ce

https://www.loom.com/share/9e08684e93dc48f5b4944bc7cbb12fea

image

image

image

image


YatsukBogdan1 (approved):
looks good to me

edmundito:
@matter-q from looking at the changes, the namespace/stream name prefix fields also need to be updated so that they are consistent with the current table.

matter-q:
> @matter-q from looking at the changes, the namespace/stream name prefix fields also need to be updated so that they are consistent with the current table.

I didn't quite understand what needed to be updated... 
Could you be more specific?

edmundito:
When the new streams table feature is disabled, there are more fields.

Also, while creating a new connection, there are extra name and data residency fields. Here's what it looks like from the new connection page:

image

Also, the vertical alignment of the items should be centered but it is not:
image

edmundito (commented):
The changes look good, and tested locally. There is one more thing in terms of how the fields are presented in the new design. Required fields no longer need the `*` next to it and instead the optional fields are marked as `optional` (there is a prop in the ControlLabels to do this). Here's an example with the Destination Stream Prefix:

Screen Shot 2023-01-09 at 10 36 10

airbyte-webapp/src/views/Connection/ConnectionForm/ConnectionFormFields.tsx
```diff
@@ -72,14 +72,15 @@ export const ConnectionFormFields: React.FC<ConnectionFormFieldsProps> = ({ valu
             {values.namespaceDefinition === NamespaceDefinitionType.customformat && (
               <Field name="namespaceFormat">
                 {({ field, meta }: FieldProps<string>) => (
-                  <FlexContainer alignItems="flex-start">
+                  <FlexContainer alignItems="center">
                     <div className={styles.leftFieldCol}>
                       <ControlLabels
                         className={styles.namespaceFormatLabel}
                         nextLine
+                        optional
```
edmundito, line 80: Not optional

```suggestion
```

edmundito (changes requested):
Tested locally, but noticed one small thing...

edmundito:
Sorry, just tested the cloud version and noticed a couple of things on the connection settings page:

image

1. The tooltip should now use the "learn more" link for tooltips: https://63053d1047904393cf8d8faa-hpmeogsbgz.chromatic.com/?path=/story/ui-tooltip--with-learn-more-url
2. The spacing between the label and the dropdown needs to be adjusted. The dropdown now takes whatever is left of the card, but it should be a specific size (300px) and looks like removing the description broke the layout

edmundito (approved):

PR #21047, opened 2023-01-05
Switch order of namespaces and make Destination Default is the defaul…

### LINKED ISSUE #21030: Change the default replication type from "Mirror Source Structure" to "Destination Default" on UI
Change the default replication type from "Mirror Source Structure" to "Destination Default" on UI. This would be a workaround to decrease onCall issues related to this issue: https://github.com/airbytehq/airbyte/issues/20561

Additionally, we may add some notification warn msg for a user that he needs to make sure that the destination doesn't already contain the same schema name as source is going to replicate as part of the "Mirror Source Structure" sync approach. 

Currently, we have a "Check" connection stage in destination connectors, but we check the connection for the schema provided in the connector's config. 

Having the "Mirror Source Structure" by default we may end up with the case when we passed the "Check" connection stage, but then fail because will try to write to another schema that may be not allowed for us. 

Thanks

Selection_189

grishick:
Fixes https://github.com/airbytehq/airbyte/issues/21030

## What

- Change the order of options to make Destination Default the first option
- Make Destination Default the option selected by default instead of Mirror Source Structure

### DIFF
--- a/airbyte-webapp/src/components/connection/DestinationNamespaceModal/DestinationNamespaceModal.tsx
+++ b/airbyte-webapp/src/components/connection/DestinationNamespaceModal/DestinationNamespaceModal.tsx
@@ -69,22 +69,6 @@ export const DestinationNamespaceModal: React.FC<DestinationNamespaceModalProps>
                   <LabeledRadioButton
                     {...field}
                     className={styles.radioButton}
-                    id="destinationNamespace.source"
-                    label={
-                      <Text as="span">
-                        <FormattedMessage id="connectionForm.modal.destinationNamespace.option.source" />
-                      </Text>
-                    }
-                    value={NamespaceDefinitionType.source}
-                    checked={field.value === NamespaceDefinitionType.source}
-                  />
-                )}
-              </Field>
-              <Field name="namespaceDefinition">
-                {({ field }: FieldProps<string>) => (
-                  <LabeledRadioButton
-                    {...field}
-                    className={styles.radioButton}
                     id="destinationNamespace.destination"
                     label={
                       <Text as="span">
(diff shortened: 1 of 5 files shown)


evantahler:
Your frontend tests are failing for the right reasons! the order of the options in the form has changed.  

You can re-record those snapshots with `./node_modules/.bin/jest --updateSnapshot`

grishick:
> Your frontend tests are failing for the right reasons! the order of the options in the form has changed.
> 
> You can re-record those snapshots with `./node_modules/.bin/jest --updateSnapshot`. Check in the updates.

Thanks!

airbyte-webapp/src/components/connection/DestinationNamespaceModal/DestinationNamespaceModal.tsx
```diff
@@ -48,7 +48,7 @@ export const DestinationNamespaceModal: React.FC<DestinationNamespaceModalProps>
   return (
     <Formik
       initialValues={{
-        namespaceDefinition: initialValues?.namespaceDefinition ?? NamespaceDefinitionType.source,
+        namespaceDefinition: initialValues?.namespaceDefinition ?? NamespaceDefinitionType.destination,
```
josephkmh, line 51: This will correctly set `NamespaceDefinitionType.destination` as the default, but it still appears as the second option in the list due to the order in the HTML below:

image

To make it the first option, you could move the destination `<Field>` component (lines 83-98) before the source component (lines 67-82).

Note: this is only visible in the upcoming stream table redesign, so you need to set `REACT_APP_NEW_STREAMS_TABLE=true` in `airbyte-webapp/.env.development` before spinning up the server with `npm run start` or `npm run start:cloud` to see this new modal UI.

airbyte-webapp/src/views/Connection/ConnectionForm/formConfig.tsx
```diff
@@ -138,8 +138,8 @@ export const createConnectionValidationSchema = ({
       namespaceDefinition: yup
         .string()
         .oneOf([
-          NamespaceDefinitionType.source,
           NamespaceDefinitionType.destination,
+          NamespaceDefinitionType.source,
           NamespaceDefinitionType.customformat,
```
josephkmh, line 143: Same here, not strictly necessary to change the order for the validation schema, but doesn't hurt.

joeykmh (approved):
Looks good! Pulled and tested locally, works as expected. I made one optional suggestion for reordering some radio inputs.

PR #21048, opened 2023-01-05
Remove workspace helper from fetchConfigActivity

alovew:
- Remove WorkspaceHelper, which requires configRepository, and replace it with WorkspaceApi
- Remove getStandardSync method from fetchConfigActivity, which relies on configRepository
- Add getBreakingChange method to fetch the breaking change attr from a connection using the connectionApi

### DIFF
--- a/airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionNotificationWorkflowImpl.java
+++ b/airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionNotificationWorkflowImpl.java
@@ -16,6 +16,7 @@ import io.airbyte.workers.temporal.annotations.TemporalActivityStub;
 import io.airbyte.workers.temporal.scheduling.activities.ConfigFetchActivity;
 import io.airbyte.workers.temporal.scheduling.activities.NotifySchemaChangeActivity;
 import io.airbyte.workers.temporal.scheduling.activities.SlackConfigActivity;
+import io.temporal.workflow.Workflow;
 import java.io.IOException;
 import java.util.Optional;
 import java.util.UUID;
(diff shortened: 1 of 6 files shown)


airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionNotificationWorkflowImpl.java
```diff
@@ -35,14 +34,14 @@ public class ConnectionNotificationWorkflowImpl implements ConnectionNotificatio
   @Override
   public boolean sendSchemaChangeNotification(final UUID connectionId)
       throws IOException, InterruptedException, ApiException, ConfigNotFoundException, JsonValidationException {
-    final StandardSync standardSync = configFetchActivity.getStandardSync(connectionId);
+    final Optional<Boolean> breakingChange = configFetchActivity.getBreakingChange(connectionId);
     final Optional<SlackNotificationConfiguration> slackConfig = slackConfigActivity.fetchSlackConfiguration(connectionId);
-    if (slackConfig.isPresent()) {
+    if (slackConfig.isPresent() && breakingChange.isPresent()) {
```
benmoriceau, line 39: Do we need to check the NonBreakingChange preference?

alovew: no, at this point we're just using the breakingChange boolean

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionNotificationWorkflowImpl.java
```diff
@@ -35,14 +34,14 @@ public class ConnectionNotificationWorkflowImpl implements ConnectionNotificatio
   @Override
   public boolean sendSchemaChangeNotification(final UUID connectionId)
       throws IOException, InterruptedException, ApiException, ConfigNotFoundException, JsonValidationException {
-    final StandardSync standardSync = configFetchActivity.getStandardSync(connectionId);
+    final Optional<Boolean> breakingChange = configFetchActivity.getBreakingChange(connectionId);
```
benmoriceau, line 37: I think that there is a (small) risk to have workflow being stuck because of a lack of version. could we add one in order to avoid any issue.

benmoriceau (approved):

PR #21050, opened 2023-01-05
[Low-Code CDK] Construct declarative components from Pydantic models

### LINKED ISSUE #19411: [Low-Code CDK] Construct Declarative Components from the parsed Pydantic Models
Low Code Schema Refactor Phase 4

While parsing a manifest, once the manifest dictionary has been turned into a set of corresponding Pydantic models via `parse_obj()`, they need to be constructed into the declarative components that will be used at runtime. Once this phase is complete, there will be no need for the auto-generated schema or the original `factory.py` + `create_partial.py` which used to be responsible for instantiating components.

As part of https://github.com/airbytehq/airbyte/issues/20045, `manifest_decalarative_source.py`, will have the code to create Pydantic model instances and stand up much of the boilerplate for where component creation will be done. What is needed in this ticket is to define a _new_ constructor/factory class that can take in a model and return the generated declarative component. This runtime component should be functionally identical to the old flow. This change revolves around how we parse and create components only.

## Implementation Details

This will involve traversing the models to create every subcomponent. We need to support creation of components starting with the base instances used by ManifestDeclarativeSource methods. The generic entry point can be `ComponentFactory.create_component_from_model(BaseModel, Config`:

- CheckStream
- DeclarativeStream(s)
- Spec

This will be a large sized task where we will have to write constructors for every single declarative component. The work can only be merged in an all or nothing release, but we can parallelize writing some of the component constructors. Some initial general thoughts around the implementation

- Each component will have a corresponding constructor. Take HttpRequester for example: `create_http_requester(HttpRequesterModel, Config) -> HttpRequester`
- Subcomponents are created by invoking their constructor and then used as parameters to initialize the current component.
- Since there are multiple types of authenticators/pagination strategies/etc only known at runtime, we need a map of declarative models to component constructors
- It fits more naturally to put each component constructor in the classes themselves. However, this poses a potential issue with circular dependencies on this map resolving constructors. The factory might be a bit large, but it should still be easy to reason about as one could follow the path of each create method
- Custom components might be able to share the same create and will follow a basic pattern. Resolve the type from the `class_name`, get type hints, match model fields to type hints, invoke the type's constructor
- It is worth seeing if the existing `test_factory.py` can be leveraged to validate most of our changes using the new flow since it covers most components and some edge cases

## Acceptance Criteria

- [ ] Base models (CheckStream, DeclarativeStream, Spec) from ManifestDeclarativeSource can be transformed into respective declarative components
- [ ] Custom components are supported
- [ ] `spec` and `read` commands are run successfully against 5 connectors (Greenhouse, Delighted, Square, Posthog, Zenloop)

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/19411

## What
This PR migrates us from the old flow that relies on the legacy `factory.py` to take dereferenced components in the form of a dictionary, and directly turns them into declarative components. The new flow now takes the dereferenced components to turn them into pydantic models (done in the previous [PR](https://github.com/airbytehq/airbyte/pull/20747) but gated to not be used) and then turn these models into the declarative components. There should be parity between the new and old flows, although there are a few cases where we modify come  yaml config

### General Approach

- As is from the previous changes,`ModelToComponentFactory.create_component()` serves as the entry point invoked by Declarative Source to take a manifest dictionary and turn them into the relevant component
- A bulk of the changes involve adding individual component creation methods to the `model_to_component_factory` for every component in the language.
- These constructors are invoked as we traverse a component and build out its relevant subcomponents. Once all subcomponents are initialized, then we initialize the current component.
- The `PYDANTIC_MODEL_TO_CONSTRUCTOR` serves as a registry for a Pydantic model to the respective constructor class. This registry is necessary because we don't always know the model to create until runtime. For something like a SimpleRetriever w/ one option we know, but for a pagination strategy which can vary, we would need to map the strategy to the relevant connector

### Edge Cases

Custom Components:

- This is arguably the complex edge case during the refactor and there were some interesting considerations that we needed to make in order to retain backwards compatibility with the existing flow.
- Much of this functions as it does in the legacy factory by deriving the class and type hints from `class_name` field.
- The biggest nuance to be aware of is that we need to perform a second `parse_obj()` of subcomponent objects of custom components because the original `parse_obj()` done at the beginning cannot traverse any further than the subcomponent. That is because the schema of custom components is dynamic and will assign what was supposed to be subcomponents into a general dictionary. Hence why we need to do it twice
- Also prior to the second parse, we try to infer the component type when it is missing and append it before we attempt the second parse. Long term this should not be necessary because we want to require `type` on all components, but this is used to retain backwards compatibility
- 

Post processing Pydantic models:

- In the comments there is more detail, but right now we had to add some post-processing to the generated models due to a limitation of the Pydantic framework.

## Recommended reading order
1. `generate-component-manifest-files.sh`
2. `manifest_declarative_source.py`
3. `model_to_component_factory.py`
4. `manifest_component_transformer.py`
5. small bug fixes to various components we noticed while implementing


airbyte-cdk/python/bin/generate-component-manifest-files.sh
```diff
@@ -21,6 +21,10 @@ function main() {
       --use-title-as-name \
       --disable-timestamp \
       --enum-field-as-literal one
+
+    temp_file=$(mktemp)
+    sed 's/_options:/options:/g' "$ROOT_DIR/$OUTPUT_DIR/$filename_wo_ext.py" > "${temp_file}"
+    mv "${temp_file}" "$ROOT_DIR/$OUTPUT_DIR/$filename_wo_ext.py"
```
brianjlai, line 27: There is an interesting bug in Pydantic that prevents private fields values from being accessible. This means that we cannot invoke `._options` from the component constructors. This has been widely discussed for what looks like 2+ years on the pydantic repo [issue](https://github.com/pydantic/pydantic/issues/288).

Basically why this happens is that we use the `$options` key in our YAML manifests and the pydantic model generator automatically turns special chars into `_` even the first character, so now Pydantic thinks this is a private field. 

Catherine and I have explored a bunch of different options and stack ranked them:

1. the model generator has an option to emit fields w/o prefixing _
2. post-processing of the schema models in declarative_component_schema.py [what we have here]
3. Find a way to parse `model._options` as the underlying value
4. Use parse_obj() at each level of the models
5. Replace `$options` with `options` in the schema + migrate existing manifests
6. Replace `$options` with options in the schema + under the hood transform `$options` to `option` for incoming manifests

Our ideal solution is that the model generator gets rid of the leading underscore, however I've been unable to find a flag to do that. And I haven't been able to find a way of actually getting the value of `model._options` correctly from the constructors. That is why we're left with option 2.

airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/error_handlers/default_error_handler.py
```diff
@@ -92,7 +92,6 @@ class DefaultErrorHandler(ErrorHandler, JsonSchemaMixin):
 
     DEFAULT_BACKOFF_STRATEGY = ExponentialBackoffStrategy
 
-    config: Config
```
brianjlai, line 95: noticed that we define config twice. not related to the refactor, but just fixing things i find

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_component_transformer.py
```diff
@@ -56,7 +55,11 @@
     # AddFields
     "AddFields.fields": "AddedFieldDefinition",
     # CustomStreamSlicer
+    "CustomStreamSlicer.end_datetime": "MinMaxDatetime",
```
brianjlai, line 58: noticed a couple of connectors where we need to auto-infer these values. some explicitly type them, but some connectors don't so we need this to auto insert it when it doesn't exist

airbyte-integrations/connectors/source-greenhouse/source_greenhouse/greenhouse.yaml
```diff
@@ -46,6 +46,7 @@ definitions:
         inject_into: "path"
       url_base: "*ref(definitions.requester.url_base)"
   base_stream:
+    type: DeclarativeStream
```
brianjlai, line 49: This is a case worth noting. In this case, we have a custom substream slicer component which references a specific parent stream. For normal components, we append `type: DeclarativeStream` when it is missing, however, we cannot do type inference for custom components because the fields can be dynamic. So in the greenhouse (and other custom substream components), we need to explicitly define the type. We are also shifting in the direction of explicit type definitions for all manifests. Although added verbosity, this additional inference layer does not play as nicely with connector builder UI because it's an added layer of transformation on both sides.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -41,18 +122,581 @@ def create_component(self, model_type: Type[BaseModel], component_definition: Co
...
+        return False
+    return cls.__module__ == "builtins"
+
+
+def _create_nested_component(model, model_field: str, model_value: Any, config: Config) -> Any:
+    type_name = model_value.get("type", None)
+    # if not type_name:
+    #     raise ValueError(
+    #         f"Error while parsing custom component {model.class_name}. Subcomponent field '{model_field}' should have a 'type' specified"
+    #     )
```
brianjlai, line 277: Will remove this block later, but wanted to leave this in as a discussion point. Ideally we would like to flag any custom components that do not have a type, however we can't differentiate between a general dictionary mapping vs. a subcomponent that is missing it's type. If anyone has a clever idea for this, otherwise we just aren't able to surface the error quite as nicely as we would like

clnoll: Hmm so if `not type_name`, we'll basically always hit the `ValueError` on line 283. Seems to me that'd also be incorrect in the case of a dictionary. Is that right?

clnoll (commented):
Awesome, just a few comments/questions for you @brianjlai.

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, source_config: ConnectionDefinition, debug: bool = False, con
         self.logger = logging.getLogger(f"airbyte.{self.name}")
 
         # Controls whether we build components using the manual handwritten schema and Pydantic models or the legacy flow
-        self.construct_using_pydantic_models = construct_using_pydantic_models
+        self.construct_using_pydantic_models = True
```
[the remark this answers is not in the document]
brianjlai: Leaving it in actually so that we can still keep the legacy flow in the short term. But post-refactor we can clean this up and remove the old legacy flow and this conditional lock

airbyte-integrations/connectors/source-greenhouse/source_greenhouse/greenhouse.yaml
brianjlai: That's a good question. Something that I don't love is that existing connectors are inconsistent about usage of `type` vs inferred types. But going back and adding it to every connector will muddle up an already complicated change more. My first thought is we should modify existing templates and docs i this PR to always have type specified. 

And we should go back and add `type` in a follow up PR for connectors written, pre-beta release.

And just to note as to why this affected only `source-greenhouse` in this case is because this is the only case of a custom stream slicer which therefore is unable to infer the `DeclarativeStream` type. All other cases of substream slicer use the first class component which will automatically populate the `type: DeclarativeStream` during the transformation before we do all this.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_component_transformer.py
brianjlai: The guideline so far has just been to ensure backwards compatibility so we have been adding when we catch it during our custom component testing, but once we get this change through, the templates and docs should be updated and developers should always be specifying the `type` in every component. I think our original hypothesis of type inference might not have paid off because even though it's less verbose, we have lots of these little edge cases especially around custom components

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -41,18 +122,581 @@ def create_component(self, model_type: Type[BaseModel], component_definition: Co
...
+
+def create_component_from_model(model: BaseModel, config: Config, **kwargs) -> Any:
+    if model.__class__ not in PYDANTIC_MODEL_TO_CONSTRUCTOR:
+        raise ValueError(f"{model.__class__} with attributes {model} is not a valid component type")
+    component_constructor = PYDANTIC_MODEL_TO_CONSTRUCTOR.get(model.__class__)
+    return component_constructor(model=model, config=config, **kwargs)
+
+
+def create_added_field_definition(model: AddedFieldDefinitionModel, config: Config, **kwargs) -> AddedFieldDefinition:
+    interpolated_value = InterpolatedString.create(model.value, options=model.options)
```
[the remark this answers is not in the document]
brianjlai: In my opinion yes. I think the complexity of building should be encapsulated here and not the object in `__post_init__`, but trying to reduce the scope of changes in this PR for something we can clean up during a follow up

airbyte-cdk/python/bin/generate-component-manifest-files.sh
brianjlai: yup will add!

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
clnoll: Not that I'm aware of, but maybe I'm misunderstanding. I thought that was a situation you were referring to when you said "we can't differentiate between a general dictionary mapping vs. a subcomponent that is missing it's type".

clnoll: Not that I'm aware of @girarda but seems that it's theoretically possible and that's what we're protecting against - can you confirm @brianjlai?

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -88,6 +94,10 @@ definitions:
         type: string
       password:
         type: string
+        default: ""
```
maxi297, line 97: Why would we have an empty string by default here? Based on a quick check, my guess is that `InterpolatedString` does not take `None` as an input. If this is the case, does `None` and an empty string means the same thing for Basic Auth?

maxi297 (approved):

maxi297:
Note that the `Airbyte CI / Connectors Base: Build` was failing because of the twitter maven issue. I re-ran it just not and maybe the result will be more interesting

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
brianjlai: The general reasoning was that the [basic http authenticator implementation](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/token.py#L68) (even prior to low-code) was a concatenated string of username + password that was base64 encoded. So rather than have to translate password None into "" in that code, we just opted for the empty string default. And that carried into our low code implementation

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: I haven't seen any usage of it so far, but it is not an inconceivable use case. Right now in our schema, the only non-component raw dictionary usage we have is for request body/headers/parameters (request_options_provider) and refresh_request_body (oauth authenticator).

brianjlai: @clnoll that's a good catch... actually if we don't have a `type` we'll end up with `None`. My thoughts are that if type name ends up being None, then just return the original dictionary back, otherwise we continue to do the parse object resolution

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -41,18 +122,581 @@ def create_component(self, model_type: Type[BaseModel], component_definition: Co
...
+    SingleSliceModel: create_single_slice,
     SpecModel: create_spec,
+    SubstreamSlicerModel: create_substream_slicer,
+    WaitTimeFromHeaderModel: create_wait_time_from_header,
+    WaitUntilTimeFromHeaderModel: create_wait_until_time_from_header,
 }
+
+
+# Needed for the case where we need to perform a second parse on the fields of a custom component
+TYPE_NAME_TO_MODEL = {cls.__name__.partition("Model")[0]: cls for cls in PYDANTIC_MODEL_TO_CONSTRUCTOR}
```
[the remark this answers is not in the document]
brianjlai: good point. had not considered that, fixing!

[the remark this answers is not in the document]
brianjlai: actually on further examination, we actually don't even need to partition at all. The class name of the models don't actually end in Model, the model is aliased just to differentiate the models from the real components, so the following w/o any partitioning should work:
```
{cls.__name__: cls for cls in PYDANTIC_MODEL_TO_CONSTRUCTOR}
```

airbyte-integrations/connectors/source-zenloop/source_zenloop/components.py
```diff
@@ -8,11 +8,14 @@
...
 from airbyte_cdk.models import SyncMode
 from airbyte_cdk.sources.declarative.stream_slicers.substream_slicer import SubstreamSlicer
-from airbyte_cdk.sources.declarative.types import StreamSlice, StreamState
+from airbyte_cdk.sources.declarative.types import Config, StreamSlice, StreamState
 
 
 @dataclass
 class ZenloopSubstreamSlicer(SubstreamSlicer):
+
+    config: Config
```
brianjlai, line 17: This was an interesting case. We used to extract the config via options because it was stored there. This seemed like a misuse of the options vs. config separation, instead like other components, `config` can be passed via parameter separately and we retrieve it that way for this custom component

PR #21051, opened 2023-01-05
Better escape the select query for a case a schema name starts with a number

rodireich:
## What
Safer escaping for schema names starting with a number.


subodh1810 (approved):
Approved once we add a test case where we have a schema starting with a digit

PR #21058, opened 2023-01-05
🪟 🎉 Add column selection UI to new stream table

### LINKED ISSUE #20584: Add column selection UI to new stream table
There is a new stream table in the works: https://github.com/airbytehq/airbyte/issues/18187

Currently the column selection UI is only implemented in the "old" stream table. We need to integrate column selection into the new stream table before it can be rolled out.

Figma designs: https://www.figma.com/proto/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?node-id=972%3A36212&scaling=min-zoom&page-id=972%3A36208&starting-point-node-id=972%3A36212&show-proto-sidebar=1

joeykmh:
## What
Closes https://github.com/airbytehq/airbyte/issues/20584

Adds column selection UI to the new stream table design.

https://user-images.githubusercontent.com/7550957/212195952-c4dcb4c8-b859-4219-ad7a-02187c630803.mov

## How
- Adds a column to the new `StreamFieldsTable` for a "Sync" checkbox for each top-level field
- Updates the stream config when the checkbox is toggled
- Adds a checkbox in the header to select/deselect all fields
- Does some conditional checks based on `syncMode` and `destinationSyncMode` - primary keys and cursors in these modes should not be deselected
- Adds unit tests for various scenarios of selecting/deselecting fields


airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
```diff
@@ -112,44 +112,44 @@ const CatalogSectionInner: React.FC<CatalogSectionInnerProps> = ({
...
+  const onToggleFieldSelected = useCallback(
+    (fieldPath: string[], isSelected: boolean) => {
+      if (!config) {
+        return;
+      }
+      const updatedConfig = updateFieldSelected({ config, fields, fieldPath, isSelected, numberOfFieldsInStream });
+      updateStreamWithConfig(updatedConfig);
+    },
+    [config, fields, numberOfFieldsInStream, updateStreamWithConfig]
+  );
```
josephkmh, line 151: This has been extracted to `streamConfigHelpers` to make unit testing easier (see `streamConfigHelpers.test.ts`)

airbyte-webapp/src/components/connection/CatalogTree/streamConfigHelpers/streamConfigHelpers.test.ts
```diff
@@ -15,34 +18,54 @@ const mockStreamConfiguration: AirbyteStreamConfiguration = {
...
+  cleanedName: "field_two",
+  key: "field_two",
+  type: "string",
+};
+const FIELD_THREE: SyncSchemaField = {
+  cleanedName: "field_three",
+  type: "todo",
+  key: "field_three",
+  path: ["field_three"],
+};
```
josephkmh, line 38: I redefined `FIELD_ONE`, `FIELD_TWO` and `FIELD_THREE` in this file, which is one of the reasons for the large diff.

airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
```diff
@@ -51,8 +51,14 @@ const CatalogSectionInner: React.FC<CatalogSectionInnerProps> = ({
...
   changedSelected,
 }) => {
+  const { stream, config } = streamNode;
   const isNewTableDesignEnabled = useNewTableDesignExperiment();
 
+  const fields = useMemo(() => {
+    const traversedFields = traverseSchemaToField(stream?.jsonSchema, stream?.name);
+    return traversedFields.sort(naturalComparatorBy((field) => field.cleanedName));
+  }, [stream?.jsonSchema, stream?.name]);
+
```
josephkmh, line 61: This has just been moved up in the file, not changed!

airbyte-webapp/src/components/connection/CatalogTree/next/StreamDetailsPanel/StreamFieldsTable/StreamFieldsTable.tsx
```diff
@@ -28,32 +36,48 @@ export interface TableStream {
...
   );
+  const checkIsFieldSelected = useCallback(
+    (field: SyncSchemaField): boolean => {
+      // All fields are implicitly selected if field selection is disabled
+      if (!config?.fieldSelectionEnabled) {
+        return true;
+      }
+
+      // path[0] is the top-level field name for all nested fields
+      return !!config?.selectedFields?.find((f) => isEqual(f.fieldPath, [field.path[0]]));
```
edmundito, line 77: Nit: You could also use `.some` instead of `!!object.find`

airbyte-webapp/src/components/connection/CatalogTree/streamConfigHelpers/streamConfigHelpers.test.ts
```diff
@@ -283,33 +306,159 @@ describe(`${toggleFieldInPrimaryKey.name}`, () => {
...
-        primaryKey: [FIELD_ONE, FIELD_THREE],
+        primaryKey: [FIELD_ONE.path, FIELD_THREE.path],
         selectedFields: [],
         fieldSelectionEnabled: false,
       });
     });
   });
 });
+
+describe(`${updateFieldSelected.name}`, () => {
```
edmundito, line 340: I've been using this convention:

Function: `#${myFunction.name}`
Component: `<${MyComponent.name} />`

edmundito (commented):
Code looks good but there were a few issues I noticed during testing.

Can you follow up with Nico about the "Sync" column? I'm not sure if that is outdated design so I'm not sure what exactly is the intent there (previous designs had a switch instead of checkbox)
image

The field count needs to appear in the main table. Is this intended to be done separately?
image

I tried just unselecting my primary key and saving and got an error. But the error is a little vague. It works for one stream, but what about multiple streams? One option is to invalidate the selected primary key immediately so that the user has to update the key right away (Probably can be addressed separately.)

image

After you delete a connection and go back to it (we call this readonly mode), you can still check and uncheck the checkboxes: 
image

When I check and uncheck the boxes, the save changes button is enabled instead of going back to being disabled:
image

Let me know which ones you want to address in this PR and which will be addressed separately.

josephkmh:
> Can you follow up with Nico about the "Sync" column? I'm not sure if that is outdated design so I'm not sure what exactly is the intent there (previous designs had a switch instead of checkbox)

Confirmed with @Upmitt that checkboxes is correct, latest figma designs [here](https://www.figma.com/proto/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?node-id=972%3A36212&scaling=min-zoom&page-id=972%3A36208&starting-point-node-id=972%3A36212&show-proto-sidebar=1)

josephkmh:
> I tried just unselecting my primary key and saving and got an error. But the error is a little vague. It works for one stream, but what about multiple streams? One option is to invalidate the selected primary key immediately so that the user has to update the key right away (Probably can be addressed separately.)
> 
> image


This is definitely unexpected, and deselection of the primary key or cursor field should be prevented on the client side already: 
image

Could you let me know steps to reproduce this?

josephkmh:
> When I check and uncheck the boxes, the save changes button is enabled instead of going back to being disabled
Good catch, this one is because the backend does not return `selectedFields` in `stream.config` if `fieldSelectionEnabled: false`. I'm checking with @mfsiega-airbyte if we should reliably return this field from the backend. If not, we can fix this by removing `selectedFields: []` when `fieldSelectionEnabled: false`

airbyte-webapp/src/components/connection/CatalogTree/streamConfigHelpers/streamConfigHelpers.test.ts
josephkmh: Not sure I quite follow here - are you suggesting that I put a `#` before the function name in the test title? Something like this?
```
describe(`#${updateFieldSelected.name}`, () => {
```
I can't find any other usage like this in the codebase, so I think I'm misunderstanding here.

airbyte-webapp/src/components/connection/CatalogTree/next/StreamDetailsPanel/StreamFieldsTable/SyncFieldCell/SyncFieldCell.tsx
```diff
@@ -0,0 +1,66 @@
...
+    <>
+      {!isDisabled && (
+        <CheckBox
+          checkboxSize="sm"
+          checked={isFieldSelected}
+          onChange={() => handleFieldToggle(field.path, !isFieldSelected)}
+        />
+      )}
+      {isDisabled && (
+        <Tooltip control={<CheckBox checkboxSize="sm" disabled checked={isFieldSelected} readOnly />}>
```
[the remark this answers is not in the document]
josephkmh: I think I understand, and I think I generally agree when there's a lot of JSX, but I'm not sure if it would be any clearer in this instance. This component's render function basically does one thing only: checks the boolean `isDisabled` and returns one or the other bit of JSX. If I split that into another variable, wouldn't that same logic be repeated in the variable? What would the benefit be?

josephkmh:
> After you delete a connection and go back to it (we call this readonly mode), you can still check and uncheck the checkboxes:

Thanks for catching this! I did not have `readonly` mode on my radar. Here's a fix for that: https://github.com/airbytehq/airbyte/pull/21058/commits/706163df018f8448082600cb47ca607db638b2a6

josephkmh:
> I tried just unselecting my primary key and saving and got an error. But the error is a little vague. It works for one stream, but what about multiple streams? One option is to invalidate the selected primary key immediately so that the user has to update the key right away (Probably can be addressed separately.)

This has been fixed here: https://github.com/airbytehq/airbyte/pull/21058/commits/427a3b697108ca293a98d40fc91d46ef1b39aac0

dizel852 (commented):
Overall - LGTM 👍 
But have a question: do we support stream selection functionality during connection creation?

Since I got an error when tried to create a connection with unselected field:
image

josephkmh:
@dizel852 thank you for pointing that out, it looks like a backend error that we're looking into 👍

josephkmh:
@dizel852 this has been fixed in https://github.com/airbytehq/airbyte/pull/21647 !

airbyte-webapp/src/components/connection/CatalogTree/next/StreamDetailsPanel/StreamFieldsTable/StreamFieldsTable.tsx
josephkmh: True, I will leave this though since I don't think it makes any functional difference

airbyte-webapp/src/components/connection/CatalogTree/next/StreamDetailsPanel/StreamFieldsTable/SyncFieldCell/SyncFieldCell.tsx
```diff
@@ -0,0 +1,76 @@
...
+  destinationSyncMode,
+}) => {
+  const { mode } = useConnectionFormService();
+  const isNestedField = SyncSchemaFieldObject.isNestedField(field);
+  const isCursor = checkIsCursor(field.path);
+  const isChildFieldCursor = checkIsChildFieldCursor(field.path);
+  const isPrimaryKey = checkIsPrimaryKey(field.path);
+  const isChildFieldPrimaryKey = checkIsChildFieldPrimaryKey(field.path);
+  const isDisabled =
+    mode === "readonly" ||
```
edmundito, line 41: I'm still able to click on the checkboxes. https://127.0.0.1:3001/workspaces/10c578a0-250f-4d2a-9cf9-192107956407/connections/89114a84-9537-4ce9-8e2e-78e2b24431d9/replication

The logic is a bit off here, but we only disable if there's a tooltip.
```
isDisabled = mode === 'readonly'
showTooltip = isDisabled && mode !=== 'readonly'

-> mode === 'readonly' && mode !=== 'readonly' -> false

edmundito (commented):
I did some testing and noticed something 👀

airbyte-webapp/src/components/connection/CatalogTree/next/StreamDetailsPanel/StreamFieldsTable/SyncFieldCell/SyncFieldCell.tsx
josephkmh: Good catch, addressed in https://github.com/airbytehq/airbyte/pull/21058/commits/22f08b4e09e9f1de5eaf18982ff46ccef044894b and https://github.com/airbytehq/airbyte/pull/21058/commits/b5769c841062ef3288959dbd67a0f627cecfe335

edmundito (approved):

edoang:
@josephkmh i'm trying to test this feature on the latest version (0.44.0) but i don't find the way to select columns. I'm using postgres src and dest. am i missing something?

josephkmh:
@edoang this is indeed only available in cloud currently. It will be coming to OSS, but we don't have a timeline for that just yet!

PR #21059, opened 2023-01-05
🪟🐛 Connector form: Fill in all default values on oneOf switch

flash1293:
Slack thread for context: https://airbytehq-team.slack.com/archives/C02TL38U5L7/p1672859912283969

This PR fixes the problem that default values of nested properties are not populated when switching between oneOf conditions.

For example:
* Go to big query destination
* Switch to GCS staging
* "Credential" should default to "HMAC", but defaults to empty

This is fixed by selectively running the existing recursive `setDefaultValues` logic when picking a new condition (respecting all properties that have been set already to preserve existing behavior).

The old logic did only update the "const" values on the highest level, ignoring all nested default values.


flash1293:
In review, just opened today

lmossman (approved):
Changes LGTM. Tested locally and the problem is fixed

PR #21073, opened 2023-01-05
Migrate airbyte-bootloader to Micronaut

### LINKED ISSUE #12392: Annotate Dependency Services Utilized by airbyte-bootloader

The  project makes use of the constructor pattern and manually managed lifecycles to create and use objects.  In order to make the code more dependency injection ready and to reduce the amount of work required to convert the services within the project to an application framework, the existing code base should be prepared for use in a dependency injection enabled container.  In addition to this preparation work, the conversion of these objects into dependency injection ready, container managed objects will improve the modularity of the code base.  Allowing the framework to manage the lifecycle of these objects also makes it possible to introduce interfaces, which will allow us to leverage the application framework to inject different implementations in different environments.

To prepare the project to leverage dependency, all objects that are currently created manually and provided to other objects should be identified and annotated as a bean.  This will be done using either the [JSR-330 annotations](http://javax-inject.github.io/javax-inject/api/javax/inject/package-summary.html) annotations.  This includes marking the class as a `@Singleton` so that the framework will identify it and manage it at runtime and by marking any required fields as injectable (e.g. `@Inject`).  This solution will NOT change or remove constructors – that will occur when converting applications to use a framework.

In addition, any classes that should be managed by the dependency injection framework but will need to be refactored due to how they are created (e.g. via static builder methods or other static utility methods) should be documented.  A follow-on ticket will be created to address these classes.

### More Details

* The GAV for the [JSR-330 annotations](http://javax-inject.github.io/javax-inject/api/javax/inject/package-summary.html) is `'javax.inject:javax.inject:1'`
* Constructors should **NOT** be removed as part of this work.  That will be done later.
* Conversion of configuration from the [Configs.java](https://github.com/airbytehq/airbyte/blob/master/airbyte-config/models/src/main/java/io/airbyte/config/Configs.java) to framework configuration will be done later.  Ignore the modification of the [Configs.java](https://github.com/airbytehq/airbyte/blob/master/airbyte-config/models/src/main/java/io/airbyte/config/Configs.java).

## Acceptance Criteria

- [ ] All objects utilized by the [airbyte-bootloader](https://github.com/airbytehq/airbyte/tree/master/airbyte-bootloader) service should be annotated as managed beans
- [ ] All members used by a managed bean should be annotated for injection
- [ ] Any beans that cannot be easily created via an empty constructor should be documented for follow-on work
- [ ] Project should build and run without any issues

### LINKED ISSUE #13270: Convert airbyte-bootloader to a Micronaut service

The introduction of an application framework reduces the amount of time engineers must spend on "glue code" that provides low business value.  In addition, the lack of an application framework leads to inconsistent implementations across services, does not prevent code duplication and requires the re-invention of industry-solved problems.

The `airbyte-bootloader` service should be migrated to a [Micronaut](https://micronaut.io/) application in order to take advantage of consistent features and implementations across all services in the Airbyte project.

## Acceptance Criteria

- [ ] Service is converted to a Micronaut service
- [ ] Service still functions as expected

jdpgrailsdev:
## What
* Convert airbyte-bootloader to use the Micronaut framework
* fix #13270 
* fix #12392

## How
* Convert airbyte-bootloader to a Micronaut application
* Move configuration to Micronaut property source
* Convert existing classes to singletons/use dependency injection
* Add missing secret persistence type to airbyte-workers

## Recommended reading order
1. `Application.java`
2. `Bootloader.java`
3. `application.yml`
4. `build.gradle`
5. `ProtocolVersionChecker.java`
6. `ApplyDefinitionsHelper.java`
7. `RemoteDefinitionsProvider.java`
8. `PostLoadExecutor.java`
9. `DefaultPostLoadExecutor.java`
10. `SecretPersistence.java`
11.  `LocalTestingSecretPersistence.java`
12. `SecretMigrator.java`
13. `docker-compose.yml`
14. `.env/.env.dev`

## Tests

* All unit tests pass
* Updated unit tests pass
* Application builds locally
* Application runs locally
* Application runs successfully as part of a local Docker deployment
* Tested creating and running a sync locally using database bootstraped by airbyte-bootloader

**N.B.:** ~This is a preview PR.  I still need to integrate this into airbyte-cloud to verify everything works.  Once I have done that, this PR will be moved from a draft to ready for review.  However, I don't expect it to change very much, so feel free to do a first pass review.~


.env
```diff
@@ -39,6 +39,10 @@ BASIC_AUTH_PASSWORD=password
 BASIC_AUTH_PROXY_TIMEOUT=600
 
 ### DATABASE ###
+# Airbyte Bootloader environment variables
+BOOTLOADER_MIGRATION_BASELINE_VERSION=0.29.0.001
+RUN_DATABASE_MIGRATION_ON_STARTUP=true
```
gosusnp, line 44: Any cases where we do not want to run migrations?

airbyte-bootloader/build.gradle
```diff
@@ -11,19 +26,42 @@ dependencies {
     implementation project(':airbyte-protocol:protocol-models')
     implementation project(':airbyte-persistence:job-persistence')
 
-    implementation libs.temporal.sdk
-    implementation libs.flyway.core
+    testAnnotationProcessor platform(libs.micronaut.bom)
+    testAnnotationProcessor libs.bundles.micronaut.test.annotation.processor
```
gosusnp, line 30: Out of curiosity, what happened to `libs.flyway.core`?

airbyte-bootloader/src/main/java/io/airbyte/bootloader/Bootloader.java
```diff
@@ -0,0 +1,291 @@
...
+  private final DatabaseInitializer jobsDatabaseInitializer;
+  private final DatabaseMigrator jobsDatabaseMigrator;
+  private final JobPersistence jobPersistence;
+  private Runnable postLoadExecution;
+  private final ProtocolVersionChecker protocolVersionChecker;
+  private final boolean runMigrationOnStartup;
+  private final SecretMigrator secretMigrator;
+
+  public Bootloader(
+                    @Value("${airbyte.protocol.auto-upgrade-connectors}") final boolean autoUpgradeConnectors,
```
gosusnp, line 58: `airbyte.protocol` looks off, this should be a platform settings, not protocol.

airbyte-bootloader/src/main/java/io/airbyte/bootloader/Bootloader.java
```diff
@@ -0,0 +1,291 @@
...
+                    final JobPersistence jobPersistence,
+                    final ProtocolVersionChecker protocolVersionChecker,
+                    @Value("${airbyte.run-migration-on-startup}") final boolean runMigrationOnStartup,
+                    final SecretMigrator secretMigrator) {
+    this(autoUpgradeConnectors, configRepository, configsDatabaseInitializer, configsDatabaseMigrator, currentAirbyteVersion,
+        definitionsProvider, featureFlags, jobsDatabaseInitializer, jobsDatabaseMigrator, jobPersistence, protocolVersionChecker,
+        runMigrationOnStartup, secretMigrator, null);
+  }
+
+  @VisibleForTesting
```
gosusnp, line 76: I think this needs to be public, the `postLoadExecution` is overridden in cloud.

airbyte-bootloader/src/main/java/io/airbyte/bootloader/Bootloader.java
```diff
@@ -0,0 +1,291 @@
...
+    this.secretMigrator = secretMigrator;
+    this.postLoadExecution = postLoadExecution;
+  }
+
+  /**
+   * Ensures additional initialization is performed after all dependency injection has completed.
+   */
+  @PostConstruct
+  public void afterInitialization() {
+    postLoadExecution = () -> {
```
gosusnp, line 113: This was initialized in the constructor because cloud had a different one. With the new `ApplyDefinitionsHelper` I think we should be close to have the different behaviors parameterized rather than injected.

airbyte-bootloader/src/main/resources/application.yml
```diff
@@ -0,0 +1,115 @@
...
+airbyte:
+  bootloader:
+    migration-baseline-version: ${BOOTLOADER_MIGRATION_BASELINE_VERSION}
+  flyway:
+    configs:
+      initialization-timeout-ms: ${CONFIGS_DATABASE_INITIALIZATION_TIMEOUT_MS:60000}
+    jobs:
+      initialization-timeout-ms: ${JOBS_DATABASE_INITIALIZATION_TIMEOUT_MS:60000}
+  protocol:
+    auto-upgrade-connectors: ${AUTO_UPGRADE_CONNECTORS_PROTOCOL:false}
```
gosusnp, line 16: I'd move this to the `airbyte.bootloader`, also, this flag might be a bit misleading currently.
Our default behavior is to auto upgrade connectors that are not used in a sync. Cloud has a post-load function override to be able to force upgrade connectors all connectors. When I see `auto-upgrade`, I do not know which behavior it ties to.
We may want to address this when/if we rationalize those behaviors.

airbyte-bootloader/src/main/resources/application.yml
```diff
@@ -0,0 +1,115 @@
...
+    sensitive: false
+  threaddump:
+    enabled: true
+    sensitive: true
+
+flyway:
+  enabled: true
+  datasources:
+    config:
+      enabled: false
```
gosusnp, line 89: why are `config` and `jobs` disabled here?

airbyte-bootloader/src/test/java/io/airbyte/bootloader/ProtocolVersionCheckerTest.java
```diff
@@ -199,13 +213,17 @@ void testValidateAllConnectorsAreUpgraded(final boolean supportAutoUpgrade) thro
         Map.entry(dest2, V1_0_0),
         Map.entry(dest3, V2_0_0)));
 
+    final ProtocolVersionChecker protocolVersionChecker =
+        new ProtocolVersionChecker(jobPersistence, expectedVersionRange, configRepository, Optional.of(definitionsProvider));
```
gosusnp, line 217: nit: I find targetVersionRange clearer because it is the input of the checker, not just the expected result.

airbyte-bootloader/src/main/java/io/airbyte/bootloader/Bootloader.java
```diff
@@ -0,0 +1,291 @@
...
+    this.protocolVersionChecker = protocolVersionChecker;
+    this.runMigrationOnStartup = runMigrationOnStartup;
+    this.secretMigrator = secretMigrator;
+    this.postLoadExecution = postLoadExecution;
+  }
+
+  /**
+   * Ensures additional initialization is performed after all dependency injection has completed.
+   */
+  @PostConstruct
```
gosusnp, line 111: I don't think we can have this as a postConstruct.
We have a few validation in the `.load` that may decide to abort the process. I understand that `PostConstruct` would run before the `.load` which means that we would always update connectors even if we decide to abort the bootloader.

gosusnp (changes requested):
Exciting to see a bootloader on micronaut, we're getting there!

I don't know if you tested the integration with cloud, but looking at some changes on the bootloader constructor, we may run into some issues.

.env
jdpgrailsdev: @gosusnp Don't know.  This env var exists for some reason, so I suspect it was added for a particular case.  We can obviously remove this in the future if we deem that it is not needed.

airbyte-bootloader/build.gradle
jdpgrailsdev: It got moved up in this block to ensure that we get the correct version.

airbyte-bootloader/src/main/java/io/airbyte/bootloader/Bootloader.java
jdpgrailsdev: In DI frameworks, it is not advisable to initialize code that depends on other singletons in a constructor, as those objects may also not be completely initialized yet.  Regardless, I think the plan will be to move this to a concrete class that implements `Runnable` and is its own singleton, so that cloud can just provide a replacement.

airbyte-bootloader/src/main/resources/application.yml
jdpgrailsdev: @gosusnp That controls whether or not Micronaut should automatically run the migrations itself.  We want to manually run them via our own code in the `Bootloader` class, so it is disabled to prevent conflict.  See https://micronaut-projects.github.io/micronaut-flyway/latest/guide/#configuration for more details.

airbyte-bootloader/src/main/java/io/airbyte/bootloader/Bootloader.java
jdpgrailsdev: We are only setting the variable to the lambda/runnable, not executing it in the post construct.  That being said, I think it makes sense to create a class that represents the runnable so we can treat it like a singleton and replace it more easily in cloud.

airbyte-bootloader/src/main/java/io/airbyte/bootloader/Bootloader.java
gosusnp: Yes, different implementation was what I was thinking. I think some recent changes took us halfway already since both behaviors now live in this repo. I guess the next question might be wether we want to keep the flexibility of a post load hook that can do anything.

airbyte-bootloader/src/main/java/io/airbyte/bootloader/Bootloader.java
gosusnp: Indeed, I read it as apply the function rather than set.

airbyte-bootloader/src/main/java/io/airbyte/bootloader/Bootloader.java
jdpgrailsdev: This has now been extracted to an interface/implementation that can be overridden in cloud.

benmoriceau (approved):

gosusnp (approved):

airbyte-config/init/src/main/java/io/airbyte/config/init/ApplyDefinitionsHelper.java
```diff
@@ -20,26 +22,21 @@
...
-  @Deprecated(forRemoval = true)
-  public ApplyDefinitionsHelper(final ConfigRepository configRepository, final DefinitionsProvider definitionsProvider) {
-    this.configRepository = configRepository;
-    this.definitionsProvider = definitionsProvider;
-    this.jobPersistence = null;
-  }
-
   public ApplyDefinitionsHelper(final ConfigRepository configRepository,
-                                final DefinitionsProvider definitionsProvider,
+                                final Optional<DefinitionsProvider> definitionsProviderOptional,
```
gosusnp, line 36: Just saw this afterwards, why do we have this `@Requires` and `Optional` distinction here? Shouldn't all three arguments always be required?

airbyte-config/init/src/main/java/io/airbyte/config/init/RemoteDefinitionsProvider.java
```diff
@@ -21,32 +27,34 @@
...
 import java.util.UUID;
 import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
 
 /**
  * This provider pulls the definitions from a remotely hosted catalog.
  */
-final public class RemoteDefinitionsProvider implements DefinitionsProvider {
+@Singleton
+@Primary
```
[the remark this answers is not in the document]
jdpgrailsdev: @gosusnp We could, but I don't think it really helps anything.  I suppose it would let us remove the `@Primary` annotation.

airbyte-config/init/src/main/java/io/airbyte/config/init/ApplyDefinitionsHelper.java
jdpgrailsdev: @gosusnp Based on the existing code, there is a case where no definition provider is configured.  I'm not sure why that is the case, but for this first pass, I am just trying to maintain the existing functionality.

airbyte-bootloader/src/main/resources/application.yml
```diff
@@ -0,0 +1,121 @@
...
+        prefix: ${VAULT_PREFIX:}
+        token: ${VAULT_AUTH_TOKEN:}
+  version: ${AIRBYTE_VERSION}
+
+datasources:
+  config:
+    connection-test-query: SELECT 1
+    connection-timeout: 30000
+    idle-timeout: 600000
+    initialization-fail-timeout: -1   # Disable fail fast checking to avoid issues due to other pods not being started in time
```
davinchia, line 46: +1 for the comment

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/split_secrets/SecretPersistence.java
```diff
@@ -18,6 +18,15 @@
 @SuppressWarnings("PMD.MissingOverride")
 public interface SecretPersistence extends ReadOnlySecretPersistence {
 
+  /**
+   * Performs any initialization prior to utilization of the persistence object. This exists to make
```
davinchia, line 22: +1

airbyte-config/init/build.gradle
```diff
@@ -3,14 +3,17 @@ plugins {
 }
 
 dependencies {
-    implementation 'commons-cli:commons-cli:1.4'
+    annotationProcessor libs.bundles.micronaut.annotation.processor
+    api libs.bundles.micronaut.annotation
 
+    implementation 'commons-cli:commons-cli:1.4'
```
davinchia, line 9: huh, didn't realise this wasn't already in a deps.toml.

airbyte-config/init/src/main/java/io/airbyte/config/init/RemoteDefinitionsProvider.java
```diff
@@ -21,32 +27,34 @@
...
 
-  public RemoteDefinitionsProvider(final URI remoteDefinitionCatalogUrl, final Duration timeout) throws InterruptedException, IOException {
-    this.remoteDefinitionCatalogUrl = remoteDefinitionCatalogUrl;
-    this.timeout = timeout;
-    // TODO remove this call once dependency injection framework manages object creation
-    initialize();
-  }
-
-  // TODO will be called automatically by the dependency injection framework on object creation
+  @PostConstruct
```
davinchia, line 57: Curious, what does this do?

airbyte-server/src/main/resources/application.yml
```diff
@@ -105,6 +108,7 @@ datasources:
     maximum-pool-size: 10
     minimum-idle: 0
     idle-timeout: 600000
+    initialization-fail-timeout: -1   # Disable fail fast checking to avoid issues due to other pods not being started in time
```
davinchia, line 111: Curious, was this also failing in the server? Or are we adding this just in case. Trying to figure out how this change related to the rest of the changeset.

airbyte-workers/src/main/java/io/airbyte/workers/config/SecretPersistenceBeanFactory.java
```diff
@@ -33,6 +34,8 @@ public class SecretPersistenceBeanFactory {
             pattern = "(?i)^(?!google_secret_manager).*")
   @Requires(property = "airbyte.secret.persistence",
             pattern = "(?i)^(?!vault).*")
+  @Requires(property = "airbyte.secret.persistence",
+            pattern = "(?i)^(?!aws_secret_manager).*")
```
davinchia, line 38: oh my first time seeing this, what does the pattern do here?

docker-compose.yaml
```diff
@@ -24,14 +24,10 @@ services:
     container_name: airbyte-bootloader
     environment:
       - AIRBYTE_VERSION=${VERSION}
-      - CONFIG_DATABASE_PASSWORD=${CONFIG_DATABASE_PASSWORD:-}
```
davinchia, line 27: why are we removing these?

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/split_secrets/LocalTestingSecretPersistence.java
```diff
@@ -15,15 +16,21 @@ public class LocalTestingSecretPersistence implements SecretPersistence {
...
 
+  private boolean initialized = false;
+
   public LocalTestingSecretPersistence(final Database configDatabase) {
     this.configDatabase = configDatabase;
+  }
 
-    Exceptions.toRuntime(() -> {
+  @Override
+  public void initialize() throws SQLException {
```
davinchia, line 26: I understand your previous comment on the test set up now

airbyte-bootloader/src/test/java/io/airbyte/bootloader/ProtocolVersionCheckerTest.java
```diff
@@ -292,10 +314,10 @@ void testValidateFailsOnProtocolRangeChangeWithoutDefinitionsProvider(final bool
   @ParameterizedTest
   @ValueSource(booleans = {true, false})
   void testValidateSucceedsWhenNoProtocolRangeChangeWithoutDefinitionsProvider(final boolean supportAutoUpgrade) throws Exception {
-    protocolVersionChecker = new ProtocolVersionChecker(jobPersistence, configs, configRepository, Optional.empty());
-
+    final AirbyteProtocolVersionRange expectedTargetVersionRange = new AirbyteProtocolVersionRange(V0_0_0, V2_0_0);
     setCurrentProtocolRangeRange(V0_0_0, V2_0_0);
-    setTargetProtocolRangeRange(V0_0_0, V2_0_0);
+    final ProtocolVersionChecker protocolVersionChecker =
```
davinchia, line 319: I appreciate you cleaning up the tests as you go

davinchia (approved):
Nice work!

I appreciate the detailed write up, the comments and the minor clean ups here and there.

airbyte-config/init/src/main/java/io/airbyte/config/init/RemoteDefinitionsProvider.java
jdpgrailsdev: @davinchia This tells the DI framework to run this method after all required singletons used by this singleton are created.  It's basically a hook to run some additional init logic.

airbyte-server/src/main/resources/application.yml
jdpgrailsdev: @davinchia Yes, but I think that our pod retry policy made it not obvious, as the pod would keep trying and eventually the db pod would be available.

airbyte-workers/src/main/java/io/airbyte/workers/config/SecretPersistenceBeanFactory.java
jdpgrailsdev: If the value does NOT equal the string (case insensitive).

airbyte-server/src/main/resources/application.yml
davinchia: got it, thanks

docker-compose.yaml
jdpgrailsdev: They are no longer used.  We reference the `DATABASE_X` env vars only.  We could reintroduce the separate ones if necessary.

davinchia: ah. if these are no longer used, we should also remove them from the EnvConfigs file.

although, maybe a good idea to wait on that since we haven't yet figure out how we want to track all configuration information globally. i.e. a master list of configuration that our micronaut application yamls accept.

davinchia: no need to action now. thinking out loud with you

airbyte-workers/src/main/java/io/airbyte/workers/config/SecretPersistenceBeanFactory.java
jdpgrailsdev: @davinchia I should add that once this is merged, I plan on doing a second pass to clean this up/consolidate this so that we don't have this in multiple places.

docker-compose.yaml
jdpgrailsdev: @davinchia Agreed.  Once everything is converted to Micronaut, there will be a pass to remove EnvConfigs entirely and clean up the environment variables.

airbyte-workers/src/main/java/io/airbyte/workers/config/SecretPersistenceBeanFactory.java
davinchia: sounds good. just from a reader pov, it's not immediately clear why we need the inverse match here.

docker-compose.yaml
davinchia: and I hope consolidate into one master list for reference :)

PR #20716, opened 2022-12-20
Update to Micronaut 3.8.0

jdpgrailsdev:
## What
* Stay up to date with latest framework release

## How
* Update to latest Micronaut release and corresponding dependencies
* ~https://github.com/micronaut-projects/micronaut-core/releases/tag/v3.7.5~
* https://micronaut.io/2022/12/27/micronaut-framework-3-8-0-released/

## Recommended reading order
1. `deps.toml`

PR #21077, opened 2023-01-05
Upgrade to Micronaut 3.8.0 and related dependencies

jdpgrailsdev:
## What
* Re-apply https://github.com/airbytehq/airbyte/pull/20716

## How
* Upgrade dependencies to latest Micronaut release and related dependencies

## Recommended reading order
1. `deps.toml`


deps.toml
```diff
@@ -95,7 +95,7 @@ platform-testcontainers-jdbc = { module = "org.testcontainers:jdbc", version.ref
 platform-testcontainers-postgresql = { module = "org.testcontainers:postgresql", version.ref = "platform-testcontainers" }
 postgresql = { module = "org.postgresql:postgresql", version.ref = "postgresql" }
 quartz-scheduler = { module = "org.quartz-scheduler:quartz", version = "2.3.2" }
-s3 = { module = "software.amazon.awssdk:s3", version = "2.16.84" }
+s3 = { module = "software.amazon.awssdk:s3", version = "2.19.4" }
```
benmoriceau, line 98: I guess that we will see with the acceptance test but we might change how we call this lib. THis was likely the cause of the issue previously.

gosusnp: The unknown host exception is the same error I was seeing when I was testing with `3.8.0`.
To clarify what I was saying earlier, I think the issues I saw when I tested`3.7.5` may have been fixed by the changes I merged yesterday.

jdpgrailsdev:
@davinchia After doing some investigation, I discovered that the issue is due to a change in how the AWS SDK generates the URL when requesting S3 objects.  From what I can tell, the client use to use the format `http://<hostname>/<bucket name>`.  Now, it uses the following format:  `<bucket name>.<hostname>`.  This collides with the default DNS generation for K8 services, specifically for our Minio S3 stand-in.  I have tweaked the configuration files to ensure that we can handle the new format and verified that all of the K8 acceptance tests now pass locally.  I also manually created a connection and ran a sync a few times in my local minikube to verify it works.  I could use another set of eyes on this approach to see if there is a better way to do it.

davinchia (approved):
Changeset looks good.

Should wait for the helm build to pass before merging in.

kube/resources/airbyte-minio.yaml
```diff
@@ -34,12 +34,16 @@ spec:
...
               value: "minio123"
+            - name: MINIO_DOMAIN
+              value: "airbyte-minio-svc"
           ports:
             - containerPort: 9000
           # Mount the volume into the pod
           volumeMounts:
             - name: storage # must match the volume name, above
               mountPath: "/storage"
+      hostname: airbyte-dev-logs
```
gosusnp, line 45: I think the other error I saw is tied to the config: `STATE_STORAGE_MINIO_BUCKET_NAME`. In our configs, we have two buckets, one for logs, one for states which should to be state of the orchestrator.
I did a quick search, it seems that both are set with the same value `airbyte-dev-logs`. As long as we do not have different values for minio, this should work.

gosusnp (approved):

kube/resources/airbyte-minio.yaml
jdpgrailsdev: Agreed that this is a risk/something that probably needs to be refactored if/when they diverge.

benmoriceau (approved):
LGTM, I wonder if we will need something similar in cloud to support their acceptance tests

jdpgrailsdev:
> LGTM, I wonder if we will need something similar in cloud to support their acceptance tests

@benmoriceau @gosusnp brought this up as well.  I am going to test the cloud acceptance tests locally to see if those tests are also impacted.

jdpgrailsdev:
@davinchia I have updated the PR to ensure that we use the older versions of the AWS SDK dependencies to avoid the behavior change in the URL used to access objects.  This should allow us to move forward while we try to figure out how best to re-configure the local Minio pod.

airbyte-container-orchestrator/build.gradle
```diff
@@ -9,6 +9,15 @@ dependencies {
     implementation platform(libs.micronaut.bom)
     implementation libs.bundles.micronaut
 
+    // Ensure that the versions defined in deps.toml are used
+    // instead of versions from transitive dependencies
+    implementation(libs.s3) {
```
davinchia, line 14: can we add comments in all the spots where we need to force s3 due to the break for context?

airbyte-server/build.gradle
```diff
@@ -7,6 +7,26 @@ configurations.all {
...
+    implementation platform(libs.micronaut.bom)
+    implementation libs.bundles.micronaut
+    implementation libs.micronaut.jaxrs.server
+
+    // Ensure that the versions defined in deps.toml are used
+    // instead of versions from transitive dependencies
+    implementation(libs.flyway.core) {
+        force = true
+    }
+    implementation(libs.s3) {
```
davinchia, line 23: same comment here

airbyte-workers/build.gradle
```diff
@@ -25,6 +25,12 @@ dependencies {
     implementation (libs.jooq) {
         force = true
     }
+    implementation(libs.s3) {
```
davinchia, line 28: and here

davinchia (approved):
Good to merge after adding comments!

Should we also create an issue to track this?

jdpgrailsdev:
> Good to merge after adding comments!
> 
> Should we also create an issue to track this?

Definitely.  I have saved the Kubernetes changes off to the side.  I'll create a ticket to track this.

airbyte-container-orchestrator/build.gradle
jdpgrailsdev: @davinchia An additional comment beyond the one listed that talks about avoiding updates from transitive dependencies?

davinchia: yes, that in this case the transitive dependencies from micronaut 3.8 and higher are incompatible with our current helm setup. We could also link to the created issue for better visibility.

PR #21085, opened 2023-01-05
Move hadoop-lzo to test dependency

grishick:
## What
destination-s3 does not have a direct compile dependency on hadoop-lzo library, so it can be removed


grishick:
destination-s3 run here: https://github.com/airbytehq/airbyte/actions/runs/3851924067/jobs/6563610583

cgardens (approved):
save us! approved once builds pass!

PR #21087, opened 2023-01-05
Improve Authentication Error handling 

### LINKED ISSUE #20092: Staging destinations (bigquery, etc): Handle S3 auth error as a ConfigException
e.g. https://sentry.io/organizations/airbytehq/issues/3804018246/?project=6527718&referrer=slack

This code probably lives in base-java-s3, so we can roll it out for all our staging destinations (bigquery, snowflake, redshift, ...). This should get wrapped into a ConfigException so that it doesn't trigger alerting.

I'm pretty sure we already detect this in `check`, but would be good to confirm that also.

example stacktrace:
```
com.amazonaws.services.s3.model.AmazonS3Exception: Forbidden (Service: Amazon S3; Status Code: 403; Error Code: 403 Forbidden; Request ID: null; S3 Extended Request ID: null; Proxy: null), S3 Extended Request ID: null
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.handleErrorResponse(AmazonHttpClient.java:1819)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.handleServiceErrorResponse(AmazonHttpClient.java:1403)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeOneRequest(AmazonHttpClient.java:1372)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeHelper(AmazonHttpClient.java:1145)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.doExecute(AmazonHttpClient.java:802)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeWithTimer(AmazonHttpClient.java:770)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.execute(AmazonHttpClient.java:744)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.access$500(AmazonHttpClient.java:704)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutionBuilderImpl.execute(AmazonHttpClient.java:686)
	at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:550)
	at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:530)
	at com.amazonaws.services.s3.AmazonS3Client.invoke(AmazonS3Client.java:5437)
	at com.amazonaws.services.s3.AmazonS3Client.invoke(AmazonS3Client.java:5384)
	at com.amazonaws.services.s3.AmazonS3Client.getObjectMetadata(AmazonS3Client.java:1367)
	at com.amazonaws.services.s3.AmazonS3Client.getObjectMetadata(AmazonS3Client.java:1341)
	at com.amazonaws.services.s3.AmazonS3Client.doesObjectExist(AmazonS3Client.java:1422)
	at io.airbyte.integrations.destination.s3.S3StorageOperations.createBucketObjectIfNotExists(S3StorageOperations.java:110)
	at io.airbyte.integrations.destination.bigquery.BigQueryGcsOperations.createStageIfNotExists(BigQueryGcsOperations.java:102)
	at io.airbyte.integrations.destination.bigquery.BigQueryStagingConsumerFactory.lambda$onStartFunction$3(BigQueryStagingConsumerFactory.java:105)
	at io.airbyte.commons.concurrency.VoidCallable.call(VoidCallable.java:15)
	at io.airbyte.integrations.destination.buffered_stream_consumer.BufferedStreamConsumer.startTracked(BufferedStreamConsumer.java:118)
	at io.airbyte.integrations.base.FailureTrackingAirbyteMessageConsumer.start(FailureTrackingAirbyteMessageConsumer.java:34)
	at io.airbyte.integrations.base.IntegrationRunner.consumeWriteStream(IntegrationRunner.java:233)
	at io.airbyte.integrations.base.IntegrationRunner.lambda$runConsumer$2(IntegrationRunner.java:241)
	at io.airbyte.integrations.base.IntegrationRunner.watchForOrphanThreads(IntegrationRunner.java:270)
	at io.airbyte.integrations.base.IntegrationRunner.runConsumer(IntegrationRunner.java:240)
	at io.airbyte.integrations.base.IntegrationRunner.runInternal(IntegrationRunner.java:150)
	at io.airbyte.integrations.base.IntegrationRunner.run(IntegrationRunner.java:100)
	at io.airbyte.integrations.destination.bigquery.BigQueryDestination.main(BigQueryDestination.java:327)
```

jbfbell:
## What
From #20092, an authentication error which happens not in the check method can trigger our alerting. This change is to catch Authentication errors in S3 operations and repackage these as ConfigExceptions

## How
For Blob storage operations and things which use the S3 base, specifically look for authentication errors and rethrow these as config exceptions.

## 🚨 User Impact 🚨
User's will now see a config exception if their credentials do not have


airbyte-integrations/connectors/destination-redshift/src/main/java/io/airbyte/integrations/destination/redshift/RedshiftStagingS3Destination.java
```diff
@@ -60,44 +61,49 @@ private boolean isEphemeralKeysAndPurgingStagingData(final JsonNode config, fina
...
+      } finally {
+        try {
+          DataSourceFactory.close(dataSource);
+        } catch (final Exception e) {
+          LOGGER.warn("Unable to close data source.", e);
+        }
       }
+    } catch (final Exception e) {
+      LOGGER.error("Check failed.", e);
+      throw new ConfigErrorException(e.getMessage() != null ? e.getMessage() : e.toString());
```
edgao, line 106: what does the error message look like? (just want to check that it's something a user could reasonably understand + figure out how to fix)

airbyte-integrations/connectors/destination-redshift/src/main/java/io/airbyte/integrations/destination/redshift/RedshiftStagingS3Destination.java
```diff
@@ -60,44 +61,49 @@ private boolean isEphemeralKeysAndPurgingStagingData(final JsonNode config, fina
...
+            .withStatus(AirbyteConnectionStatus.Status.FAILED)
+            .withMessage("Could not connect with provided configuration. \n" + e.getMessage());
+      } finally {
+        try {
+          DataSourceFactory.close(dataSource);
+        } catch (final Exception e) {
+          LOGGER.warn("Unable to close data source.", e);
+        }
       }
+    } catch (final Exception e) {
```
edgao, line 104: @ryankfu does this match the semantics of a config error? Do we need to catch a more specific exception type?

my understanding is that we should only throw a config error if it's user error. I.e. if this is some sort of weird network thing (e.g. we messed up our GCP firewall rules and actually can't connect to S3) then we shouldn't wrap in a config error

* assuming we want to catch a more specific error type -> can we put this try-catch somewhere in base-java-s3? Maybe in S3BaseChecks or something. That would allow us to solve all S3-based destinations in one shot.
* or if we're ok catching generic Exception -> Can we just do that in IntegrationRunner? (https://github.com/airbytehq/airbyte/blob/3c52168f80cdc14cf346b4cadf5660fff0ad4936/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java#L123)

airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/BigQueryDestinationTest.java
```diff
@@ -81,7 +81,7 @@ class BigQueryDestinationTest {
   protected static final Path CREDENTIALS_WITH_GCS_STAGING_PATH =
       Path.of("secrets/credentials-gcs-staging.json");
 
-  protected static final Path[] ALL_PATHS = {CREDENTIALS_WITH_GCS_STAGING_PATH, CREDENTIALS_BAD_PROJECT_PATH, CREDENTIALS_NO_DATASET_CREATION_PATH,
+  protected static final Path[] ALL_PATHS = {CREDENTIALS_STANDARD_INSERT_PATH, CREDENTIALS_BAD_PROJECT_PATH, CREDENTIALS_NO_DATASET_CREATION_PATH,
```
edgao, line 84: nice catch 🤦

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/S3StorageOperations.java
```diff
@@ -95,23 +95,15 @@ public String getBucketObjectPath(final String namespace, final String streamNam
...
-  public void createBucketObjectIfNotExists(final String objectPath) {
+  public void createBucketIfNotExists() {
     final String bucket = s3Config.getBucketName();
-    final String folderPath = objectPath.endsWith("/") ? objectPath : objectPath + "/";
     if (!doesBucketExist(bucket)) {
       LOGGER.info("Bucket {} does not exist; creating...", bucket);
       s3Client.createBucket(bucket);
       LOGGER.info("Bucket {} has been created.", bucket);
     }
-    if (!s3Client.doesObjectExist(bucket, folderPath)) {
```
edgao, line 110: not sure I understand why we're removing this - do we not need to verify the ability to create a new s3 object?

edgao (commented):
some generic comments:
* sign the CLA ;) (I should have mentioned that on your previous PR)
* is there some way we can unit test this? (e.g. inject a mocked s3 client and verify that `check` throws a ConfigError)

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/S3StorageOperations.java
jbfbell: IIUC - S3 (and GCS) have a simulated folder hierarchy, i.e. while are some helpers in the client tooling that simulate listing objects under a certain "directory", these tools are essentially just looking at a string prefix. The removed lines below are creating a zero content object at the `folderPath` which is unnecessary because the a parent path doesn't need to exist before objects are placed since there are no real folders/directories. After removing those lines the `objectPath` parameter is no longer required because its purpose was to create the unnecessary empty object at `folderPath` and the spirit of the method changes to ensuring the bucket exists.

edgao: ah got it, that makes sense.

also it looks like we check object create+delete in a separate method https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/S3BaseChecks.java#L36

docs/integrations/destinations/redshift.md
```diff
@@ -178,5 +178,3 @@ Each stream will be output into its own raw table in Redshift. Each table will c
 | 0.3.13  | 2021-09-02 | [5745](https://github.com/airbytehq/airbyte/pull/5745)     | Disable STATUPDATE flag when using S3 staging to speed up performance                                                                                                                                            |
 | 0.3.12  | 2021-07-21 | [3555](https://github.com/airbytehq/airbyte/pull/3555)     | Enable partial checkpointing for halfway syncs                                                                                                                                                                   |
 | 0.3.11  | 2021-07-20 | [4874](https://github.com/airbytehq/airbyte/pull/4874)     | allow `additionalProperties` in connector spec                                                                                                                                                                   |
-
```
jbfbell, line 181: Honestly not sure how this happened, I have a blank new line at the end of my file and it says its up to date.

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/S3StorageOperations.java
```diff
@@ -138,7 +132,13 @@ public String uploadRecordsToBucket(final SerializableBuffer recordsData,
         exceptionsThrown.add(e);
       }
     }
-    throw new RuntimeException(String.format("Exceptions thrown while uploading records into storage: %s", Strings.join(exceptionsThrown, "\n")));
+    final boolean areAllExceptionsAuthExceptions = exceptionsThrown.stream().filter(e -> e instanceof AmazonS3Exception)
```
ryankfu, line 135: What's the rationale behind checking that all exceptions are authentication issues?

Similar to `BigQueryGcsOperations` can this Auth Error Code be a constant so anybody not sure how come `403` can easily understand the value behind the integer?

airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/BigQueryGcsOperations.java
```diff
@@ -84,7 +87,15 @@ public String getStagingFullPath(final String datasetId, final String stream) {
   public void createSchemaIfNotExists(final String datasetId, final String datasetLocation) {
     if (!existingSchemas.contains(datasetId)) {
       LOGGER.info("Creating dataset {}", datasetId);
-      BigQueryUtils.getOrCreateDataset(bigQuery, datasetId, datasetLocation);
+      try {
+        BigQueryUtils.getOrCreateDataset(bigQuery, datasetId, datasetLocation);
+      } catch (BigQueryException bqe) {
```
ryankfu, line 92: Super nit: since we're only catching one exception type, it doesn't seem necessary to have an Exception value that differs from the rest of the code 

```java
BigQueryException e
```

That said, it's not blocking. It's mostly to keep consistency across our code base.

Also in the super class of `BigQueryException`, `BaseServiceException` there's a method `getReason()` that might be more usage to return back to the user for how come something caused this exception

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/S3StorageOperations.java
jbfbell: +1 on making it a constant

Rationale for checking that all errors are auth errors is that theoretically there could be a transient issue with permissions thats resolved on a retry. All issues being auth issues was the behavior I observed when permissions were not set correctly and it seemed safer to make a judgement call on this being a configuration issue when this occurs since matching `any` as opposed to `all` might hide a different issue

ryankfu: Got it, that makes sense. I would add that in as a comment if you can since I can see someone looking at this code and coming with a different conclusion and just checking if `any` of the exceptions are Auth and calling it a day

Further more in the `BigQueryException` class they have a section that goes like
```java
  private static final Set<Error> RETRYABLE_ERRORS =
      ImmutableSet.of(
          new Error(500, null), new Error(502, null), new Error(503, null), new Error(504, null));
```

Might be worth considering naming the errors as `AUTH_ERRORS` and placing them into some general utility file like `JdbcUtils.java` so they don't need to be recreated across each destination connector

airbyte-integrations/connectors/destination-redshift/src/main/java/io/airbyte/integrations/destination/redshift/RedshiftStagingS3Destination.java
ryankfu: I didn't catch this tag before the changes were outdated but we don't need to catch a specific exception type so as long as the general exception type is directly associated with user-related errors

As you highlighted, firewall or transient issues _should not_ be captured as `ConfigErrorException`s

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/S3StorageOperations.java
grishick, line 135: As far as I can see, this line checks if all exceptions are instance of `AmazonS3Exception`. Wouldn't this also include non-authentication related exceptions?

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/S3StorageOperations.java
```diff
@@ -130,7 +132,13 @@ public String uploadRecordsToBucket(final SerializableBuffer recordsData,
...
```
edgao, line 135: if I'm reading this correctly, this currently checks that all `AmazonS3Exception` are auth errors, but ignores any non-amazon exceptions. E.g. if there's a random `IllegalStateException` in this list, it would be treated as though it's an auth error

I think it might be better to bundle the filter + map together?
```java
.map(e -> {
  if (e instanceof AmazonS3Exception s3e) {
    // ...
  } else {
    // assume non-amazon exceptions are unrelated to authentication
    return false;
  }
})
```

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/S3StorageOperations.java
```diff
@@ -130,7 +132,13 @@ public String uploadRecordsToBucket(final SerializableBuffer recordsData,
...
+        .map(s3e -> ((AmazonS3Exception) s3e).getStatusCode()).allMatch(code -> Integer.valueOf(403).equals(code));
```
edgao, line 136: also formatting nit, I'd prefer to have the `allMatch` on a newline to make this easier to read, i.e.
```suggestion
        .map(s3e -> ((AmazonS3Exception) s3e).getStatusCode())
        .allMatch(code -> Integer.valueOf(403).equals(code));
```

(I find it hard to keep track of which parentheses surround which expressions)

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/S3StorageOperations.java
jbfbell: It took me awhile but I think I see what you and @grishick are saying - updating this to ensure that really all of the errors are auth errors

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/S3StorageOperations.java
jbfbell: +1 on the multiline styling, I wasn't sure what the preference here was so I let the autoformatter decide

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/S3StorageOperations.java
jbfbell: updated this to ensure that when the list is filtered it verifies that the length matches the original length

grishick (approved):

edgao (approved):
:shipit:

ryankfu (approved):
LGTM

jbfbell:
/publish connector=connectors/destination-snowflake,connectors/destination-bigquery,connectors/destination-bigquery-denormalized,connectors/destination-gcs,connectors/destination-s3,connectors/destination-redshift

jbfbell:
/publish connector=connectors/destination-snowflake,connectors/destination-bigquery,connectors/destination-bigquery-denormalized,connectors/destination-gcs,connectors/destination-s3,connectors/destination-redshift

jbfbell:
/publish connector=connectors/destination-snowflake,connectors/destination-bigquery,connectors/destination-bigquery-denormalized,connectors/destination-gcs,connectors/destination-s3,connectors/destination-redshift

jbfbell:
/publish connector=connectors/destination-snowflake,connectors/destination-bigquery,connectors/destination-bigquery-denormalized,connectors/destination-gcs,connectors/destination-s3,connectors/destination-redshift parallel=true

grishick:
@jbfbell it looks like you have already published destination-bigquery in this operation: https://github.com/airbytehq/airbyte/pull/21087#issuecomment-1397500613 and the publish command has updated seed files successfully, so all subsequent attempts to (re)publish `destination-bigquery` will fail.

jbfbell:
/publish connector=connectors/destination-bigquery-denormalized

Issue #17426, opened 2022-09-30
Do not attempt to parse JSON messages that are too darn big


evantahler:
_Inspired by a [comment in this document](https://docs.google.com/document/d/1z7ngxncrX44inDnGnBQJQvkCKew_4MQb2Fr_slxzGoI) by @jdpgrailsdev, this PR potentially addresses https://github.com/airbytehq/oncall/issues/525 and https://github.com/airbytehq/airbyte/issues/16031 (and maybe https://github.com/airbytehq/airbyte/issues/12532)._

In our replication activity, very bad things happen if the Worker or Replication pod crash - we loose job state, we loose metrics, and we have no idea why.  One theory of why we might be having OOM crashes in the replication worker is that we are trying to parse STATE or RECORD messages that are "very large"™️.  We can prevent that type of OOM by halting the replication activity if we encounter a record that is "very large"™️.

Todo:
* In our [parsing method](https://github.com/airbytehq/airbyte/blob/1e2f701fe71e78361ae1f6750885bf79fd013fb8/airbyte-commons/src/main/java/io/airbyte/commons/json/Jsons.java#L72-L86), we should first check the byte length of the record we are trying to parse - we already have [a method for that](https://github.com/airbytehq/airbyte/blob/1e2f701fe71e78361ae1f6750885bf79fd013fb8/airbyte-commons/src/main/java/io/airbyte/commons/json/Jsons.java#L129-L131)!  If the size is larger than ~60% of the container's available RAM (or, perhaps, 1GB), don't try to parse it.
  * How could we pass the container's total RAM to the worker process - is there a system call or ENV variable for this?
* Fail the replication process (attributable to the source) if this happens - **do not** log the big record - that's PII and would make the logs huge.

evantahler:
An assumption to test is that the try/catch blocks around JSON deserialization don't help us if the failure is OOM.

PR #21090, opened 2023-01-06
Avoid parsing a message if this message is too big

benmoriceau:
## What
Address: https://github.com/airbytehq/airbyte/issues/17426

In order to avoid OOM when trying to parse a record, we need to check that the size of the messages received by the orchestrator are not too big. If they are, the replication is failed by throwing an `IllegalStateException`.


airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
```diff
@@ -283,7 +283,14 @@ static void watchForOrphanThreads(final Procedure runMethod,
...
    */
   @VisibleForTesting
-  static void consumeMessage(final AirbyteMessageConsumer consumer, final String inputString) throws Exception {
+  static void consumeMessage(final AirbyteMessageConsumer consumer, final String inputString, final long runnerMaximumMemoryInBytes)
+      throws Exception {
+
+    if (inputString.getBytes().length > (runnerMaximumMemoryInBytes * 0.6)) {
+      LOGGER.error("One messge is too big, the replication will fail in order to avoid OOM. The size of the message is: {} bytes",
+          inputString.getBytes().length);
+      throw new IllegalStateException("Invalid message, the message is too big");
```
evantahler, line 292: I think this error will eventually bubble up to the user/UI. If that's true, perhaps we should provide a little more context to try and learn as much as we can about the message:

```
Airbyte has received a message from the source(!) at 2022-01-01 01:02:03(!) which is larger than xxxMB(!).  The sync has been failed to prevent running out of memory.
```

Everything with (!) is dynamic.  Ideally, I'd like to use a regular expression or similar to grab the stream name, but we can add that later if this becomes a problem.

evantahler (approved):
Happy to see this getting addressed! 

I added some suggestions, but I don't think they are blocking.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
evantahler: Will this exception properly be attributed to the source and build a FailureReason?

benmoriceau: I'll check that.

benmoriceau: During my test, it turns out that adding that here will only be apply on the destination. I'll rework it and test the error message and make sure that the getMaxMemory is only call once.

benmoriceau: @evantahler , I made some change (it's a draft because I need to clean it up, I'll re-request a review when its ready). In my test it looked like 
image

evantahler: Nice!

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteStreamFactory.java
```diff
@@ -62,6 +86,23 @@ public Stream<AirbyteMessage> create(final BufferedReader bufferedReader) {
     return bufferedReader
         .lines()
         .peek(str -> metricClient.distribution(OssMetricsRegistry.JSON_STRING_LENGTH, str.length()))
+        .peek(str -> {
+          if (exceptionClass.isPresent()) {
+            long messageSize = str.getBytes(StandardCharsets.UTF_8).length;
+            if (messageSize > maxMemory * 0.8) {
```
gosusnp, line 92: Would be nice to have some comments about this magic `0.8`

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteStreamFactory.java
```diff
@@ -62,6 +86,23 @@ public Stream<AirbyteMessage> create(final BufferedReader bufferedReader) {
...
+              try {
+                String errorMessage = String.format(
```
gosusnp, line 94: nit: `final` and for `messageSize` as well

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteStreamFactory.java
```diff
@@ -62,6 +86,23 @@ public Stream<AirbyteMessage> create(final BufferedReader bufferedReader) {
     return bufferedReader
         .lines()
         .peek(str -> metricClient.distribution(OssMetricsRegistry.JSON_STRING_LENGTH, str.length()))
+        .peek(str -> {
+          if (exceptionClass.isPresent()) {
+            long messageSize = str.getBytes(StandardCharsets.UTF_8).length;
```
gosusnp, line 91: Why are we using a different method here to get the length than line 88?

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteStreamFactory.java
```diff
@@ -38,21 +44,39 @@ public class DefaultAirbyteStreamFactory implements AirbyteStreamFactory {
...
   public DefaultAirbyteStreamFactory(final MdcScope.Builder containerLogMdcBuilder) {
-    this(new AirbyteProtocolPredicate(), LOGGER, containerLogMdcBuilder);
+    this(new AirbyteProtocolPredicate(), LOGGER, containerLogMdcBuilder, Optional.empty());
   }
 
   DefaultAirbyteStreamFactory(final AirbyteProtocolPredicate protocolPredicate,
                               final Logger logger,
-                              final MdcScope.Builder containerLogMdcBuilder) {
+                              final MdcScope.Builder containerLogMdcBuilder,
+                              final Optional<Class<? extends RuntimeException>> exceptionClass) {
```
gosusnp, line 61: Can we add some comments about why we have this `Optional<RuntimeException>`?
Also, this API feels a bit unusual to use as it requires the client to tell what exception to throw. Would throwing a specific exception and let the client code handle the exception wrapping if needed work?

gosusnp (commented):
Approach looks good to me, left some comments/questions.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteStreamFactory.java
benmoriceau: I agree, I was questioning that as well. The original ticket was mentioning a 60% ratio. I put 80% out of safety, I wonder if it could make a connection to fail even if it's able to process.

evantahler: I don't think that there is one yet, but we modify a source's `resourceRequirements` and give it a ton of ram.  Parsing the message would work on the source, but not in the worker in that case.  Not that it's likely, but we know that we've encountered users who store binary data (images) in their database, and the source turns them into base64 strings.  That's the kind of message-parsing crash we are trying to avoid with this change.

benmoriceau: Ok, that PR should prevent that but should we push for the connectors implement that as limitation as well. In this PR a `source error` means that the platform think it won't be able to parse a message from the source. If we want to plug that error, it needs to be directly in the source read method that is call in the IntegrationRunner [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java#L138)

evantahler: Agree - but there will always be differences in implementation or OSS sources which don't have this limit in place.

gosusnp: Should we consider having a hard limit on the message size?

gosusnp (approved):

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteStreamFactory.java
```diff
@@ -34,25 +40,49 @@
...
   DefaultAirbyteStreamFactory(final AirbyteProtocolPredicate protocolPredicate,
                               final Logger logger,
-                              final MdcScope.Builder containerLogMdcBuilder) {
+                              final MdcScope.Builder containerLogMdcBuilder,
+                              final Optional<Class<? extends RuntimeException>> messageSizeExceptionClass) {
     protocolValidator = protocolPredicate;
     this.logger = logger;
     this.containerLogMdcBuilder = containerLogMdcBuilder;
+    this.exceptionClass = messageSizeExceptionClass;
+    this.maxMemory = Runtime.getRuntime().maxMemory();
```
grishick, line 72: AFAIK `maxMemory()` returns the total amount of memory reserved for the VM (including the unallocated memory). Judging by the rest of the code, there is an assumption here that 80% of that will be available for local variables - how is this assumption made and validated? 
I might suggest that we use `freeMemory()` instead

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteStreamFactory.java
```diff
@@ -61,7 +91,24 @@ public Stream<AirbyteMessage> create(final BufferedReader bufferedReader) {
     final var metricClient = MetricClientFactory.getMetricClient();
     return bufferedReader
         .lines()
-        .peek(str -> metricClient.distribution(OssMetricsRegistry.JSON_STRING_LENGTH, str.length()))
+        .peek(str -> metricClient.distribution(OssMetricsRegistry.JSON_STRING_LENGTH, str.getBytes(StandardCharsets.UTF_8).length))
+        .peek(str -> {
+          if (exceptionClass.isPresent()) {
+            final long messageSize = str.getBytes(StandardCharsets.UTF_8).length;
```
grishick, line 97: at this point, isn't `str` already occupying as many as `str.getBytes(StandardCharsets.UTF_8).length` bytes in memory?

PR #21091, opened 2023-01-06
add feature-flag client

### LINKED ISSUE #20830: create a feature-flag client
Add a new module that adds a feature-flag client that supports multiple environments.

colesnodgrass:
## What
- fix #20830 
- add an `airbyte-featureflag` module
- add a Feature-Flag client; supports
  - local feature-flags (via config file) 
  - remote feature-flags (currently via LaunchDarkly)
  - env-var feature-flags (intended to only be supported during the migration phase)
- introduces `Kotlin` 1.8.0
- Feature-Flag Client is not actually used as part of this PR

## How
See the [details](#details) section below

## Todo
- [x] finalize naming of classes
- [x] show how to use this (working on this as a separate PR)
- [ ] figure out where `flags.yml` should exist

## Recommended reading order
1. `Client.kt`
2. `Flags.kt`
3. `Context.kt`
4. `Factory.kt`

## Details
There is a lot going on in this PR, I'm going to attempt to break it down into bite-size descriptive chunks below for ease of understanding.

#### Client
The goal of this feature-flag client is to support both cloud-hosted and non-cloud-hosted configurations.  

This has been accomplished by defining a single interface
```kotlin
/**
 * Feature-Flag Client interface.
 */
sealed interface FeatureFlagClient {
  /**
   * Returns true if the flag with the provided context should be enabled. Returns false otherwise.
   */
  fun enabled(flag: Flag, ctx: Context): Boolean
}
```
with multiple implementations; `ConfigFileClient` and `LaunchDarklyClient`.  
```mermaid
graph TD
    I(FeatureFlagClient) --> P(ConfigFileClient)
    I --> C(LaunchDarklyClient)
    C --> L(LaunchDarkly)
    P --> Y(flags.yml)
    P -.-> E(Env Vars)
    C -.-> E
    style E stroke-dasharray: 3
```
To know the current state of a feature-flag, provide a `Flag`object  and a `Context` object (more information about these concepts in their sections below) to the `enabled` method. The returned boolean indicates the current status of the `Flag` with the `Context`.

In order for this `Client` to determine that status of the requested `Flag` and `Context`, the `ConfigFileClient` relies on a `flags.yml` configuration file while the `LaunchDarklyClient` relies on an external service (LaunchDarkly).

As this repo already has some notion of [environment based feature-flags](https://github.com/airbytehq/airbyte/blob/9f930a0de76b947adda10c0971dc686f5409182a/airbyte-commons/src/main/java/io/airbyte/commons/features/EnvVariableFeatureFlags.java), all implementations of the `FeatureFlagClient` will initially support flags managed by environment-variables.  The near-term goal is to replace all environment-variables flags with `flags.yml` entries and LaunchDarkly flags.

This PR contains a [Micronaut @Factory](https://github.com/airbytehq/airbyte/blob/9f930a0de76b947adda10c0971dc686f5409182a/airbyte-featureflag/src/main/kotlin/config/Factory.kt) for automatically determining which `FeatureFlagClient` should be used. If the application configuration `airbyte.feature-flag.path` is defined, the `ConfigFileClient` will be utilized.  If `airbyte.feature-flag.api-key` is defined, then `LaunchDarklyClient` will be utilized. 

>**Note**
> What should happen if _both_ fields are defined? What about if _neither_ are defined?
> Are these both error states?

New `Client` types cannot be defined outside of the `airbyte-featureflag` module as the `Client` type is `sealed` with no `open` implementations.

#### Flags
A `Flag` is defined as
```kotlin
/**
 * Flag is a sealed class that all feature-flags must inherit from.
 *
 * @param [key] is the globally unique identifier for identifying this specific feature-flag.
 * @param [default] is the default value of the flag.
 * @param [team] is the team that is responsible for the feature-flag, defaults to Unknown.
 */
sealed class Flag(
  internal val key: String,
  internal val default: Boolean = false,
  internal val team: Team = Team.UNKNOWN,
)
```
Where the `key` is the unique identifier of the flag (must match the name defined in both the `flags.yml` and LaunchDarkly).  The `default` value is the value returned if the flag's state cannot be found or determined.  The `team` is the owning team responsible for managing this feature-flag and only exists for ownership and tracking reasons.

There are currently three types of feature flags supported by this client; `Permanent`, `Temporary`, and `EnvVar`.

`Permanent` flags are those which will always exist and are not necessarily part of a _feature_.  Presently there is only a single permanent flag which is _currently_ controlled by the env-var `LOG_CONNECTOR_MESSAGES` (this should migrate away from being env-var controlled soon after this PR is merged).  Permanent flags should extend the `Flag` type directly
```kotlin
object LogConnectorMessages : Flag(key = "LOG_CONNECTOR_MESSAGES", team = Team.PLATFORM_MOVE)
```

`Temporary` flags will be the majority of feature flags, as these flags will only exist for a finite period of time, being removed entirely once a feature has been completely and successfully released.  Temporary flags should extend the `Temporary` type
```kotlin
object AutoDetectSchema : Temporary(key = "AUTO_DETECT_SCHEMA", team = Team.PLATFORM_MOVE)
```
There is no real difference between a `Permanent` and `Temporary` flag other than an indication of the lifecycle intent of the flag itself.

`EnvVar` flags should only exist for a small period of time as the current env-var based feature-flag solution is replaced with this more well defined and managed solution.  More bluntly the `EnvVar` class itself will be removed in the near future.  `EnvVar` flag statuses are determined entirely by the environment-variable they are watching and cannot be changed without restarting the services.  EnvVar flags must extend the `EnvVar` type
```kotlin
object NeedStateValidation : EnvVar(envVar = "NEED_STATE_VALIDATION")

Ideally all feature flags would be defined within the `Flags` file, but this will be be enforced (at least initially) for `Temporary` and `EnvVar` flags.  This is why both of those classes are defined as `open`.  Any `Permanent` flag that extends the `Flag` type _must_ be [declared within the `airbyte-featureflag` module](https://kotlinlang.org/docs/sealed-classes.html#location-of-direct-subclasses) as the `Flag` type is `sealed`.
```

#### Contexts
A `Context` adds additional _context_ to the current request of the status of a feature-flag.  For example, the user or workspace that is making the request.  As feature-flags may only be enabled for a subset of users or workspaces, this additional information is necessary in order to determine the correct state of the feature-flag being requested. 

A `Context` is defined as 
```kotlin
sealed interface Context {
  val kind: String
  val key: String
}
```
There are currently two implementations; `User` and `Workspace`.  `User` represents a user as identified by their unique-id and `Workspace` represents a workspace as identified by its unique-id (and optionally a user-id as well). 

>**Note** 
> I'm not sure if supporting a  `User` context is necessary, as discussions around feature-flags tend to focus on handling them at the Workspace level.

New `Context` types cannot be defined outside of the `airbyte-featureflag` module as the `Context` type is `sealed` with no `open` implementations.

#### flags.yml
The configuration file for determining the status of a feature-flag for non-cloud deployments must be in the format of
```yaml
flags:
  - name: feature-flag-example-one
    enabled: true
  - name: feature-flag-example-two
    enabled: false
```

Currently flags defined in the `flags.yml` do not respect `Contexts`.  This may be added at a later date.

### Kotlin
This PR introduces Kotlin to the platform, specifically [Kotlin 1.8.0](https://kotlinlang.org/docs/whatsnew18.html).  There has been multiple discussions regarding adding Kotlin into our code base and this new feature-flag client seemed like the opportune time.

#### Kotlin Features Utilized
There are a number of Kotlin specific features that are utilized within this PR.  As this is the first Kotlin PR I'm going to briefly touch on a few of them.  If there are any further questions, please reach out to me.

* Multiple Types Defined Within a Single File
   From [Kotlin's coding conventions](https://kotlinlang.org/docs/coding-conventions.html#source-file-organization)
   ```
   Placing multiple declarations (classes, top-level functions or properties) in the same Kotlin source 
   file is encouraged as long as these declarations are closely related to each other semantically, 
   and the file size remains reasonable (not exceeding a few hundred lines)
   ```
   This is utilized in nearly every file added in this PR.  Technically Java does allow this, but only [for sealed classes](https://github.com/airbytehq/airbyte/blob/045323d61e8a0c918d6350ba2031625b251967f9/airbyte-metrics/reporter/src/main/java/io/airbyte/metrics/reporter/Emitter.java), which we are already utilizing.

* Data Classes
  Data classes are similar to Java's record type, but with more flexibility. There are only a few [data](https://github.com/airbytehq/airbyte/blob/f2cec1891942c4f032fa85d2bb111d54c6e8bc8f/airbyte-featureflag/src/main/kotlin/Context.kt#L23-L43) [classes](https://github.com/airbytehq/airbyte/blob/0838a3f689484e1ca57cd6318b33d79064b34b4f/airbyte-featureflag/src/main/kotlin/Client.kt#L124-L129) in this PR. 

* init Blocks
   A primary constructor in Kotlin cannot contain any code, [init blocks](https://github.com/airbytehq/airbyte/blob/0838a3f689484e1ca57cd6318b33d79064b34b4f/airbyte-featureflag/src/main/kotlin/Client.kt#L56-L65) are utilized for these specific purposes.

* Internal Visibility
   Kotlin supports [different visibility modifiers](https://kotlinlang.org/docs/visibility-modifiers.html) than Java and this PR takes advantage of them; specifically `private` and `internal`.  Where `private` is only visible within the file its defined in (like these [data classes](https://github.com/airbytehq/airbyte/blob/0838a3f689484e1ca57cd6318b33d79064b34b4f/airbyte-featureflag/src/main/kotlin/Client.kt#L114-L129) used for parsing the config file) and `internal` is visible to the entire module (like this [primary constructor](https://github.com/airbytehq/airbyte/blob/0838a3f689484e1ca57cd6318b33d79064b34b4f/airbyte-featureflag/src/main/kotlin/Flags.kt#L74-L79) which can only be called from within the `airbyte-featureflag` module).

* Default Parameters, Named Parameters, and @JvmOverloads
   Kotlin supports both defaults parameters and named parameters. Both of which reduce the number of methods necessary to support optional parameters or for ensuring that method parameters are passed in a specific order. 
   For example, the following Kotlin code
   ```kotlin
   open class Temporary @JvmOverloads constructor(
     key: String,
     default: Boolean = false,
     team: Team = Team.UNKNOWN,
   ) : Flag(key = key, default = default, team = team)
   ```
   can be called from other Kotlin code via `Temporary("key")` or `Temporary(default=true, key="key")`. However, as Java does not support default or named parameters, this code has also been annotated with `@JvmOverloads` to auto generate the appropriate methods for Java callers
   ```java
   Temporary(String key, boolean default, Team team)
   Temporary(String key, boolean default)   
   Temporary(String key)
   ```

* Object Types
   Kotlin has explicit support for [singleton types](https://kotlinlang.org/docs/object-declarations.html#object-declarations-overview), called `object`s.  These are utilized for defining the [feature-flags](https://github.com/airbytehq/airbyte/blob/0838a3f689484e1ca57cd6318b33d79064b34b4f/airbyte-featureflag/src/main/kotlin/Flags.kt#L12-L17).  When referenced within Kotlin code, they are referenced simply by their object name.  However, as Java doesn't have this support natively, any `object`s referenced from Java code will need to use the object's auto-generated `.INSTANCE`; i.e. `LogConnectorMessages.INSTANCE` 
   >**Note**
   > This will become more apparent in the related _how to use this client_ PR.

* Extension Functions
   Kotlin supports [extension functions](https://kotlinlang.org/docs/extensions.html) which are utilized a few times in this PR. One adds a [Path.onChange](https://github.com/airbytehq/airbyte/blob/0838a3f689484e1ca57cd6318b33d79064b34b4f/airbyte-featureflag/src/main/kotlin/Client.kt#L153) function that executes a lambda anytime that Java Path is modified (for watching the `flags.yml` file for changes). The other two add [helper methods](https://github.com/airbytehq/airbyte/blob/0838a3f689484e1ca57cd6318b33d79064b34b4f/airbyte-featureflag/src/main/kotlin/Client.kt#L177-L186) for converting the `Context` type to LaunchDarkly specific contexts.  All three of these extension functions are defined with a `private` visibility and therefore can only be utilized by code within the `Client` file.

* thread Block
   Kotlin has a lot of builtin helpers for dealing with common Java patterns. One [used in this PR](https://github.com/airbytehq/airbyte/blob/0838a3f689484e1ca57cd6318b33d79064b34b4f/airbyte-featureflag/src/main/kotlin/Client.kt#L158-L166) is a `thread` block that creates a background thread for monitoring file changes.


colesnodgrass:
I'm planning on moving this out of draft status tomorrow, but feel free to review it before that time as I don't plan on making any big changes.

airbyte-featureflag/src/main/kotlin/Context.kt
```diff
@@ -0,0 +1,43 @@
...
+ * I'm still playing around with this.  Basically the idea is to define our own custom context types
+ * (by implementing this sealed interface) to ensure that we are consistently using the same identifiers
+ * throughout the code.
+ *
+ * @property [kind] determines the kind of context the implementation is,
+ * must be consistent for each type and should not be set by the caller of a context
+ * @property [key] is the unique identifier for the specific context, e.g. a user-id or workspace-id
+ */
+sealed interface Context {
+    val kind: String
```
gosusnp, line 19: `kind` feels a bit redundant with the actual `data class`. The overhead isn't big, I was wondering if you considered not having it?

airbyte-featureflag/src/main/kotlin/Flags.kt
```diff
@@ -0,0 +1,104 @@
...
+ * There are two types of feature-flags; permanent and temporary. Permanent flags should inherit from the Flag class directly
+ * while temporary flags should inherit from the Temporary class (which it itself inherits from the Flag class).
+ *
+ * @param [key] is the globally unique identifier for identifying this specific feature-flag.
+ * @param [default] is the default value of the flag.
+ * @param [team] is the team that is responsible for the feature-flag, defaults to Unknown.
+ */
+sealed class Flag(
+    internal val key: String,
+    internal val default: Boolean = false,
```
gosusnp, line 42: That's probably fine for now but would it make sense to consider different value types:
```
sealed class Flag<T>(internal val default: T)
```

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+import kotlin.concurrent.thread
+import kotlin.concurrent.write
+import kotlin.io.path.isRegularFile
+
+/**
+ * Open Questions:
+ * 1. Can the context and flag be combined into one class?
+ *   - i.e. Does it make sense to one a feature-flag apply to two different context types?
+ * 2. Should the default value be merged into the Flag type instead of specified on each check?
+ *   - Or maybe still allow it to be specified, but make it an optional parameter
```
cgardens, line 29: what are we trying to achieve with the optional? it seems like whoever declares the flag is well situated to decide what the default is. it's not obvious to me why the person checking it would get any say in what the default is (even optionally).

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+ * Data wrapper around OSS feature-flag configuration file.
+ *
+ * The file has the format of:
+ * flags:
+ *  - name: feature-one
+ *    enabled: true
+ *  - name: feature-two
+ *    enabled: false
+ */
+private data class PlatformFlags(val flags: List<PlatformFlag>)
```
gosusnp, line 124: I don't think there's much shared code across the different implemtations, I'd be somehow leaning towards having one file per implementation and its privates for clarity.

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+    .let { it.flags }
+    .associateBy { it.name }
+
+/**
+ * Monitors a [Path] for changes, calling [block] when a change is detected.
+ *
+ * @receiver Path
+ * @param [block] function called anytime a change is detected on this [Path]
+ */
+private fun Path.onChange(block: () -> Unit) {
```
gosusnp, line 153: I thought it was some fancy kotlin builtin when I was reading the usage above ;)

airbyte-featureflag/src/main/kotlin/Context.kt
cgardens, line 19: why make this key a string as opposed to an enum? it seems like that's how it's functionally being used anyway.

airbyte-featureflag/src/main/kotlin/Client.kt
cgardens, line 153: If i understand right, this overrides behavior on the `Path` object. What is the scope of that override? Is it scoped just to this file?

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+/**
+ * Cloud feature-flag client. Feature-flags are derived from an external source (the LDClient).
+ * Also supports flags defined via environment-variables via the [EnvVar] class.
+ *
+ * @param [client] the Launch-Darkly client for interfacing with Launch-Darkly.
+ */
+class CloudClient(private val client: LDClient) : FeatureFlagClient {
+    override fun enabled(flag: Flag, ctx: Context): Boolean {
+        return when (flag) {
+            is EnvVar -> flag.enabled()
```
cgardens, line 84: it's a little confusing to me why the caller of `enabled` knows if the flag is if an EnvFlag or not. i assumed the point of the abstraction was that the caller has a "flag" (roughly a string) and wants to know if it's turn on or not, but doesn't need to know where that flag is set. passing in `EnvFlag` breaks that abstraction. how are you thinking about it.

cgardens (commented):
very cool! kotlin is here! he finally couldn't resist.

airbyte-featureflag/src/main/kotlin/Context.kt
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * I'm still playing around with this.  Basically the idea is to define our own custom context types
+ * (by implementing this sealed interface) to ensure that we are consistently using the same identifiers
+ * throughout the code.
+ *
+ * @property [kind] determines the kind of context the implementation is,
+ * must be consistent for each type and should not be set by the caller of a context
+ * @property [key] is the unique identifier for the specific context, e.g. a user-id or workspace-id
+ */
+sealed interface Context {
```
gosusnp, line 18: I am wondering if this implementation of `Context` is too restrictive.
When I am thinking of feature flags, eventually, it would be nice to be able to define a general context (User, Workspace, Connection, Source, Destination), and then in the feature flag definition, be able to say if user matches X, or if source is Y etc. Looking at this, it feels like context is currently only one dimension at a time.

Or maybe, adjusting the Client interface to work with a list of context could be a future extension.

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+    /** lock is used for ensuring access to the flags map is handled correctly when the map is being updated. */
+    private val lock = ReentrantReadWriteLock()
+
+    init {
+        if (!config.isRegularFile()) {
+            throw IllegalArgumentException("config must reference a file")
+        }
+
+        config.onChange {
+            val tempFlags = readConfig(config)
```
gosusnp, line 62: This will probably fall under the code style section, but especially when defining variables from a function call, I tend to prefer explicit typing.

airbyte-featureflag/src/main/kotlin/Flags.kt
```diff
@@ -0,0 +1,104 @@
...
+ * Intended only to be used in a transitory manner as the platform migrates to an official feature-flag solution.
+ * Every instance of this class should be migrated over to the Temporary class.
+ *
+ * @param [team] the team that controls this flag
+ * @param [envVar] the environment variable to check for the status of this flag
+ * @param [default] the default value of this flag, if the environment variable is not defined
+ * @param [fetcher] the function used to retrieve the environment-variable, overrideable for testing purposes only
+ * @constructor an internal constructor for testing purposes, users must be the public constructor
+ */
+open class EnvVar internal constructor(
```
gosusnp, line 74: This looked a bit confusing to me at first, thinking of env vars as an implementation of the `Client` rather than a subclass of `Flag` felt easier to comprehend.
Even though it might not make sense most of the time, I am wondering if we should still keep an env var feature flags implementation around. If it were to stay, I'd advocate for a `Client` implementation over `Flag`.

gosusnp (commented):
I like the general approach, left some questions/suggestions that are more details.

For a first review of kotlin code, it felt pretty natural.
Looking forward to see how it integrates in the other parts of the codebase.

airbyte-featureflag/src/main/kotlin/Client.kt
colesnodgrass: This comment was my open thoughts on an earlier implementation of this code and is no longer applicable.   

The LaunchDarkly SDK has you provide the default response _every time_ you call to check the status of the flag, and my first (rough) approach followed this same design.  It's not a good design, as it's possible that two sections in the code checking the same feature-flag would have two different defaults (if that flag no longer existed or couldn't be found), which is not a situation that should be supported.  I merged this into the Flag definition and it cannot be overwritten outside of the flag definition.

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+}
+
+/**
+ * Platform feature-flag client. Feature-flag are derived from a yaml config file.
+ * Also supports flags defined via environment-variables via the [EnvVar] class.
+ *
+ * @param [config] the location of the yaml config file that contains the feature-flag definitions.
+ * The [config] will be watched for changes and the internal representation of the [config] will be updated to match.
+ */
+class PlatformClient(config: Path) : FeatureFlagClient {
```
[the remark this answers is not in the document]
colesnodgrass: Seems logical, one of my `todo` items listed in this PR is to `finalize naming of classes`

airbyte-featureflag/src/main/kotlin/Context.kt
colesnodgrass: It's a string primarily for two reasons:
1.  LaunchDarkly requires a string to be provided in [their v6 api](https://github.com/airbytehq/airbyte/blob/9cb32d3381a12ba55f58379d4ddeeaccd2343984/airbyte-featureflag/src/main/kotlin/Client.kt#L187), so making it an Enum adds one more layer of abstraction without really giving us anything in return
2. As this is a `sealed interface` the`kind` value can only be set by the implementations of this interface within the `airbyte-featureflag` module (that is to say, no implementation of this interface can be defined outside of this module), so there is no risk of these values being used anywhere else

They could be an enum, but I question what value that would add.

airbyte-featureflag/src/main/kotlin/Client.kt
colesnodgrass: This doesn't override existing behavior on the `Path` object, it adds behavior to it.  Really it's just syntactical-sugar that Kotlin provides which roughly equates to the following java code:
```java
private static void onChange(final Path path, final Supplier<Void> block) {
  // ...
}
```

As this added behavior (aka extension function) is defined as `private`, it can only be called from within this file.  If it was defined as `internal` then it could be called from anywhere within the `airbyte-featureflag` module, and if defined as `public` (which is the default visibility if not defined) then it could be called from anywhere that this module was included as a dependency.

airbyte-featureflag/src/main/kotlin/Client.kt
colesnodgrass: > it's a little confusing to me why the caller of enabled knows if the flag is if an EnvFlag or not

The caller shouldn't care if the flag they're using is an `EnvVar` flag or not.  The idea is that all flags are defined as singletons in a central location and the caller would simply reference the specific flags by their singleton, ignoring what type they actually are.

E.g. we have the following flag currently defined
```kotlin
object LogConnectorMessages : EnvVar(envVar = "LOG_CONNECTOR_MESSAGES", team = Team.PLATFORM_MOVE)
```
which currently is an `EnvVar` type flag, but the user of this flag doesn't know that when providing it to the `enabled` method, they simply call
``` kotlin
val client: FeatureFlagClient = /* whatever implementation they're using */
client.enabled(LogConnectorMessages, WorkspaceContext("workspace-id")
```
As these flags are migrated away from environment-variables, the definition will be updated to 
```kotlin
object LogConnectorMessages : Flag(key = "LOG_CONNECTOR_MESSAGES", team = Team.PLATFORM_MOVE)
```
and the caller will be none the wiser.

The `FeatureFlagClient` implementations themselves need to be aware if the Flag is of type `EnvVar` as environment-variables based flags will never be found in the `flags.yml` file or in the LaunchDarkly API.  The client needs to know that these flags are special cases and require special handling.  This is why the `EnvVar.enabled()` method is defined as `internal`, it isn't visible to any code outside of this module.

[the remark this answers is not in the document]
colesnodgrass: The `EnvVar` flags will ideally be a temporary stop gap on the migration to a config-file/LaunchDarkly and I feel that the added complexity of a `CompositeClient` probably isn't worth it.  Additionally during the migration to using new feature-flags there will be times when both the environment-variable and the externally-controlled feature-flag would both be defined, but we only would want one source of truth for these and probably on a flag-by-flag basis.

Originally I wasn't planning on having `EnvVar` flag support in all the clients, but ultimately realized it was necessary.  It might be worth looking into changing the `sealed interface FeatureFlagClient` into a `sealed class FeatureFlagClient` with a method on it that could dedupe some of this.

In my ideal world, the `EnvVar` code is removed within weeks of us moving over to LaunchDarkly.

airbyte-featureflag/src/main/kotlin/Context.kt
colesnodgrass: Originally I didn't have it, for this very reason, it's redundant. 

The problem I ran into is that LaunchDarkly wants a string passed to their [ContextKind.of](https://javadoc.io/doc/com.launchdarkly/launchdarkly-java-server-sdk/latest/com/launchdarkly/sdk/ContextKind.html) method (a requirement for their v6 sdk) which ended up with the code having to check which type of `Context` was provided and then assigning that value to a string to pass to `ContextKind.of`
```kotlin
private fun Context.toLDContext(): LDContext {
  val kind = when(this) {
    is Workspace -> "workspace"
    is User -> "user"
  }
  val builder = LDContext.builder(ContextKind.of(kind), key)
  // ...
}
```

I found it cleaner to place this unique-string-identifer into the `data class` itself.

airbyte-featureflag/src/main/kotlin/Flags.kt
colesnodgrass: Yes.  

There are different potentials for what a feature-flag could return.  A boolean, an int, a double, a string.  Initially I'm only supporting boolean values as that is all I see the need for.  

Adding additional value types would also change the `FeatureFlagClient` interface to include more than the single `enabled(/*...*/): Boolean)` method.

I went with the simplest solution for now.

airbyte-featureflag/src/main/kotlin/Context.kt
colesnodgrass: It should be working this way already (with the addition of the new `Context` types you called out).

It's up to the evaluator of the feature-flag itself to determine _given this feature-flag and this context, is this feature-flag enabled or not_, which is exactly what LaunchDarkly does.  

In LaunchDarkly you would configure _this_ feature-flag for _that_ User-Context to return true.   

For the `flags.yml` based client we would need to build the context support ourselves, and I'm hoping we can avoid that for as long as possible.

airbyte-featureflag/src/main/kotlin/Flags.kt
colesnodgrass: The reason to not have an `EnvVarClient` is that it limits you to _only_ have feature-flags defined as environment-variables.  Which is not what we want, we want to migrate to LaunchDarkly (or `flag.yml` for local).  Defining this at the `Flag` level lets the `FeatureFlagClient` handle both the deprecated (env-var) flags and the new (externally-managed, dynamically evaluated) flags within the same client.  

There still could be a `EnvVarClient` but I think that just forced the complexity into  the user's hands as they would have to manage multiple `FeatureFlagClient`s within the code base and would need to be aware of what feature-flag is defined where and which client should be used.

For now these flags are defined as singletons within the code and can change from being environment-variable based, to config-file based, to LaunchDarkly based, to whatever-is-next based without the user of the `FeatureFlagClient` being aware or having to make any code changes (they would have to pull in a newer `airbyte-featureflag` depedency).

airbyte-featureflag/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+
+dependencies {
+    annotationProcessor(platform(libs.micronaut.bom))
+    annotationProcessor(libs.bundles.micronaut.annotation.processor)
+
+    implementation(platform(libs.micronaut.bom))
+    implementation(libs.micronaut.inject)
+    implementation("com.jayway.jsonpath:json-path:2.7.0")
+    implementation("com.launchdarkly:launchdarkly-java-server-sdk:6.0.1")
+    implementation("com.fasterxml.jackson.core:jackson-databind:2.14.0")
```
jdpgrailsdev, line 17: We should add these in to the deps.toml, as there is already a managed jackson version in there.  This will prevent us getting out of sync.

airbyte-featureflag/src/main/kotlin/Flags.kt
```diff
@@ -0,0 +1,104 @@
...
+object StreamCapableState : EnvVar(envVar = "USE_STREAM_CAPABLE_STATE")
+object AutoDetectSchema : EnvVar(envVar = "AUTO_DETECT_SCHEMA", team = Team.PLATFORM_MOVE)
+object NeedStateValidation : EnvVar(envVar = "NEED_STATE_VALIDATION")
+object ApplyFieldSelection : EnvVar(envVar = "APPLY_FIELD_SELECTION")
+
+/**
+ * The team that created and is responsible for individual feature-flags.
+ * TODO: finalize the options here!
+ */
+enum class Team {
```
jdpgrailsdev, line 23: I know I was the one who suggested this, but I wonder if this should be more of an "area" or "focus" than team, simply because we may end up with different concerns that are owned by the same team (thinking out loud here).  Ultimately, we are trying to achieve some sort of categorization or identification to make it easy for those reading the code to know who to talk to about what a flag is being used for and/or if it can be removed.  I don't hold this too strongly -- I just wonder that this won't mean anything to open source users, as they know nothing about our internal team structure.

airbyte-featureflag/src/test/kotlin/ClientTest.kt
```diff
@@ -0,0 +1,229 @@
+/*
+ * Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+ */
+
+import com.launchdarkly.sdk.LDUser
+import com.launchdarkly.sdk.server.LDClient
+import io.airbyte.featureflag.*
+import io.mockk.*
```
jdpgrailsdev, line 8: Nit:  avoid * imports.

airbyte-featureflag/src/main/kotlin/Client.kt
cgardens: Thanks! Makes sense.

airbyte-featureflag/src/main/kotlin/Client.kt
benmoriceau: +1

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+            throw IllegalArgumentException("config must reference a file")
+        }
+
+        config.onChange {
+            val tempFlags = readConfig(config)
+            lock.write { flags = tempFlags }
+        }
+    }
+
+    override fun enabled(flag: Flag, ctx: Context): Boolean {
```
benmoriceau, line 67: Should we add another function to return a string or another value type. We currently have a string feature flag: https://github.com/airbytehq/airbyte/blob/master/airbyte-commons/src/main/java/io/airbyte/commons/features/FeatureFlags.java#L38 Maybe it should have been move the the EnvConfig though

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+        return when (flag) {
+            is EnvVar -> flag.enabled()
+            else -> lock.read { flags[flag.key]?.enabled ?: flag.default }
```
benmoriceau, line 70: After your presentation the ?/?: is clear but I think that it would worth doing a presentation to the team to introduce this to the entire engineer team

cgardens:
@colesnodgrass for Kotlin, what's the guidance you would give on how to scope what code goes in a single file?

Vanilla Java is very restrictive--one top-level class per file. There are certainly down sides there, but it also nudges developers in a decent direction of not yeeting tons of code into one file. It's a nudge, not a silver bullet--obviously people still write Java classes that are too big (and we have a few in our code base).

In a previous life, I worked in a large production Ruby code base. One of the major code quality issues was gigantic files. I _think_ the source of this is that over time, in a language like Ruby, when you want to add additional functionality, it's so easy to just add another function at the bottom of the file. Java counteracts that a little because if the new functionality is a new class, you can't easily just add it to an existing file. I know nothing I'm saying here is 100% enforced, but I think language features that nudge developers in the right direction do a lot.

Kotlin gives us flexibility in terms of what constructs can go in one file. That flexibility is definitely great in some places, but also could be a foot gun in others. Especially since our team is not very experienced with this language, I would be curious what rule of thumb, convention, or even static analysis that you might suggest to prevent misuse.

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+    .associateBy { it.name }
+
+/**
+ * Monitors a [Path] for changes, calling [block] when a change is detected.
+ *
+ * @receiver Path
+ * @param [block] function called anytime a change is detected on this [Path]
+ */
+private fun Path.onChange(block: () -> Unit) {
+    val watcher = fileSystem.newWatchService()
```
benmoriceau, line 154: Could we make the type to be explicit when it val comes from a method return, it makes it easier to understand when reviewing. IMO:
val string = "str" is ok 
but
val methodReturn = test()
should be
val methodReturn: Type = test()

This comes from the experience I have with scala, non explicit types can become a pain.

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+/** The yaml mapper is used for reading the feature-flag configuration file. */
+private val yamlMapper = ObjectMapper(YAMLFactory()).registerKotlinModule()
+
+/**
+ * Reads a yaml configuration file, converting it into a map of flag name to flag configuration.
+ *
+ * @param [path] to yaml config file
+ * @return map of feature-flag name to feature-flag config
+ */
+//private fun readConfig(path: Path): Map<String, OSSFlag> = yamlMapper.readValue<List<OSSFlag>>(path.toFile())
```
benmoriceau, line 141: Nit: we should probably remove the comments here

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+private fun Path.onChange(block: () -> Unit) {
+    val watcher = fileSystem.newWatchService()
+    // The watcher services requires a directory to be provided, hence the `parent` reference here as `this` path should be a file.
+    parent.register(watcher, StandardWatchEventKinds.ENTRY_MODIFY, StandardWatchEventKinds.ENTRY_CREATE)
+
+    thread(isDaemon = true, name = "feature-flag-watcher", priority = MIN_PRIORITY) {
+        val key = watcher.take()
+        // The context on the poll-events for ENTRY_MODIFY events should not be null, this is a sanity check more than anything else
+        key.pollEvents().mapNotNull { it.context() as? Path }
+            .filter { this.endsWith(it) }
```
benmoriceau, line 162: What is this filter doing, I don't understand the mix of `it` and `this`. Also in the same spirit of making the type explicit, should we avoid using `it` when we can have  a more explicit name?

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+ */
+private fun Path.onChange(block: () -> Unit) {
+    val watcher = fileSystem.newWatchService()
+    // The watcher services requires a directory to be provided, hence the `parent` reference here as `this` path should be a file.
+    parent.register(watcher, StandardWatchEventKinds.ENTRY_MODIFY, StandardWatchEventKinds.ENTRY_CREATE)
+
+    thread(isDaemon = true, name = "feature-flag-watcher", priority = MIN_PRIORITY) {
+        val key = watcher.take()
+        // The context on the poll-events for ENTRY_MODIFY events should not be null, this is a sanity check more than anything else
+        key.pollEvents().mapNotNull { it.context() as? Path }
```
benmoriceau, line 161: Nit/QQ about kotlin, is the `?` needed in the `as?` since we know that the map only map the non null value.
Also do we need to filter out null values or is it included in the `mapNonNull` function?

airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -0,0 +1,196 @@
...
+    val builder = LDContext.builder(ContextKind.of(kind), key)
+    when (this) {
+        is Workspace -> {
+            user?.let { builder.set("user", it) }
+        }
+
+        is User -> Unit
+    }
+    return builder.build()
+}
```
benmoriceau, line 196: Nit missing new line

airbyte-featureflag/src/main/kotlin/Context.kt
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/**
+ * Context for representing a user.
+ *
+ * @param [key] the unique identifying value of this uer
+ */
+data class User(override val key: String) : Context {
+    override val kind = "user"
+}
```
benmoriceau, line 43: Nit: newline

airbyte-featureflag/src/main/kotlin/Context.kt
```diff
@@ -0,0 +1,43 @@
...
+) : Context {
+    override val kind = "workspace"
+}
+
+/**
+ * Context for representing a user.
+ *
+ * @param [key] the unique identifying value of this uer
+ */
+data class User(override val key: String) : Context {
```
benmoriceau, line 41: Should we add a context for connection as well

colesnodgrass:
@cgardens Kotlin publishes their own [coding conventions](https://kotlinlang.org/docs/coding-conventions.html) and mentions this specifically under the [source file organization](https://kotlinlang.org/docs/coding-conventions.html#source-file-organization) section:
> Placing multiple declarations (classes, top-level functions or properties) in the same Kotlin source file is encouraged as long as these declarations are closely related to each other semantically, and the file size remains reasonable (not exceeding a few hundred lines).

And then an additional paragraph specific to [extension functions](https://kotlinlang.org/docs/extensions.html)
> In particular, when defining extension functions for a class which are relevant for all clients of this class, put them in the same file with the class itself. When defining extension functions that make sense only for a specific client, put them next to the code of that client. Avoid creating files just to hold all extensions of some class.

I think this convention provides good guidance; are the types closely related and meager in size.  I would also consider visibility in this as well; does this type need to be accessed by code outside of these classes/module?

airbyte-featureflag/src/main/kotlin/Client.kt
colesnodgrass: Good call, I missed this during my research.  I think it makes sense to add in a string version as well, given that we already appear to be using it.

airbyte-featureflag/src/main/kotlin/Client.kt
colesnodgrass: Kotlin allows this, but I would prefer to have a discussion about these prior pain-points. Both you and Jimmy have brought this up in this PR (and also asked this same question when I gave the presentation months ago).   

Myself, and my previous teams, having written in kotlin, go, and java (using `var`) without running into issues related to not explicitly including the type.  I'm not against adding it, but I want to learn _why_ and _what_ problem that solves.

In this PR I only included the type (only in tests) when casting a class to the interface it implements:
```kotlin
var client: FeatureFlagClient = PlatformClient(/*...*/)
```

I'll schedule some time next week for us to discuss this.

airbyte-featureflag/src/main/kotlin/Client.kt
colesnodgrass: > is the ? needed in the as?

As Kotlin requires null values to be explicitly marked as nullable, this also applies when casting one class to another.
If `context()` returned a null (which it can according to the [javadocs](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/WatchEvent.html#context())) then the following code would result in an exception when `context()` was `null`
```kotlin
context() as Path
```

Kotlin as a ["Safe" (nullable) cast operator](https://kotlinlang.org/docs/typecasts.html#safe-nullable-cast-operator) (rolls right off the tonguq) for handling this scenario.  If `context()` returns `null`, the result will be `null` instead of an exception. If `context()` is not-null then it will be cast to a `Path` (which could also throw an exception if the type isn't a `Path`, but according to the javadocs, it should be a `Path`).

> do we need to filter out null values or is it included in the mapNonNull function

The [mapNotNull(https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map-not-null.html) filters out null values. I had originally written this as two separate methods (one for the map and one for filtering out nulls) and intellij told me to collapse them into this one method call.

airbyte-featureflag/src/main/kotlin/Context.kt
colesnodgrass: I think so.  Jimmy called out some additional Contexts that should also be added.

airbyte-featureflag/src/main/kotlin/Client.kt
colesnodgrass: I'll add a comment to this code to make it more explicit what is going on here, as I agree you can't look at this code and know what is happening.

In Kotlin, lambdas have an [implicit lambda parameter](https://kotlinlang.org/docs/lambdas.html#it-implicit-name-of-a-single-parameter) `it` that is scoped to the lambda body.  You can provide a name to the parameter (`name ->`) but in my experience it just adds more bloat to the code with questionable benefit. Most Kotlin code uses the provided `it` parameter. 

There are situations where a named lambda parameter makes sense, such as nested lambdas, or a really long lambda block (which should be avoided).

---

The `this` in this section of code is referring to the `Path` object itself as this function is an [extension function](https://kotlinlang.org/docs/extensions.html).
```kotlin
private fun Path.onChange(block: () -> Unit) { /* ... */ }
```
This adds an `onChange` function to the `Path` object, but if this function needs to reference the `Path` object itself, it does so via `this`.  The `this.endsWith` is calling `Path.endsWith` as `this` is a `Path` type. 

Extension functions are Kotlin syntactical-sugar that essentially equate to the following java code
```
private static void onChange(final Path this, final Supplier<Void> block) { /* ... */ }
```

airbyte-featureflag/build.gradle.kts
colesnodgrass: 👍🏻 I was already planning on doing this

airbyte-featureflag/src/test/kotlin/ClientTest.kt
colesnodgrass: #21136 
I updated our `code style` docs to set this for Kotlin correctly

airbyte-featureflag/src/main/kotlin/Client.kt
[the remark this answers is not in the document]
colesnodgrass: I renamed the first two as requested but left the `TestClient` alone, as it should really only be used in unit-test situations.

airbyte-featureflag/src/main/kotlin/Flags.kt
gosusnp: This is indeed very internal, at the same time, the current feature flagging approach is not really targeted towards open source users. Usage of those flags may change quickly and I don't think we plan on offering any guarantees in terms API stability.

airbyte-featureflag/src/main/kotlin/Client.kt
colesnodgrass: @benmoriceau I dug into this a little further and I don't think we need to support string feature-flags yet. The `FIELD_SELECTION_WORKSPACES` contains a comma-delimited list of workspace identifiers that should have field selection enabled.  This maps to the `Workspace` Context concept that this PR introduces.

Currently the env-var is defined similar to `FIELD_SELECTION_WORKSPACES: "uuid-0000, uuid-0001"`
which is used to determine which workspaces should have this functionality enabled.  Instead of having the workspace-ids defined in this configuration you would grab the current workspace-id, wrap it in the `Workspace` context and then pass it to the feature-flag client. 

The hard-coded workspace-ids would now reside on the LD side (and would be evaluated against the `Workspace` context provided) instead of being defined within the configuration.

Basically, in LD, if the provided workspace (from the context) matches one of the workspaces listed in LD, then that feature-flag would evaluate to `true`, otherwise `false`.

gosusnp (approved):

PR #21094, opened 2023-01-06
Extend connection list filtering

### LINKED ISSUE #21024: web_backend/v1/connections/list should allow filtering by multiple source or destination id's
#19850 enabled filtering by a single source or destination id.
This can be leveraged on the source and destination item pages since we only need to filter on a single id, however the list pages have multiple source or destination id's and we would like to be able to filter connections returned to just those.

Ideally the endpoint would take a string[], but if a comma-delimited list is easier that's doable as well.

gosusnp:
## What

Enable querying connections with a list of sources/destinations.

Closes #21024 

## How
* Update DB layer to support list of sources and destinations
* Update API definition


krishnaglick (approved):
Tested locally, LGTM! Did an implementation [here](https://github.com/airbytehq/airbyte/pull/21119)
The [existing PR](https://github.com/airbytehq/airbyte/pull/21027/files) for this work will generate conflicts but that'll be easy for me to take care of 👍

jdpgrailsdev (approved):
:shipit:

PR #21101, opened 2023-01-06
🪟🐛 Connector form: Render name as first control in the form

flash1293:
## What

This PR fixes a bug introduced with https://github.com/airbytehq/airbyte/pull/20809 - the "name" input in the connector form is rendered in the last position instead of the first position.

## How

Javascript remembers the order of insertion of properties in an object. The old code would set the `connectionConfiguration` first, then add the "name" if the connector form is used in the source or destination page. As it's added last, it will be rendered last.

The new code will rebuild the wrapping `properties` object and put name first before spreading the rest. This also gets rid of an unnecessary null assertion.


lmossman (approved):
Changes LGTM, tested locally and it fixes the issue

PR #21108, opened 2023-01-06
🪟 🐛 🎨 Fix stream table icon checkboxes and icons

### LINKED ISSUE #20974: Disabled streams in streams table collapse their checkboxes
image

When you disable a stream it collapses the checkbox to start bulk-editing that stream. The checkbox should hold it's size.

This issue applies to both the current and new streams table.

edmundito:
## What
Fixes #20974 

This fixes an issue where the checkboxes in the streams table (current and new) would collapse when the stream was changed in any way. Additionally, it updates the icons to use the SVGs provided in the design so that the design more closely matches the Figma.

Current table (ignore the gap between rows, that's a separate bug):
Screen Shot 2023-01-06 at 09 25 28

New table:

Screen Shot 2023-01-06 at 09 26 44

## How
For the current table: CSS adjustments and icon swapping. The icon container has a fixed size so that the spacing remains there even when there's no icon.

For the new table: Fixed size icon container and added new icons. While reworking the icon component, I added the ability to fade in and out the current icon by updating the status from an effect and using opacity transitions.

## Recommended reading order
From top to bottom

### DIFF
--- a/airbyte-webapp/src/components/CreateConnection/__snapshots__/CreateConnectionForm.test.tsx.snap
+++ b/airbyte-webapp/src/components/CreateConnection/__snapshots__/CreateConnectionForm.test.tsx.snap
@@ -635,6 +635,9 @@ exports[`CreateConnectionForm should render 1`] = `
                         <div
                           class="<removed-for-snapshot-test>"
                         >
+                          <div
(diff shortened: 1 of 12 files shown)


krishnaglick (approved):
Tested locally, LGTM

PR #21109, opened 2023-01-06
🪟 🔧 UI Table migration to v8

### LINKED ISSUE #17941: Migrate Table UI component to react-table v8
The Table component we're having has pretty broken types at the moment. It does not know about any `cell.value` type (they are all `any`), also there are a lot of generic types from `react-table` used which never gets a generic parameter passed in and defaulting to `any`.

The types of this component should get a complete cleanup, since it's at the moment very dangerous to change anything around it, since the types are not giving much type-safteyness.

YatsukBogdan1:
## What
Closes airbytehq/airbyte#17941 

## How
Create a new component (`NextTable` temporary name) to use instead of `Table`, and migrate all `Table` occurrences to use new one


airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
```diff
@@ -1,43 +1,87 @@
...
+const getStyleFromColumnMeta = <T,>(meta: ColumnMeta<T, unknown>): CSSProperties => {
+  const style: CSSProperties = {};
+  if (meta?.customWidth) {
+    style.width = `${meta?.customWidth}%`;
+  }
+  if (meta?.customPadding?.left) {
+    style.paddingLeft = `${meta?.customPadding?.left}px`;
+  }
+  if (meta?.customPadding?.right) {
+    style.paddingRight = `${meta?.customPadding?.right}px`;
```
krishnaglick, line 30: Nit: You don't need to `?` within the if block.

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
```diff
@@ -1,43 +1,87 @@
...
-                colSpan={header.colSpan}
-                className={classNames(styles.th, header.column.columnDef.meta?.thClassName)}
-                key={`table-column-${headerGroup.id}-${header.id}`}
-              >
-                {flexRender(header.column.columnDef.header, header.getContext())}
-              </th>
-            ))}
+            {headerGroup.headers.map((header) => {
+              const { meta } = header.column.columnDef;
+              const style: CSSProperties = getStyleFromColumnMeta<TData<T>>(meta as ColumnMeta<T, unknown>);
```
krishnaglick, line 56: Nit: Don't need CSSProperties here

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
```diff
@@ -1,43 +1,87 @@
...
+              return (
+                <th
+                  colSpan={header.colSpan}
+                  className={classNames(styles.th, meta?.thClassName, {
```
krishnaglick, line 60: Does this filter out undefined or will it apply "undefined" as a class name?

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx
```diff
@@ -44,7 +45,7 @@ const UsagePerConnectionTable: React.FC<UsagePerConnectionTableProps> = ({ credi
   const { sourceDefinitions } = useSourceDefinitionList();
   const { destinationDefinitions } = useDestinationDefinitionList();
 
-  const creditConsumptionWithPercent = React.useMemo(() => {
+  const creditConsumptionWithPercent = React.useMemo((): FullTableProps[] => {
```
krishnaglick, line 48: Nit:
```suggestion
  const creditConsumptionWithPercent = React.useMemo<FullTableProps[]>(() => {
```

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx
```diff
@@ -102,14 +104,14 @@ const UsagePerConnectionTable: React.FC<UsagePerConnectionTableProps> = ({ credi
   );
 
   const sortingData = React.useMemo(
-    () => creditConsumptionWithPercent.sort(sortData),
+    (): FullTableProps[] => creditConsumptionWithPercent.sort(sortData),
```
krishnaglick, line 107: Nit: Move above to `useMemo<FullTableProps[]>`

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx
```diff
@@ -102,14 +104,14 @@ const UsagePerConnectionTable: React.FC<UsagePerConnectionTableProps> = ({ credi
...
     [sortData, creditConsumptionWithPercent]
   );
 
   const columns = React.useMemo(
-    () => [
+    (): Array<ColumnDef<FullTableProps>> => [
```
krishnaglick, line 112: Nit: Same here

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx
```diff
@@ -139,31 +145,45 @@ const UsagePerConnectionTable: React.FC<UsagePerConnectionTableProps> = ({ credi
...
+        },
       },
     ],
     [onSortClick, sortBy, sortOrder]
   );
 
   return (
     <Content>
-      <Table columns={columns} data={sortingData} light />
+      <NextTable<FullTableProps> columns={columns} data={sortingData} light />
```
krishnaglick, line 186: I did some testing to remove the @ts-ignore but ended up with an odd error here:
Screenshot 2023-01-09 at 1 54 45 PM
Seems like something can be improved in the type chain.

krishnaglick: [Here](https://github.com/airbytehq/airbyte/blob/2713fa86192a20ed0c3964be47fbfd8669fda331/airbyte-webapp/src/components/ui/NextTable/NextTable.tsx#L14)
`columns: Array<ColumnDef<TData<T>, string>>;` seems to fix it.

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
```diff
@@ -1,43 +1,87 @@
...
+  data: Array<TData<T>>;
+  erroredRows?: boolean;
+  light?: boolean;
+  onClickRow?: (data: TData<T>) => void;
 }
 
-export const NextTable = <TData,>({ columns, data, onClickRow, className }: PropsWithChildren<TableProps<TData>>) => {
+const getStyleFromColumnMeta = <T,>(meta: ColumnMeta<T, unknown>): CSSProperties => {
+  const style: CSSProperties = {};
+  if (meta?.customWidth) {
```
dizel852, line 23: any reason why we introduced the style props?
We can pass the desired style via css in the meta

airbyte-webapp/src/types/react-table.d.ts
```diff
@@ -7,7 +7,14 @@ import "@tanstack/react-table";
  */
 declare module "@tanstack/table-core" {
   interface ColumnMeta {
-    thClassName?: string;
+    collapse?: boolean;
+    customPadding?: {
```
dizel852, line 11: I'd prefer to avoid adding style props here since after some time we will get the same result as we have in the current implementation.

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
YatsukBogdan1: just checked this. seems like `classNames` filters out `undefined` value, so it is not going into DOM

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
```diff
@@ -1,43 +1,72 @@
...
-  columns: Array<ColumnDef<TData>>;
-  onClickRow?: (data: unknown) => void;
+type TData<T> = T & {
+  error?: unknown;
+};
+
+export interface TableProps<T> {
   className?: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  columns: Array<ColumnDef<TData<T>, any>>;
```
krishnaglick, line 14: Don't love this any but unsure if we can do anything about it. It's fine, just would like it if it was unnecessary.

krishnaglick (commented):
Nothing stands out to me w/ the current state of things!

airbyte-webapp/src/components/ui/NextTable/NextTable.module.scss
```diff
@@ -18,23 +23,48 @@
...
+    &.light {
+      padding-left: vars.$spacing-lg;
+    }
   }
 
   &:last-child {
-    padding-left: vars.$spacing-lg;
-    border-radius: 0 vars.$border-radius-md 0 0;
+    padding-left: 15px;
+    border-radius: 0 10px 0 0;
```
dizel852, line 63: Please fix with css variables

airbyte-webapp/src/components/ui/NextTable/NextTable.module.scss
```diff
@@ -18,23 +23,48 @@
...
+  &.collapse {
+    /* stylelint-disable-next-line number-max-precision */
+    width: 0.0000000001%;
+  }
 
   &:first-child {
-    padding-left: calc(vars.$spacing-2xl + vars.$spacing-sm);
-    border-radius: vars.$border-radius-md 0 0;
+    padding-left: vars.$spacing-2xl;
+    border-radius: 10px 0 0;
```
dizel852, line 54: css variables was removed

airbyte-webapp/src/components/EntityTable/ConnectionTable.tsx
```diff
@@ -25,6 +27,16 @@ interface IProps {
...
+type ColumnDefs = [
+  ColumnDef<ITableDataItem, string>,
+  ColumnDef<ITableDataItem, string>,
+  ColumnDef<ITableDataItem, string>,
+  ColumnDef<ITableDataItem, ConnectionScheduleData>,
+  ColumnDef<ITableDataItem, number>,
+  ColumnDef<ITableDataItem, boolean>,
+  ColumnDef<ITableDataItem, string>
+];
+
```
dizel852, line 39: Please, use `columnHelper()` from `@tanstack/react-table` so we don't need to redefine types by ourselves.
Doc: https://tanstack.com/table/v8/docs/guide/column-defs#column-helpers

airbyte-webapp/src/components/ui/NextTable/NextTable.module.scss
```diff
@@ -18,23 +23,48 @@
...
+  background: colors.$dark-blue;
   color: colors.$grey;
   text-transform: uppercase;
   border-bottom: none;
   text-align: left;
   font-size: 10px;
   line-height: 12px;
   font-weight: 600;
+  width: auto;
+  padding: 9px 13px 10px;
```
dizel852, line 35: css variables

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
```diff
@@ -1,43 +1,72 @@
...
-  data: TData[];
-  columns: Array<ColumnDef<TData>>;
-  onClickRow?: (data: unknown) => void;
+type TData<T> = T & {
+  error?: unknown;
+};
+
+export interface TableProps<T> {
   className?: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
dizel852, line 13: I'm pretty sure we don't need to provide `any` type here since the goal of refactoring was to make out base table more strictly typed

airbyte-webapp/src/components/EntityTable/ConnectionTable.tsx
```diff
@@ -146,39 +173,43 @@ const ConnectionTable: React.FC<IProps> = ({ data, entity, onClickRow, onSync })
...
+          thClassName: styles.thConnectionId,
+        },
+        cell: (props: CellContext<ITableDataItem, string>) => <ConnectionSettingsCell id={props.cell.getValue()} />,
       },
     ],
     [sortBy, sortOrder, entity, onSortClick, onSync, allowSync, allowAutoDetectSchema]
   );
 
-  return <Table columns={columns} data={sortingData} onClickRow={onClickRow} erroredRows />;
+  return <NextTable columns={columns} data={sortingData} onClickRow={onClickRow} erroredRows />;
```
dizel852, line 212: `erroredRows` - I don't the this in props, how do we get it?

airbyte-webapp/src/types/react-table.d.ts
```diff
@@ -7,7 +7,9 @@ import "@tanstack/react-table";
  */
 declare module "@tanstack/table-core" {
   interface ColumnMeta {
-    thClassName?: string;
+    collapse?: boolean;
+    headerHighlighted?: boolean;
```
dizel852, line 11: I'm not sure I understand what is `headerHighlighted`. 
Can we style instead?

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx
```diff
@@ -139,31 +151,37 @@ const UsagePerConnectionTable: React.FC<UsagePerConnectionTableProps> = ({ credi
...
-        customPadding: { left: 0 },
-        Cell: ({ cell }: CellProps<FullTableProps>) => <UsageCell percent={cell.value} />,
+        header: "",
+        accessorKey: "creditsConsumedPercent",
+        meta: {
+          thClassName: styles.thCreditsConsumedPercent,
+        },
+        cell: (props) => <UsageCell percent={props.cell.getValue()} />,
       },
       // TODO: Replace to Grow column
```
dizel852, line 169: Is this comment relevant?

airbyte-webapp/src/components/ui/NextTable/NextTable.module.scss
```diff
@@ -18,23 +23,48 @@
...
+    text-align: left;
+    font-size: 11px;
+    line-height: 12px;
+    font-weight: 400;
+  }
+
+  &.collapse {
+    /* stylelint-disable-next-line number-max-precision */
+    width: 0.0000000001%;
+  }
```
dizel852, line 50: Just curious, this sounds to be like a row collapse.
Is this style just set minimum width?

dizel852 (changes requested):
Did a review, found some issues that need to be addressed:
1. We need to be really carefully updating base table styles, since it affects all other tables. Streams details panel is changed now:
Before changes:
Screenshot at Jan 20 17-39-24

After changes:
Screenshot at Jan 20 17-41-55

2. Some props, like `erroredRows` and `collapse`, really sounds to me like something table-specific(not all tables can have errors). But we can extend the base component to change the `<tr/>` background color if we need. Except "light", since we use 2 styles of table header(blue and white)

3. Base table should be as strictly typed as as is can be. This will save us a lot of time in future

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
YatsukBogdan1: I left a comment in the `NextTable` component. We tried yesterday with @dizel852 to get rid of `any` here, but we didn't find a way. It also seems to be type safely, since the `useReactTable` `options.columns` itself expects `Array<ColumnDef<T, any>>`

YatsukBogdan1:
Fixed most of the comments, but there is still some questions I need answers for. I think @krishnaglick or @edmundito can give an answer
- do we need to keep `erroredRows` table property? seems like it should color rows which have `error: true` in red, but yesterday we were looking with @dizel852 into tables that used this property, and seems there is no actual results with this styling
- do we need to keep `headerHighlighted` inside column `meta`? seems this is not table specific feature, and can be done through direct styling of header
- I was trying to change all usages of magic numbers inside SCSS to variables, but some of values was specific, so I used the  closest values from SCSS variables; is this ok for us, or should we play around to make all spacing pixel perfect, as it was in original `Table` component?
- there is comment inside `src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx` is it still relevant for us? and how we should handle it?
```
// TODO: Replace to Grow column
columnHelper.accessor("connectionId", {
  header: "",
  cell: () => <div />,
  meta: {
    thClassName: styles.thConnectionId,
  },
}),
```
Also:
- we investigated with @dizel852 on `react-table` `columns` typization and seems safely to have `NextTable` property `columns: Array<ColumnDef<TData<T>, any>>;` with `any` since `useReactTable` `options.columns` hook itself expects `Array<ColumnDef<T, any>>`

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
dizel852: Confirm, double checked that

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
```diff
@@ -1,43 +1,72 @@
...
         ))}
       </thead>
       <tbody>
         {table.getRowModel().rows.map((row) => (
-          <tr className={styles.tr} key={`table-row-${row.id}`} onClick={() => onClickRow?.(row.original)}>
+          <tr
+            className={classNames(styles.tr, {
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore
+              [styles.withError]: erroredRows && !!row.original.error,
```
tealjulia, line 64: I can't find a code path where we would need/want this.  I may be wrong here... but I _think_ this may be removable.

airbyte-webapp/src/components/EntityTable/components/FrequencyCell.tsx
```diff
@@ -27,7 +27,7 @@ const FrequencyCell: React.FC<FrequencyCellProps> = ({ value, enabled, scheduleT
     <Content enabled={enabled}>
       <FormattedMessage
         id={`frequency.${value ? value.basicSchedule?.timeUnit : "manual"}`}
```
tealjulia, line 29:
```suggestion
        id={`frequency.${value?.basicSchedule?.timeUnit ?? "manual"}`}
```
more consistent since we're optional chaining below

airbyte-webapp/src/components/EntityTable/components/LastSyncCell.tsx
```diff
@@ -7,7 +7,7 @@ const Content = styled.div<{ enabled?: boolean }>`
 `;
 
 interface IProps {
-  timeInSecond: number;
+  timeInSecond?: number | null;
```
tealjulia, line 10: nit: `timeInSeconds`

airbyte-webapp/src/components/EntityTable/components/LastSyncCell.tsx
```diff
@@ -7,7 +7,7 @@ const Content = styled.div<{ enabled?: boolean }>`
 `;
 
 interface IProps {
```
tealjulia, line 9: nit: let's name the prop instead of IProps

airbyte-webapp/src/types/react-table.d.ts
tealjulia: Agree, this should probably be styled using a classname.
I have a separate Issue that I created to clean up the styling on the tables (header alignment, header highlighting, etc.) https://github.com/airbytehq/airbyte/issues/21941

teallarson (changes requested):
Looks like we lost sortability on the credit usage table (credits page on cloud mode)

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
```diff
@@ -1,43 +1,72 @@
...
-  data: TData[];
-  columns: Array<ColumnDef<TData>>;
-  onClickRow?: (data: unknown) => void;
+export interface TableProps<T> {
   className?: string;
+  // We can leave type any here since useReactTable options.columns itself is waiting for Array<ColumnDef<T, any>>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  columns: Array<ColumnDef<T, any>>;
+  data: T[];
+  erroredRows?: boolean;
```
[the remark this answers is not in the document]
dizel852: I guess it came from an old design(probably) - For the case when the connection sync failed and we needed to color it red. And I also stand for removing it, since it doesn't make any effect.

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
dizel852: +1 agree

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
dizel852: Double-checked - we can use `any` here since under the hood react-table uses the same interface

tealjulia:
> there is comment inside src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx is it still relevant for us? and how we should handle it?

@YatsukBogdan1 I'm not sure you got an answer to this one.  Compose team is implementing a full redesign of this table.  It should be ok to get rid of for now.  The main thing on this page is what I mentioned above (table no longer sorts).

airbyte-webapp/src/components/EntityTable/ConnectionTable.tsx
```diff
@@ -16,12 +16,14 @@ import { ConnectorNameCell } from "./components/ConnectorNameCell";
 import { FrequencyCell } from "./components/FrequencyCell";
 import { LastSyncCell } from "./components/LastSyncCell";
 import { StatusCell } from "./components/StatusCell";
-import { ITableDataItem, SortOrderEnum } from "./types";
+import styles from "./ConnectionTable.module.scss";
+import { ConnectionTableDataItem, SortOrderEnum } from "./types";
+import { NextTable } from "../ui/NextTable";
 
 interface IProps {
```
dizel852, line 23:
```suggestion
interface ConnectionTableProps {
```

YatsukBogdan1:
Changes (besides resolved comments):
- merged @matter-q PR for supporting responsive columns; and applied those changes on `<NextTable />`
- fixed styles for `<StremFieldsTable />` component
- removed unused `columnSort` property for `<NextTable />` component (probably will be implemented later while migrating to `react-table` sorting mechanism)

dizel852 (approved):
Did another review - LGTM 👍 
Also, tested locally, on OSS and Cloud envs - looks great. 
Seems like all the comments are fixed. 👍 

Passing to @edmundito and @tealjulia for final review

teallarson (commented):
Is this alignment correct?  Should the checkbox be under the sync title?

Screenshot 2023-01-31 at 12 51 33 PM

dizel852:
> Is this alignment correct? Should the checkbox be under the sync title?
> 
> Screenshot 2023-01-31 at 12 51 33 PM

Hmm, can't say for sure, but in `master` we have the same alignment. 
The Figma design is a little bit different: we don't have there a checkbox for select/unselect all streams

I've asked Bohdan to make the width of the Sync column a little narrower than we have in `master`.

teallarson (approved):
Lgtm!  Excited to get the new components in place.

PR #21117, opened 2023-01-06
allow the cursors and primary keys to be deselected when the sync mode does not require it

mfsiega-airbyte:
## What
Allows the cursor field to be deselected if the sync mode is FULL_REFRESH and the primary key to be deselected if the destination sync mode is not DEDUP.


joeykmh (approved):
Logic looks good to me 👍

PR #21126, opened 2023-01-06
In Cloud, out of date connectors call always returns 0

krishnaglick:
## What
In Cloud, out of date connectors call always returns 0

## How
The app checks if it's running in Cloud mode, and if so doesn't make the call to the server to get out of date connectors; it just returns 0 out of date connectors.

Additionally, the check now properly returns a boolean.

### DIFF
--- a/airbyte-webapp/src/services/connector/ConnectorService.ts
+++ b/airbyte-webapp/src/services/connector/ConnectorService.ts
@@ -1,11 +1,15 @@
 import { useConfig } from "config";
-import { webBackendCheckUpdates } from "core/request/AirbyteClient";
+import { webBackendCheckUpdates, WebBackendCheckUpdatesRead } from "core/request/AirbyteClient";
 import { AirbyteRequestService } from "core/request/AirbyteRequestService";
 import { useDefaultRequestMiddlewares } from "services/useDefaultRequestMiddlewares";
 import { useInitService } from "services/useInitService";
+import { isCloudApp } from "utils/app";
 
 class ConnectorService extends AirbyteRequestService {
-  checkUpdates() {
+  checkUpdates(): Promise<WebBackendCheckUpdatesRead> {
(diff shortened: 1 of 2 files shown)


lmossman (approved):
Changes LGTM

airbyte-webapp/src/services/connector/ConnectorService.ts
```diff
@@ -1,11 +1,15 @@
 import { useConfig } from "config";
-import { webBackendCheckUpdates } from "core/request/AirbyteClient";
+import { webBackendCheckUpdates, WebBackendCheckUpdatesRead } from "core/request/AirbyteClient";
 import { AirbyteRequestService } from "core/request/AirbyteRequestService";
 import { useDefaultRequestMiddlewares } from "services/useDefaultRequestMiddlewares";
 import { useInitService } from "services/useInitService";
+import { isCloudApp } from "utils/app";
```
ambirdsall, line 6: This is great as a hotfix—small, simple, direct, effective—but I do think that a good follow-up would be editing the conditional to check if `FeatureItem.AllowUpdateConnectors` is enabled instead of checking `isCloud()`. It looks like we don't really have an API for that besides `useFeature`, which is a bummer: right now this module doesn't introduce any react-specific context like hooks impose.

krishnaglick: Doing conditionals with hooks always gets awkward, but yeah.
We're going to revert the backend change Monday so maybe @josephkmh can make that change if he gets to it before I'm online!

josephkmh: @ambirdsall agreed - I have an open issue [here](https://github.com/airbytehq/airbyte-cloud/issues/3972) already proposing that approach. I'll keep that issue open as a refactor of this hotfix.

PR #21128, opened 2023-01-06
🪟 Add e2e test for upgrading and downgrading connector versions

teallarson:
## What

closes airbytehq/airbyte-internal-issues#1276

Adds e2e tests for upgrading and downgrading connectors in the settings page.

## How
Adds e2e tests that a connector can be upgraded and downgraded using each of the possible methods to do so in the UI.

Also adds some more descriptive testid's to `airbyte-webapp` for this purpose.

## Recommended reading order
1. `airbyte-webapp-e2e-tests/pages/settingsPage`
2. `airbyte-webapp-e2e-tests/integration/settings.spec.ts`
3. others

### DIFF
--- a/airbyte-webapp-e2e-tests/cypress/integration/settings.spec.ts
+++ b/airbyte-webapp-e2e-tests/cypress/integration/settings.spec.ts
@@ -1,0 +1,33 @@
+import { initialSetupCompleted } from "commands/workspaces";
+import {
+  clickUpgradeAllButton,
+  editVersionByConnectorName,
+  goToSettingsDestinationTab,
+  goToSettingsSourcesTab,
+} from "pages/settingsPage";
+
+describe("Source settings main actions", () => {
+  beforeEach(() => {
+    initialSetupCompleted();
+    goToSettingsSourcesTab();
+  });
+  it("downgrades a source version and updates it through the input", () => {
+    editVersionByConnectorName("Postgres", "0.1.0");
+    editVersionByConnectorName("Postgres", "0.1.1");
+  });
+  it("downgrades a source version and updates it through 'upgrade all' button", () => {
+    editVersionByConnectorName("Postgres", "0.1.0");
+    clickUpgradeAllButton();
+  });
+});
+
+describe("Destination settings main actions", () => {
+  beforeEach(() => {
+    initialSetupCompleted();
+    goToSettingsDestinationTab();
+  });
+  it("downgrades a destination version and updates it through the input", () => {
+    editVersionByConnectorName("Postgres", "0.1.0");
+    editVersionByConnectorName("Postgres", "0.1.1");
+  });
+});
--- a/airbyte-webapp-e2e-tests/cypress/integration/settings.spec.ts
+++ b/airbyte-webapp-e2e-tests/cypress/integration/settings.spec.ts
@@ -1,4 +1,5 @@
-import { connectorsIds } from "commands/common";
+// import { connectorsIds } from "commands/common";
+import { ConnectorIds } from "../../../airbyte-webapp/src/utils/connectors/constants";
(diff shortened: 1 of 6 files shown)


airbyte-webapp-e2e-tests/cypress/integration/settings.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+describe("Source settings main actions", () => {
+  beforeEach(() => {
+    initialSetupCompleted();
+    goToSettingsSourcesTab();
+  });
+  it("downgrades a source version and updates it through the input", () => {
+    editVersionByConnectorName("Postgres", "0.1.0");
+    editVersionByConnectorName("Postgres", "0.1.1");
+  });
+  it("downgrades a source version and updates it through 'upgrade all' button", () => {
```
tealjulia, line 18: Not sure if it is intentional, but an "Upgrade all" button is only present on the `sources` tab, hence why only this describe block has this test.

airbyte-webapp-e2e-tests/cypress/pages/settingsPage.ts
```diff
@@ -0,0 +1,23 @@
...
+  cy.intercept("/api/v1/destination_definitions/list_for_workspace").as("getDestinationDefinitionsList");
+  cy.visit("/settings/destination");
+  cy.wait(5000);
+};
+
+export const editVersionByConnectorName = (connectorName: string, version: string) => {
+  cy.get("[data-testid='connectorNameCell']")
+    .contains(connectorName) // div with icon and text
+    .parents("tr")
+    .find("[data-testid='version-input']")
```
tealjulia, line 17: We could change the test id to include the connector name, but I don't see us using dynamic testid's anywhere so went with this.  Finds the div with the connector name, gets its parent row, finds the input within that row.

airbyte-webapp-e2e-tests/cypress/pages/settingsPage.ts
```diff
@@ -0,0 +1,33 @@
...
+export const editVersionByConnectorName = (connectorName: string, version: string) => {
+  cy.get("[data-testid='connectorNameCell']")
+    .contains(connectorName)
+    .parents("tr")
+    .find("[data-testid='version-input']")
+    .focus()
+    .type(`{selectall} {backspace} ${version}`, { force: true })
+    .parents("tr")
+    .find("[data-testid='versionButton']")
+    .click({ force: true });
```
tealjulia, line 22: This hackiness with `.focus()` and `({ force: true })` shouldn't be needed and aren't working anyways... somehow cypress is finding the element but usually (not always) losing it.  There are a _lot_ of re-renders when cypress is going through these steps...

airbyte-webapp-e2e-tests/cypress/pages/settingsPage.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const editVersionByConnectorName = (
+  connectorType: "source" | "destination",
+  connectorId: string,
+  version?: string
+) => {
+  cy.intercept(`/api/v1/${connectorType}_definitions/update`).as("updateConnectorDefinitions");
+
+  if (version) {
+    cy.get(`[data-testid='${connectorId}-versionInput']`).clear().wait(1000);
```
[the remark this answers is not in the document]
tealjulia: Dang.  I am not entirely sure.  It seems like there's something funky with re-renders going on and I am not entirely sure the cause.  I can spend some time looking into it.

tealjulia:
reverting this one to a draft until i have more time to dig into that page

tealjulia:
Saw some recent work on this page so I checked in.  It does not look like we're at the point where it's stable enough to run these tests (elements are still being "hidden" according to cypress)

PR #21132, opened 2023-01-06
Revert Convert the server to micronaut

benmoriceau:
Reverts airbytehq/airbyte#19194

It is blocking a p0 resolution, reverting + there is some CORS issues detected bu @krishnaglick

PR #21133, opened 2023-01-06
Revert "Revert Convert the server to micronaut"

benmoriceau:
Reverts airbytehq/airbyte#21132


krishnaglick (approved):
Can confirm the FE doesn't get CORS errors during dev!

PR #21142, opened 2023-01-07
[Connector Builder]Add ability to convert from YAML manifest to UI

### LINKED ISSUE #21285: [Connector Builder] Productionize YAML -> UI conversion
## Context

As a result of the POC for this functionality, we are deciding to move forward with implementing this feature.

A PR that lays the groundwork for this can be found here, though it is incomplete: https://github.com/airbytehq/airbyte/pull/21142

This is blocked by completion of https://github.com/airbytehq/airbyte/issues/21211, so that YAMLs using $refs and $options are able to be converted.

## Acceptance Criteria
- YAML manifests that don't encode behavior unsupported by the Connector Builder UI can be converted into UI form values in the Connector Builder UI (by clicking the UI | YAML toggle)
- YAML manifests that do things the Connector Builder UI does not support are rejected during toggling
- YAMLs containing $refs and $options will have those macros resolved and flattened by the endpoint added in https://github.com/airbytehq/airbyte/issues/21211,

lmossman:
## What
Resolves https://github.com/airbytehq/airbyte/issues/21285

Adds logic to handle converting from a YAML manifest to the connector builder UI form values.

Also updates the confirmation modal service to allow sending FormattedMessage values along with the i18n id


flash1293 (commented):
A few things I noticed that are not checked (probably missing some):
* Nested cartesian slicers
* Whether a requestor is a custom requester
* Whether an extractor is custom extractor
* checkpoint interval and transformations which can be defined on a stream and would get lost
* error handler can be defined on the retriever and would get lost

Not sure whether it makes sense, but would it make sense to do the following:
* Define a scoped down json schema which defines only the parts we want to support - then as a first step in the conversion logic we check against this scoped down schema. This will get rid of a lot of custom checks, probably easier to maintain
* For the additional props we don't support in the UI - make sure they are kept around even if they are not editable (would work e.g. for the transformations or for the requester error handler). This should work already for the spec but it should probably become a common thing.
* For the inferred inputs: Your suggestion sounds good to me
* For the substream slicers: Yeah, this is unfortunately a little more complex. This is what I had in mind:
  * Recursively go through all streams with its substream slicers (not actually converting them yet) and store all streams in serialized form in a set (`JSON.stringify(streamDeclaration)`
  * For the actual conversion, instead of iterating over `manifest.streams`, iterate over the set of collected streams (id can be the index as it's unique)
  * If a substream slicer is encountered, stringify the parent stream definition again and look it up from the set - this will give you the parent stream id.

lmossman:
@flash1293 I've pushed up some more commits here to handle converting the spec and auth components from YAML to UI. Could you take a look and let me know what you think?

I haven't addressed the other parts of your feedback yet for a couple reasons:
1. This POC is starting to drag on due to the complexity of converting from YAML to UI, and I don't want to spend too much time on implementing these changes when it feels like there is other higher-priority work
2. This is already starting to feel pretty complex and prone to weird behavior, and I haven't even got to the other parts where I expect even more complexity (keeping around additional props not supported in the UI, and the substream slicer issue)
3. It feels like maintaining all of this complex conversion logic as more changes are made to the low-code YAML will be burdensome.

What do you think? Am I overemphasizing the complexity of this? 
I just feel that the value gained from this may not outweigh the complexity in making all of this work and maintaining it going forward.

airbyte-webapp/src/components/connectorBuilder/types.ts
```diff
@@ -115,117 +122,130 @@ export const DEFAULT_BUILDER_STREAM_VALUES: Omit<BuilderStream, "id"> = {
...
-        required: true,
-        definition: {
-          type: "string",
-          title: "API Key",
-          airbyte_secret: true,
-        },
+const authTypeToKeyToInferredInput: Record<string, Record<string, BuilderFormInput>> = {
+  NoAuth: {},
+  ApiKeyAuthenticator: {
+    api_token: {
```
flash1293, line 128: Looks like these property names always need to match the `key` otherwise it won't recognize inferred inputs vs regular inputs.

airbyte-webapp/src/components/connectorBuilder/types.ts
```diff
@@ -607,3 +674,246 @@ export const convertToManifest = (values: BuilderFormValues): ConnectorManifest
...
+
+  if (manifestSpec === undefined) {
+    return result;
+  }
+
+  const required = manifestSpec.connection_specification.required as string[];
+  result.inputs = Object.entries(
+    manifestSpec.connection_specification.properties as Record<string, AirbyteJSONSchema>
+  ).flatMap(([key, definition]) => {
+    if (Object.keys(authTypeToKeyToInferredInput[result.auth.type]).includes(key)) {
```
flash1293, line 817: looks like the keys of that object are the properties of the authenticator object, not the name of the input (in this case it's causing a mismatch between api_key and api_token). Should probably be
```
if (Object.values(authTypeToKeyToInferredInput[result.auth.type])
        .map((input) => input.key).includes(key)) {
```

Otherwise you will end up with two api_keys when switching around

airbyte-webapp/src/components/connectorBuilder/types.ts
```diff
@@ -607,3 +674,246 @@ export const convertToManifest = (values: BuilderFormValues): ConnectorManifest
...
+    throw new ManifestCompatibilityError(
+      streamName,
+      "schema_loader type is unsupported. Only InlineSchemaLoader is supported."
+    );
+  }
+
+  const inlineSchemaLoader = manifestSchemaLoader as InlineSchemaLoader;
+
+  if (inlineSchemaLoader.schema) {
+    return JSON.stringify(inlineSchemaLoader.schema);
```
flash1293, line 740: should use nicely formatted schema:
Screenshot 2023-01-11 at 10 58 16

flash1293:
> What do you think? Am I overemphasizing the complexity of this?
I just feel that the value gained from this may not outweigh the complexity in making all of this work and maintaining it going forward.

I think it's worth it for two reasons: 
* It brings UI and YAML pretty close together, allowing user to use it a little bit if it makes sense, but not being limited by it. Having it strictly one-way means a lot of advanced users will drop it completely because it limits them too much.
* We wouldn't strictly need to have this logic right now, but it's a nice canary to make it less likely for the manifest and the UI to diverge over time as it forces us to think about the conversion every time we add new functionality on either side.

> But if you feel differently and we do decide to move forward with this, then we can create separate tickets for the other parts you mentioned

Definitely, we can just fail on these conditions for now and get better in conversion later - it's lossy anyway, so it's not a fundamental change.

lmossman:
@flash1293 I think I've addressed all of the feedback here - this should be working with both inputs and substream slicers now.

Also, I didn't end up needing to check for nested Cartesian slicers, because the manifest schema already doesn't allow that. Notice that Cartsian isn't listed here: https://github.com/airbytehq/airbyte/blob/189891c12f6b12a3162b545f0855a620426bead8/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml#L134-L138

Keeping the PR in draft because I still want to add unit tests for the conversion logic, but I think it would be useful to get another review on the implementation.

airbyte-webapp/src/components/connectorBuilder/manifestToBuilderForm.ts
```diff
@@ -0,0 +1,386 @@
...
+  BuilderPaginator,
+  BuilderStream,
+  BuilderSubstreamSlicer,
+  DEFAULT_BUILDER_FORM_VALUES,
+  DEFAULT_BUILDER_STREAM_VALUES,
+  isInterpolatedConfigKey,
+} from "./types";
+import { formatJson } from "./utils";
+
+export const convertToBuilderFormValues = (
```
lmossman, line 36: Decided to move this and its related methods into its own file, because types.ts was starting to get pretty large

airbyte-webapp/src/components/connectorBuilder/manifestToBuilderForm.ts
```diff
@@ -0,0 +1,384 @@
...
+    if (errorMessage[0] === '"') {
+      errorMessage = errorMessage.substring(1, errorMessage.length);
+    }
+    if (errorMessage.slice(-1) === '"') {
+      errorMessage = errorMessage.substring(0, errorMessage.length - 1);
+    }
+    const tracebackIndex = errorMessage.indexOf(" - Traceback");
+    if (tracebackIndex >= 0) {
+      errorMessage = errorMessage.substring(0, tracebackIndex);
+    }
```
lmossman, line 53: This can be removed once https://github.com/airbytehq/airbyte/pull/21443 is merged

flash1293: Looks like this is merged in

airbyte-webapp/src/components/connectorBuilder/manifestToBuilderForm.ts
```diff
@@ -0,0 +1,384 @@
...
+      stream.name
+    );
+  }
+
+  return {
+    ...DEFAULT_BUILDER_STREAM_VALUES,
+    id: index.toString(),
+    name: stream.name ?? "",
+    urlPath: requester.path,
+    httpMethod: (requester.http_method as "GET" | "POST" | undefined) ?? "GET",
```
flash1293, line 139: Not broken by you, but I just realized we do not set the http_method on the requester when doing form state -> manifest conversion. This is a bug right?

airbyte-webapp/src/components/connectorBuilder/YamlEditor/YamlEditor.tsx
```diff
@@ -1,18 +1,24 @@
...
 import debounce from "lodash/debounce";
-import isMatch from "lodash/isMatch";
+import isEqual from "lodash/isEqual";
 import { editor } from "monaco-editor/esm/vs/editor/editor.api";
 import { useEffect, useMemo, useRef, useState } from "react";
 
 import { CodeEditor } from "components/ui/CodeEditor";
 
 import { ConnectorManifest } from "core/request/ConnectorManifest";
+// import { useConfirmationModalService } from "hooks/services/ConfirmationModal";
```
flash1293, line 12: leftover

flash1293:
Overall this looks great! Due to the high redundancy in the configuration I don't really see users switching between UI and YAML at the moment in a super effective way, but that's something we can gradually improve over time in multiple ways. It should work just fine for the initial use case we had in mind about only storing the manifest and not the builder form values along with it.

airbyte-webapp/src/components/connectorBuilder/manifestToBuilderForm.ts
lmossman: Yep, good catch. I'll include the fix in this PR since it's just one line

airbyte-webapp/src/components/connectorBuilder/manifestToBuilderForm.test.ts
```diff
@@ -0,0 +1,434 @@
...
+  retriever: {
+    name: "stream2",
+    requester: {
+      name: "stream2",
+      path: "/stream-2-path",
+    },
+  },
+});
+
+describe("Conversion throws error when", () => {
```
lmossman, line 52: I originally started to add tests for every case that throws a ManifestCompatibilityError, but that felt overly excessive and just means that we would be repeating all of those cases in yet another place, so I decided against it.

I think its more important that we test the more complex conversion cases anyway.

flash1293 (approved):
Tested and works really well! LGTM

Issue #19981, opened 2022-12-01
Automate version bump and package publishing for `airbyte-cdk`


clnoll:
## Tell us about the problem you're trying to solve
There are two issues when publishing the CDK: 
1. Many people could be working on the CDK in parallel, which results in version conflicts between in-flight PRs for the `airbyte-cdk` package
2. Publishing the CDK is done from a branch, which can lead to nasty race conditions, not to mention being tedious/manual 

## Describe the solution you’d like
We could solve the two problems at once, or we could solve just one or the other. 
To solve 1 we need to add a GitHub Action to bump the version and publish the package on merge to `master`. If we do this, we'll want a way to indicate whether it's a major/minor version change. One possible solution could be to use the commit message, or possibly the PR description since the template indicates whether there are breaking changes.

To solve 2 we need a Github action that checks whether the `airbyte-cdk` directory was changed on merge to master, and runs the publish workflow if so. We will also need a clear path for error recovery if the publishing fails for any reason. 

The implementer of this issue is free to think about whether to solve both issues vs. just one, and the sequencing of a solution. 

## Describe the alternative you’ve considered or used
We could queue up the PRs that would require a version bump and merge them once a day. We might end up wanting to do this eventually anyway, with or without automatically bumping the version, so it's somewhat orthogonal.

maxi297:
## Current situation
<img width="852" alt="image" src="https://user-images.githubusercontent.com/3360483/211082493-83b084a7-cb46-415f-aebf-c6cf38ca29ad.png">

## Challenges with proposed solution
There were a couple challenges making the release automatic on merge:
* What happens if the version in the setup.py has been updated as part of the PR? Do we generate an error and block the merge? Do we release on that version even though it would allow race conditions as stated in the issue?
* Input needs to be provided by the dev. However, it is not very intuitive when this input should be provided when merging the PR. We could check the last comment with a specific format but what happens if the dev forgets before merging?
* What happens if the pipeline on master fails? Should we still publish the new version?

## New Proposed Solution
Let's decouple two problems: 
* Race condition generated by releasing in the feature branches
* Having to manually do something to release

Given that releasing needs user input, let's ignore the second problem for now. This enables us to create this flow:
<img width="598" alt="image" src="https://user-images.githubusercontent.com/3360483/211082571-535427a2-6b54-4ee3-ba7e-aab9477dff65.png">

This gives more flexibility to the dev:
[...]

PR #21157, opened 2023-01-09
[ISSUE #19981] adding notifications to 'Publish CDK Manually'

maxi297:
## What
Adding notifications on release


clnoll:
@maxi297 do you have screenshots of the notifications that you could add to the PR description?

maxi297:
> @maxi297 do you have screenshots of the notifications that you could add to the PR description?

@clnoll I did a test on `#issue-19981_test-release-notification` that looks like this:
image

.github/workflows/publish-cdk-command-manually.yml
```diff
@@ -58,6 +58,34 @@ jobs:
...
           commit_user_email: octavia-squidington-iii@users.noreply.github.com
+      - name: Post failure to Slack channel dev-connectors-extensibility
+        if: !success()
+        uses: slackapi/slack-github-action@v1.23.0
+        continue-on-error: true
+        with:
+          channel-id: C02TL38U5L7
+          payload: |
+            {
+                "text": "Error while publishing Airbyte CDK!"
```
clnoll, line 69: Any reason not to give a more specific message depending on whether the error occurred during `bump-version`, `build-cdk`, or `publish-cdk`?

clnoll (changes requested):
Looks good, just 1 q for you @maxi297.

.github/workflows/publish-cdk-command-manually.yml
maxi297: We can. I was checking the process and I think that in any case, the user will have to click on the link of the workflow to have more detail and therefore, he will have the information anyway. If we think it has value to be added earlier (like in the Slack message), let me know and I can add it

clnoll: I think that extra bit of detail would be nice to have.

.github/workflows/publish-cdk-command-manually.yml
```diff
@@ -58,6 +58,34 @@ jobs:
           commit_message: 🤖 Bump ${{ github.event.inputs.release-type }} version of Airbyte CDK
           commit_user_name: Octavia Squidington III
           commit_user_email: octavia-squidington-iii@users.noreply.github.com
+      - name: Post failure to Slack channel dev-connectors-extensibility
```
brianjlai, line 61: My main worry about posting this to our main channel is that we might start to spam or clutter it if we have a lot of deploys or failed workflows in any given day. How would you feel about a side channel #airbyte-cdk-releases or something like that?

brianjlai: sounds good thank you. if alerts ultimately are not too noisy i have no concerns moving it to main channel.

clnoll (approved):
LGTM!

PR #21158, opened 2023-01-09
ci-connector-ops: change required reviewers logic

alafanechere:
## What
The way [the action-required-review](https://github.com/Automattic/action-required-review) works mean that if any team in the list of required reviewer reviews the check passes.
On GA connectors we explicitly want a connector team to review a PR when these changes happen: 
- `backward_compatibility`
- `test_strictness_level`

When none of the change above happen but the connector changed is GA we want GL Python's connector team to review.
 
## How
Change this function logic:
https://github.com/airbytehq/airbyte/blob/bdcbee6f7d41aa446dfe3b426f994ae3aebe5b18/tools/ci_connector_ops/ci_connector_ops/sat_config_checks.py#L53


erohmensing (approved):
This is a bummer, because I went to suggest what I'd expect to be the default behavior, and it was exactly what you'd implemented before 😄 (I.e. GL always is requested for GA connectors, connectors teams are added as required as necessary). 

Based on the way the action works, I think you're right in that this is the best way to implement it now unless we wanted to add 2 separate checks - one for GL on GA connectors and one for Connectors team as necessary. I guess connectors teams will have to add GL as necessary after they do their reviews re: strictness level or compatibility, though I wouldn't be surprised if GL devs had already requested reviews from their colleages anyway.

PR #21161, opened 2023-01-09
🪟🐛 Connector builder UI: Fix datetime stream slicer

flash1293:
The 0.16 upgrade of the CDK changed some fields in the datetime stream slicer. This PR adjusts the UI accordingly:
* Add `cursor_granularity`
* Remove wrong pattern matching and example for lookback window and step

Didn't add a better pattern for these inputs because the ISO standard for durations is a little more complex and I think a wrong pattern is worse than none, especially as the server is doing validation as well.

We can add one back in a later step.


airbyte-webapp/src/components/connectorBuilder/Builder/StreamSlicerSection.tsx
```diff
@@ -113,8 +118,7 @@ export const StreamSlicerSection: React.FC<StreamSlicerSectionProps> = ({ stream
             type="string"
             path={buildPath("step")}
             label="Step"
-            tooltip="Time interval for which to break up stream into slices, e.g. 1d"
-            pattern={timeDeltaRegex}
+            tooltip="Time interval for which to break up stream into slices, e.g. P1D for daily slices"
```
lmossman, line 121: nit: add the name of the format to the description
```suggestion
            tooltip="Time interval (ISO 8601 duration) for which to break up stream into slices, e.g. P1D for daily slices"
```

airbyte-webapp/src/components/connectorBuilder/Builder/StreamSlicerSection.tsx
```diff
@@ -127,7 +131,7 @@ export const StreamSlicerSection: React.FC<StreamSlicerSectionProps> = ({ stream
               type="string"
               path={buildPath("lookback_window")}
               label="Lookback window"
-              tooltip="How many days before the start_datetime to read data for, e.g. 31d"
+              tooltip="How many days before the start_datetime to read data for, e.g. P1M for looking back one month"
```
lmossman, line 134:
```suggestion
              tooltip="Time interval (ISO 8601 duration) before the start_datetime to read data for, e.g. P1M for looking back one month"
```

lmossman (approved):
Tested locally and it works as expected. Had a couple small nits on the descriptions but otherwise LGTM

flash1293:
You can see how it failed the build here (before the fix with the unit tests was merged in): https://github.com/airbytehq/airbyte/actions/runs/3882948666/jobs/6623703102

PR #21171, opened 2023-01-09
convert EnvVariableFeatureFlag to an injected dependency

### LINKED ISSUE #21139: update EnvVariableFeatureFlags usages
The `EnvVariableFeatureFlags` class is currently being instantiated via code similar to:
```java
class Example {
  private final boolean logConnectorMessages = new EnvVariableFeatureFlags().logConnectorMessages();
}
```
This makes it difficult to migrate to the new [feature-flag implementation](#21091) due to the new client expecting to be injected as a dependency where necessary. 

This story is to update the usage of the `EnvVariableFeatureFlags` class to be dependency injected so that it can be swapped out easily with the new feature-flag work.

colesnodgrass:
## What
- fix #21139
- made it easier to change the feature-flag implementation by injecting the dependency instead of instantiating within the various classes modified in this PR
- does not modify integrations (will do as a separate PR)

## How
- replace `new EnvVariableFeatureFlags()` calls with injected `FeatureFlag` dependency


jdpgrailsdev (approved):
:shipit:

gosusnp (approved):

airbyte-commons-worker/src/test/java/io/airbyte/workers/general/ReplicationWorkerPerformanceTest.java
```diff
@@ -65,7 +66,7 @@ public class ReplicationWorkerPerformanceTest {
   public void executeOneSync() throws InterruptedException {
     final var perSource = new LimitedAirbyteSource();
     final var perDestination = new EmptyAirbyteDestination();
-    final var messageTracker = new AirbyteMessageTracker();
+    final var messageTracker = new AirbyteMessageTracker(new EnvVariableFeatureFlags());
```
benmoriceau, line 69: Nit in the test should we use mock? The test might be affected by the env of where the test are run.

benmoriceau (approved):
on Nit comment on the test. Apart from that, I merged the server before your commit it is more FYI since this shouldn't impact you change.

airbyte-commons-worker/src/test/java/io/airbyte/workers/general/ReplicationWorkerPerformanceTest.java
colesnodgrass: The tests should use a mock, and when I replace this with the new feature-flag client I'll add mocks where necessary.  

The code (and the tests) were and are already referencing the _real_ environment variables.  So the changes in this PR shouldn't have any unexpected side-effects that didn't already exist.

PR #21174, opened 2023-01-09
🪟 🔧 Use jest directly as test runner

timroes:
## What

This is part of moving our build system from CRA to Vite. Since Vite doesn't have any test runner build in we need a solution for running our tests without Create React App. We use jest directly for doing so (elaboration see below).

This PR already switches the tests to run via Jest directly, since we can extract that into a separate PR, since it does not require Vite in any way.

## How

This changes the `npm` scripts to run tests from using CRA (via `craco`) to use `jest` as a command directly. It also adds the required configuration files we'll need to run the tests.

We continue to use babel as a transpiler for jest (instead of using esbuild that we plan to use with vite for actual bundling), since the jest babel integration works better and supports jest mocks hoisting mechanism properly (which esbuild doesn't).

For more details view my inline comments.

## Why jest?

We also looked at [vitest](https://vitest.dev/) as an alternative to run tests. We decided to stay with jest/babel for tests for some reasons:

* Vitest is still very new, while jest is already established.
* We haven't seen issues using `jest` in the past - never change a running system
* Vitest requires a different syntax for mocking, so we'd need to change way more code to get it running.

We also looked at integrating jest with vite via [`vite-jest`](https://www.npmjs.com/package/vite-jest). Given the rare updates of the library, low download rate and felt immaturity, we decided not to go that route.


airbyte-webapp/jest.config.ts
```diff
@@ -0,0 +1,21 @@
+// eslint-disable-next-line jest/no-jest-import
+import type { Config } from "jest";
```
timroes, line 2: :information_source: Actual jest config, compiled from our previous `package.json` plus some additional config we need now, since CRA doesn't take care of it anymore.

airbyte-webapp/jest.config.ts
```diff
@@ -0,0 +1,21 @@
...
+  snapshotSerializers: ["./src/test-utils/classname-serializer.js"],
+  coveragePathIgnorePatterns: ["\\.stories\\.tsx$"],
+  modulePathIgnorePatterns: ["src/.*/__mocks__"],
+  testEnvironment: "jsdom",
+  moduleDirectories: ["node_modules", "src"],
+  moduleNameMapper: {
+    "\\.module\\.scss$": "test-utils/mock-data/mockIdentity.js",
+    "\\.(css|png|svg|scss)$": "test-utils/mock-data/mockEmpty.js",
+  },
+  setupFilesAfterEnv: ["./src/test-utils/setup-tests.ts"],
```
timroes, line 18: :information_source: Moved this file (and the `classname-serializer`) into `src/test-utils`, since we no longer require it to be at `src/setupTests.ts`, which CRA forced us to.

airbyte-webapp/jest.config.ts
```diff
@@ -0,0 +1,21 @@
...
+  // but several node_modules are not transpiled so they would fail without babel transformation running
+  transformIgnorePatterns: [],
+  snapshotSerializers: ["./src/test-utils/classname-serializer.js"],
+  coveragePathIgnorePatterns: ["\\.stories\\.tsx$"],
+  modulePathIgnorePatterns: ["src/.*/__mocks__"],
+  testEnvironment: "jsdom",
+  moduleDirectories: ["node_modules", "src"],
+  moduleNameMapper: {
+    "\\.module\\.scss$": "test-utils/mock-data/mockIdentity.js",
+    "\\.(css|png|svg|scss)$": "test-utils/mock-data/mockEmpty.js",
```
timroes, line 16: :information_source: Make sure we mock asset imports in a way that jest test run properly.

airbyte-webapp/src/test-utils/setup-tests.ts
```diff
@@ -0,0 +1,22 @@
+// jest-dom adds custom jest matchers for asserting on DOM nodes.
+// allows you to do things like:
+// expect(element).toHaveTextContent(/react/i)
+// learn more: https://github.com/testing-library/jest-dom
+import "@testing-library/jest-dom/extend-expect";
+import "@testing-library/jest-dom";
+import "../globals";
+
+// jsdom doesn't mock `matchMedia`, which is required by react-slick
+Object.defineProperty(window, "matchMedia", {
```
timroes, line 10: :information_source: Rewrote this mock slightly from what we had beforehand here, to align with what `jest` actually has in their documentation: https://jestjs.io/docs/26.x/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom

joeykmh (approved):
Looks good, tests pass ✅

PR #21175, opened 2023-01-09
🪟 🎉 Verify auth status on tab focus

teallarson:
## What
Closes airbytehq/airbyte-cloud#3735 and airbytehq/airbyte-cloud#3992

Previously, a user could have two tabs open, logout in one, and could still click in the UI in their other (now out of date) tab and eventually see a vague/ominous "You do not have permissions" error.

Now, they are simply redirected to login upon tab focus.

## How
Adds a useEffect() with a focus listener to our AuthService that watches for auth status changes whenever the window gains focus.


joeykmh (approved):
Nice solution - works well! 

Just noticed while I was testing that we have a similar problem when logging in. If you have two tabs open to the login (or signup) page, then authenticate in one of them, the other one would ideally also automatically log you in. One more thing we could add is a similar mechanism in `src/packages/cloud/views/auth/Auth.tsx` to redirect users from unauthenticated routes (login, signup) to the authenticated routes.

I just created another ticket for the login issue: https://github.com/airbytehq/airbyte-cloud/issues/3992

tealjulia:
Hmm. I did notice just now testing it again (and seeing what it would take to get the login watch to work too... the answer is not much) that we lose the `from=*` query param that nicely redirects back to where you were when you log back in.

I'm going to see if I can get that working again without too much trouble.  It's a nice touch

tealjulia:
@josephkmh that turned out to be fairly trivial to support in this PR as well, and falls under the category "verifying auth on tab focus" so I threw it in here.

I also adjusted the logic for the redirect URL so this flow also redirects users to the page they came from when logged back in.

airbyte-webapp/src/packages/cloud/views/auth/Auth.tsx
```diff
@@ -54,7 +54,17 @@ export const Auth: React.FC = () => {
...
               <Route
                 path="*"
-                element={<Navigate to={`${CloudRoutes.Login}${loggedOut ? "" : `?from=${pathname}`}`} />}
+                element={
+                  <Navigate
+                    to={
+                      user
+                        ? `${pathname}`
+                        : `${CloudRoutes.Login}${
+                            loggedOut && pathname.includes("/settings/account") ? "" : `?from=${pathname}`
```
tealjulia, line 63: I have a feeling there may be a cleaner way to determine whether to include the `?from=` path.  I do think it's nice to include here though.

josephkmh: I agree that it would be nice for the auth service to remain unconcerned with the routing.

I'm not sure I follow the checking of `user` here - wouldn't it always be `null`? `<Auth>` itself is only rendered if `!user` in [cloudRoutes.tsx](https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/packages/cloud/cloudRoutes.tsx#L177)

tealjulia: You're right!  I had thought it was needed to handle the redirect upon login, but it is not.  Addressed in https://github.com/airbytehq/airbyte/pull/21175/commits/e196d89d57938ea79252d00247c55d0cb1ab1dc0

PR #21180, opened 2023-01-10
Shady Faker (example of a connector which sends data to an untrusted URL)

### LINKED ISSUE #21120: Make a shady connector
To test the `allowed_hosts` and network isolation work, we need a connector that does something shady that we can test with.

1. Make an echo server that logs and replays HTTP request payloads it receives
2. Make a connector that does normal source things, but also tries to send a message to the server made in step 1

Later, we'll make sure that this connector would be blocked.

evantahler:
### ⚠️ This PR SHOULD NEVER BE MERGED ⚠️ 

This PR demonstrates a connector being modified to call a remote URL at boot, demonstrating that it has access to the filesystem and contents of `secrets.json`.  This source was chosen for this test as it does not produce or have access to any *real* PII.

This version of the connector is published as `source-faker:shady` via the command:
```
docker buildx build . --platform "linux/amd64,linux/arm64" --tag airbyte/source-faker:shady --push
```

The URL which this connector sends information to is `echo.evantahler.com`, run by @evantahler.  The codebase for this server is here: https://github.com/evantahler/echo-server.  

Ideally, if the network request to `echo.evantahler.com` fails, the connector should fail to run (`- e`), giving us a strong signal that network isolation was successful. 

Closes #21120

### DIFF
--- a/airbyte-integrations/connectors/source-faker/Dockerfile
+++ b/airbyte-integrations/connectors/source-faker/Dockerfile
@@ -27,12 +27,16 @@ RUN echo "Etc/UTC" > /etc/timezone
 # bash is installed for more convenient debugging.
 RUN apk --no-cache add bash
 
+# install shady packages
+RUN apk --no-cache add curl
+RUN curl --version
+
 # copy payload code only
 COPY main.py ./
 COPY source_faker ./source_faker
 
 ENV AIRBYTE_ENTRYPOINT "python /airbyte/integration_code/main.py"
-ENTRYPOINT ["python", "/airbyte/integration_code/main.py"]
+ENTRYPOINT ["bash", "/airbyte/integration_code/source_faker/launcher.sh"]
 
 LABEL io.airbyte.version=2.0.0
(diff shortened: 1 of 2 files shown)


evantahler:
The echo server stores the last 1000 messages it receives in redis.  These messages can be viewed by visiting https://echo.evantahler.com

Screenshot 2023-01-09 at 4 37 26 PM

evantahler:
@davinchia in lieu of a PR review, can you comment if you think the test plan outlined here is sufficient?

pedroslopez (changes requested):
👍🏻 ack - commenting to acknowledge the review request but definitely shouldn't merge 😄

davinchia:
@evantahler yes this is a decent sanity check. A sophisticated attacker can bypass this with DNS hijacking. The risk of that is low given we primary rely on Google's network (both for DNS servers and to route requests). Def good enough to check that request are routing to what 'we think' is the right endpoint.

davinchia:
I'm closing this to clean up PRs.

PR #21207, opened 2023-01-10
Remove excess get DestinationDefinition calls

krishnaglick:
## What
The destination definition was already available from the connection form service, no need to make the call again on the transformation view.

## How
The destination definition is now pulled out of the connection form service, instead of being loaded again.


edmundito (approved):
Did not test.

PR #21218, opened 2023-01-10
🪟 🐛 Hide no credits banner for no billing accounts

timroes:
## What

Hide the "We stopped all your syncs because you ran out of credits" for accounts that actually are still working (no billing accounts). They will still have creditStatus on "positive" even when ran out. There's not real way we know those accounts before they ran out of credit, so we keep showing the "Low credit warning" still, but at least hide the banner that tells we stopped their syncs (which would not happen).


joeykmh (approved):
I'm not sure I follow why  a workspace could have `remainingCredits <= 0` and simultaneously `CreditStatus.POSITIVE`? Are these special accounts where we are allowing negative credit balance but continuing syncs?

Approving because the change looks fine, I'm just missing some context.

PR #21219, opened 2023-01-10
🪟 🐛 Fix broken switch UI state - when checked state provided as a "value"

dizel852:
## What
Fixed broken `<Switch />` UI state

## How
Handle the `checked` state if it was provided as a `value`

CPT2301102049-572x571

### DIFF
--- a/airbyte-webapp/src/components/ui/Switch/Switch.tsx
+++ b/airbyte-webapp/src/components/ui/Switch/Switch.tsx
@@ -44,7 +44,7 @@ export const Switch: React.FC<SwitchProps> = ({
     <label className={labelStyle}>
       <input
         {...props}
-        aria-checked={indeterminate ? "mixed" : checked || !!value}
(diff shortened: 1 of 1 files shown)


airbyte-webapp/src/components/ui/Switch/Switch.tsx
```diff
@@ -44,7 +44,7 @@ export const Switch: React.FC<SwitchProps> = ({
     <label className={labelStyle}>
       <input
         {...props}
-        aria-checked={indeterminate ? "mixed" : checked}
+        aria-checked={indeterminate ? "mixed" : checked || !!value}
```
krishnaglick, line 47:
```suggestion
        aria-checked={(indeterminate ? "mixed" : checked) ?? !!value}
```
This feels slightly safer.

m19berger (approved):
LGTM!

krishnaglick (approved):

PR #21227, opened 2023-01-10
🪟 🧹 Shared SideBar component for Cloud and OSS

### LINKED ISSUE #13766: Create a single SideBar (main menu) component
There's a lot of repeated code between the cloud package [SideBar](https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/packages/cloud/views/layout/SideBar/SideBar.tsx) and the [OSS SideBar](https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/views/layout/SideBar/SideBar.tsx).  It makes updates a more cumbersome and prone to human error.

These can be combined such that cloud is able to use the OSS sidebar and pass in the additional components it needs rendered.

teallarson:
## What
closes #13766

Makes `<SideBar />` a reusable component rather than having two nearly identical but entirely separate SideBars for cloud and OSS.

## How

The `<SideBar />` component and its subcomponents are now rendered more declaratively.

Also:
- Adds a `<NavItem/>` component and `<NavDropdown/>` component to reduce repeated code and enable easier redesigns.
- Tears out as much .scss as possible by using `<Text/>` and `<Flex*/>` components

There some minor visual changes with spacing due to the flex layout components and spacing variables, but highlighting, fonts, icons, etc. should not be affected 

AND last but not least... fixes some redundant naming.
OSS is the base case here (MainView) and Cloud components are named as such (CloudMainView)

## Recommended reading order
1. `SideBar`
2. `MainNav`
3. `MainView`
4. `CloudMainView`
5. others


airbyte-webapp/jest.config.ts
```diff
@@ -13,7 +13,8 @@ const jestConfig: Config = {
   moduleDirectories: ["node_modules", "src"],
   moduleNameMapper: {
     "\\.module\\.scss$": "test-utils/mock-data/mockIdentity.js",
-    "\\.(css|png|svg|scss)$": "test-utils/mock-data/mockEmpty.js",
+    "\\.(css|png|scss)$": "test-utils/mock-data/mockEmpty.js",
+    "\\.svg": "test-utils/mock-data/mockSvg.js",
```
tealjulia, line 17: The old method wasn't working with svgs that we import using `import { ReactComponent as...`

Solved via the recommended solution from SVGR: https://react-svgr.com/docs/jest/

airbyte-webapp/src/views/layout/SideBar/SideBar.tsx
```diff
@@ -36,36 +36,99 @@ export const useCalculateSidebarStyles = () => {
...
+    <NavLink className={navLinkClassName} to={RoutePaths.Settings}>
+      <React.Suspense fallback={null}>
+        <NotificationIndicator />
+      </React.Suspense>
+      <SettingsIcon />
+      <Text className={styles.text} size="sm">
+        <FormattedMessage id="sidebar.settings" />
+      </Text>
+    </NavLink>,
+  ];
```
tealjulia, line 98: I debated wether to create an `OSSSideBar` component that passes these in (as a parallel to `CloudSideBar`) to keep the main component clean... on the fence as to whether that's warranted.  

It seems ok to me to treat OSS as the base case like this but I could be convinced otherwise.

airbyte-webapp/jest.config.ts
timroes: Minor nitpick:

```suggestion
    "\\.(css|png|scss)$": "test-utils/mock-data/mockEmpty.js",
    "\\.svg$": "test-utils/mock-data/mockSvg.js",
```

airbyte-webapp/src/views/layout/MainView/MainView.tsx
```diff
@@ -5,7 +5,8 @@ import { LoadingPage } from "components";
 
 import { ResourceNotFoundErrorBoundary } from "views/common/ResorceNotFoundErrorBoundary";
 import { StartOverErrorView } from "views/common/StartOverErrorView";
-import SideBar from "views/layout/SideBar";
+
+import { SideBar } from "../SideBar/SideBar";
```
josephkmh, line 9: nit: 
```suggestion
import { SideBar } from "../SideBar";
```

airbyte-webapp/src/packages/cloud/views/layout/SideBar/CloudSideBar.tsx
```diff
@@ -0,0 +1,143 @@
...
+import { links } from "utils/links";
+import ChatIcon from "views/layout/SideBar/components/ChatIcon";
+import SettingsIcon from "views/layout/SideBar/components/SettingsIcon";
+import StatusIcon from "views/layout/SideBar/components/StatusIcon";
+import { NotificationIndicator } from "views/layout/SideBar/NotificationIndicator";
+import { SideBar, useCalculateSidebarStyles } from "views/layout/SideBar/SideBar";
+
+import styles from "./CloudSideBar.module.scss";
+import { LOW_BALANCE_CREDIT_TRESHOLD } from "../../credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint";
+import { WorkspacePopout } from "../../workspaces/WorkspacePopout";
```
josephkmh, line 30: nit: I prefer making import paths explicit. This makes it easier to know where something comes from, and makes moving this file easier, since there are no relative paths.

```suggestion
import { LOW_BALANCE_CREDIT_TRESHOLD } from "packages/cloud/views/credits/CreditsPage/components/LowCreditBalanceHint/LowCreditBalanceHint";
import { WorkspacePopout } from "packages/cloud/views/workspaces/WorkspacePopout";
```

airbyte-webapp/src/views/layout/SideBar/SideBar.tsx
```diff
@@ -36,36 +36,99 @@ export const useCalculateSidebarStyles = () => {
   return ({ isActive }: { isActive: boolean }) => menuItemStyle(isActive);
 };
 
-const SideBar: React.FC = () => {
+interface SideBarProps {
+  additionalTopItems?: JSX.Element;
+  bottomMenuItems?: JSX.Element[];
```
josephkmh, line 41: What about using `JSX.Element` for both of these? You could wrap `bottomMenuItems` in a react fragment to avoid needing to loop over an array of elements manually.

joeykmh (changes requested):
Just left a few comments, but generally looks good! This reduces a lot of duplicate code.

airbyte-webapp/src/views/layout/SideBar/SideBar.tsx
tealjulia: That makes sense!

airbyte-webapp/src/views/layout/MainView/MainView.tsx
```diff
@@ -5,7 +5,8 @@ import { LoadingPage } from "components";
 
 import { ResourceNotFoundErrorBoundary } from "views/common/ResorceNotFoundErrorBoundary";
 import { StartOverErrorView } from "views/common/StartOverErrorView";
-import SideBar from "views/layout/SideBar";
+
+import { SideBar } from "./SideBar";
```
tealjulia, line 9: not entirely sure why my linter wants this now...

airbyte-webapp/src/views/layout/SideBar/components/MenuContent.tsx
```diff
@@ -0,0 +1,19 @@
+import { PropsWithChildren } from "react";
+
+import { FlexContainer } from "components/ui/Flex";
+
+import styles from "./MenuContent.module.scss";
+
+export const MenuContent: React.FC<PropsWithChildren<unknown>> = ({ children }) => {
```
tealjulia, line 7: essentially just a wrapper for a full-height full-width flex container so I wasn't constantly overriding in separate files with the same className.  

Does make me wonder whether we want that as a prop on the `FlexContainer` itself... but _really_ trying to watch scope creep at this point as this PR is getting gnarly.

airbyte-webapp/src/views/layout/SideBar/components/NavDropdown.module.scss
```diff
@@ -0,0 +1,26 @@
+@use "scss/variables";
+@use "scss/colors";
+@forward "./NavItem.module.scss";
+
+.dropdownMenuButton {
```
tealjulia, line 5: TIL that you can't use a functional component as the base for these dropdowns (so cannot use a `NavItem` and get rid of this, though I do reuse some of the classnames to reduce redundancy)

airbyte-webapp/jest.config.ts
```diff
@@ -13,7 +13,8 @@ const jestConfig: Config = {
   moduleDirectories: ["node_modules", "src"],
   moduleNameMapper: {
     "\\.module\\.scss$": "test-utils/mock-data/mockIdentity.js",
-    "\\.(css|png|svg|scss)$": "test-utils/mock-data/mockEmpty.js",
+    "\\.(css|png|scss)$": "test-utils/mock-data/mockEmpty.js",
+    "\\.svg$": "test-utils/mock-data/mockSvg.js",
```
tealjulia, line 17: now we can test components using svg's imported using `import { ReactComponent as...` 🎉

airbyte-webapp/src/views/layout/SideBar/components/NavDropdown.module.scss
josephkmh: I don't understand why we can't use functional components for `DropdownMenu` button... sunk 30 mins into it, but did not figure it out. This works fine too 🙂

joeykmh (commented):
Looking good, just left a few comments related to CSS/syntax

airbyte-webapp/src/views/layout/SideBar/components/MenuContent.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <FlexContainer
+      direction="column"
+      gap="xs"
+      alignItems="center"
+      justifyContent="space-between"
+      className={styles.menuContent}
```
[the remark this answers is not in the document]
tealjulia: Ooh!  I missed this when I migrated over to a "MenuContent" component to cut down on forwarded scss.  Big oops.

airbyte-webapp/src/views/layout/SideBar/components/MenuContent.module.scss
```diff
@@ -0,0 +1,5 @@
+.menuContent {
+  height: 100%;
+  padding: 0;
+  width: 100%;
```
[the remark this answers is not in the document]
tealjulia: Maybe?  I tried it and it adds quite a bit more repeated code adding `<FlexContainer>`s, and there were still a few places I had to add a new stylesheet just to add this same "reset"... so keeping it encapsulated in one reusable component seems to make sense.

tealjulia:
@josephkmh @ambirdsall one piece of work that is outstanding here is to get rid of `/components/`, but that will ruin the diff.  I would like to make that a separate PR, unless one of you feels strongly against it.

airbyte-webapp/src/packages/cloud/views/layout/CloudMainView/CloudMainView.module.scss
```diff
@@ -1,10 +1,47 @@
...
+      margin-top: 50px;
+
+      .dataBlock {
+        height: calc(100% - 50px);
+      }
+    }
+  }
+}
+
 .workspaceButton {
```
josephkmh, line 45: I think we'll need a width definition here to avoid overflowing:

image

joeykmh (changes requested):
A couple styles are off, almost there though!

ambirdsall (commented):
I do wonder if we need the extra layer of indirection from bundling up subsets of the nav items in `MainNavItems` and `BottomItems`: if the different conventional nav items are wrapped in their own components the way `ResourcesDropdown` is, then the cloud and OSS sidebars will read like tables of contents in a way that feels (subjectively) simple and obvious:
```tsx
// OSS
<SideBar>
  <MenuContent>
    <div className="topItems">
      <AirbyteHomeLink />
      <ConnectionsLink />
      <SourcesLink />
      <DestinationsLink />
    </div>
    <div className="bottomItems">
      <UpdateLink />
      <ResourcesDropdown />
      <SettingsLink>
      {version && <Version primary />}
    </div>
  </MenuContent>
</SideBar>

// cloud
<SideBar>
  <MenuContent>
    <div className="topItems">
      <AirbyteHomeLink />
      <WorkspacePopout>{/* redacted */}</WorkspacePopout>
      <ConnectionsLink />
      <SourcesLink />
      <DestinationsLink />
    </div>
    <div className="bottomItems">
      <CreditsLink />
      <CloudResourcesDropdown />
      <CloudSupportDropdown />
      <CloudSettingsDropdown />
    </div>
  </MenuContent>
</SideBar>
```

A side bonus of using the simpler flat structure would be that the `MenuContent` doesn't really need to exist as a separate component, since each version of the sidebar only needs a single flex container to put `space-between` the top and bottom link groups.

Any thoughts on that idea, @tealjulia? Besides the clear and present fact that it's a bit obnoxious to swoop in and suggest big structural changes after a PR has been open for so long :sweat_smile:

edmundito (commented):
I had some comments and meant to post them earlier. Whoops.

airbyte-webapp/src/views/layout/SideBar/components/MenuContent.tsx
tealjulia: It does seem like a common use case, however when I looked through our code I only saw 2 places we were sending a classname with a width or height of 100% via a `className` prop... so I am on the fence...

airbyte-webapp/src/packages/cloud/views/layout/CloudMainView/CloudMainView.module.scss
```diff
@@ -1,16 +1,53 @@
...
   overflow: hidden;
   cursor: pointer;
   display: block;
   height: 21px;
-  width: 100%;
+  width: 85px;
```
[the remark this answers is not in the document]
tealjulia: That does not appear to solve it there, but let me see if there's somewhere else on the chain that it might work.

airbyte-webapp/src/packages/cloud/views/layout/CloudMainView/CloudBottomItems.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { CloudResourcesDropdown } from "./CloudResourcesDropdown";
+import { CloudSupportDropdown } from "./CloudSupportDropdown";
+
+export const CloudBottomItems: React.FC = () => {
+  const workspace = useCurrentWorkspace();
+  const cloudWorkspace = useGetCloudWorkspace(workspace.workspaceId);
+  const isAllowUpdateConnectorsEnabled = useFeature(FeatureItem.AllowUpdateConnectors);
+
+  return (
+    <ul>
```
josephkmh, line 24: Noticed while reading Edmundo's comment - `<div>` is actually not a valid child element of a `<ul>`. It needs to contain only `<li>`. Could we make `<MenuContent>` a `<ul>` to avoid this? It might mean we can't use the `<FlexContainer>` 🤷

airbyte-webapp/src/packages/cloud/views/layout/CloudMainView/CloudBottomItems.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <MenuContent>
```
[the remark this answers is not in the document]
josephkmh: I'm not sure I follow here - what's the benefit of mapping over the children like this? If it's just to avoid typing out 4x `<li>` I'm not sure the indirection is worth it.

airbyte-webapp/src/packages/cloud/views/layout/CloudMainView/CloudBottomItems.tsx
tealjulia: Oh!  Good catch!
1. Yes with a tad of reworking
2. This kind of thing seems (valid html children) like it should be lintable but I couldn't find anything on it

tealjulia: 3. as long as I'm tackling that, it makes sense to add the children.map like Edmundo outlined (since I'm having to switch up the one place I was using this as a div anyways)

airbyte-webapp/src/views/layout/SideBar/components/NavDropdown.module.scss
josephkmh: Thanks for the explanation! I knew something funky was going on with that render prop pattern, but I couldn't figure it out. That makes sense!

airbyte-webapp/src/packages/cloud/views/layout/CloudMainView/CloudBottomItems.tsx
tealjulia: addressed in https://github.com/airbytehq/airbyte/pull/21227/commits/23f818a2bb790415023f030d343e9bbe14d11546

airbyte-webapp/src/packages/cloud/views/layout/CloudMainView/CloudBottomItems.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { RoutePaths } from "pages/routePaths";
+import { useCurrentWorkspace } from "services/workspaces/WorkspacesService";
+import { MenuContent } from "views/layout/SideBar/components/MenuContent";
+import { NavItem } from "views/layout/SideBar/components/NavItem";
+import SettingsIcon from "views/layout/SideBar/components/SettingsIcon";
+
+import { CloudResourcesDropdown } from "./CloudResourcesDropdown";
+import { CloudSupportDropdown } from "./CloudSupportDropdown";
+
+export const CloudBottomItems: React.FC = () => {
```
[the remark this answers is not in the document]
tealjulia: removed in https://github.com/airbytehq/airbyte/pull/21227/commits/23f818a2bb790415023f030d343e9bbe14d11546

airbyte-webapp/src/packages/cloud/views/layout/CloudMainView/CloudBottomItems.tsx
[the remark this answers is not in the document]
tealjulia: resolved in https://github.com/airbytehq/airbyte/pull/21227/commits/23f818a2bb790415023f030d343e9bbe14d11546

airbyte-webapp/src/packages/cloud/views/layout/CloudMainView/CloudBottomItems.tsx
tealjulia: addressed in https://github.com/airbytehq/airbyte/pull/21227/commits/23f818a2bb790415023f030d343e9bbe14d11546

edmundito (commented):
Code changes look good. ✅ Did not test. Will defer approval.

airbyte-webapp/src/views/layout/SideBar/AirbyteHomeLink.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { PropsWithChildren } from "react";
+import { useIntl } from "react-intl";
+import { Link } from "react-router-dom";
+
+import { RoutePaths } from "pages/routePaths";
+
+import styles from "./AirbyteHomeLink.module.scss";
+import { ReactComponent as AirbyteLogo } from "./airbyteLogo.svg";
+
+export const AirbyteHomeLink: React.FC<PropsWithChildren<unknown>> = () => {
```
josephkmh, line 10: nit: this component doesn't actually have children
```suggestion
export const AirbyteHomeLink: React.FC = () => {
```

airbyte-webapp/src/views/layout/SideBar/components/MenuContent.module.scss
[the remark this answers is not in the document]
josephkmh: I was just thinking of adding `display: flex;` to the existing stylesheet (not using `<FlexContainer>`, but either way is fine for me 👍

airbyte-webapp/src/views/layout/SideBar/components/MenuContent.tsx
```diff
@@ -0,0 +1,13 @@
+import { PropsWithChildren } from "react";
+
+import styles from "./MenuContent.module.scss";
+
+export const MenuContent: React.FC<PropsWithChildren<{ children: React.ReactNode[] }>> = ({ children }) => {
+  return (
+    <ul className={styles.menuContent}>
+      {children.map((child, index) => {
```
josephkmh, line 8: Not a huge fan of this approach, since overriding the `children` property like this means this component can't be used with a single child, which seems weird. We could also remove the `PropsWithChildren` helper, since we are overriding the `children` type it provides anyway.

airbyte-webapp/src/views/layout/SideBar/components/NavItem.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const isChild = location.pathname.split("/").length > 4 && location.pathname.split("/")[3] !== "settings";
+
+    return classNames(styles.menuItem, className, {
+      [styles.active]: isActive,
+      [styles.activeChild]: isChild && isActive,
+    });
+  };
+
+  return ({ isActive }: { isActive: boolean }) => menuItemStyle(isActive);
+};
```
josephkmh, line 50: Any reason to extract this into a separate hook? It's only being used one place

joeykmh (approved):
Approving with a few non-blocking comments, so you can decide whether to address them or not. Feel free to tag me for a re-review if you do. Nice work!

airbyte-webapp/src/views/layout/SideBar/components/MenuContent.tsx
tealjulia: Fair enough!  I think it's better to keep the helper as it sounds important for an eventual React18 upgrade.  Switched things up so it supports one or more children here.

josephkmh: I think for the React 18 upgrade it's only important that `children` has _some_ sort of type, so it would be fine to remove here. If you get rid of it completely (just use `React.FC`) you'll notice that `children` still has the correct type:

Screenshot 2023-02-13 at 9 43 15 PM

In React 18 this ^ will not work, since the implicit typing of `children` has been removed. But since you are defining it manually anyway, no need to use the `PropsWithChildren` helper type here (under the hood, it's just `type PropsWithChildren<P> = P & { children?: ReactNode | undefined };`) More info here if you're interested! https://solverfox.dev/writing/no-implicit-children/

tealjulia: Got it!  I instead kept `PropsWithChildren` and added an array check before the map... which I think should solve it.

airbyte-webapp/src/views/layout/SideBar/MainNavItems.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+      <NavItem
+        label={<FormattedMessage id="sidebar.destinations" />}
+        icon={<DestinationIcon />}
+        testId="destinationsLink"
+        to={RoutePaths.Destination}
+      />
+      {showBuilderNavigationLinks && (
+        <NavItem label="sidebar.builder" icon={<BuilderIcon />} testId="builderLink" to={RoutePaths.ConnectorBuilder} />
+      )}
```
tealjulia, line 40: @lmossman and @flash1293 I saw you recently added this SideBar item for connector builder.  I just wanted to call out that it is moving to this shared component in this refactor!

joeykmh (approved):

octavia-approvington:

octavia-approvington (approved):

PR #21230, opened 2023-01-10
🪟 🔧 Use experiment for new streams table design

### LINKED ISSUE #20196: Add webapp feature flag for new table redesign
Update feature flagging to use LaunchDarkly and the environment variable (for local testing.)

edmundito:
## What
Resolves #20196 

Updates flagging for new table design to use LaunchDarkly experiment.

## How
Creates a helper hook that checks for the experiment or the current env variable. While there is work to make this easier, I decided to use this custom solution to enable this in cloud sooner.

Some changes were made to how the `validationSchema` was created to use the hook. Creating the schema validation is also a hook so that the flags and experiments don't need to be passed directly.

`tidyConnectionFormValues` also now takes the generated validation schema instead of re-generating it.

## Recommended reading order
Top top bottom

### DIFF
--- a/airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionReplicationTab.tsx
+++ b/airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionReplicationTab.tsx
@@ -16,7 +16,6 @@ import {
   tidyConnectionFormValues,
   useConnectionFormService,
 } from "hooks/services/ConnectionForm/ConnectionFormService";
-import { FeatureItem, useFeature } from "hooks/services/Feature";
 import { useModalService } from "hooks/services/Modal";
 import { useConnectionService, ValuesProps } from "hooks/services/useConnectionHook";
 import { useCurrentWorkspaceId } from "services/workspaces/WorkspacesService";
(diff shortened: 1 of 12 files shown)


airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionReplicationTab.tsx
```diff
@@ -59,7 +54,7 @@ export const ConnectionReplicationTab: React.FC = () => {
   const { connection, schemaRefreshing, schemaHasBeenRefreshed, updateConnection, discardRefreshedSchema } =
     useConnectionEditService();
   const { initialValues, mode, schemaError, getErrorMessage, setSubmitError } = useConnectionFormService();
-  const allowSubOneHourCronExpressions = useFeature(FeatureItem.AllowSyncSubOneHourCronExpressions);
+  const validationSchema = useConnectionValidationSchema({ mode });
```
krishnaglick, line 57: Since this is done both here and in the create form I think this should be moved to the Connection Form Service.

krishnaglick (approved):
LGTM, did not test locally but since this is mostly a swap-in-place I expect tests and build to catch any errors.

Issue #17398, opened 2022-09-29
Autogenerate low-code reference docs for all components


girarda:
# Problem
As a user writing or using YAML, I want to comprehensively see all the components I can use and the documentation of their parameters. 

# Proposed Solution

#21231 added a link to that docs 

But it would be a better experience to have a CI hook which embeds this into the docs directly, using something like https://coveooss.github.io/json-schema-for-humans -- whatever we use should be easy to navigate and should feel like. using a javadoc or an autogenerated sphinx reference doc.

girarda:
closing because the manifest docs is generated

PR #21231, opened 2023-01-10
Lowcode docs: add link to YAML reference

sherifnada:
## What
Adds a link to the YAML reference to the docs so users can always see the full options available. 

In the future, it would be a better experience to have a CI hook which embeds this into the docs directly, using something like https://coveooss.github.io/json-schema-for-humans. 

in the interest of "shipped is better than perfect", putting this here as a stopgap.


lmossman (approved):
Looks fine to me for now. Want to create an issue for embedding the json in the docs directly?

sherifnada:
updated this issue's description https://github.com/airbytehq/airbyte/issues/17398

PR #21235, opened 2023-01-10
Bmoric/restore old interface

benmoriceau:
## What
After bumping that oss version that contains the micronaut migration of the oss server, we ran into issue such as ```A provider io.airbyte.server.errors.InvalidInputExceptionMapper registered in SERVER runtime does not implement any provider interfaces applicable in the SERVER runtime. Due to constraint configuration problems the provider io.airbyte.server.errors.InvalidInputExceptionMapper will be ignored. ```

In order to avoid having those mapper to not be present, we need to re-introduce the old mappers even if they are not being used in the OSS project anymore.


colesnodgrass (commented):
This PR is also making changes to existing files and is not just re-introducing removed files.  Is there a way to differentiate between the two cases or do we need to undo all of the Micronaut Mapper changes?

benmoriceau:
> This PR is also making changes to existing files and is not just re-introducing removed files. Is there a way to differentiate between the two cases or do we need to undo all of the Micronaut Mapper changes?

The `Mapper` are restoring what they used to be before migration and the micronaut migration now goes into handlers. The file were not remove but re-implemented :/

colesnodgrass (approved):

PR #21242, opened 2023-01-11
update: new input for source commcare

siddhant3030:
## What
Adding the new optional input field for the Commcare connector

## Why

This field is an array that takes multiple fields to remove it from each data and pulls the limited dataset for the Form stream.


siddhant3030:
@marcosmarxm @sajarin Need a review on this PR.

airbyte-integrations/connectors/source-commcare/source_commcare/spec.yaml
```diff
@@ -36,3 +36,8 @@ connectionSpecification:
       description: >-
         UTC date and time in the format 2017-01-25T00:00:00Z. Only records after this date will be replicated.
       order: 3
+    form_fields:
```
marcosmarxm, line 39:
```suggestion
    form_fields_to_exclude:
```

airbyte-integrations/connectors/source-commcare/source_commcare/spec.yaml
```diff
@@ -36,3 +36,8 @@ connectionSpecification:
...
+      type: array
+      title: Unwanted Fields
```
marcosmarxm, line 41:
```suggestion
      title: Form Fields to be Excluded
```

airbyte-integrations/connectors/source-commcare/source_commcare/source.py
```diff
@@ -34,15 +35,22 @@ def url_base(self) -> str:
...
     schemas = {}
     unwantedfields = re.compile(r"^(case_|update_|meta|create_|commcare_).*$")
-
+  
     @property
     def dateformat(self):
         return "%Y-%m-%dT%H:%M:%S.%f"
 
     def scrubUnwantedFields(self, form):
-        newform = {k: v for k, v in form.items() if not self.unwantedfields.match(k)}
```
marcosmarxm, line 43: Why are you removing the system excluding fields? Please keep this as is today.

airbyte-integrations/connectors/source-commcare/source_commcare/source.py
```diff
@@ -34,15 +35,22 @@ def url_base(self) -> str:
...
-
+        new_dict = {}
+        for key, value in form.items():
+            if key in self.form_fields:
+             continue
+            if isinstance(value, dict):
+             new_dict[key] = self.scrubUnwantedFields(value)
+            else:
+                new_dict[key] = value
+        return new_dict
```
marcosmarxm, line 52: Here you're keeping the values from the `form_fields` which should be deleted right?

marcosmarxm (changes requested):
Some comments. Can you give an example of fields you want to remove from records?

airbyte-integrations/connectors/source-commcare/source_commcare/source.py
siddhant3030: We don't need this. The idea is to not pull all the data from forms. and it was hard coded so I have made it dynamic and the user can choose which fields he doesn't want. It pulls a lot of unnecessary data.

airbyte-integrations/connectors/source-commcare/source_commcare/source.py
siddhant3030: This is why we changed to this. So it can be empty or user can choose which fields he doesn't want.

airbyte-integrations/connectors/source-commcare/source_commcare/source.py
marcosmarxm: Yes, but the previous version is already removing the system fields. In your suggestion the user will must define ALL system fields to be removed.

siddhant3030: It will be an optional field. Either they can pull all the data or remove some of the fields. It works with both

siddhant3030: @marcosmarxm I have made some changes that you suggested. The idea is to excludes some form_fields and it can be optional
Screenshot 2023-01-31 at 1 22 54 PM
here users can add the fields which they don't want to pull. So earlier it was hardcoded we've just made it dynamic.

PR #21244, opened 2023-01-11
🐛 🎉 Source Amplitude:  additional `len` check for `series` for `ActiveUsers` stream

### LINKED ISSUE #15208: Amplitude: IndexError when reading ActiveUsers stream
## Current Behavior
The Amplitude connectors is sometimes throwing an index out of bound error when zipping the time series in the ActiveUsers stream.
https://sentry.io/organizations/airbytehq/issues/3418815568

According to the docs at https://www.docs.developers.amplitude.com/analytics/apis/dashboard-rest-api/#response_1 , the length of xValues and the length of an element of the series array should be the same, but the issue occurred twice over the last month so it would be worth handling it better.

The root cause is unknown because I couldn't reproduce it with our test account. It's also possible that this is caused by issues on Amplitude's end.

Created this issue so we don't lose track of it, but I don't think it's high priority because it's only intermittent.

bazarnov:
## What
Resolving: 
- https://github.com/airbytehq/airbyte/issues/15208

## How
- enabled  `high` strictness for SAT, with bypassed  `basic_read`, please read the reason in the test config
- added an additional `len` check for the `series` object for the `ActiveUsers`  stream, as it can return `None` instead of `[]`, which causes the `Out Of Bound` issues while yielding records.

## 🚨 User Impact 🚨
No impact is expected.

## Pre-merge Checklist

- [X] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [X] Code reviews completed
- [X] Documentation updated
    - [X] Connector's `README.md`
    - [X] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - [X] Changelog updated in `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)

### Airbyter

- [ ] Build is successful

</details>


arsenlosenko (approved):

airbyte-integrations/connectors/source-amplitude/acceptance-test-config.yml
```diff
@@ -1,24 +1,35 @@
...
   incremental:
-    - config_path: "secrets/config.json"
-      configured_catalog_path: "integration_tests/configured_catalog_without_events.json"
-      # Unable to use 'state_path' because Amplitude returns an error when specifying a date in the future.
-      # future_state_path: "integration_tests/abnormal_state.json"
+    tests:
+      - config_path: "secrets/config.json"
+        configured_catalog_path: "integration_tests/configured_catalog_without_events.json"
+        future_state:
+          bypass_reason: "Unable to use 'state_path' because Amplitude returns an error when specifying a date in the future."
```
alafanechere, line 31: Can we handle this error and make the test pass please?

airbyte-integrations/connectors/source-amplitude/acceptance-test-config.yml
```diff
@@ -1,24 +1,35 @@
...
-      configured_catalog_path: "integration_tests/streams_with_output_records_catalog.json"
+    tests:
+      - config_path: "secrets/config.json"
+        expect_records:
+          path: "integration_tests/expected_records.txt"
+        empty_streams:
+          - name: "events"
+            bypass_reason: "The stream is too large, skipped."
+          - name: "annotations"
+            bypass_reason: "Stream has no records"
```
alafanechere, line 25: What'd be the effort to seed this stream with data?

airbyte-integrations/connectors/source-amplitude/acceptance-test-config.yml
```diff
@@ -1,24 +1,35 @@
...
   basic_read:
-    - config_path: "secrets/config.json"
-      configured_catalog_path: "integration_tests/streams_with_output_records_catalog.json"
+    tests:
+      - config_path: "secrets/config.json"
+        expect_records:
+          path: "integration_tests/expected_records.txt"
+        empty_streams:
+          - name: "events"
+            bypass_reason: "The stream is too large, skipped."
```
alafanechere, line 23: What do you mean by "its too large"? The SAT run is too long on this stream? Could we reduce the volume of events on the test account to test this stream?

alafanechere (changes requested):

airbyte-integrations/connectors/source-amplitude/acceptance-test-config.yml
bazarnov: Will check.

airbyte-integrations/connectors/source-amplitude/acceptance-test-config.yml
bazarnov: "Too large" means the live account generates millions of thousands of records/hour (lol). In fact, in order to sync for at least 1 day, we should spend almost 2 hours for a single stream (which is at least 300k+ records/slice),  where the slice is 24 hours, and the minimum required is 3 hours slice for API to accept the call.
I believe we cannot reduce the number of records for this stream since this is a live account used for tests.

sherifnada:
removing extensibility since Augustin is reviewing the SAT portions

airbyte-integrations/connectors/source-amplitude/acceptance-test-config.yml
bazarnov: Done.

airbyte-integrations/connectors/source-amplitude/acceptance-test-config.yml
bazarnov: Done.

airbyte-integrations/connectors/source-amplitude/acceptance-test-config.yml
```diff
@@ -1,24 +1,33 @@
...
   basic_read:
-    - config_path: "secrets/config.json"
-      configured_catalog_path: "integration_tests/streams_with_output_records_catalog.json"
+    tests:
+      - config_path: "secrets/config.json"
+        expect_records:
+          path: "integration_tests/expected_records.txt"
+        empty_streams:
+          - name: "events"
+            bypass_reason: "The stream is too large and frequently changing, due to test with live account."
```
alafanechere, line 23: We must not use a live account! 
(cc @evantahler)

alafanechere (commented):
Thank you for the changes that allowed less test bypass.
We should not use a live account for testing. 
I thought the credentials we have were using our integration-test email address. This is why I'm surprised this is a live account used for testing.
I created this [issue](https://github.com/airbytehq/airbyte/issues/21320) to create a sandbox account if the current one is indeed a live account.

airbyte-integrations/connectors/source-amplitude/acceptance-test-config.yml
bazarnov: Please follow this link: https://analytics.amplitude.com/airbyteinc/home to check out for yourself, whether or not we use the live account or test one.
Use the LastPass creds for `integration-test@airbyte.io`

bazarnov: @alafanechere 
Are we freezing this one until we can use the sandbox?

alafanechere: I confirm the Amplitude account is a live account for airbyte on which we have the `integration-test-@airbyte.io` user.
I believe the `expected_records.json` file is not containing sensitive data at the moment but I'd like a double-check from @evantahler . 
In the short term, I think it's OK to keep the live account for testing but in long run, the test might fail after unrelated actions on the Amplitude account.

evantahler: Access has been removed

evantahler: Information about the test account has been removed form  LastPass and GSM. SAT will fail until a new sandbox account & secrets are added.

evantahler (changes requested):
Please update the tests to use a new sandbox account

bazarnov:
@evantahler @alafanechere @arsenlosenko 

The https://github.com/airbytehq/airbyte/issues/19036 is updated to the sandbox and would be released within this PR:
https://github.com/airbytehq/airbyte/pull/23192. 

This PR will keep tracking the other stream issues it fixes.

airbyte-integrations/connectors/source-amplitude/acceptance-test-config.yml
alafanechere, line 23: Do we still need to declare `events` as an empty stream now that we have a dedicated sandbox account?

alafanechere (changes requested):
Requesting change to declare the events stream as not empty now that we have a sandbox account for Amplitude.

bazarnov:
> Requesting change to declare the events stream as not empty now that we have a sandbox account for Amplitude.

@alafanechere please check this comment: https://github.com/airbytehq/airbyte/pull/21244#issuecomment-1434545493 there is no need to request changes for SAT in this PR.

PR #19344, opened 2022-11-11
🪟 🔧  Use CSS Custom Properties for colors

joeykmh:
## What
Converts our SASS color variables from using hex codes to using CSS custom properties

## How
- Adds `_customProperties.scss`
- Adjusts `_colors.scss` to use CSS Custom Properties instead of hex codes
- Removes `colors.$blue-transparent` in [this commit](https://github.com/airbytehq/airbyte/pull/19344/commits/2498effa82348659adb4e4f3d0e8b7016c9b859c) to avoid SVG  `<linear-gradient>` choking on this syntax: `rgba(--var-color-blue, .1)`
- Minor adjustments to [CodeEditor](https://github.com/airbytehq/airbyte/pull/19344/commits/b209bc7c2c0771ce77527ef3b627c18d7e793d9e) to make it work with custom props
- Removes transparent color variants in the `Tooltip` and `TooltipTable` components
- Adds box-shadows to the theme/SCSS variables

There are a few changes in the appearance of the tooltips specifically. They are not longer slightly transparent, and are now instead fully opaque. Instead of the inline transparent values, I used colors from our theme instead. There may be very minor differences, but they should be barely noticeable:

Before:
image
image
image
image

After:
image
image
image
image

## Recommended reading order
Top to bottom

PR #21246, opened 2023-01-11
🪟 🐛 Fix auto detect schema change backdrop color CSS

edmundito:
## What
Fixes a CSS rule that prevents rendering the background that covers the replication settings when schema changes have been detected.

Caused by #19344

## How
`rgba(var(--color-name), 50%)` is not supported, so the color must be extracted into the RGB values first.

### DIFF
--- a/airbyte-webapp/src/views/Connection/ConnectionForm/components/SchemaChangeBackdrop.module.scss
+++ b/airbyte-webapp/src/views/Connection/ConnectionForm/components/SchemaChangeBackdrop.module.scss
@@ -16,7 +16,7 @@
     @extend %cover;
 
     position: absolute;
-    background: linear-gradient(180deg, rgba(colors.$grey-50, 0.5) 0%, colors.$white 92.54%);
+    background: linear-gradient(180deg, rgba(248, 248, 250, 50% /* grey-50 50% */) 0%, colors.$white 92.54%);
     z-index: z-indices.$schemaChangesBackdrop;
     border-radius: variables.$border-radius-lg;
     display: flex;


airbyte-webapp/src/views/Connection/ConnectionForm/components/SchemaChangeBackdrop.module.scss
```diff
@@ -16,7 +16,7 @@
     @extend %cover;
 
     position: absolute;
-    background: linear-gradient(180deg, rgba(colors.$grey-50, 0.5) 0%, colors.$white 92.54%);
+    background: linear-gradient(180deg, rgba(248, 248, 250, 50% /* grey-50 50% */) 0%, colors.$white 92.54%);
```
krishnaglick, line 19: Fun bug

krishnaglick (approved):
LGTM, did not test locally

airbyte-webapp/src/views/Connection/ConnectionForm/components/SchemaChangeBackdrop.module.scss
josephkmh: Yeah using `rgba()` with color variables will no longer work. My suggestion is to treat opaque variants of theme colors as their own variables (see e.g. the overlay background color [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/scss/_theme-light.scss#L91)) or, if they're really one-off colors, to define them inline like [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/packages/cloud/views/auth/components/PersonQuoteCover/PersonQuoteCover.module.scss#L26-L32). 

In this case I'm not sure which is more appropriate, probably best to define it inline like you did. But basically I would decouple it from `grey-50`, since it's a different color.

PR #21253, opened 2023-01-11
Add tooltips to overflow text on streams table

### LINKED ISSUE #20450: All cells in streams/fields tables should have ellipses and tooltips for text overflow
Per design pass with Nico.  Any cell with truncated text should have a tooltip (including the dropdown for sync mode, destination namespace, etc.)

I do not believe this needs to be hidden behind the new table feature flag.

krishnaglick:
## What
Resolves #20450
On the new streams table, fields that may have text that overflows now have a tooltip that shows the full text.

## How
Adds a `withTooltip` property to the CatalogTreeTableCell component to enable the tooltip functionality.

https://www.loom.com/share/711bed4eb0434808a3f736cb861833ca


airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableHeader.tsx
```diff
@@ -80,9 +80,6 @@ export const CatalogTreeTableHeader: React.FC = () => {
       <HeaderCell size="small">
         <FormattedMessage id="sources.sync" />
       </HeaderCell>
-      {/* <TextCell>
```
tealjulia, line 83: Suppose this is fine to remove for now?  We do want to have this field once we introduce column selection to the new table (https://github.com/airbytehq/airbyte/pull/21058) we want this column.  Unless the designs have changed and I'm not up to date.

airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableRow.tsx
```diff
@@ -66,18 +66,17 @@ export const CatalogTreeTableRow: React.FC<StreamHeaderProps> = ({
       <CatalogTreeTableCell size="small">
         <Switch size="sm" checked={stream.config?.selected} onChange={onSelectStream} disabled={disabled} />
       </CatalogTreeTableCell>
-      {/* <Cell>{fieldCount}</Cell> */}
```
tealjulia, line 69: Same comment as above

airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableRow.tsx
```diff
@@ -66,18 +66,17 @@ export const CatalogTreeTableRow: React.FC<StreamHeaderProps> = ({
...
         </Text>
       </CatalogTreeTableCell>
-      <CatalogTreeTableCell>
+      <CatalogTreeTableCell withTooltip>
         <Text size="md" className={styles.cellText}>
           {stream.stream?.name}
         </Text>
       </CatalogTreeTableCell>
-      <CatalogTreeTableCell size="large">
+      <CatalogTreeTableCell size="large" withTooltip={disabled}>
```
tealjulia, line 79: This is one of the main reasons we're introducing the Tooltip -- the sync mode "incremental | deduped + history" always has overflow.  We want a Tooltip on it.

teallarson (changes requested):
So many tips!

A few notes:
- My understanding is we _only_ want the Tooltip if there is overflow, but maybe check in with Nico on that if you think I'm off there.
- I did not see a Tooltip on the cursor and primary key cells.
- I think all of our other Tooltips use a dark variant.  I'm not sure if we want dark or light here? No strong feeling.

Left a few other, more specific thoughts, too!

airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableRow.tsx
krishnaglick: So we want the value of the dropdown as a tooltip here?

tealjulia: Yeah!  That was the idea as I understood it.

krishnaglick:
>     * I think all of our other Tooltips use a dark variant.  I'm not sure if we want dark or light here? No strong feeling.


I thought light looked better, but honestly they're both fine. I'll go w/ dark since that's the default.

teallarson (approved):
Nice!  I did notice one case where if I resize the window, the text wraps, at least in this case (this is a source-defined primary key if the cell type matters).  I'm not sure whether we want to solve that in this PR or if it's something for another PR.

Screenshot 2023-01-17 at 12 30 53 PM

airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableHeader.tsx
krishnaglick: I just removed it since it had been sitting around. I can re-add 👍

tealjulia:
Super close!
Still not seeing ellipses on overflowing source-defined primary keys.
Screenshot 2023-01-19 at 2 34 42 PM

Also, would be nice to parse the actual character here!
Screenshot 2023-01-19 at 2 25 34 PM

teallarson (approved):
Tested locally and both of the issues from earlier were resolved 🎉

PR #21266, opened 2023-01-11
Run Airbyte CI on pull_request

c-p-b:
## What
Currently this only runs on push. This causes external contributor pull requests to never trigger even though they are required.

## How
Update the ci yaml to also trigger on pull request. Also make Airbyte CI only run on pushes to master instead of any branch to reduce amount of extraneous ci


.github/workflows/gradle.yml
```diff
@@ -19,8 +19,9 @@ on:
   schedule:
     - cron: "0 */1 * * *"
   push:
-    branches-ignore:
-      - "gitbook/v1"
+    branches:
+      - master
```
perangel, line 23: Given your concern about this having an adverse effect on other workflows, would you rather just revert this and just add the `pull_request` trigger?

cpdeethree: Actually, this being set up like this dates back to the very first commit of ci in airbyte. I think it's pretty safe to say that the original reason it was set up like that probably doesn't matter anymore: https://github.com/airbytehq/airbyte/commit/51920f229ab220ca1e6bd0be74e43db467cdf24f

perangel (approved):

PR #21269, opened 2023-01-11
🪟 🔧 Add stylelint plugin & no-color-variables-in-rgba rule

joeykmh:
## What
Recently we [merged a PR](https://github.com/airbytehq/airbyte/pull/19344) that makes using our SASS color variables incompatible with the `rgba()` function. Unfortunately there is nothing blocking a developer from using this syntax, it just silently fails. This PR introduces a stylelint plugin & rule, `no-color-variables-in-rgba`, which checks our source (S)CSS files for incompatible usage.

For example, if you write some CSS like this:

```scss
@use "scss/colors";

.myClass {
  background-color: rgba(colors.$blue, 50%);
```

Our pre-commit hook, which runs `stylelint`, will fail with this error message:

image

## Considerations
The rule checks using a regex `/rgba\([^)]*colors\.\$/` that looks specifically for `rgba(colors.$` to detect if the rule is violated. If we change the name of our `scss/colors` import for whatever reason, this could break.


airbyte-webapp/packages/stylelint-plugin/index.js
```diff
@@ -0,0 +1,12 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const stylelint = require("stylelint");
+
+const rules = {
+  "no-color-variables-in-rgba": require("./no-color-variables-in-rgba"),
+};
```
josephkmh, line 6: If we want to add more stylelint plugins later, we can put them in here

josephkmh:
Tested that this works on CI during the frontend build step as well: https://github.com/airbytehq/airbyte/actions/runs/3895729403/jobs/6651434662#step:8:1853

airbyte-webapp/packages/stylelint-plugin/no-color-variables-in-rgba.js
```diff
@@ -0,0 +1,35 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const stylelint = require("stylelint");
+const { ruleMessages } = stylelint.utils;
+const ruleName = "airbyte/no-color-variables-in-rgba";
+const messages = ruleMessages(ruleName, {
+  variableFoundInRgba: () => `A color variable should not be used within an rgba() function.`,
```
timroes, line 6: Maybe we should be more explicit here that we "can't" use them inside instead of just "shouldn't", since they basically won't work. I also wonder if we can reference to any place that explains why this is the case?

timroes (approved):
Tested locally, code LGTM. Just left a comment about potential change in wording, but non blocking for me.

airbyte-webapp/packages/stylelint-plugin/no-color-variables-in-rgba.js
josephkmh: Good point, updated to change to "can't".

As for an explanation, I thought about linking either to this PR in the error message (I could add some more context in the description) or to the `STYLEGUIDE.md`. Any preference? I would add a new section in the style guide, `## Styling`

timroes: I think as well the Styleguide might be the best place for now to put this and link to.

josephkmh: Cool, added!

PR #21271, opened 2023-01-11
🪟 🎨 Add focus state mixin to Button and Switch components

teallarson:
## What
Attempted to address airbytehq/airbyte-internal-issues#1170

However, the work has been re-scoped and the original intent of that issue is largely unaddressed here :) 

Instead... we now have a focus ring on our Switch and Button

## Note

During implementation, it was decided that the scope of this work is greater than previously considered and will require reworking how we handle nested interactive elements.  I am leaving the original issue open for that reason.

However, the focus-ring itself is still a useful feature and does not cause problems with the Button and Switch components as implemented.

## How
- adds a mixin that is utilized by Button and can be implemented elsewhere as well
- utilizes the styles directly in Switch due to some implementation differences there

## Recommended reading order
1. `_mixins.scss`
2.  others

### DIFF
--- a/airbyte-webapp/src/components/connection/CatalogTree/StreamHeader.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogTree/StreamHeader.tsx
@@ -95,14 +95,14 @@ export const StreamHeader: React.FC<StreamHeaderProps> = ({
     [styles.purpleBackground]: isSelected,
     [styles.redBorder]: hasError,
   });
-  const checkboxCellCustomStyle = classnames(styles.checkboxCell, { [styles.streamRowCheckboxCell]: true });
+  const checkboxCellCustomStyle = classnames(styles.checkboxCell, styles.streamRowCheckboxCell);
 
   return (
     <Row className={styles.catalogSectionRow}>
       {!disabled && (
         <div className={checkboxCellCustomStyle}>
           <div className={iconStyle}>{changedSelected && (isStreamEnabled ? <PlusIcon /> : <MinusIcon />)}</div>
-          <CheckBox checked={isSelected} onChange={selectForBulkEdit} />
+          <CheckBox checked={isSelected} className={styles.checkBox} onChange={selectForBulkEdit} />
         </div>
       )}
       <ArrowCell>
--- a/airbyte-webapp/src/components/connection/CatalogTree/StreamHeader.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogTree/StreamHeader.tsx
@@ -102,7 +102,7 @@ export const StreamHeader: React.FC<StreamHeaderProps> = ({
       {!disabled && (
         <div className={checkboxCellCustomStyle}>
           <div className={iconStyle}>{changedSelected && (isStreamEnabled ? <PlusIcon /> : <MinusIcon />)}</div>
-          <CheckBox checked={isSelected} className={styles.checkBox} onChange={selectForBulkEdit} />
+          <CheckBox checked={isSelected} onChange={selectForBulkEdit} />
         </div>
       )}
       <ArrowCell>
--- a/airbyte-webapp/src/components/connection/CatalogTree/StreamHeader.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogTree/StreamHeader.tsx
@@ -95,7 +95,7 @@ export const StreamHeader: React.FC<StreamHeaderProps> = ({
     [styles.purpleBackground]: isSelected,
     [styles.redBorder]: hasError,
   });
-  const checkboxCellCustomStyle = classnames(styles.checkboxCell, styles.streamRowCheckboxCell);
+  const checkboxCellCustomStyle = classnames(styles.checkboxCell, { [styles.streamRowCheckboxCell]: true });
 
   return (
     <Row className={styles.catalogSectionRow}>
--- a/airbyte-webapp/src/components/ui/Button/Button.module.scss
+++ b/airbyte-webapp/src/components/ui/Button/Button.module.scss
@@ -151,6 +151,10 @@
     &:active {
       background-color: colors.$blue-600;
     }
+
+    &:focus-visible {
+      outline-offset: 1px;
+    }
   }
 
   &.typeSecondary {
--- a/airbyte-webapp/src/components/ui/Button/Button.module.scss
+++ b/airbyte-webapp/src/components/ui/Button/Button.module.scss
@@ -1,7 +1,10 @@
 @use "scss/colors";
 @use "scss/variables";
+@use "scss/mixins";
 
 .button {
+  @include mixins.focus-ring;
+
   display: inline-flex;
   align-items: center;
   justify-content: center;
@@ -24,14 +27,9 @@
     pointer-events: none;
   }
 
-  &:focus-visible,
   &:hover,
   &:not(.typeLight) {
     box-shadow: 0 3px 4px -2px rgba(26, 25, 77, 21%);
-  }
-
-  &:focus-visible {
-    outline: 3px solid colors.$blue-50;
   }
 
   &:disabled:not(.isLoading) {
--- a/airbyte-webapp/src/components/ui/Button/Button.module.scss
+++ b/airbyte-webapp/src/components/ui/Button/Button.module.scss
@@ -151,10 +151,6 @@
     &:active {
       background-color: colors.$blue-600;
     }
-
-    &:focus-visible {
-      outline-offset: 1px;
-    }
   }
 
   &.typeSecondary {
--- a/airbyte-webapp/src/components/ui/CheckBox/CheckBox.module.scss
+++ b/airbyte-webapp/src/components/ui/CheckBox/CheckBox.module.scss
@@ -19,8 +19,7 @@ $sm-size: 14px;
   @include mixins.focus-ring;
 
   input {
-    position: absolute;
-    opacity: 0;
+    @include mixins.sr-only;
   }
 
   &.disabled {
--- a/airbyte-webapp/src/components/ui/CheckBox/CheckBox.module.scss
+++ b/airbyte-webapp/src/components/ui/CheckBox/CheckBox.module.scss
@@ -1,5 +1,6 @@
 @use "scss/colors";
 @use "scss/variables";
+@use "scss/mixins";
 
 $lg-size: 18px;
 $sm-size: 14px;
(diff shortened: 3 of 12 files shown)


airbyte-webapp/src/components/ui/CheckBox/CheckBox.module.scss
```diff
@@ -14,9 +15,10 @@ $sm-size: 14px;
   color: colors.$white;
   background-color: colors.$white;
   cursor: pointer;
+  @include mixins.focus-ring;
 
   input {
-    display: none;
+    opacity: 0;
```
tealjulia, line 21: display: none also hides the focus

airbyte-webapp/src/components/ui/Switch/Switch.module.scss
```diff
@@ -112,6 +114,10 @@
     width: 0;
     height: 0;
 
+    &:focus-visible ~ .slider {
+      @include mixins.focus-ring-style;
```
tealjulia, line 118: here's where it's a little funky.

I attempted adding an optional argument to the mixin where we could pass a class and therefore apply the mixin directly to .switch, but spiraled.  Can take another stab if there's a strong feeling that we should do that!

timroes: I think putting this inline here sounds like the better idea, than overengineering the mixin for potentially very rare use-cases :)

airbyte-webapp/src/components/ui/Button/Button.module.scss
```diff
@@ -153,6 +151,10 @@
     &:active {
       background-color: colors.$blue-600;
     }
+
+    &:focus-visible {
+      outline-offset: 1px;
```
tealjulia, line 156: For primary buttons, give a 1px offset so there is contrast.

Open to _always_ applying the offset, but the designs Nico has, they are hugging the element.

airbyte-webapp/src/components/ui/CheckBox/CheckBox.module.scss
```diff
@@ -15,8 +16,10 @@ $sm-size: 14px;
   background-color: colors.$white;
   cursor: pointer;
 
+  @include mixins.focus-ring;
+
   input {
-    display: none;
+    @include mixins.sr-only;
```
tealjulia, line 22: This makes our checkboxes hidden to the eye but visible to a screen-reader.

airbyte-webapp/src/components/ui/CheckBox/CheckBox.tsx
```diff
@@ -10,9 +10,10 @@ type CheckBoxSize = "lg" | "sm";
 export interface CheckBoxProps extends InputHTMLAttributes<HTMLInputElement> {
   indeterminate?: boolean;
   checkboxSize?: CheckBoxSize;
+  tabIndex?: number;
 }
 
-export const CheckBox: React.FC<CheckBoxProps> = ({ indeterminate, checkboxSize = "lg", ...inputProps }) => {
+export const CheckBox: React.FC<CheckBoxProps> = ({ indeterminate, checkboxSize = "lg", tabIndex, ...inputProps }) => {
```
tealjulia, line 16: This is specifically to allow us to remove the checkbox from the regular tabbing flow (ie: when it's inside of a dropdown :) )

airbyte-webapp/src/components/ui/Switch/Switch.module.scss
```diff
@@ -44,6 +45,7 @@
 .switch {
   position: relative;
   display: inline-block;
+  margin: variables.$spacing-xs;
```
tealjulia, line 48: so the focus-ring is visible

airbyte-webapp/src/components/ui/CheckBox/CheckBox.tsx
```diff
@@ -29,7 +30,13 @@ export const CheckBox: React.FC<CheckBoxProps> = ({ indeterminate, checkboxSize
         className
       )}
     >
-      <input type="checkbox" aria-checked={indeterminate ? "mixed" : checked} {...inputProps} />
+      <input
+        type="checkbox"
+        aria-checked={indeterminate ? "mixed" : checked}
+        {...inputProps}
+        onClick={(e) => e.stopPropagation()}
```
josephkmh, line 37: Looks like this is not working as expected, or does this PR need to be rebased/merge master? Right now clicking a checkbox is also triggering e.g. the stream panel to open.

https://user-images.githubusercontent.com/7550957/215805187-71b7e032-af64-4817-8311-a04a2b47c96a.mov

joeykmh (changes requested):
Left a few comments and noticed one problem with the click behavior on checkboxes ✔️

airbyte-webapp/src/scss/_mixins.scss
```diff
@@ -11,3 +12,28 @@
 @mixin right-shadow {
   box-shadow: variables.$box-shadow-right;
 }
+
+@mixin focus-ring-style {
+  outline: 1px solid colors.$blue;
```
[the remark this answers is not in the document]
tealjulia: fine by me!

airbyte-webapp/src/components/ui/Button/Button.module.scss
```diff
@@ -1,7 +1,10 @@
 @use "scss/colors";
 @use "scss/variables";
+@use "scss/mixins";
 
 .button {
+  @include mixins.focus-ring;
```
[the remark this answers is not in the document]
tealjulia: I agree.  Are you proposing there is a way to remove the transition from _only_ the outline?

tealjulia:
Thanks for the review!  Will take a look at the propagation. 

Yeah... Perhaps it would be good to revisit with Nico before continuing to move forward.  Difficult to create a focus-ring that meets accessibility standards (either is contrasting enough or has an offset).

airbyte-webapp/src/components/ui/CheckBox/CheckBox.tsx
tealjulia: Fun fact: `e.stopPropagation()` works here for keyboard nav but not mouse... and `event.nativeElement.stopImmediatePropagation()` works here for mouse but not keyboard.  Neat.
(Also TIL [nativeElement.stopImmediatePropagation()](https://gist.github.com/ggregoire/ce7bc946212920c0a6bad8125567d001)).  To be continued....

airbyte-webapp/src/components/connection/CatalogTree/StreamHeader.tsx
```diff
@@ -95,14 +95,14 @@ export const StreamHeader: React.FC<StreamHeaderProps> = ({
...
-  const checkboxCellCustomStyle = classnames(styles.checkboxCell, { [styles.streamRowCheckboxCell]: true });
+  const checkboxCellCustomStyle = classnames(styles.checkboxCell, styles.streamRowCheckboxCell);
 
   return (
     <Row className={styles.catalogSectionRow}>
       {!disabled && (
         <div className={checkboxCellCustomStyle}>
           <div className={iconStyle}>{changedSelected && (isStreamEnabled ? <PlusIcon /> : <MinusIcon />)}</div>
-          <CheckBox checked={isSelected} onChange={selectForBulkEdit} />
+          <CheckBox checked={isSelected} className={styles.checkBox} onChange={selectForBulkEdit} />
```
edmundito, line 105: Is there a `checkBox` style in the CSS? I didn't see one.

airbyte-webapp/src/components/ui/Switch/Switch.tsx
```diff
@@ -50,6 +50,7 @@ export const Switch: React.FC<SwitchProps> = ({
         value={value}
         disabled={loading || disabled}
         checked={checked || !!value}
+        onClick={(e) => e.stopPropagation()}
```
edmundito, line 53: Because the Props can include an onClick handler, this should call:

```ts
(e) => {
  e.stopPropagation();
  props?.onClick?.(e);
}
```

airbyte-webapp/src/components/ui/CheckBox/CheckBox.tsx
edmundito, line 37: At the very least, the onClick handler from the input props should be allowed:

```
(e) => {
  e.stopPropagation();
  props?.onClick?.(e);
}
```

edmundito (commented):
Overall, it's looking good. I had a couple of notes.

I tested only in the storybook but I noticed that when clicking on the checkbox multiple times the focus ring disappears sometimes. Aside from that I also feel that the focus ring is too strong and maybe should use a lighter color (though outside of the scope of this change)

airbyte-webapp/src/components/connection/CatalogTree/StreamHeader.tsx
tealjulia: No, thank you!  That was missed cleanup from something I tried out.

airbyte-webapp/src/components/ui/CheckBox/CheckBox.tsx
timroes: I'm in general a bit hesistant to put the stopPropagation logic into the core component itself. Us nesting interactive inputs (against best practice), I think should not cought on the base component levels (and thus prevent with standard browser behavior of inputs), but I'd rather suggest that we actually let the component bubble the event freely, and the consuming component that has problems caused by it (because it nested it inside another interactive element), must make sure to handle this properly (e.g. by checking in it's `onClick` event that the actual `ev.target` wasn't actually another interactive element).

Happy though to put that up for discussion and get more people's feeling, if we should or shouldn't handle event propagation stops inside core UI components.

tealjulia: That makes sense... I wasn't able to get it working this way anyways, even after pairing with @ambirdsall on it for quite some time.

airbyte-webapp/src/components/ui/Button/Button.module.scss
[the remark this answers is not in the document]
tealjulia: Yes!  That was odd, but makes sense, that the focus-ring would inherit the transition rule of whatever it's applied to
https://github.com/airbytehq/airbyte/pull/21271/commits/42ee0df97e8574b5aacdf4466f023409cae433eb

airbyte-webapp/src/components/ui/CheckBox/CheckBox.tsx
tealjulia: Leaving this open for posterity's sake, but all bits related to checkboxes have been removed from this PR.

PR #21272, opened 2023-01-11
Add missing info tooltips to new streams table headers

### LINKED ISSUE #20451: Add missing info tooltips to new streams table headers
Image

All headers except "Sync" should have info tooltips. We need them on Namespace, Stream name, and Primary Key.

All text is already contained in the localization file but may need to be split for the new table (do not change the existing tooltips).

Here's an example for the destination tooltip in the old table that can be split into the destination namespace and stream name:
Image

m19berger:
## What
Closes [#20451](https://github.com/airbytehq/airbyte/issues/20451)

## How
Added info tooltips to the new streams table headers

Fixed minor issues:
- "Learn more" link (color and margin-top)
- Relative import

_For test purposes, use the environment variable: REACT_APP_NEW_STREAMS_TABLE=true_

## Loom
https://www.loom.com/share/c67fd06ed3da478a9ffbdf6be98839c3

image

### DIFF
--- a/airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableHeader.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableHeader.tsx
@@ -86,13 +86,13 @@ export const CatalogTreeTableHeader: React.FC = () => {
       <HeaderCell>
         <FormattedMessage id="form.namespace" />
         <InfoTooltip>
-          <FormattedMessage id="connectionForm.destinationName.info" />
+          <FormattedMessage id="connectionForm.sourceNamespace.info" />
         </InfoTooltip>
       </HeaderCell>
       <HeaderCell>
         <FormattedMessage id="form.streamName" />
         <InfoTooltip>
-          <FormattedMessage id="connectionForm.destinationStream.info" />
+          <FormattedMessage id="connectionForm.sourceStreamName.info" />
         </InfoTooltip>
       </HeaderCell>
       <HeaderCell size="large">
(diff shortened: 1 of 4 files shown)


dizel852 (approved):
Code LGTM 👍 
Also tested locally - works as should

Passing to @edmundito for another review

airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableHeader.tsx
```diff
@@ -85,9 +85,15 @@ export const CatalogTreeTableHeader: React.FC = () => {
...
       <HeaderCell>
         <FormattedMessage id="form.namespace" />
+        <InfoTooltip>
+          <FormattedMessage id="connectionForm.destinationName.info" />
+        </InfoTooltip>
       </HeaderCell>
       <HeaderCell>
         <FormattedMessage id="form.streamName" />
+        <InfoTooltip>
+          <FormattedMessage id="connectionForm.destinationStream.info" />
```
edmundito, line 95: We need to create new strings for sourceName and souceStream because these talk about the destination, but these are about the source.

edmundito (changes requested):
Tested locally, looks good, but we need some new strings.

airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableHeader.tsx
matter-q: Fixed with temporary tooltip strings

edmundito (approved):
Tested locally on current and new table.

Issue #21074, opened 2023-01-05
UI for free connector components


ambirdsall:
New/updated UI components:
- status pill variants with "free" overlay for alpha and beta
- banners:
  - big showy marketing banner, images, the whole shebang
  - more subtle/interstitial banner for mid-connection-creation workflow
- signup modal with marketing images and stripe setup payment CTA

This new UI should all be conditionally-rendered, but the conditions can be hardcoded booleans behind a sensible API boundary during initial development. Similarly, clicking the stripe setup payment CTA can be a noop until the new endpoint it will hit is ready.
 
![image](https://media.tenor.com/faPfYg--3D0AAAAC/indiana-jones-harisson-ford.gif)

pmossman:
Refining notes:
- Alex can split out more granular tickets from this to divide up the work

Issue #20133, opened 2022-12-06
Create Icon component


timroes:
We would like to use more custom SVG icons over just using Font Awesome. We should create a custom `Icon` component that aligns handling those icons. At the moment we're just exporting custom React components for each of them, which all come with a bit different behavior (regarding size, how we handle colors, etc.).

I think the new icon component should do the following:

* Allow specifying the icon type as a string, so it's easier usable, e.g. `<Icon type="caret" />`
* Should allow specifying size of the icon in shirt sizes, e.g. `<Icon size="md" />` to match our regular icon sizes we have (and align them.
* Allow specifying color for the icon from a predefined set of colors we usually use or `text` for using `currentColor`.
* Allow specifying a `className` as we should for all base UI components.

That could allow us to manage the actual icons even in pure `svg` files if we want. We could consider loading all icons in one svg once only on the page and referencing them via `symbol` and `use` so we don't need to duplicate the icon SVG if we have an icon used multiple times on the same page.

```[tasklist]
- [ ] #22433
- [x] Support color and size
- [ ] https://github.com/airbytehq/airbyte/issues/24594
- [ ] https://github.com/airbytehq/airbyte/issues/24595
- [ ] https://github.com/airbytehq/airbyte/issues/24596
- [ ] https://github.com/airbytehq/airbyte/issues/24598
```

krishnaglick:
@timroes What do you mean about color. Are just generic css color's fine?

timroes:
@krishnaglick I think we want to have a limited set of colors only available, like "primary", "subdued", or such. We'd need to actually check what colors we're at the moment using in our UI, and what colors we can break that down to. Might be worth syncing with Nico about which colors we'll need exactly.

Upmitt:
@krishnaglick @timroes - I did this list of colors used for icons https://www.figma.com/file/gMNRYVJav25pdsoxAyW7Qp/00_01_WEB-APP-LIBRARIES?node-id=2514%3A4578&t=2dhs2PhfGlCBSMY6-4 - do you want to plan a meeting to talk about it?

timroes:
> @krishnaglick @timroes - I did this list of colors used for icons https://www.figma.com/file/gMNRYVJav25pdsoxAyW7Qp/00_01_WEB-APP-LIBRARIES?node-id=2514%3A4578&t=2dhs2PhfGlCBSMY6-4 - do you want to plan a meeting to talk about it?

Thanks for the list.

[...]

PR #21273, opened 2023-01-11
<Callout /> component

teallarson:
## What

Antecedent to some of our work in https://github.com/airbytehq/airbyte/issues/21074 I've made some improvements (and a rename) to the artist formerly known as `<InfoBox />`

Screenshot 2023-01-12 at 3 05 39 PM

## How
- The component only takes children and a classname now
- It uses scss modules instead of styled components now
- Updated the implementation of the component to included a few needed custom CSS overrides (see `WarningBox`) and make the icons children rather than props

## Recommended reading order
1. `InfoBox.*`
2. others


airbyte-webapp/src/components/ui/Callout/Callout.tsx
```diff
@@ -0,0 +1,23 @@
+import classNames from "classnames";
+
+import styles from "./Callout.module.scss";
+
+interface CalloutProps {
+  className?: string;
+  variant?: "yellow" | "red" | "blue";
```
tealjulia, line 7: Not in use anywhere yet, but it is in some Figma designs so am adding it now to use shortly.

josephkmh: Is there some semantic meaning we can attach instead of color names? Like `warning | error | default`? This could help us consistently use them for the right reason, and also makes changing the color name later easier, if we want to!

airbyte-webapp/src/components/ui/Callout/Callout.tsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export const Callout: React.FC<React.PropsWithChildren<CalloutProps>> = ({
+  children,
+  className,
+
```
josephkmh, line 13: nit: empty line?

airbyte-webapp/src/components/ui/Callout/Callout.tsx
```diff
@@ -0,0 +1,23 @@
...
+  variant = "yellow",
+}) => {
+  const containerStyles = classNames(styles.container, className, {
+    [styles.yellow]: variant === "yellow",
+    [styles.red]: variant === "red",
+    [styles.blue]: variant === "blue",
+  });
+
+  return <div className={containerStyles}>{children}</div>;
```
josephkmh, line 22: Take it or leave it: should we add a prop for an optional icon? I imagine a lot of callouts have an icon, and that would make it easy to ensure spacing between icon & text (or whatever is in `{children}`) is consistent.

joeykmh (changes requested):
Looks good, I would just suggest we use semantic names instead of colors for `variant`. WDYT?

airbyte-webapp/src/components/ui/Callout/Callout.tsx
tealjulia: Yep! Happy to switch that back.

tealjulia:
My reason for taking the semantic names out is that we don't have a consistent pattern for them app-wide currently. I don't feel strongly on it and agree it makes it easier to change theming later.

airbyte-webapp/src/components/ui/Callout/Callout.tsx
tealjulia: IMO no.  I took out the existing prop for that as it was only used in 1-2 places.  

There is also a spot where I plan on adding a button to a Callout, etc.  Once I started considering different ways this _might_ be used, I ended up deciding the important part here is really just the basic container.

I'm more in favor of keeping this very basic and unconcerned with what its children are personally.  It does have flex, so it does space out its children consistently by default (which can be overridden by a classname if wanted)

airbyte-webapp/src/components/ui/Callout/Callout.tsx
tealjulia: https://github.com/airbytehq/airbyte/pull/21273/commits/657236cad5c5918135215da8906b0260e76c3a67

airbyte-webapp/src/components/ui/Callout/Callout.tsx
tealjulia: Having said that I just saw one more usage when I rebased.

However, with the move away from FA towards custom icons, we'd likely be passing in a ReactNode or a svg, at which point it feels to me like we may as well just accept it as a child... I could be swayed there though.

josephkmh: Fine for me to skip this for now! I think when we do eventually move towards custom icons, we'll still want to enforce sizing constrains so that different `<Callout>` usages don't have different icon sizes, something like `<Callout icon="exclamation-mark">{ ... some text or JSX ... }</Callout>`

But no need to address that yet - we can probably enforce that later if/when we implement a custom `Icon` component: https://github.com/airbytehq/airbyte/issues/20133

joeykmh (approved):
Looks great! Approving and crossing my fingers that CI will be green 😅

PR #21286, opened 2023-01-12
add attempt metrics

cgardens:
## What
I want to be able to build this dashboard (below). I want to know what the failure rate is with connector failures removed. Ideally this is what we would use to trigger pages for the _platform_  team, instead of triggering based on failure rate with connectors included.
Screen Shot 2023-01-11 at 5 26 29 PM ([link](https://app.datadoghq.com/dashboard/ad6-yqt-54d/charles-kpi-dashboard?fullscreen_end_ts=1673487186475&fullscreen_paused=false&fullscreen_section=overview&fullscreen_start_ts=1672882386475&fullscreen_widget=3379745404882158&from_ts=1672877320065&to_ts=1673482120065&live=true))

I can't build it because we don't emit one event per attempt or job. We emit n events based on the number of release stages and failure origins. Because of this, I can't really get numbers that reflect the reality. They get skewed by the number of stages and failure origins. I end up with a result I don't really trust.

In addition, I find dashboards like this [one](https://app.datadoghq.com/dashboard/fai-itf-f54/chriss-dashboard?fullscreen_end_ts=1673486919717&fullscreen_paused=false&fullscreen_section=edit&fullscreen_start_ts=1670808519717&fullscreen_widget=7929707339489212&from_ts=1670803915040&to_ts=1673482315040&live=true) that are wrong, because they double count jobs since for each job there could be 1 or 2 release stages.

## How
* I am proposing emitting new events for attempts. There will be 1 created and 1 completed event for each attempt.
* I get around the multiple release stage issue by having a min and max release stage field. Any graph we could build with the old one we could build using this new scheme and not have confusing cardinality issues.
* For failure origin and failure type, just emit the first one (if present).
* Suggests normalizing  the way we refer to data planes. Using just geography isn't going to cut it. We need something. Not too picky on what.
* Added a constructor in `OssMetricRepository` so that we can declare what dimensions we expect to ship with a metric. Not sure this is a good idea yet. The declaration is good, but it currently can't be enforced.
* I did it for attempts and not jobs (at least for now) because that's really what I need for the current dashboard, but we could do the same for jobs if we like it.
* Skipped doing pending since it's not really meaningful. I would love to figure out how to get a meaningful pending metric though.

## To d
* Still need to wire up the metrics to actually get emitted. Will do this if I get a :+1: on the approach.

### DIFF
--- a/airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/OssMetricsRegistry.java
+++ b/airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/OssMetricsRegistry.java
@@ -166,6 +166,8 @@ public enum OssMetricsRegistry implements MetricsRegistry {
   private final String metricName;
   private final String metricDescription;
 
+  // added this field to declare metric attributes, but we never read them.
+  @SuppressWarnings("FieldCanBeLocal")
   private final List<String> metricTags;
 
   OssMetricsRegistry(final MetricEmittingApp application,
--- a/airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/OssMetricsRegistry.java
+++ b/airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/OssMetricsRegistry.java
@@ -141,33 +141,26 @@ public enum OssMetricsRegistry implements MetricsRegistry {
       "reset_request",
       "number of requested resets"),
 
-  ATTEMPT_CREATED(
+  ATTEMPTS_CREATED(
       MetricEmittingApps.WORKER,
       "attempt_created",
       "increments when a new attempt is created. one is emitted per attempt",
-      MetricTags.DATA_PLANE_ID,
-      MetricTags.MIN_CONNECTOR_RELEASE_STATE, // max
-      MetricTags.MAX_CONNECTOR_RELEASE_STATE // min
-  ),
-  // ATTEMPT_PENDING (doesn't actually mean anything)
-  ATTEMPTS_RUNNING(
-      MetricEmittingApps.METRICS_REPORTER,
-      "attempts_running",
-      "number attempts in status running at a moment in time.",
-      MetricTags.DATA_PLANE_ID, // really should be data-plane id
-      MetricTags.MIN_CONNECTOR_RELEASE_STATE, // max
-      MetricTags.MAX_CONNECTOR_RELEASE_STATE // min
-  ),
-  ATTEMPT_COMPLETED(
+      MetricTags.GEOGRAPHY,
+      MetricTags.ATTEMPT_NUMBER,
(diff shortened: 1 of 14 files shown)


airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/OssMetricsRegistry.java
```diff
@@ -137,21 +139,53 @@ public enum OssMetricsRegistry implements MetricsRegistry {
...
   private final String metricName;
   private final String metricDescription;
 
+  private final List<String> metricTags;
+
   OssMetricsRegistry(final MetricEmittingApp application,
                      final String metricName,
-                     final String metricDescription) {
+                     final String metricDescription,
+                     final String... metricTags) {
```
davinchia, line 181: i like this idea even though we can't strictly enforce it. Perhaps we can do slightly better by turning this into a list of metric tags and having a static constant named `NO_METRIC_TAGS` to make it obvious.

davinchia:
Sounds reasonable. I like the min/max approach. I think we should give it a shot. It's a 2-way door so should be easy to try and pick what we like best.

Curious, why do you say pending isn't meaningful? If I recall correctly, that alert has helped us in the past when there are issues with the worker deployment.

airbyte-metrics/metrics-lib/src/main/java/io/airbyte/metrics/lib/OssMetricsRegistry.java
jdpgrailsdev: @davinchia That would add additional logic to check the list to see if it only has one entry and it's value is `NO_METRIC_TAGS`.  That seems more complicated than using the ellipses to indicate that the list/array can simply not be provided if there are no tags.  This also aligns with how the OpenTracing/DataDog tracing classes work.

jdpgrailsdev (approved):
:shipit:

cgardens:
@davinchia 
> Curious, why do you say pending isn't meaningful? If I recall correctly, that alert has helped us in the past when there are issues with the worker deployment.

```
      workflowInternalState.setJobId(getOrCreateJobId(connectionUpdaterInput));
      workflowInternalState.setAttemptNumber(createAttempt(workflowInternalState.getJobId()));
```
A job is set to pending on line 208 of `ConnectionManagerWorkflowImpl`. It is then set to started on line 209. So all `pending` tells us is that it line 208 has executed but not 209, which doesn't seem all that useful. 😛 It would be great to have a way of actually tracking `pending`, aka we have a job we want to schedule but it hasn't started yet.

cgardens:
Update: I removed the running attempts metric. Because we use a gauge, it doesn't work well to add several dimensions. For now I think nailing the creation and completion is really the key anyway.

@jdpgrailsdev mind taking one more look?

PR #21005, opened 2023-01-04
normalization: delete supportsDbt and supportsNormalization from DestinationDefinitionSpecificationRead

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/18239

Now that the source of truth for `supportsDbt` and `normalizationConfig` is in the destination definition and not in it's specification we need to remove these fields from the `DestinationDefintionSpecificationRead` model.

## How
1. Remove `supportsDbt` and `supportsNormalization` from `DestinationDefintionSpecificationRead` model
2. Expose `supportsDbt` and `normalizationConfig`in `DestinationDefinitionRead` model as it is [required by the front-end for the `ConnectionTransformationTab`](https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionTransformationTab.tsx#L102). 
3. Update `airbyte-webapp` according to API changes
4. Update `octavia-cli` according to API changes 

## Recommended reading order
1. `airbyte-api` Open API configuration changes
2. `airbyte-server` Handler logic update
3. `airbyte-webapp` changes to reflect API updates
4. `octavia-cli` changes to reflect API updates.

## 🚨 User Impact 🚨
No user impact is expected.

Issue #14276, opened 2022-06-29
[MOVE EPIC] Airbyte Protocol Versioning


benmoriceau:
Product Requirements Doc - https://docs.google.com/document/d/1RMCSZKt9GT_RMR0hI4jjb7WBcppkzTTQ87hPqSqzwT8/edit#

> Airbyte does not currently support transitions of the Airbyte Protocol evolution. As a result, we require both the platform and the connectors to always support the same version of the protocol and we lack the ability to clearly delineate which connectors work with which versions of the platform, and communicate this to our users. We want to enable progressive update of the platform and the connectors. 
> … enter Airbyte Protocol Versioning

## Airbyte Protocol Versions

**v0.0.0** - initial version of Airbyte Protocol 
**v0.1.0** - AirbyteTraceMessage - new feature, non breaking, optional 
**v0.2.0** - New State messages introduced (Aug 2022 - here now)
but still backwards compatible with connectors dealing with both versions of STATE
**v1.0.0** - connectors advertise their version of the protocol (or we assume a version)
***v1.1.0** - New datatypes in Protocol (additive change)*
***v1.2.0** - New AirbyteTraceMessage (optional additive change)*
**v2.0.0** - legacy state removed (breaking change - Feb 1, 2023)

davinchia:
This work has been deprioritised.

PR #21295, opened 2023-01-12
airbyte-protocol: remove `supportsNormalization` and `supportsDbt` from `ConnectorSpecification`

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte-internal-issues/issues/2536
We are removing the `supportsNormalization` and `supportsDbt` from the connector specification and moved these metadata to destination definitions. (https://github.com/airbytehq/airbyte/pull/21005)

## How
* Remove these two fields from the protocols `ConnectorSpecification` model in `airbyte_protocol.yaml`
* Update mocks using `withSupportsDbt` or `withSupportsNormalization`
* Update the CDK models

### DIFF
--- a/airbyte-protocol/protocol-models/src/main/resources/airbyte_protocol/airbyte_protocol.yaml
+++ b/airbyte-protocol/protocol-models/src/main/resources/airbyte_protocol/airbyte_protocol.yaml
@@ -452,16 +452,6 @@ definitions:
       supportsIncremental:
         description: (deprecated) If the connector supports incremental mode or not.
         type: boolean
-      # Destination Connectors Properties
-      # Normalization is currently implemented using dbt, so it requires `supportsDBT` to be true for this to be true.
-      supportsNormalization:
-        description: If the connector supports normalization or not.
-        type: boolean
-        default: false
-      supportsDBT:
-        description: If the connector supports DBT or not.
-        type: boolean
-        default: false
       supported_destination_sync_modes:
         description: List of destination sync modes supported by the connector
         type: array
--- a/airbyte-cdk/python/airbyte_cdk/models/airbyte_protocol.py
+++ b/airbyte-cdk/python/airbyte_cdk/models/airbyte_protocol.py
@@ -327,8 +327,6 @@ class ConnectorSpecification(BaseModel):
         None,
         description="(deprecated) If the connector supports incremental mode or not.",
     )
-    supportsNormalization: Optional[bool] = Field(False, description="If the connector supports normalization or not.")
-    supportsDBT: Optional[bool] = Field(False, description="If the connector supports DBT or not.")
     supported_destination_sync_modes: Optional[List[DestinationSyncMode]] = Field(
         None, description="List of destination sync modes supported by the connector"
     )
--- a/airbyte-config/config-persistence/src/test/java/io/airbyte/config/persistence/MockData.java
+++ b/airbyte-config/config-persistence/src/test/java/io/airbyte/config/persistence/MockData.java
@@ -263,9 +263,7 @@ public class MockData {
         .withAdvancedAuth(null)
         .withChangelogUrl(URI.create("whatever"))
         .withSupportedDestinationSyncModes(Arrays.asList(DestinationSyncMode.APPEND, DestinationSyncMode.OVERWRITE, DestinationSyncMode.APPEND_DEDUP))
-        .withSupportsDBT(true)
-        .withSupportsIncremental(true)
-        .withSupportsNormalization(true);
+        .withSupportsIncremental(true);
   }
 
   public static StandardDestinationDefinition publicDestinationDefinition() {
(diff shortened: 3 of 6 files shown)


alafanechere:
@gosusnp `testDowngrade` does not pass. I guess this might have something to do with protocol versioning / migration. Could you please tell me what I could do to make this test pass or point me to documentation about protocol updates with migrations?

cgardens:
woohoo!

gosusnp:
@alafanechere, looking at the test failures, I'd expect it comes from the input of the test. The fields that were removed are now in the `additionalProperties` field. Most likely because the updated model doesn't know about them.

In terms of protocol version, this PR removes fields so it would be a breaking change and should result in a major bump. Generally, I would recommend to delay this change to avoid bumping majors with only clean up of old fields.
One suggestion is that since we are currently working on getting v1 of the protocol with new data types out. We could take that opportunity to remove those fields as well.

alafanechere:
Thank you @gosusnp for the details
> Generally, I would recommend to delay this change to avoid bumping majors with only clean up of old fields.
I'm not entirely sure about what you're suggesting here. 
Should the other changes I made on this PR (the mock updates) happen before we make the protocol change?

> One suggestion is that since we are currently working on getting v1 of the protocol with new data types out. We could take that opportunity to remove those fields as well.

Is this happening on another PR on which I could push these changes?

airbyte-protocol/protocol-models/src/main/resources/airbyte_protocol/airbyte_protocol.yaml
```diff
@@ -452,16 +452,6 @@ definitions:
...
-      # Destination Connectors Properties
-      # Normalization is currently implemented using dbt, so it requires `supportsDBT` to be true for this to be true.
-      supportsNormalization:
-        description: If the connector supports normalization or not.
-        type: boolean
-        default: false
-      supportsDBT:
-        description: If the connector supports DBT or not.
-        type: boolean
-        default: false
```
evantahler, line 464: There are probably some related updates to make in https://github.com/airbytehq/airbyte/blob/master/docs/understanding-airbyte/airbyte-protocol.md#actor-specification

alafanechere: Done, thanks for the suggestion.

evantahler:
@gosusnp can you help me understand why this is a breaking change?  Perhaps this is about message validation.  As supportsDbt and supportsNormalization still required platform changes to work, we don't expect OSS users to have made connectors that made use of these flags.  Ideally we can remove them without breaking things for users.

* Do we need to change `additionalProperties:true` in all the destination specs to avoid parse errors?  I don't think so, because the specs themselves are definitions in jsonSchema, maybe?
* Do we need to change the model for ConnectorSpecification?  [This](https://github.com/airbytehq/airbyte/blob/master/airbyte-config/config-models/src/main/resources/types/StandardGetSpecOutput.yaml) might imply that perhaps we should change `additionalProperties:true` here, so that when a connector returns the now-removed supportsDbt property, we don't fail validation

gosusnp:
@evantahler, my comment is purely on the nature of the change, removing a field should fall under a breaking change. Now, that doesn't mean it will actually break something.

I think the default for `additionalProperties` is `true` so it may be redundant to add it everywhere. The test Augustin made shows that we parsed those attributes into the `additionalProperties` so we are not going to reject the spec. Given the breath of connectors, we should benefit from being more fault resilient when it comes to parsing the messages and probably should only fail on attributes that are key to a correct replication.

`StandardGetSpecOutput` is a different use case, I do not think those objects should be considered public. Being stricter here enables us to catch our internal errors sooner. Some of those objects currently bleed into the connectors mostly from how we test, after recent discussions, I am leaning towards providing dedicated tooling for this purpose which would remove this dependency.

This change might be safe, but a goal of versioning was to take the guess work out of it. Unless we have a strong argument for having to clean it now, I would generally recommend to flag for removal and remove in the next version.

sherifnada (approved):

alafanechere:
> This change might be safe, but a goal of versioning was to take the guess work out of it. Unless we have a strong argument for having to clean it now.


My only argument to clean it now is, but I'm not sure it's a strong one :smile :
- Java connectors rely on the protocols models to build the `spec` output. If we're not removing these fields in the protocol they will fall back to the default `false` value, which will be misleading (https://github.com/airbytehq/airbyte/pull/21306). We'll have `spec` outputs with `supportsNormalization: false` on destination bigquery. It won't break the connector as the normalization support is now stored in the definition, but it will create misleading `spec` outputs for the connector.

>  I would generally recommend to flag for removal and remove in the next version

How can I flag these change to be made in the next protocol version?

alafanechere:
As @gosusnp mentions it's a breaking change I'm closing this PR and adding the original issue to this [EPIC](https://github.com/airbytehq/airbyte/issues/14276) to flag it for removal in the next major version.

PR #21306, opened 2023-01-12
GA destinations: remove `supportsDbt` and `supportsNormalization` from spec.json

alafanechere:
## What
Issue: https://github.com/airbytehq/airbyte-internal-issues/issues/2539
The source of truth for  `supportsDbt` and `supportsNormalization`  is now in the destination definition.
We should remove it from the destination spec. 
This PR is doing it for GA destinations.


alafanechere:
NB: I don't think it's safe to publish it until cloud is deployed with OSS >= `0.43.18`

airbyte-integrations/connectors/destination-bigquery/Dockerfile
```diff
@@ -17,5 +17,5 @@ ENV ENABLE_SENTRY true
 
 COPY --from=build /airbyte /airbyte
 
-LABEL io.airbyte.version=1.2.9
+LABEL io.airbyte.version=1.2.10
```
alafanechere, line 20: I should introduce a major version bump as this will likely break normalization on old OSS versions

evantahler (approved):
Nice!
* Major version bump (semver!)
* Docs explain why & what version of the platform is needed
* connector tests all pass

alafanechere:
Cloud is now on >= 0.43.18.

alafanechere:
@airbytehq/use please checkout the changelog. 
This will probably make these connectors not work on OSS versions <= 0.43.8. 
Feel free to organize a community outreach.

alafanechere:
So the only benefit from the current PR would be confirming the platform 💯 relies on definitions to manage connector normalization. But I'm not sure it's worth the confusion of releasing connectors with `supportsNormalization = false`  in the `spec` output.

alafanechere:
@davydov-d let me know once you created the OAuth parameters on the Cloud DB and I'll approve the PR so that we can merge it tommorow.

grishick (approved):

alafanechere:
Hey @grishick do you mind if I merge these? It will lead to misleading spec outputs according to what I explained [here](https://github.com/airbytehq/airbyte/pull/21295#issuecomment-1403822541). I'd understand if you preferred to hold these changes until the protocol update.

grishick:
> Hey @grishick do you mind if I merge these? It will lead to misleading spec outputs according to what I explained [here](https://github.com/airbytehq/airbyte/pull/21295#issuecomment-1403822541). I'd understand if you preferred to hold these changes until the protocol update.

I am confused as to how these changes affect the destinations. From the comment that you linked is sounds like if you merge this PR, normalization will be disabled for BigQuery and Snowflake destinations, which is not good.

alafanechere:
> is sounds like if you merge this PR, normalization will be disabled for BigQuery and Snowflake destinations, which is not good.

We now pull the information about the normalization / dbt support from the connector definition, so the connector spec is not used by the platform to determine if normlization is supported. So merging this PR should not break anything. My concern is that the spec output is going to be misleading: as the we remove these field from these connector spec their value will fallback to the default defined in the protocol model. So running `spec` on BigQuery will output `supportsNormalization: false` which is not great. 
We can hold until the protocol change.

grishick:
> So running spec on BigQuery will output supportsNormalization: false which is not great.

This seems dangerous.

Another question. If I am adding a new connector with normalization support - how do I do that without breaking the CI? I cannot add the connector to `destination_definitions.yaml` or `destination_specs.yaml` before I publish it, because the CI will complain about a missing Docker image. All integration tests that use normalization will be skipped for this connector, because it has no way of declaring that it supports normalization outside of the seed files.

alafanechere:
Closing to wait for the protocol update tracked [here](https://github.com/airbytehq/airbyte/pull/21295) or the decoupling of java connectors from protocol models.

PR #21311, opened 2023-01-12
survey cto updates: code improvement 

siddhant3030:
## What
Changing the SurveyCto connector for improving the code and we have put some checks on it.


airbyte-integrations/connectors/source-surveycto/source_surveycto/helpers.py
```diff
@@ -13,6 +13,20 @@
 
 
 class Helpers(object):
+
+    @staticmethod
+    def modify_date(response_json):
+        data_mod = ["startdata", "enddata", "CompletionDate", "SubmissionDate"]
```
jakejh, line 19: Shouldn't these be "starttime" and "endtime" instead of "startdata" and "enddata"?

siddhant3030: done

airbyte-integrations/connectors/source-surveycto/source_surveycto/helpers.py
```diff
@@ -13,6 +13,20 @@
...
+        data_mod = ["starttime", "endtime", "CompletionDate", "SubmissionDate"]
+        dateformat_in = "%b %d, %Y %I:%M:%S %p"
+        dateformat_out = "%Y-%m-%dT%H:%M:%S+00:00"
+
+        for data in response_json:
+            for key in data_mod:
+                if key in data:
+                    data[key] = datetime.strptime(data[key], dateformat_in).strftime(dateformat_out)
+
+                yield data
```
marcosmarxm, line 28: You're creating a function to use only one time. Also now you're doing a double for loop in your data besides only one. I don't think it is the best strategy here.

airbyte-integrations/connectors/source-surveycto/source_surveycto/helpers.py
```diff
@@ -38,20 +52,8 @@ def call_survey_cto(config, form_id):
         if response.status_code != 200 and response_json["error"]:
             message = response_json["error"]["message"]
             raise Exception(message)
```
marcosmarxm, line 54: My suggestion would be to improve the code.
```python
        dateformat_in = "%b %d, %Y %I:%M:%S %p"
        dateformat_out = "%Y-%m-%dT%H:%M:%S+00:00"
        date_keys_to_transform = ["starttime", "endtime", "CompletionDate", "SubmissionDate"]
        for data in response_json:
            try:
                for k in date_keys_to_transform:
                    data[k] = datetime.strptime(data[k], dateformat_in).strftime(dateformat_out)
                yield data
            except Exception as e:
                raise e
        return data
```

docs/integrations/sources/surveycto.md
```diff
@@ -47,3 +47,4 @@ The Commcare source connector supports the following streams:
 
 | Version | Date | Pull Request | Subject |
 | 0.1.0   | 2022-11-16 | [19371](https://github.com/airbytehq/airbyte/pull/19371)   | SurveyCTO Source Connector |
```
marcosmarxm, line 49: Check others connectors how they made the changelog, this is not the standard.

marcosmarxm (changes requested):
Some comments.

jakejh:
@siddhant3030 What do you think of the various changes I proposed in https://github.com/siddhant3030/airbyte/pull/1/files?

A few are rewording and clarifying documentation, but in my mind at least the following are more substantive:

1. Change "startdata" and "enddata" (which don't exist in my SurveyCTO forms) to "starttime" and "endtime" (which do exist).
2. Use "SubmissionDate" as the `cursor_field` instead of "CompletionDate". I have forms that have the former, but not the latter.
3. Clarify that time-stamps are UTC.
4. Fix typo "Dic" to "Dec" in the regex pattern for `start_date`.

PR #21317, opened 2023-01-12
remove supportsNormalization + supportsDbt from sources spec

alafanechere:
## What
Issue: https://github.com/airbytehq/airbyte-internal-issues/issues/2539
Some source connectors have `supportsNormalization` and `supportsDbt` fields.
These field are useless for sources and are now defined in destination definition.

## How
Remove these fields from the `spec.json` file
Update the `expected_spec.json` used in testing.


alafanechere:
@evantahler do I have to publish the connectors or only updating the `source_spec.json` file is OK?

evantahler (commented):
I think this PR and https://github.com/airbytehq/airbyte/pull/21317 will have some collisions (e.g. both change `airbyte-integrations/connectors/source-bigquery/src/main/resources/spec.json` the same way).  

We probably should publish the connectors (with a major semver bump + changelog entry like in https://github.com/airbytehq/airbyte/pull/21317)

alafanechere:
@evantahler I think you rather wanted to mention PR https://github.com/airbytehq/airbyte/pull/21306 . The current PR is addressing source connectors only as the others are touching destinations. 
I don't think the two fields I removed are used anywhere for sources, this I why I think it's safe to remove these without a major bump. I would also like to consider not publishing / version bumping these source connectors and only updating the `source_spec.json`. Wdyt?

alafanechere:
For the same reasons I detailed [here](https://github.com/airbytehq/airbyte/pull/21319#issuecomment-1381624197) I think it's safe to merge these changes without version bump / publish.
These fields are moreover not used for `sources` in the platform.

evantahler (approved):

PR #21319, opened 2023-01-12
Remove `supportsNormalization` `supportsDbt` from non GA destination spec

alafanechere:
## What
Issue https://github.com/airbytehq/airbyte-internal-issues/issues/2539

Removing `supportsNormalization` `supportsDbt` for non ga destination connector specs.


airbyte-config/init/src/main/resources/seed/destination_definitions.yaml
```diff
@@ -135,6 +135,7 @@
   documentationUrl: https://docs.airbyte.com/integrations/destinations/firebolt
   icon: firebolt.svg
   releaseStage: alpha
+  supportsDbt: true
 - name: Google Cloud Storage (GCS)
```
evantahler, line 139: good catch!

evantahler (approved):
👍 

~~Why did you split this pr and https://github.com/airbytehq/airbyte/pull/21319 ?~~ Edit: GA vs Non-GA, I get it now!

alafanechere:
I'm merging without publishing because:
- All these connectors have `supportsNormalization` and `supportsDbt` set to false
- `false` is the default value on these fields according to the protocol `ConnectorSpecification` model 
- So the `spec.json` edit I made are not changing the output of `spec` command
- This is the reason why the `destination_spec.json` is not updated.

I'm reverting the `firebolt` spec change as this is the single connector that has `supportsDbt = true`

PR #21322, opened 2023-01-12
Source Gnews: reorganize inline schema in gnews.yaml

clnoll:
## What
Improve readability of the gnews manifest.

## How
Now that we're supporting forward references, we can move the `schemas` key in to the bottom of the manifest, when using inline schemas.

## Pre-merge Checklist

- [x] Changelog updated
- [x] New Connector version released on Dockerhub and connector version bumped by running the `/publish` command described [here](https://docs.airbyte.io/connector-development#updating-an-existing-connector)


girarda (approved):
much better :)

PR #21327, opened 2023-01-12
store env-var response

colesnodgrass:
## What
- address excess logging messages

## How
- store results of the feature-flag, instead of checking them constantly

## Note
This is the quick fix, I'll create a long-term fix later today that will cache the env-var results within the `EnvVariableFeatureFlags` class itself.  That way the real env-var is only checked once and the log message would only be generated once.


benmoriceau (approved):
Should we set the logs to debug instead?

colesnodgrass:
> Should we set the logs to debug instead?

@benmoriceau I think we should. I'm planning on cleaning this quick-fix up in a separate pr (if I can ever get these checks to pass on this pr 😞)

gosusnp (approved):

PR #21329, opened 2023-01-12
Ensure Local Secret Persistence is Initialized

jdpgrailsdev:
## What
* Ensure that local testing persistence is initialized

## How
* Call `intialize()` before read/write

## Recommended reading order
1. `LocalTestingSecretPersistence.java`

### DIFF
--- a/docker-compose-cloud.buildx.yaml
+++ b/docker-compose-cloud.buildx.yaml
@@ -35,7 +35,7 @@ services:
           - airbyte/server:${ALT_TAG:-${VERSION}}
         platforms:
           - linux/amd64
-          - linux/arm64          
+          - linux/arm64
   worker:
     image: airbyte/worker:${VERSION}
     build:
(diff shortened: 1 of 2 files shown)


docker-compose-cloud.buildx.yaml
```diff
@@ -35,7 +35,7 @@ services:
           - airbyte/server:${ALT_TAG:-${VERSION}}
         platforms:
           - linux/amd64
-          - linux/arm64          
+          - linux/arm64
```
benmoriceau, line 38: is this duplication needed?

benmoriceau (approved):

docker-compose-cloud.buildx.yaml
jdpgrailsdev: @benmoriceau Could you be more specific?  The two entries are different (amd64 vs arm64)

colesnodgrass (approved):

davinchia (approved):

PR #21340, opened 2023-01-12
Added nginx reverse proxy example config

JulianRommel:
## What
I added an nginx example config for people who want to put their local installation of Airbyte behind a reverse proxy.
This may be important for people who want to use Airbyte in production with SSL encryption

## How
A new entry in the docs showing config for nginx.


marcosmarxm (approved):
Thanks for the addition @JulianRommel

PR #21342, opened 2023-01-12
JDBC Destinations: improve error message for conflicting streams

edgao:
## What
see https://github.com/airbytehq/airbyte-internal-issues/issues/2542 for information

I'm only planning to publish destination-snowflake; other destinations can get published whenever :shrug:

## How
In jdbc destination connectors, detect multiple tables writing into the same table (currently this just throws IllegalStateException) and throw a more informative config error.

## Recommended reading order
yes

## 🚨 User Impact 🚨
none, this is just improving how we handle an already broken situation

- [x] Code reviews completed
- [x] Documentation updated
    - [x] Connector's `README.md`
    - [x] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)

### Airbyter

- [x] Create a non-forked branch based on this PR and test the below items on it
- [ ] Build is successful
- [x] If new credentials are required for use in CI, add them to GSM. [Instructions](https://docs.airbyte.io/connector-development#using-credentials-in-ci).


airbyte-integrations/connectors/destination-jdbc/src/main/java/io/airbyte/integrations/destination/staging/StagingConsumerFactory.java
```diff
@@ -160,15 +167,32 @@ private static AirbyteStreamNameNamespacePair toNameNamespacePair(final WriteCon
...
+  CheckedBiConsumer<AirbyteStreamNameNamespacePair, SerializableBuffer, Exception> flushBufferFunction(
                                                                                                                final JdbcDatabase database,
                                                                                                                final StagingOperations stagingOperations,
                                                                                                                final List<WriteConfig> writeConfigs,
                                                                                                                final ConfiguredAirbyteCatalog catalog) {
-    final Map<AirbyteStreamNameNamespacePair, WriteConfig> pairToWriteConfig =
-        writeConfigs.stream()
-            .collect(Collectors.toUnmodifiableMap(
-                StagingConsumerFactory::toNameNamespacePair, Function.identity()));
+    final Set<WriteConfig> conflictingStreams = new HashSet<>();
```
ryankfu, line 176: This block seems to make sense in the `onStartFunction` since none of the values you're accessing here seem to be only isolated to the `flushBufferFunction` section

The reason that this makes sense to me in the `onStartFunction` is primarily because you already have the writeConfig values (since we're setting up the temporary tables for each stream (synonymous with writeConfig) and the metadata for each `streamIdentifier` will be known at setup time

It also fits with what `onStartFunction` means which is set up so this is only called _once_ as opposed to many times throughout the flushing of the buffer (N times where N can be infinitely large)

edgao: this block happens before the lambda gets created (i.e. it's not actually part of the flush buffer function), so I think it only gets executed once per sync

ryankfu: That does make sense, although I would be hesitant to have this in the `flushBufferFunction` since it muddies the singular responsibility of the buffer flush. Also, since we know that the lambda for `onStartFunction` is called only once and it already includes `writeConfigs` that it would be another suitable location

That said, this isn't blocking

ryankfu (approved):

edgao:
submitted https://github.com/airbytehq/airbyte/issues/21351 to solve for non-jdbc destinations

PR #21345, opened 2023-01-12
[ISSUE #19981] use github user with access to push on master

maxi297:
## What
Currently, the publish pipeline can't be executed in master because it uses the default GITHUB_TOKEN (the robot users that triggers workflows) which does not have access to push on master. Based on Peter Hu's comment:
* airbyteio is admin for the airbytehq/airbyte repository
* admins have access to push on protected branches


brianjlai (approved):

brianjlai:
just noting some failing CI checks but at least the connectors base stuff seems unrelated to your change

PR #21347, opened 2023-01-12
Make documentation_url optional in a declarative connector spec

lmossman:
## What
Make the `documentation_url` field of a declarative connector's spec optional. This is because users will not always have a documentation url to set on low-code connectors, and the documentation url is already optional in the airbyte protocol, so this brings the declarative CDK in alignment with that.

See slack thread for context: https://airbytehq-team.slack.com/archives/C03TP6W8081/p1673474183580999?thread_ts=1673463486.033949&cid=C03TP6W8081

## How
Make `documentation_url` non-required in the declarative component schema, as well as in the declarative spec.py file.

This has been tested on a low-code connector that doesn't set the documentation_url, as well as one that sets it to the empty string, and both cases work properly with `python main.py spec` returning the Spec that does not contain the documentation URL.
It was also tested that a low-code connector which _does_ set the documentation_url still has that url being returned as part of the spec for the above command.


airbyte-cdk/python/airbyte_cdk/sources/declarative/spec/spec.py
```diff
@@ -15,20 +15,23 @@ class Spec(JsonSchemaMixin):
...
+    documentation_url: Optional[str] = None
 
     def generate_spec(self) -> ConnectorSpecification:
         """
         Returns the connector specification according the spec block defined in the low code connector manifest.
         """
 
+        obj = {"connectionSpecification": self.connection_specification}
+
+        if self.documentation_url is not None and self.documentation_url != '':
```
girarda, line 33: nit: this could be simplified to
`if self.documentation_url:`

girarda (approved):
lgtm. next steps would be to update the [changelog](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/CHANGELOG.md) and [update the CDK version](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/setup.py#L18) before releasing the change

PR #21352, opened 2023-01-12
Update create connection page to full width, and fix trivial UI issues

### LINKED ISSUE #21338: Update create connection page to fit new streams table
Update the create connection page to have a full-width container and fix any outstanding issues from the change. The select/create source or destination steps should continue to be a fixed width.

edmundito:
## What
Resolves #21338 

Updates the create connection page screen full width on the create connection step (not the source or destination step). Additionally:

* Updates the Normalizations & Transformations section heading to match the rest of the headings
* Removes extra 20px padding from the Replication page
* Adds page bottom padding to the FormPageContent layout

## How
CSS changes


krishnaglick (approved):
Tested locally, LGTM!

PR #21357, opened 2023-01-12
Add Docker Debugging Fixture 

### LINKED ISSUE #21350: Add Fixture for Local Docker Debugging using Environment Variables
It is possible to connect a remote JVM debugger to a running docker container by passing along specific options and environment variables. The current solution requires making a change in the Dockerfile and rebuilding the image.

However, its possible to do this without rebuilding the image by making changes to the `docker-compose.debug.yaml` as well as the worker docker image code which spawns other containers, this debugging logic can live safely in the code, so developers don't accidentally commit the Dockerfile changes.

jbfbell:
- docker debugging options
- refactor image shortening method

## What
Add options to allow for Environment Variables to set debugging options on docker containers

## How
Docker Compose can inherit environment variables from the host as well as compose multiple yaml files together. For multivalue (lists) in Docker compose it will concat the values together. By adding additional environment variables in the `docker-compose.debug.yaml` we can add additonal options which are not activated with the standard `docker-compose.yaml`. 

This approach also does not require rebuilding the image since the environment variables are passed in at runtime.


docker-compose.debug.yaml
```diff
@@ -27,3 +27,13 @@ services:
     networks:
       - airbyte_internal
       - airbyte_public
+  worker:
+    environment:
+      - DEBUG_CONTAINER_IMAGE=${DEBUG_CONTAINER_IMAGE}
+      - DEBUG_CONTAINER_JAVA_OPTS=-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:5005
+  server:
+    # You will need to create a remote JVM debugging Run Configuration
```
etsybaev, line 35: I personally use the Ubuntu OS and never had issues with remote debugging (just added this var to the required dockerfile), but for MAC users there may be some issues with IPs. I've heard that this tool may help https://github.com/chipmk/docker-mac-net-connect
Maybe it would be worth also this link to notes. @suhomud knows better how it works on MAC.

jbfbell: @etsybaev good call out, definitely need that tool for the mac setup

The intent of this change is that you would no longer need to make any changes to `Dockerfile`s since those changes should never be committed.

grishick (approved):

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/DockerProcessFactory.java
```diff
@@ -169,6 +173,31 @@ public Process create(final String jobType,
...
+   * docker-compose.debug.yaml up ``` Additionally you may have to update the image version of your
+   * target image to 'dev' in the UI of your local airbyte platform. See the
+   * `docker-compose.debug.yaml` file for more context.
+   *
+   * @param containerName the name of the container which could be debugged.
+   * @return A list with debugging arguments or an empty list
+   */
+  static List<String> localDebuggingOptions(final String containerName) {
+    final boolean shouldAddDebuggerOptions =
+        Optional.ofNullable(System.getenv("DEBUG_CONTAINER_IMAGE")).filter(StringUtils::isNotEmpty)
```
gosusnp, line 191: Can you add the environment lookups to https://github.com/airbytehq/airbyte/blob/fca3940d2cf608cfa653a1695196d623bde4876a/airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java#L38?

For context, all the config goes through `EnvConfigs`, we're in the process of migrating to micronaut which should also be handling configuration for us: https://github.com/airbytehq/airbyte/blob/master/airbyte-workers/src/main/resources/application.yml. I am thinking `airbyte.container.docker.debug-container-java-opts` might make sense for the time being.

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/DockerProcessFactory.java
```diff
@@ -169,6 +173,31 @@ public Process create(final String jobType,
...
+            .map(ProcessFactory.extractShortImageName(containerName)::equals).orElse(false)
+            && Optional.ofNullable(System.getenv("DEBUG_CONTAINER_JAVA_OPTS")).isPresent();
+    if (shouldAddDebuggerOptions) {
+      return List.of("-e", "JAVA_TOOL_OPTIONS=" + System.getenv("DEBUG_CONTAINER_JAVA_OPTS"), "-p5005:5005");
```
gosusnp, line 195: optional: We may want to extract the port to a specific config as well.

gosusnp, line 195: Not entirely sure how this `JAVA_TOOL_OPTIONS=X` is getting rebuilt, do we need some extra quotes here?

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/DockerProcessFactory.java
```diff
@@ -169,6 +173,31 @@ public Process create(final String jobType,
     }
   }
 
+  /**
+   * !! ONLY FOR DEBUGGING, SHOULD NOT BE USED IN PRODUCTION !! If you set the DEBUG_CONTAINER_IMAGE
+   * environment variable, and it matches the image name of a spawned container, this method will add
+   * the necessary params to connect a debugger. For example, to enable this for
```
gosusnp, line 179: It could be worth adding this information to some dev docs if you know where they are, quick search sent me through all the specific connector's readme.

gosusnp (approved):
I do not have a better suggestion at this time so let's start with this approach. I highlighted a few changes so that it fits better within how we are reading configs currently, will be good for me once those are addressed.

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/DockerProcessFactory.java
jbfbell: I think its ok as is, this is all getting passed into a command List/Array to be sent to the docker run command.  Specifically not using `JAVA_TOOL_OPTIONS` as the env variable so that the container which is running this doesn't try and process it, but the result of this command should effectively be a List with 3 strings

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/DockerProcessFactory.java
jbfbell: can do

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/DockerProcessFactory.java
jbfbell: I'll poke around and see if I can find a good docs home

airbyte-commons-worker/src/main/java/io/airbyte/workers/process/DockerProcessFactory.java
jbfbell: Opting to not implement this change - this is for debugging only and probably should be more broadly integrated into the standard execution process. Could add as a follow on

Issue #1210, opened 2020-12-05
MSSQL source connector seems broken


johnlafleur:
https://analytics.amplitude.com/daxtarity/chart/46qlv54
Stats show that attempts end in failure most of the time

This might not be the case, but seems worth checking.

cgardens:
decided there is likely not a problem and the stats aren't quite right here.

PR #21361, opened 2023-01-12
Add back in Auto detect schema functionality in sync workflow

alovew:
Add back in Auto Detect Schema functionality that was removed due to OC issue #1210

ConfigFetchActivityImpl no longer relies on the ConfigRepository

### DIFF
--- a/airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
+++ b/airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
@@ -12,6 +12,7 @@ import static io.airbyte.metrics.lib.ApmTraceConstants.Tags.SOURCE_DOCKER_IMAGE_
 import static io.airbyte.metrics.lib.ApmTraceConstants.WORKFLOW_TRACE_OPERATION_NAME;
 
 import datadog.trace.api.Trace;
+import io.airbyte.api.client.model.generated.ConnectionStatus;
 import io.airbyte.commons.temporal.scheduling.SyncWorkflow;
 import io.airbyte.config.NormalizationInput;
 import io.airbyte.config.NormalizationSummary;
@@ -21,12 +22,16 @@ import io.airbyte.config.StandardSyncInput;
 import io.airbyte.config.StandardSyncOperation;
 import io.airbyte.config.StandardSyncOperation.OperatorType;
 import io.airbyte.config.StandardSyncOutput;
+import io.airbyte.config.StandardSyncSummary;
+import io.airbyte.config.StandardSyncSummary.ReplicationStatus;
+import io.airbyte.config.SyncStats;
 import io.airbyte.config.WebhookOperationSummary;
 import io.airbyte.metrics.lib.ApmTraceUtils;
 import io.airbyte.persistence.job.models.IntegrationLauncherConfig;
 import io.airbyte.persistence.job.models.JobRunConfig;
 import io.airbyte.protocol.models.ConfiguredAirbyteCatalog;
 import io.airbyte.workers.temporal.annotations.TemporalActivityStub;
+import io.airbyte.workers.temporal.scheduling.activities.ConfigFetchActivity;
 import io.temporal.workflow.Workflow;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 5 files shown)


airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
```diff
@@ -80,30 +84,26 @@ public StandardSyncOutput run(final JobRunConfig jobRunConfig,
     final int version = Workflow.getVersion(VERSION_LABEL, Workflow.DEFAULT_VERSION, CURRENT_VERSION);
     final String taskQueue = Workflow.getInfo().getTaskQueue();
 
-    // Temporarily suppressed to address OC issue #1210
-    @SuppressWarnings("PMD.UnusedLocalVariable")
     final int autoDetectSchemaVersion =
         Workflow.getVersion(AUTO_DETECT_SCHEMA_TAG, Workflow.DEFAULT_VERSION, AUTO_DETECT_SCHEMA_VERSION);
 
-    // Temporarily disabled to address OC issue #1210
```
benmoriceau, line 88: Just to double check, are all the issue related to this ticket solved? and is there a way to test it in dev?

alovew: Yes, the config repository is removed from all activities (config fetch activity & refresh schema activity). I'm not sure how to test this in the data plane. @jdpgrailsdev do you know?

jdpgrailsdev: We can at least test it locally to verify that the application starts up in "data plane" mode.  This will at least test the problem we saw in production with the missing data source.

jdpgrailsdev (approved):
:shipit:

alovew:
@benmoriceau @jdpgrailsdev I'm continually getting this error when I try to run this locally in data plane mode. The pods are stuck in pending, not showing errors. I'm pretty sure it's not related to these changes since i've always gotten this intermittently, but now it is happening every time.

```
BUILD SUCCESSFUL in 1m
265 actionable tasks: 129 executed, 128 from cache, 8 up-to-date
S3 cache 2s wasted on misses, reads: 8, elapsed: 1527ms
Finished build, proceeding with local deploy...
pod "airbyte-minio-create-bucket" force deleted
Creating jobs namespace if necessary...
Cleaning up any existing jobs pods to avoid job id conflicts...
Warning: Immediate deletion does not wait for confirmation that the running resource has been terminated. The resource may continue to run on the cluster indefinitely.
No resources found
Killing any existing processes running on desired ports...
Deploying airbyte cloud to local kube cluster...
Release "ab" does not exist. Installing it now.
Error: failed post-install: timed out waiting for the condition
Switching to docker-desktop kube context...
Switched to context "docker-desktop".
Tearing down local helm deployment...
release "ab" uninstalled
Killing port-forward of webapp proxy...
```

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
```diff
@@ -80,30 +84,26 @@ public StandardSyncOutput run(final JobRunConfig jobRunConfig,
...
-    // if (!status.isEmpty() && Status.INACTIVE == status.get()) {
-    // LOGGER.info("Connection is disabled. Cancelling run.");
-    // final StandardSyncOutput output =
-    // new StandardSyncOutput()
-    // .withStandardSyncSummary(new
-    // StandardSyncSummary().withStatus(ReplicationStatus.CANCELLED).withTotalStats(new SyncStats()));
-    // return output;
-    // }
-    // }
+    if (autoDetectSchemaVersion >= AUTO_DETECT_SCHEMA_VERSION) {
```
benmoriceau, line 90: I think that we will need a new version here. Since it wasn't commented out, the version will be in the history and even old workflow will run into undeterministic exception.

benmoriceau (changes requested):

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
alovew: I updated AUTO_DETECT_SCHEMA_VERSION to 2. is that what you mean?

benmoriceau: Yes that will work

benmoriceau (approved):

PR #21363, opened 2023-01-13
Add AllowedHosts to Actor Definitions and Config Database

### LINKED ISSUE #21183: Read `allowed_hosts` from actor definitions yaml and store in database
From the [tech spec](https://docs.google.com/document/d/1NHmdzMSBtF2dZWd_XXDRpvpJtF2DUFSdXXdhY2Tz6dI/edit#):

ActorDefintions are updated so that connectors can declare which hosts/IPs they intend to connect to.  This is a new optional field in [source_definitions.yaml](https://github.com/airbytehq/airbyte/blob/master/airbyte-config/init/src/main/resources/seed/source_definitions.yaml) and  [destination_definitions.yaml](https://github.com/airbytehq/airbyte/blob/master/airbyte-config/init/src/main/resources/seed/destination_definitions.yaml).

```yaml
- name: HubSpot
  sourceDefinitionId: 36c891d9-4bd9-43ac-bad2-10e12756272c
  dockerRepository: airbyte/source-hubspot
  dockerImageTag: 0.3.0
  documentationUrl: https://docs.airbyte.com/integrations/sources/hubspot
  icon: hubspot.svg
  sourceType: api
  # ~~~ NEW ~~~
  allowed_hosts:
    - *.hubspot.com 
```

The new “allowed_hosts” property is an array.  It’s values can be URL matches (per the example above, e.g. "*.instagram.com"), IP address and/or CIDR ranges, or a reference to a value that will be completed by the user completing the connector’s configuration.  The reference use-case is required for those connectors who accept a dynamic hostname or IP address to connect to (e.g. a database hostname or an API sub-domain).  A reference specification could look like:

```yaml
allowed_hosts:
 - “${#/definitions/config/hostname}”
```

Finally, we can combine the approaches for when there’s a customer-specific subdomain in use with interpolation:

```yaml
allowed_hosts:
- “${#/definitions/config/subdomain}.snowflake.com”
```

**This information should be loaded by the bootloader and stored in the actor_defintions table in the database in a new column as a JSON array.**

evantahler:
Part of the [Network Isolation](https://docs.google.com/document/d/1NHmdzMSBtF2dZWd_XXDRpvpJtF2DUFSdXXdhY2Tz6dI/edit#) epic.  Closes https://github.com/airbytehq/airbyte/issues/21183

This PR updates the actor definitions to provide `allowedHosts.hosts`, which is an array of URLs (or config matchers) which instances of this connector will be allowed to connect to.

There are a few logical cases:
* If `allowedHosts` is not present on the actor definition, this connector does not have any network restrictions.
* If `allowedHosts.hosts` is present on the actor definition, and is an empty array (see example included for `source-faker`), this connector is denied all network access. 
* If `allowedHosts.hosts` is present on the actor definition, and the array has content, access will be granted to those hosts, IPs, or patterns


airbyte-config/init/src/main/resources/seed/source_definitions.yaml
```diff
@@ -501,6 +501,7 @@
   icon: faker.svg
   sourceType: api
   releaseStage: alpha
+  allowedHosts: []
```
evantahler, line 504: This is an example of denying the source network access entirely.

evantahler:
I wanted to add some validation that allowedHosts match some known formats (URLs, IPs, etc), but until we know how we will be doing templating to reference config values (e.g. `"${#/definitions/config/subdomain}.snowflake.com"`), we should wait on that part of the story.

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
```diff
@@ -501,6 +501,8 @@
   icon: faker.svg
   sourceType: api
   releaseStage: alpha
+  allowedHosts:
+    hosts: []
```
evantahler, line 505: This example denies `source-faker` internet access entirely

git-phu (approved):
Lgtm!

git-phu:
Hmm now that I think about it, actor definition makes sense for API connectors, but would we also want to consider supporting allowed hosts for DB connectors?

The existing (non custom) DB connector definitions wouldn't know beforehand what the IP of the target DB is for a specific instance of a connector, but in the context of custom definitions, would we require the DB endpoint to be baked into the definition, or do we have something else in mind here?

git-phu:
Ah ok I see, thanks!

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigWriter.java
```diff
@@ -105,6 +105,8 @@ static void writeStandardSourceDefinition(final List<StandardSourceDefinition> c
             .set(Tables.ACTOR_DEFINITION.RESOURCE_REQUIREMENTS,
                 standardSourceDefinition.getResourceRequirements() == null ? null
                     : JSONB.valueOf(Jsons.serialize(standardSourceDefinition.getResourceRequirements())))
+            .set(ACTOR_DEFINITION.ALLOWED_HOSTS, standardSourceDefinition.getAllowedHosts() == null ? null
```
pedroslopez, line 108: Not exactly sure what the difference is, but I see other entries here are doing this: should this be `Tables.ACTOR_DEFINITION`?

pedroslopez (approved):
LGTM!

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigWriter.java
evantahler: Might as well match! https://github.com/airbytehq/airbyte/pull/21363/commits/cac0426da7b9187caecab9a1552f92b13ab4db9f

PR #21378, opened 2023-01-13
🪟🐛 Connector builder: Fix validation race condition

flash1293:
This PR fixes a small problem with the validation I noticed when running e2e tests. The error message for a field is shown if there is an error (which comes from the validation and the field is touched as well).

 As the validation is debounced by 200ms, a fast user can run into the following situation:
* Typing in a required field - during typing no validation is triggered because new change events are constantly coming in
* The field is initially in the error state because the field is required but not set
* The user stops typing and in under 200ms blurs the field by hitting tab
* Hitting tab causes the field to become touched, which means the error message is shown
* After 200ms, the validation is rerun and realizes the field is set now and no error message needs to be shown
* The error message disappears

What the user is seeing is a briefly blinking error message.

This PR is fixing this by doing an additional validation run on blur


lmossman (approved):
Funny, I just ran into this issue myself today right before I looked at this PR, thanks for putting up the fix!

LGTM, tested locally and it fixed the issue.

PR #21380, opened 2023-01-13
Connector builder: Fix e2e tests

flash1293:
Quick fix for the connector builder e2e tests

It looks like the connector builder server is not returning records anymore if no schema is passed along. This PR applies to quick-fix to the frontend, passing an empty schema along until it's fixed on the backend.


joeykmh (approved):
LGTM

PR #21404, opened 2023-01-13
Improvements to edge cases of CheckStream

erohmensing:
## What
CheckStream is meant to check that all of the configuration for connection is provided (e.g. API tokens, headers, etc.). We do this by trying to grab the first record from a stream, which uses all of the connection configuration to perform the record grabbing. 
* If there is no record (the stream is empty), it currently fails, even though we were able to successfully use the configuration to access the empty stream. This PR makes it so that this case is now considered a success.
* If there are no stream slices (Note: not when stream slice is None!), it currently fails when trying to use the stream slice within the path. This PR makes it so that if there are no stream slices, the check fails, since there is no way to check the complete path of the stream for access.

Fixes KeyError issue surfaced by the introduction of AvailabilityStrategy (see https://github.com/airbytehq/airbyte/pull/20891 and https://github.com/airbytehq/airbyte/issues/20677 for more context) 

## 🚨 User Impact 🚨

Users who try to perform a check() operation on a source using CheckStream, where at least one of the `streams_to_check` is empty, will now have success instead of failure checking.


girarda (approved):
🎉

erohmensing:
@pedroslopez thanks for asking about which edge cases we're trying to handle - adding you if curious. Hope the comments help with that!

airbyte-cdk/python/airbyte_cdk/sources/streams/utils/stream_helper.py
```diff
@@ -0,0 +1,36 @@
...
+    slices = iter(
+        stream.stream_slices(
+            cursor_field=stream.cursor_field,
+            sync_mode=SyncMode.full_refresh,
+        )
+    )
+    return next(slices)
+
+
+def get_first_record_for_slice(stream: Stream, stream_slice: Optional[Mapping[str, Any]]) -> StreamData:
```
girarda, line 28: is it worth mentioning this will raise an exception if there are no records?
alternatively, this could return an `Optional[StreamData]`

girarda (approved):

airbyte-cdk/python/airbyte_cdk/sources/streams/utils/stream_helper.py
erohmensing: I tried to name it so that it is more intuitive, but clearly not intuitive enough if we're asking the question! will add it to the function description

erohmensing: Optional would imply that `None` is a valid record, which I don't think is what we want

PR #21408, opened 2023-01-13
🪟 🐛 Fix jest tests to fail build correctly

timroes:
## What

Jests tests currently don't fail the `:airbyte-webapp:test` task correctly, since specifying `--watch` and `--watchAll=false` to jest will cause it to always return 0 exit codes, despite the `--watch` being overwritten otherwise in behavior by `--watchAll=false`.

## How

This creates a new `test:ci` npm task that will be called from gradle, to make sure we're not specifying both arguments.
To test you can make sure that `SUB_BUILD=PLATFORM ./gradlew :airbyte-webapp:test` will now correctly fail if there's a jest test failure.

### DIFF
--- a/airbyte-webapp/build.gradle
+++ b/airbyte-webapp/build.gradle
@@ -34,7 +34,7 @@ npm_run_build {
 task test(type: NpmTask) {
     dependsOn assemble
 
-    args = ['run', 'test', '--', '--watchAll=false', '--silent']
+    args = ['run', 'test:ci']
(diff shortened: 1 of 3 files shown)


krishnaglick (approved):
Failed locally w/ a failing test!

timroes:
Failure: https://github.com/airbytehq/airbyte/actions/runs/3913901513/jobs/6690366504#step:8:4208

PR #21410, opened 2023-01-13
fix loading remote definitions via cron

### LINKED ISSUE #21401: Airbyte Cron really does re-fetch new connector definitions every 30s
Details in this[ thread :thread:](https://airbytehq-team.slack.com/archives/C03VDJ4FMJB/p1673570100386949)

This PR seems to be the culprit https://github.com/airbytehq/airbyte/pull/21073

Previously we had been instantiating a new RemoteDefinitionsProvider on each cron invocation, but now we're injecting the dependency so it's only ever instantiated once. Definitions are loaded from the remote source when instantiating the class, so in practice we only load the new values once, when cron boots. (diff)

A fix could be to change this back to instantiate on each call, or the right solution is probably to make fetching the catalog an explicit method call rather than only doing it on the constructor - this way it's more clear when the fetching happens.

pedroslopez:
## What

fixes #21401 

After https://github.com/airbytehq/airbyte/pull/21073 moved the definitions provider to be injected as a singleton, the behavior of only fetching definitions in the RemoteDefinitionProvider on class initialization meant that we were no longer fetching the latest definitions each time the cron ran. 

## How

Refactor the RemoteDefinitionsProvider to let use leverage micronaut caching to determine when to re-fetch the data. Each getter now calls the method to fetch the data, but we're caching the results of the method for 15s on airbyte-cron.


evantahler (approved):
👍 

Just to check my understanding - in Cloud, there will be a PR to call `definitionsProvider.loadDefinitions()`?

airbyte-config/init/src/main/java/io/airbyte/config/init/LocalDefinitionsProvider.java
```diff
@@ -48,22 +48,26 @@ public LocalDefinitionsProvider(final Class<?> seedResourceClass) throws IOExcep
...
 
   // TODO will be called automatically by the dependency injection framework on object creation
-  public void initialize() throws IOException {
-    this.sourceDefinitions =
-        parseDefinitions(this.seedResourceClass, SeedType.STANDARD_SOURCE_DEFINITION.getResourcePath(), SeedType.SOURCE_SPEC.getResourcePath(),
-            SeedType.STANDARD_SOURCE_DEFINITION.getIdName(), SeedType.SOURCE_SPEC.getIdName(), StandardSourceDefinition.class);
-    this.destinationDefinitions = parseDefinitions(this.seedResourceClass, SeedType.STANDARD_DESTINATION_DEFINITION.getResourcePath(),
-        SeedType.DESTINATION_SPEC.getResourcePath(), SeedType.STANDARD_DESTINATION_DEFINITION.getIdName(), SeedType.DESTINATION_SPEC.getIdName(),
-        StandardDestinationDefinition.class);
+  public void loadDefinitions() {
```
gosusnp, line 55: Since local definitions are read from the resources, we could annotate here with `@PostConstruct`

airbyte-config/init/src/main/java/io/airbyte/config/init/LocalDefinitionsProvider.java
```diff
@@ -48,22 +48,26 @@ public LocalDefinitionsProvider(final Class<?> seedResourceClass) throws IOExcep
     this.seedResourceClass = seedResourceClass;
 
     // TODO remove this call once dependency injection framework manages object creation
-    initialize();
+    loadDefinitions();
```
gosusnp, line 51: We do not need this call here.

airbyte-config/init/src/main/java/io/airbyte/config/init/RemoteDefinitionsProvider.java
```diff
@@ -55,16 +55,20 @@ public RemoteDefinitionsProvider(@Value("${airbyte.platform.remote-connector-cat
...
-    final CombinedConnectorCatalog catalog = getRemoteDefinitionCatalog(this.remoteDefinitionCatalogUrl, this.timeout);
-    this.sourceDefinitions = catalog.getSources().stream().collect(Collectors.toMap(
-        StandardSourceDefinition::getSourceDefinitionId,
-        source -> source.withProtocolVersion(
-            AirbyteProtocolVersion.getWithDefault(source.getSpec() != null ? source.getSpec().getProtocolVersion() : null).serialize())));
-    this.destinationDefinitions = catalog.getDestinations().stream().collect(Collectors.toMap(
-        StandardDestinationDefinition::getDestinationDefinitionId,
-        destination -> destination.withProtocolVersion(
-            AirbyteProtocolVersion.getWithDefault(destination.getSpec() != null ? destination.getSpec().getProtocolVersion() : null).serialize())));
+  public void loadDefinitions() {
```
gosusnp, line 58: Add a @CachePut("remoteDefinitions") annotation here instead of `@PostConstruct`

all the getters in this provider should call `loadDefinitions()`

airbyte-config/init/src/main/java/io/airbyte/config/init/ApplyDefinitionsHelper.java
```diff
@@ -54,6 +54,8 @@ public void apply(final boolean updateAll) throws JsonValidationException, IOExc
       final DefinitionsProvider definitionsProvider = definitionsProviderOptional.get();
       final Optional<AirbyteProtocolVersionRange> currentProtocolRange = getCurrentProtocolRange();
 
+      definitionsProvider.loadDefinitions();
```
gosusnp, line 57: This would no longer be needed.

airbyte-config/init/src/main/java/io/airbyte/config/init/DefinitionsProvider.java
```diff
@@ -12,6 +12,8 @@
 
 public interface DefinitionsProvider {
 
+  void loadDefinitions();
```
gosusnp, line 15: This should remain internal to the implementation.

gosusnp (commented):
I am wondering if we could leverage micronaut caching here instead of requiring the client to decide when to actually load the definitions. I added comments trying to highlight what this could look like.
edit: link to [micronaut doc](https://guides.micronaut.io/latest/micronaut-cache-gradle-java.html)

The other missing piece would be updating the `application.yml` with:
```
micronaut:
  caches:
    remoteDefinitions: 
      charset: 'UTF-8'
      expire-after-write: 1min
```

@jdpgrailsdev, do you think that would be a valid approach to manage how frequently we refresh definitions?

jdpgrailsdev:
> I am wondering if we could leverage micronaut caching here instead of requiring the client to decide when to actually load the definitions. I added comments trying to highlight what this could look like. edit: link to [micronaut doc](https://guides.micronaut.io/latest/micronaut-cache-gradle-java.html)
> 
> The other missing piece would be updating the `application.yml` with:
> 
> ```
> micronaut:
>   caches:
>     remoteDefinitions: 
>       charset: 'UTF-8'
>       expire-after-write: 1min
> ```
> 
> @jdpgrailsdev, do you think that would be a valid approach to manage how frequently we refresh definitions?

@gosusnp Yes, but I think it is a little more involved.  The Micronaut Cache abstraction is just that -- an abstraction over the creation and management of an in memory cache, such as one you could create using Guava, Ehcache or Caffeine.  You would still need to mark the methods that will cache their return values with the `@Cachable` annotation.  This is also not a distributed cache, so its possible that different instances of the application may return different results depending on when they were called even with caching.

gosusnp (approved):

pedroslopez:
Ok! Moved to using micronaut to handle the caching and reverted to the previous interface. Set to cache for 15s (cron runs every 30s).

gosusnp:
@jdpgrailsdev, I agree, it is a restriction from having an in-memory cache, I don't think it is a blocker for the current use case. The goal is to refresh frequently and I don't think we want to have concurrent updates so it should remain restricted to a single instance.

PR #21411, opened 2023-01-13
Fix links in connector builder UI docs

lmossman:
## What
Fix the broken links in the connector builder UI docs.

## How
Apparently relative paths work differently in our docs page than they do in my local VS code markdown previewing tool. I've made these changes based on the locations that the links in our docs page was sending the user to, so they should be correct for that page.


sherifnada (approved):
Do you know why this didn't break the docusaurus build? 

It seems we already throw https://github.com/airbytehq/airbyte/blob/master/docusaurus/docusaurus.config.js#L16

Approving this PR assuming you tested it locally. But really this sort of bug shouldn't make it past CI. Could you create an issue for that?

sherifnada:
left some context: https://airbytehq-team.slack.com/archives/C046KQF5GBT/p1673652656319819

lmossman:
Created an issue: https://github.com/airbytehq/airbyte/issues/21438

PR #21416, opened 2023-01-13
Allow the OSS server only on OSS

benmoriceau:
## What
The server endpoints are being declared twice in cloud because it depends on the server.
Ideally we should extract the handler in their own module and remove the dependency from the server wrapped to the server to avoid having requirement but this is a temporary since we will have one server soon. So the "dirty" solution is acceptable.

Without that the wrapped server can't run its health check properly because of:
```
benoit@Benoits-MacBook-Pro-2 ~/w/airbuild (master|✔) [7]> k exec airbyte-server-64cb54bb9b-qb8nf -- curl http://localhost:8001/api/v1/health
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0{"message":"Bad Request","_links":{"self":{"href":"/api/v1/health","templated":false}},"_embedded":{"errors":[{"message":"More than 1 100   262  100   262    0     0  36449      0 --:--:-- --:--:-- --:--:-- 43666h: GET - /api/v1/health, GET - /api/v1/health"}]}}
```

### DIFF
--- a/airbyte-server/src/main/java/io/airbyte/server/apis/AttemptApiController.java
+++ b/airbyte-server/src/main/java/io/airbyte/server/apis/AttemptApiController.java
@@ -9,12 +9,15 @@ import io.airbyte.api.model.generated.InternalOperationResult;
 import io.airbyte.api.model.generated.SaveStatsRequestBody;
 import io.airbyte.api.model.generated.SetWorkflowInAttemptRequestBody;
 import io.airbyte.server.handlers.AttemptHandler;
+import io.micronaut.context.annotation.Requires;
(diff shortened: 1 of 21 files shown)


colesnodgrass (approved):

jdpgrailsdev (approved):
:shipit:

octavia-approvington:

octavia-approvington (approved):

PR #21417, opened 2023-01-13
Add query to determine whether a workspace has an alpha or beta connector

pmossman:
## What
The Free Connector Program info endpoint needs to return a boolean indicating whether or not a workspace contains at least one alpha or beta connector (The front-end will use this info to conditionally render a banner about the program).

This PR adds a method to the ConfigRepository to efficiently query for this info.

## How
Add a method that joins actor with actor_definition and counts records that contain an alpha or beta release stage. This is in OSS because all tables involved are OSS tables in the config DB. The Cloud API handler will call this method to serve the free_connector_program_info endpoint.


airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -1517,6 +1517,30 @@ public Geography getGeographyForConnection(final UUID connectionId) throws IOExc
         .fetchOneInto(Geography.class);
   }
 
+  /**
+   * Specialized query for efficiently determining eligibility for the Free Connector Program. If a
```
davinchia, line 1521: Side thought I want to confirm:

If a workspace user signs up for the free connector program and also has connections with both GA source and destination, we should only consider connections with alpha/beta sources/destinations as free yes?

If a user does not sign up, but already has payment information with us, are they automatically enrolled in the free program?

We might have covered this already. So sorry if I missed it. Checking my understanding.

davinchia (approved):

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
pmossman: > If a workspace user signs up for the free connector program and also has connections with both GA source and destination, we should only consider connections with alpha/beta sources/destinations as free yes?

Yes, the credit processing cron will only set free_usage: true on Orb events for jobs that contain an alpha/beta connector.

> If a user does not sign up, but already has payment information with us, are they automatically enrolled in the free program?

They are not. Currently our Stripe terms of use for credit payments only allow a single charge to complete that one purchase. So to enroll for the program, they need to explicitly step through the 'setup' mode Stripe checkout which contains a broader agreement for allowing Airbyte to charge that payment in the future

PR #21421, opened 2023-01-13
🪟 🔧 Migrate from react-scripts to Vite

timroes:
## What

Removed react-scripts (Create React App) as a build system and instead use Vite.

## Why?

Because it's :zap: faster, :gear: more configurable and way better :construction_worker: maintained.

## How

I've tried to document most files inline to give context why changes have happened. Some general remarks:

### :closed_lock_with_key: HTTPS required

Since Vite loads modules as individual files during development, we want to make sure to use HTTP2 to not run into parallel file loading limits of HTTP1. Thus HTTPS is required now also during development, i.e. the development page will now be available via `https://localhost:3000` or `https://127.0.0.1:3000`. Since this will use self-signed certificates you need to ignore an "unsecure" warning in your browser. Due to the way Vite does caching it's also not necessary to disable HTTP caching during development and will be faster if this setting is off in your Dev Tools.

### No code changes and further work

There has been **no code change** on actual source code for this. The only difference would be that Vite used `import.meta.env` instead of `process.env`. Since we still run tests via babel and Storybook via Webpack, I made sure `process.env` continues to work. That allows us to migrate Storybook and Jest in separate PRs, since both of them are quiet a bit of work to convert (and for Storybook we likely want to wait for the 7.0 release).

### :mag: Code quality scans

Stylelint, ESLint and TypeScript runs in Vite (via the checker plugin) async in different workers. That means you might see warnings/errors coming from those in the UI with a bit of a delay, because the actual code already is delivered immediately, while the checks are still running. The new error overlay also doesn't pop up by default, but can be opened via a button on the bottom (Errors will still be printed to the terminal as well):

screenshot-20230117-120844

## Testing

I've validated that ESLint (warnings as well as errors), Stylelint and TypeScript will fail the build with a non zero status code, to make sure they won't pass CI. Also tested Cloud mode and local development working, as well as OSS mode building proper docker images.


airbyte-webapp/src/hooks/useSegment.tsx
```diff
@@ -1,3 +1,5 @@
+/// <reference types="@types/segment-analytics" />
```
timroes, line 1: :information_source: Without `types` in the `tsconfig.json` (which was the case beforehand) all `@types/...` packages get automatically imported. That's not the case anymore now, so we'll reference the only missing one in this file explicitally.

airbyte-webapp/build.gradle
```diff
@@ -106,16 +106,8 @@ task copyNginx(type: Copy) {
     into "build/docker/bin/nginx"
 }
 
-task stylelint(type: NpmTask) {
```
timroes, line 109: :information_source: Stylelint now runs as part of the Vite (via the checker plugin). So it doesn't need a separate Gradle task anymore.

airbyte-webapp/package.json
```diff
@@ -125,26 +124,28 @@
...
     "express": "^4.18.1",
     "husky": "^8.0.1",
     "jest": "^29.3.0",
     "jest-environment-jsdom": "^29.3.1",
     "license-checker": "^25.0.1",
     "lint-staged": "^12.3.7",
-    "mini-css-extract-plugin": "^2.6.1",
+    "meow": "^9.0.0",
     "node-fetch": "^2.6.7",
+    "optionator": "^0.9.1",
```
timroes, line 146: :information_source: Peer dependencies required by the styellint and eslint plugin for vite checker.

airbyte-webapp/package.json
```diff
@@ -125,26 +124,28 @@
...
     "orval": "^6.11.0-alpha.10",
     "prettier": "^2.6.2",
-    "react-scripts": "^5.0.1",
```
timroes, line 147: :wave: bye, bye, you will for sure not be missed.

timroes:
Successfull failure on eslint warning: https://github.com/airbytehq/airbyte/actions/runs/3939246270/jobs/6738917898

timroes:
Currently blocked on https://github.com/airbytehq/airbyte/pull/20887

dizel852:
Unblocked - just merged https://github.com/airbytehq/airbyte/pull/20887 🎉

flash1293:
I'm probably doing something wrong, but if that's the case it's easy to run into it. This is what I did:
* Check out this branch
* Run `npm install`
* Go to https://localhost:3000/ (need to allow unsafe connection)
* Fails like this:
Screenshot 2023-01-20 at 13 59 53

timroes:
Thanks for catching this. This is actually a valid thing I need to fix. Currently we set `apiUrl` to the same procotol in the `defaultConfig`, which we shouldn't since the local backend server does only listen to http. Will address that.

josephkmh:
CORS issue is fixed in https://github.com/airbytehq/airbyte/pull/21668, merged that in locally while I'm testing and request from `https://localhost:3000` are working well again 👍

timroes:
@flash1293 This should be solved now. You'll need to rebuild the server though (`:airbyte-server:assemble`) and restart your docker-compose to get it solved.

airbyte-webapp/src/config/defaultConfig.ts
```diff
@@ -4,8 +4,8 @@ const defaultConfig: Config = {
   segment: { enabled: true, token: "" },
   healthCheckInterval: 20000,
   version: "dev",
-  apiUrl: `${window.location.protocol}//${window.location.hostname}:8001/api`,
-  connectorBuilderApiUrl: `${window.location.protocol}//${window.location.hostname}:8003`,
+  apiUrl: `http://${window.location.hostname}:8001/api`,
+  connectorBuilderApiUrl: `http://${window.location.hostname}:8003`,
```
timroes, line 8: @josephkmh Could you make sure you also make those adjustments in https://github.com/airbytehq/airbyte/pull/21456. It should be fine merging them also before Vite, so it shouldn't matter in which order we'll merge the PRs eventually.

josephkmh: Added here https://github.com/airbytehq/airbyte/pull/21456/commits/8e751cf65c9139c6b2d0181d626a73fd993dd40d

joeykmh (commented):
Looks good to me! I tested locally in dev mode on `https://localhost:3000` and everything worked smoothly. Same with a fresh build served via docker on `http://localhost:8000`

flash1293 (approved):
LGTM; tested and seems to work fine.

I have one complaint, but I'm pretty sure that happened before, so no need to fix it on this PR: When defining a class in the scss module which is never referenced, it shows the expected error:
```
 ERROR(ESLint)  Unused classes found in BuilderField.module.scss: errorXXX, abc (css-modules/no-unused-class)
 FILE  /Users/joereuter/Clones/airbyte/airbyte-webapp/src/components/connectorBuilder/Builder/StreamReferenceField.tsx:10:8

     8 |
     9 | import { BuilderStream } from "../types";
  > 10 | import styles from "./BuilderField.module.scss";
```

However, when fixing the problem, this error doesn't go away - it sticks around until I restart the dev server.

timroes:
> However, when fixing the problem, this error doesn't go away - it sticks around until I restart the dev server.

This is more related to eslint and indeed currently present. You should though not need to actually kill the dev server, but you'll need to save the TS file that includes that SASS again. Since this rule is part of ESLint, it won't rerun if something in the SASS file changes, only when the TS file actually has a change.

krishnaglick (approved):

timroes:
The only check that failed, was tearing down the CI runner in the end, so treating this CI run as successful for merging.

PR #21424, opened 2023-01-13
🪟 🔧 remove cyclic dependencies with cloud routes

teallarson:
## What
Removes cyclic dependency in our cloud routes to unblock part of https://github.com/airbytehq/airbyte/pull/21421

## How
Updates our cloud routes to use `cloudRoutePaths` in the same manner that OSS uses `routePaths`

## Recommended reading order
1. `cloudRoutePaths.tsx`
2. others

### DIFF
--- a/airbyte-webapp/src/packages/cloud/cloudRoutePaths.tsx
+++ b/airbyte-webapp/src/packages/cloud/cloudRoutePaths.tsx
@@ -1,0 +1,19 @@
+export enum CloudRoutes {
+  Root = "/",
+  AuthFlow = "/auth_flow",
+
+  Metrics = "metrics",
+  SelectWorkspace = "workspaces",
+  Credits = "credits",
+
+  // Auth routes
+  Signup = "/signup",
+  Login = "/login",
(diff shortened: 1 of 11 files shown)


airbyte-webapp/src/packages/cloud/cloudRoutePaths.tsx
```diff
@@ -0,0 +1,19 @@
...
+  Signup = "/signup",
+  Login = "/login",
+  ResetPassword = "/reset-password",
+
+  // Firebase action routes
+  // These URLs come from Firebase emails, and all have the same
+  // action URL ("/verify-email") with different "mode" parameter
+  // TODO: use a better action URL in Firebase email template
+  FirebaseAction = "/verify-email",
+}
```
tealjulia, line 19: I can't think of a reason this needs to be an object `as const` as opposed to an enum (and the OSS `routePaths` is an enum so it seems right?)

timroes (approved):
Code LGTM, tested locally for some routes and nothing seems to be broken by the change.

PR #21428, opened 2023-01-13
🪟 🔧 Run FE e2e tests on separate runner

timroes:
## What

This runs the frontend end2end tests on a separate runner, so the Frontend Build and End2End test shop won't be serialized anymore (as they are today), but instead can run in parallel, since they have no need for serialization.


.github/workflows/gradle.yml
```diff
@@ -442,11 +442,79 @@ jobs:
...
+            ${{ secrets.DAVINCHIA_PAT }}
+      - name: Start AWS Runner
+        id: start-ec2-runner
+        uses: ./.github/actions/start-aws-runner
+        with:
+          aws-access-key-id: ${{ secrets.SELF_RUNNER_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.SELF_RUNNER_AWS_SECRET_ACCESS_KEY }}
+          github-token: ${{ env.PAT }}
+
   frontend-test:
```
josephkmh, line 513: Should we name this `frontend-e2e-test`? If I understand correctly, unit tests (`jest`) are running during the `frontend-build` action via gradle, right?

joeykmh (approved):
Looks good - just had one suggestion to make the test workflow more explicit

davinchia (approved):
Thanks @timroes sorry I haven't had time to follow up yet. Busy with the performance work and roadmap planning!

PR #21430, opened 2023-01-13
Fix CreateConnectionForm test snapshot

edmundito:
## What
Fixes a broken jest snapshot that causes the webapp unit tests to fail.

Caused by #21352
Related: #21408

## How
Updates snapshot


teallarson (approved):
lgtm assuming CI is green!

Issue #19410, opened 2022-11-15
[Low-Code CDK] Remove request_options_provider from the manifest language


brianjlai:
## Tell us about the problem you're trying to solve
**BREAKING CHANGE**

The `request_options_provider` component was originally implemented to organize all relevant request parameter fields under a common component during sync processing. However, from the perspective of a developer writing a manifest, this just feels like an arbitrary component that doesn't provide much value. 

We should promote the following fields from the `request_options_provider` component to the parent `http_requester` component and remove `request_options_provider` from the language:

- request_parameters
- request_headers
- request_body_data
- request_body_json

## Describe the solution you would like

A sample of the manifest for the `http_request` component is as follows:
```
requester:
  url_base: "https://sample-state.com"
  http_method: "GET"
  authenticator:
    type: BearerAuthenticator
    api_token: "{{ config['api_key'] }}"
  request_parameters:
    created_after: "{{ config['start_date'] }}"
  request_headers:
    company_id: "{{ config['company_id'] }}"
  request_body_json:
    body_key: "body_val"
  request_body_data:
    body_data: "body_str_val"
```

Any documentation related to the request_options_provider should be removed and any code examples should be modified to reflect the various request options being defined the requester itself.

Given that very few connectors actually use `request_options_provider`, it might be simpler to just un-indent the various fields and remove the `request_options_provider` field.

The low-code generator templates don't have `request_options_provider` defined so they should not need to be modified.

## Acceptance Criteria

- [ ] request_parameters, request_headers, request_body_json, and request_body_data should be fields of the requester and not be nested underneath a request_options_provider
- [ ] Existing low-code connector manifests should reflect the new changes (either via migration or handwritten)
- [ ] documentation and code samples should be updated accordingly
- [ ] Update connector builder

## Implementation Approach
Under the hood, we should retain the `InterpolatedRequestOptionsProvider` class because it still useful for consolidate request input evaluation under a common object for organization purposes. However, we can avoid forcing developers to specify it in the language by changing the `HttpRequester` dataclass. We can replace the `request_options_provider` field with fields for `request_parameters`, `request_headers`, `request_body_data`, and `request_body_json`. And `__post_init__()` we can use the inputs to instantiate an `InterpolatedRequestOptionsProvider`. That allows us to still use the structure behind the scenes but not expose it to developers.

PR #21434, opened 2023-01-13
🚨🚨 [Low Code CDK] Update `*ref` format to `#/`

clnoll:
Closes [19409](https://app.zenhub.com/workspaces/connector-extensibility-6257455decf3920012f4e872/issues/gh/airbytehq/airbyte/19409).

## What
Updates the reference format used in manifests from our custom format (`*ref(.*)`, with `.`-separated path components) to the format used by JSON Schema (`#/...`, with `/`-separated path components).

## How
Updates the manifest reference resolver to resolve references starting with `#/`.

Also includes a script to update all manifests to use the new reference format. When run, it will update 91 connectors to use the new format.

Note: with our original reference format, we were also handling references that index into a list element, which took a bracket notation (e.g. `*ref(definitions.mycomponent[0])`). In JSON Schema references, the brackets would be replaced with `/` (e.g. mycomponent/0). However, I didn't see any existing connectors using this feature so for the sake of simplicity haven't included it in the script.

## 🚨 User Impact 🚨
All manifests that use references will need to be updated to use the new reference format or else their references will not be resolved. The current thinking is that we will run the script and update the manifests, but won't publish a new version of the connector; the updated manifest will be present whenever the next update is ready for publishing.

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
@@ -13,10 +13,10 @@ class ManifestReferenceResolver:
     An incoming manifest can contain references to values previously defined.
     This parser will dereference these values to produce a complete ConnectionDefinition.
 
-    References can be defined using a *ref(<arg>) string.
+    References can be defined using a #/<arg> string.
     ```
     key: 1234
-    reference: "*ref(key)"
+    reference: "#/key"
     ```
     will produce the following definition:
     ```
@@ -28,7 +28,7 @@ class ManifestReferenceResolver:
     key_value_pairs:
       k1: v1
       k2: v2
-    same_key_value_pairs: "*ref(key_value_pairs)"
+    same_key_value_pairs: "#/key_value_pairs"
     ```
     will produce the following definition:
     ```
@@ -46,7 +46,7 @@ class ManifestReferenceResolver:
       k1: v1
       k2: v2
     same_key_value_pairs:
-      $ref: "*ref(key_value_pairs)"
+      $ref: "#/key_value_pairs"
       k3: v3
     ```
     will produce the following definition:
@@ -66,7 +66,7 @@ class ManifestReferenceResolver:
     ```
     dict:
         limit: 50
-    limit_ref: "*ref(dict.limit)"
+    limit_ref: "#/dict/limit"
     ```
     will produce the following definition:
     ```
@@ -75,18 +75,18 @@ class ManifestReferenceResolver:
     limit-ref: 50
     ```
 
-    whereas here we want to access the `nested.path` value.
+    whereas here we want to access the `nested/path` value.
     ```
     nested:
         path: "first one"
-    nested.path: "uh oh"
-    value: "ref(nested.path)
+    nested/path: "uh oh"
+    value: "#/nested/path
     ```
     will produce the following definition:
     ```
     nested:
         path: "first one"
-    nested.path: "uh oh"
+    nested/path: "uh oh"
     value: "uh oh"
     ```
 
@@ -118,7 +118,7 @@ class ManifestReferenceResolver:
                 return evaluated_dict
         elif isinstance(node, list):
             return [self._evaluate_node(v, manifest, visited) for v in node]
-        elif isinstance(node, str) and node.startswith("*ref("):
+        elif self._is_ref(node):
             if node in visited:
                 raise CircularReferenceException(node)
             visited.add(node)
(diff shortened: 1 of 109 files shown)


airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
```diff
@@ -75,40 +77,38 @@ class ManifestReferenceResolver:
...
         :return:
         """
         return self._evaluate_node(manifest, manifest, set())
 
-    def _evaluate_node(self, node: Any, manifest: Mapping[str, Any], visited: Set):
+    def _evaluate_node(self, node: Any, manifest: Mapping[str, Any], visited: Set) -> Any:
         if isinstance(node, dict):
             evaluated_dict = {k: self._evaluate_node(v, manifest, visited) for k, v in node.items() if not self._is_ref_key(k)}
-            if self.ref_tag in node:
+            if REF_TAG in node:
```
sherifnada, line 109: just a q: why are we keeping backwards compatibility?

clnoll: We are no longer supporting the `*ref` format, so aren't keeping backwards compatibility there. The `$ref` format is supported by JSON Schema so we're still using it to identify refs whose value should be splatted into the referencer.

sherifnada: ah got it. Thanks for the 🥁 pointer 🥁

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
```diff
@@ -7,16 +7,18 @@
 
 from airbyte_cdk.sources.declarative.parsers.custom_exceptions import CircularReferenceException, UndefinedReferenceException
 
+REF_TAG = "$ref"
```
maxi297, line 10: Why extract this outside of the class boundaries? Shouldn't this be private to `ManifestReferenceResolver`?

maxi297 (approved):
LGTM! When will we migrate the manifest using the migration script?

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_reference_resolver.py
clnoll: I could go either way. It allowed me to clean things up a little - making both `is_ref` and `is_ref_key` `staticmethod`s, plus I didn't feel strongly that it should be owned by `ManifestReferenceResolver`. That said, it is only _used_ by `ManifestReferenceResolver`, so I can see an argument for keeping it private.

clnoll:
> When will we migrate the manifest using the migration script?

The migration will be done in a separate PR. Debating between waiting until other breaking changes are done and doing all of them in one pass to keep the connectors' commit histories cleaner, or running it just after we merge to `low_code_cdk_to_beta` so the commit histories of each connector explicitly state each breaking change that was applied to them. Leaning toward the latter, but WDYT?

cc @maxi297 @brianjlai

brianjlai (approved):
changes look good to me. nice work!

Have you run the `./gradlew validateSourceYamlManifest` command which verifies the `spec` command still works for all low code connectors? And it might be also worth testing a `read` on 1-2 connectors as a sanity check if you haven't already done so.

with regards to the migration, yeah I think I'm also leaning towards running the migration on the base `low_code_cdk_to_beta` branch. Just makes the reviews a little easier to manage and we can run the included script which can constitute individual commits as we look at the final branch

brianjlai:
I'm also not sure what's going on w/ the Platform and Frontend CI checks. Strange

clnoll:
@brianjlai just want to confirm that `./gradlew validateSourceYamlManifest` passes and the spot check of `read`s passed too.

clnoll:
@maxi297 would you mind taking a look at my latest commit, [bdf1446](https://github.com/airbytehq/airbyte/pull/21434/commits/bdf1446f929df05cf7b9931312f181de85c8fd44)? I added it because I was getting a [build failure](https://github.com/airbytehq/airbyte/actions/runs/3962117292/jobs/6788427876#step:8:1024) with an error, `TS2305: Module '"core/request/ConnectorManifest"' has no exported member 'InterpolatedRequestOptionsProvider'.`, which I think is related to https://github.com/airbytehq/airbyte/issues/19410.

brianjlai:
I'm thinking this is because the frontend hasn't been updated to reflect the now removed component since we got rid of it in the manifest. I think FE auto-generates their type definitions from the same manifest which is now missing that component. We need to file a ticket so Lake and Joe are aware, which should now be part of our acceptance criteria. But if we want to focus on building these faster, we can probably just ignore these tests when we merge to the base branch since we can tolerate the tests not working in the immediate until we update the front end code.

clnoll:
Okay thanks @brianjlai in that case I'll revert [bdf1446](https://github.com/airbytehq/airbyte/pull/21434/commits/bdf1446f929df05cf7b9931312f181de85c8fd44).

PR #21437, opened 2023-01-13
Protocol Change: Suggested Streams

### LINKED ISSUE #10872: Create a notion of streams which are "selected by default"
A connector like JIRA may have 100 streams. Most users typically care about 20 of those. But since all streams are selected by default, a user has to specifically opt out of the streams they don't care about. Since the path of least resistance is do nothing/keep all streams selected, user syncs often end up with way more data than they care about or need. 

The downside for this is: 
1. Increased sync times, sometimes significantly e.g: nested REST resources which may take a long time may be unnecessarily slowing down the sync.
2. More API quota usage for the user

When setting up a connection, expose a way to display the streams which are included by default in the sync, and have all others deselected. 

Thus, when a user goes to setup a connector in a connection, some streams are selected by default and others need to be opted into by the user. 

This could be implemented at the protocol/connector level (e.g adding a `selected_by_default` keyword in the catalog output from discovery) or it could be at the server level. Not sure

evantahler:
Closes https://github.com/airbytehq/airbyte/issues/10872, (inspired by https://github.com/airbytehq/oncall/issues/1303#issuecomment-1381772903 OC issue).  [Protocol Change Process Document](https://docs.google.com/document/d/1RnoFFr12ZX0WGsjicyC_ErvSib5aP8A0DZ-IPzZcB1g/edit#).

## What?

This PR allows connectors to suggest which streams are pre-selected for users creating a new connection.  This is accomplished by adding a new “suggested" (boolean) property to each stream returned by the DISCOVER command, and a top-level “suggestingStreams" (boolean) to opt-into the feature.  This change is backwards-compatible and does not require the re-publishing of existing connectors. 

For many of our API sources, the connector provides *a lot* of streams.  This is great because Airbyte connectors come with "batteries included" to enable the widest range of data-syncing use-cases.  However, for some sources, most users probably don't want all of the streams.  Some of those streams may be extraneous, but worse, they may be slow, expensive, and often make the syncing experience for the more "primary" streams worse. Perhaps this stream has a slower API, consumes more API "credits", or harms the global rate limit in a worse-than-normal way.   

This PR allows each connector to tune the default stream selections to provide a better first-time sync experience.  

Screenshot 2023-01-13 at 3 47 38 PM

> The Convenience Principle: People will do what's convenient, and they will avoid what's inconvenient. All else being equal, to increase how many people do something, make it more convenient, and to decrease, make it less convenient
> - @sherifnada 

This PR comes with an example for `soruce-github`, a source with particularly egregious streams, which we don't think many people would opt-into.

## Technical notes:

* This is an Airbyte Protocol change, adding new optional parameters to the response of a DISCOVER command. 
  * `AirbyteCatalog` gains `suggestingStreams(bool)` - Does this catalog suggest a subset of streams to be selected when discovered?  If false or null, the previous behavior of pre-selecting all streams should be used.
  * `AirbyteStream` gains `suggested(bool)` - Should this stream be enabled by default?  This property is considered when the catalog suggesting this stream has suggestingStreams=true
* No API or webapp changes needed
* All existing connectors will work the same way - defaulting to selecting all streams for a new connection
* The CDK updates maintain the default behavior - defaulting to selecting all streams for a new connection
* All that connector developers need to do is supply `suggested = False` on their streams when desired.
 
## Notes for @evantahler 

This PR requires a publication of the CDK and source-github


airbyte-integrations/connectors/source-github/source_github/streams.py
```diff
@@ -319,6 +319,8 @@ class RepositoryStats(GithubStream):
     API docs: https://docs.github.com/en/rest/reference/repos#get-a-repository
     """
 
+    suggested = False
+
```
evantahler, line 323: This is all it takes to demote a CDK stream to un-suggested.

airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
```diff
@@ -105,22 +105,34 @@ public static io.airbyte.api.model.generated.AirbyteCatalog toApi(final io.airby
...
         .selected(true);
+
+    if (catalog.getSuggestingStreams()) {
+      final io.airbyte.protocol.models.AirbyteStream modelStream = catalog.getStreams().stream()
+          .filter(s -> s.getName().equals(stream.getName()) && (s.getNamespace() == null || s.getNamespace().equals(stream.getNamespace())))
+          .findFirst().orElseThrow();
+      if (!modelStream.getSuggested()) {
+        result.setSelected(false);
+      }
+    }
```
evantahler, line 128: There's probably a better ☕ way to do this... Also, the difference between protocol models and api models was confusing.

airbyte-protocol/protocol-models/src/main/resources/airbyte_protocol/airbyte_protocol.yaml
```diff
@@ -328,6 +331,9 @@ definitions:
       namespace:
         description: Optional Source-defined namespace. Currently only used by JDBC destinations to determine what schema to write to. Airbyte streams from the same sources should have the same namespace.
         type: string
+      suggested:
```
sherifnada, line 334: naming nitpick: can we call it `sync_by_default` or `enabled_by_default`? makes the name more self explanatory

sherifnada (changes requested):
exciting! LGTM minus one request about naming

evantahler:
Closing in favor of https://github.com/airbytehq/airbyte/pull/21577, which moves the storage of SuggestedStreams to the Actor Definition

PR #21443, opened 2023-01-14
Don't return stacktrace as part of the response

girarda:
## What
* Remove the stacktrace from the connector builder's error response

Before
Screen Shot 2023-01-13 at 5 27 16 PM

After:
Screen Shot 2023-01-13 at 5 38 21 PM


lmossman (approved):
LGTM

brianjlai (approved):
lgtm!

Issue #18741, opened 2022-10-31
Add acceptance test for correct format on date fields


timroes:
We currently don't set the `format` property on all `type: string` fields in connectors that accept a date. The `format` property has currently no effect, but will be once we have date pickers in the UI, since it's the one determining if a string field is actually a date.

We should add an acceptance test, that tests that:

a) if a `type: string` field has the pattern `^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$`, that it must also have a `format: date-time`
b) if a `type: string` field has the pattern `^[0-9]{4}-[0-9]{2}-[0-9]{2}$`, it must also have `format: date`

erohmensing:
@timroes as far as I understand, the current state is:
* When a field should be a date, it's stored as a `string`
* In order to help users input the correct format for this string, we define a `pattern` in the spec, and use that regular expression to validate on the frontend
* The `format` field is available to use in the spec and readable/usable from the frontend
* We may or may not (?) already set the `format` field either manually or injected in the frontend (this one needs clarifying) 

I think you're asking for:
* If we read that a `string` field has a `pattern` that we (somewhat arbitrarily) think is a date, SAT should fail and specify that that connector needs a `format: <date/date-time>` field

What I think you're trying to get to is "If something should explicitly be a date, it needs to be marked with the `format: date` field" so that we can use the date-picker. Makes sense to me but I don't think we need to look at the regexes for that, we need to update the existing connectors and the existing documentation. As far as I can tell, the regexes are only in place because we hadn't implemented the datepicker yet! Being explicit about definining what's a date can avoid accidentally pulling up the date picker for a field that only looks (via regex) like a date. 

Here's what I'd suggest:

* Introduce `format: date` and `format: date-time` if frontend needs to distinguish these two for date picker and date-time picker
* Optional: If we need to dump the date from the frontend into a specific format, introduce a field where we let the connector specify which date format it likes (e.g. `YYYY-MM-DD` out of a list of ones we support/can dump from the frontend based on whatever library we do to do the dumping). If this were the case we probably don't need both `date` and `date-time`.
* Wherever our current connectors use the `pattern` to indicate dates currently, manually update those to use one of the `format`s instead, removing the `pattern`. 
[...]

PR #21451, opened 2023-01-16
Extend SATs to capture UI limitations

### LINKED ISSUE #20063: SAT: Add tests for all JSON schema features that are not supported by UI
JSON schema is more flexible than what the current connector form UI can handle. However, it's currently not clearly documented what will work and what won't, so a connector developer is left with trial and error.

The SATs should check whether the spec is "UI-compliant" and the documentation should mention that test.

Tasks
* Find all features not supported today. List so far:
   * Pattern properties https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties
   * allOf https://json-schema.org/understanding-json-schema/reference/combining.html#allof
   * anyOf https://json-schema.org/understanding-json-schema/reference/combining.html#anyof
   * not https://json-schema.org/understanding-json-schema/reference/combining.html#not
   * if-then-else https://json-schema.org/understanding-json-schema/reference/conditionals.html#if-then-else
   * dependentSchemas https://json-schema.org/understanding-json-schema/reference/conditionals.html#id5
   * special date and date-time cases (there should be a pattern defined along with the format)
* Extend SATs
* Reference in documentation

flash1293:
Fixes https://github.com/airbytehq/airbyte/issues/20063

Changes to the SATs:
* Only strings, integers and numbers can be secrets, arrays and objects do not work correctly in the UI. There was an existing related test already which would also accept objects and arrays - made it stricter to not accept them anymore
* "type" can't be an array, except for the case where it's two values and the second one is "null" - all other cases are not handled by the UI
* The array "items" type can't be an array
* The array "items" type either needs to be unspecified which defaults to string, explicitly string, object, number, integer or has an enum defined (other cases won't work in UI)
* not, anyOf, itemPrefixes, patternProperties, allOf, if, then, else, dependentSchemas and dependentRequired is not allowed as it won't be rendered as expected
* If format: date/date-time is specified, it should have the corresponding pattern set as well and the other way around - this is only a warning as it's not a hard requirement and the UI still works (also enforcing this would require a lot of breaking changes)
* An object or array type always needs a valid definition of the sub types (items in case of array, properties in case of object)

There were a few connectors with problematic specs.
For problems that were present before this PR I opened an issue here: https://github.com/airbytehq/airbyte-internal-issues/issues/1291
For new problems I set up a PR here: https://github.com/airbytehq/airbyte/pull/21587


girarda (commented):
@alafanechere is there any documentation that should be updated when adding new SATs?

flash1293:
@girarda Good point, I will set up a separate PR for that!

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,149 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+        """
+        Properties with format date or date-time should always have a pattern defined how the date/date-time should be formatted
+        that corresponds with the format the datepicker component is creating.
+        """
+        specification = connector_spec_dict["connectionSpecification"]
+        for format_path, format in dpath.util.search(specification, "**/format", yielded=True):
+            if not isinstance(format, str):
+                # format is not a format definition here but a property named format
+                continue
+            if format == "date" or format == "date-time":
```
girarda, line 390: nit: this if statement is superfluous since we compare the format string on line 393 and 397

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,149 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+                property_definition = self._get_parent(specification, format_path)
+                pattern = property_definition.get("pattern")
+                if format == "date" and not pattern == DATE_PATTERN:
+                    detailed_logger.warning(
```
girarda, line 394: is the plan to migrate existing connectors to always use these patterns and then change this warning to a failure?

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,149 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+        for pattern_path, pattern in dpath.util.search(specification, "**/pattern", yielded=True):
+            if not isinstance(pattern, str):
+                # pattern is not a pattern definition here but a property named pattern
+                continue
+            if pattern == DATE_PATTERN or pattern == DATETIME_PATTERN:
+                property_definition = self._get_parent(specification, pattern_path)
+                format = property_definition.get("format")
+                if not format == "date" and pattern == DATE_PATTERN:
+                    detailed_logger.warning(
+                        f"{pattern_path} is defining a pattern that looks like a date without setting the format to `date`. Consider specifying the format to make it easier for users to edit this field in the UI."
```
girarda, line 416: should this fail? What would be a valid format that would trigger a warning?

girarda (approved):
lgtm. Added a couple of non-blocking questions and requested a review from Augustin in case he has concerns about the new tests

girarda (changes requested):
synced offline: arrays of integers are also valid

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,149 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
 
+    def test_property_type_is_not_array(self, connector_spec_dict: dict):
+        """
+        Each field has one or multiple types, but the UI only supports a single type and optionally "null" as a second type.
+        """
+        specification = connector_spec_dict["connectionSpecification"]
+        errors = []
+        for type_path, value in dpath.util.search(specification, "**/properties/*/type", yielded=True):
+            if isinstance(value, List):
+                if len(value) != 2:
```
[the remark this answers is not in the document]
flash1293: I think it would work but I would like to be strict here as there isn't a good reason type would just be a single-value list. One case less to worry about.

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,149 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+        absolute_path = f"/{path}"
+        parent_path, _ = absolute_path.rsplit(sep="/", maxsplit=1)
+        if parent_path == "":
+            return obj
+        return dpath.util.get(obj, parent_path)
+
+    def test_object_not_empty(self, connector_spec_dict: dict):
+        """
+        Each object field needs to have at least one property as the UI won't be able to show them otherwise.
+        If the whole spec is empty, it's allowed to have a single empty object at the top level
```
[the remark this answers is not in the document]
flash1293: I started with this but actually we have some today, e.g. https://github.com/airbytehq/airbyte/blob/e3ff75f3d6c4ce4b418dac8cb5e97f5ae1941fb3/airbyte-integrations/connectors/source-whisky-hunter/source_whisky_hunter/spec.yaml#L7

I can see how this is an edge case, but it might be handy for very simple cases where it's just about loading a json response from an endpoint.

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
flash1293: It would be nice for consistency, but it felt like too much at the moment. I will create an issue for this.

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
flash1293: I was on the fence about this, too - I decided this way because of the following reasoning:
* Showing a date picker for a field instead of requiring the user to type it out is more of a presentational/convenience thing than a "just plain wrong/not configurable at all in the UI" thing (like the other checks)
* Semantically, a pattern doesn't necessarily mean it's a date - it is very likely for this specific pattern (hence the warning), but is it the right call the enforce this relationship?
* One example could be a sort of serial number which has the same format, but is semantically not a date - you wouldn't want to force a connector developer to render a date picker for that.

This is definitely not a strong reasoning, I will put it in the issue I mentioned above about the check the other way.

flash1293:
Thanks for the review @girarda ! I updated the checks here and made integers/number valid for arrays

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,148 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
                     {properties}"""
             )
 
+    def test_property_type_is_not_array(self, connector_spec_dict: dict):
+        """
+        Each field has one or multiple types, but the UI only supports a single type and optionally "null" as a second type.
+        """
+        specification = connector_spec_dict["connectionSpecification"]
+        errors = []
+        for type_path, value in dpath.util.search(specification, "**/properties/*/type", yielded=True):
```
alafanechere, line 286:
```suggestion
        for type_path, type_value in dpath.util.search(specification, "**/properties/*/type", yielded=True):
```
nit: for consistency

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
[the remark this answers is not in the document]
alafanechere: I would prefer to still allow single value list. I agree there's no good reason to use it. 
But this might cause connector build failures on existing connectors that our teams will have to fix.
If it works with a single value list, let's not make the acceptance tests fail.

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,148 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+            if isinstance(value, List):
+                if len(value) != 2:
+                    errors.append(f"{type_path} is not either a simple type or an array of a simple type plus null: {value}")
```
alafanechere, line 289: I would suggest adding an example in the error message to show what an array of a simple type plus null is :D I know it sounds obvious, but the more explicit we make the errors the less we receive community solicitations for troubleshooting.

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,148 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+                if value[1] != "null":
```
alafanechere, line 290: This condition will likely require an update if you agree to allow single value lists.

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,148 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+    @staticmethod
+    def _get_parent(obj: dict, path: str) -> Any:
+        """
+        Returns the parent dict of a given path within the `obj` dict
+        """
+        absolute_path = f"/{path}"
+        parent_path, _ = absolute_path.rsplit(sep="/", maxsplit=1)
+        if parent_path == "":
+            return obj
+        return dpath.util.get(obj, parent_path)
```
alafanechere, line 306: nit: this can be reused in other context. Feel free to move it to this [module](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/utils/json_schema_helper.py#L13).

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,148 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
                     {properties}"""
             )
 
+    def test_property_type_is_not_array(self, connector_spec_dict: dict):
+        """
+        Each field has one or multiple types, but the UI only supports a single type and optionally "null" as a second type.
+        """
+        specification = connector_spec_dict["connectionSpecification"]
```
alafanechere, line 284: `connectionSpecification` could be a class fixture as you reuse it quite often. Scoping it to `class` will avoid reloading it for each test.

```suggestion
    def test_property_type_is_not_array(self, connection_specification: dict):
        """
        Each field has one or multiple types, but the UI only supports a single type and optionally "null" as a second type.
        """
```

You can declare it in the class with:
```python
@pytest.fixture(scope="class")
def connection_specification(self, connector_spec_dict):
    return connector_spec_dict["connectionSpecification"]
```

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,148 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+        return dpath.util.get(obj, parent_path)
+
+    def test_object_not_empty(self, connector_spec_dict: dict):
+        """
+        Each object field needs to have at least one property as the UI won't be able to show them otherwise.
+        If the whole spec is empty, it's allowed to have a single empty object at the top level
+        """
+        specification = connector_spec_dict["connectionSpecification"]
+        errors = []
+        for type_path, value in dpath.util.search(specification, "**/type", yielded=True):
```
alafanechere, line 315:
```suggestion
        for type_path, type_value in dpath.util.search(specification, "**/type", yielded=True):
```
nit: for consistency

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,148 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+                # allow empty root object
+                continue
+            if value == "object":
+                property = self._get_parent(specification, type_path)
+                if "oneOf" not in property and ("properties" not in property or len(property["properties"]) == 0):
+                    errors.append(
+                        f"{type_path} is an empty object which will not be represented correctly in the UI. Either remove or add specific properties"
+                    )
+        if len(errors) > 0:
+            pytest.fail("\n".join(errors))
```
alafanechere, line 326: nit: wrap this in a function as you reuse this logic often.

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,148 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+
+    def test_array_type(self, connector_spec_dict: dict):
+        """
+        Each array has one or multiple types for its items, but the UI only supports a single type which can either be object, string or an enum
+        """
+        specification = connector_spec_dict["connectionSpecification"]
+        errors = []
+        for type_path, type in dpath.util.search(specification, "**/type", yielded=True):
```
alafanechere, line 334: nit: this search operation is the same as in `test_object_not_empty`, could be worth adding it as a class fixture.

alafanechere, line 334:
```suggestion
        for type_path, type_value in dpath.util.search(specification, "**/type", yielded=True):
```
For consistency, moreover `type` is a reserved python keyword.

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,148 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+            property_definition = self._get_parent(specification, type_path)
+            if type != "array":
+                # unrelated "items", not an array definition
+                continue
+            items_value = property_definition.get("items", None)
+            if items_value is None:
+                continue
```
alafanechere, line 341: Isn't it a requirement to declare the `items` field when you declare an array?

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +277,148 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
...
+    def test_date_pattern(self, connector_spec_dict: dict, detailed_logger):
+        """
+        Properties with format date or date-time should always have a pattern defined how the date/date-time should be formatted
+        that corresponds with the format the datepicker component is creating.
+        """
+        specification = connector_spec_dict["connectionSpecification"]
+        for format_path, format in dpath.util.search(specification, "**/format", yielded=True):
+            if not isinstance(format, str):
+                # format is not a format definition here but a property named format
+                continue
```
alafanechere, line 389: Could you please add a test  case to `test_date_format`  unit test to validate this behavior?

alafanechere (changes requested):
Thanks for these great additions!
I made minor comments.
Requesting change for the single value array topic and `type` variable name which is a reserved python keyword.
Could you please also share in #dev-connectors-gl when you merge this? This will likely break existing connectors, so if our maintainers can be aware of this addition it will be helpful to debug faster.

flash1293:
Thanks for your comments @alafanechere , I will address them!

> This will likely break existing connectors, so if our maintainers can be aware of this addition it will be helpful to debug faster

Hopefully not, I ran these new tests for all sources locally and the only problems that showed up are fixed here: https://github.com/airbytehq/airbyte/pull/21587

alafanechere:
> Hopefully not, I ran these new tests for all sources locally and the only problems that showed up are fixed here: #21587
Ok then I can approve. Please address the `type` variable naming and the other comments if you like.

alafanechere (approved):

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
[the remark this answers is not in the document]
flash1293: As mentioned below I tested all existing sources, but if you want to be safe I'm happy to pull it in, no strong objections.

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -268,7 +269,7 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
 
         if non_secrets_hidden:
             properties = "\n".join(non_secrets_hidden)
-            detailed_logger.warning(
+            pytest.fail(
```
girarda, line 272: <3

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
```diff
@@ -280,6 +281,139 @@ def test_secret_is_properly_marked(self, connector_spec_dict: dict, detailed_log
                     {properties}"""
             )
 
+    def _fail_on_errors(self, errors: List[str]):
+        if len(errors) > 0:
+            pytest.fail("\n".join(errors))
+
+    def test_property_type_is_not_array(self, connector_specification: dict):
```
girarda, line 288: nit: the type should be `Dict`

girarda (approved):

flash1293:

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
bazarnov: @girarda @flash1293 

This test fails with :
```python
      def test_property_type_is_not_array(self, connector_specification: dict):
E       fixture 'connector_specification' not found
>       available fixtures: acceptance_test_config, actual_connector_spec, base_path, cache, cache_discovered_catalog, cached_schemas, capfd, capfdbinary, caplog, capsys, capsysbinary, class_mocker, configured_catalog, configured_catalog_path, connection_specification, connector_config, connector_config_path, connector_setup, connector_spec, connector_spec_dict, connector_spec_path, cov, detailed_logger, discovered_catalog, docker_runner, doctest_namespace, empty_streams, expect_records_config, expected_records_by_stream, image_tag, invalid_connector_config, invalid_connector_config_path, malformed_connector_config, mocker, module_mocker, monkeypatch, no_cover, package_mocker, previous_cached_schemas, previous_connector_docker_runner, previous_connector_image_name, previous_connector_spec, previous_discovered_catalog, pull_docker_image, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, requests_mock, secret_property_names, session_mocker, skip_backward_compatibility_tests, test_strictness_level, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```

airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_core.py
erohmensing: +1 to @flash1293's reasoning - I had similar reasoning [here](https://github.com/airbytehq/airbyte/issues/18741) re: we can't enforce date-formatting on things that just look like they could be dates - its up to the source to define whether we should interpret it as a date or not

PR #21454, opened 2023-01-16
🪟 🐛 Fix duplicate createUser calls on signup

joeykmh:
## What
`onAfterAuth()` was being called multiple times. This in turn was causing multiple calls to `web_backend/users/create` with the same user information, which caused many 409s (conflicting user names) and spamming our oncall with failed requests.

## How
Reverts the change introduced in https://github.com/airbytehq/airbyte/pull/21175

## Considerations
The auth service is missing tests, which could help avoid this. The component is complex and depends on many other modules, so testing this is out of scope for this PR.

### DIFF
--- a/airbyte-webapp/src/packages/cloud/services/auth/AuthService.tsx
+++ b/airbyte-webapp/src/packages/cloud/services/auth/AuthService.tsx
@@ -1,5 +1,5 @@
 import { User as FirebaseUser } from "firebase/auth";
-import React, { useCallback, useContext, useEffect, useMemo, useRef } from "react";
+import React, { useCallback, useContext, useMemo, useRef } from "react";
 import { useQueryClient } from "react-query";
 import { useEffectOnce } from "react-use";
 import { Observable, Subject } from "rxjs";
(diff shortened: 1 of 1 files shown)


airbyte-webapp/src/packages/cloud/services/auth/AuthService.tsx
```diff
@@ -147,26 +147,34 @@ export const AuthenticationProvider: React.FC<React.PropsWithChildren<unknown>>
   const stateRef = useRef(state);
   stateRef.current = state;
 
-  useEffectOnce(() => {
-    return auth.onAuthStateChanged(async (currentUser) => {
-      // We want to run this effect only once on initial page opening
+  const onAuthStateChange = useCallback(
```
josephkmh, line 150: Extracted since this logic is used in two places

josephkmh:
@tealjulia FYI

airbyte-webapp/src/packages/cloud/services/auth/AuthService.tsx
```diff
@@ -147,26 +147,34 @@ export const AuthenticationProvider: React.FC<React.PropsWithChildren<unknown>>
...
 
+  // Check auth state on window focus, in case the user logged out in a different tab
   useEffect(() => {
     const onFocus = async () => {
       return auth.onAuthStateChanged(async (currentUser) => {
         if (!currentUser) {
           loggedOut();
         } else {
-          await onAfterAuth(currentUser);
+          onAuthStateChange(currentUser);
```
yeomann, line 177: with this new PR changes, async is redundant at `const onFocus = async` and `return auth.onAuthStateChanged(async`

josephkmh: Thanks for catching that, indeed there are several unnecessary `async` keywords, will clean those up 🙂

yeomann: no worries, i just came across the repo today and was checking some pull requests.

flash1293 (approved):
LGTM; as discussed in Slack this is a simple revert of https://github.com/airbytehq/airbyte/pull/21175 - a new issue will be created to track the new proper fix of that

PR #21456, opened 2023-01-16
🪟 🔧 Refactor webapp configuration provider

### LINKED ISSUE #10955: Generalize configProvider
Currently the config provider in the FE does mix it's configuration from build time environment variables (`REACT_APP_`), variables assigned to `window` and a file loaded from `config.json`. 

We want to narrow this down to two different providers with different priority, i.e. `window` or `config.json` should exist and be able to overwrite the build time environment variables. The code should be generic to be able to overwrite any part of the configuration without requiring hard-coded references for each individual variable.

joeykmh:
## What
Simplifies our app config to use a single, central configuration object.

Closes https://github.com/airbytehq/airbyte/issues/10955

### Consolidates many config files into one
Previously we had configuration split across many files, organized by where the value was coming from (`.env` file, `window` object, `config.json` file, etc). These many configuration files have been consolidated into a single object where we can see all config values: `src/config/config.ts`.

For the moment, we are still setting some configuration properties via the `window` object. This allows us to "inject" runtime overrides in the [nginx container](https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/nginx/default.conf.template#L24-L29). These will be removed in a future PR, in favor of setting all required configuration values via environment variables. This will allow us to configure each environment at build time (without any runtime overrides) and ultimately deploy the application as simple static files on a CDN.

### Consolidates two `useConfig()` hooks into one
Previously we had two `useConfig()` hooks - one in `src/packages/cloud/services/config` and one in `src/config`. Since the config object has been consolidated, we don't need these two hooks anymore. We will only use the one located in `src/config`.

This has some implications: some services depend on configuration values that are only available in cloud environments (such as Firebase or Intercom configuration, a `cloudApiUrl`, etc). These services will now throw a `MissingConfigError` if their required config is not available:

```js
  if (!intercom.appId) {
    throw new MissingConfigError("Missing required configuration intercom.appId");
  }
```

> Note: I tried implementing a `Proxy` here to throw this error whenever `config` is accessed with a property that is undefined. Unfortunately TypeScript does not properly infer types from proxies ([issue](https://github.com/microsoft/TypeScript/issues/20846)) which means the "optional" cloud-only types remain `string | undefined`. While we could force TS to cooperate here, I believe this could also give consumers the wrong impression that _all_ config properties are always available, while in fact attempting to access an `undefined` property could throw an error at runtime. I think it's best to keep the `string | undefined` type for those config properties that are not guaranteed to be set.

## What's next?

This is Phase 1, Step 1 of the plan outlined here: https://docs.google.com/document/d/1QIDhYP90VSsmL-ddn-2iL5QD-y0fDYgkDd0IErBXYBY/edit#heading=h.wc4fbc3h3qyl

The next step will be Phase 1, Step 2: Removing nginx runtime environment variables

## 🚨 User Impact 🚨
These changes could potentially break a variety of integrations and functionality, so they should be tested carefully.

Configuration is verified to be working in:
- [x] Local OSS/docker deployment
- [ ] Local kubernetes deployment
- [ ] frontend-dev
- [ ] dev
- [ ] dev-1
- [x] dev-2
- [ ] dev-3
- [ ] staging
- [ ] production


airbyte-webapp/src/config/types.ts
```diff
@@ -13,21 +13,30 @@ declare global {
...
     REACT_APP_INTEGRATION_DOCS_URLS?: string;
-    SEGMENT_TOKEN?: string;
+    SEGMENT_TOKEN: string;
     LAUNCHDARKLY_KEY?: string;
     analytics: SegmentAnalytics.AnalyticsJS;
   }
 }
 
-export interface Config {
+export interface AirbyteWebappConfig {
```
josephkmh, line 22: Renamed to make it more clear what this is a config for

airbyte-webapp/src/packages/cloud/services/AppServicesProvider.tsx
```diff
@@ -43,6 +43,10 @@ const ServiceOverrides: React.FC<React.PropsWithChildren<unknown>> = React.memo(
 
   const { cloudApiUrl } = useConfig();
 
+  if (!cloudApiUrl) {
+    throw new MissingConfigError("Missing required configuration cloudApiUrl");
+  }
```
josephkmh, line 48: Another option would be to log the error with our `AppMonitoringService()` instead of throwing it here. The downside is that this would not be surfaced to the user in environments without datadog.

airbyte-webapp/src/config/types.ts
```diff
@@ -13,20 +13,36 @@ declare global {
     REACT_APP_WEBAPP_TAG?: string;
     REACT_APP_INTERCOM_APP_ID?: string;
     REACT_APP_INTEGRATION_DOCS_URLS?: string;
-    SEGMENT_TOKEN?: string;
+    SEGMENT_TOKEN: string;
```
josephkmh, line 16: As far as I can tell, we have segment enabled in every environment. For example, in OSS we set `REACT_APP_SEGMENT_TOKEN` in `airbyte-webapp/.env`

timroes: While the token should always be set (and I would prefer we even don't rely on this), this here is the typing for the `window` object. We're not having the SEGMENT_TOKEN always set on the `window` object, since e.g. in OSS it's coming in via the `.env` file and thus `process.env`.

josephkmh: Right, then my question is irrelevant, whoops. I'll make it optional on `window` again.

airbyte-webapp/src/config/config.ts
```diff
@@ -0,0 +1,36 @@
...
+    token: window.SEGMENT_TOKEN ?? process.env.REACT_APP_SEGMENT_TOKEN,
+    enabled: window.TRACKING_STRATEGY === "segment",
+  },
+  apiUrl:
+    window.API_URL ??
+    process.env.REACT_APP_API_URL ??
+    `${window.location.protocol}//${window.location.hostname}:8001/api`,
+  connectorBuilderApiUrl:
+    process.env.REACT_APP_CONNECTOR_BUILDER_API_URL ?? `${window.location.protocol}//${window.location.hostname}:8003`,
+  healthCheckInterval: 20000,
```
timroes, line 14: Given that this is never actually configurable, I suggest we just move that as a constant into the `HealthPollService.tsx`

airbyte-webapp/src/config/types.ts
```diff
@@ -38,4 +62,4 @@ export type Provider<T> = () => T;
 
 export type ValueProvider<T> = Array<ProviderAsync<DeepPartial<T>>>;
 
-export type ConfigProvider<T extends Config = Config> = ProviderAsync<DeepPartial<T>>;
+export type ConfigProvider<T extends AirbyteWebappConfig = AirbyteWebappConfig> = ProviderAsync<DeepPartial<T>>;
```
timroes, line 65: I think most of those types could still be removed or the Provider which seems to be used in the auth service, just inlined there instead (and most likely doesn't need a named type).

josephkmh: Thanks, forgot that I wanted to clean up (remove) these unnecessary types!

josephkmh: addressed here https://github.com/airbytehq/airbyte/pull/21456/commits/1550b3a594946cdb242ec8c438503c5ca1969ffe

timroes (approved):
Code LGTM. Have tested locally in cloud mode. Everything seems to be applied as expected.

josephkmh:
The e2e tests are legitimately failing here, will need to refactor `useFreeConnectorProgram()` a bit before merging this PR

ambirdsall:
I was too occupied to get in a timely review, but I'm glad I kept the tab open: this is great stuff, _much_ cleaner.

PR #21466, opened 2023-01-16
feat: return whether configuration was updated as part of api response

### LINKED ISSUE #20913: Better handle control messages while editing the connector through UI (Part 1: API update)

#20894 introduced updating connector configuration in check commands when the connector emits a config control message. This is mainly used for supporting single-use refresh tokens in an OAuth implementation, but can also be used for config migrations.

If the `check` operation changed your configuration (whether due to token refresh or config migration), the values in the form would become outdated. Because the UI calls an `update` after a successful check, the (non-secret) values updated by the control message would be reverted to those set on the form.

This is mostly a problem in the case of a reauthenticating a connector that always refreshes tokens ("eager token refresh"), since the `check` will invalidate the credentials and the UI will save the invalid credentials again. But it also means we're reverting any other changes that may be happening.

image

We should:
1. Update the `check_connection_for_update` endpoint to return whether the configuration has changed as a result of the check. We can do this by comparing the provided config with the currently stored config after the check has completed.
2. On the frontend, if the above returns `true`, don't issue the `update` call (since something else updated the config) and show a message indicating to the user that the configuration has changed and they should reload the page.

Instead of showing a message telling users to reload, we could return the config and update the frontend to reflect the new configuration, but this is a bit more complex and doesn't seem high priority right now.

pedroslopez:
## What

Returns a new `connectorConfigurationUpdated` boolean as part of the job metadata in synchronous job responses.
This will allow us to give the user the appropriate feedback / fetch the updated config on the frontend when jobs like `check` emit an updated configuration.

I opted to do this on the worker side and send the output rather than comparing configs on the api handler since it actually turned out to be simpler than dealing with hydrating the config with secrets and other things that came up with that approach. Since the worker knows what was the initial config, doing the comparison on that side is pretty straightforward.

We only really need this for `check` at the moment, but since any jobs that use configs can update configs (e.g. discover), this was added across all synchronous jobs. This also made it easier to pass this info along regardless of success/failure statues.

close #20913 

Example response for a check that updated config:

```json
{
  "status": "succeeded",
  "jobInfo": {
    "id": "502e0205-e178-4890-9ba6-ba02d8265091",
    "configType": "check_connection_source",
    "configId": "Optional[ef69ef6e-aa7f-4af1-a01d-ef775033524e]",
    "createdAt": 1673917879434,
    "endedAt": 1673917893148,
    "succeeded": true,
    "connectorConfigurationUpdated": true,
    "logs": {
      "logLines": [
        "2023-01-17 01:11:21 \u001b[32mINFO\u001b[m i.a.c.i.LineGobbler(voidCall):114 - ----- START CHECK -----",
        "...",
      ]
    }
  }
}
```

...and for one that didn't:

```json
{
  "status": "succeeded",
  "message": "Source config: ContinuousFeedConfig{maxMessages=100, seed=0, messageIntervalMs=Optional.empty, mockCatalog=io.airbyte.protocol.models.AirbyteCatalog@56e07a08[streams=[io.airbyte.protocol.models.AirbyteStream@35d6ca49[name=data_stream,jsonSchema={\"type\":\"object\",\"properties\":{\"column1\":{\"type\":\"string\"},\"column2\":{\"type\":\"array\",\"items\":{\"type\":\"string\"}},\"column3\":{\"type\":\"number\"}}},supportedSyncModes=[full_refresh],sourceDefinedCursor=<null>,defaultCursorField=[],sourceDefinedPrimaryKey=[],namespace=<null>,additionalProperties={}], io.airbyte.protocol.models.AirbyteStream@3b9d6699[name=stream2,jsonSchema={\"type\":\"object\",\"properties\":{\"field1\":{\"type\":\"object\",\"properties\":{}}}},supportedSyncModes=[full_refresh],sourceDefinedCursor=<null>,defaultCursorField=[],sourceDefinedPrimaryKey=[],namespace=<null>,additionalProperties={}]],additionalProperties={}]}",
  "jobInfo": {
    "id": "bc0baed9-550c-43b0-8051-72184a6a3aef",
    "configType": "check_connection_source",
    "configId": "Optional[d53f9084-fa6b-4a5a-976c-5b8392f4ad8a]",
    "createdAt": 1673917270865,
    "endedAt": 1673917305828,
    "succeeded": true,
    "connectorConfigurationUpdated": false,
    "logs": {
      "logLines": [
        "2023-01-17 01:01:16 \u001b[32mINFO\u001b[m i.a.c.i.LineGobbler(voidCall):114 - ----- START CHECK -----",
        "...",
      ]
    }
  }
}
```

## How

- Added `connectorConfigurationUpdated` as part of the `ConnectorJobOutput`
- Check / Discover workers now set the above appropriately if they processed a config message and this config message updated the configuration (if the config message contains the same configuration that was set as the job input, `connectorConfigurationUpdated` would be false)
- Set `connectorConfigurationUpdated` as part of the `SynchronousJobMetadata`
- Added `connectorConfigUpdated` to the `SynchronousJobRead` api schema (and pass value rom above)


pedroslopez:
I'm not quite sure what to do with this coverage comment 😅 I don't think my changes made it go down this much.
Also, I'm pretty sure the coverage report on DefaultCheckConnectionWorker and DefaultDiscoverCatalogWorker is wrong - each of these do have tests.

alafanechere (approved):
Nice!

alafanechere (approved):
Nice!

airbyte-api/src/main/openapi/config.yaml
```diff
@@ -4200,6 +4200,9 @@ components:
           format: int64
         succeeded:
           type: boolean
+        connectorConfigurationUpdated:
```
evantahler, line 4203: nit: Could `connectorConfigurationUpdated` and `didUpdateConfiguration` both be the same name?  Perhaps `didUpdateConnectorConfiguration` works in both cases?

airbyte-commons-worker/src/main/java/io/airbyte/workers/WorkerUtils.java
```diff
@@ -140,6 +142,12 @@ private static Optional<AirbyteTraceMessage> getTraceMessageFromMessagesByType(f
         .findFirst();
   }
 
+  public static Boolean getDidControlMessageChangeConfig(final JsonNode initialConfigJson, final AirbyteControlConnectorConfigMessage configMessage) {
+    final Config newConfig = configMessage.getConfig();
+    final JsonNode newConfigJson = Jsons.jsonNode(newConfig);
+    return !initialConfigJson.equals(newConfigJson);
```
evantahler, line 148: ~~This is OK, but calling out that this would would fail if the config JSON was re-ordered by the connector.  However, the connector shouldn't just re-order it's config without a good reason...~~

Actually, `jackson.equals()` should be positionally independant!

evantahler (approved):

airbyte-api/src/main/openapi/config.yaml
pedroslopez: I wanted to use `didUpdateConnectorConfiguration`, but PMD or some other check yelled at me because it wanted the get method on SynchronousJobMetadata to start with `is` :(

I can change the ConnectorJobOutput to be connectorConfigurationUpdated though

pedroslopez: Updated in [37b734d](https://github.com/airbytehq/airbyte/pull/21466/commits/37b734da2aa0ed63f324f3c55437378b0029508c)

mfsiega-airbyte:
@pedroslopez just a note - the intention of the code coverage comment is just to prompt you to consider whether you've sufficiently tested your changes here, and provide _some_ signal.

As far as why it doesn't know that those files do have test coverage - the workers themselves live in `airbyte-commons-workers`, but the tests live in `airbyte-workers`. I'd guess that's what's confusing the coverage reporter.

PR #21472, opened 2023-01-17
airbyte-webapp: use the `supported` field to define `supportsNormalization`

alafanechere:
## What
Now that the `normalizationConfig` object has a `supported` required boolean property we can use it to define the `supportsNormalization` variable.

 
## How
Update `supportsNormalization` declaration in `ConnectionTransformationPage.tsx`


evantahler (approved):
👍

PR #21474, opened 2023-01-17
ci-connector-ops: split workflows

alafanechere:
## What
When I added the `check-review-requirements`  I added a workflow trigger on pull request review.
It meant that on any PR on the Airbyte repo both `check-review-requirements` and `test-strictness-level` jobs were running. 
It's not great because we don't want to run the `test-strictness-level` job on PR not related to `sources`. 

## How
Split the original workflow in two workflow with different triggering logic.


evantahler (changes requested):
Both workflows now work only for sources:

```
on:
  pull_request:
    paths:
      - "airbyte-integrations/connectors/source-**"
```

Perhaps one of these should be for all connectors?

evantahler (approved):
(edit) I see the removal of `workflow_dispatch` now!

PR #21475, opened 2023-01-17
Updating start-aws-runner-action to fetch the latest AMI

Mykyta-Serbynevskyi:
In scope of https://github.com/airbytehq/airbyte-cloud/issues/3612 a new process of AMI building was introduced.
This PR updates `start-aws-runner-action` to fetch the latest freshest AMI in AWS for gh self-hosted runners.

### DIFF
--- a/.github/actions/start-aws-runner/action.yml
+++ b/.github/actions/start-aws-runner/action.yml
@@ -8,7 +8,7 @@ inputs:
   github-token:
     required: true
   ec2-image-id:
-    # https://internal-docs.airbyte.io/Things-To-Know/Build-Runner-Images#ami-017f6fe645968f7e0
+    # https://internal-docs.airbyte.io/Things-To-Know/Build-Runner-Images
     default: "ami-017f6fe645968f7e0"
     required: true
   ec2-instance-type:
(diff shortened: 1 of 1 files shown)


supertopher (approved):
this looks great Mykyta.  Did we test this beforehand or are we testing in prod?

even if we have tested this I want to keep an eye in prod since I imagine if this fails it's going to cause alot of drama.

ty

Mykyta-Serbynevskyi:
> this looks great Mykyta. Did we test this beforehand or are we testing in prod?
> 
> even if we have tested this I want to keep an eye in prod since I imagine if this fails it's going to cause alot of drama.

Yes, it has been tested by checks in this PR and they are all `green`. 

Moreover, I can see those checks used the freshest AMI that had been created via [Build AMI](https://github.com/airbytehq/airbyte-cloud/actions/runs/3934351287/jobs/6729013142#step:7:104) gh workflow in `airbyte-cloud` repo:

https://github.com/airbytehq/airbyte/actions/runs/3938364826/jobs/6736973450#step:4:53

Thus there should not be any issues.
Anyway, I will keep an eye on this for a while.

PR #21481, opened 2023-01-17
Source GitHub: #970 handle 502 Bad Gateway

roman-yermilov-gl:
## What
*Handle 502 Bad Gateway with proper message*

## How
*If there is an 502 error then we try to do few more delayed requests and finally (if not succeeded) provide a client with proper log message*

## Recommended reading order
1. `x.java`
2. `y.python`


lazebnyi (commented):
Need add record to change log

bazarnov (approved):

roman-yermilov-gl:
> Need add record to change log

Done. I didn't do it before since I was waiting for another PR in which version also bumped to 0.3.12

lazebnyi (approved):

PR #21484, opened 2023-01-17
Reintroduce AvailabilityStrategy into the CDK (HttpAvailabilityStrategy default not turned on yet)

### LINKED ISSUE #21426: Reintroduce AvailabilityStrategy into the CDK

erohmensing:
See [original PR](https://github.com/airbytehq/airbyte/pull/19977) for feature description.

Closes #21426 

Differences from original implementation:
* CheckStream now explicitly uses AvailabilityStrategy (behavior was previously the same for both)
* Add [test](https://github.com/airbytehq/airbyte/commit/7c17351631ece93fe78fc4a1a730c120cc132012) from post-release bug fix (bug fix has since been handled separately in the CDK)
* Bug edge case code from [existing CheckStream fix](https://github.com/airbytehq/airbyte/pull/21404) now exists in the HttpAvailabilityStrategy, + tests from these cases

Note: Do not merge until [dependent PR](https://github.com/airbytehq/airbyte/pull/21488) turning off the default is merged into this branch.


airbyte-cdk/python/airbyte_cdk/sources/streams/http/availability_strategy.py
```diff
@@ -0,0 +1,133 @@
...
+        :param source: optional (source)
+        :return: A dictionary of (status code, reason) where the 'reason' explains
+        why 'status code' may have occurred and how the user can resolve that
+        error, if applicable.
+        """
+        forbidden_error_message = f"The endpoint to access stream '{stream.name}' returned 403: Forbidden. "
+        forbidden_error_message += "This is most likely due to insufficient permissions on the credentials in use. "
+        forbidden_error_message += self._visit_docs_message(logger, source)
+
+        reasons_for_codes: Dict[int, str] = {requests.codes.FORBIDDEN: forbidden_error_message}
```
girarda, line 102: it might be worth parameterizing the class on this Dict. some APIs might return a 404 instead of a 403 if a set of credentials have insufficient permissions

airbyte-cdk/python/airbyte_cdk/sources/utils/stream_helpers.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        Gets the first record for a stream.
+
+        :param stream: stream
+        :return: StreamData containing the first record in the stream
+        """
+        # Some streams need a stream slice to read records (e.g. if they have a SubstreamSlicer)
+        stream_slice = self.get_stream_slice(stream)
+        records = stream.read_records(sync_mode=SyncMode.full_refresh, stream_slice=stream_slice)
+        next(records)
```
girarda, line 23: `return next(records)`? This method will always return `None` right now

airbyte-cdk/python/unit_tests/sources/streams/http/test_availability_strategy.py
```diff
@@ -0,0 +1,163 @@
...
+    mocker.patch.object(requests.Session, "send", return_value=req, json={"error": "Oh no!"})
+
+    stream_is_available, reason = http_stream.check_availability(logger, source)
+    assert not stream_is_available
+
+    expected_messages = [
+        f"The endpoint to access stream '{http_stream.name}' returned 403: Forbidden.",
+        "This is most likely due to insufficient permissions on the credentials in use.",
+        f"Please visit https://docs.airbyte.com/integrations/sources/{source.name} to learn more.",
+        # "Oh no!",
```
girarda, line 105: nit: I'm not sure how to interpret this comment

airbyte-cdk/python/unit_tests/sources/test_source.py
```diff
@@ -7,10 +7,10 @@
 import tempfile
 from collections import defaultdict
 from contextlib import nullcontext as does_not_raise
-from typing import Any, List, Mapping, MutableMapping, Optional, Tuple
-from unittest.mock import MagicMock
```
girarda, line 11: why is this import deleted?

girarda (changes requested):

airbyte-cdk/python/airbyte_cdk/sources/utils/stream_helpers.py
erohmensing: This whole file isn't supposed to be here - It's not being used anywhere, made it into accidentally from the merge. Thanks for flagging! The real utils file is the one we added last week (hence not in the diff) imported [here](https://github.com/airbytehq/airbyte/pull/21484/files/91affb8c69d7c532438ac4f3f2ffc1f6387fbb33#diff-a9dd5cf1051977e9a7c58329dbac9bded08d9f53f314ae2bdeef073f360820bcR12). Removing this one

airbyte-cdk/python/unit_tests/sources/test_source.py
erohmensing: Using the `MagicMock()` already included in `mocker` instead of importing it was a [review suggestion](https://github.com/airbytehq/airbyte/pull/19977#discussion_r1044325754) in the previous PR

airbyte-cdk/python/airbyte_cdk/sources/streams/http/availability_strategy.py
erohmensing: In that case, they'd also want to update the error message - in which case they can override the entire `reasons_for_unavailable_status_codes` method. Wdyt? Is it specifically declarative sources that makes us prefer a class variable over an overrideable method?

airbyte-cdk/python/unit_tests/sources/streams/http/test_availability_strategy.py
erohmensing: Fixed - was something that was supposed to be asserted on (that the error from the json response exists in the error message, but wasn't being checked. Merged with the test above via parametrize, now is being tested

airbyte-cdk/python/airbyte_cdk/sources/streams/http/availability_strategy.py
girarda: I try to avoid redefining methods using inheritance because long inheritance chains can be tricky, but I see it's difficult to avoid because we need to Source object to produce the error message. Not something worth blocking on

girarda (approved):

airbyte-cdk/python/airbyte_cdk/sources/streams/http/availability_strategy.py
erohmensing: Haha, I do hear that - I had to write a separate doc just to figure out how all the inheritance was working for AbstractSource 😄

PR #21495, opened 2023-01-17
Remove unused protocol version envvar

gosusnp:
## What

* We are not using and not planning to use the `AIRBYTE_PROTOCOL_VERSION_MAX` and `AIRBYTE_PROTOCOL_VERSION_MIN` environment variables.
* The lookup generates noise in our logs as we are using the expected default.

## How

Replace the var lookups by the hardcoded default directly to remove noise from the logs.

### DIFF
--- a/airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
+++ b/airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
@@ -42,8 +42,6 @@ public class EnvConfigs implements Configs {
   // env variable names
   public static final String AIRBYTE_ROLE = "AIRBYTE_ROLE";
   public static final String AIRBYTE_VERSION = "AIRBYTE_VERSION";
-  public static final String AIRBYTE_PROTOCOL_VERSION_MAX = "AIRBYTE_PROTOCOL_VERSION_MAX";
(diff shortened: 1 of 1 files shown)


airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
```diff
@@ -312,12 +310,12 @@ public AirbyteVersion getAirbyteVersion() {
 
   @Override
   public Version getAirbyteProtocolVersionMax() {
-    return new Version(getEnvOrDefault(AIRBYTE_PROTOCOL_VERSION_MAX, "0.3.0"));
+    return new Version("0.3.0");
```
benmoriceau, line 313: Micronaut is expecting an envvar in the server it default to those values. But I think that we should check is other micronaut application.yml also specify a default value and don't rely on the return of this value.

benmoriceau: https://github.com/airbytehq/airbyte/blob/master/airbyte-server/src/main/resources/application.yml#L90

gosusnp: When we get rid of `EnvConfigs` we should directly use the values from the micronaut config. Not sure where we are on this, this is a quick fix to remove logging noise.

benmoriceau (approved):

jdpgrailsdev (approved):
:shipit:

PR #21514, opened 2023-01-18
:window: :wrench: :cloud: Check email verification status for free connector enrollment

ambirdsall:
## What
A few improvements to the frontend for free connector enrollment:
- passes `emailVerified` and `sendEmailVerification` into `EnrollmentModalContent` as props (via `useAuthService()`, which can't be called within the modal content component itself but can be from the `useShowEnrollmentModal` hook)
- if and only if the user hasn't verified their email:
  * conditionally renders a new callout with a link to resend the email verification
  * disables the CTA
- minor refactors to the `useFreeConnectorProgram` hook
  * extracts the experiment flag check into the hook, making it a single source of truth to override or update
  * rename the function to match its filename
 
Screenshot of the modal for a user with unvalidated email:
fcp-email-unverified-modal

## Okay, but why is `CreditsPage/components/EmailVerificationHint.tsx` in this PR
Right, that. That component was the only place in the app that used the `sendEmailVerification` function from `useAuthService`. It also added a bunch of error handling that we'll want every time we use this API, so before I used `sendEmailVerification` in the FCP enrollment modal, I extracted the error handling into `useAuthService` itself, so we would get it "for free". As of now, the error handling strategy is hardcoded in the hook; if the need for context-dependent error handling arises, we will have to rewrite `sendEmailVerification` to return some well-typed error object instead of handling errors right inside the function. But I'm [not sure we're gonna need it](https://martinfowler.com/bliki/Yagni.html).

## Recommended reading order
There are three pairs of files whose changes are linked; you can read them in either order, but should probably consider them together.
1. `useShowEnrollmentModal.tsx` / `EnrollmentModal.tsx`
2. `useFreeConnectorProgram.ts` / `ConnectionPageTitle.tsx`
3. `EmailVerificationHint.tsx` / `AuthService.tsx`
4. the rest

### DIFF
--- a/airbyte-webapp/src/locales/en.json
+++ b/airbyte-webapp/src/locales/en.json
@@ -745,12 +745,5 @@
 
   "jobs.noAttemptsFailure": "Failed to start job.",
 
-  "cloudApi.loginCallbackUrlError": "There was an error connecting to the developer portal. Please try again.",
-
-  "freeConnectorProgram.enrollmentModal.title": "Free connector program",
-  "freeConnectorProgram.enrollmentModal.free": "<p1>Alpha and Beta connectors are free while you're in the program.</p1><p2>The whole connection is free until both the source and destination connector have moved into General Availability (GA).</p2>",
-  "freeConnectorProgram.enrollmentModal.emailNotification": "We will email you before both connectors in a connection move to GA.",
-  "freeConnectorProgram.enrollmentModal.cardOnFile": "When both connectors are in GA, the connection will no longer be free. You'll need to have a credit card on file to enroll so Airbyte can handle a connection's transition to paid service.",
-  "freeConnectorProgram.enrollmentModal.cancelButtonText": "Cancel",
-  "freeConnectorProgram.enrollmentModal.enrollButtonText": "Enroll now!"
+  "cloudApi.loginCallbackUrlError": "There was an error connecting to the developer portal. Please try again."
 }
--- a/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/EnrollmentModal/EnrollmentModal.tsx
+++ b/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/EnrollmentModal/EnrollmentModal.tsx
@@ -21,12 +21,19 @@ interface EnrollmentModalContentProps {
   closeModal: () => void;
   createCheckout: (p: StripeCheckoutSessionCreate) => Promise<StripeCheckoutSessionRead>;
   workspaceId: string;
+  emailVerified: boolean;
+  sendEmailVerification: () => void;
 }
 
+// we have to pass the email verification data and functions in as props, rather than
+// directly using useAuthService(), because the modal renders outside of the
+// AuthenticationProvider context.
 export const EnrollmentModalContent: React.FC<EnrollmentModalContentProps> = ({
   closeModal,
   createCheckout,
   workspaceId,
+  emailVerified,
+  sendEmailVerification,
 }) => {
   const isMountedRef = useRef(false);
   const [isLoading, setIsLoading] = useState(false);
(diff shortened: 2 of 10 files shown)


airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/EnrollmentModal/EnrollmentModal.tsx
```diff
@@ -57,6 +64,32 @@ export const EnrollmentModalContent: React.FC<EnrollmentModalContentProps> = ({
...
+  const EnrollmentEmailVerificationWarning = () => {
+    const WarningContent = () => (
+      <FlexContainer>
+        <FlexContainer justifyContent="center" className={styles.iconContainer}>
+          <MailSVG />
+        </FlexContainer>
+        <Text size="lg" className={styles.warning}>
+          <FormattedMessage id="freeConnectorProgram.enrollmentModal.unvalidatedEmailWarning" />
+        </Text>
+      </FlexContainer>
```
tealjulia, line 87: I think the `<Callout/>` mentioned in Slack would be less visually harsh than the red text here.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/hooks/useFreeConnectorProgram.ts
```diff
@@ -1,26 +1,26 @@
...
   const requestOptions = { config, middlewares };
+  const freeConnectorProgramEnabled = useExperiment("workspace.freeConnectorsProgram.visible", false);
 
   return useQuery(["freeConnectorProgramInfo", workspaceId], () =>
     webBackendGetFreeConnectorProgramInfoForWorkspace({ workspaceId }, requestOptions).then(
       ({ hasEligibleConnector, hasPaymentAccountSaved }) => {
-        const showEnrollmentUi = !hasPaymentAccountSaved && hasEligibleConnector;
-        // TODO hardcoding this value to allow testing while data source gets sorted out
-        const needsEmailVerification = false;
+        const eligibleToEnroll = !hasPaymentAccountSaved && hasEligibleConnector;
```
tealjulia, line 21: I mean, wouldn't a user technically be ineligible if they don't have the experiment on?

teallarson (commented):
Could use a rebase to remove changes from the inline banner.  

Other than that, one comment and one question but looks pretty much good to go.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/hooks/useFreeConnectorProgram.ts
ambirdsall: You're not wrong! The intended meaning was more like "meets eligibility requirements", but really I just wanted to define the business logic part of the combined condition as its own thing. LMK if you think it would be clearer with a different name.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/EnrollmentModal/EnrollmentModal.tsx
ambirdsall: Agreed, definitely nicer.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/hooks/useFreeConnectorProgram.ts
ambirdsall: > I think it'd be cleaner on the other end to just have something that tells the UI "should we show this thing"?

Perhaps I'm misunderstanding, but that's is what the function returns: a single boolean that tells the UI whether to show a thing. It would behave identically if we combined all the input conditions in one big boolean expression, `freeConnectorProgramEnabled && !hasPaymentAccountSaved && hasEligibleConnector`; the fact that the function gives a separate name to the `!hasPaymentAccountSaved && hasEligibleConnector` subset of that is an internal implementation detail.

tealjulia: disregard.  i've read this file too many times.

airbyte-webapp/src/locales/en.json
```diff
@@ -745,12 +745,5 @@
 
   "jobs.noAttemptsFailure": "Failed to start job.",
 
-  "cloudApi.loginCallbackUrlError": "There was an error connecting to the developer portal. Please try again.",
-
-  "freeConnectorProgram.enrollmentModal.title": "Free connector program",
```
ambirdsall, line 750: these are not being deleted, just moved to the cloud-specific locale file

teallarson (commented):
Nice!

Some thoughts I had trying it out:
- Could use a success message on send ("verification email sent!" or something along the lines)
- I have to refresh the page after verifying to remove the warning

ambirdsall:
@tealjulia there's [a slack discussion](https://airbytehq-team.slack.com/archives/C049XDAS19C/p1674139899224119) on the confirmation feedback; definitely going to be added, though I might hold that for a follow-up for easier reviewing.

The other issue, "have to refresh the page after verifying to remove the warning", would have to be solved by polling `AuthService` outside of the modal and re-triggering it if the value changes. Definitely possible, but might be enough of a headache to accept the need to refresh (at least for the time being).

tealjulia:
> The other issue, "have to refresh the page after verifying to remove the warning", would have to be solved by polling AuthService outside of the modal and re-triggering it if the value changes. Definitely possible, but might be enough of a headache to accept the need to refresh (at least for the time being).

Reasonable!  I wouldn't say it's blocking.  Possibly a related solution to https://github.com/airbytehq/airbyte-cloud/issues/3992 (redirecting when a user logs out or in in a separate tab.

teallarson (approved):

PR #21516, opened 2023-01-18
[Low Code CDK] Pass DeclarativeStream's `name` into DefaultSchemaLoader

### LINKED ISSUE #21381: Connector builder server: No records returned if no schema loader is defined for a stream
The connector builder server is not returning records if no schema loader is defined as part of the manifest even though the schema loader is marked as optional in the typings.

To reproduce:

Issue a `read` request with a payload like this (defines a valid stream but without a schema loader defined
```
{
    "manifest": {
        "version": "0.1.0",
        "type": "DeclarativeSource",
        "check": {
            "type": "CheckStream",
            "stream_names": []
        },
        "streams": [
            {
                "type": "DeclarativeStream",
                "name": "test",
                "primary_key": [],
                "retriever": {
                    "type": "SimpleRetriever",
                    "name": "test",
                    "primary_key": [],
                    "requester": {
                        "type": "HttpRequester",
                        "name": "test",
                        "url_base": "http://dummy_api:6767/",
                        "path": "items/",
                        "request_options_provider": {
                            "request_parameters": {},
                            "request_headers": {},
                            "request_body_json": {}
                        },
                        "authenticator": {
                            "type": "BearerAuthenticator",
                            "api_token": "{{ config['api_key'] }}"
                        }
                    },
                    "record_selector": {
                        "type": "RecordSelector",
                        "extractor": {
                            "type": "DpathExtractor",
                            "field_pointer": [
                                "items"
                            ]
                        }
                    },
                    "paginator": {
                        "type": "NoPagination"
                    }
                }
            }
        ],
        "spec": {
            "connection_specification": {
                "type": "object",
                "required": [
                    "api_key"
                ],
                "properties": {
                    "api_key": {
                        "type": "string",
                        "title": "API Key",
                        "airbyte_secret": true
                    }
                },
                "additionalProperties": true
            },
            "documentation_url": "",
            "type": "Spec"
        }
    },
    "stream": "test",
    "config": {
        "api_key": "theauthkey"
    }
}
```

Builder server logs the following:
```
airbyte-connector-builder-server  | {"type": "DEBUG", "message": "Making outbound API request", "data": {"headers": "{'User-Agent': 'python-requests/2.28.2', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive', 'Authorization': 'Bearer theauthkey'}", "url": "http://dummy_api:6767/items", "request_body": "None"}}

airbyte-connector-builder-server  | {"type": "DEBUG", "message": "Receiving response", "data": {"body": "{\"items\":[{\"name\":\"abc\"},{\"name\":\"def\"}]}", "status": "200", "headers": "{'Content-Type': 'application/json', 'Date': 'Fri, 13 Jan 2023 11:20:43 GMT', 'Connection': 'keep-alive', 'Keep-Alive': 'timeout=5', 'Transfer-Encoding': 'chunked'}"}}

airbyte-connector-builder-server  | {"type": "LOG", "log": {"level": "ERROR", "message": "Encountered an exception while reading stream test\nTraceback (most recent call last):\n  File \"/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/declarative/schema/default_schema_loader.py\", line 40, in get_json_schema\n    return self.default_loader.get_json_schema()\n  File \"/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/declarative/schema/json_file_schema_loader.py\", line 62, in get_json_schema\n    raise IOError(f\"Cannot find file {json_schema_path}\")\nOSError: Cannot find file ./.json\n\nDuring handling of the above exception, another exception occurred:\n\nTraceback (most recent call last):\n  File \"/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/abstract_source.py\", line 111, in read\n    yield from self._read_stream(\n  File \"/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/abstract_source.py\", line 180, in _read_stream\n    for record in record_iterator:\n  File \"/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/abstract_source.py\", line 291, in _read_full_refresh\n    message = self._get_message(record_data_or_message, stream_instance)\n  File \"/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/abstract_source.py\", line 325, in _get_message\n    return stream_data_to_airbyte_message(stream.name, record_data_or_message, stream.transformer, stream.get_json_schema())\n  File \"/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/declarative/declarative_stream.py\", line 132, in get_json_schema\n    return self._schema_loader.get_json_schema()\n  File \"/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/declarative/schema/default_schema_loader.py\", line 44, in get_json_schema\n    stream_name = self._options.get(\"name\", \"\")\nAttributeError: 'NoneType' object has no attribute 'get'"}}
```

Then returns an empty records array.

clnoll:
Closes https://github.com/airbytehq/airbyte/issues/21381

## What
Fixes an issue where the CDK is not identifying the default schema location if a schema loader is not defined.

## How
Problem: when `streams` is invoked from https://github.com/airbytehq/airbyte/blob/81be661a1d1e40b60ab6614f5533e7ec72065c64/airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py#L104 with a manifest that does not define a schema loader, `self._constructor.create_component` creates a `DeclarativeStream` object; however, it is not created with the `options` attribute. The `DeclarativeStream` object creates a `DefaultSchemaLoader` downstream, which resolves to a `JsonSchemaLoader`. `JsonSchemaLoader` attempts to identify the path to local stream schema files; it does this by reading the `"name"` key from `options`. But, when `options` is not present, we return `""`, and so are attempting to find a file called `.json`, whereas we want the file name to default to `<stream name>.json`.

This fixes the problem by passing `DeclarativeStream.name` to the `DefaultSchemaLoader`.

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -359,7 +359,7 @@ def create_declarative_stream(model: DeclarativeStreamModel, config: Config, **k
     if model.schema_loader:
         schema_loader = _create_component_from_model(model=model.schema_loader, config=config)
     else:
-        schema_loader = DefaultSchemaLoader(config=config, options=model.options)
+        schema_loader = DefaultSchemaLoader(config=config, name=model.name, options=model.options)
 
     transformations = []
     if model.transformations:
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/schema/default_schema_loader.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/schema/default_schema_loader.py
@@ -23,9 +23,13 @@ class DefaultSchemaLoader(SchemaLoader, JsonSchemaMixin):
     """
 
     config: Config
+    name: InitVar[str]
     options: InitVar[Mapping[str, Any]]
 
-    def __post_init__(self, options: Mapping[str, Any]):
+    def __post_init__(self, name: str, options: Mapping[str, Any]):
+        options = options or {}
+        if "name" not in options:
+            options["name"] = name
         self._options = options
         self.default_loader = JsonFileSchemaLoader(options=options, config=self.config)
 
--- a/airbyte-cdk/python/unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
+++ b/airbyte-cdk/python/unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
@@ -1011,3 +1011,37 @@ class TestCreateTransformations:
             )
         ]
         assert stream.transformations == expected
+
+    def test_default_schema_loader(self):
+        component_definition = {
+            "type": "DeclarativeStream",
+            "name": "test",
+            "primary_key": [],
+            "retriever": {
+                "type": "SimpleRetriever",
+                "name": "test",
+                "primary_key": [],
+                "requester": {
+                    "type": "HttpRequester",
+                    "name": "test",
+                    "url_base": "http://localhost:6767/",
+                    "path": "items/",
+                    "request_options_provider": {
+                        "request_parameters": {},
+                        "request_headers": {},
+                        "request_body_json": {},
+                        "type": "InterpolatedRequestOptionsProvider",
+                    },
+                    "authenticator": {"type": "BearerAuthenticator", "api_token": "{{ config['api_key'] }}"},
+                },
+                "record_selector": {"type": "RecordSelector", "extractor": {"type": "DpathExtractor", "field_pointer": ["items"]}},
+                "paginator": {"type": "NoPagination"},
+            },
+        }
+        resolved_manifest = resolver.preprocess_manifest(component_definition)
+        propagated_source_config = ManifestComponentTransformer().propagate_types_and_options("", resolved_manifest, {})
+        stream = factory.create_component(
+            model_type=DeclarativeStreamModel, component_definition=propagated_source_config, config=input_config
+        )
+        schema_loader = stream.schema_loader
+        assert schema_loader.default_loader._get_json_filepath() == f"./{stream.name}.json"


airbyte-cdk/python/airbyte_cdk/sources/declarative/schema/default_schema_loader.py
```diff
@@ -23,9 +23,13 @@ class DefaultSchemaLoader(SchemaLoader, JsonSchemaMixin):
...
 
     config: Config
+    name: InitVar[str]
     options: InitVar[Mapping[str, Any]]
 
-    def __post_init__(self, options: Mapping[str, Any]):
+    def __post_init__(self, name: str, options: Mapping[str, Any]):
+        options = options or {}
+        if "name" not in options:
+            options["name"] = name
```
maxi297, line 32: If name is defined, why not use `self.name` directly at https://github.com/airbytehq/airbyte/pull/21516/files#diff-eebfee995e9a963a52d517c0ebcab9579872dee179cf961719740e6c3fdddc9fR48 instead of `self._options.get("name", "")` that was generating the error? Do we need the options at all?

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -359,7 +359,7 @@ def create_declarative_stream(model: DeclarativeStreamModel, config: Config, **k
     if model.schema_loader:
         schema_loader = _create_component_from_model(model=model.schema_loader, config=config)
     else:
-        schema_loader = DefaultSchemaLoader(config=config, options=model.options)
+        schema_loader = DefaultSchemaLoader(config=config, name=model.name, options=model.options)
```
maxi297, line 362: In the case of the connector-builder, it seems like the stream name is passed [in the request body](https://github.com/airbytehq/airbyte/blob/fcf193db16b65c25f56bf7650e6537658aafc1b8/airbyte-connector-builder-server/connector_builder/generated/models/stream_read_request_body.py#L27). Could we use this to populate the name downstream through the options? 

To be clear, I would prefer to have "name" if "name" is necessary for this class to do the job. Right now, it seems only for debugging purposes though

maxi297 (approved):
Adding to my previous review:
Conditional approval: it feels like there is overlapping concerns with `options` and `name` and to avoid breaking the interface, I'm wondering if we could just have a different error message [here](https://github.com/airbytehq/airbyte/pull/21516/files#diff-eebfee995e9a963a52d517c0ebcab9579872dee179cf961719740e6c3fdddc9fR49) when we don't have the name. If it would be possible to clean this overlapping concerns or just remove one of the fields, I think it would be great. In any case, I trust your judgement to change this or not

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: The connector builder is actually something of a unique case. The main point is that connectors that use the connector builder UI don't actually support this type Schema Loader. It has to use an inline schema loader since the builder doesn't support multiple files.

But specifically for the `stream` name that is passed in via the request to test YAML that you pointed out, we don't want that to directly impact the YAML being tests. Think of that parameter as telling us which stream we want to test rather than something we inject into the manifest itself. So for example, we might have 4-5 streams in the YAML manifest. And the `stream` parameter of StreamReadRequestBody tells us to take the given manifest and test that stream.

airbyte-cdk/python/unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
```diff
@@ -1011,3 +1011,37 @@ def test_add_fields(self):
...
+                "paginator": {"type": "NoPagination"},
+            },
+        }
+        resolved_manifest = resolver.preprocess_manifest(component_definition)
+        propagated_source_config = ManifestComponentTransformer().propagate_types_and_options("", resolved_manifest, {})
+        stream = factory.create_component(
+            model_type=DeclarativeStreamModel, component_definition=propagated_source_config, config=input_config
+        )
+        schema_loader = stream.schema_loader
+        assert schema_loader.default_loader._get_json_filepath().split("/")[-1] == f"{stream.name}.json"
```
brianjlai, line 1047: Are we able to test the entire string, not just the final part just to be comprehensive?

brianjlai (approved):
A few comments, but nothing blocking from my side

airbyte-cdk/python/unit_tests/sources/declarative/parsers/test_model_to_component_factory.py
clnoll: Yep! Done.

airbyte-cdk/python/airbyte_cdk/sources/declarative/schema/default_schema_loader.py
clnoll: `name` isn't available on `self` here, and at least for now for consistency with the rest of the factory we do want to override `model.name` if `options` is provided and includes it. Removing `options` is a possible refactor but felt out of scope for this PR (we've talked about doing a larger effort to remove it where it's not needed as there are various places where that's the case).

PR #1313, opened 2020-12-13
Add secrets handling in UI

jamakase:
Closes #1055 

# What
adds `type=password` for secrets

PR #21517, opened 2023-01-18
#1313 source google ads: write less logs

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1313

## How
Filter out each 10th log record


airbyte-integrations/connectors/source-google-ads/source_google_ads/streams.py
```diff
@@ -89,8 +108,10 @@ class GoogleAdsStream(Stream, ABC):
     CATCH_API_ERRORS = True
 
     def __init__(self, api: GoogleAds, customers: List[Customer]):
+        super().__init__()
```
brianjlai, line 111: why do we need to invoke the paren't `init` method when we were not before this PR?

airbyte-integrations/connectors/source-google-ads/source_google_ads/streams.py
```diff
@@ -154,6 +177,7 @@ def current_state(self, customer_id, default=None):
 
     def stream_slices(self, stream_state: Mapping[str, Any] = None, **kwargs) -> Iterable[Optional[MutableMapping[str, any]]]:
         for customer in self.customers:
+            logger = cyclic_sieve(self.logger, 10)
```
brianjlai, line 180: Can we avoid initializing a new `cyclic_sieve` since it should have already been defined on the stream as `self.incremental_sieve_logger`?

airbyte-integrations/connectors/source-google-ads/unit_tests/test_streams.py
```diff
@@ -218,3 +219,14 @@ def test_retry_transient_errors(mocker, config, customers, error_cls):
         records = list(stream.read_records(sync_mode=SyncMode.incremental, cursor_field=["segments.date"], stream_slice=stream_slice))
     assert mocked_search.call_count == 5
     assert records == []
+
+
+def test_cyclic_sieve(caplog):
```
brianjlai, line 224: thanks for adding tests for the new logger?

brianjlai (commented):
A few questions about some specifics of how we use the sieve, but overall the approach looks good.

airbyte-integrations/connectors/source-google-ads/source_google_ads/streams.py
davydov-d: Technically we can, but this was done on purpose. The idea was to filter out not just 9 out of 10 messages, but 9 out of 10 message cycles to have more consistent logs so an engineer shouldn't have to stick together messages from different routine calls / cycles / exceptions / if statements and so on when debugging. Since `stream_slices()` is called once per stream before reading records, and `read_records()` is called once per stream slice, these are two different log cycles.
If you still insist, I can reuse `self.incremental_sieve_logger` here, technically it is not a problem

airbyte-integrations/connectors/source-google-ads/source_google_ads/streams.py
davydov-d: I overlooked it, thanks. Removed

airbyte-integrations/connectors/source-google-ads/source_google_ads/streams.py
brianjlai: Ah got it, that makes sense and I agree that for a better experience we wouldn't want to share the same log cycles. Can you add a comment for that here just mentioning that's why a logger sieve is initialized here.

airbyte-integrations/connectors/source-google-ads/source_google_ads/streams.py
```diff
@@ -189,7 +213,8 @@ def read_records(
         and update `start_date` key in the `stream_slice` with the latest read record's cursor value, then retry the sync.
         """
         while True:
-            self.logger.info("Starting a while loop iteration")
+            self.incremental_sieve_logger.bump()
```
brianjlai, line 216: Going off the prior comment about separate slice and read_records log cycles, do you think it's also worth initializing a new sieve every time read_records is invoked? If the sieve is shared on the stream, we will emit exactly 1/10 of logs, but it could be a little confusing why certain cycles are skipped on different read_records invocations if a prior one has a partially bumped count.

I'm open to either option and this is non-blocking on approval, but I think worth highlighting when the sieve is shared at this point? If we do decide to have a separate sieve per `read_records` invocation, we could probably remove it from `__init__()` and do something similar to slices where it is created within this method.

brianjlai (approved):
Overall no blocking comments, but had one discussion point about whether we want a new sieve per-read_records invocation if that made interpreting the logs more intuitive

airbyte-integrations/connectors/source-google-ads/source_google_ads/streams.py
davydov-d: That's a tricky point cause normally a while loop here contains a single iteration. It only goes further if there were errors during the read. So I do not think there's much sense in initializing new sieve each time `read_records` is invoked. On the other hand I agree that logs may become inconsistent if there are two or more while loop iterations, therefore I moved `incremental_sieve_logger.bump()` outside the loop to remain consistent

PR #21519, opened 2023-01-18
🪟 🔧 Fixes webhook updating logic

### LINKED ISSUE #21256: Unable to clear the hook url from the notification settings
image

Actual:
When trying to remove the URL from the notification settings and saving, the user gets an error.

Expected:
The user should be able to clear the notification settings when removing the URL.

Steps to reproduce:
1. Go to notification settings
2. Add URL and save
3. Remove URL and save

YatsukBogdan1:
## What
Fixes #21256 

## How
Fixes validation of webhook, so in case if webhook is empty, it doesn't test webhook, but just saves empty string (clearing webhook)


edmundito (approved):
Tested locally

PR #21520, opened 2023-01-18
🪟 ✍🏼 Update catalog diff modal confirm label to 'OK'

edmundito:
## What

Updates the confirm button text in the catalog diff modal from "Confirm" to "OK".


dizel852 (approved):
LGTM, not tested locally

Issue #20771, opened 2022-12-21
[connector builder] limit the number of requests submitted to the endpoint


girarda:
## Tell us about the problem you're trying to solve
We should limit the number of requests the connector builder can send as part of a "test read" to limit how expensive a read is. 

As a result: 
* you'll have to wait at the UI a long time
* we could use too much API quota
* In cloud it could be exploited 

## Describe the solution you’d like
A way to limit the number of requests submitted to an API endpoint because of a single "test read"

## Acceptance criteria
- The server has a hard limit on the number of requests per slice and number of slices (initial proposal: 5)
- If a read call attempts to do more than the limit of allowed requests, it will stop and return the data gathered so far. A flag on the response is set so the UI can react to the fact the read command got limited
- If this flag is set, the UI shows a warning callout that only a certain number of requests has been made

## Implementation details
- This will require updating the `HttpStream` so it stops paginating after X requests
- (to validate) this will require update the `AbstractSource` so it stops iterating over the stream slices after X requests

## Open questions
- Do we also need this limit on OSS?
  - To keep things simple IMHO we shouldn't have different behavior on OSS
- Should the be configurable (with a max)?
  - Not configurable for now IMHO
- What should the max be? 
  - IMHO going with 5 - 10 for now would be sufficient
- What about secondary requests like for oauth/session token auth?
  - Ignore for the scope of this ticket
  - Create separate issue for consolidating request handling
- What about complex slice/page combinations where 10 is too little and it wouldn't hit the second slice / be not representative in some way?
  - Apply the limit not the overall number of requests, but
    - Limit requests per slice
    - Limit slices

flash1293:
Related to that and also important:
* Fail faster if API is not available
* Right now default behavior is doing really long exponential backoff which makes sense during production but not while iterating
* Maybe we can solve this in a similar way?

flash1293:
Side note: Should we cancel the read once the upstream connection gets closed? Might be separate issue

PR #21525, opened 2023-01-18
[ISSUE #20771] limiting the number of requests performed to the backe…

maxi297:
…nd without flag

## What
This addresses the first part of #20771 where we want to limit the number of requests done to the back. As for this PR, the flag is excluded and another PR will follow.

## How
Change the behaviour at runtime of the paginator and the simple retriever.

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py`
2. `airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/default_paginator.py`
3.  The rest

## 🚨 User Impact 🚨
Users of the connector builder should see a great improvement in the response time when reading a stream. However, the response can now be partial if there is more than 5 slices of 5 pages


airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/default_paginator.py
```diff
@@ -160,3 +160,68 @@ def _get_request_options(self, option_type: RequestOptionType) -> Mapping[str, A
             if option_type != RequestOptionType.path:
                 options[self.page_size_option.field_name] = self.pagination_strategy.get_page_size()
         return options
+
+
+class PaginatorTestReadDecorator(Paginator):
```
maxi297, line 165: Is there a reason why Paginator does not inherit ABC? As of now, we won't get a test error if a new method is added to Paginator

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -116,8 +116,70 @@
 
 
 class ModelToComponentFactory:
-    @staticmethod
-    def create_component(model_type: Type[BaseModel], component_definition: ComponentDefinition, config: Config) -> type:
+    def __init__(self, is_test_read=False):
```
girarda, line 119: what do you think about passing two flags so the connector builder can control the number of pages and the number of slices independently without having to republish the CDK?

girarda (approved):
minor request to use two separate flags instead of a generic `is_test_read` flag, but looks good otherwise

airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/default_paginator.py
```diff
@@ -160,3 +160,68 @@ def _get_request_options(self, option_type: RequestOptionType) -> Mapping[str, A
...
+    pages that are queried throughout a read command.
+    """
+
+    _PAGE_COUNT_BEFORE_FIRST_NEXT_CALL = 1
+    _DEFAULT_PAGINATION_LIMIT = 5
+
+    def __init__(self, decorated, maximum_number_of_pages: int = None):
+        self._decorated = decorated
+        self._page_count = self._PAGE_COUNT_BEFORE_FIRST_NEXT_CALL
+        self._maximum_number_of_pages = maximum_number_of_pages if maximum_number_of_pages else self._DEFAULT_PAGINATION_LIMIT
```
brianjlai, line 177: Instead of `maximum_number_of_pages: int = None` and the ternary, can we just set the default for `maximum_number_of_pages: int = _DEFAULT_PAGINATION_LIMIT` directly in the signature?

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: seconding this

brianjlai (approved):
a small nit comment, but otherwise looks good to me!

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
maxi297: This is part of https://github.com/airbytehq/airbyte/pull/21619 where we pass the maximum number of requests we want to allow. This allows us to calculate the flag

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -236,7 +239,8 @@ def _read_incremental(
         has_slices = False
         for _slice in slices:
             has_slices = True
-            logger.debug("Processing stream slice", extra={"slice": _slice})
+            if logger.isEnabledFor(logging.DEBUG):
+                yield AirbyteMessage(type=MessageType.LOG, log=AirbyteLogMessage(level=Level.INFO, message=f"slice:{json.dumps(_slice)}"))
```
girarda, line 243: nit: should "slice:", "request:", and "response:" be constants?

airbyte-connector-builder-server/connector_builder/impl/default_api.py
```diff
@@ -164,33 +177,53 @@ def _get_message_groups(self, messages: Iterator[AirbyteMessage], schema_inferre
...
         current_page_response: Optional[HttpResponse] = None
 
-        while len(current_records) < limit and (message := next(messages, None)):
-            if first_page and message.type == Type.LOG and message.log.message.startswith("request:"):
-                first_page = False
-                request = self._create_request_from_log_message(message.log)
-                current_page_request = request
+        while records_count < limit and (message := next(messages, None)):
+            if self._need_to_close_page(at_least_one_page_in_group, message):
+                self._close_page(current_page_request, current_page_response, current_slice_pages, current_page_records)
```
girarda, line 189: should this branch also set `current_page_request` and `current_page_response` to `None` to protect ourselves against a malformed stream of messages (due to a bug in the CDK)?

example scenario:
1. request
2. response
3. record
4. (no request)
5. response

girarda (approved):
looks good. only a couple of small questions 🚢

PR #21531, opened 2023-01-18
🪟🔧 Refactor ConnectorsView

### LINKED ISSUE #20706: Refactor ConnectorsView
The ConnectorsView component is using mostly custom styling, which causes it to look slightly different than the rest of the webapp in some cases and be less responsive than possible. The component should be cleaned up, both from a technical perspective and a usability perspective:
* Get rid of styled components
* Get rid of large margins and paddings to layout the form, instead rely on default table layouting
* Use standard form components for the "New connector" modal
* Properly show loading state while connector version is upgraded
* Do not conditionally hide the "Upgrade all" button, instead make it disabled with a tooltip to explain everything is up to date

## Acceptance criteria

* No styled components and as little hardcoded widths and custom styles as possible, relying on component library
* It's obvious when a connector is currently in the process of being updated
* The upgrade button does not disappear after clicking

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/20706

## How

The following changes have been done:
* Remove styled component from and extend the indicator component so it can render invisible - this makes it easy for things to be aligned which have a leading optional indicator like in the connectors table
* Instead of rendering a custom string for custom connectors, use a badge like for alpha and beta connectors
* Show a loading indicator in the "Upgrade" button while a single connector is updating
* Do not show the loading indicator for connectors that don't actually update when hitting "Upgrade all" (only the ones that are actually upgradable at the moment)
* Remove styled components from different parts of the connectors table
* To prevent often occurring re-render of full table, introduce a context to pass down the loading state to each individual loading button
* Use standard components for the "Add custom connector" dialog
* Always show the "Upgrade all" button so the layout doesn't jump due to it


airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/SourcesPage.tsx
```diff
@@ -15,33 +15,39 @@ const SourcesPage: React.FC = () => {
 
   const [isUpdateSuccess, setIsUpdateSuccess] = useState(false);
   const [feedbackList, setFeedbackList] = useState<Record<string, string>>({});
+  const feedbackListRef = useRef(feedbackList);
+  feedbackListRef.current = feedbackList;
```
lmossman, line 19: Why was this change made? And is it intentional that this change was only made to SourcesPage and not also to DestinationsPage?

Side note: it seems like the logic in those two components is almost exactly the same, so a lot of it could be moved into the `ConnectorsView` component, with props to differentiate between the source- or destination-specific methods. What do you think? Could also be done in a separate PR if you'd prefer - just thought I'd call it out here since we are already refactoring in this PR

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/CreateConnectorModal.tsx
```diff
@@ -98,167 +43,149 @@ const CreateConnectorModal: React.FC<CreateConnectorModalProps> = ({ onClose, on
...
-                    <FormattedMessage id="form.add" />
-                  </Button>
-                </div>
-              </ButtonContent>
-            </Form>
-          )}
-        </Formik>
-      </Content>
-    </Modal>
+    <Formik
```
lmossman, line 46: Just curious, why did you move the `Formik` outside of the `Modal`?

lmossman (approved):
A couple questions but nothing truly blocking. LGTM and seems to work well from local testing

airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/SourcesPage.tsx
flash1293: I meant to do the same change for the destination, thanks for catching this. This is done to prevent a full re-render of the table on state change of the feedback list. If onUpdateVersion changes, then renderColumns changes here https://github.com/airbytehq/airbyte/blob/fd9185363e94b8c7241c23e6d3bd18b37454174c/airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/ConnectorsView.tsx#L124

which in turn means the columns object passed to the table changes here: https://github.com/airbytehq/airbyte/blob/fd9185363e94b8c7241c23e6d3bd18b37454174c/airbyte-webapp/src/pages/SettingsPage/pages/ConnectorsPage/components/ConnectorsView.tsx#L151

which should be avoided (the docs say that the columns should be memoized: https://react-table-v7.tanstack.com/docs/api/useTable which means if they change reference it's pretty expensive)

Issue #21492, opened 2023-01-17
Performance - background json schema validation


colesnodgrass:
This is a potential low-hanging issue that was discovered in one of Davin's tests, and it may be responsible for some of the speed-up seen when parallelizing the message-processing.

I'm going to attempt to do this behind a feature-flag, so I'll be going through the process of getting the proper accounts setup to get this working.

colesnodgrass:
Fixed in airbytehq/airbyte-platform-internal#4758

PR #21534, opened 2023-01-18
add perf feature-flag; ensure compatability with LaunchDarkly (LDv5) users

colesnodgrass:
## What
- part of #21492 
- update feature-flag client capabilities
    - updated `ConfigFileClient` to support passing in a `null` config file path
        - if a `null` config is provided, the requested feature-flag default values will always be returned
        - this was changed to make instantiating the `FeatureFlagClient` easier
    - enable the `Multi` contexts to work with LDv5
        - convert the context-key to LDv5 user attributes so feature-flags can be correctly targeted 
    - pass LDv6 contexts as LDv5 user attributes
        - as our LaunchDarkly account does not yet have access to contexts, this ensures that our `Context` attributes are set correctly on the LaunchDarkly user
- add `performance.backgroundJsonSchemaValidation` feature flag
    - to both the `flags.yml` and `Flags.kt` files

## Recommended reading order
1. `Flags.kt`
2. `flags.yml`
3. `Context.kt`
4. `Client.kt`
5. tests


airbyte-featureflag/src/main/kotlin/config/Factory.kt
```diff
@@ -20,22 +19,22 @@ internal const val CONFIG_OSS_KEY = "airbyte.feature-flag.path"
...
-    @Requirements(
-        Requires(property = CONFIG_OSS_KEY),
-        Requires(missingProperty = CONFIG_LD_KEY),
-    )
-    fun Platform(@Property(name = CONFIG_OSS_KEY) configPath: String): FeatureFlagClient {
-        val path = Path.of(configPath)
+    @Requires(missingProperty = CONFIG_LD_KEY)
+    @Singleton
+    fun ConfigFileClient(@Property(name = CONFIG_OSS_KEY) configPath: String): FeatureFlagClient {
+        val path: Path? = if (configPath.isNotBlank()) {
```
jdpgrailsdev, line 33: Out of curiosity, Kotlin doesn't have a more succinct way to represent this type of if/else?  Or, are you just avoiding using a ternary or elvis operator for stylistic/readability reasons?

jdpgrailsdev (approved):
:shipit:

benmoriceau (approved):

gosusnp (approved):

PR #21541, opened 2023-01-18
[Connector Builder] Move start/end time options out of optional block

lmossman:
## What
As discussed in this [slack thread](https://airbytehq-team.slack.com/archives/C03TP6W8081/p1673563857960709?thread_ts=1673562210.818799&cid=C03TP6W8081), we want to make the start/end time request option configurations more visible by default by moving them out of the Optional block.

## How
Rearrange the fields in the stream slicer component to be outside of the BuilderOptional block.


flash1293 (approved):
LGTM

PR #21543, opened 2023-01-18
[ISSUE #19981] fix unexpected tag error

maxi297:
## What
Fix this issue https://github.com/airbytehq/airbyte/actions/runs/3951883480

### DIFF
--- a/.github/workflows/publish-cdk-command-manually.yml
+++ b/.github/workflows/publish-cdk-command-manually.yml
@@ -52,7 +52,6 @@ jobs:
           new_version="$(grep -i 'current_version = ' .bumpversion.cfg | sed -e 's/.* = //')"
           awk -v NEW_VERSION="$new_version" -v CHANGELOG_MESSAGE="${{ github.event.inputs.changelog-message }}" 'NR==3{print "## " NEW_VERSION "\n" CHANGELOG_MESSAGE "\n"}1' CHANGELOG.md > tmp && mv tmp CHANGELOG.md
       - name: Commit and Push Changes
-        if: success()
         uses: stefanzweifel/git-auto-commit-action@v4
         with:
           file_pattern: airbyte-cdk/python/setup.py airbyte-cdk/python/.bumpversion.cfg airbyte-cdk/python/CHANGELOG.md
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-cdk-command-manually.yml
```diff
@@ -167,7 +166,6 @@ jobs:
           TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
           TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
       - name: Post success to Slack channel dev-connectors-extensibility
-        if: success()
```
maxi297, line 170: Based on https://docs.github.com/en/actions/learn-github-actions/expressions#success, this is the default condition so no need to specify it here

.github/workflows/publish-cdk-command-manually.yml
```diff
@@ -52,15 +52,14 @@ jobs:
           new_version="$(grep -i 'current_version = ' .bumpversion.cfg | sed -e 's/.* = //')"
           awk -v NEW_VERSION="$new_version" -v CHANGELOG_MESSAGE="${{ github.event.inputs.changelog-message }}" 'NR==3{print "## " NEW_VERSION "\n" CHANGELOG_MESSAGE "\n"}1' CHANGELOG.md > tmp && mv tmp CHANGELOG.md
       - name: Commit and Push Changes
-        if: success()
```
maxi297, line 55: Based on https://docs.github.com/en/actions/learn-github-actions/expressions#success, this is the default condition so no need to specify it here

.github/workflows/publish-cdk-command-manually.yml
```diff
@@ -195,7 +193,7 @@ jobs:
         env:
           SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN_AIRBYTE_TEAM }}
       - name: Post failure to Slack channel dev-connectors-extensibility
-        if: !success()
+        if: ${{ failure() }}
```
erohmensing, line 196: nit: `if: failure()` should work too

.github/workflows/publish-cdk-command-manually.yml
erohmensing, line 170: I think we still want to keep this one - We don't want to post success to slack every case, only if we were successful (Different from the one you removed in `Commit and push changes` , which we want it to attempt to do each time)

erohmensing (changes requested):

erohmensing (approved):

PR #20960, opened 2023-01-03
🎉 New Source: Railz

Allexik:
## What
Issue: https://github.com/airbytehq/airbyte/issues/17604
Added new source for railz.ai

## How
In low-code

## Recommended reading order
1. `railz_ai.yaml`
2. `y.python`

## Pre-merge Checklist

<details><summary><strong>New Connector</strong></summary>

### Community member or Airbyter

- [ ] **Community member?** Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [x] Secrets in the connector's spec are annotated with `airbyte_secret`
- [x] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [x] Documentation updated
    - [x] Connector's `README.md`
    - [x] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - [x] `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
    - [x] `docs/integrations/README.md`
    - [x] `airbyte-integrations/builds.md`
- [x] PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests)

### Airbyter

If this is a community PR, the Airbyte engineer reviewing this PR is responsible for the below items.

- [x] Create a non-forked branch based on this PR and test the below items on it
- [x] Build is successful
- [x] If new credentials are required for use in CI, add them to GSM. [Instructions](https://docs.airbyte.io/connector-development#using-credentials-in-ci).
- [x] [`/test connector=connectors/<name>` command](https://docs.airbyte.io/connector-development#updating-an-existing-connector) is passing
- [ ] New Connector version released on Dockerhub by running the `/publish` command described [here](https://docs.airbyte.io/connector-development#updating-an-existing-connector)
[...]

PR #21546, opened 2023-01-18
Low-Code CDK: AccessTokenAuthenticator added

grubberr:

## What
- This auth method will be used in `source-railz-ai` https://github.com/airbytehq/airbyte/pull/20960
- In really I don't like name `AccessTokeAuthenticator` for this auth component. Can you advice your solution?


brianjlai (changes requested):
For your initial question, I think something like ShortLivedTokenAuthenticator would be a good name since this is basically implementing a specialized flow to refresh an access token via username/password.

I think a significant concern I have around this new component is that it feels very specialized to railz-ai's API implementation. I've seen a general process for short lived access token + a refresh token flow, but this kind of flow where we insert client_id and client_secret into the username/password to get an access token does not seem to be a common practice. Have you been seeing other APIs that are being implemented in low-code following this same pattern?

My thinking is that we should have a high standard for what components we introduce into the language and their interface should support the use case of many connectors not just one. I apologize since I should have flagged this earlier in the prior railz-ai review, but based on my review, this seems to be a better candidate for a custom component. Functionally this acts a lot like a BasicHttpAuthenticator in that its primary inputs are username = client_id, password = secret_key (and a authentication url). And it performs potentially two steps in the `token()` function as you have it implemented here. So unless we have a clear set of connectors that all require this type of token flow, we should start with this as a custom component and promote it into the language at a later time.

grubberr:
@brianjlai After your review 
I have moved the custom component "ShortLivedTokenAuthenticator" to the Railz connector
as you proposed
I will close this PR as soon as [PR](https://github.com/airbytehq/airbyte/pull/20960) with the new connector will be merged to master

brianjlai:
@grubberr thank you for the explanation and that plan sounds good to me

PR #21554, opened 2023-01-18
🪟 🎨 Fix stream table heading design issues

edmundito:
## What
* Makes info icon 11px, which more closely matches design
* Update table heading to use same components as stream details panel
* Move checkbox to align horizontally with source icon

image


krishnaglick (approved):
Code LGTM, did not test locally. Looks like there's some snapshot failures.

dizel852 (approved):
LGTM 👍 
Tested locally - works as expected
image

PR #14452, opened 2022-07-06
GitHub Actions - workflow `Connector Integration Tests` - add retry

grubberr:
Signed-off-by: Sergey Chvalyuk <grubberr@gmail.com>

## What
1. Rewrite shell script `ci_integration_workflow_launcher.sh` to python script `ci_integration_workflow_launcher.py`

2. add UUID for workflow `integration-test` to track progress
    https://stackoverflow.com/questions/69479400/get-run-id-after-triggering-a-github-workflow-dispatch-event
    
3. `ci_integration_workflow_launcher.py` run all workflows, sleep for 20 minutes and re-run AWS failed workflows again

4. `tools/bin/build_report.py` - date added for report

PR #21562, opened 2023-01-18
CI: build alpha sources and destinations

alafanechere:
## What
The nightly builds we run on connector is not building alpha connectors.
We need to get the build status for all connectors for the [Deploy all eligible connectors to Cloud project](https://github.com/airbytehq/airbyte/issues/20553).

@grubberr could you please explain why alpha connectors are not currently built? The release stage filtering happened in https://github.com/airbytehq/airbyte/pull/14452 but no details are shared why we only build GA and beta connectors.

## How
Explicitly ask the `ci_integration_workflow_launcher.py` to build alpha connectors.


evantahler (approved):
Yep... this is what we want to happen!

I imagine we'll need to increase some timeouts in the script to build more connectors?  Do we need more AWS action runners?  

cc @YowanR, this will likely make the data you see in #connector-health increase to include alpha connectors as well.  Triaging that list in a tool other than slack will probably need to come sooner rather than later.

alafanechere:
I double checked the code that's posting the Slack message on connector health. A filter Ain release stage is happening g there too. So no change to the slack message is expected.

PR #21565, opened 2023-01-18
[Connector Builder Server] add endpoint to resolve $refs and $options

### LINKED ISSUE #21211: Connector builder server endpoint to resolve $refs and $options
## Context

In order to convert YAML manifests into UI form values in the connector builder, the frontend needs a way to resolve all $refs and $options in the YAML so that the values can be filled out in the builder forms.

Since the backend server is already doing this when running a manifest, this is a natural place to implement this logic.

## Acceptance Criteria
- A connector builder server endpoint exists which takes in a JSON manifest and returns a JSON manifest with all of the $refs and $options resolved and flattened (e.g. `/v1/manifest/resolve`)
  - This should also infer `type` values and add them for each component to the returned manifest
  - Returns an error response if the manifest is invalid
  - **Note**: this should _not_ resolve interpolated values like {{config[]}}, since we want to propagate those values to the UI forms

clnoll:
Closes https://github.com/airbytehq/airbyte/issues/21211.

## What
Provide an endpoint for the connector builder to resolve all $refs and $options in a manifest so that the values can be filled out in the builder forms.

## How
Adds a new endpoint, `manifest/resolve`, to the connector builder server that takes a manifest in the request body and returns a manifest with all $refs and $options resolved.

The resolving of $refs and $options is done in a new method of `ManifestDeclarativeSource`, `resolved_manifest`. Although this method is currently only used by the connector builder server (i.e. doesn't need to be in the CDK), this felt like a natural place for this logic given that the acceptance criteria requires that an error response is returned if the manifest is invalid; in addition to resolving $refs and $options, `ManifestDeclarativeSource` does some validation on instantiation, and sets the resolved manifest as an attribute on the instance.


airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -81,6 +81,9 @@ def __init__(self, source_config: ConnectionDefinition, debug: bool = False, con
         if unknown_fields:
             raise InvalidConnectorDefinitionException(f"Found unknown top-level fields: {unknown_fields}")
 
+    def resolved_manifest(self) -> Mapping[str, Any]:
```
maxi297, line 84: nit: I don't know about the team conventions: should this be a `@property`? If not, should it be rename using an action verb like `resolve_manifest`?

airbyte-connector-builder-server/connector_builder/impl/default_api.py
```diff
@@ -144,11 +148,34 @@ async def read_stream(self, stream_read_request_body: StreamReadRequestBody = Bo
...
+        try:
+            return ManifestResolve(
+                manifest=ManifestDeclarativeSource(
+                    manifest_resolve_request_body.manifest, construct_using_pydantic_models=True
+                ).resolved_manifest()
+            )
+        except Exception as error:
+            self.logger.error(f"Could not resolve manifest with with error: {error.args[0]} - {self._get_stacktrace_as_string(error)}")
+            raise HTTPException(
+                status_code=400,
```
maxi297, line 172: It seems weird to me to assume that this is a user input error (i.e. a `400` error) and not an internal server error. I know @lmossman wanted to track stream test errors. Will this enable us to do this? 

Note: This can probably be scoped out of the PR. My main goal is to bring awareness and make sure we are aligned on short term incoming changes

airbyte-connector-builder-server/src/main/openapi/openapi.yaml
```diff
@@ -69,6 +69,27 @@ paths:
...
                 description: Connector manifest template string
+  /v1/manifest/resolve:
+    post:
+      summary: Given a JSON manifest, returns a JSON manifest with all of the $refs and $options resolved and flattened
+      operationId: resolveManifest
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/ManifestResolveRequestBody"
```
maxi297, line 80: nit: The endpoint is called `resolve_manifest`. Should the input be called `ResolveManifestRequestBody` instead of `ManifestResolveRequestBody`?

maxi297 (approved):
The seems to be a lot of change that are associated to formatting. Should we clarify the tooling regarding formatting for the Connector Builder Server?

airbyte-connector-builder-server/connector_builder/impl/default_api.py
lmossman: Does the exception thrown by the `ManifestDeclarativeSource.resolved_manifest()` method provide any indication as to what type of error occurred? Ideally this would return a 400 if the manifest is invalid in some way, but it would return a 500 if some other unrelated issue happened (e.g. some bug in the code causing an uncaught exception)

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
clnoll: I like the idea of making it an `@property`, will do that (and so will keep it as a noun).

airbyte-connector-builder-server/src/main/openapi/openapi.yaml
clnoll: 👍 makes sense to me! Updating.

airbyte-connector-builder-server/connector_builder/impl/default_api.py
clnoll: Currently the `4**` errors are returned for invalid manifests & requests, and an unhandled exception in the code would return a 500, openapi's default.

I took a closer look at the errors surfaced by `resolved_manifest`, which are all raised during initialization of `ManifestDeclarativeSource`. They have varying levels of usefulness - a manifest with an unresolvable reference will return an error stating such (as in the `test_resolve_manifest_unresolvable_references` test case), but manifests that are invalid for other reasons (e.g. missing some fields, as in the `test_resolve_manifest_invalid` test case) may have more generic error messages. If there is a category of errors that you'd like special handling for I'm happy to add it.

PR #21566, opened 2023-01-18
don't fail on test-reporter action and add comments on failures

c-p-b:
## What
This change updates two github actions:

https://github.com/EnricoMi/publish-unit-test-result-action
https://github.com/dorny/test-reporter

The first change forces test-reporter to not fail when the build fails. The reasoning behind this change is that when the build fails, you typically want to see that. However, if test-reporter fails, that's the first thing you see when you click in, which can erroneously lead people to believe that the reporting action failed to upload, when in reality it was the build itself.

The second change updates publish-unit-test-result-action to add a pr comment pointing to specific test failures when they occur. This is to increase visibility of failures to end users.

## How
Update gradle.yml


colesnodgrass (approved):
Curious what this will actually look like, changes seem fine.

PR #21569, opened 2023-01-19
implement PerfBackgroundJsonValidation feature-flag

colesnodgrass:
## What
- builds on #21534 
- add support for checking the `PerfBackgroundJsonValidation` feature-flag within the `RecordSchemaValidator` class

## How
- start plumbing in the `FeatureFlagClient` in all the necessary places to get feature-flag support within the `RecordSchemaValidator`
    - the existing `FeatureFlags` is still present in these files as I wanted to limit the scope changes to implement this single `PerfBackgroundJsonValidation` flag

- this feature-flag doesn't do anything other than change a log message, once functionality has been verified, this will be updated to do real work

- the feature-flag check is done in the following way:
    ```java
    if (workspaceId != null) {
      if (ffClient.enabled(PerfBackgroundJsonValidation.INSTANCE, new Workspace(workspaceId))) {
        log.info("feature flag enabled for workspace {}", workspaceId);
      } else {
        log.info("feature flag disabled for workspace {}", workspaceId);
      }
    } else {
      log.info("workspace id is null");
    }
    ```
    I am unsure if a `workspaceId` will exist at this point in the code (the `syncInput` object supports a `workspaceId` but doesn't require one)

### Micronaut property checking issues
Micronaut does not handle the following situation well:
> if property P is defined and P is a non-empty string, then return bean A, else return bean B

Instead redefine the situation as:
> if property P is defined and equals a well known value, then return bean A, else return B_

To summarize use `@Requires(property=P, value=WELL_DEFINED)` with `@Requires(property=P, notEquals=WELL_DEFINED)`.  *Do not use* `@Requires(property=P)` with `@Requires(missingProperty=P)` or `@Requires(property=P, pattern=NON_EMPTY_REGEX)` with `@Requires(property=P, pattern=EMPTY_REGEX)`

This also appears to be related to this [Micronaut issue](https://github.com/micronaut-projects/micronaut-core/issues/7362)

### Micronaut Kotlin Dependencies
I spent a lot of time with trying to get the Kotlin beans injectable into the java code.  What should _just work_ doesn't and it well documented on the Micronaut side as to what exactly is required (or at least I never found the documentation).  After many attempts I and looking at the auto-generated kotlin Micronaut projects I landed on `kapt` being necessary in kotlin projects.  Even this isn't entirely true.

If the Kotlin project is defined with the following gradle setup, the beans will *not* be visible to java projects (and possibly other projects as well):
```kts
plugins {
    kotlin("jvm") version "1.8.0"
}

dependencies {
    annotationProcessor(platform(libs.micronaut.bom))
    annotationProcessor(libs.bundles.micronaut.annotation.processor)
}
```

But they will be visible with the following setup:   
```kts
plugins {
    kotlin("jvm") version "1.8.0"
    id("io.micronaut.minimal.application") version "3.7.0"
}

dependencies {
    annotationProcessor(platform(libs.micronaut.bom))
    annotationProcessor(libs.bundles.micronaut.annotation.processor)
}

micronaut {
  version("3.8.2")
}
```
_and_ also visible with the following setup:
```kts
plugins {
    kotlin("jvm") version "1.8.0"
    kotlin("kapt") version "1.8.0"
}

dependencies {
    kapt(platform(libs.micronaut.bom))
    kapt(libs.bundles.micronaut.annotation.processor)
}
```

I went with the `kapt` solution. 

It's worth noting that `kapt` is deprecated and is being replaced with `kps` in Micronaut 4 (this is a good thing).


airbyte-commons-worker/src/main/java/io/airbyte/workers/RecordSchemaValidator.java
```diff
@@ -6,26 +6,40 @@
...
+
+  private final FeatureFlagClient ffClient;
+  private final UUID workspaceId;
   private final Map<AirbyteStreamNameNamespacePair, JsonNode> streams;
 
-  public RecordSchemaValidator(final Map<AirbyteStreamNameNamespacePair, JsonNode> streamNamesToSchemas) {
+  public RecordSchemaValidator(final FeatureFlagClient ffClient,
+                               final UUID workspaceId,
+                               final Map<AirbyteStreamNameNamespacePair, JsonNode> streamNamesToSchemas) {
+    this.ffClient = ffClient;
```
benmoriceau, line 41: Nit: I'm not a big fan of the use of abbreviation for variable name, it can make the code hard to understand if we use too much of them. Could we name this variable `featureFlagClient`?

airbyte-workers/build.gradle
```diff
@@ -40,7 +40,7 @@ dependencies {
     implementation 'com.auth0:java-jwt:3.19.2'
     implementation 'io.fabric8:kubernetes-client:5.12.2'
     implementation libs.guava
-    implementation (libs.temporal.sdk) {
+    implementation(libs.temporal.sdk) {
```
davinchia, line 43: curious, why are these changes happening?

colesnodgrass: It's part of the formatting rules we have defined, which I have configured to run on any file that has been saved.  And this is consistent with [other build.gradle file formats](https://github.com/airbytehq/airbyte/blob/31d7eb39c5dd1b078654b47cec3833ec0d1e99c4/airbyte-server/build.gradle#L20-L28).

My guess is that the `gradle format` task only targets java files, which is why these slight formatting differences only fail PR builds.

airbyte-commons-worker/src/main/java/io/airbyte/workers/RecordSchemaValidator.java
```diff
@@ -6,26 +6,40 @@
...
+import org.slf4j.LoggerFactory;
 
 /**
  * Validates that AirbyteRecordMessage data conforms to the JSON schema defined by the source's
  * configured catalog
  */
-
 public class RecordSchemaValidator {
 
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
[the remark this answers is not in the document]
colesnodgrass: I do this entirely for copy/paste friendliness.

airbyte-container-orchestrator/src/test/java/io/airbyte/container_orchestrator/config/ContainerOrchestratorFactoryTest.java
```diff
@@ -33,6 +38,10 @@ class ContainerOrchestratorFactoryTest {
   @Inject
   FeatureFlags featureFlags;
 
+  @Bean
+  @Replaces(FeatureFlagClient.class)
```
davinchia, line 42: What is the 'Replaces' annotation?

davinchia (commented):
Looks sane to me! 

I'm not fully caught up to speed on the actual feature flagging yet. How do we actually update the feature flag? Through the launch darkly UI?

colesnodgrass:
@davinchia 
> I'm not fully caught up to speed on the actual feature flagging yet. How do we actually update the feature flag? Through the launch darkly UI?

LaunchDarkly is used for the cloud deployment. The [flags.yml](https://github.com/airbytehq/airbyte/blob/2b2d6266e8bcc96ea575d076002b514073f37fca/configs/flags.yml) file is used for local/oss deployments.

airbyte-container-orchestrator/src/test/java/io/airbyte/container_orchestrator/config/ContainerOrchestratorFactoryTest.java
colesnodgrass: This tells Micronaut to use _this_ object to replace any `FeatureFlagClient` object it knows about for dependency injection purposes.  Specifically here I'm telling Micronaut to use the `TestClient` implementation of the `FeatureFlagClient` instead of the `LaunchDarklyClient` or `ConfigFileClient` that it would have provided.

davinchia: Thanks for the explanation.

airbyte-featureflag/build.gradle.kts
```diff
@@ -4,11 +4,12 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 plugins {
     `java-library`
     kotlin("jvm") version "1.8.0"
+    kotlin("kapt") version "1.8.0"
```
davinchia, line 7: is this a koitlin specific annotation processor?

airbyte-commons-worker/src/main/java/io/airbyte/workers/RecordSchemaValidator.java
```diff
@@ -43,6 +57,16 @@ public void validateSchema(final AirbyteRecordMessage message, final AirbyteStre
     final JsonNode messageData = message.getData();
     final JsonNode matchingSchema = streams.get(messageStream);
 
+    if (workspaceId != null) {
+      if (featureFlagClient.enabled(PerfBackgroundJsonValidation.INSTANCE, new Workspace(workspaceId))) {
+        log.info("feature flag enabled for workspace {}", workspaceId);
```
davinchia, line 62: what about the separate threading changes? Or does this PR only include the feature flag changes?

davinchia: also to confirm, we'll opt in workspaces via adding the workspace id in LauchDarkly?

colesnodgrass: This only includes the feature-flag changes.  A quick, follow-up PR will add in the actual changes that will be contained within the feature-flag. 

This allows me to vet that the client is working as expected and seeing updated from LaunchDarkly without having an impact on any users.

davinchia (approved):
@colesnodgrass I think we are going to have a merge conflict. Do you mind letting me merge in my PR first before we merge in yours?

PR #21570, opened 2023-01-19
Make destination-clickhouse-strict-encrypt use the correct class

grishick:
## What
destination-clickhouse-strict-encrypt is using `ClickhouseDestination` as the `mainClass` in the build, which means that it is not removing SSL option from the UI. This change fixes `build.gradle` and bumps the version of destination-clickhouse-strict-encrypt.
This change also bumps the version of destination-clickhouse in order to keep it in sync with the strict encrypt version


ryankfu (approved):
LGTM, pending publishing connector

grishick:
/publish connector=connectors/destination-clickhouse run-tests=false

grishick:
/publish connector=connectors/destination-clickhouse run-tests=false

grishick:
/publish connector=connectors/destination-clickhouse-strict-encrypt run-tests=false

PR #21577, opened 2023-01-19
Suggested Streams via Actor Definition

### LINKED ISSUE #10872: Create a notion of streams which are "selected by default"

evantahler:
Closes https://github.com/airbytehq/airbyte/issues/10872, (inspired by https://github.com/airbytehq/oncall/issues/1303#issuecomment-1381772903 OC issue).  [Protocol Change Process Document](https://docs.google.com/document/d/1RnoFFr12ZX0WGsjicyC_ErvSib5aP8A0DZ-IPzZcB1g/edit#).

## What?

This PR allows connectors to suggest which streams are pre-selected for users creating a new connection.  This is accomplished by adding a new “suggestedStreams" block to the source actor definitions. This change is backwards-compatible and does not require the re-publishing of existing connectors. 

For many of our API sources, the connector provides *a lot* of streams.  This is great because Airbyte connectors come with "batteries included" to enable the widest range of data-syncing use-cases.  However, for some sources, most users probably don't want all of the streams.  Some of those streams may be extraneous, but worse, they may be slow, expensive, and often make the syncing experience for the more "primary" streams worse. Perhaps this stream has a slower API, consumes more API "credits", or harms the global rate limit in a worse-than-normal way.   

This PR allows each connector to tune the default stream selections to provide a better first-time sync experience.  

Screenshot 2023-01-18 at 9 36 49 PM

> The Convenience Principle: People will do what's convenient, and they will avoid what's inconvenient. All else being equal, to increase how many people do something, make it more convenient, and to decrease, make it less convenient
> - @sherifnada 

This PR comes with an example for `soruce-github`, a source with particularly egregious streams, which we don't think many people would opt-into.

## Technical notes:

* No API or webapp changes needed
* All existing connectors will work the same way - defaulting to selecting all streams for a new connection
* All that connector developers need to do is supply a list of suggestedStreams in `source_definitions.yaml` if they want to select only some streams.  Otherwise, all streams will remain suggested as they are today.


airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/DbConverter.java
```diff
@@ -160,7 +162,13 @@ public static StandardSourceDefinition buildStandardSourceDefinition(final Recor
             : record.get(ACTOR_DEFINITION.RELEASE_DATE).toString())
         .withResourceRequirements(record.get(ACTOR_DEFINITION.RESOURCE_REQUIREMENTS) == null
             ? null
-            : Jsons.deserialize(record.get(ACTOR_DEFINITION.RESOURCE_REQUIREMENTS).data(), ActorDefinitionResourceRequirements.class));
+            : Jsons.deserialize(record.get(ACTOR_DEFINITION.RESOURCE_REQUIREMENTS).data(), ActorDefinitionResourceRequirements.class))
+        .withAllowedHosts(record.get(ACTOR_DEFINITION.ALLOWED_HOSTS) == null
+            ? null
+            : Jsons.deserialize(record.get(ACTOR_DEFINITION.ALLOWED_HOSTS).data(), AllowedHosts.class))
```
evantahler, line 168: https://github.com/airbytehq/airbyte/pull/21363 forgot this part...

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
```diff
@@ -602,6 +604,18 @@
...
+      - branches
+      - comments
+      - issues
+      - organizations
+      - pull_requests
+      - repositories
+      - stargazers
+      - tags
+      - teams
+      - users
```
evantahler, line 618: Here's the new interface for connector developers who only want some streams selected by default

airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java
```diff
@@ -98,29 +99,51 @@ private static io.airbyte.protocol.models.AirbyteStream toProtocol(final Airbyte
         .withNamespace(stream.getNamespace());
   }
 
-  public static io.airbyte.api.model.generated.AirbyteCatalog toApi(final io.airbyte.protocol.models.AirbyteCatalog catalog) {
+  public static io.airbyte.api.model.generated.AirbyteCatalog toApi(final io.airbyte.protocol.models.AirbyteCatalog catalog,
```
jdpgrailsdev, line 102: There appears to be a formatting issue in this file, based on the build checks.

airbyte-db/db-lib/src/main/java/io/airbyte/db/instance/configs/migrations/V0_40_28_001__AddSuggestedStreams.java
```diff
@@ -0,0 +1,39 @@
...
+    // As database schema changes, the generated jOOQ code can be deprecated. So
+    // old migration may not compile if there is any generated code.
+    try (final DSLContext ctx = DSL.using(context.getConnection())) {
+      addSuggestedStreams(ctx);
+    }
+  }
+
+  private static void addSuggestedStreams(final DSLContext ctx) {
+    ctx.alterTable("actor_definition")
+        .addColumnIfNotExists(DSL.field(
```
sherifnada, line 33: I think a naming which uses `default` like `defaultEnabledStreams` or `streamsSyncedByDefault` would be a clearer naming but ultimately a nit

sherifnada (changes requested):
This change makes sense to me! But shouldn't we also return it from the API? That's my only change request.

evantahler:
> This change makes sense to me! But shouldn't we also return it from the API? That's my only change request.

You got it @sherifnada !
Screenshot 2023-01-24 at 1 56 56 PM

evantahler:
Ping! @airbytehq/connector-operations - any feedback?

sherifnada (approved):
LGTM but the change is mostly happening in platform code so they should be the main reviewers

airbyte-api/src/main/openapi/config.yaml
```diff
@@ -3879,7 +3879,10 @@ components:
           description: Alias name to the stream to be used in the destination
           type: string
         selected:
-          description: If this is true, the stream is selected with all of its properties.
+          description: If this is true, the stream is selected with all of its properties. For new connections, this considered if the stream is suggested or not
```
pedroslopez, line 3882:
```suggestion
          description: If this is true, the stream is selected with all of its properties. For new connections, this considers if the stream is suggested or not
```

pedroslopez (approved):
🦦

LGTM as long as tests are passing 😄

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/helpers/CatalogConverter.java
```diff
@@ -103,29 +104,57 @@ private static io.airbyte.protocol.models.AirbyteStream toConfiguredProtocol(fin
...
         .cursorField(stream.getDefaultCursorField())
         .destinationSyncMode(io.airbyte.api.model.generated.DestinationSyncMode.APPEND)
         .primaryKey(stream.getSourceDefinedPrimaryKey())
-        .selected(true);
+        .selected(!suggestingStreams)
+        .suggested(true);
+
+    if (suggestingStreams) {
+      if (suggestedStreams.contains(stream.getName())) {
+        result.setSelected(true);
```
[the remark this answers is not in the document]
evantahler: It's set right above the `if()` block

timroes:
Just to double check (since I haven't ran this locally, just looked at the code): This selection only takes place when we newly create a connection for the first time, not every time we refresh the catalog? I.e. an already saved connection where the user does a "refresh from source" won't get all their stream selected states overwritten by this logic?

evantahler:
> Just to double check (since I haven't ran this locally, just looked at the code): This selection only takes place when we newly create a connection for the first time, not every time we refresh the catalog? I.e. an already saved connection where the user does a "refresh from source" won't get all their stream selected states overwritten by this logic?

Yep!

evantahler:
> Just to double check (since I haven't ran this locally, just looked at the code): This selection only takes place when we newly create a connection for the first time, not every time we refresh the catalog? I.e. an already saved connection where the user does a "refresh from source" won't get all their stream selected states overwritten by this logic?

Yep!
I've also added additional detail to the API response so in the future we can highlight "suggested by not selected" streams if we want

airbyte-config/config-models/src/main/resources/types/SuggestedStreams.yaml
```diff
@@ -0,0 +1,14 @@
...
+"$id": https://github.com/airbytehq/airbyte/blob/master/airbyte-config/models/src/main/resources/types/SuggestedStreams.yaml
+title: SuggestedStreams
+description: A source's suggested streams.  These will be suggested by default for new connections using this source.  Otherwise, all streams will be selected.
+type: object
+required:
+additionalProperties: true
+properties:
+  streams:
+    type: array
+    description: An array of streams that this connector suggests the average user will want.  SuggestedStreams not being present for the source means that all streams are suggested.  An empty list here means that no streams are suggested.
```
erohmensing, line 12: Thanks for thinking this through and laying it out, I'm sure it'll be helpful in some weird issue a year from now

erohmensing (commented):
Just checking this out, looks exciting!

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/helpers/CatalogConverter.java
```diff
@@ -103,29 +104,57 @@ private static io.airbyte.protocol.models.AirbyteStream toConfiguredProtocol(fin
...
     final io.airbyte.api.model.generated.AirbyteStreamConfiguration result = new io.airbyte.api.model.generated.AirbyteStreamConfiguration()
         .aliasName(Names.toAlphanumericAndUnderscore(stream.getName()))
         .cursorField(stream.getDefaultCursorField())
         .destinationSyncMode(io.airbyte.api.model.generated.DestinationSyncMode.APPEND)
         .primaryKey(stream.getSourceDefinedPrimaryKey())
-        .selected(true);
+        .selected(!suggestingStreams)
+        .suggested(true);
+
+    if (suggestingStreams) {
```
mfsiega-airbyte, line 144: Nit: what about `result.setSelected(suggestingStreams && suggestedStreams.contains(stream.getName()))`? In general, I think `setValue(x)` is more readable than `if (x) setValue(true)`.

mfsiega-airbyte: 🤦 noticed that it's setting different fields on the the result.

Not sure if that says anything about whether it could be made more readable, or that I should just read more carefully.

airbyte-config/config-models/src/main/resources/types/SuggestedStreams.yaml
```diff
@@ -0,0 +1,14 @@
+---
```
malikdiarra, line 1: Can we add more metadata?
like a description or a title that would indicate what use case the suggested streams are trying to achieve?  
It seems your github example exclude some streams related to workflow for example. Those can be useful to investigate CI. 
(There might be some connectors that would also benefit from multiple suggestedStream with different use case each but that might be beyond the scope here)

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
malikdiarra: I am a bit surprised that this is in the definition yaml file when the schema itself is not present in that file but requires calling a connector function and the suggestedStreams are definitely tied to the schema.

I'm not advocating moving that to a connector function but that might be a good signal that we would benefit from having the schema show up in the definition for API connectors.

malikdiarra (commented):
Thanks for that, I think the idea is good. I think we can do additional changes after that to make it even better. One of them being highlighting a bit better in the UI which streams are on and not require people to scroll past dozen of deactivated streams.

airbyte-config/config-models/src/main/resources/types/SuggestedStreams.yaml
evantahler: For sure! I added more context in c025313a0c6.

I like the idea of context-aware suggestions for the future!  That's why this is an object.  @michel-tricot and I brainstormed a `streams-matcher` array which might one day contain the logic for database sources to deselect tables with foreign keys, or any stream coming from a non-public namespace/scheams.

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
evantahler: Yeah... that was the original proposal (https://github.com/airbytehq/airbyte/pull/21437) complete with protocol change.  For posterity, I added a comment in that PR as to why we chose to move this out of the protocol, and you can read more details [here](https://docs.google.com/document/d/1RnoFFr12ZX0WGsjicyC_ErvSib5aP8A0DZ-IPzZcB1g/edit).  

In the near future, all of "actor definitions" will be replaced with "connector metadata", and things will hopefully begin to make more sense.

evantahler:
> Thanks for that, I think the idea is good. I think we can do additional changes after that to make it even better. One of them being highlighting a bit better in the UI which streams are on and not require people to scroll past dozen of deactivated streams.

Exactly! That's why the API is now returning some additional information which can enable that

PR #21587, opened 2023-01-19
Small spec fixes to make sure they work with connector form UI

flash1293:
https://github.com/airbytehq/airbyte/pull/21451 is making the SATs more strict w.r.t. UI limitations - some schemas are valid JSON schema, but will cause small issues in the UI. This PR is fixing the connectors causing issues:

## s3

Enum was encoded within an `allOf` (see https://github.com/pydantic/pydantic/issues/2088) - added a rewrite logic to fix the schema so it works correctly in the UI.

As this changed the previous schema I had to disable the compatibility check

## sendgrid

Date had either integer or string set as value - the UI will always produce strings anyway, reflected this in the schema


airbyte-integrations/connectors/source-amazon-ads/integration_tests/spec.json
```diff
@@ -51,7 +51,7 @@
         "order": 6,
         "type": "array",
         "items": {
-          "type": "integer"
+          "type": "string"
```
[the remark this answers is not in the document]
flash1293: Thanks for flagging this - I wondered how these configs got created and realized that I missed the “array of integers” support in the UI which means I can revert this change. I will update this PR and the SAT extension.

flash1293:
@girarda For the sendgrid one I checked and right now users have to use the integer format (specifying a unix timestamp). So we should run a migration for these. Alternatively we could just take away the string format (because so far users can't use that), but it seems like only allowing strings for a start time is a much better choice than only allowing unix timestamps.

airbyte-integrations/connectors/source-s3/Dockerfile
```diff
@@ -17,5 +17,5 @@ COPY source_s3 ./source_s3
 ENV AIRBYTE_ENTRYPOINT "python /airbyte/integration_code/main.py"
 ENTRYPOINT ["python", "/airbyte/integration_code/main.py"]
 
-LABEL io.airbyte.version=0.1.28
+LABEL io.airbyte.version=0.1.29
```
girarda, line 20: should this be a minor update / is this a breaking change?

girarda (approved):
approved assuming the change is S3 is not breaking

airbyte-integrations/connectors/source-s3/Dockerfile
flash1293: no, that's actually a patch change - it doesn't change what's allowed for this field, it just makes sure the UI understands correctly.

PR #21589, opened 2023-01-19
Revert: CI  nightly build alpha sources and destinations

alafanechere:
This reverts commit 61f88f30134df1a71466cd63e0d73757a7ad1d8d.

## What
There's a timeout in the GHA running Integration tests for connector.
All connectors's can't be built in this time window.

PR #21591, opened 2023-01-19
Connector daily integration tests: new job for alpha connectors

alafanechere:
## What
Declare a separate job to run `alpha` connectors build only in the daily Connector Integration Tests run.
I reverted https://github.com/airbytehq/airbyte/pull/21589 because all the builds can't run in 30mn.
For orchestration and timeout reasons I prefer to declare it in a second job.

### DIFF
--- a/.github/workflows/connector_integration_tests.yml
+++ b/.github/workflows/connector_integration_tests.yml
@@ -29,3 +29,22 @@ jobs:
         run: python ./tools/bin/ci_integration_workflow_launcher.py base-normalization source-acceptance-test source:alpha source:beta source:GA destination:alpha destination:beta destination:GA
         env:
           GITHUB_TOKEN: ${{ secrets.SLASH_COMMAND_PAT }}
+  launch_integration_tests_alpha_only:
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
+    steps:
+      - name: Checkout Airbyte
+        uses: actions/checkout@v3
+      - name: Install Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: "zulu"
+          java-version: "17"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install PyYAML requests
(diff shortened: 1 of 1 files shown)


.github/workflows/connector_integration_tests.yml
```diff
@@ -29,3 +29,22 @@ jobs:
         run: python ./tools/bin/ci_integration_workflow_launcher.py base-normalization source-acceptance-test source:alpha source:beta source:GA destination:alpha destination:beta destination:GA
```
erohmensing, line 29: Hm, the alpha connectors are still in here, maybe you need to merge master in to this branch? Just to validate

.github/workflows/connector_integration_tests.yml
```diff
@@ -29,3 +29,22 @@ jobs:
...
+      - name: Install Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: "zulu"
+          java-version: "17"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install PyYAML requests
+      - name: Launch Integration Tests
```
erohmensing, line 47:
```suggestion
      - name: Launch Integration Tests (Alpha connectors)
```
Nit: titling the "Launch Integration Tests" job steps for the jobs more clearly

erohmensing (approved):
I also like having them as 2 separate jobs, in case we end up wanting to point them to different places. Wondering what led us to pick 30 minutes as a timeout for the initial job, and should we have some sort of timeout for this one, even if a longer one?

alafanechere:
> Wondering what led us to pick 30 minutes as a timeout for the initial job.
I don't know exactly but I think it's to have a kind of guarantee that all the test run in 30mn to build the report that is sent on slack afterward in a different action.

> And should we have some sort of timeout for this one, even if a longer one?
I hope the default 360mn is good enough.
The scripts of this workflow triggers /test workflow in parallel, I hope that we can build all our  alpha connectors in parallel in less than 6hours...
I'll figure out tomorrow 😄

PR #21594, opened 2023-01-19
🪟 🎉 Adds extra information while deleting source or destination

### LINKED ISSUE #19237: Show affected connections/sources/destinations on delete
A [p0](https://airbytehq-team.slack.com/archives/C03AS1GAQV6/p1668014071550469) popped up where a user deleted a very important destination.
A helpful way to mitigate that would be to show affected sources/destinations/connections before allowing the delete.

YatsukBogdan1:
## What
Closes #19237 

## How
- I added a new `ConfirmationModal` property called `extra` which allows us to add additional information needed by the user.
- I decided to put additional info only while deleting the source or destination since when deleting the connection seems like the user should know what is the result of his actions.
- Also for now there is no styling for extra info content, but maybe we need somehow to highlight this information, it is the design question from this point.

## Screenshots
Screenshot 2023-01-19 at 16 03 56
Screenshot 2023-01-19 at 16 03 43


edmundito:
cc @andyjih for copy changes

airbyte-webapp/src/pages/destination/DestinationSettingsPage/DestinationSettingsPage.tsx
```diff
@@ -57,7 +69,7 @@ export const DestinationSettingsPage: React.FC = () => {
         connector={destination}
         onSubmit={onSubmitForm}
       />
-      <DeleteBlock type="destination" onDelete={onDelete} />
+      <DeleteBlock extraModal={extraModal} type="destination" onDelete={onDelete} />
```
dizel852, line 72: Hm... Can we replace `extraModal` and `extra` prop name with something more descriptive, like "additionalText", or "additionalInfo"?

airbyte-webapp/src/components/common/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -15,12 +15,14 @@ export interface ConfirmationModalProps {
...
   submitButtonDataId?: string;
   cancelButtonText?: string;
+  extra?: React.ReactNode;
 }
 
 export const ConfirmationModal: React.FC<ConfirmationModalProps> = ({
   onClose,
   title,
   text,
+  extra = null,
```
dizel852, line 25: Just curious, why do we set `null` as a default value?

dizel852 (changes requested):
Did a review - left some comments.
Also, tests locally - seems like work as expected. 

@edmundito I've not found any figma  design for such a case. Probably would be better to highlight the affected connection names(or at least make them `bold`) WDYT?

airbyte-webapp/src/locales/en.json
```diff
@@ -485,6 +485,8 @@
   "tables.connectionDeleteConfirm": "Confirm connection deletion",
   "tables.sourceDeleteModalText": "Deleting a source cannot be undone without a full re-sync. Note that:\n  - All connections with this source will be deleted, including their past logs and configurations.\n  - No existing data in the destination will be altered",
   "tables.destinationDeleteModalText": "Deleting a destination cannot be undone. Note that:\n  - All connections with this destination will be deleted, including their past logs and configurations.\n  - No existing data in the destination will be altered.",
+  "tables.affectedConnectionsOnSourceDeletion": "This connections will be affected if you delete this source:\n",
+  "tables.affectedConnectionsOnDestinationDeletion": "This connections will be affected if you delete this destination:\n",
```
edmundito, line 489: It should be "The following connection will be deleted" instead of "This connections will be affected"

airbyte-webapp/src/components/common/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -15,12 +15,14 @@ export interface ConfirmationModalProps {
   onSubmit: () => void;
   submitButtonDataId?: string;
   cancelButtonText?: string;
+  extra?: React.ReactNode;
```
edmundito, line 18: `additionalContent` would be a better name.

airbyte-webapp/src/components/common/ConfirmationModal/ConfirmationModal.tsx
edmundito: This may not be necessary

airbyte-webapp/src/pages/destination/DestinationSettingsPage/DestinationSettingsPage.tsx
edmundito: I suggested `addtionalContent` above

edmundito:
@dizel852 There is no specific design for this. In this case, we expect the developer to use their best judgment.

airbyte-webapp/src/pages/destination/DestinationSettingsPage/DestinationSettingsPage.tsx
krishnaglick: Could this be `footer`?

edmundito: The modal's footer is where the button go

airbyte-webapp/src/locales/en.json
andyjih: Do we have a library for pluralizing? Ideally it would be
1 connection: "The following connection..."
2+ connections: "The following connections..."

andyjih (dismissed):
Copy feedback. Thanks!

airbyte-webapp/src/locales/en.json
edmundito: Yes, the i18n library can pluralize

airbyte-webapp/src/components/common/ConfirmationModal/ConfirmationModal.tsx
krishnaglick: I think `undefined` will render out the same as `null`

airbyte-webapp/src/pages/destination/DestinationSettingsPage/DestinationSettingsPage.tsx
krishnaglick: +1 to `addtionalContent`

YatsukBogdan1:
Next updates were made:
- Added pluralization for connection message
- Changed `extra` to `additionalContent`
- Added `bold` for connection names to make this information recognizable
Screenshot 2023-01-25 at 14 59 06
Screenshot 2023-01-25 at 14 59 23

dizel852 (approved):
LGTM 👍 

Passing to @edmundito for final review

airbyte-webapp/src/pages/SourcesPage/pages/SourceItemPage/components/SourceSettings.tsx
```diff
@@ -54,6 +54,20 @@ const SourceSettings: React.FC<SourceSettingsProps> = ({ currentSource, connecti
...
 
+  const modalAdditionalContent = useMemo<React.ReactNode>(() => {
+    if (connectionsWithSource.length === 0) {
+      return null;
+    }
+    return (
+      <p>
+        <FormattedMessage id="tables.affectedConnectionsOnDeletion" values={{ count: connectionsWithSource.length }} />
+        {connectionsWithSource.map((connection) => (
+          <b>- {`${connection.name}\n`}</b>
```
edmundito, line 65:
```suggestion
          - <strong>{`${connection.name}\n`}</strong>
```

airbyte-webapp/src/pages/destination/DestinationSettingsPage/DestinationSettingsPage.tsx
```diff
@@ -44,6 +44,23 @@ export const DestinationSettingsPage: React.FC = () => {
...
+      return null;
+    }
+    return (
+      <p>
+        <FormattedMessage
+          id="tables.affectedConnectionsOnDeletion"
+          values={{ count: connectionsWithDestination.length }}
+        />
+        {connectionsWithDestination.map((connection) => (
+          <b>- {`${connection.name}\n`}</b>
```
edmundito, line 58:
```suggestion
          - <strong>{`${connection.name}\n`}</strong>
```

edmundito (approved):
Tested locally and looks good, just had a couple of suggestions before merge.

airbyte-webapp/src/locales/en.json
```diff
@@ -486,6 +486,7 @@
   "tables.connectionDeleteConfirm": "Confirm connection deletion",
   "tables.sourceDeleteModalText": "Deleting a source cannot be undone without a full re-sync. Note that:\n  - All connections with this source will be deleted, including their past logs and configurations.\n  - No existing data in the destination will be altered",
   "tables.destinationDeleteModalText": "Deleting a destination cannot be undone. Note that:\n  - All connections with this destination will be deleted, including their past logs and configurations.\n  - No existing data in the destination will be altered.",
```
edmundito, line 488: One more thing 😅 : I noticed that while just deleting sources or destinations, the message is now confusing because no connections will be deleted if they don't appear. We can simplify this text to:

```
"Deleting a source cannot be undone without a full re-sync. No existing data in the destination will be altered."

"Deleting a destination cannot be undone. No existing data in the destination will be altered."
```

This way the user only worries about connections being deleted when there are any.

edmundito (approved):

PR #21602, opened 2023-01-19
Source Instagram: test strictness level high

davydov-d:
## What
Set `high` test strictness level of Instagram source 

## How
- add expected records
- fix minor bug
- refactor test config file
- add abnormal state
- add incremental catalog


lazebnyi (approved):

darynaishchenko (approved):

evantahler (approved):
Assuming the business account ID is a sandbox account, 👍!

PR #21604, opened 2023-01-19
Source S3: keep processing but warn if OSError happen

roman-yermilov-gl:
## What
*Handle OSError which means no access to key in a bucket. Warn and keep working on another files which we have an access to*

### DIFF
--- a/airbyte-integrations/connectors/source-s3/source_s3/s3file.py
+++ b/airbyte-integrations/connectors/source-s3/source_s3/s3file.py
@@ -65,8 +65,10 @@ class S3File(StorageFile):
         try:
             result = smart_open.open(f"s3://{bucket}/{self.url}", transport_params=params, mode=mode)
         except OSError as e:
-            self.logger.warn(f"We don't have access to {self.url}. "
-                             f"Check whether key {self.url} exists in `{bucket}` bucket and/or has proper ACL permissions")
+            self.logger.warn(
+                f"We don't have access to {self.url}. "
+                f"Check whether key {self.url} exists in `{bucket}` bucket and/or has proper ACL permissions"
+            )
             raise e
         # see https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager for why we do this
         try:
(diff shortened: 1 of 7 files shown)


airbyte-integrations/connectors/source-s3/source_s3/s3file.py
```diff
@@ -59,8 +59,15 @@ def open(self, binary: bool) -> Iterator[Union[TextIO, BinaryIO]]:
...
-
+        # There are rare cases when some keys become unreachable during sync
+        # and we don't know about it, because catalog has been initially formed only once at the beginning
+        # This is happen for example if a file was deleted/moved (or anything else) while we proceed with another file
+        try:
+            result = smart_open.open(f"s3://{bucket}/{self.url}", transport_params=params, mode=mode)
+        except OSError as e:
+            self.logger.warn(f"We don't have access to {self.url}. "
+                             f"Check whether key {self.url} exists in `{bucket}` bucket and/or has proper ACL permissions")
+            raise e
```
evantahler, line 70: Does this produce an AirbyteTraceError message?  Just logging to stdout won't help users learn what to do

lazebnyi (approved):

airbyte-integrations/connectors/source-s3/source_s3/s3file.py
evantahler: cc @brianjlai maybe you recall the answer - does throwing an exception here end up making a useable AirbyteTraceErrorMessage?

darynaishchenko (approved):

airbyte-integrations/connectors/source-s3/source_s3/s3file.py
roman-yermilov-gl: It does. But in this case I catch it in two possible places where it can be raised and handle it

evantahler: Ok, this matches the philosophy that @erohmensing is working on with streams being unavailable - we sync what we can and log an error for the streams we can't reach.  👍

evantahler (approved):

Issue #3078, opened 2021-04-27
CDK: add flag to turn on debugging mode


sherifnada:
## Tell us about the problem you're trying to solve
It would be nice to have a flag `--debug` which outputs a bunch of extra logs like: 

* what the stream slices are
* which URL endpoints/params/etc.. are being hit
* etc. basically various information that can give me visibility about what's happening in the connector

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1200367912513076/1200368227754448) by [Unito](https://www.unito.io)

girarda:
few additional things that would be useful to expose for troubleshooting:
* raw response (need to make sure the secrets are scrubbed)
* fully expanded config (for low-code)
* cursor value

PR #21605, opened 2023-01-19
[ISSUE #20771] adding slices to connector builder read request

maxi297:
## What
In order to validate if the number of requests was limited, we want to be able to know how many pages and slices were queries. We already had the information for the pages but didn't for the slices. This PR addresses the slices part.

## How
Emit a log message when a new slice is being started and catch this log message to create the slices properly

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py`: how the slice message is logged
2. `airbyte-connector-builder-server/connector_builder/impl/default_api.py`: how we use this log message to create the slices on the connector-builder server

## 🚨 User Impact 🚨
@flash1293, will this be supported by the UI? Although not a breaking API change, there were some assumptions in the backend that there will always be only one slice.

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
@@ -237,7 +237,6 @@ class AbstractSource(Source, ABC):
             has_slices = True
             if logger.isEnabledFor(logging.DEBUG):
                 yield AirbyteMessage(type=MessageType.LOG, log=AirbyteLogMessage(level=Level.INFO, message=f"slice:{json.dumps(_slice)}"))
-            logger.debug("Processing stream slice", extra={"slice": _slice})
             records = stream_instance.read_records(
                 sync_mode=SyncMode.incremental,
                 stream_slice=_slice,
@@ -288,7 +287,6 @@ class AbstractSource(Source, ABC):
         for _slice in slices:
             if logger.isEnabledFor(logging.DEBUG):
                 yield AirbyteMessage(type=MessageType.LOG, log=AirbyteLogMessage(level=Level.INFO, message=f"slice:{json.dumps(_slice)}"))
-            logger.debug("Processing stream slice", extra={"slice": _slice})
             record_data_or_messages = stream_instance.read_records(
                 stream_slice=_slice,
                 sync_mode=SyncMode.full_refresh,
--- a/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
@@ -6,6 +6,7 @@ import json
 import logging
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Iterator, List, Mapping, MutableMapping, Optional, Tuple, Union
+
 from airbyte_cdk.models import (
     AirbyteCatalog,
     AirbyteConnectionStatus,
--- a/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
@@ -2,17 +2,19 @@
 # Copyright (c) 2022 Airbyte, Inc., all rights reserved.
 #
 
+import json
 import logging
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Iterator, List, Mapping, MutableMapping, Optional, Tuple, Union
-
 from airbyte_cdk.models import (
     AirbyteCatalog,
     AirbyteConnectionStatus,
+    AirbyteLogMessage,
     AirbyteMessage,
     AirbyteStateMessage,
     ConfiguredAirbyteCatalog,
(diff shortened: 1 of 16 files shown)


maxi297:
There is no test that validates that `ManifestDeclarativeSource` will configure the logger to debug. Is this something we should have?

airbyte-connector-builder-server/connector_builder/impl/default_api.py
```diff
@@ -190,7 +198,8 @@ def _get_message_groups(self, messages: Iterator[AirbyteMessage], schema_inferre
         else:
             if not current_page_request or not current_page_response:
                 raise ValueError("Every message grouping should have at least one request and response")
-            yield StreamReadPages(request=current_page_request, response=current_page_response, records=current_records)
+            current_slice_pages.append(StreamReadPages(request=current_page_request, response=current_page_response, records=current_records))
+            yield StreamReadSlices(pages=current_slice_pages)
```
girarda, line 202: should we also set the slice_descriptor and the state now that we're partitioning by slices?

@lmossman there aren't any short-term plans to use these fields, right?

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -232,6 +235,8 @@ def _read_incremental(
         has_slices = False
         for _slice in slices:
             has_slices = True
+            if logger.isEnabledFor(logging.DEBUG):
+                yield AirbyteMessage(type=MessageType.LOG, log=AirbyteLogMessage(level=Level.INFO, message=f"slice:{json.dumps(_slice)}"))
             logger.debug("Processing stream slice", extra={"slice": _slice})
```
girarda, line 240: open question: would it make sense to not using the logger here and only yield a the log message? I doubt there's a lot of value in logging both when running a read with the `--debug` flag
eg
```
{"type": "LOG", "log": {"level": "INFO", "message": "slice:{\"created_after\": \"1970-01-01T00:00:00.000Z\"}"}}
{"type": "DEBUG", "message": "Processing stream slice", "data": {"slice": "{'created_after': '1970-01-01T00:00:00.000Z'}"}}
```

girarda (approved):
couple of questions that maybe span outside of the scope of the PR. lgtm 🚢

airbyte-connector-builder-server/connector_builder/impl/default_api.py
lmossman: We already have a dropdown implemented for selecting the slice implemented which will be rendered in the UI if there is more than one slice being returned, so you should be able to see what that looks like on this PR branch. That dropdown uses the slice descriptor to decide what values to show in the dropdown.

But I don't believe we currently have any plans for `state`

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
maxi297: I had the exact same question. In the end, my conclusion was "I don't have the knowledge to know what was the use of this log". If it is operational, we would need to migrate the tooling associated to it. If we are confident it's not used, it will be my pleasure to remove it

girarda: the debug flag is only used when manually troubleshooting commands (eg `python main.py read --config secrets/config.json --catalog integration_tests/configured_catalog.json --debug`)

Here's the [issue](https://github.com/airbytehq/airbyte/issues/3078) and it's corresponding [PR](https://github.com/airbytehq/airbyte/issues/3078) for extra context.

yielding an airbyte log message will be functionally equivalent since those messages will also show up in the terminal.

airbyte-connector-builder-server/connector_builder/impl/default_api.py
maxi297: I am still unclear as to how we should populate the `slice_descriptor` but I think this can be part of another issue if/when needed

flash1293: Agreed, it doesn't look too bad to me

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
maxi297: Perfect! I'll remove the `logger.debug` lines then. Thanks!

airbyte-connector-builder-server/connector_builder/impl/default_api.py
lmossman: A couple ideas for populating the slice_descriptor:
- If using a datetime stream slicer, it could be set to the starting datetime of the current slice, e.g. `1 Jan 2023`
- If using a list stream slicer, it could be set to the current list item, e.g. `comments`
- If using a cartesian product of datetime and list, it could be a concatenation of both, e.g. `1 Jan 2023 | comments`

Not as sure about substream, maybe it could be set to the parent slice key or something like that?

PR #21608, opened 2023-01-19
Remove source and destination definition calls that were used only to get the icon in Streams Field Table

### LINKED ISSUE #21622: Remove source and destination definition calls from Streams Field Table
In the Streams Field Table component we make a call for both the source and destination definition only to get the icon. This is available on source and destination objects already and should be used from them instead.

krishnaglick:
Resolves https://github.com/airbytehq/airbyte/issues/21622

## What
Remove source and destination definition calls that were used only to get the icon.

## How
Reference the icon on the Source and Destination directly instead of their definition


dizel852 (approved):
LGTM
Tested locally - no issues 👍

PR #21619, opened 2023-01-19
[ISSUE #20771] set flag when limit requests reached

maxi297:
## What
Now that we know how many pages and slices were queries, we can set the flag in the response so that the UI can display it.

## How
Count the number of slices and the number of pages per slices. If any of these exceeds the limit, set flag to true.

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py`: how the slice message is logged
2. `airbyte-connector-builder-server/connector_builder/impl/default_api.py`: how we use this log message to create the slices on the connector-builder server

## 🚨 User Impact 🚨
@flash1293, I wouldn't mind trying to add this to the UI but I'm not sure what you had in mind in terms of UX (where to display the warning, content of the warning message, etc...) Do you mind if we sync on this?

### DIFF
--- a/airbyte-cdk/python/unit_tests/sources/test_source.py
+++ b/airbyte-cdk/python/unit_tests/sources/test_source.py
@@ -390,6 +390,7 @@ def test_internal_config_limit(abstract_source, catalog):
     logger_mock.level = logging.DEBUG
     del catalog.streams[1]
     STREAM_LIMIT = 2
+    SLICE_DEBUG_LOG_COUNT = 1
     FULL_RECORDS_NUMBER = 3
     streams = abstract_source.streams(None)
     http_stream = streams[0]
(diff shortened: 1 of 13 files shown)


airbyte-cdk/python/unit_tests/sources/test_source.py
```diff
@@ -390,6 +390,7 @@ def test_internal_config_limit(abstract_source, catalog):
     logger_mock.level = logging.DEBUG
     del catalog.streams[1]
     STREAM_LIMIT = 2
+    SLICE_DEBUG_LOG_COUNT = 1
```
maxi297, line 393: These tests should have been fixed in https://github.com/airbytehq/airbyte/pull/21525 but I'm doing this here for simplicity's sake

airbyte-connector-builder-server/connector_builder/impl/default_api.py
```diff
@@ -237,7 +249,7 @@ def _create_response_from_log_message(self, log_message: AirbyteLogMessage) -> O
 
     def _create_low_code_adapter(self, manifest: Dict[str, Any]) -> CdkAdapter:
         try:
-            return self.adapter_cls(manifest=manifest)
+            return self.adapter_cls(manifest, self._MAXIMUM_NUMBER_OF_PAGES_PER_SLICE, self._MAXIMUM_NUMBER_OF_SLICES)
```
maxi297, line 252: It's a little weird to call the __init__ method of the `adapter_cls` when it is not in the API. It feels to me that there might be a factory missing but I've seen very few factories in the code so I'm curious about how the team feels about that

girarda: I agree that a factory would make this easier to maintain

girarda (approved):

PR #21621, opened 2023-01-19
add @Ignore to test that can be unpredicatable

colesnodgrass:
## What
- skip test that can fail due to timing issues with watching for file updates
- this has bit a few people already, so disabling this test until it can be relocated to a non-unit-test suite

## How
- `@Ignore` test that is unpredictable


benmoriceau (approved):

jdpgrailsdev (approved):
:shipit:

PR #21623, opened 2023-01-19
:window: :handshake: Free connectors program confirmation UI

### LINKED ISSUE #21616: Show confirmation upon successfull enrollment in Free Connector Program and verification email re-send
Part of https://github.com/airbytehq/airbyte/issues/21074. The stripe `successUrl` appends a specific query param to the current URL, so the enrollment banner components which serve as the enrollment entry points (and are therefore always part of the `successUrl`'s UI) can check for that query param on page load; if present, they will display a success `Toast` and avoid rendering the banner UI.

Similarly, there should be a `Toast` triggered when the request to firebase to re-send the address validation email returns successfully.

ambirdsall:
## What
Suppress enrollment banners and show confirmation `Toast` when user returns to the `successUrl` from an enrollment stripe checkout session; dismiss modal and show confirmation `Toast` when a user requests a verification email resend from the enrollment modal. Closes https://github.com/airbytehq/airbyte/issues/21616.

## screen recordings
Enrollment success. In retrospect, I should have started recording the gif before opening the modal, to make the difference more apparent when the enrollment banner is hidden. Ah well, c'est la vie. Gifs are hidden behind `<details>` tags because they are large and distracting:
<details>
fcp-enrollment-success
</details>

Email validation resend requested
<details>
fcp-email-validation-confirmation
</details>

## How
Pretty straightforward implementation. One key but nonobvious detail: this changes the query param appended to the stripe `successUrl`, in order to distinguish between a Free Connector Program enrollment and a one-off transaction on the credits page.

## Open question about a future UX improvement
Should the email verification popup mention that the user has to refresh the page after confirming their email? Simply retriggering the modal won't reflect any changes that may have happened if the banner that triggers it hasn't rerendered, because the current modal implementation cannot directly use the required hooks (it renders outside of the necessary context provider components).

Options:
- tell users to refresh the page in the popup
- start polling email validation status from `AuthService` in the banner components as soon as the user requests email validation
- other??

### DIFF
--- a/airbyte-webapp/src/components/ui/Toast/Toast.module.scss
+++ b/airbyte-webapp/src/components/ui/Toast/Toast.module.scss
@@ -43,6 +43,7 @@ $toast-bottom-margin: 27px;
   position: fixed;
   box-sizing: border-box;
   bottom: $toast-bottom-margin;
+  margin-left: calc(vars.$width-size-menu / 2);
   left: 50%;
   transform: translate(-50%, 0);
   z-index: z-indices.$notification;
--- a/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/EnrollmentModal/useShowEnrollmentModal.tsx
+++ b/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/EnrollmentModal/useShowEnrollmentModal.tsx
@@ -1,4 +1,9 @@
+import { useIntl } from "react-intl";
+
+import { ToastType } from "components/ui/Toast";
+
 import { useModalService } from "hooks/services/Modal";
+import { useNotificationService } from "hooks/services/Notification";
 import { useAuthService } from "packages/cloud/services/auth/AuthService";
 import { useStripeCheckout } from "packages/cloud/services/stripe/StripeService";
 import { useCurrentWorkspaceId } from "services/workspaces/WorkspacesService";
@@ -10,6 +15,18 @@ export const useShowEnrollmentModal = () => {
   const { mutateAsync: createCheckout } = useStripeCheckout();
   const workspaceId = useCurrentWorkspaceId();
   const { emailVerified, sendEmailVerification } = useAuthService();
+  const { formatMessage } = useIntl();
+  const { registerNotification } = useNotificationService();
+
+  const verifyEmail = () =>
+    sendEmailVerification().then(() => {
+      registerNotification({
+        id: "fcp/verify-email",
+        text: formatMessage({ id: "freeConnectorProgram.enrollmentModal.validationEmailConfirmation" }),
+        type: ToastType.INFO,
+      });
+      closeModal();
+    });
 
   return {
     showEnrollmentModal: () => {
@@ -21,7 +38,7 @@ export const useShowEnrollmentModal = () => {
             createCheckout={createCheckout}
             closeModal={closeModal}
             emailVerified={emailVerified}
-            sendEmailVerification={sendEmailVerification}
+            sendEmailVerification={verifyEmail}
           />
         ),
       });
--- a/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/EnrollmentModal/useShowEnrollmentModal.tsx
+++ b/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/EnrollmentModal/useShowEnrollmentModal.tsx
@@ -19,14 +19,15 @@ export const useShowEnrollmentModal = () => {
   const { registerNotification } = useNotificationService();
 
   const verifyEmail = () =>
-    sendEmailVerification().then(() => {
-      registerNotification({
-        id: "fcp/verify-email",
-        text: formatMessage({ id: "freeConnectorProgram.enrollmentModal.validationEmailConfirmation" }),
-        type: ToastType.INFO,
-      });
-      closeModal();
-    });
+    sendEmailVerification()
+      .then(() => {
+        registerNotification({
+          id: "fcp/verify-email",
+          text: formatMessage({ id: "freeConnectorProgram.enrollmentModal.validationEmailConfirmation" }),
+          type: ToastType.INFO,
+        });
+      })
+      .catch(); // don't crash the page on error
 
   return {
     showEnrollmentModal: () => {
--- a/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/InlineEnrollmentCallout.tsx
+++ b/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/InlineEnrollmentCallout.tsx
@@ -5,6 +5,7 @@ import { Callout } from "components/ui/Callout";
 import { Text } from "components/ui/Text";
 
 import { useShowEnrollmentModal } from "./EnrollmentModal";
+import { useFreeConnectorProgram } from "./hooks/useFreeConnectorProgram";
 import styles from "./InlineEnrollmentCallout.module.scss";
 
 export const EnrollLink: React.FC<PropsWithChildren<unknown>> = ({ children }) => {
@@ -17,7 +18,12 @@ export const EnrollLink: React.FC<PropsWithChildren<unknown>> = ({ children }) =
   );
 };
 export const InlineEnrollmentCallout: React.FC = () => {
-  return (
+  const { userDidEnroll } = useFreeConnectorProgram();
+
+  return userDidEnroll ? (
+    // eslint-disable-next-line react/jsx-no-useless-fragment
+    <></>
+  ) : (
     <Callout variant="info" className={styles.container}>
       <Text size="sm">
         <FormattedMessage
--- a/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/InlineEnrollmentCallout.tsx
+++ b/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/InlineEnrollmentCallout.tsx
@@ -20,10 +20,7 @@ export const EnrollLink: React.FC<PropsWithChildren<unknown>> = ({ children }) =
 export const InlineEnrollmentCallout: React.FC = () => {
   const { userDidEnroll } = useFreeConnectorProgram();
 
-  return userDidEnroll ? (
-    // eslint-disable-next-line react/jsx-no-useless-fragment
-    <></>
-  ) : (
+  return userDidEnroll ? null : (
     <Callout variant="info" className={styles.container}>
       <Text size="sm">
         <FormattedMessage
--- a/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/LargeEnrollmentCallout.tsx
+++ b/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/LargeEnrollmentCallout.tsx
@@ -16,10 +16,7 @@ export const LargeEnrollmentCallout: React.FC = () => {
   const { showEnrollmentModal } = useShowEnrollmentModal();
   const { userDidEnroll } = useFreeConnectorProgram();
 
-  return userDidEnroll ? (
-    // eslint-disable-next-line react/jsx-no-useless-fragment
-    <></>
-  ) : (
+  return userDidEnroll ? null : (
     <Callout variant="boldInfo" className={styles.container}>
       <FlexContainer direction="row" alignItems="center" className={styles.flexRow}>
         <FlexItem grow={false} alignSelf="center">
--- a/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/LargeEnrollmentCallout.tsx
+++ b/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/LargeEnrollmentCallout.tsx
@@ -9,12 +9,17 @@ import { Text } from "components/ui/Text";
 
 import { ReactComponent as ConnectorsBadges } from "./connectors-badges.svg";
 import { useShowEnrollmentModal } from "./EnrollmentModal";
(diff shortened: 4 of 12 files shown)


airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/LargeEnrollmentCallout.tsx
```diff
@@ -9,12 +9,17 @@ import { Text } from "components/ui/Text";
...
+import { useFreeConnectorProgram } from "./hooks/useFreeConnectorProgram";
 import styles from "./LargeEnrollmentCallout.module.scss";
 
 export const LargeEnrollmentCallout: React.FC = () => {
   const { showEnrollmentModal } = useShowEnrollmentModal();
+  const { userDidEnroll } = useFreeConnectorProgram();
 
-  return (
+  return userDidEnroll ? (
+    // eslint-disable-next-line react/jsx-no-useless-fragment
```
ambirdsall, line 20: If the whole UI subtree should be rendered or not based on some condition, I think keeping the conditional logic outside the UI it toggles, like this:
```tsx
return someCond ? (
  <></>
) : (
  <TheActualComponent>
    <And>
      <ItsChildren />
    </And>
  </TheActualComponent>
);
```
is _much_ nicer than sticking it inside the nominal UI it's trying to toggle:
```tsx
return (
  <>
    {someCond ? null : (
      <TheActualComponent>
        <And>
          <ItsChildren />
        </And>
      </TheActualComponent>
    )}
  </>
)
```

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/InlineEnrollmentCallout.tsx
```diff
@@ -17,7 +18,12 @@ export const EnrollLink: React.FC<PropsWithChildren<unknown>> = ({ children }) =
   );
 };
 export const InlineEnrollmentCallout: React.FC = () => {
-  return (
+  const { userDidEnroll } = useFreeConnectorProgram();
+
+  return userDidEnroll ? (
+    // eslint-disable-next-line react/jsx-no-useless-fragment
+    <></>
+  ) : (
```
josephkmh, line 26:
```suggestion
  return userDidEnroll ? null :
```

joeykmh (changes requested):
Just a couple small comments. I think we should find a way to re-check whether a user's email has been verified when the window regains focus. That would make the UX a lot smoother. But I think that will require a refactor of the `AuthService`, so we can make that a separate PR.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/EnrollmentModal/useShowEnrollmentModal.tsx
```diff
@@ -10,6 +15,18 @@ export const useShowEnrollmentModal = () => {
...
+  const { registerNotification } = useNotificationService();
+
+  const verifyEmail = () =>
+    sendEmailVerification().then(() => {
+      registerNotification({
+        id: "fcp/verify-email",
+        text: formatMessage({ id: "freeConnectorProgram.enrollmentModal.validationEmailConfirmation" }),
+        type: ToastType.INFO,
+      });
+      closeModal();
```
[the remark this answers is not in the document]
ambirdsall: > I think closing the modal is kind of weird UX

Agreed. Unfortunately, because the modal is at a higher z-index than the toast, leaving it open means the only way to confirm the action is rendering something inline in the modal, which also seemed weird.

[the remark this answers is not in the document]
timroes: I think we can switch those z-indexes around. In general I'd assume an incoming notification should be higher even than an open modal?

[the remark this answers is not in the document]
timroes: And for posterity (talked offline with Joey about that): We should be fine reordering the providers, but we might need to check which orders work since there are "some" dependencies between them.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/InlineEnrollmentCallout.tsx
ambirdsall: huh, somehow assumed that would change the type; I've run into type issues trying to pass `null` to code that expects jsx, and inappropriately generalized the learning.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/EnrollmentModal/useShowEnrollmentModal.tsx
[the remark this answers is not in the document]
ambirdsall: :+1: to both. To keep this particular PR's diff focused, though, I'll leave the providers reastructuring and clean up associated cleanup for a follow-up PR.

ambirdsall:
Updates since the last round of reviews:
- while keeping the shared error handling in `sendEmailVerification`, also reraise caught errors. Client code (all two files of it lol) now need to use `.catch()` if they don't want to risk crashing the page, but it means a chained `.then` callback will only be run on success instead of being a de facto `.finally`
- updated the toast z-index to be higher than the modal's. Also, for readability, sorted the z-index variables in order of, well, z-index.
- centered the toast popup vis a vis the modal.

Before tweaking the toast alignment:
2023-01-20-12:22:36-screenshot
2023-01-20-12:22:48-screenshot

and after:
2023-01-20-12:51:16-screenshot
2023-01-20-12:51:29-screenshot

airbyte-webapp/src/components/ui/Toast/Toast.module.scss
```diff
@@ -43,6 +43,7 @@ $toast-bottom-margin: 27px;
   position: fixed;
   box-sizing: border-box;
   bottom: $toast-bottom-margin;
+  margin-left: calc(vars.$width-size-menu / 2);
```
tealjulia, line 46: This variable name doesn't make much sense to me as we aren't using it anywhere for a menu, and I can't tell why a menu's width would be a relevant benchmark for this.  

Is there a meaningful reason to use 2 * width-size-menu vs. the number itself?

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/EnrollmentModal/useShowEnrollmentModal.tsx
```diff
@@ -10,6 +15,19 @@ export const useShowEnrollmentModal = () => {
...
+  const verifyEmail = () =>
+    sendEmailVerification()
+      .then(() => {
+        registerNotification({
+          id: "fcp/verify-email",
+          text: formatMessage({ id: "freeConnectorProgram.enrollmentModal.validationEmailConfirmation" }),
+          type: ToastType.INFO,
+        });
+      })
+      .catch(); // don't crash the page on error
```
tealjulia, line 30: If we're housing the email verification logic here rather than within the modal component itself, does it make sense to also move `startStripeCheckout()` here from within the modal?  They feel parallel-ish to me.

Edit: actually, it looks like we're doing more there with refs and redirects and that may not be as simple a change as it first looked.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/hooks/useFreeConnectorProgram.ts
```diff
@@ -1,21 +1,47 @@
...
+  const [userDidEnroll, setUserDidEnroll] = useState(false);
+  const { formatMessage } = useIntl();
+  const { registerNotification } = useNotificationService();
 
-  return useQuery(["freeConnectorProgramInfo", workspaceId], () =>
+  useEffectOnce(() => {
+    if (searchParams.has(STRIPE_SUCCESS_QUERY)) {
+      // Remove the stripe parameter from the URL
+      setSearchParams({}, { replace: true });
+      setUserDidEnroll(true);
```
tealjulia, line 35: `userDidEnroll` is essentially a temporary place for us to know that a user has enrolled, but we haven't done a full re-query of the backend yet so we can show the Toast and remove the enrollment materials, right?

airbyte-webapp/src/pages/connections/ConnectionPage/ConnectionPageTitle.tsx
```diff
@@ -26,7 +26,8 @@ export const ConnectionPageTitle: React.FC = () => {
 
   const { connection } = useConnectionEditService();
 
-  const { data: freeConnectorProgramInfo } = useFreeConnectorProgram();
+  const { enrollmentStatusQuery } = useFreeConnectorProgram();
+  const { data: freeConnectorProgramInfo } = enrollmentStatusQuery;
   const displayEnrollmentCallout = freeConnectorProgramInfo?.showEnrollmentUi;
```
tealjulia, line 31: can this be simplified to two lines instead of three by doing 

```suggestion
  const { enrollmentStatusQuery } = useFreeConnectorProgram();
  const displayEnrollmentCallout = enrollmentStatusQuery?.data?.freeConnectorProgramInfo?.showEnrollmentUi;
```
or something of the like? Just seems redundant as is with the intermediate steps there.

airbyte-webapp/src/scss/_z-indices.scss
```diff
@@ -1,11 +1,11 @@
...
-$panelSplitter: 0;
-$dropdownMenu: 2;
-$notification: 20;
-$schemaChangesBackdrop: 3;
 $schemaChangesBackdropContent: 4;
+$schemaChangesBackdrop: 3;
+$dropdownMenu: 2;
 $switchSliderBefore: 1;
 $tableScroll: 1;
+$panelSplitter: 0;
```
tealjulia, line 11: 🙏

teallarson (commented):
I'll defer to Joey on a review here!  I mostly just had questions.

airbyte-webapp/src/components/ui/Toast/Toast.module.scss
ambirdsall: Agreed, it's a weird naming choice. Here, I used that variable purely because the modal does, to keep the two nicely-aligned vis a vis each other.

Zooming out a little, it's not used very many places:
2023-01-20-13:31:07-screenshot
I think it's a fine candidate to be pulled out or renamed or something, I just didn't want to have to verify the correctness of `components/connection/CatalogTree/next/StreamDetailsPanel` in order to ship the free connectors program :grimacing:

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/hooks/useFreeConnectorProgram.ts
ambirdsall: "Almost right, but the situation is slightly worse than that. After a user successfully finishes the stripe checkout flow: 
1) Stripe redirects the user directly to the `successUrl` their checkout session was configured with (in this case, the airbyte URL the user came from with `STRIPE_SUCCESS_QUERY` appended)
2) fires off a webhook to update our backend

That means that even if we do a full re-query of the backend right away (which in fact we do, since we're reloading the whole webapp), there's a race condition: we cannot rely on the backend to have already processed the enrollment at the moment we ought to show the user a confirmation.

airbyte-webapp/src/packages/cloud/services/auth/AuthService.tsx
```diff
@@ -273,6 +273,7 @@ export const AuthenticationProvider: React.FC<React.PropsWithChildren<unknown>>
                 type: ToastType.ERROR,
               });
           }
+          throw error;
```
josephkmh, line 276: I find this error handling hard to grok, but it works 🙂

joeykmh (approved):
I think the error handling here is kind of hard to reason about (throwing & catching without handling seems... unintuitive?) but it works for now, and we've got a deadline. So happy to merge it the way it is!

airbyte-webapp/src/packages/cloud/services/auth/AuthService.tsx
ambirdsall: I certainly did when I first pulled it out of `EmailVerificationHint.tsx` (it's basically unchanged otherwise), but at least it's structurally simple.

The explanation, I guess for posterity:
1) wrap the firebase call in a `try`/`catch`
1) match against a firebase-defined code (if you want to get fancy, it effectively makes the error type a tagged union) with `switch(error.code)`
2) for each `case` in that switch statement, match the error code against the `AuthErrorCodes` object imported from `"firebase/auth"` and give the contextually-appropriate notification

As Joey asked on Slack, a reasonable follow-up question is whether the success handling should also be centralized. Maybe! We'd just have to verify that this behavior also makes sense in the context of `EmailValidationHint` first, but I didn't want to hold up the free connectors work by going off on that sort of tangent.

josephkmh: My confusion was more about throwing the error again on line 276, only to have empty `catch()` statements in the consumer. I don't quite get why that's necessary?

ambirdsall: Ahh, that is a better question, thanks for clarifying. It was necessary (emphasis on "was") because I assumed the code I was extracting was structurally appropriate to the use case :facepalm: (a reasonable default assumption, given that the internal API defined in `authService` had an identical type signature to the `firebase/auth` function it was wrapping, but incorrect). It's [quite a simple fix](https://github.com/airbytehq/airbyte/pull/21763), though.

PR #21628, opened 2023-01-19
remove dependency declared twice

colesnodgrass:
## What
- remove dependency that was declared twice

## How
- delete key * ~20


colesnodgrass:
Found the same dependency duplicated in another place as well.  Which also forced some groovy code to reformat.

jdpgrailsdev (approved):
:shipit:

benmoriceau (approved):

Issue #20912, opened 2022-12-28
Refresh connector configuration between attempts / activities


pedroslopez:
## Tell us about the problem you're trying to solve

#19811 introduces the ability for connectors to update configuration throughout the sync in order to support single-use refresh tokens. However, configuration is currently loaded at the start of the sync and passed around between attempts via temporal. This means that even though the database has updated configuration, subsequent attempts will not have that data.

I also believe the same problem exists between CHECK / SYNC, if a check occurs before the sync that emits a control message, the sync that runs afterwards will not have the updated configuration.

This is particularly bad because:
* If there's a failure
* the next attempt will start with a check and get new tokens
* but those new tokens won't be available to the SYNC part of the attempt, and we keep failing/looping

## Describe the solution you’d like

Subsequent sync attempts should use the latest configuration. 
* don't pass configs around via temporal
* don't store job config statically in the jobs table
* the orchestrator really does load the config each attempt/job from the database.  This may be a new internal API as we need to hydrate the actor info with secrets.

evantahler:
Adding the Platform Workflow team, as this also falls into their wheelhouse.

evantahler:
I spoke with @davinchia, @colesnodgrass, and @jdpgrailsdev about this issue.  We all agree that there's a platform-level [problem](https://docs.google.com/document/d/1eCfNGsbJju0MwzcB7_MB9Xuao8PgNAyQhzuya7c8nrs/edit#heading=h.n4bpb53gyck9) we need to solve around the mutability of inputs between temporal activities & job attempts.  It is likely that as the platform transitions away from "sync-first" to "stream-first" syncing, this will become less of an issue... but that doesn't help us close this issue out in the near term.  Today, the 'sync' is the atomic operation (inclusive of all subsequent activities), rather than the 'attempt', which is why inputs changing between attempts is hard for the platform to handle.

I think there are 2 potential paths forward:
1. Re-hydrate config (and state) at the beginning of the replication activity
2. Throw a new "restart-the-sync-now" exception

### Re-hydrate config (and state) at the beginning of the replication activity

This pathway involves /always/ refreshing the contents of the config and state objects, as an atomic operation, as the firs step of the replication activity.  If we can get this working, we'll be sure that the sync is using the latest inputs (STATE and CONFIG) from the database.  

[...]

PR #21629, opened 2023-01-19
fix: refresh actor configuration and state between sync attempts

### LINKED ISSUE #17774: Airbyte does not use state from last attempt
## Environment
- **Airbyte version**:  0.40.14
- **OS Version / Instance**: Arch Linux
- **Deployment**: docker
- **Source Connector and version**: any
- **Destination Connector and version**: (any
- **Step where error happened**: Sync Job

## Current Behavior
If an error occurs during first attempt but the connector submited an state message, Airbyte does not use that state message on the second attempt.

If I run another sync, after the first one failed all 3 attempts, it continues from where the first attempt of the first sync failed.

## Expected Behavior
Airbyte should continue from where the first attempt stopped. I do think this is an extremely urgent bug 

## Logs

[First job](https://github.com/airbytehq/airbyte/files/9742752/a977e1f0_1275_47c8_9995_70be8c2bb76d_logs_22_txt.txt)

[Second job](https://github.com/airbytehq/airbyte/files/9742791/a977e1f0_1275_47c8_9995_70be8c2bb76d_logs_24_txt.txt)

## Steps to Reproduce
1. Use this source:
```python
#
#


import json
import typing
from datetime import datetime
from typing import Generator

from airbyte_cdk.logger import AirbyteLogger
from airbyte_cdk.models import (
    AirbyteCatalog,
    AirbyteConnectionStatus,
    AirbyteMessage,
    AirbyteRecordMessage,
    AirbyteStream,
    ConfiguredAirbyteCatalog,
    Status,
    Type, AirbyteStateMessage, AirbyteStateType, AirbyteStreamState, StreamDescriptor, AirbyteStateBlob, SyncMode,
)
from airbyte_cdk.sources import Source


class SourceStateBug(Source):
    def check(self, logger: AirbyteLogger, config: json) -> AirbyteConnectionStatus:
        try:
            return AirbyteConnectionStatus(status=Status.SUCCEEDED)
        except Exception as e:
            return AirbyteConnectionStatus(status=Status.FAILED, message=f"An exception occurred: {str(e)}")

    def discover(self, logger: AirbyteLogger, config: json) -> AirbyteCatalog:
        streams = []

        stream_name = "Test"
        json_schema = {
            "type": "object",
            "properties": {"test": {"type": "string"}}
        }

        streams.append(AirbyteStream(name=stream_name, json_schema=json_schema, supported_sync_modes=[SyncMode.full_refresh, SyncMode.incremental]))
        return AirbyteCatalog(streams=streams)

    def read(
        self, logger: AirbyteLogger, config: json, catalog: ConfiguredAirbyteCatalog, state: typing.List[AirbyteStateMessage]
    ) -> Generator[AirbyteMessage, None, None]:

        stream_name = "Test"
        fail_on_number = config["fail-on-number"]

        last_number = 1 if not state else state[0].stream.stream_state.dict()['last_number']
        print(f"Last number: {last_number}")

        for i in range(last_number, 10):
            print(f"Loop variable: {i}")
            if last_number == 1 and fail_on_number == i:
                raise Exception("Stopping for expected exception")
            yield AirbyteMessage(
                type=Type.RECORD,
                record=AirbyteRecordMessage(stream=stream_name, data={"test": f"Number: {i}"}, emitted_at=int(datetime.now().timestamp()) * 1000),
            )
            yield AirbyteMessage(
                type=Type.STATE,
                state=AirbyteStateMessage(
                    type=AirbyteStateType.STREAM,
                    stream=AirbyteStreamState(stream_descriptor=StreamDescriptor(name="Test"), stream_state=AirbyteStateBlob.parse_obj(
                        {"last_number": i}))
                )
            )

```
spec:
```yaml
documentationUrl: https://docsurl.com
connectionSpecification:
  $schema: http://json-schema.org/draft-07/schema#
  title: State Bug Spec
  type: object
  required:
    - fail-on-number
  properties:
    fail-on-number:
      type: number
      description: Number to fail on
      default: 5
```
2. Run a connection, the second attempt should continue from the number 5

Yes, but I think this is a design issue, so I would need to discuss a solution

pedroslopez:
## What

This is part 1 of #20912, focused on removing the actor configuration and state from `JobSyncConfig` and moving into a new `AttemptSyncConfig` at the Attempt level. It will also mean that config/state is fetched when building the sync input rather than retrieving it from the job's JobSyncConig.

Two issues are fixed with these changes:
1. Refreshing state between sync attempts (fixes checkpointing! - closes #17774)
2. Refreshing actor configuration between sync attempts (part of #20912)

However, it does not address when the CHECK that can run before the SyncWorkflow makes changes to the configuration. This will be addressed via a follow up PR (part 2) via the solutions outlined in the tech spec for #20912, specifically the part on changing the ConnectionManagerWorkflow activities that generate inputs.

## How

- Add new `AttemptSyncConfig` pojo with `sourceConfiguration` `destinationConfiguration` and `state` fields. 
- Remove above fields from `JobSyncConfig`
- Add `attempt_sync_config` column to the `attempts` table.
- Build `AttemptSyncConfig` when generating the sync inputs by fetching the latest actor configuration + state, and write the values to the above column.
- Introduce a new internal api endpoint to update attempt sync config
- Update references that used the JobSyncConfig to instead use the values from AttemptSyncConfig


airbyte-config/config-models/src/main/resources/types/JobResetConnectionConfig.yaml
```diff
@@ -49,9 +44,6 @@ properties:
     existingJavaType: io.airbyte.config.ResourceRequirements
   resetSourceConfiguration:
     "$ref": ResetSourceConfiguration.yaml
-  state:
-    description: optional current state of the connection
-    "$ref": State.yaml
```
pedroslopez, line 54: source/destination configuration and state are removed from JobResetConnectionConfig

airbyte-config/config-models/src/main/resources/types/AttemptSyncConfig.yaml
```diff
@@ -0,0 +1,22 @@
...
+    description: Integration specific blob. Must be a valid JSON string.
+    type: object
+    existingJavaType: com.fasterxml.jackson.databind.JsonNode
+  destinationConfiguration:
+    description: Integration specific blob. Must be a valid JSON string.
+    type: object
+    existingJavaType: com.fasterxml.jackson.databind.JsonNode
+  state:
+    description: optional state of the previous run. this object is defined per integration.
+    "$ref": State.yaml
```
pedroslopez, line 22: This new model, AttemptSyncConfig, will store the source/destination configuration and state per attempt

airbyte-config/config-models/src/main/resources/types/JobSyncConfig.yaml
```diff
@@ -64,9 +54,6 @@ properties:
     description: The webhook operation configs belonging to this workspace. Must conform to WebhookOperationConfigs.yaml.
     type: object
     existingJavaType: com.fasterxml.jackson.databind.JsonNode
-  state:
-    description: optional state of the previous run. this object is defined per integration.
-    "$ref": State.yaml
```
pedroslopez, line 69: source/destination configuration and state are removed from the JobSyncConfig

airbyte-server/src/main/java/io/airbyte/server/apis/AttemptApiController.java
```diff
@@ -46,4 +47,12 @@ public InternalOperationResult setWorkflowInAttempt(@Body final SetWorkflowInAtt
...
   }
 
+  @Override
+  @Post(uri = "/save_sync_config",
+        processes = MediaType.APPLICATION_JSON)
+  @Secured({ADMIN})
+  public InternalOperationResult saveSyncConfig(@Body final SaveAttemptSyncConfigRequestBody requestBody) {
+    return ApiHelper.execute(() -> attemptHandler.saveSyncConfig(requestBody));
+  }
+
```
pedroslopez, line 57: New api route to save the AttemptSyncConfig, marked as internal and admin-only

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityImpl.java
```diff
@@ -128,25 +179,27 @@ public GeneratedJobInput getSyncWorkflowInput(final SyncInput input) {
...
       final StandardSyncInput syncInput = new StandardSyncInput()
           .withNamespaceDefinition(config.getNamespaceDefinition())
           .withNamespaceFormat(config.getNamespaceFormat())
           .withPrefix(config.getPrefix())
           .withSourceId(standardSync.getSourceId())
           .withDestinationId(standardSync.getDestinationId())
-          .withSourceConfiguration(config.getSourceConfiguration())
-          .withDestinationConfiguration(config.getDestinationConfiguration())
+          .withSourceConfiguration(attemptSyncConfig.getSourceConfiguration())
+          .withDestinationConfiguration(attemptSyncConfig.getDestinationConfiguration())
```
pedroslopez, line 191: all references that previously used the JobSyncConfig to retrieve config and state now move to using the attemptSyncConfig

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityImpl.java
```diff
@@ -128,25 +179,27 @@ public GeneratedJobInput getSyncWorkflowInput(final SyncInput input) {
...
           .withWebhookOperationConfigs(config.getWebhookOperationConfigs())
           .withCatalog(config.getConfiguredAirbyteCatalog())
-          .withState(config.getState())
+          .withState(attemptSyncConfig.getState())
           .withResourceRequirements(config.getResourceRequirements())
           .withSourceResourceRequirements(config.getSourceResourceRequirements())
           .withDestinationResourceRequirements(config.getDestinationResourceRequirements())
           .withWorkspaceId(config.getWorkspaceId());
 
+      saveAttemptSyncConfig(jobId, attempt, connectionId, attemptSyncConfig);
```
pedroslopez, line 201: After generating the input, we persist the actor configuration and state used. This is done via an internal api call.

airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/TemporalClient.java
```diff
@@ -372,7 +373,11 @@ public TemporalResponse<ConnectorJobOutput> submitDiscoverSchema(final UUID jobI
         () -> getWorkflowStub(DiscoverCatalogWorkflow.class, TemporalJobType.DISCOVER_SCHEMA).run(jobRunConfig, launcherConfig, input));
   }
 
-  public TemporalResponse<StandardSyncOutput> submitSync(final long jobId, final int attempt, final JobSyncConfig config, final UUID connectionId) {
+  public TemporalResponse<StandardSyncOutput> submitSync(final long jobId,
+                                                         final int attempt,
+                                                         final JobSyncConfig config,
+                                                         final AttemptSyncConfig attemptConfig,
+                                                         final UUID connectionId) {
```
pedroslopez, line 380: I made the necessary changes here because we're building a sync input, but I couldn't really find where the temporal client's submitSync method actually runs. To me it seems like in practice we're using the GenerateInputActivity to do this.

Maybe @benmoriceau knows more about why this is here?

pedroslopez: Notably, this sync input is missing other recent additions, like source/destination actor IDs and allowed hosts

airbyte-persistence/job-persistence/src/test/java/io/airbyte/persistence/job/DefaultJobPersistenceTest.java
```diff
@@ -312,6 +317,25 @@ void testWriteOutput() throws IOException {
     assertEquals(List.of(failureReason1, failureReason2), storedNormalizationSummary.getFailures());
   }
 
+  @Test
+  @DisplayName("Should be able to read AttemptSyncConfig that was written")
```
evantahler, line 321: Should we add a test confirming that secrets are obfuscated in the normal way?

evantahler (approved):
👍 from me, but I think that @benmoriceau and @gosusnp should review this too.  They are also working on updating state as it is ACKED from the destination... hopefully they can re-use this APIs!

gosusnp (commented):
Sorry for the lag here... I haven't gotten to the implementation details of the PR yet.

I am trying to understand the high level change, why do we need to save the AttemptSyncConfig?
It looks like we need to be more dynamic in how we are reading the data to make sure we always run with the most recent configs and states. I'd expect it should be the same for the ConfiguredCatalog.

Also, instead of a saveSyncConfig, why not directly save the source/dest config when applicable? This is what's going to happen for the states pretty soon.

pedroslopez:
> I am trying to understand the high level change, why do we need to save the AttemptSyncConfig?

@gosusnp The main reason is just to be able to look back at a given attempt and see what configuration the attempt ran on. It's not really needed functionally, but it was brought up by Charles as a requirement that we should be able to retroactively understand the inputs for the sync.

> Also, instead of a saveSyncConfig, why not directly save the source/dest config when applicable? This is what's going to happen for the states pretty soon.

I'm not sure I follow. We are directly saving the new configurations to sources and destinations as part of the sync and other jobs. The new endpoint added here, `/attempt/save_sync_config`, is specifically to persist the AttemptSyncConfig at the attempt level for the reason outlined above, when building up the sync input. Nothing is changed here for the actor configurations, we're just retrieving it and persisting it to the attempt.

I added a new endpoint rather than just calling on the jobPersistence because of a concern from benoit/jonathan on the tech spec to avoid introducing more dependencies on direct db access, and moving towards internal APIs.

airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/TemporalClient.java
benmoriceau: I think that it is some legacy from before we had the ConnectionManagerWorkflow was not existing.
The WorkerRun (that call that method upon creation) seems to be only useful to get the jobRoot...

This type of starting a temporal job is still useful for check connection and discover schema.

benmoriceau (approved):
Sorry for the delay, I missed this PR. It LGTM, it doesn't seems that there is a concurrency issue with this PR. About the comment where you tagged me, I believe that it is some legacy code that should be removed. This shouldn't be part of your PR and I'll create an issue for for that.

Issue #19408, opened 2022-11-15
[Low-Code CDK] Rename $options keyword to $parameter in the connector manifest


brianjlai:
## Tell us about the problem you're trying to solve
**BREAKING CHANGE**

The keyword `$options` is a very vague and overloaded term that doesn't do a good job of conveying its purpose.
This field keyword is used to propagate values accessible at runtime for string interpolation and to be assigned to component fields. We should replace this with the `$parameters` keyword which is a more accurate term

## Describe the solution you would like
As explained above, all usages of `$options` should be replaced with `$parameters`. An example of what the new manifest should look like:
```
sample_stream:
  type: DeclarativeStream
  $parameters:
    name: sample
  primary_key: id
  schema_loader:
    $ref: "*ref(definitions.schema_loader)"
  retriever:
    $ref: "*ref(definitions.retriever)"
    record_selector:
      $ref: "*ref(definitions.results_selector)"
    requester:
      $ref: "*ref(definitions.requester)"
      path: "/messages/{{ stream_slice.id }}/history"
```

The low-code documentation should also be reflected to use the new `$parameter` keyword.

The sample manifest file `{{snakeCase name}}.yaml.hbs` in the low-code generator should also have `$options` switched to `$parameters`.

A well tested script to replace all instances of `$options` with `$parameters` should be written and ready for use when we make the major version bump. The script should also check for any usage of options within interpolation blocks such as `{{ options['name'] }}` and replace it as `{{ parameters['name'] }}`

## Acceptance Criteria

- [ ] The parser should now interpret $parameters instead of $options keyword
- [ ] The declarative component dataclasses should be renamed from options to parameters and they should be passed to interpolated classes as parameters
- [ ] A migration should transform all usage of the $options keyword to $parameters
- [ ] A migration should transform all existing interpolation templates that use options to use parameters
- [ ] Documentation should be updated to reflect the change

## Implementation Approach
The basic flow for how a low code connector transforms a YAML spec into a declarative source that can handle connector operations is made up of two steps:

1. Parse the source's manifest YAML file into a mapping of component definition mappings.
2. Depending on the operation (spec, check, read), we use the `DeclarativeComponentFactory` to construct the relevant declarative components. It is during this create component step that we read apply `$options` at the current level and pass it down to child components. 

In `factory.py` and `create_partial.py` anywhere we use the keyword `$options` should be replaced with `$parameters`.

For consistency, we should also rename the `options` parameter on all declarative components to `parameters`.

[...]

PR #21632, opened 2023-01-20
[Low-Code CDK] Replace the $options keyword with $parameters

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/19408

## What
The `$options` keywords is a mostly nondescript word which is not really representative of how it's used within the declarative framework. `$parameters` is more apt since it is being passed through to various components. This change replaces all the usages of options with parameters in the following areas:

- The developer facing schema
- The component construction flow
- At runtime during a sync, using `parameters` instead of `options`
- Updated documentation
- Updated low code templates
- Migrating existing connectors (will be added in the next commit)

## How
This change looks pretty big but its mostly renaming and refactoring. I'll break this into a couple different functional areas that had code changes:

- I updated the declarative_component_schema.yaml to reflect the language change. I also regenerated the Pydantic models and updated the swap to replace `_parameters` with `parameters`.
- For the Pydantic flow `manifest_component_transformer.py` and `model_to_component_factory.py` are updated to apply and construct using `parameters`.
- For the legacy flow, we now key on `parameters` to retain the existing validation flow.
- Every declarative component used to receive `options` from the parent. It will now arrive from the parent as `parameters`
- At runtime, to maintain consistency, the interpolation keyword has also been replaced. For example, `{{ parameters.name }}` can be specified at runtime.
- A lot of test updates since `$options` is used so widely
- All the documentation and tutorials have been updated.
- Templates also reference `$options` which was updated
- I included the migration script I will run, but I haven't included the resulting manifests post-migration. That will follow soon

## Recommended reading order
2. `manifest_component_transformer.py`
3. `model_to_component_factory.py`
4.  `factory.py`
5. All the dataclass components whose constructors were affected
6. `generate-component-manifest.sh`
7. `migrate_options_to_parameters.sh`
8. `/docs`


airbyte-cdk/python/bin/migrate_options_to_parameters.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+echo "Starting migration"
+
+[ -z "$ROOT_DIR" ] && exit 1
+
+CONNECTORS_DIR=$ROOT_DIR/airbyte-integrations/connectors
+CDK_DIR=$ROOT_DIR/airbyte-cdk/python/
```
girarda, line 8: should we also migrate the connector builder server?

girarda (approved):
PR looks good but I found a [few more instances](https://github.com/airbytehq/airbyte/commit/b802bc5c7ca3f024682e9a6c575e6b0d590d55cf) of `options` I think we should replace

airbyte-cdk/python/bin/migrate_options_to_parameters.sh
brianjlai: Do you mean updating the starter manifest code in `default_api.py`? If so yes I agree that is a good suggestion

airbyte-cdk/python/unit_tests/sources/declarative/requesters/paginators/test_offset_increment.py
```diff
@@ -41,11 +37,7 @@ def test_offset_increment_paginator_strategy(test_name, page_size, options, expe
...
-    paginator_strategy = OffsetIncrement(
-        page_size="{{ options['page_size'] }}",
-        options={"page_size": "invalid value"},
-        config={}
-    )
+    paginator_strategy = OffsetIncrement(page_size="{{ parameters['page_size'] }}", parameters={"page_size": "invalid value"}, config={})
     with pytest.raises(Exception) as exc:
         paginator_strategy.get_page_size()
-    assert str(exc.value) == 'invalid value is of type <class \'str\'>. Expected <class \'int\'>'
+    assert str(exc.value) == "invalid value is of type <class 'str'>. Expected <class 'int'>"
```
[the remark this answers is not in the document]
brianjlai: Yeah this likely came as a result of `SUB_BUILD=CONNECTORS_BASE ./gradlew format` which we use to regenerate the catalog. I didn't change this myself

airbyte-cdk/python/bin/migrate_options_to_parameters.sh
```diff
@@ -0,0 +1,23 @@
...
+
+for directory in $CONNECTORS_DIR/source-* ; do
+  MANIFEST_DIRECTORY=$(basename $directory | tr - _)
+  SOURCE_NAME=${MANIFEST_DIRECTORY#source_}
+  FILEPATH=$directory/source_$SOURCE_NAME/$SOURCE_NAME.yaml
+
+  echo "Migrating manifest located at $FILEPATH"
+  if test -f $FILEPATH; then
+    # In place replacement of $options to $parameters
+    sed -i '' -E 's/\$options/\$parameters/' $FILEPATH
```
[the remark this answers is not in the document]
brianjlai: Fair point, I can kind of see it going either way. It might overdoing it to make a really generic search replace for what might be a short term set of simple migrations. But might revisit this if it starts to get unwieldly

brianjlai:
Thanks for highlighting the instances I missed, I'll push a commit w/ those addressed

airbyte-cdk/python/unit_tests/sources/declarative/auth/test_session_token_auth.py
```diff
@@ -129,56 +128,55 @@ def test_get_token_username_password(requests_mock):
...
         login_url=login_url,
-        validate_session_url=validate_session_url
+        validate_session_url=validate_session_url,
     ).is_valid_session_token()
 
 
 def test_get_new_session_token(requests_mock):
-    requests_mock.post(f"{config['instance_api_url']}session", headers={"Content-Type": "application/json"},
-                       json={"id": "some session id"})
+    requests_mock.post(f"{config['instance_api_url']}session", headers={"Content-Type": "application/json"}, json={"id": "some session id"})
```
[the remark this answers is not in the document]
brianjlai: not 100% on this one, its right on the final max length of a line, this was automated from the format command though, i'm more surprised that it didn't get flagged in our CI because that should rely on the format command I ran

airbyte-cdk/python/bin/migrate_options_to_parameters.sh
girarda: yes, and the comments in the openapi spec https://github.com/airbytehq/airbyte/blob/master/airbyte-connector-builder-server/src/main/openapi/openapi.yaml#L74

brianjlai: ah right adding!

brianjlai:
noting that running `./gradlew validateSourceYamlManifest` was successful after running the migration script and updating some custom components that relied on options. merging this into the branch

PR #21634, opened 2023-01-20
fix-postgres-source: get rid of short lived objects

subodh1810:
Issue : https://github.com/airbytehq/airbyte/issues/21806


airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
```diff
@@ -60,6 +63,13 @@ public class PostgresSourceOperations extends AbstractJdbcCompatibleSourceOperat
   private static final String TIMESTAMPTZ = "timestamptz";
   private static final String TIMETZ = "timetz";
   private static final ObjectMapper OBJECT_MAPPER = MoreMappers.initMapper();
+  private static final Map<Integer, PostgresType> POSTGRES_TYPE_DICT = new HashMap<>();
+  private Map<String, Map<String, ColumnInfo>> streamColumnInfo = new HashMap<>();
+
+  public PostgresSourceOperations() {
+    super();
+    Arrays.stream(PostgresType.class.getEnumConstants()).forEach(c -> POSTGRES_TYPE_DICT.put(c.type, c));
```
bleonard, line 71: I think this is weird that you are doing in the constructor beacause you are filling up a global static. Maybe multiple times.
does this work in Java?
```java
// at the top where it is now.
private static final Map<Integer, PostgresType> POSTGRES_TYPE_DICT = new HashMap<>(); 
static
    {
        Arrays.stream(PostgresType.class.getEnumConstants()).forEach(c -> POSTGRES_TYPE_DICT.put(c.type, c));
    }
```

If it has to be in a constructor, then at least make sure you are only doing it once by it not being empty or with a boolean.

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
```diff
@@ -60,6 +63,12 @@ public class PostgresSourceOperations extends AbstractJdbcCompatibleSourceOperat
   private static final String TIMESTAMPTZ = "timestamptz";
   private static final String TIMETZ = "timetz";
   private static final ObjectMapper OBJECT_MAPPER = MoreMappers.initMapper();
+  private static final Map<Integer, PostgresType> POSTGRES_TYPE_DICT = new HashMap<>();
+  private final Map<String, Map<String, ColumnInfo>> streamColumnInfo = new HashMap<>();
```
akashkulk, line 67: Can you just add a comment describing what this streamColumnInfo does

akashkulk (approved):
LGTM - just adds comments explaining the caching/optimization

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
```diff
@@ -199,22 +187,22 @@ public void copyToJsonField(final ResultSet resultSet, final int colIndex, final
         case "_timetz" -> putTimeTzArray(json, columnName, resultSet, colIndex);
         case "_time" -> putTimeArray(json, columnName, resultSet, colIndex);
         default -> {
-          switch (columnType) {
-            case BOOLEAN -> putBoolean(json, columnName, resultSet, colIndex);
+          switch (columnInfo.columnType) {
+            case BOOLEAN -> json.put(columnName, value.equalsIgnoreCase("t"));
```
rodireich, line 191: why not `putBoolean()`?

rodireich (approved):

PR #21643, opened 2023-01-20
🪟🐛 Connector builder UI: Fix inputs modal

flash1293:
## What

This PR fixes two small issues with the inputs modal:
* The default value input type is not updated correctly when the type changes
* Placeholder is not reflected in the test input form

## How

For the first one, I tried for a bit to find a "real" solution, but it became way to complex to justify it here. The problem is that formik will not re-render the inner field if the value doesn't change which means changes in the props after the initial render might not be picked up correctly. I tried to come up with a generic way to enforce an update for any of the props, but there are two issues with this:
* There are a lot of props which are not memoized, so we would need to take care of that first which makes the change pretty big (otherwise the full thing would just re-render all the time as before)
* It's pretty difficult to force formik to re-render in this situation and the complexity doesn't seem justified - we basically need to re-implement the regular formik fastfield "shouldUpdate" logic, then do a standard memoize check on the passed down props as well. To do this in a performant way isn't super simple, there are libraries for that but it doesn't look like it's worth it.

The solution I went with is adding a key prop to the fastfield with the type value so it will definitely re-render in this situation. It fixes the problem at hand and I hope we can soon refactor this code - with react-hook-form we won't have this problem.

For the second one - I checked and you are right, we don't support placeholders like this. I think this was a left-over from working from the original designs that included that field. I removed it.


lmossman (approved):
Changes seem fine to me. I agree this doesn't feel like the "correct" way to solve this, but it is very minimal and should be okay for now until we switch to a different form library.

Tested locally and it seems to fix the issue

lmossman:
Merging now since I am going to release OSS and I want this fix included

octavia-approvington:
Lets do it! <br/> giddddy-up

octavia-approvington (approved):

PR #21647, opened 2023-01-20
fix column selection handling for syncs without cursors

mfsiega-airbyte:
## What
Fix the column selection handling for syncs without cursors.

The way we handle syncs when column selection is enabled *assumes* that there will be a cursor field present, and all of our test cases supply it.

## How
Correctly stop relying on a cursor being configured.

## Testing
Added unit test coverage, also deployed to dev-3 and verified that it worked there.


joeykmh (commented):
I'll leave the approval to @pmossman, but the logic in `CatalogConverter.java` looks good, now I see where the bug was!

pmossman (approved):
Logic checks out to me, and appreciate the test refactoring!

PR #21655, opened 2023-01-20
Add No content status code

benmoriceau:
## What
Some method we returning 204 with jaxRs but 200 with Micronaut, this is fixing it.


jdpgrailsdev (approved):
:shipit:

PR #21660, opened 2023-01-20
[Low Code CDK] Update ref format from `*ref(.*)` -> `#/` in all low code source manifests

clnoll:
## What
Updates all manifests to the new `ref` format, `#/`.

## How
Runs the script from https://github.com/airbytehq/airbyte/pull/21434/files#diff-ecce74fe42149d709b5dd2f582fed104fdd21c81fd2603b06d117182ed49d4e3.

## 🚨 User Impact 🚨
Currently we are not planning to create a new release for all of the sources with updated manifests; the updated manifests will be in place and the sources can be re-relased as needed.

## Pre-merge Checklist

- [x] Spot check sources with the [`/test connector=connectors/<name>` command](https://docs.airbyte.io/connector-development#updating-an-existing-connector)
- [x] Run `./gradlew validateSourceYamlManifest`

### DIFF
--- a/airbyte-integrations/connectors/source-activecampaign/source_activecampaign/activecampaign.yaml
+++ b/airbyte-integrations/connectors/source-activecampaign/source_activecampaign/activecampaign.yaml
@@ -17,12 +17,12 @@ definitions:
 
   retriever:
     record_selector:
-      $ref: "*ref(definitions.selector)"
+      $ref: "#/definitions/selector"
 
     # API Docs: https://developers.activecampaign.com/reference/pagination
     paginator:
       type: DefaultPaginator
-      url_base: "*ref(definitions.requester.url_base)"
+      url_base: "#/definitions/requester/url_base"
       page_size_option:
         inject_into: "request_parameter"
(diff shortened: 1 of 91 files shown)


brianjlai (approved):
Spot checking a handful of the manifests and looks good to me. And for everything else `./gradlew validateSourceYamlManifest` should sus out a lot of cases since a ref that isn't working should likely cause an invalid manifest in most cases.

Not quite a block on approval, but one weird thing I'm seeing is -6 lines (+2097, -2103). I would have expected this to edit the same number of lines since we're doing an in place swap of the ref format so I would think this would have equal number of added/removed on the PR, do you know why we are mismatching?

clnoll:
Good call-out @brianjlai. This is because of some formatting changes I had to make to `source-facebook-pages` to get the linter to pass.

And just want to confirm that my spot checks and `./gradlew validateSourceYamlManifest` were clean.

brianjlai:
makes sense and yup I see the formatting in that connector. thanks for confirming! no other questions on my end!

PR #21664, opened 2023-01-20
Add required roles to API endpoints

### LINKED ISSUE #21596: Support authentication/authorization for airbyte-server endpoints in OSS
Currently, the OSS API server (airbyte-server) does not use any authentication or authorization for its exposed endpoints.  This conflicts with the Cloud version of airbyte-server, which does enforce authentication and authorization.  To reduce this disparity and make it easier to work with both modules, the OSS version of airbyte-server should utilize Micronaut's security framework.

A basic authentication/authorization provider should be implemented using Micronaut that is only used by OSS.  This implementation should always grant the highest privilege to any authenticated OSS user.  All controller endpoints should be annotated with the correct roles, as currently defined in the Cloud version of the airbyte-server.

jdpgrailsdev:
## What
* Add required roles to API endpoints
* fix #21596

## How
* Ensure that `@Secured` annotations are added to each API endpoint.  Required role should match those currently required in Cloud implementation.

## Recommended reading order
1. `AuthRole.java` (copy from Cloud)
2. `AuthRoleConstants.java`
3. `*Controller.java`
4. `application.yml`

## More Context

Originally, the intent was to implement a basic authentication provider that would be replaced in Cloud.  However, after attempting to do that, I discovered that because the system does not have a consistent auth story currently (e.g. we have an nginx proxy that requires auth in Docker, but doesn't exist when running OSS in K8), it is not easy to handle both cases.  Ultimately, we want to enable basic auth consistently in OSS and the introduction of a basic authentication provider can be added at that time.  For now, we will simply add the annotations so that Cloud doesn't need to extend every controller endpoint and disable auth via Micronaut by default in airbyte-server in OSS.

## Testing

* Verified that `@Secured` annotations do not impact OSS with Micronaut security disabled by running system locally via docker-compose
* All tests pass locally
* All acceptance tests pass locally


airbyte-server/src/main/resources/application.yml
```diff
@@ -2,23 +2,13 @@ micronaut:
...
-      - pattern: /**
-        httpMethod: POST
-        access:
-          - isAnonymous()
-      - pattern: /**
-        httpMethod: HEAD
-        access:
-          - isAnonymous()
+    authentication-provider-strategy: ALL
+    enabled: ${API_AUTHORIZATION_ENABLED:false}
```
benmoriceau, line 6: If I understand well, this will disable the authentication on OSS

airbyte-server/src/main/resources/application.yml
```diff
@@ -134,3 +150,10 @@ jooq:
     jobs:
       jackson-converter-enabled: true
       sql-dialect: POSTGRES
+
+logger:
+  levels:
+  # Uncomment to help resolve issues with conditional beans
+  #    io.micronaut.context.condition: DEBUG
+  # Uncomment to help resolve issues with security beans
+  #    io.micronaut.security: DEBUG
```
benmoriceau, line 159: Nit: new line

airbyte-server/src/main/java/io/airbyte/server/apis/AttemptApiController.java
```diff
@@ -36,6 +41,7 @@ public InternalOperationResult saveStats(final SaveStatsRequestBody requestBody)
   @Override
```
benmoriceau, line 41: Should there be some authorization on the saveStats as well?

benmoriceau (commented):
I have a question: I haven't added a comment everywhere but should we add at least `@Secured({AUTHENTICATED_USER})` to all the endpoints? Some don't have a Secured annotation and I wonder why.

airbyte-server/src/main/resources/application.yml
jdpgrailsdev: Correct.  By default, it will disable Micronaut's built-in security support.  If we set it to true in OSS and/or in Cloud, it will enable it.  The plan will be to set the env var in the Helm charts for the airbyte-server in Cloud to `true`.

jdpgrailsdev:
> I have a question: I haven't added a comment everywhere but should we add at least `@Secured({AUTHENTICATED_USER})` to all the endpoints? Some don't have a Secured annotation and I wonder why.

I think that we need to confer with the compose team on this.  I'm not sure why some endpoints do not have a check in cloud.

airbyte-server/src/main/java/io/airbyte/server/apis/AttemptApiController.java
jdpgrailsdev: Maybe.  This endpoint is unimplemented in Cloud and therefore there is nothing to port over.  If we can come up with what we want on those unimplemented endpoints, we can set it.  However, for the sake of porting this over, my plan is to match exactly what is in Cloud at the moment.

benmoriceau (approved):
Approved since the un-annotated endpoints are not supported in cloud.

PR #21668, opened 2023-01-20
Allow all origin in cors

benmoriceau:
## What
Micronaut has some check in the CORS filter specifiv to the localhost. This will allow all the Cors requests, including the one coming from localhost.


jdpgrailsdev (approved):
:shipit:

Issue #20969, opened 2023-01-03
Update connection page scrolling behavior


edmundito:
## Tell us about the problem you're trying to solve
The design for the updated connection page has a layout that sizes to the height of the screen where only the streams table scrolls. This is not the way that the page is currently implemented because there are a few areas to consider:

* The top area (above the streams table) takes a significant amount of vertical space, meaning that depending on the height of the user's browser window, they can have little space for the streams
* The name field was designed only to show it as one line with an ellipsis when it was too long. Currently, the implementation does not support this (tracked in #19227)
* This only works well with the new streams table (since it has fewer form fields above the table.) 

## Describe the solution you’d like
1. The proposed design needs more work. A better solution needs to be discussed with design, one that can work with users that have smaller screens
2. Implement new design 

## Describe the alternative you’ve considered or used
Leave it as it is: The pages are sized based on the length of the table.

## Additional context
[Figma](https://www.figma.com/proto/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?page-id=972%3A36208&node-id=972%3A36212&viewport=446%2C338%2C0.09&scaling=min-zoom&starting-point-node-id=972%3A36212&show-proto-sidebar=1)

```[tasklist]
- [ ] #21880
- [ ] #22173
- [ ] #22524
- [ ] https://github.com/airbytehq/airbyte-platform/pull/72
```

PR #21673, opened 2023-01-20
🪟 🎨  Update page bottom margins to be consistent

edmundito:
## What
Related to #20969

Updates the bottom margin of pages to be a consistent 20px, or 88px in cloud (to accommodate for Intercom.). Pages impacted include:

* Connection (mainly Replication and Transformations pages)
* Connection, source, and destination lists
* Create new Connection
* Create source or destination
* Source or destination settings

## How
Removes extraneous bottom paddings in a few pages and leaves the padding to the layout components (MainPageWithScroll or FormPageContent). The connectors settings pages have significant cleanup so that they use FlexContainer and less custom scss, and lets the container control the bottom margin.

For cloud, it checks if it's cloud before adding the extra padding.


krishnaglick (approved):
Code LGTM, didn't notice any issues w/ padding on the stated paged.

PR #21675, opened 2023-01-20
🪟 🎨 🐛 Fix streams table card margins

edmundito:
## What
Fixes an issue where the streams table card were wider than the rest of the cards

## How
This was caused by a race condition between the Section and the ConectionFormFields CSS where the Section CSS would override the custom CSS. Now the Section controls the flushing.


dizel852 (approved):
LGTM 👍 
Tested locally - no issues

PR #21677, opened 2023-01-20
Source Amazon Ads: fix reports stream records primary keys

roman-yermilov-gl:
## What
*Because of non-unique values are defined as a composite primary key of the report streams, we are missing a bunch of records per each sync if dedup+history strategy chosen. This PR contains a fix but also adds some possibly incompatible changes that needs to be discussed. At [this point](https://github.com/airbytehq/airbyte/compare/ryermilov/source-amazon-ads-fix-report-stream-record-primary-keys?expand=1#diff-4e3b79abb6fc723d5f9e51b1ff286901bb4bfb259b525fdcd02b5e1a44ed5d10R377) I splitted `asins` report to two reports `asins_keywords` and `asins_targets`. We need this separation for everything to be consistent but maybe it require some preparation before delivery to cloud.*


airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/streams/report_streams/report_streams.py
```diff
@@ -89,7 +89,7 @@ class ReportStream(BasicAmazonAdsStream, ABC):
     Common base class for report streams
     """
 
-    primary_key = ["profileId", "recordType", "reportDate", "updatedAt"]
+    primary_key = ["profileId", "recordType", "reportDate", "updatedAt", "recordId"]
```
erohmensing, line 92: Are all of these still needed for uniqueness? Trying to understand where `reportDate` fits in here. I could see "profileid's report about recordtype:recordID" being an identifier for one report, and reportDate potentially being a way to distinguish multiple reports about the same recordtype:recordID, but couldn't you create multiple on the same date?

And doesn't `updatedAt` as part of the PK make updating an existing report effectively turn it into a new report from the perspective of the deduper? (I think this was mentioned in the issue)

erohmensing:
>  This PR contains a fix but also adds some possibly incompatible changes that needs to be discussed.

Can you lay out the possible incompatible changes?

airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/streams/report_streams/products_report.py
```diff
@@ -243,6 +243,17 @@
...
 
+METRICS_TYPE_TO__MAP = {
+    "campaigns": "campaignId",
+    "adGroups": "adGroupId",
+    "keywords": "keywordId",
+    "productAds": "adId",
+    "asins_keywords": "asin",
+    "asins_targets": "asin",
+    "targets": "targetId",
+}
```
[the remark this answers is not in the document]
roman-yermilov-gl: As we can see there is an abstract `metrics_map` property that must be implemented by every successor of the `ReportStream` class. Now we need to extend it by adding an id mapping for every `metrics_map`. If we put this mapping to the base class then it turns out that we put implementation to the base class, but it tied to implementation of its successors. Then we need to combine all the `metrics_map` and also move them to the base class... It seems not correct to me. I like the idea that it is better to extend a logic instead of making some kind of god object (at least in this case). And last but not least sometimes key/values pairs are differ:
- products report: `"keywords": "keywordId"`
- brands video report: `"keywords": "keywordBid"`

airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/streams/report_streams/report_streams.py
roman-yermilov-gl: You are right. We need `updatedAt` to distinguish multiple syncs of the same report done at the same date. But `updatedAt` is a `reportDate` + `datetime`, so we have `reportDate` added two times to primary key for no reason. It breaks nothing for now but should be avoided

airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/streams/report_streams/products_report.py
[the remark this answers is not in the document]
erohmensing: Got it - in the case that they sometimes differ, I think your implementation makes sense. I wasn't sure if the metrics_map was also shared between all the subclasses as well and could be combined into the base class. Thanks for the clarification :)

airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/streams/report_streams/report_streams.py
erohmensing: > distinguish multiple syncs of the same report done at the same date

In that case its still the same report though right? In which case it should probably be a cursor, not part of the primary key? Unless I'm missing something

roman-yermilov-gl: Yes, this is true. https://github.com/airbytehq/airbyte/pull/21677/files#diff-4e3b79abb6fc723d5f9e51b1ff286901bb4bfb259b525fdcd02b5e1a44ed5d10R100

roman-yermilov-gl:
> > This PR contains a fix but also adds some possibly incompatible changes that needs to be discussed.
> 
> Can you lay out the possible incompatible changes?

It's on description. Not sure what else can I add more, unless concrete question

lazebnyi (approved):

erohmensing:
> We need this separation for everything to be consistent but maybe it require some preparation before delivery to cloud.

What is the preparation needed - is it on our end or on the customer's end? If it's on the customer's end we should figure out how many people are impacted by this so that TCS can coordinate

airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/streams/report_streams/report_streams.py
erohmensing: This doesn't seem to link to a diff anymore - I'm suggesting that I think `updatedAt` doesn't belong in the composite primary key

girarda: I agree with Ella. `updatedAt` should be a cursor field used to track updates on the daily report (keyed off the `reportDate` field)

girarda (changes requested):

airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/streams/report_streams/report_streams.py
grubberr: Please keep in mind that `updatedAt` is artificial field which is added when real sync was done 
it's possible to have such object:

```json
{
 "reportDate": "2023-01-10"
 "updatedAt": "2023-01-26T00:00:00Z"
}
```

it means we synced report object for `2023-01-10`, but real sync was done `2023-01-26T00:00:00Z`

We use
```
cursor_field: reportDate
```

to keep track what last report was synced.

davydov-d:
Hey guys, Roman is unavailable for a couple of days so I will pick up this issue. Serhii and I discussed this problem and the suggested solution and decided that it needs to be thinked over one more time and probably reimplemented. Suggested solution fixes existing problem but most probably introduces a new one.
Will keep you updated

erohmensing:
Thank you for the update @davydov-d1

airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/streams/report_streams/report_streams.py
davydov-d: So.. let me start with one implicit problem we're trying to solve as well. Since the cursor field we have is of a `date` type and the dedup sync may be running once in a couple of hours, we're not sure which record with the same primary key and cursor value would be deleted and which would be stored. If the one with bigger `emitted_at` value is stored, as Edward [supposes](https://airbytehq.slack.com/archives/C02UF50V9HA/p1674834152323399?thread_ts=1674819243.778729&cid=C02UF50V9HA), then we're fine to drop `updatedAt` both from the primary key and from the record. Otherwise we can not guarantee that the latest data is stored, thus can not guarantee correctness of a dedup sync mode and need to remove both `updatedAt`, `recordId` and the primary key itself.

@erohmensing @girarda please let me know your opinion

davydov-d: btw, `updatedAt` is an artificial field and is actually a time when the record is emitted

erohmensing: > If the one with bigger emitted_at value is stored, as Edward [supposes](https://airbytehq.slack.com/archives/C02UF50V9HA/p1674834152323399?thread_ts=1674819243.778729&cid=C02UF50V9HA), then we're fine to drop updatedAt both from the primary key and from the record.

This sounds like the expected behavior to me, and I think we'd be seeing a lot more on call issues if it weren't the case. Can we test that that is what happens when the cursor field is date type and being synced on the level of hours? If so, I think dropping `updatedAt` from the PK, and from the record if emitted_at is already inserted into the record

davydov-d: makes sense, I will test it

davydov-d: So I checked it and yes, it works as expected so I'm gonna remove `updatedAt` field

airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/streams/report_streams/products_report.py
```diff
@@ -252,6 +263,7 @@ def report_init_endpoint(self, record_type: str) -> str:
         return f"/v2/sp/{record_type}/report"
 
     metrics_map = METRICS_MAP
+    metrics_type_to_id_map = METRICS_TYPE_TO__MAP
```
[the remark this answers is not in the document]
davydov-d: fixed

airbyte-integrations/connectors/source-amazon-ads/source_amazon_ads/streams/report_streams/report_streams.py
davydov-d: done

davydov-d:
@erohmensing PR updated, please review

davydov-d:
@erohmensing Sorry, I overlooked it, must have been pre-commit hooks. Removed non related changes.

erohmensing (approved):
Please make sure to update TCS with any changes customers will have to make and get their OK before merging (they may want to schedule it if it is still breaking) - thanks for adopting this PR!

girarda (approved):
lgtm. Reminder there is a breaking change so we'll need to coordinate with TCS to release this to Cloud

PR #21680, opened 2023-01-20
Trigger notifications for changed schemas

alovew:
NOTE: this feature is turned OFF and will not be enabled yet when this PR is merged

Add functionality to trigger notifications when schema changes are detected

- Notification triggering is added in `SchedulerHandler.java`
- Some logic is moved out of the ConnectionNotificationWorkflow and into the NotifySchemaChangeActivityImpl
- New property on SourceDiscoverySchemaRequestBody: notifySchemaChange. It is set to true when schema refreshes are triggered from the SyncWorkflow, but set to false when schema refreshes are triggered manually.


airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SchedulerHandler.java
```diff
@@ -393,10 +411,13 @@ public JobInfoRead cancelJob(final JobIdRequestBody jobIdRequestBody) throws IOE
...
   // wants the connection disabled when non-breaking changes are detected. If so, disable that
   // connection. Modify the current discoveredSchema object to add a CatalogDiff,
-  // containsBreakingChange paramter, and connectionStatus parameter.
+  // containsBreakingChange parameter, and connectionStatus parameter.
   private void generateCatalogDiffsAndDisableConnectionsIfNeeded(final SourceDiscoverSchemaRead discoveredSchema,
-                                                                 final SourceDiscoverSchemaRequestBody discoverSchemaRequestBody)
+                                                                 final SourceDiscoverSchemaRequestBody discoverSchemaRequestBody,
+                                                                 final UUID workspaceId)
       throws JsonValidationException, ConfigNotFoundException, IOException {
+    final ConnectionNotificationWorkflow notificationWorkflow =
```
gosusnp, line 419: So far, I think the temporal access have been generally abstracted ([ConnectionManagerUtils](https://github.com/airbytehq/airbyte/blob/3c52168f80cdc14cf346b4cadf5660fff0ad4936/airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/ConnectionManagerUtils.java)), should we either extend this object or have a similar abstraction for the notifications (NotificationClient could be a fit)?

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SchedulerHandler.java
```diff
@@ -417,12 +438,27 @@ private void generateCatalogDiffsAndDisableConnectionsIfNeeded(final SourceDisco
       }
       updateObject.status(connectionStatus);
       connectionsHandler.updateConnection(updateObject);
+      if (shouldNotifySchemaChange(diff, connectionRead, discoverSchemaRequestBody)) {
+        final String url = webUrlHelper.getConnectionUrl(workspaceId, connectionRead.getConnectionId());
+        try {
+          notificationWorkflow.sendSchemaChangeNotification(connectionRead.getConnectionId(), url);
```
gosusnp, line 444: Shouldn't the notification also take the information from the update that could happen line 450?

airbyte-api/src/main/openapi/config.yaml
```diff
@@ -2789,6 +2789,8 @@ components:
           format: uuid
         disable_cache:
           type: boolean
+        notifySchemaChange:
```
gosusnp, line 2792: This doesn't feel right, I don't think we want to expose this flag here, it keeps on adding to the fact that our APIs are sometimes doing too many things and we should try to rationalize into simpler operations.
It might be a good time to extract the notification code into its own class to avoid this, what do you think?

gosusnp (commented):
My main concern is about having to expose extra config in the API for something that seems to be "platform internal".

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SchedulerHandler.java
alovew: which information are you thinking? I was hoping to add the diff info afterward, but right now the notification includes whether or not it's a breaking change and a link to the connection page

airbyte-api/src/main/openapi/config.yaml
alovew: What do you mean exactly? this is here because we do want to notify for automatic schema changes but not for manual schema changes, so it's an option tied to the source schema refresh

gosusnp: Do we want to keep going through the API for the automatic refresh? I thought this should only be used for manual schema change.
We currently have handlers depending on handlers which is a dependencies we should remove, it looks like we're exposing more internals here due to this bad dependency. I am wondering if we can have a solution that doesn't involve exposing more config not intended to be exposed in the first place.

alovew: I think at this point, yes, partly since we want to get this out this week. Maybe we should jump on a call about this since I'm not sure exactly what kind of refactor you're thinking of - but is it ok if we refactor afterward?

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SchedulerHandler.java
alovew: Yes, breaking change is included in the notification - diff isn't yet

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/SchedulerHandler.java
benmoriceau: Yes I agree, the scheduler shouldn't be aware of temporal. They should launch the temporal action through an interface/utils that could be replace by something else than temporal

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/WebBackendConnectionsHandler.java
```diff
@@ -413,7 +413,7 @@ private Optional<SourceDiscoverSchemaRead> getRefreshedSchema(final UUID sourceI
     final SourceDiscoverSchemaRequestBody discoverSchemaReadReq = new SourceDiscoverSchemaRequestBody()
         .sourceId(sourceId)
         .disableCache(true)
-        .connectionId(connectionId);
+        .connectionId(connectionId).notifySchemaChange(false);
```
benmoriceau, line 416: nit, can it be on 2 lines

airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
```diff
@@ -1042,7 +1042,7 @@ public boolean shouldRunConnectionManagerWorkflows() {
 
   @Override
   public boolean shouldRunNotifyWorkflows() {
-    return getEnvOrDefault(SHOULD_RUN_NOTIFY_WORKFLOWS, false);
+    return getEnvOrDefault(SHOULD_RUN_NOTIFY_WORKFLOWS, true);
```
benmoriceau, line 1045: Is there anything else in the env that needs to be configured in order to be able to make a notification. If so we should leave the default as false.

airbyte-server/src/main/java/io/airbyte/server/ServerApp.java
```diff
@@ -265,7 +267,7 @@ public static ServerRunnable getServer(final ServerFactory apiFactory,
         configs.getLogConfigs(),
         eventRunner,
         connectionsHandler,
-        envVariableFeatureFlags);
+        envVariableFeatureFlags, workflowClient, webUrlHelper);
```
benmoriceau, line 270: nit: missing new lines

airbyte-api/src/main/openapi/config.yaml
benmoriceau: @gosusnp I think that the main issue here is that the API is doing a lot here. Another fix would be to also have another endpoint that would be internal (maybe this one and another one in the webbackend for the manual operation). Should we exclude that from the scope of the ticket?

airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
alovew: I'm not sure - that's why I set the default to true

benmoriceau: So should the default be false then so we don't try to send a notification if the webhook is not configured in an OSS deployment?

benmoriceau (commented):
Still got a small concern about  the default value of the feature flag. Apart from that it LGTM.

airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
alovew: all of the other workflow env variables are defaulting to true so I thought I'd follow the same pattern - is there a reason not to?

airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/NotificationUtils.java
```diff
@@ -0,0 +1,33 @@
...
+import io.airbyte.config.persistence.ConfigNotFoundException;
+import io.airbyte.validation.json.JsonValidationException;
+import io.temporal.client.WorkflowClient;
+import jakarta.inject.Singleton;
+import java.io.IOException;
+import java.util.UUID;
+import lombok.NoArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+
+@NoArgsConstructor
```
jdpgrailsdev, line 18: Nit:  we should avoid using this with singletons.  This is mainly because if/when we need to add annotations to the parameters, we cannot do that with the lombok generated constructor.  The suggestion here is to manually create the constructor, even if it takes no parameters at the moment for singletons.  Not a blocker to this PR.

airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
jdpgrailsdev: We should not be using this anymore.  Instead, we should be using configuration injected by Micronaut.  If we need to change the default to true, we should do so in the `application.yaml` files for the appropriate environment.

alovew: so all of the workflows should default to false?

alovew: talked to @jdpgrailsdev separately - this was about where we're setting the env variable. I'm setting it to true in all places so we don't have potential conflicts. @benmoriceau is it ok to leave this to true like other workflows? users have a separate setting for turning off these notifications. I'm worried it might be confusing if this is the only workflow that defaults to false

benmoriceau: The notification is in the payload, so it should be ok to set it as true. I think that generally it is safer to have an opt-in strategy in OSS rather until we are confident that something is working after having it live in cloud for a little bit.

benmoriceau (approved):

Issue #21499, opened 2023-01-17
Source Postgres : Use fast select to get table size estimate in Postgres 


akashkulk:
While calculating the estimated rows to be synced for Source Postgres connector the `count()` operation is currently used. However, this is not optimal as : 
- This operation locks the table, so it cannot be performed while reading the data
- This operation can result in a full table scan (slow)

There is a faster way to get a very rough estimate : 
`select reltuples::int8 as count from pg_class c JOIN pg_catalog.pg_namespace n ON n.oid=c.relnamespace where nspname='20m_users' AND relname='users';`

However, the issue is that this is wildly incorrect (have seen it return -1) for smaller tables. We'd like to try the fast query and fall back to the slow count if it is negative. 

See discussion : https://github.com/airbytehq/airbyte/pull/20783

bleonard:
It seems like this not get the right number when incremental because we aren't fetching to full table. is that right or am I missing something?

akashkulk:
No, this query is just an estimation. So, for extremely small tables this query returns invalid results (e.g. -1 rows). 

For incremental mode, we do calculate the right number of rows and bytes based on the 
`select count(*) where cursor_field > cursor_value` query

PR #21683, opened 2023-01-20
Source Postgres : Fast query for estimate messages

akashkulk:
As part of https://github.com/airbytehq/airbyte/pull/20783, trace estimate messages were added to source-postgres. 

One of the follow-up items https://github.com/airbytehq/airbyte/issues/21499 identified in that PR was to use a fast query to calculate number of rows instead of `select count(*)` which can be slow. This was initially thought to not be a problem : see https://github.com/airbytehq/airbyte/pull/20783#issuecomment-1384698480, however we are seeing user reports of this query being slow : https://github.com/airbytehq/airbyte/pull/20783#issuecomment-1399031995. 

In this PR, we migrate to the fast query. Initially, this was held off because invalid results were returned for smaller tables. But thinking more on this, we can skip estimate traces altogether for small tables (the sync will probably complete quickly anyways). This can be determined if the fast table row count estimate is something invalid (e.g. a negative value)

The only open question here is whether we should disable estimate trace messages for non-CDC incremental mode as I can't think of another way to get around this other than to issue a `select count(*) where cursor_id > cursor_value` query, which would have the same issues as a `select count(*)`


airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresQueryUtils.java
```diff
@@ -26,7 +26,7 @@ public class PostgresQueryUtils {
 
   public static final String TABLE_ESTIMATE_QUERY =
       """
-            SELECT (SELECT COUNT(*) FROM %s) AS %s,
+            SELECT (select reltuples::int8 as count from pg_class c JOIN pg_catalog.pg_namespace n ON n.oid=c.relnamespace where nspname='%s' AND relname='%s') AS %s,
```
evantahler, line 29: Maybe both versions should be kept and done in a try/catch... I wonder if there are certain older versions of postgres that don't have a `pg_catalog` table?

evantahler (approved):
👍 from me, but a question below

evantahler (approved):

akashkulk:

akashkulk:

akashkulk:

PR #21676, opened 2023-01-20
Pass allowedHosts to container runners

evantahler:
This PR follows https://github.com/airbytehq/airbyte/pull/21363 and closes #21797, passing the AllowedHosts to the docker/K8s container executors.  This PR stops at logging the AllowedHost information as the container is started, showing that the information is being passed though the Temporal jobs properly.  Usage of this information in the K8s launcher will be done in a later story by @git-phu.

As part of this work, we needed to resolve dynamic allowed hosts from the connector's configuration.  I went with [StringSubstitutor](https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringSubstitutor.html) for this purpose.  As an example, you can see how `source-postgres` now has an allowed host limitation of `"${host}"` which is the value of "host" from the configuration of that instance of the connector.  this can be verified via the logs, as "host.docker.internal" is my configuration:

Screenshot 2023-01-20 at 3 42 29 PM

```
airbyte-worker | 2023-01-20 23:35:35 INFO i.a.w.p.DockerProcessFactory(create):122 - Creating docker container = source-postgres-read-8-0-evhis with resources io.airbyte.config.ResourceRequirements@5f1218b0[cpuRequest=,cpuLimit=,memoryRequest=,memoryLimit=] and allowed hosts io.airbyte.config.AllowedHosts@4d32a0ff[hosts=[host.docker.internal],additionalProperties={}]
```

PR #21684, opened 2023-01-21
Github Action to enforce allowed hosts are set for GA and Beta connectors

### LINKED ISSUE #21184: Test that Beta and GA connectors have `allowed_hosts`
Using our `ci_connector_ops` tools, we can enforce that Beta and Alpha connectors which do yet `allowed_hosts` in their actor definitions will fail a CI check.

evantahler:
This PR adds a check to the renamed `Connector Ops CI - Connector Metadata Checks` to ensure that GA connectors include `allowed_hosts` in their actor definition. 

Requires https://github.com/airbytehq/airbyte/pull/21676 to be merged first
Closes #21184


tools/ci_connector_ops/ci_connector_ops/allowed_hosts_checks.py
```diff
@@ -0,0 +1,51 @@
...
+        connector_release_stage = utils.get_connector_release_stage(connector_name)
+        if connector_release_stage in RELEASE_STAGES_TO_CHECK:
+          missing = not connector_has_allowed_hosts(connector_name)
+          if missing:
+            connectors_missing_allowed_hosts.append(connector_name)
+
+    return connectors_missing_allowed_hosts
+
+def connector_has_allowed_hosts(connector_name: str) -> bool:
+  definition = utils.get_connector_definition(connector_name)
```
evantahler, line 31: @alafanechere these utilities are great!

.github/workflows/connector_metadata_checks.yml
```diff
@@ -1,12 +1,12 @@
-name: Connector Ops CI - Check SAT strictness level
+name: Connector Ops CI - Connector Metadata Checks
```
evantahler, line 1: Renaming and extending our existing Github Action

tools/ci_connector_ops/setup.py
```diff
@@ -21,7 +21,8 @@
         "console_scripts": [
```
alafanechere, line 21: Would you mind bumping the package version? It's not released anywhere so versionning it a good practice only at the moment.

tools/ci_connector_ops/ci_connector_ops/allowed_hosts_checks.py
```diff
@@ -0,0 +1,47 @@
...
+  # print("----- " + connector_name  + " -----")
+  # print(definition)
```
alafanechere, line 29: You can remove these debugs comments, don't you?

tools/ci_connector_ops/ci_connector_ops/allowed_hosts_checks.py
```diff
@@ -0,0 +1,47 @@
...
+  try:
+    return definition["allowedHosts"] != None
+  except KeyError:
+    return False
```
alafanechere, line 33: More pythonic
```suggestion
    return definition.get("allowedHosts") is not None
```

alafanechere (approved):
Nice! I only have a couple of minor suggestions.
I think we might want to eventually move this to a QA check as part of `/test`. This module is ready to welcome it if you like:
https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/qa_checks.py#L71

But as it is release stage specific check I'm fine keeping it here 👍

erohmensing (approved):

PR #21685, opened 2023-01-21
Fix an npe on a sort fuction

benmoriceau:
## What
An Npe exception during the sort of the release stages was causing confusing on build. This error was not terminal and would likely pass the test. This is filtering null values before sorting.


jdpgrailsdev (approved):
:shipit:

cgardens (approved):

cgardens:
thanks @benmoriceau

PR #21686, opened 2023-01-21
[Connector Builder] Add Segment event tracking

### LINKED ISSUE #20863: Connector builder: Add event tracking
To be able to tell how users are working with the connector builder UI, there should be events for common interactions:
* Trigger stream test
* Test success / failure
* Add stream
* Remove stream
* Copy stream
* Copy component
* Enable/disable component
* Switch to YAML/UI
* Export manifest

Events can be tracked using the analytics service (`useAnalyticsService`)

lmossman:
## What
Resolves https://github.com/airbytehq/airbyte/issues/20863

Captures the events as listed in the `Track (Connector Builder)` tab of this spreadsheet: https://docs.google.com/spreadsheets/d/1lGLmLIhiSPt_-oaEf3CpK-IxXnCO0NRHurvmWldoA2w/edit#gid=1297906000
(except for the Connector Created event, since that functionality does not exist yet)

## How
Uses the AnalyticsService to send various events to Segment. I followed the existing convention of not using i18n for the action descriptions as these are not displayed to users and only used for internal tracking.

In order to test this properly, you need to sign into [Amplitude](https://analytics.amplitude.com/airbyteinc) and search for events starting with `Airbyte.UI.ConnectorBuilder`.
Here is a screenshot with a bunch of these tracked events shown in that platform (the CSV export does not actually export these events, just the single user list, and I couldn't find a great way to export these as text)
Screen Shot 2023-01-20 at 5 50 23 PM


airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/StreamTester.tsx
```diff
@@ -44,6 +52,25 @@ export const StreamTester: React.FC<{
     }
   }, [isError]);
 
+  useEffect(() => {
+    // This will only be true if the data was manually refetched by the user clicking the Test button,
+    // so the analytics events won't fire just from the user switching between streams, as desired
+    if (isFetchedAfterMount) {
+      if (errorMessage) {
+        analyticsService.track(Namespace.CONNECTOR_BUILDER, Action.STREAM_TEST_FAILURE, {
```
flash1293, line 60: This is called on every state update:
* Make an actual stream test which results in failure: works as expected
* Afterwards, change something in the UI - another failure event is recorded (even if the user didn't click on test).

This happens because `jsonManifest.streams` changes its reference on every state update.

Idea to fix this: Use the streams via ref and re-run the hook on `streamReadData`

flash1293 (changes requested):
It seems like stream test success and failure are recorded, but the user action "Stream test" is not - I guess we should add that on button click.

airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/StreamTester.tsx
lmossman: @flash1293 since we separately decided to not send the stream JSON body in these events, I no longer had the issue of this depending on `jsonManifest.streams` and re-firing for every state change. However, I did need to add a couple more changes here to have this `useEffect` be dependent on `dataUpdatedAt` and `errorUpdatedAt` from the react-query, to make the events be tracked when a user clicks Test again and the results don't change. See commit: https://github.com/airbytehq/airbyte/pull/21686/commits/09967f01a770cde0321f2f855e58f1cac96d8434

With these changes, I believe this is now working properly, but definitely lmk if you think anything else should be changed here.

lmossman:
@flash1293 I believe I've addressed the feedback here and it should be ready for another look

flash1293:
Looks like the e2e test failure is legit - as the route is part of the workspace now, it shows the signup form: 
Screenshot 2023-01-24 at 16 22 46

flash1293 (approved):
LGTM, there is one thing that probably should be adjusted but I don't think another round of review is necessary - when switching the stream view by switching the stream name in the testing panel, no event is fired, but it's essentially the same thing as switching from the left hand side bar.

PR #21687, opened 2023-01-21
🪟 🐛 Force refresh of firebase token after verifying email

joeykmh:
## What
Forces firebase to fetch a new ID token (JWT) after a user verifies their email. This new token will have `email_verified: true`, to reflect the new user status. By default, firebase will not get a new token until 5 mins before it expires, so we need to explicitly refresh it in this case.


timroes (approved):
I've tested that this will successfully reload the JWT token to actually contain email_verified: true in it after verifying the mail. There is potentially some drawback, that another window (where the user maybe already have the dialog open and clicked "resend validation mail") will not see that update immediately. I think we should though not handle that, and require in between tab communication just for that use-case.

Code LGTM

ambirdsall:
:+1:

PR #21695, opened 2023-01-23
Postgres Source : Support JSONB datatype

### LINKED ISSUE #9028: 🐛  Source Postgres: JSONB Postgres field shows as string
## Environment
- **Airbyte version**:  v0.35.0-alpha
- **OS Version / Instance**: Ubuntu
- **Deployment**: Docker
- **Source Connector and version**: 0.4.0
- **Severity**: Medium
- **Step where error happened**: Connection

## Current Behavior
In Postgres, I have a JSONB field. This shows as string in airbyte, where as I want this to be JSONB so I can have the field as VARIANT in Snowflake.

## Expected Behavior
It shows as variant.

## Steps to Reproduce
1. Create a postgres table with a JSONB column
2. It shows as string
3. It exports as string

VitaliiMaltsev:
## What
Postgres JSONB columns shows as string in destination, the field should be as VARIANT in Snowflake, SUPER in Redshift, JSONB in Postgres, etc

## How
Updated Json schema mapping to 
```
    "test_column": {
      "type": "object",
      "oneOf": [
        {
          "type": "array"
        },
        {
          "type": "object"
        },
        {
          "type": "string"
        },
        {
          "type": "number"
        },
        {
          "type": "boolean"
        }
      ]
    }
```

## Recommended reading order
1. `JsonSchemaType.java`
2. `PostgresSourceOperations.java`

## 🚨 User Impact 🚨

## Pre-merge Checklist

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>

- [ ] Code reviews completed
- [x] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>

</details>

## Tests

</details>

</details>

</details>


adam-bloom:
Just tried this out (I hit a use case end of last week that requires jsonb be handled correctly) - I'm still getting strings in the destination. Stream config is updated along with the source connector image.

VitaliiMaltsev:
> Just tried this out (I hit a use case end of last week that requires jsonb be handled correctly) - I'm still getting strings in the destination. Stream config is updated along with the source connector image.

@adam-bloom in which destinations did you get string?

adam-bloom:
> @adam-bloom in which destinations did you get string?
@VitaliiMaltsev we use redshift as our destination. I ended up putting a workaround in a custom redshift destination in the meantime, but am definitely willing to test this build again in the future. I'm on the airbyte slack as well if you want more details.

VitaliiMaltsev:
> > @adam-bloom in which destinations did you get string?
> 
> @VitaliiMaltsev we use redshift as our destination. I ended up putting a workaround in a custom redshift destination in the meantime, but am definitely willing to test this build again in the future. I'm on the airbyte slack as well if you want more details.


@adam-bloom tested Redshift destination and values replicated as SUPER datatype not String
reshift-super

adam-bloom:
> @adam-bloom tested Redshift destination and values replicated as SUPER datatype not String

What is in the column though? The column may be a super, but a super can hold any type of data (objects, arrays, strings, etc). My testing showed that the data was being stored as a string inside the super, not as an object, which limits the size of the data to 65KB. Try a large jsonb row and you should see that it is currently (correctly) dropped by the redshift destination as too large for a SUPER varchar field. Note that a SUPER has a limit of 1MB (potentially going up to 16 MB soon...docs aren't clear on this, but just noticed them starting to use 16 MB).

I'm not using normalization, but that shouldn't impact this.

VitaliiMaltsev:
> > @adam-bloom tested Redshift destination and values replicated as SUPER datatype not String
> 
> What is in the column though? The column may be a super, but a super can hold any type of data (objects, arrays, strings, etc). My testing showed that the data was being stored as a string inside the super, not as an object, which limits the size of the data to 65KB. Try a large jsonb row and you should see that it is currently (correctly) dropped by the redshift destination as too large for a SUPER varchar field. Note that a SUPER has a limit of 1MB (potentially going up to 16 MB soon...docs aren't clear on this, but just noticed them starting to use 16 MB).
> 
> I'm not using normalization, but that shouldn't impact this.

@adam-bloom i have updated values mapping 
redshift-super-values
Please pull changes from this branch and check if it works for you now

akashkulk (commented):
Can you add a test (or provide an example) for what the catalog looks like for the JSONB type?

VitaliiMaltsev:
> @VitaliiMaltsev confirmed that it is behaving as expected now. Thanks!
> 
> My workaround was a custom redshift destination build that coerced to JSON the same way :). So debezium is outputting strings for jsonb then, and not objects?

@adam-bloom tested Postgres JSONB <> Redshift Destination in both modes (STANDARD and CDC). In both cases results are the same - SUPER datatype with objects values

VitaliiMaltsev:
> Can you add a test (or provide an example) for what the catalog looks like for the JSONB type?

@akashkulk here is an example of json schema for JSONB

`    "test_column": {
      "type": "object",
      "oneOf": [
        {
          "type": "array"
        },
        {
          "type": "object"
        },
        {
          "type": "string"
        },
        {
          "type": "number"
        },
        {
          "type": "boolean"
        }
      ]
    }`
this construction allows us to have successful syncs for normalization of JDBC destinations and for S3 avro/parquet conversions

VitaliiMaltsev:
@akashkulk @edgao please advise how to update this PR to support V1 protocol as well

VitaliiMaltsev:
> @VitaliiMaltsev I agree that it works - my question was why`ObjectMapper` is even necessary - where is the json being stringified in the first place? I ask because other data types (numeric data types) are written as strings too and are required to be converted, which doesn't make much sense. I can open a separate issue for that.

@adam-bloom Got it. No, these issues are not related, we have different mapping for each datatype. Thanks for creating a separate issue

akashkulk:
> 

Postgres source doesn't currently support v1. But in v1-land, the catalog should look like this : 

```
"test_column": {
    "type": "object",
    "oneOf": [
        {
            "type": "array"
        },
        {
            "type": "object"
        },
        {
            "type": "AirbyteTypes.yaml#definitions/String"
        },
        {
            "type": "AirbyteTypes.yaml#definitions/Integer"
        },
        {
            "type": "AirbyteTypes.yaml#definitions/boolean"
        }
    ]
}
```

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
```diff
@@ -508,7 +520,12 @@ public JsonSchemaType getAirbyteType(final PostgresType jdbcType) {
       case DATE_ARRAY -> JsonSchemaType.builder(JsonSchemaPrimitive.ARRAY)
           .withItems(JsonSchemaType.STRING_DATE)
           .build();
-
+      case JSONB -> JsonSchemaType.builder(JsonSchemaPrimitive.ARRAY,
```
akashkulk, line 523: Instead of having the builder take multiple values, can you just define a JsonSchemaPrimitive.JSONB (which under the hood will build the same structure/oneof that you've been doing)?

airbyte-protocol/protocol-models/src/main/java/io/airbyte/protocol/models/JsonSchemaType.java
```diff
@@ -106,15 +110,29 @@ public static class Builder {
 
     private final ImmutableMap.Builder<String, Object> typeMapBuilder;
 
-    private Builder(final JsonSchemaPrimitive type) {
-      typeMapBuilder = ImmutableMap.builder();
-      if (JsonSchemaPrimitiveUtil.isV0Schema(type)) {
-        typeMapBuilder.put(TYPE, type.name().toLowerCase());
+    private Builder(final JsonSchemaPrimitive... types) {
+      final List<JsonSchemaPrimitive> schemaPrimitives = Arrays.asList(types);
+      if (hasMultipleTypes(schemaPrimitives)) {
```
akashkulk, line 115: And there you can check for a `JsonSchemaType.JSON` and build the same oneof structure. 

When we update to V1, I can just make the change to have the V1 versions of `JsonSchemaType.STRING, JsonSchemaType.BOOLEAN, JsonSchemaType.NUMBER`

airbyte-integrations/connectors/source-postgres/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/AbstractPostgresSourceDatatypeTest.java
```diff
@@ -253,9 +253,14 @@ protected void initTests() {
...
-            .airbyteType(JsonSchemaType.STRING)
-            .addInsertValues("null", "'[1, 2, 3]'::jsonb")
-            .addExpectedValues(null, "[1, 2, 3]")
+            .airbyteType(JsonSchemaType.builder(JsonSchemaPrimitive.ARRAY,
+                JsonSchemaPrimitive.OBJECT,
+                JsonSchemaPrimitive.STRING,
+                JsonSchemaPrimitive.BOOLEAN,
+                JsonSchemaPrimitive.NUMBER)
+                .build())
+            .addInsertValues("null", "'10000'::jsonb", "'true'::jsonb")
```
akashkulk, line 262: Can you test some more nested json objects here?

VitaliiMaltsev: > Can you test some more nested json objects here?

done

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
VitaliiMaltsev: > Instead of having the builder take multiple values, can you just define a JsonSchemaPrimitive.JSONB (which under the hood will build the same structure/oneof that you've been doing)?

@akashkulk updated with your proposal

airbyte-protocol/protocol-models/src/main/java/io/airbyte/protocol/models/JsonSchemaType.java
```diff
@@ -87,6 +90,7 @@ public class JsonSchemaType {
       JsonSchemaType.builder(JsonSchemaPrimitive.STRING)
           .withLegacyAirbyteTypeProperty("big_integer")
           .build();
+  public static final JsonSchemaType JSONB = JsonSchemaType.builder(JsonSchemaPrimitive.JSONB).build();
```
akashkulk, line 93: was talking to Edward and I think we should add a legacy_airbyte_type = "json" here for debuggability

akashkulk:
Couple of thoughts after chatting with Edward : 
1) Have you tested this against v1 normalization & platform as Edward mentioned [here](https://airbytehq-team.slack.com/archives/C043JHEEYKG/p1674786523315659?thread_ts=1674741542.369019&cid=C043JHEEYKG) : 
The upgrade to v1 has led to some issues at this point, so perhaps it's worth holding off on this until that is in?
2) Is the top-level object required? Does having the test column as ``test_column: {oneOf: [...], airbyte_type: jsonb}`` work? The way you have the test_column setup currently it seems to be kicking off normalization's unknown type logic which defaults to jsonb.

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
```diff
@@ -185,6 +185,7 @@ public void copyToJsonField(final ResultSet resultSet, final int colIndex, final
         case "path" -> putObject(json, columnName, resultSet, colIndex, PGpath.class);
         case "point" -> putObject(json, columnName, resultSet, colIndex, PGpoint.class);
         case "polygon" -> putObject(json, columnName, resultSet, colIndex, PGpolygon.class);
+        case "jsonb" -> putJsonb(json, columnName, resultSet, colIndex);
```
rodireich, line 188: Should we also take care of an array of jsonb? i.e `_jsonb`

rodireich:
> field should be as VARIANT in Snowflake, SUPER in Redshift, JSONB in Postgres

What here tells these destinations to creates these types?
What will happen for connectors that don't have a specialized json type?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
VitaliiMaltsev: > Should we also take care of an array of jsonb? i.e `_jsonb`

@rodireich added support for json[] datatype
Note: 
**STANDARD SYNC** load data to the proper JsonNodes
**CDC** load data to the String. 
_A Postgres' json[] or jsonb[] column doesn't really describe any schema. So when reading such a column, Debezium converts it in string. It means that part of message is structured, part is nested as a JSON encoded string._
Currently it is a limitation of debezium and Kafka Connect API

VitaliiMaltsev:
> > field should be as VARIANT in Snowflake, SUPER in Redshift, JSONB in Postgres
> 
> What here tells these destinations to creates these types? What will happen for connectors that don't have a specialized json type?

i believe dbt macros are responsible to put values to the proper datatypes. You can find them in [datatypes.sql](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-normalization/dbt-project-template/macros/cross_db_utils/datatypes.sql) file

airbyte-protocol/protocol-models/src/main/java/io/airbyte/protocol/models/JsonSchemaType.java
VitaliiMaltsev: > legacy_airbyte_type = "json"

added

VitaliiMaltsev:
> Couple of thoughts after chatting with Edward :
> 
> 1. Have you tested this against v1 normalization & platform as Edward mentioned [here](https://airbytehq-team.slack.com/archives/C043JHEEYKG/p1674786523315659?thread_ts=1674741542.369019&cid=C043JHEEYKG) :
>    The upgrade to v1 has led to some issues at this point, so perhaps it's worth holding off on this until that is in?
> 2. Is the top-level object required? Does having the test column as `test_column: {oneOf: [...], airbyte_type: jsonb}` work? The way you have the test_column setup currently it seems to be kicking off normalization's unknown type logic which defaults to jsonb.

@akashkulk 
1. Not yet. Will take a look
2. I agree that such a strange schema is a little confusing
This schema allows us to make successful syncs with normalization (for this top level type object is needed) and also to successfully convert avro records (for this we need an oneOf structure)

I came up with exactly this json schema through numerous tests with various destinations, since it showed the best result at the moment

Full tests results:

**Redshift** JSONB ----> SUPER 
Standart inserts - ok 
S3 staging - ok

**Snowflake** JSONB ----> VARIANT
 Internal Staging - ok A
WS S3 Staging (CSV) - ok 
GCS Staging (CSV) - ok 
Azure blob Staging (CSV) -ok

**BigQuery**  
Standart inserts JSONB ----> STRING(ok) 
GCS Staging JSONB ----> STRING(ok)

**S3**
 CSV - ok 
JSONL -ok 
AVRO - ok 
Parquet - ok

**MsSQL** JSONB ----> VARCHAR (ok)
**Postgres** JSONB ----> JSONB (ok)
**MySQL** JSONB ----> JSON (ok)

**BigQuery Denormalized** 
Standard inserts JSONB ----> STRING(failed)
GCS Staging JSONB ----> STRING(failed)

For BigQuery Denormalized created the separate [issue](https://github.com/airbytehq/airbyte/issues/21977)

rodireich (approved):

airbyte-protocol/protocol-models/src/main/java/io/airbyte/protocol/models/JsonSchemaType.java
```diff
@@ -109,12 +113,25 @@ public static class Builder {
...
+          buildJsonbSchema();
+        } else {
+          typeMapBuilder.put(TYPE, type.name().toLowerCase());
+        }
       } else {
         typeMapBuilder.put(REF, PRIMITIVE_TO_REFERENCE_BIMAP.get(type));
       }
     }
 
+    private void buildJsonbSchema() {
```
edgao, line 126: random thought - can you try with this schema?
```json
{
  "type": ["object", "array", "string", "number", "boolean"],
  "airbyte_type": "jsonb"
}
```

I think that avoids the misleading `type: object`, and maybe works better for normalization

(sorry, I didn't think of this until now)

edgao (commented):
had some questions about the source behavior itself, plus one new idea about the schema declaration

airbyte-protocol/protocol-models/src/main/java/io/airbyte/protocol/models/JsonSchemaType.java
VitaliiMaltsev: > random thought - can you try with this schema?
> 
> ```json
> {
>   "type": ["object", "array", "string", "number", "boolean"],
>   "airbyte_type": "jsonb"
> }
> ```
> 
> I think that avoids the misleading `type: object`, and maybe works better for normalization
> 
> (sorry, I didn't think of this until now)

@edgao Yes, I tried this schema but the results are worse then
For example, sync fails fwith such a schema for destination Bigquery
You can see the results with my proposed schema [here](https://github.com/airbytehq/airbyte/pull/21695#issuecomment-1414208540).

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/PostgresConverter.java
```diff
@@ -69,13 +72,28 @@ public void converterFor(final RelationalColumn field, final ConverterRegistrati
...
     }
   }
 
+  private void registerJsonb(RelationalColumn field, ConverterRegistration<SchemaBuilder> registration) {
+    registration.register(SchemaBuilder.string().optional(), x -> {
+      if (x == null) {
+        return DebeziumConverterUtils.convertDefaultValue(field);
+      }
+      try {
+        return new ObjectMapper().readTree(x.toString()).toString();
```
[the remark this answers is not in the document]
VitaliiMaltsev: @edgao replaced ObjectMapper to the global variable
I use this construction here not for json validation, but for the result to be similar for СDC and standard sync
The call here is just toString() adds unnecessary spaces to json, ObjectMapper doesn't.
For example
```
                      x.toString()                                           ===> [1, 2, 3]
ObjectMapper().readTree(x.toString()).toString()  ===> [1,2,3]
```

airbyte-integrations/connectors/source-postgres/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/AbstractPostgresSourceDatatypeTest.java
```diff
@@ -46,6 +48,19 @@ public boolean testCatalog() {
     return true;
   }
 
+  protected String getValueFromJsonNode(final JsonNode jsonNode) throws IOException {
+    if (jsonNode != null) {
+      if (jsonNode.isArray() || jsonNode.isObject()) {
+        return jsonNode.toString();
+      }
+
+      String value = (jsonNode.isBinary() ? Arrays.toString(jsonNode.binaryValue()) : jsonNode.asText());
```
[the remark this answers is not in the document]
VitaliiMaltsev: > how do we get a binary node? What does that even represent?
> 
> I'm not sure that `Arrays.toString` is quite what we want here (it'll produce a string like `"[1, 2, 3]"`), but I don't really know what the expected behavior _is_.

good point! Removed redundant binary node check

VitaliiMaltsev:
> can we put something in the cdc value converter to parse these into json nodes? I think it would be surprising for users if > we read different values based on whether we're in cdc or select * mode

@edgao Unfortunately, we cannot, this is a limitation of the Kafka Connect API that is used by Debezium, so the construction _**registration.register(SchemaBuilder.string().optional(), x -> {)**_ is used for jsonb

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/PostgresConverter.java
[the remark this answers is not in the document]
edgao: wait, what are we putting into the raw table? Is it something like this:
```json
{
  "the_jsonb_field": [1,2,3]
}
```
or more like this:
```json
{
  "the_jsonb_field": "[1,2,3]"
}
```

airbyte-protocol/protocol-models/src/main/java/io/airbyte/protocol/models/JsonSchemaType.java
edgao: was that test using `oneof: [{type: object}, ...]` or `type: [object, ...]`?

my main concern is that right now the `type: object` is explicitly false - e.g. `[1, 2, 3]` is a valid jsonb, but it won't match the json schema. Platform is planning to fail syncs with incorrect schemas, so this is putting us in a slightly awkward position

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/PostgresConverter.java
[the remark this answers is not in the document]
VitaliiMaltsev: > wait, what are we putting into the raw table? Is it something like this:
> 
> ```json
> {
>   "the_jsonb_field": [1,2,3]
> }
> ```
> 
> or more like this:
> 
> ```json
> {
>   "the_jsonb_field": "[1,2,3]"
> }
> ```

@edgao for STANDARD sync it's 
```
{
  "the_jsonb_field": [1,2,3]
}
```
for CDC 
```
{
  "the_jsonb_field": "[1,2,3]"
}
```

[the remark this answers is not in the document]
adam-bloom: As a user, this is both unexpected and problematic - it means we can’t effectively use CDC with large jsonb columns and the Redshift destination, as strings have a much shorter maximum supported size than objects. I was hoping to switch one of our large jsonb tables from full refresh to CDC soon, but sounds like that won’t actually work. (The build of the source connector based on this branch is otherwise running great!)

airbyte-protocol/protocol-models/src/main/java/io/airbyte/protocol/models/JsonSchemaType.java
VitaliiMaltsev: > was that test using `oneof: [{type: object}, ...]` or `type: [object, ...]`?
> 
> my main concern is that right now the `type: object` is explicitly false - e.g. `[1, 2, 3]` is a valid jsonb, but it won't match the json schema. Platform is planning to fail syncs with incorrect schemas, so this is putting us in a slightly awkward position

i see. In case of platform will validate every record we definitely can not use current json schema
@edgao i have updated my PR with the following json schema
```
{
  "type": ["object", "array", "string", "number", "boolean"],
  "airbyte_type": "jsonb"
}
```
**Original Postgres table**

postgres-original-json-table

## Tests results

**Postgres Destination** ✅ 
JSONB ----> JSONB 

**MySQL Destination** ✅ 
JSONB ----> JSON 

**MSSQL Destination** ❌ 
JSONB ----> VARCHAR 

**_Integers and booleans values are null_**

mssql-jsonb-int-and-bool-null

**Redshift Destination** 
JSONB ----> SUPER 
Standart inserts ✅ 
 S3 staging ✅ 

**Snowflake Destination**
JSONB ----> VARIANT 
Internal Staging ✅ 
AWS S3 Staging (CSV) ✅
GCS Staging (CSV) ✅
 Azure blob Staging (CSV) ✅

**S3 Destination**
 CSV ✅
JSONL ✅ 
AVRO ✅ 
Parquet ✅

**Oracle Destination** ❌ 
JSONB ----> VARCHAR 

**_All values are null except booleans and integers_**

oracle-jsonb-all-null-without-int-and-bool

**Clickhouse Destination** ❌ 
JSONB ----> Array(String) 

**_All values are empty arrays except original arrays_**

clickhouse-jsonb-all-values-empty-except-arrays

**BigQuery  Destination**
JSONB ----> STRING
Standard inserts ❌
 GCS Staging ❌

**_All values are null except original arrays_**

bigquery-jsonb-all-null-except-arrays

**BigQuery Denormalized Destination**
JSONB ----> STRING

 Standard inserts  ❌❌❌ **failed sync**
`Error while reading data, error message: JSON parsing error in row starting at position 0: Array specified for non-repeated field: test_column.`

GCS Staging - ❌❌❌ **failed sync**
`Caused by: com.google.cloud.bigquery.BigQueryException: Field test_column has incompatible types. Configured schema: string; Avro file: record.`

Summing up the test results, the problems that are observed in MSSQL Destination, Oracle Destination, ClickHouse Destination and BigQuery Destination are definitely normalization issues since data populated correctly into the airbyte_raw table

edgao: @evantahler you mentioned that platform will start failing syncs with schema mismatches “soon” - how soon is that? I.e. imminent doom, a couple weeks, a couple months?

from vitalli's research it looks like there's a few paths forward here:
1. Use `{type: object}` - this schema is wrong but has the best outcome in normalization today
     1. We'd need to prioritize adding a proper JSONB type to the protocol, with varying urgency based on when syncs will start failing. This has a dependency on the protocol v1 rollout, which is expected to extend into Q2.
2. Use `{type: [object, array, number, ....]}` - this is a valid schema, but has worse output
     1. Maybe we can figure out how to make this work better in normalization? But there's already a lot of stuff going on in that module right now, and I don't know if we have bandwidth

option 1 is a bit more appealing, but if the timeline is "next month" then that changes things

evantahler: Update on "soon" - we decided not to go that route, but instead, open up Sentry issues whenever we find records that don't match the schema.  That, in turn, will be funneled into on-call issues to be worked on.  TBD on the priority, but GA connectors will likely be higher priority than the rest.

This is a Q1B effort for the Connector Ops team - "Track & fix schema validation errors"

edgao: neat! Sounds like worst-case we'll have some noise in OC, which... I feel OK with, since supporting jsonb is pretty valuable. We can just silence the postgres alert when it fires. Once we have a protocol-native jsonb type, I think this largely stops mattering (since the platform will internally upgrade all messages to v2, so it shouldn't see the hacky protocol v0 schema)

@VitaliiMaltsev I think the `{type: object, oneOf: [...]}` schema is actually fine, sorry for the flip flop here

VitaliiMaltsev: @edgao reverted to json schema `{type: object, oneOf: [...]}`

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/PostgresConverter.java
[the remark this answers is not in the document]
bleonard: @VitaliiMaltsev what's the current CDC behavior? I think it's ok to move forward and make another issue if it was strings and it's still strings. That is, this only improves the beahvior. Of course, if we could make it the same all the better.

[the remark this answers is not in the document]
VitaliiMaltsev: @bleonard CDC behaviour is still strings

VitaliiMaltsev:
@akashkulk @edgao please advise if anything else should be done in scope of this issue of we can proceed with publishing/merging?

edgao (approved):
lgtm since this is strict improvement over current behavior (once tests are passing). Let's add something to the docs about the difference between standard + CDC mode.

we should submit a followup issue to improve the cdc experience as well.

Issue #21041, opened 2023-01-04
Better handle control messages while editing the connector through UI (Part 2: UI update)


evantahler:
After https://github.com/airbytehq/airbyte/issues/20913 is complete, this story is to update the front-end to re-hydrate the connector setup form if the `check_for_update` API response indicated that there was a configuration change during that check (e.g. an oAuth token has changed).  See https://github.com/airbytehq/airbyte/issues/20913 for a full description.

<img width="1295" alt="image" src="https://user-images.githubusercontent.com/4368928/209881958-1cdac252-c1f1-4ea4-941c-007f18bca74e.png">

From https://github.com/airbytehq/airbyte/issues/20913#issuecomment-1369563163: 

> Instead of returning the updated config in the response of the check_for_update call, what about sending the flag whether it updated automatically and have the frontend re-fetch the values of the form and repopulate it by calling (sources|destinations)/get (instead of showing a message to the user to reload manually)? This way no further backend changes would be needed and this kind of change is relatively simple in the frontend with our setup. This way the additional complexity of the connector updating its config during check is completely transparent to the user.

## Acceptance criteria

* If the configuration is updated on the "testing connection" call, the form is updated with the most recent data as if the user had refreshed the page

PR #21698, opened 2023-01-23
🪟🎨 Clean up connector form form controls 

### LINKED ISSUE #20967: Connector form: Refactor form controls and improve UX
The controls part of the connector form needs to be refactored in a bunch of ways:
* Remove leftover styled components
* Get rid of hardcoded pixel values for sizing (instead rely on flexbox layout)
* Restructure component hierarchy and props better to capture all different cases

Right now for the edit case there are two buttons: "Retest source" and "Save changes and test". The former one always tests the last saved version even if there have been changes to the form which is very confusing. This should be simplified to only show a single button - if there are no changes to the form, it says "Retest source", if there are changes to the form it says "Save changes and test" and acts accordingly.

In the source or destination create pages, the submit button is placed inside the card:

The change proposed is to move the button outside of the card to match the current design:

Prototype:
https://www.figma.com/proto/mFVUWYYjDm5zz6FFx8hTP6/FORMS-FIELDS?page-id=1%3A4&node-id=301%3A10567&viewport=606%2C64%2C0.16&scaling=min-zoom&starting-point-node-id=301%3A10567

In the source settings, the save and delete buttons also move. See in design:
https://www.figma.com/file/8aM4gBoW95whyshI5kwVct/01_06_SOURCES?node-id=222%3A2479

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/20967

* Remove legacy styled components
* Unify look of edit connector form and new connector form
* Make it possible to check connection without saving
* Make "retest" button test current form state instead of saved state

### Old look

Empty
Screenshot 2023-01-23 at 12 40 01

Selected connector
Screenshot 2023-01-23 at 12 40 11

Check in progress
Screenshot 2023-01-23 at 12 40 19

Error on unsaved connector
Screenshot 2023-01-23 at 12 40 26

Passed checks:
Screenshot 2023-01-23 at 12 40 35

Edit form:
Screenshot 2023-01-23 at 12 40 44

Error on edit form:
Screenshot 2023-01-23 at 12 40 52

### New look

Empty
Screenshot 2023-01-23 at 12 41 05

Selected connector
Screenshot 2023-01-23 at 12 41 15

Check in progress
Screenshot 2023-01-23 at 12 41 20

Error on unsaved connector
Screenshot 2023-01-23 at 12 41 24

Passed checks:
Screenshot 2023-01-23 at 12 41 31

Edit form:
Screenshot 2023-01-23 at 12 41 36

Error on edit form:
Screenshot 2023-01-23 at 12 41 42

## How
* To pull the controls out of the card, the ConnectorForm had to be changed:
  * Do not render the card as part of `ConnectorCard` anymore (move it down to the `FormRoot` level)
  * Pass down the connector definition select as `headerBlock`, so it can be rendered inside the card within `FormRoot`
  * Introduce a `renderFooter` render prop which is responsible for rendering the form controls:
    * in case of the connector card, it is rendering the testing card as well as the buttons for saving/deleting/canceling (depending on whether it's editing or a new actor)
    * in case of connector builder, it's rendering reset, cancel and save buttons
* As the connector form is always rendered now even if there is no definition selected yet, the selector specification definition gets optional (which needs to be handled throughout the component)
* Custom styling of various elements (success message, error message, spacings) are replaced by standard components (flex container and item, text, callout)


airbyte-webapp/src/views/Connector/ConnectorCard/components/TestCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+            >
+              <FormattedMessage id="form.cancel" />
+            </Button>
+          ) : (
+            <Button
+              type="button"
+              onClick={onRetestClick}
+              variant="secondary"
+              icon={<FontAwesomeIcon icon={faRefresh} />}
+              disabled={!isValid || (!isEditMode && !dirty)}
```
lmossman, line 103:
```suggestion
              disabled={!isValid || (isEditMode && !dirty)}
```
I think this was a mistake. The current logic means that if you set up a new connector and don't change any of the values (e.g set up an E2E Testing destination and don't change any of the defaults), then it won't let you test or submit.

airbyte-webapp/src/views/Connector/ConnectorCard/components/Controls.tsx
```diff
@@ -0,0 +1,73 @@
...
+              <FormattedMessage id={`tables.${formType}Delete`} />
+            </Button>
+          )}
+        </FlexItem>
+        {isEditMode && (
+          <Button type="button" variant="secondary" disabled={isSubmitting || !dirty} onClick={onCancelClick}>
+            <FormattedMessage id="form.cancel" />
+          </Button>
+        )}
+        <Button type="submit" disabled={isSubmitting || !dirty}>
```
lmossman, line 63: You'll need to fix this `disabled` logic in a similar way, so that if we arent in edit mode, the form can still be submitted even if its not dirty (see above comment)

lmossman (changes requested):
Left some comments, main changes requested are about issues I found when locally testing.

Overall this is looking great though!

airbyte-webapp/src/views/Connector/ConnectorForm/ConnectorForm.tsx
```diff
@@ -24,17 +24,31 @@ export interface ConnectorFormProps {
...
   connectionTestSuccess?: boolean;
   errorMessage?: React.ReactNode;
-  successMessage?: React.ReactNode;
   connectorId?: string;
   footerClassName?: string;
   bodyClassName?: string;
   submitLabel?: string;
+  title?: React.ReactNode;
+  description?: React.ReactNode;
+  full?: boolean;
```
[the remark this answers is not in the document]
flash1293: I specified this in the types using a union of an interface for props with `renderWithCard: true` and title/description/full and one with `renderWithCard?: false` without them - we use the same pattern for destinations vs. sources.

Also while doing this I deduped the type between form root and connector form and removed a bunch of stuff that's not used anymore.

airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/ConfigMenu.tsx
```diff
@@ -112,13 +113,35 @@ export const ConfigMenu: React.FC<ConfigMenuProps> = ({ className, testInputJson
...
                     setIsOpen(false);
                   }}
+                  renderFooter={({ dirty, isSubmitting }) => (
+                    <div className={styles.inputFormModalFooter}>
+                      <FlexContainer>
+                        <FlexItem grow>
+                          <Button
+                            onClick={() => {
+                              setTestInputJson({});
+                            }}
```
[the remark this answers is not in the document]
flash1293: Good catch, it needs to call `resetConnectorForm` which is passed in as well, forgot about that.

airbyte-webapp/src/views/Connector/ConnectorCard/components/TestCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+              <Button
+                variant="clear"
+                type="button"
+                icon={<FontAwesomeIcon icon={logsVisible ? faChevronDown : faChevronRight} />}
+                onClick={() => {
+                  setLogsVisible(!logsVisible);
+                }}
+              >
+                <FormattedMessage id="connector.testLogs" />
+              </Button>
```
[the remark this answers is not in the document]
flash1293: This is existing behavior but I agree it doesn't make sense. Unified the connection check behavior on submit vs. the test button (we need to set the `errorStatusRequest` state)

airbyte-webapp/src/views/Connector/ConnectorCard/components/TestCard.tsx
flash1293: Good catch, actually it's even simpler - it should always be possible to re-test as long as the form is valid right now.

airbyte-webapp/src/views/Connector/ConnectorCard/components/Controls.tsx
flash1293: Adjustted

flash1293:
Thanks for the review @lmossman , addressed your points

lmossman (approved):
Tested out the changes and it looks like all of the things I pointed out have been fixed 👍

flash1293:
I had to make a change because I realized we can't use the updated config values when testing an existing source/destination because with https://github.com/airbytehq/airbyte/issues/21041 the check_for_update call can potentially update the actor in the database.

This is what I did for now (but we should probably re-visit this in the future):
* Make the test button test the saved config again (reverting to the existing behavior)
* Adjust the copy of the button to make it clearer what's happening
* Disable the button when the form is dirty to avoid people thinking they test the updated state

lmossman:
@flash1293 to confirm my understanding here:
1. We cannot call the `scheduler/sources/check_connection` endpoint (i.e. the one we use when testing new connectors) for testing changes to existing connectors, because we don't have the actual secret values in the webapp, as they are saved to the database
2. We cannot call the `check_connection_for_update` endpoint for testing changes to existing connectors, because if there is a control message returned then that could result in the server writing the unsaved form values to the actor database record

If the above is correct, then I think the changes you made make sense to me as a **temporary** solution.

However, it still feels pretty bad to me that users have no way of testing changes to an existing connector without saving those changes. To solve this, it seems like we need a new API endpoint, or a boolean param added to `check_connection_for_update`, which causes the API to just run the connection check with the provided values, ignoring control messages and not updating the actor config in the database. Since that is exactly what we want to be able to do with the `Test` button in the UI.

cc @evantahler @pedroslopez to get your thoughts on the above

lmossman (approved):
Approving the current state of the PR as a good intermediate solution (see my comment above for my thoughts on follow-up changes here). 

> Also, this change makes me think that we might just want to hide the testing card for new actors and only pop them in if the user tries to save

I think the current solution is good, since as a user I like being able to test my configuration before I commit to actually saving it.

Tested locally and it works well. There is a merge conflict that needs to be fixed before merging.

pedroslopez:
> However, it still feels pretty bad to me that users have no way of testing changes to an existing connector without saving those changes. To solve this, it seems like we need a new API endpoint, or a boolean param added to check_connection_for_update, which causes the API to just run the connection check with the provided values, ignoring control messages and not updating the actor config in the database. Since that is exactly what we want to be able to do with the Test button in the UI.

In general I agree that it's _weird_ that invalid configurations could potentially be saved. However, this comes up because of single-use refresh tokens, which are part of the config. Once that refresh token is used, the previous config would be invalidated. Because of the way the UI worked (I haven't reviewed the actual changes in this PR), you could get into a situation where the check runs, refreshes tokens, and then the UI would call an "update" with the (invalid) values that were in the form. The problem is described more in https://github.com/airbytehq/airbyte/issues/20913.

To at least make it clearer to the user we introduced a `connectorConfigurationUpdated` bool in the response on https://github.com/airbytehq/airbyte/pull/21466 so that we could at least show the updated values to the user. 

We could add an input to the check that avoids persisting config updates, but this would have to be on some separate button so that we make sure your credentials are valid and properly update refresh tokens if necessary on the check that runs when saving.

flash1293:
@pedroslopez thanks for taking a look!

> Because of the way the UI worked (I haven't reviewed the actual changes in this PR), you could get into a situation where the check runs, refreshes tokens, and then the UI would call an "update" with the (invalid) values that were in the form.

Yes, this is handled on a separate PR

> We could add an input to the check that avoids persisting config updates, but this would have to be on some separate button so that we make sure your credentials are valid and properly update refresh tokens if necessary on the check that runs when saving.

There is a separate button in the UI for this:
Screenshot 2023-01-27 at 11 14 59

I can open a separate issue for this, I don't think it's a super pressing task though - the UI worked like this probably since the beginning.

flash1293:
@pedroslopez created an issue here: https://github.com/airbytehq/airbyte/issues/21960

PR #21699, opened 2023-01-23
CI connector ops: introduce QA checks in /test

### LINKED ISSUE #21694: Define where and how to declare QA tests for connectors
Find the best place to add new QA tests that should assert things like:
- Changelog was updated
- Documentation file is available
- Icon is available
...

It should be triggerable by the /test command, but not part of SAT (because SAT is source only). Failed QA tests should lead to a failed build in the build reports like in https://dnsgjos7lj2fu.cloudfront.net/tests/summary/source-pokeapi/

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/21694
I want to run QA checks during the integration test workflow run.
These checks will run on `/test` and nightly builds.
Their failure will lead to a build failure in the CI. 
They are not meant to be run manually by connector developers to test actual connector behavior.
They are run in the CI to asses whether a connector meets all the QA requirements to join our connector catalog.

## How
* Introduce the `run-qa-checks` commands in `ci-connector-ops` python package
* Create empty functions returning True for all the planned checks
* Run the `run-qa-checks` command in https://github.com/airbytehq/airbyte/blob/7f5e5d9e851b50edc90b2230bc6458097c6a41c6/tools/bin/ci_integration_test.sh#L35
* Discard the use of virtualenv in https://github.com/airbytehq/airbyte/blob/7f5e5d9e851b50edc90b2230bc6458097c6a41c6/.github/workflows/test-command.yml#L1
They are not required and their absence simplifies the availability of `run-qa-checks` command in path.

## 🚨 User Impact 🚨
None: all the checks are forced to successful and yet to be implemented in future PRs.


alafanechere:
An example of a `/test` execution on this branch can be found [here](https://github.com/airbytehq/airbyte/actions/runs/3986148089/jobs/6834406913). Check the head of the `Test connectors/source-facebook-marketing` logs.

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
```diff
@@ -0,0 +1,93 @@
...
+    connector_name = sys.argv[1]
+    qa_check_results = []
+    print(f"Running QA checks for {connector_name}")
+    for qa_check in QA_CHECKS:
+        successful_check = qa_check(connector_name)
+        check_result_prefix = "✅" if successful_check else "❌"
+        print(f"{check_result_prefix} - {qa_check.__name__}")
+        qa_check_results.append(successful_check)
+    if not all(qa_check_results):
+        print(f"QA checks failed for {connector_name}")
```
erohmensing, line 89: Non-blocking, but ere, I find the naming of `successful_check` a bit confusing, and we're only storing the boolean result, so we have to rely on the print statements to interpret which one failed. WDYT about a dictionary of check_results? `{"check_name": bool}` e.g. `{"check_connector_icon_is_available": True, "check_changelog_entry_is_updated": False}`

erohmensing (commented):
> These checks will run on /test and nightly builds.
Their failure will lead to a build failure in the CI.
They are not meant to be run manually by connector developers to test actual connector behavior.
They are run in the CI to asses whether a connector meets all the QA requirements to join our connector catalog.

Just wondering where we plan to run/consume this info in the future - since we're tying it to a `/test` failure; if we're using it to remove connectors from cloud, does that also then mean we'd remove them if their integration tests fail, even if QA checks are passing?

This is my only open question re: approval

alafanechere:
> Just wondering where we plan to run/consume this info in the future - since we're tying it to a /test failure; if we're using it to remove connectors from cloud, does that also then mean we'd remove them if their integration tests fail, even if QA checks are passing?

These new checks will be run on each `/test` command dispatch. This slash command is run manually from PR or during the nightly builds. 
These new checks are not meant to decide if a connector gets added or removed from Cloud (this will be a different action based on these test results and other logic).
The role of this PR is only to add a new set of QA checks that are run on connectors built in the CI that will make this build fail if the QA requirements are not met.

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
alafanechere: I improved it as you suggested 😄

erohmensing (approved):
> These new checks are not meant to decide if a connector gets added or removed from Cloud (this will be a different action based on these test results and other logic).

Got it - that was the clarification I needed!

PR #21703, opened 2023-01-23
[Low Code CDK] surface the resolved manifest in the CDK

clnoll:
This is the CDK update portion of https://github.com/airbytehq/airbyte/pull/21565; the CDK update needs to be published before the connector builder server CI will pass.


maxi297 (approved):
LGTM! And out of curiosity, why not update the connector builder server in the same PR?

clnoll:
@maxi297 the connector builder server CI tests run against the latest published version of the CDK. Since the CB changes rely on changes to the CDK, CB tests will fail until the new version of the CDK is published. To make CDK & CB changes in one PR, I'd have to publish the CDK off of the same branch that CB changes are in, then rerun the CI tests once publishing is done, and then merge the code. Since the new workflow is 1) merge to master, then 2) publish the new CDK version, I don't think that will plan will work anymore. Even if it does still work, it would mean publishing the CDK despite getting a red build. Although the test failures are due to the CB and not the CDK, that still makes me a bit uncomfortable.

PR #21708, opened 2023-01-23
Connector builder server: Pin cdk version

flash1293:
This is a small adjustment of the dependencies of the connector builder server. Right now the connector builder server is implicitly using the latest version of the CDK. This can lead to broken CI on master as the connector builder server is not automatically rebuilt if the CDK changes, so potential breakages would not be caught in time.

This PR pins the cdk dependency down to the patch version so updating it will touch the connector builder server code and run its tests to make sure the upgrade will actually work.

Another approach would be to automatically test the connector builder server with the current state of cdk as part of CI on the PR itself, but this has several disadvantages:
* Higher effort to set up in the current system
* Requires an engineer working on the cdk to fix all problems of the connector builder on the same PR instead of decoupling these streams of work

Just putting this into place for now, but happy to discuss and revisit later on.


clnoll (approved):
LGTM @flash1293 - makes sense that we'd want to do this so we can be more careful as we roll out updates to the connector builder.

PR #21725, opened 2023-01-23
Checkboxes in dropdowns now correctly visually update when clicked

### LINKED ISSUE #21707: PathPopout dropdown checkboxes not working as expected
PathPopout is not behaving as expected for a multiselect

([Loom](https://www.loom.com/share/e38c32c703f24ff196278c6dea40ed14))

- I have to click the label to get it to actually check/uncheck
- Clicking a checked checkbox unselects but then reselects it (no visual hint of this, but if you look at the order of the primary key it seems this is what's happening)
- Clicking an unchecked checkbox does nothing

krishnaglick:
## What
Resolves #21707
Checkboxes in the select dropdown triggered logic, but did not cause a visual change.

## How
preventDefault causes the checkboxes to stop toggling themselves and to instead only toggle based on the data.


teallarson (approved):
tested locally 👍

Issue #21208, opened 2023-01-10
Actor Catalog Fetch Event Duplicate Key exception


michelgalle:
## Environment
- **Airbyte version**:  0.40.26
- **OS Version / Instance**: AMI 2 / AWS EC2
- **Deployment**: Docker
- **Step where error happened**: After Setting up a connector

## Current Behavior
After setting up a connector (unfortunately I don't have the logs nor remember which connector it was), I started getting the following error in the ui:
![image](https://user-images.githubusercontent.com/25638577/211614386-ac6975a0-7e56-4385-bc65-17d1a255fa85.png)

## Expected Behavior
The ui behave as expected.

## Logs
The only logs I can get are the ones posted on a slack [thread](https://airbytehq.slack.com/archives/C021JANJ6TY/p1671745430683249).

```
2022-12-22 21:49:01 ERROR i.a.s.a.ApiHelper(execute):28 - Unexpected Exception
java.lang.IllegalStateException: Duplicate key b601e028-e003-4290-9acc-5fc72ca44776 (attempted merging values io.airbyte.config.ActorCatalogFetchEvent@50671700[id=<null>,actorId=b601e028-e003-4290-9acc-5fc72ca44776,actorCatalogId=29181719-0c50-4489-b5aa-03c0e10cb227,configHash=<null>,connectorVersion=<null>,createdAt=1671739270] and io.airbyte.config.ActorCatalogFetchEvent@69081582[id=<null>,actorId=b601e028-e003-4290-9acc-5fc72ca44776,actorCatalogId=d664d0dd-2ded-4a9e-9737-fba55e1e79eb,configHash=<null>,connectorVersion=<null>,createdAt=1671739270])
```

## Steps to Reproduce
1. Create a new ingest.
2. During schema discovery somehow airbyte executed/added to the metadata db 2 rows to `actor catalog fetch event` with the same timestamp. Not sure why it happened.

[...]

PR #21726, opened 2023-01-23
fix query to get most recent actor catalog fetch event for some sources

mfsiega-airbyte:
## What
Fix the query to get the most recent actor catalog fetch event for some sources.

This query is both inefficient -- fails to apply the source id filter -- and possibly incorrect -- if two rows exist with the same `actor_id, actor_catalog_id, created_at` values, it will crash. This is unlikely since `created_at` is a timestamp, but the frontend actually ends up issuing two simultaneous discovery workflows, so it ends up happening more often than you'd expect. See https://github.com/airbytehq/airbyte/issues/21208.

## How
- Use `row_number()` instead of `rank()`, which handles duplicates.
- Applies `sourceId` filter.


benmoriceau (approved):

mfsiega-airbyte:
@benmoriceau just FYI - I noticed that the query was choking on an empty list of source ids. Short-circuiting seemed simplest, do you mind taking another look?

benmoriceau (approved):

Issue #21009, opened 2023-01-04
CDC : Emit state messages more frequently 


subodh1810:
Currently in CDC mode for all the databases we emit a state message at the end of the sync. With checkpointing coming in the destination connectors, we should change our CDC  implementation to emit state messages more frequently so that we get the benefits of checkpointing.

bleonard:
Let's emit our LSN state when either of the follow occurs
1. It's been 10 minutes since we last did it
2. ~1000 (you pick) have been emitted.

Consider doing it in the heartbeat for the timing component, but don't overcomplicate things if that's not easy.

prateekmukhedkar:
Plan to release this on week of April 17.

PR #21727, opened 2023-01-23
feat: Add CDC sync checkpointing based on time or records

sergio-ropero:
This PR adds new functionality that generates checkpoints when doing CDC synchronisation: #21009

To doing so, instead of creating a new iterator on top of the current one, we are just changing the way we process the events.
We iterate over `ChangeEvent`s and we figure out:
- The record is not part of the snapshot
- The `lsn` of the record is *higher* than the one read by Debezium
- There has been sent +10000 records or has past more than 15 minutes since last state message (values that can change)

If all the previous conditions are met, a new state message is going to be sent in the next iteration. If it's also the end of the iterator, the state message is sent only once.

⚠️⚠️⚠️⚠️⚠️⚠️
It's possible that **there will NOT be a checkpoint save every X records/seconds** as expected.
The process needs to validate another conditions before setting sending the `STATE` message (like is not snapshot load, or the record we are publishing is actually after the offset we are going to send).
⚠️⚠️⚠️⚠️⚠️⚠️


sergio-ropero:
source-alloydb-strict-encrypt

sergio-ropero:
source-alloydb

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
```diff
@@ -0,0 +1,169 @@
...
+
+    if (sendCheckpointMessage) {
+      AirbyteMessage stateMessage = createStateMessage(checkpointOffset);
+      initializeCheckpointValues();
+      return stateMessage;
+    }
+
+    if (changeEventIterator.hasNext()) {
+      try {
+        ChangeEvent<String, String> event = changeEventIterator.next();
```
subodh1810, line 121: This can be final

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
```diff
@@ -0,0 +1,169 @@
...
+        recordsLastSync++;
+
+        if (checkpointOffset == null &&
+            (recordsLastSync >= syncCheckpointRecords ||
+                Duration.between(dateTimeLastSync, OffsetDateTime.now()).compareTo(syncCheckpointDuration) > 0)) {
+          checkpointOffset = offsetManager.read();
+        }
+
+        if (checkpointOffset != null
```
subodh1810, line 130: Let's re-arrange this as 
```
if (checkpointOffset != null
            && changeEventIterator.hasNext()
            && !cdcStateHandler.isSnapshotEvent(event)
            && cdcStateHandler.isRecordBehindOffset(checkpointOffset, event)) {
          sendCheckpointMessage = true;
        }

```

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
```diff
@@ -0,0 +1,169 @@
...
+            && cdcStateHandler.isRecordBehindOffset(checkpointOffset, event)
+            && !cdcStateHandler.isSnapshotEvent(event)
+            && changeEventIterator.hasNext()) {
+          sendCheckpointMessage = true;
+        }
+
+        return DebeziumEventUtils.toAirbyteMessage(event, cdcMetadataInjector, emittedAt);
+      } catch (final Exception e) {
+        LOGGER.error("Message iterator failed to read next record. {}", e.getMessage());
```
subodh1810, line 139: We should not swallow exception. 
`throw new RuntimeException(e);`

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    isSyncFinished = true;
+    return createStateMessage(null);
+  }
+
+  /**
+   * Initialize or reset the checkpoint variables.
+   */
+  private void initializeCheckpointValues() {
```
subodh1810, line 150: Can we call this method `resetCheckpointValues`

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * This class encapsulates an iterator and adds the required functionality to create checkpoints for
+ * CDC replications. That way, if the process fails in the middle of a long sync, the process is
+ * able to recover for any acknowledged checkpoint in the following syncs.
+ */
+public class DebeziumStateDecoratingIterator extends AbstractIterator<AirbyteMessage> implements Iterator<AirbyteMessage> {
+
+  public static final Duration SYNC_CHECKPOINT_DURATION = Duration.ofMinutes(10);
+  public static final int SYNC_CHECKPOINT_RECORDS = 1000;
```
subodh1810, line 30: I think the default should be 10K records and 15 minutes duration

subodh1810 (commented):
We need to edit the config to add custom values for frequency so that we can edit the tests and assert values

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
sergio-ropero: That was my first idea for the name but as it's also used in initialisation, I came out with that name. As I don't have strong feelings, just changing to reset :)

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
```diff
@@ -0,0 +1,179 @@
...
+      try {
+        final ChangeEvent<String, String> event = changeEventIterator.next();
+        recordsLastSync++;
+
+        if (checkpointOffset.size() == 0 &&
+            (recordsLastSync >= syncCheckpointRecords ||
+                Duration.between(dateTimeLastSync, OffsetDateTime.now()).compareTo(syncCheckpointDuration) > 0)) {
+          checkpointOffset.putAll(offsetManager.read());
+          if (!previousCheckpointOffset.isEmpty() &&
+              cdcStateHandler.isSameOffset(checkpointOffset, previousCheckpointOffset)) {
```
rodireich, line 130: This can happen if 15 minutes passed with no record?

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
```diff
@@ -0,0 +1,179 @@
...
+            checkpointOffset.clear();
+          }
+        }
+
+        if (checkpointOffset.size() == 1
```
rodireich, line 135: Can you explain what is this checkpointOffset

airbyte-integrations/connectors/source-mssql/src/main/java/io/airbyte/integrations/source/mssql/MssqlCdcStateHandler.java
```diff
@@ -55,4 +57,22 @@ public AirbyteMessage saveStateAfterCompletionOfSnapshotOfNewStreams() {
     throw new RuntimeException("Snapshot of individual tables is not implemented in MSSQL");
   }
 
+  @Override
+  public boolean isSnapshotEvent(ChangeEvent<String, String> event){
```
rodireich, line 61: suggestion: implement as `default` interface method in `CdcStateHandler`

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
sergio-ropero: I will add some comments on the variable and probably it's easier to rename to `checkpointOffsetToSend` so it's clearer that it's the one we can't to propagate.

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
sergio-ropero: What do you mean? If there are no records since we started the sync, both `previousCheckpointOffset` and `checkpointOffset` will be at the same offset (in both cases we are assigning the value `offsetManager.read()`).

airbyte-integrations/connectors/source-mssql/src/main/java/io/airbyte/integrations/source/mssql/MssqlCdcStateHandler.java
sergio-ropero: Added the new functions as `default` so only `source-postgres` is overriding them 😄

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
rodireich: You answer what I meant I think. If no records since last sync then `cdcStateHandler.isSameOffset(checkpointOffset, previousCheckpointOffset)` is `true`.

What I'm also thinking, in case of a very big db where almost no records relate to the stream we sync, there might pass 15 minutes of reading CDC logs without any single relevant ChangeEvent coming (all from tables we do not sync)

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
rodireich: Thank you. I get it now.
Looking good

rodireich (approved):

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
sergio-ropero: In that case, it will be trying to get the offset from the records included in the stream to sync... otherwise it will just set the final offset, but I got your point. BTW, if that's the case... the client should re-think how they organise their data 🤣 

I'm not sure if we can play with the heartbeat one.. 🤔

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/AirbyteDebeziumHandler.java
```diff
@@ -113,30 +106,20 @@ public AutoCloseableIterator<AirbyteMessage> getIncrementalIterators(final Confi
...
-
-    // wrap the supplier in an iterator so that we can concat it to the message iterator.
-    final Iterator<AirbyteMessage> stateMessageIterator = MoreIterators.singletonIteratorFromSupplier(stateMessageSupplier);
-
-    // this structure guarantees that the debezium engine will be closed, before we attempt to emit the
-    // state file. we want this so that we have a guarantee that the debezium offset file (which we use
-    // to produce the state file) is up-to-date.
-
-    return AutoCloseableIterators.concatWithEagerClose(messageIterator, AutoCloseableIterators.fromIterator(stateMessageIterator));
+    Integer sync_checkpoint_seconds = config.get("sync_checkpoint_seconds") != null ? config.get("sync_checkpoint_seconds").asInt()
```
subodh1810, line 109: this can be final, also the variable naming convention we follow is camel case so `syncCheckpointSeconds`

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/AirbyteDebeziumHandler.java
```diff
@@ -113,30 +106,20 @@ public AutoCloseableIterator<AirbyteMessage> getIncrementalIterators(final Confi
...
+        : DebeziumStateDecoratingIterator.SYNC_CHECKPOINT_SECONDS;
+    Integer sync_checkpoint_records = config.get("sync_checkpoint_records") != null ? config.get("sync_checkpoint_records").asInt()
```
subodh1810, line 111: final, `syncCheckpointRecords`

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
```diff
@@ -0,0 +1,192 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class encapsulates an iterator and adds the required functionality to create checkpoints for
+ * CDC replications. That way, if the process fails in the middle of a long sync, the process is
+ * able to recover for any acknowledged checkpoint in the following syncs.
+ */
+public class DebeziumStateDecoratingIterator extends AbstractIterator<AirbyteMessage> implements Iterator<AirbyteMessage> {
+
+  public static final Integer SYNC_CHECKPOINT_SECONDS = 15 * 60;
```
subodh1810, line 27: Can we mark it as a Duration? IT makes things simpler to understand.
`public static final Duration SYNC_CHECKPOINT_SECONDS = Duration.ofMinutes(15);`

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
```diff
@@ -0,0 +1,192 @@
...
+   * threshold defined by {@code SYNC_CHECKPOINT_RECORDS}.
+   * <p/>
+   * 2. Time between checkpoints ({@code dateTimeLastSync}) is higher than a {@code Duration} defined
+   * at {@code SYNC_CHECKPOINT_SECONDS}.
+   * <p/>
+   */
+  private final Duration syncCheckpointDuration;
+  private final Integer syncCheckpointRecords;
+  private OffsetDateTime dateTimeLastSync;
+  private Integer recordsLastSync;
```
subodh1810, line 56: Lets use Long just to be safe

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
```diff
@@ -0,0 +1,192 @@
...
+   * @return {@link AirbyteStateMessage} containing CDC data or state checkpoint message.
+   */
+  @Override
+  protected AirbyteMessage computeNext() {
+    if (isSyncFinished) {
+      return endOfData();
+    }
+
+    if (sendCheckpointMessage) {
+      AirbyteMessage stateMessage = createStateMessage(checkpointOffsetToSend);
```
subodh1810, line 126: final AirbyteMessage stateMessage

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresCdcStateHandler.java
```diff
@@ -61,4 +63,46 @@ public AirbyteMessage saveStateAfterCompletionOfSnapshotOfNewStreams() {
     return new AirbyteMessage().withType(Type.STATE).withState(stateMessage);
   }
 
+  @Override
+  public boolean isSnapshotEvent(ChangeEvent<String, String> event){
```
subodh1810, line 67: final ChangeEvent<String, String> event

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresCdcStateHandler.java
```diff
@@ -61,4 +63,46 @@ public AirbyteMessage saveStateAfterCompletionOfSnapshotOfNewStreams() {
...
+
+    String offset_lsn = offsetJson.get("lsn_commit") != null ?
+        String.valueOf(offsetJson.get("lsn_commit")) :
+        String.valueOf(offsetJson.get("lsn"));
+    String event_lsn = String.valueOf(Jsons.deserialize(event.value()).get("source").get("lsn"));
+    return Integer.parseInt(event_lsn) > Integer.parseInt(offset_lsn);
+  }
+
+  @Override
+  public boolean isSameOffset(Map<String, String> offsetA, Map<String, String> offsetB) {
```
subodh1810, line 88: final Map<String, String> offset, final ChangeEvent<String, String> event

subodh1810 (approved):
Most of it looks good. I just have some styling comment. We try to declare all the variables and parameters `final` if possible. I also see that the build is currently failing cause of formatting issues (you will need to run the formatter before pushing). 

But once you address the comments and the build is green, this is good to go

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/AirbyteDebeziumHandler.java
sergio-ropero: My bad, modifying!

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
sergio-ropero: Ok, I had changed also how we set up the value as well, so then we fit all parameters with `Duration` object type:
```
    final Duration syncCheckpointSeconds = config.get("sync_checkpoint_seconds") != null ? Duration.ofSeconds(config.get("sync_checkpoint_seconds").asInt())
        : DebeziumStateDecoratingIterator.SYNC_CHECKPOINT_SECONDS;
```

sergio-ropero:

sergio-ropero:

sergio-ropero:
Manually bumped `source-postgres` connector.

PR #21731, opened 2023-01-23
Uncomment auto detect schema

alovew:
JobsConfig is removed from ConfigFetchActivityImpl, so auto detect schema can be turned back on


jdpgrailsdev (approved):
:shipit:

PR #21733, opened 2023-01-23
🪟 🐛 Hide anonymize data field on preference page

### LINKED ISSUE #21053: Disambiguate Telemetry Opt-out
## Environment
- **Airbyte version**:  v0.40.26
- **OS Version / Instance**:  AWS EC2
- **Deployment**: Docker
- **Step where error happened**: Setup

## Current Behavior
Following the instructions on [this post](https://discuss.airbyte.io/t/can-i-disable-analytics-in-airbyte/74 ), apparently no telemetry is sent. However, one is still greeted to this screen on login:

image

With anonymisation remaning as an option. It's not clear why this option remains if no communication is sent back to Airbyte's servers anyway. 

## Expected Behavior
Either the radio button for anonymisation is left out or a message appears stating that no telemetry is collected.

timroes:
## What

Fixes https://github.com/airbytehq/airbyte/issues/21053

This will make sure to hide the "anaonymize data" checkbox in the initial setup screen on OSS, if logging via Segment has actually be turned of completely, to not confuse the user.

We set the `anonymizeDataCollection` flag in the workspace to `true` in this case, so in case the user ever enabled logging later (by removing the corresponding `TRACKING_STRATEGY` from their `.env` file), they'll still have anonymized data collection, since we never gave them the choice for it otherwise. (I imagine that to anyway be a very edge-case that a user turns it on initially and then turn it off later after having gone through preference.)

**Best to reviewed with "ignore whitespaces".**


teallarson (approved):
👍

Is there anything to take care of with regards to workspaces that were set up before this change and had it default to false?

joeykmh (approved):
LGTM 👍 Agreed that enabling logging later and wanting to change anonymization settings would be an extreme edge case.

timroes:
> Is there anything to take care of with regards to workspaces that were set up before this change and had it default to false?

@tealjulia Workspaces that were setup beforehand even had the chance to select this flag (even if they had the segment integration off altogether), so I think we're fine here, with whatever they chose.

Issue #21732, opened 2023-01-23
Extract the shared code present in the airbyte-server module to a common module


benmoriceau:
## What
Currently the airbyte-server module contains code that is specific to the OSS project as well as code like the handlers and the helpers. Those could be used in the other server present in cloud like what happened in: https://airbytehq-team.slack.com/archives/C044QMHSLKT/p1674159550366339

This will also help us to remove the `@Require` annotation on the OSS controllers

PR #21736, opened 2023-01-23
Extract split the server into 2 modules

benmoriceau:
## What
Isolate the micronaut application of the server. No other module will need to to depends on the server.

This will break cloud build and OSS need to be bump once this is merged

## How
Move the common class of the server to a common module.
Everything is moved except:
- The micronaut controller (api module)
- the micronaut factories (config module)
- The request handlers and mappers (errors module), the exceptions themself are moved to the common module.

Related issue: https://github.com/airbytehq/airbyte/issues/21732


gradle.properties
```diff
@@ -3,7 +3,7 @@
 
 org.gradle.parallel=true
 
-org.gradle.jvmargs=-Xmx4g -Xss4m
+org.gradle.jvmargs=-Xmx8g -Xss4m
```
benmoriceau, line 6: This was needed to make gradle to be able to import the new module

benmoriceau:
https://github.com/airbytehq/airbyte-cloud/pull/4161 => Cloud PR

gradle.properties
jdpgrailsdev: Very strange.  It seems odd that just moving the code to a library would cause Gradle to need more memory.  We should probably look into this a bit, because needing 8 GB to build the server seems like that may be an issue for OSS users.

jdpgrailsdev (approved):
:shipit:

gradle.properties
benmoriceau: I agree, I ran into java heap memory issue. I cleaned the cache which didn't work then re-clone the project which ran into the same issue, it was the only way I found to include the new module. I haven't tried with a lower memory setting.

PR #21741, opened 2023-01-23
authenticate when calling setup-python

c-p-b:
## What
We have two problems currently

1. We call setup-python 9 times in our airbyte ci workflow. That's a LOT of downloading of basically the same files over and over
2. setup-python auth's using GITHUB_TOKEN[ if you don't specify a token](https://github.com/actions/setup-python/blob/a6eba85bba13ec8929458742a9c19803934c4340/docs/advanced-usage.md?plain=1#L545). This is causing rate limit issues. We want to instead use a PAT to auth which has higher rate limits.

## How
Solve 2 because it's a quick fix. Acknowledge 1 needs to be solved either by caching or some other solution. Other setup-platform actions like setup-node have fallbacks to hit other places but it doesn't look like the team behind setup-python are interested in adding that functionality to their action, unfortunately.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -124,6 +124,7 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.9"
+          token: ${{ env.PAT }}
 
       - name: Set up CI Gradle Properties
         run: |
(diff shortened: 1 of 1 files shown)


git-phu (approved):
working in this PR run, seems legit

unfortunately the log output doesn't indicate that we're using a PAT now, but seems like it should be

cpdeethree:
> working in this PR run, seems legit
> 
> unfortunately the log output doesn't indicate that we're using a PAT now, but seems like it should be

Yeah the behavior of this action is very under-the-hood, i had to dig into their source code and issues to even figure out that they were using tokens and authing to github. That "token" option isn't really front and center on their README

PR #21746, opened 2023-01-23
🪟 🔧 Include module name in CSS class names

timroes:
## What

Makes CSS class names be generated in a format to contain the original file name. See the screenshot (except the highlighted row, which is a styled component):

screenshot-20230123-213530

### DIFF
--- a/airbyte-webapp/vite.config.ts
+++ b/airbyte-webapp/vite.config.ts
@@ -69,6 +69,11 @@ export default defineConfig(({ mode }) => {
     define: {
       ...processEnv,
     },
+    css: {
+      modules: {
+        generateScopedName: "[name]__[local]__[contenthash:6]",
+      },
+    },
     resolve: {
       alias: {
         // Allow @use "scss/" imports in SASS


ambirdsall (approved):
:+1: Much nicer. :sparkles: :guitar: :fire:

One issue this highlights (to be clear, an issue that is out of scope for this PR): the `.module.scss` filename convention is, correctly, only being parsed as a `.scss` extension, so all the generated classes have a boilerplate `-module` in the middle, which I think hurts readability:
| generated class name | wouldn't it be nice, tho |
|---|---|
| `Card-module__header__ac1808` | `Card__header__ac1808` |
| `ConnectionStatusPage-module__title__ed6618`. | `ConnectionStatusPage__title__ed6618` |

Since IIUC no build mechanism relies on that convention, perhaps a change is in order?

timroes:
@ambirdsall sorry this PR was on auto merge so got merged with the approval automatically.

The .module.scss is actually needed for the build system to detect it as a CSS module and make the classnames be importable. I could though have specified a function for this property that calculates the name, but that would also require us to do the content hashing etc. ourselves. I found the minor disadvantage of having module in the classnames would outweigh adding 5-10 more lines of classnames calculation in the build script. But if we feel that the module is annoying too much in the class name, I'm open for that path as well.

Issue #21542, opened 2023-01-18
Handle protocol v1 data types in UI


edgao:
Protocol v1 will introduce a new style of declaring schemas, which mostly relies on a set of [predefined primitives](https://github.com/airbytehq/airbyte/blob/master/airbyte-protocol/protocol-models/src/main/resources/airbyte_protocol/well_known_types.yaml) rather than relying on JsonSchema's primitive types.

The UI should be updated to display those types correctly (i.e. on the replication tab of a connection). After v1, the UI will no longer need to handle v0 types (so e.g. it would _not_ need to support things like `{"type": "string", "format": "date-time"}`).

edgao:
(fyi @grishick this is a requirement that I missed)

gosusnp:
@edgao, I think the UI is directly getting the datatypes info from the catalog today: https://github.com/airbytehq/airbyte/blob/21b5eabc965a6d85cc5fe33ce15f0bed65e43959/airbyte-server/src/main/java/io/airbyte/server/handlers/helpers/CatalogConverter.java#L39
Following the data type updates, currently the UI should be getting all v1 schemas, I do not know what would be the preferred representation from a front end perspective.

gosusnp:
For context, UI should be using https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/utils/useTranslateDataType.ts today to map the received data type to what is being shown in the UI

Issue #21234, opened 2023-01-10
Migration of ConfiguredCatalog


gosusnp:
## What
For the new data types, we need to migrate the `ConfiguredAirbyteCatalog` that we currently store in the database to v1 objects of the protocol.

## How
Define a Flyway Migration that will
* backup existing configs
  * `connection.catalog` is a `ConfiguredAirbyteCatalog`
  *  `jobs.config` is a `JobConfig` which contains a field that can be a `ConfiguredAirbyteCatalog` if it is a reset or sync
  *  `actor_catalog.catalog` is an `AirbyteCatalog`, do we need to migrate for data types?
* migrate existing configs

Document steps to restore or have a draft PR ready?

krishnaglick:
Groomed, to be mob'd by platform-move backend team.

gosusnp:
Platform has been updated to support on the fly migration of persisted objects instead.

PR #21757, opened 2023-01-23
On-the-fly migrations of persisted catalogs

gosusnp:
## What
With the introduction of protocol version v1, the assumption is that all the protocol objects handled by the platform will be v1. We are persisting `AirbyteCatalog` and `ConfiguredAirbyteCatalog` in our database which breaks the previous assumption as the platform will be reading v0 objects.

Closes https://github.com/airbytehq/airbyte/issues/21234
Closes https://github.com/airbytehq/airbyte-internal-issues/issues/2535
Closes https://github.com/airbytehq/airbyte/issues/21542

## How

Add code to migrate version 0 of the `AirbyteCatalog` and `ConfiguredAirbyteCatalog` when reading from the database if needed. This is done by adding helpers to specifically detect v0 data types to infer the version of the objects and using this code in our DbConverters.

Update Frontend to support data types v1 as well.

## Recommended reading order
1. airbyte-commons-protocol/src/main/java/io/airbyte/commons/protocol/migrations/v1/CatalogMigrationV1Helper.java is the logic to detect version and apply the migration.
2. the rest is about updating the reads we need to cover


gosusnp:
@edmundito, since this change impacted the integration tests, I ended up taking a stab at https://github.com/airbytehq/airbyte/issues/21542. Could you please take a look at the front end changes?

airbyte-persistence/job-persistence/src/main/java/io/airbyte/persistence/job/DefaultJobPersistence.java
```diff
@@ -917,20 +919,32 @@ private static Job getJobFromRecord(final Record record) {
     return new Job(record.get(JOB_ID, Long.class),
         Enums.toEnum(record.get("config_type", String.class), ConfigType.class).orElseThrow(),
         record.get("scope", String.class),
-        Jsons.deserialize(record.get("config", String.class), JobConfig.class),
+        parseJobConfigFromString(record.get("config", String.class)),
```
edgao, line 922: how would we have a jobconfig with a v0 schema? Wouldn't we upgrade the schema before creating the jobconfig?

edgao (approved):
CatalogMIgrationV1Helper looks right. I skimmed through the rest of the pr and didn't see anything noticeably off, though I definitely didn't fully understand all of it :shrug:

also confirmed that UI changes have the right schema->type mappings.

airbyte-persistence/job-persistence/src/main/java/io/airbyte/persistence/job/DefaultJobPersistence.java
gosusnp: Since we're updating on the fly, there is no migrations of the DB. All jobs that were created before the upgrade will have v0 data.

airbyte-webapp/src/locales/en.json
```diff
@@ -628,6 +628,17 @@
...
+  "airbyte.datatype.WellKnownTypes.json#/definitions/String": "String",
+  "airbyte.datatype.WellKnownTypes.json#/definitions/BinaryData": "Binary Data",
+  "airbyte.datatype.WellKnownTypes.json#/definitions/Number": "Number",
+  "airbyte.datatype.WellKnownTypes.json#/definitions/Integer": "Integer",
+  "airbyte.datatype.WellKnownTypes.json#/definitions/Boolean": "Boolean",
+  "airbyte.datatype.WellKnownTypes.json#/definitions/Date": "Date",
+  "airbyte.datatype.WellKnownTypes.json#/definitions/TimestampWithTimezone": "Timestamp with Timezone",
+  "airbyte.datatype.WellKnownTypes.json#/definitions/TimestampWithoutTimezone": "Timestamp without Timezone",
+  "airbyte.datatype.WellKnownTypes.json#/definitions/TimeWithTimezone": "Time with Timezone",
+  "airbyte.datatype.WellKnownTypes.json#/definitions/TimeWithoutTimezone": "Time without Timezone",
```
edmundito, line 640: Ideally, we still want to use the `airbyte.datatype` values below this so that we don't have duplicate strings. The utility function may need a way to translate these types, for example:

`"WellKnownTypes.json#/definitions/TimestampWithTimezone" -> "airbyte.datatype.time_stamp_with_timezone"`

gosusnp: I added a map in `useTranslateDataType` utility and updated the strings here.

edmundito (approved):
Frontend changes look good and tested on areas that use the data type

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -1184,8 +1184,10 @@ private Map<UUID, AirbyteCatalog> findCatalogByHash(final String catalogHash, fi
 
     final Map<UUID, AirbyteCatalog> result = new HashMap<>();
     for (final Record record : records) {
-      final AirbyteCatalog catalog = Jsons.deserialize(
-          record.get(ACTOR_CATALOG.CATALOG).toString(), AirbyteCatalog.class);
+      // We do not apply the on-the-fly migration here because the only caller is getOrInsertActorCatalog
+      // which is using this to figure out if the catalog has already been inserted. Migrating on the fly
+      // here will cause us to add a duplicate each time we check for existence of a catalog.
```
colesnodgrass, line 1189: This seems risky, granted this is a `private` method so there is less risk, but I wonder if this code should be moved to the `getOrInsertActorCatalog` method to prevent any future issues from occurring.

airbyte-webapp/src/utils/useTranslateDataType.ts
```diff
@@ -3,19 +3,37 @@ import { useIntl } from "react-intl";
 
 export interface AirbyteConnectorData {
   type: string;
+  $ref?: string;
```
colesnodgrass, line 6: Any tests need to be updated or added for this change?

colesnodgrass (commented):
Changes look fine.  

Are we emitting any metrics when `upgradeschemaifNeeded` is need to track how often that method is actually upgrading schemas?  If we're not, that is probably something we should track and emit to DD.

airbyte-webapp/src/utils/useTranslateDataType.ts
gosusnp: The UI change happened in this PR because it was breaking some integration tests.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
gosusnp: I am open to inlining this in `getOrInsertActorCatalog`. One extra note is that this behavior (the insert only inserts when it's actually different) is currently tested.

benmoriceau (commented):
LGTM, I'll let Cole to approve

PR #21763, opened 2023-01-24
:window: :wrench: Cleaner error handling for authService's `sendVerificationEmail`

ambirdsall:
## What
When I first extracted this error handling code from `EmailVerificationHint.tsx` in https://github.com/airbytehq/airbyte/pull/21623 so the enrollment modal could take advantage of it too, I missed a basic structural fix (because the internal API defined in `authService` had an identical type signature as the upstream `firebase/auth` function it wrapped, it Seemed Obvious:tm: that it was appropriate to port its error-handling code unchanged). To wit: because it handled errors _within_ the `Promise`, rather than _via_ the `Promise`, success/error code paths were being defined in different scopes; the only way to invalidate a downstream `.then` which should only run on success was to reraise the error. This lead to arbitrary and ugly `.catch()`es getting tacked onto both usage sites just to avoid crashing the page from an already-handled error.

But, like: that's the whole reason `Promise.prototype.catch` exists, though.

## Testing
Manually tested via the enrollment modal, behaves exactly as before in both error and success states.


joeykmh (approved):
Thanks for refactoring this! Explanation makes sense, and this is much cleaner 👍

PR #21766, opened 2023-01-24
🪟 🎉 Log out if a user logs out of another tab

teallarson:
## What
closes https://github.com/airbytehq/airbyte-cloud/issues/3735

Previously, a user could click in the UI and end up with an error screen if they logged out in another tab.
Now, we log out and redirect to sign-in screen immediately if user logs out in another tab

## How
Firebase already watches for updates consistently, so added a check in our Service for a logged in user.  If there is none, tell the reducer.

The flow for automatically logging _in_ from another tab is proving a bit touchier and thus will be solved in a separate PR as it requires some more digging.

## Recommended reading order
1. the one and only

## Testing
Manual testing:
1. Log in
2. Open another tab
3. Log out in one of the tabs
4. You should be logged out in that tab
5. The other tab should be logged out too


ambirdsall (approved):

joeykmh (approved):
Tested again locally, and worked as expected 👍

PR #21770, opened 2023-01-24
Use generic `manifest.yaml` name for manifest in generated lowcode CDK connectors

sherifnada:
Making the name uniform because it allows us to simplify any automation on top of lowcode connectors and workflows like downloading a user's YAML for debugging or running a connector builder connector on the UI. 

Might merge another PR batch applying this rename to connectors


brianjlai (approved):

maxi297:
> Might merge another PR batch applying this rename to connectors

This could be interesting. I see that we are building automation around this (first thing that comes up to my head is migrating manifests but there might be more) and migrating the manifest will impact those

maxi297:
> does this impact the migration to beta?
@sherifnada Yes it does but don't hold for us. As you mentioned, not a big blocked. Just something we need to take into account so that we update our migration scripts. You can go forward with the change

PR #21771, opened 2023-01-24
Rename all lowcode connector manifest files to `manifest.yaml`

sherifnada:
## What
Following up on #21770 -- applies that change to all lowcode connectors

I've spotchecked a few connectors and this rename seems to work. It was applied via a script so it should be uniform across the board. 

for each connector, the only changes is that we change the yaml manifest name from `source_<name>.yaml` to `manifest.yaml`, and we change `source.py` to load that file instead of `source_<name>.yaml`.

I don't think a connector publish is needed


airbyte-integrations/connectors/source-zoom/source_zoom/source.py
```diff
@@ -15,4 +15,106 @@
...
 class SourceZoom(YamlDeclarativeSource):
     def __init__(self):
-        super().__init__(**{"path_to_yaml": "zoom.yaml"})
+        super().__init__(**{"path_to_yaml": "manifest.yaml"})
+
+
+
+
+
+x=["airbyte-integrations/connectors/source-activecampaign/source_activecampaign/source.py",
```
maxi297, line 24: This seems odd. Why is this part of the zoom source?

maxi297 (approved):
LGTM minus the migration script in the Zoom source

airbyte-integrations/connectors/source-zoom/source_zoom/source.py
sherifnada: fixed

PR #21789, opened 2023-01-24
improve logging of find-valid-pat script

Mykyta-Serbynevskyi:
Improve error logging of `find_non_rate_limited_PAT` script to cover case when PR changes are proposed by non-contributor

Example - https://airbytehq.slack.com/archives/C046KQF5GBT/p1673547644086219


supertopher (approved):
also looking good.  ty

PR #21791, opened 2023-01-24
[ISSUE #19981] improve changelog in slack notification

maxi297:
## What
On release, the Slack notification redirect to lines 3 and 4 of the changelog (url: https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/CHANGELOG.md?plain=1#L3-L4). This causes some issue:
* When there is a new release, this link is not update on previous Slack notification release and therefore is not useful
* If a release is done in a branch, the link still points on master (we try to avoid releasing in branches but it seems it's still something that is helpful in some cases like https://github.com/airbytehq/airbyte/pull/21703#issuecomment-1400652322)

### DIFF
--- a/.github/workflows/publish-cdk-command-manually.yml
+++ b/.github/workflows/publish-cdk-command-manually.yml
@@ -178,14 +178,14 @@ jobs:
                         "type": "section",
                         "text": {
                             "type": "mrkdwn",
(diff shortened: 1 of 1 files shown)


clnoll (approved):
LGTM!

PR #21795, opened 2023-01-24
🪟 🐛 Fix test with potentially mismatching time values

joeykmh:
## What
In the test setup there are two subsequent calls to `Date.now()` that are not guaranteed to return the same millisecond. This test failed because they were off by a millisecond: https://github.com/airbytehq/airbyte-cloud/actions/runs/3996900752/jobs/6857563137#step:3:16508

I just adjusted the `makeAttempt()` setup slightly so there is only one call to `Date.now()`


evantahler (approved):
Thanks!

PR #21798, opened 2023-01-24
POC: update source-greenhouse with option to run SATs with local CDK

clnoll:
Example of surfacing the ability to run SATs against the local CDK using an environment variable, by copying the required files to a build directory.

image

See the [updated proposal](https://docs.google.com/document/d/1Awk6h4qpEKUWYesXOKCcIQRlEIRHq-x-vcHFcuXPaT0/edit#) for running SATs against the local CDK.


airbyte-cdk/Dockerfile.sats
```diff
@@ -0,0 +1,5 @@
+FROM python:3.9.11-alpine3.15
```
girarda, line 1: do we know if/why SATs can't run on 3.9-slim? I agree that using a single image for "almost" everything would be better

clnoll: Just hit a small snag with a dependency (`apk`) so decided to punt on consolidating them until we were pretty sure we wanted to. I'll go ahead and see what it looks like to use a single base image.

clnoll:
Notes:
- Default to published package
- Definitely want to integrate into CI
- If modifying CDK, consider defaulting to local CDK
- Connectors with pinned versions should use that version
- We don't always publish a new version of the CDK on every change so should still test connectors against published package
- Select a few connectors to validate against the updated CDK (maybe do all in future)

airbyte-cdk/Dockerfile.connectors
```diff
@@ -0,0 +1,5 @@
+FROM python:3.9-slim
+
+WORKDIR /home/airbyte-cdk
+COPY . ./
```
maxi297, line 4: Why not copy `python` instead of `.`? Is there value to copy the license file and the `build` folder?

airbyte-integrations/bases/source-acceptance-test/setup.py
```diff
@@ -2,11 +2,19 @@
 # Copyright (c) 2022 Airbyte, Inc., all rights reserved.
 #
 
+import os
 
 import setuptools
 
+CDK_VERSION = "airbyte-cdk~=0.2"
+
+if os.getenv("USE_LOCAL_CDK"):
```
maxi297, line 11: Could we only rely on `LOCAL_CDK_DIR` from there on? I fear that `USE_LOCAL_CDK` is doing two things currently:
* Indicate if we should build the docker images with the proper information
* Indicate if pip should install the local version or not

I guess maybe one way to answer this is to ask: 
* Running `pip install -e .` on the `airbyte-integrations/bases/source-acceptance-test`, should I have to configure both `USE_LOCAL_CDK` and `LOCAL_CDK_DIR` in order to build with the local version of airbyte-cdk? 
* What should happen is one is defined and not the other?

airbyte-integrations/connectors/source-greenhouse/acceptance-test-docker.sh
```diff
@@ -1,10 +1,23 @@
...
+set -e
+
+CDK_DIR=../../../airbyte-cdk
+
+if [ -n "$USE_LOCAL_CDK" ]; then
+  # Build the Docker images associated with the local version of the CDK
+  docker build -f $CDK_DIR/Dockerfile.connectors $CDK_DIR --build-arg USE_LOCAL_CDK=$USE_LOCAL_CDK -t airbyte/airbyte-cdk-connectors
+  docker build -f $CDK_DIR/Dockerfile.sats $CDK_DIR --build-arg USE_LOCAL_CDK=$USE_LOCAL_CDK -t airbyte/airbyte-cdk-sats
+  # Build the SAT Docker image
+  docker build ../../bases/source-acceptance-test --build-arg USE_LOCAL_CDK=$USE_LOCAL_CDK -t airbyte/source-acceptance-test
```
maxi297, line 12: Do we need to build the SATs docker using our local version of the airbyte-cdk? It feels better for it to be aligned with the connector, but I'm not clear as to if it is needed. The reason why is that currently, when I run SATs for a source using my local cdk (so with the manual file manipulation), I don't change anything regarding the SATs docker image

maxi297 (commented):
I'm still curious about the `airbyte-integrations/connectors/source-greenhouse/setup.py` which I didn't check properly yet but I still want to give feedback as early as possible. I might add more comments later

airbyte-integrations/connectors/source-greenhouse/acceptance-test-docker.sh
girarda: I think this is useful for a different use case: modifying the SAT library. Otherwise, I agree that we should use the same SAT version when running the connector's tests

clnoll: My feeling was that if a user was using the `USE_LOCAL_CDK` option all instances of the CDK should be local, to avoid having to specify separately. But given that we've been focused on connector testing, if we want to limit the scope of this specific task to the connector that would make sense to me.

airbyte-integrations/bases/source-acceptance-test/setup.py
clnoll: This is definitely something that I wanted to circle back to - I'm working on cleaning it up. The current issue with using just `LOCAL_CDK_DIR` is that it's always defined in the base image. However, that also means that the images that inherit from the base image always have `LOCAL_CDK_DIR` set, even if the user wanted to run the code against the pypi package.

The fix that I'm working on is to only copy the local CDK to the base image if `LOCAL_CDK_DIR` is defined.

airbyte-cdk/Dockerfile.connectors
clnoll: Nope I think we're good to just copy over `python`. Will make that change.

Dockerfile.local_cdk
```diff
@@ -0,0 +1,12 @@
+FROM python:3.9-slim
+
+COPY airbyte-cdk/python ./
+
+ARG CONNECTOR_NAME
+COPY airbyte-integrations/connectors/$CONNECTOR_NAME ./$CONNECTOR_NAME
```
maxi297, line 6: Just so I can understand how we can extend the usage of this: if we want to use this for Connector Builder Server for example, we would need to define another Dockerfile that copies the airbyte-cdk/python and the airbyte-connector-builder-server, right?

maxi297 (commented):
I still feel like I need to try it out to understand it properly but I can't right now since docker fails while trying to pull airbyte/airbyte-cdk

airbyte-integrations/connectors/source-greenhouse/acceptance-test-docker.sh
```diff
@@ -1,16 +1,36 @@
...
+docker build . -t $CONNECTOR_TAG:dev
 
-# Pull latest acctest image
-docker pull airbyte/source-acceptance-test:latest
+if [ -n "$LOCAL_CDK" ]; then
+  # Build an image with the Python environment of the connector, built against the local CDK
+  docker build -f $ROOT_DIR/Dockerfile.local_cdk -t airbyte/$VOLUME_NAME --build-arg CONNECTOR_NAME=$CONNECTOR_NAME $ROOT_DIR
+  # Create the volume, seeded with the installed Python environment for the connector
+  docker volume rm $VOLUME_NAME 2>/dev/null || true
+  docker run --mount source=$VOLUME_NAME,destination=$SITE_PACKAGES_LOCATION airbyte/airbyte-cdk
```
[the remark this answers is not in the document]
clnoll: Sorry about that @maxi297 didn't realize I had some outstanding code. This should be `airbyte/$VOLUME_NAME`. I've pushed the change up.

Dockerfile.local_cdk
clnoll: That's one option, another would be to modify this file and some of the upstream code so that it works with both "normal" connectors and the connector builder server.

airbyte-cdk/Dockerfile.sats
clnoll: Resolving this as with the latest update we're not modifying existing Dockerfiles.

airbyte-integrations/bases/source-acceptance-test/setup.py
clnoll: Resolving this as the updates will only require a user to specify one environment variable.

airbyte-integrations/connectors/source-greenhouse/acceptance-test-docker.sh
clnoll: The updated POC focuses on SATs and so `LOCAL_CDK` will only refer to the CDK that the connector is running against, for now. We can modify it in the future if we want to extend the behavior to the SATs themselves.

Dockerfile.local_cdk
```diff
@@ -0,0 +1,12 @@
+FROM python:3.9-slim
```
sherifnada, line 1: A single Dockerfile like this one may be problematic if it assumes what connectors' Dockerfiles look like (the dockerfile may deviate from the "vanilla" ones). Though TBH my guess is that 95% of the time it's not an issue? 

in any case, is our life easier if we do something a _little_ boneheaded (but effective) and do something like run a shell script with code like: 

```
cp airbyte-cdk/python airbyte-integrations/connectors/$CONNECTOR/build/airbyte-cdk
cp airbyte-integrations/connectors/$CONNECTOR/Dockerfile airbyte-integrations/connectors/$CONNECTOR/build/Dockerfile.dev # Copy the connector's Dockerfile into build/Dockerfile.dev
cd airbyte-integrations/connectors/$CONNECTOR/
echo "RUN pip install -e ./build/airbyte-cdk" >> build/Dockerfile.dev # install local CDK
docker build . -f build/Dockerfile.dev -t <tag> 
cd - 
```

assuming we `.gitignore` build, it'd achieve something similar to what this is doing without polluting anything meaningful

sherifnada: This would also allow us to avoid having to `sed` stuff and fiddle with site-packages

airbyte-integrations/connectors/source-greenhouse/acceptance-test-docker.sh
```diff
@@ -1,16 +1,36 @@
...
+
+ROOT_DIR=../../..
+SITE_PACKAGES_LOCATION=/usr/local/lib/python3.9/site-packages/
+CONNECTOR_TAG=$(cat acceptance-test-config.yml | grep "connector_image" | head -n 1 | cut -d: -f2)
+CONNECTOR_NAME=$(echo $CONNECTOR_TAG | cut -d / -f 2)
+VOLUME_NAME=$CONNECTOR_NAME-local-cdk
+
 # Build latest connector image
-docker build . -t $(cat acceptance-test-config.yml | grep "connector_image" | head -n 1 | cut -d: -f2)
+docker build . -t $CONNECTOR_TAG:dev
```
sherifnada, line 12: I think in this situation the connector isn't running the latest CDK?

sherifnada (commented):
there's 3 situations where SAT is invoked: 
1. locally using `python -m pytest integration_tests -p integration_tests.acceptance`
2. `./acceptance-test-docker.sh` -- this is meant to be used by the developer locally, not in CI (differs from #1 in that it operates on the connector's docker image rather than directly testing the Python code) 
3.  `./gradlew :airbyte-integrations:connectors:source-greenhouse:sourceAcceptanceTest` which is invoked (technically we invoke `:integrationTest` whcih depends on `:sourceAcceptanceTest`)

Can you clarify how this solution would plug into #3? (arguably for 1 we don't need it at all though since we can just `pip install -e ../../../airbyte-cdk/python`, and for #2 the code in this PR makes it clear)

airbyte-integrations/bases/connector-acceptance-test/acceptance-test-docker.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env sh
+
+set -e
+
+die () {
```
sherifnada, line 5: 💀

airbyte-integrations/bases/connector-acceptance-test/acceptance-test-docker.sh
```diff
@@ -0,0 +1,59 @@
...
+CONNECTOR_TAG_DEV="$CONNECTOR_TAG:dev"
+CONNECTOR_NAME="$(echo $CONNECTOR_TAG | cut -d / -f 2)"
+CONNECTOR_DIR="$ROOT_DIR/airbyte-integrations/connectors/$CONNECTOR_NAME"
+CONNECTOR_SUBDIR="$CONNECTOR_DIR/$(echo $CONNECTOR_NAME | sed 's/-/_/g')"
+BUILD_DIR=$(mktemp -d)
+
+if [ -n "$LOCAL_CDK" ] && [ -f "$CONNECTOR_DIR/setup.py" ]; then
+  # Copy the connector files & CDK to the build directory
+  cd "$BUILD_DIR"
+  cp $CONNECTOR_DIR/setup.py .
```
sherifnada, line 24: why not copy the whole connector directory and `cd` into it instead of cherry picking files?

clnoll: There's some cruft in there like venvs etc. that, if copied into the Docker container, can cause pip to pick up a cached version of the CDK instead of installing it locally. I haven't identified exactly which file/directory/etc. was the culprit, and thought that we'd probably want to minimize the extra baggage that we copy to the container anyway, so just went with a cp of what we need. But I'm happy to keep digging and copy the whole thing if we want to mimic the local environment more closely, rather than just the CDK.

clnoll:
Closing; implemented in https://github.com/airbytehq/airbyte/pull/23084.

PR #10861, opened 2022-03-04
set memory limit/request for blob storage connectors

lmossman:
## What
Resolves https://github.com/airbytehq/airbyte-cloud/issues/1082

Set the connector-definition level memory limit/request resource requirement for all blob storage connectors to 1Gi, as recommended by the connectors team. See issue for details.

## How
Uses the sync jobSpecific resource requirement configuration in the destination definition seed yaml to set the memory limit/request to 1Gi for all blob storage connectors.

This PR also removes some resource requirement configuration from the snowflake connector, as noted in the comments below

## Recommended reading order
1. `x.java`
2. `y.python`

PR #21799, opened 2023-01-24
Remove static resourceRequirements for warehouse destinations

malikdiarra:
Resource requirements can currently be set at multiple levels:

- a default instance-wide value
- a value defined at the connector definition level
- an override on a per-connection basis.

The static resource requirements defined at the connector level will override the default resource requirements even if the default value is higher. This could lead to unexpected result if a user changes the default resource requirements. Keeping those connector definition level empty initially is a better strategy.

### DIFF
--- a/airbyte-config/init/src/main/resources/seed/destination_definitions.yaml
+++ b/airbyte-config/init/src/main/resources/seed/destination_definitions.yaml
@@ -4,12 +4,6 @@
   dockerImageTag: 0.1.6
   documentationUrl: https://docs.airbyte.com/integrations/destinations/azureblobstorage
   icon: azureblobstorage.svg
-  resourceRequirements:
-    jobSpecific:
-      - jobType: sync
-        resourceRequirements:
-          memory_limit: "1Gi"
-          memory_request: "1Gi"
   releaseStage: alpha
 - name: Amazon SQS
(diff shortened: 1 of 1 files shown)


evantahler (commented):
Adding @pedroslopez to hear his thoughts as well.

This makes sense as in Airbyte Cloud, we've increased the default resources for these/all connectors.

For OSS users, what are the default resources?  Should we increase the default OSS limits to match cloud, or leave these in for OSS users with lower defaults?

I think I'd like confirmation that for OSS K8s users, the defaults are also increased for connectors...

pedroslopez (commented):
The changes look good to me, though I agree with @evantahler that we should make sure our OSS users aren't negatively affected.

These were originally set in https://github.com/airbytehq/airbyte/pull/10861 to address a cloud issue, but from the slack thread linked in that PR it seems like those are the minimum requirements regardless of running on cloud or not?

If we didn't bump the default requirements on OSS, we should validate if these are still the resources that should still be set.  I'm not sure what the minimum resource requirements should be for these warehouse destinations, maybe @grishick can weigh in?

grishick:
We haven't done much profiling of destination memory consumption, so I would hesitate to make a change that changes resource requirements for destinations. How does this change affect OSS users?

PR #21809, opened 2023-01-24
Update FCP email notification phrasing

ambirdsall:
## What
An internal Airbyte slide used phrasing like the new version here, which seemed a bit more natural and to the point.


andyjih (approved):
👍

PR #21810, opened 2023-01-24
🪟 🧹 Stream details panel text cols should be equal width

### LINKED ISSUE #20454: Stream details panel text cols should be equal width
In the stream fields panel, columns with text should be equal width.

Columns with radio, icon, checkbox, or arrow should be static at the minimum size possible.

(if any text overflows that width, it should truncate with ellipses.  #20450 tracks adding a tooltip for any truncated text)

This should be hidden behind the new streams table feature flag

dizel852:
## What
Resolves https://github.com/airbytehq/airbyte/issues/20454

UPD: I've updated the PR since the first implementation wasn't satisfied the requirements.

https://www.loom.com/share/6e9c63f8ab3541f7a9201d6bc2428171
## How
Set range width for text cells(Source and Destination field names).
For other cells - set min-width: why not static width? - it looks nicer when we zoom out.


YatsukBogdan1 (approved):
tested locally. looks good to me

airbyte-webapp/src/components/connection/CatalogTree/next/StreamDetailsPanel/StreamFieldsTable/StreamFieldsTable.module.scss
```diff
@@ -59,35 +60,45 @@ $cell-left-padding: variables.$spacing-xl + variables.$spacing-sm;
...
   @extend %bodyCell;
 
   color: colors.$dark-blue;
+  max-width: 0;
 }
 
 .dataTypeCell {
   @extend %bodyCell;
 
+  width: $cell-fixed-width;
```
edmundito, line 83: The data type cell should maybe have a min-width and max-width. The max width should be up to 200px or so to fit the new data types introduced in #21757 (notable `Timestamp With Timezone`

edmundito (approved):
Looking really good, just one more change for long data types...

airbyte-webapp/src/components/connection/CatalogTree/next/StreamDetailsPanel/StreamFieldsTable/StreamFieldsTable.module.scss
dizel852: Double-checked: if `100px` is not enough - the cell will grow to fit the content
image

PR #18468, opened 2022-10-26
🪟 🧪 [Experiment] Show source selector on signup form

letiescanciano:
Demo: https://www.loom.com/share/f676522a48184a48adfb461232937f5e 
With Connection Flow: https://www.loom.com/share/2f34ee77a49348c3950988b72da02e55
Update: we are now able to fetch directly from cloud [catalog](https://storage.googleapis.com/prod-airbyte-cloud-connector-metadata-service/cloud_catalog.json) so ignore the part about having that json harcoded. Related [Issue](https://github.com/airbytehq/airbyte-cloud/issues/3261)

Experiment info
https://airtable.com/appIuY0uKPVnk8TWT/tbl2SxXnUwf6fVCWS/viwtQpgNYwBvoHllF/reccTmRCcw1mVLU6i?blocks=bipVSxfu1Z9WinN3B

## What
As an experiment, we are including a source selector in the signup form. We think that by selecting a source, it will force users to be more thoughtful about signing up for Airbyte and decrease the rate of which users enter the platform and do nothing.

## How
- Displaying a lightweight version of `ConnectorServiceTypeControl` in the signup form
 Screen Shot 2022-10-26 at 10 18 01 AM
- Redirecting users directly to `/source/new-source ` and preselect their source on the dropdown
Screen Shot 2022-10-26 at 10 22 02 AM

PR #21819, opened 2023-01-24
🪟 🧪 Remove signup source selection experiment

timroes:
## What

We've evaluated that having the source selector on the sign up page has none (positive) impact on users finding their way through setting up their first connection, so we're removing the experiment again.

This basically reverts https://github.com/airbytehq/airbyte/pull/18468


joeykmh (approved):
LGTM - gotta love that ratio 
image

letiescanciano (commented):
LGTM, thanks for taking care of this

Issue #21291, opened 2023-01-12
Create routing logic for discover and check workflow


xiaohansong:
Now these workflow use the following logic to define the task queue name (simply looking up the job type and translate that to string).

https://github.com/airbytehq/airbyte/blob/master/airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/TemporalWorkflowUtils.java#L52

To separate these two workflows into dataplane level we need to implement a routing logic to determine which queue it needs to go based on job id. Once that helper function is implemented, change these functions to make sure when we start the workflow we used this routing logic:

https://github.com/airbytehq/airbyte/blob/master/airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/TemporalClient.java#L339-L373

mfsiega-airbyte:
Refining notes:
- In the sync workflow we determine the appropriate queue routing as a task in the connection manager workflow. However, the discover/check workflows don't have this, so we need to do this before we dispatch to a task queue.

PR #21822, opened 2023-01-24
Add routing logic for check and discover workflow

xiaohansong:
## What
Check and discover workflow will now use a routing logic to decide which task queue to enqueue the job.

https://github.com/airbytehq/airbyte/issues/21291

## How
Expanding routeService to support check/discover.
Inject routeService to server app.

## Recommended reading order

Two major parts I want this to get a closer look:
1) RouteService injection on ServiceApp. We need to maintain two parts: micronaut injection (used in OSS) and ServerApp (used in cloud); since we want to overwrite `taskQueueMapper` in cloud, is it okay for me to inject it as a parameter?
2) RouteService logic - I created a few new functions to RouteService to support getting taskqueue by geography. In the future this might be requiring a workspaceId to get the correct queue, for a true hybrid cloud settings

## 🚨 User Impact 🚨

No - changes should be invisible to user.

Tested with https://github.com/airbytehq/airbyte-cloud/pull/4185


airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/scheduling/DefaultTaskQueueMapper.java
```diff
@@ -0,0 +1,78 @@
...
+package io.airbyte.commons.temporal.scheduling;
+
+import com.google.common.annotations.VisibleForTesting;
+import io.airbyte.commons.temporal.TemporalJobType;
+import io.airbyte.config.Geography;
+import jakarta.inject.Singleton;
+import java.util.Map;
+
+@Singleton
+public class DefaultTaskQueueMapper implements TaskQueueMapper {
```
pmossman, line 14: I wonder if it would be easier to maintain a different approach here - my rough idea is that instead of maintaining a separate map for every job type going forward and adding a new RouterService method for each job type, perhaps the RouterService and TaskQueueMapper should be more opinionated and force a standard task queue format based on the geography and TemporalJobType name.

So, instead of having a separate map and method for every type of job, we could instead have a single method `getTaskQueue(TemporalJobType, Geography)` which returns the appropriate task queue. The DefaultTaskQueueMapper could simply return the TemporalJobType name without considering the Geography at all, and the CloudTaskQueueMapper could have a map from `Geography` to `taskQueueSuffix` that it adds to the TemporalJobType name to get the appropriate task queue.

Does that make sense? I haven't reviewed this full PR but I want to leave this comment up front to see what you think - I think it will make it easier to use/test/maintain the RouterService and TaskQueueMapper classes, and it forces some level of convention around how we name task queues for different jobs and geographies. Let me know what you think!

xiaohansong: Sure, thanks for the suggestion. That's done!

airbyte-commons-server/src/main/java/io/airbyte/commons/server/scheduler/DefaultSynchronousSchedulerClient.java
```diff
@@ -80,12 +86,13 @@ public SynchronousResponse<StandardCheckConnectionOutput> createSourceCheckConne
 
     final UUID jobId = UUID.randomUUID();
     final ConnectorJobReportingContext jobReportingContext = new ConnectorJobReportingContext(jobId, dockerImage);
+    String taskQueue = routerService.getTaskQueue(source.getWorkspaceId(), TemporalJobType.CHECK_CONNECTION);
```
pmossman, line 89: nit: can be `final`

airbyte-commons-server/src/main/java/io/airbyte/commons/server/scheduler/DefaultSynchronousSchedulerClient.java
```diff
@@ -110,12 +117,13 @@ public SynchronousResponse<StandardCheckConnectionOutput> createDestinationCheck
 
     final UUID jobId = UUID.randomUUID();
     final ConnectorJobReportingContext jobReportingContext = new ConnectorJobReportingContext(jobId, dockerImage);
+    String taskQueue = routerService.getTaskQueue(destination.getWorkspaceId(), TemporalJobType.CHECK_CONNECTION);
```
pmossman, line 120: nit: can be `final`

airbyte-commons-server/src/main/java/io/airbyte/commons/server/scheduler/DefaultSynchronousSchedulerClient.java
```diff
@@ -144,11 +152,13 @@ public SynchronousResponse<UUID> createDiscoverSchemaJob(final SourceConnection
     final UUID jobId = UUID.randomUUID();
     final ConnectorJobReportingContext jobReportingContext = new ConnectorJobReportingContext(jobId, dockerImage);
 
+    String taskQueue = routerService.getTaskQueue(source.getWorkspaceId(), TemporalJobType.DISCOVER_SCHEMA);
```
pmossman, line 155: nit: can be `final`

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -1544,6 +1544,14 @@ public Geography getGeographyForConnection(final UUID connectionId) throws IOExc
         .fetchOneInto(Geography.class);
   }
 
+  public Geography getGeographyForWorkspace(final UUID workspaceId) throws IOException {
```
pmossman, line 1547: I don't see this used in this PR, is it called in a different PR or in Cloud? Also probably worth adding a quick unit test for this method

airbyte-workers/src/main/java/io/airbyte/workers/ApplicationInitializer.java
```diff
@@ -366,4 +378,28 @@ private Set<String> getSyncTaskQueue() {
     return Arrays.stream(syncTaskQueue.split(",")).collect(Collectors.toSet());
   }
 
+  /**
+   * Retrieve and parse the sync workflow task queue configuration.
```
pmossman, line 382: nit: should say "check" instead of "sync" in comment

airbyte-workers/src/main/java/io/airbyte/workers/ApplicationInitializer.java
```diff
@@ -366,4 +378,28 @@ private Set<String> getSyncTaskQueue() {
...
+   */
+  private Set<String> getCheckTaskQueue() {
+    if (StringUtils.isEmpty(checkTaskQueue)) {
+      return Set.of();
+    }
+    return Arrays.stream(checkTaskQueue.split(",")).collect(Collectors.toSet());
+  }
+
+  /**
+   * Retrieve and parse the sync workflow task queue configuration.
```
pmossman, line 394: nit: should say "discover" instead of "sync" in comment

airbyte-workers/src/test/java/io/airbyte/commons/temporal/scheduling/RouterServiceTest.java
```diff
@@ -40,21 +40,21 @@ class RouterServiceTest {
...
 
-    Mockito.when(mTaskQueueMapper.getTaskQueue(Geography.AUTO)).thenReturn(US_TASK_QUEUE);
-    Mockito.when(mTaskQueueMapper.getTaskQueue(Geography.US)).thenReturn(US_TASK_QUEUE);
-    Mockito.when(mTaskQueueMapper.getTaskQueue(Geography.EU)).thenReturn(EU_TASK_QUEUE);
+    Mockito.when(mTaskQueueMapper.getTaskQueue(Geography.AUTO, TemporalJobType.SYNC)).thenReturn(US_TASK_QUEUE);
+    Mockito.when(mTaskQueueMapper.getTaskQueue(Geography.US, TemporalJobType.SYNC)).thenReturn(US_TASK_QUEUE);
+    Mockito.when(mTaskQueueMapper.getTaskQueue(Geography.EU, TemporalJobType.SYNC)).thenReturn(EU_TASK_QUEUE);
   }
 
   @Test
```
pmossman, line 48: You could turn this into a @ParameterizedTest where it takes in an @EnumSource for `TemporalJobType` and verifies that it works correctly for everything, instead of just SYNC. Here's an example test that uses an EnumSource: https://github.com/airbytehq/airbyte-platform-internal/blob/42ac3ae31fb7a3ce16862597c637ff63e8c307f1/cloud-config/cloud-config-persistence/src/test/java/io/airbyte/cloud/config/persistence/billing/BillingPersistenceTest.java#L201-L225

pmossman (approved):
Looks good to me! I left some minor nits and suggestions

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
xiaohansong: Yes - going to be, because at the stage of discover/check we will only have workspace Id available to query which region we should use. Will add a test.

airbyte-workers/src/test/java/io/airbyte/commons/temporal/scheduling/RouterServiceTest.java
xiaohansong: I guess problem is we are not doing routing logic for all TemporalJobType enums.

I think testing sync only at RouterServiceTest is fine - the other tests can be tested within taskQueueMapper logic.

pmossman: Oh, the @EnumSource takes in a list of the values you want to test: https://www.baeldung.com/parameterized-tests-junit-5#3-enum

pmossman: But yeah, not critical to cover everything here, I'm mostly just trying to spread awareness of ParameterizedTest because of how useful it is!

PR #21825, opened 2023-01-24
Revert back to erroring out on a null cursor value

### LINKED ISSUE #17150: Database Sources: Handle when a cursor value is NULL

rodireich:
## What
Put back the original change that stopped a sync when a null cursor value is discovered.
This will reverse the temporary fix we put in [#oncall/1159](https://github.com/airbytehq/oncall/issues/1159#issuecomment-1379106478).


ryankfu (approved):
LGTM

rodireich:

rodireich:

Issue #19698, opened 2022-11-22
[Low-code CDK] Deprecate the JsonSchema component name in favor of JsonSchemaFileLoader


brianjlai:
## Tell us about the problem you're trying to solve
**BREAKING CHANGE**

We had originally named the component that handles loading the JSON schema files located in the `/schemas` directory `JsonSchema`. However, this name makes it a little ambiguous from the DX perspective as to what is actually being done. We have instead opted to rename the component `JsonFileSchemaLoader` for clarity. Under the hood, the component has been renamed, but there are still some connectors using the old name `JsonSchema`. During Hacktoberfest as a temporary stopgap, we changed `class_types_registry.py` to resolve both names to the component, but we should really only support just the new name.

## Describe the solution you would like

A sample of the manifest using the new component name:
```
schema_loader:
  type: JsonFileSchemaLoader
  file_path: "./source_sendgrid/schemas/{{ options['name'] }}.yaml"
```

We don't need a migration because there are less than 20 usages of `JsonSchema` and it can be grepped and replaced with `type: JsonFileSchemaLoader` in less than 5 minutes.

Any documentation related to json_schema should be removed and any code examples should be modified to reflect the various request options being defined the requester itself.

The low-code generator templates don't have `json_schema` defined so they should not need to be modified.

## Acceptance Criteria

- [ ] Specifying JsonSchema in the manifest should result in an error being thrown for an unknown type
- [ ] The relevant usages of existing connectors have had `type: JsonSchema` replaced with `type: JsonFileSchemaLoader`
- [ ] Existing documentation and code samples should reflect JsonFileSchemaLoader instead of JsonSchema

## Implementation Approach
We've already implemented the changes to the schema loader, all that should need to be changed is removing the old `JsonSchema` name from the map in `class_types_registry.py`

PR #21832, opened 2023-01-25
[Low-Code CDK] Remove JsonSchema type in favor of JsonSchemaFileLoader

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/19698

## What
Earlier we decided to rename JsonSchema to JsonFileSchemaLoader since this more accurately described what the component was doing. To retain backwards compatibility we supported both in the `declarative_component_schema.yaml` w/ JsonSchema acting as sort of an alias.

This is a breaking change so we are packaging this up in the promotion of low-code to Beta.

## How
When we first renamed the component, we allowed for both `JsonSchema` and `JsonFileSchemaLoader` in the string literal. But now that we're deprecating this alias, its as simple as modifying the schema YAML and rerunning the model generation.

Given the few usages of JsonSchema, I just performed a replace all from PyCharm for simplicity so no migration is included

## Recommended reading order

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -717,7 +717,7 @@ definitions:
       - type
     properties:
       type:
-        enum: [JsonFileSchemaLoader, JsonSchema] # TODO As part of Beta, remove JsonSchema and update connectors to use JsonFileSchemaLoader
+        enum: [JsonFileSchemaLoader]
       file_path:
         type: string
       $parameters:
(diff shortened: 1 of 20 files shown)


maxi297 (commented):
A couple questions:
* Should this be updated: https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/class_types_registry.py#L82
* Is there documentation to update?

maxi297 (approved):
Let me approve this PR since I don't think my questions are a blocker for anything

clnoll (approved):
LGTM pending any changes based on Maxime's comments.

brianjlai:
@maxi297 ah yes nice find, I missed that one thanks! And no docs changes needed, I didn't see any usages of `JsonSchema` (or JsonFileSchemaLoader` for that matter when I implemented

PR #21853, opened 2023-01-25
🪟 🎉 Add "Request Schema/ERD" button for schemas and ERDs

letiescanciano:
From the Growth team, we are working on a couple experiments with the goal of improving docs and increase activation.

However, we see value on having this "feature" released without a FF.

While we work on the two other experiments [Show schemas with sample data](https://airtable.com/appIuY0uKPVnk8TWT/tbl2SxXnUwf6fVCWS/viw9frYvld7ks7aNo/recDdVKOliH4DFrgM?blocks=hide) and [Show ERDs](https://airtable.com/appIuY0uKPVnk8TWT/tbl2SxXnUwf6fVCWS/viww5hSnXY1UUWb6w/recPs9W5sK2OWKFDq?blocks=hide), we want to start measuring user interest on this features.

Also, we won't have sample data/schemas or ERDs for all sources, so this will also serve as a fallback for those sources within the experiments.

Demo: https://www.loom.com/share/7336a26aa85b46b28aacaff1907bc5dc

Screen Shot 2023-01-25 at 4 46 17 PM

Screen Shot 2023-01-25 at 4 46 22 PM


airbyte-webapp/src/components/ui/StepsMenu/StepsMenu.tsx
```diff
@@ -24,9 +25,9 @@ const Content = styled.div`
   font-family: ${({ theme }) => theme.regularFont};
 `;
 
-export const StepsMenu: React.FC<StepMenuProps> = ({ data, onSelect, activeStep, lightMode }) => {
+export const StepsMenu: React.FC<StepMenuProps> = ({ data, onSelect, activeStep, lightMode, className }) => {
   return (
-    <Content>
+    <Content className={className}>
```
letiescanciano, line 30: Needed to provide a different alignment

letiescanciano, line 30: Needed to provide a different alignment

airbyte-webapp/src/views/Connector/ConnectorCard/ConnectorCard.tsx
```diff
@@ -90,7 +90,8 @@ export const ConnectorCard: React.FC<ConnectorCardCreateProps | ConnectorCardEdi
   const [isFormSubmitting, setIsFormSubmitting] = useState(false);
   const [logsVisible, setLogsVisible] = useState(false);
 
-  const { setDocumentationUrl, setDocumentationPanelOpen } = useDocumentationPanelContext();
+  const { setDocumentationUrl, setDocumentationPanelOpen, setSelectedConnectorDefinition } =
```
letiescanciano, line 93: ATM, we need to add information about the connectorDefinition to check if it's a source, but we'll use this more extensively in the next PR, as we need this information to actually fetch the schema of the selected source

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/DocumentationPanel.tsx
```diff
@@ -58,20 +66,60 @@ export const DocumentationPanel: React.FC = () => {
...
+        <StepsMenu
+          className={styles.stepsContainer}
+          lightMode
+          data={tabs}
+          onSelect={setActiveTab}
+          activeStep={activeTab}
+        />
+      )}
+
+      {match(activeTab)
```
letiescanciano, line 100: I've added `ts-pattern` because I think it's a great library and allow us to have exahustive pattern matching and also cleaner/more readable code

letiescanciano:
@timroes I had to run `eslint --fix` because of all the`import/order` warnings preventing the build.
It's all in the second commit, and I'm sure it has to be a better way to do it, I just couldn't find it

timroes:
Before going into actual reviewing some design suggestion: I would prefer if we could waste as less space for the header as possible. So I feel we might not want to have "Setup Guide" and then tabs below that. Could we maybe just have the tabs (centered) and remove the "Setup Guide" header. I'd suggest instead we just rename the first tab from "Doc" to "Setup Guide". What do you think?

letiescanciano:
@timroes any chance you could review this one?

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/ResourceNotAvailable/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { FormattedMessage } from "react-intl";
+
+import { Button } from "components/ui/Button";
+import { Text } from "components/ui/Text";
+
+import { SourceDefinitionRead } from "core/request/AirbyteClient";
+import { useDocumentationPanelContext } from "views/Connector/ConnectorDocumentationLayout/DocumentationPanelContext";
+
+import styles from "./ResourceNotAvailable.module.scss";
+import octaviaWorker from "../octavia-worker.png";
```
timroes, line 12: @Upmitt Could we make sure to get a proper illustration here (vector art!). We don't want to add new `png` images into the app, so would appreciate if we can exchange that via a proper illustration.

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/ResourceNotAvailable/index.tsx
```diff
@@ -0,0 +1,55 @@
+import { faArrowDown } from "@fortawesome/free-solid-svg-icons";
```
timroes, line 1: We tend not to put actual code in the `index` files but just reexport it (since it's harder to navigate to those files if the actual code is inside the index). Could this please be either in `ResourceNotAvailable/ResourceNotAvailable.tsx` (and just reexported from `ResourceNotAvailable/index.ts`) or just in a `ResourceNotAvailable.tsx` inside `ConnectorDocumentationLayout`.

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/DocumentationPanel.tsx
```diff
@@ -30,7 +32,12 @@ export const prepareMarkdown = (markdown: string, env: "oss" | "cloud"): string
 export const DocumentationPanel: React.FC = () => {
   const { formatMessage } = useIntl();
   const config = useConfig();
-  const { setDocumentationPanelOpen, documentationUrl } = useDocumentationPanelContext();
+  const { setDocumentationPanelOpen, documentationUrl, selectedConnectorDefinition } = useDocumentationPanelContext();
+  const isSource = Object.hasOwn(selectedConnectorDefinition, "sourceDefinitionId");
```
timroes, line 36: Should use `isSourceDefinition` from `src/core/domain/connector/source.ts`.

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/ResourceNotAvailable/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+          <Button
+            variant="primary"
+            onClick={() => {
+              trackRequest({
+                sourceDefinitionId: (selectedConnectorDefinition as SourceDefinitionRead).sourceDefinitionId,
+                connectorName: selectedConnectorDefinition.name,
+                requestType: activeTab,
+              });
+              setRequested(true);
+            }}
```
timroes, line 47:
```suggestion
            onClick={() => {
              if (isSourceDefinition(selectedConnectorDefinition)) {
                trackRequest({
                  sourceDefinitionId: selectedConnectorDefinition.sourceDefinitionId,
                  connectorName: selectedConnectorDefinition.name,
                  requestType: activeTab,
                });
              }
              setRequested(true);
            }}
```

That way we don't have manual type casts here, that might break our code if someone accidentally would use this component while a destination connector is open, and thus not realy on implicit knowledge and flow of a completely other component.

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/ResourceNotAvailable/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+import styles from "./ResourceNotAvailable.module.scss";
+import octaviaWorker from "../octavia-worker.png";
+import { useAnalyticsTrackFunctions } from "../useAnalyticsTrackFunctions";
+
+export const ResourceNotAvailable: React.FC<
+  React.PropsWithChildren<{
+    activeTab: "erd" | "schema";
+    isRequested: boolean;
+    setRequested: (val: boolean) => void;
+  }>
```
timroes, line 20: Please make this a `interface ResourceNotAvailableProps`, which we usually always do whenever there's more than one property.

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/ResourceNotAvailable/ResourceNotAvailable.module.scss
```diff
@@ -0,0 +1,10 @@
+@use "scss/variables";
+
+.requestContainer {
```
timroes, line 3: We should add an `padding: variables.$spacing-xl` to this still, since at the moment the resize panel doesn't add any padding to it's content, thus this can overlap not very nicely:

screenshot-20230213-133839

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/ResourceNotAvailable/ResourceNotAvailable.module.scss
```diff
@@ -0,0 +1,10 @@
...
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: variables.$spacing-lg;
+  justify-content: center;
+  height: 80vh;
```
timroes, line 9: I think a bit hacky solution, and only "pseudo centering" it, and likely we should rewrite the documentation panel to actually use flex box, but I think that can be done in a later PR.

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/ResourceNotAvailable/ResourceNotAvailable.module.scss
```diff
@@ -0,0 +1,10 @@
+@use "scss/variables";
+
+.requestContainer {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: variables.$spacing-lg;
+  justify-content: center;
```
timroes, line 8: We should use our `FlexContainer` component for this instead.

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/ResourceNotAvailable/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+      {isRequested ? (
+        <Text size="lg">
+          <FormattedMessage id="sources.request.thankYou" />
+        </Text>
+      ) : (
+        <>
+          <Text size="lg">
+            <FormattedMessage id="sources.request.prioritize" />
+          </Text>
+          <FontAwesomeIcon icon={faArrowDown} />
```
timroes, line 37: Design nitpick: I think we should just remove this icon. I don't think we have something with this arrow leading to the button anywhere else, and I feel even without this looks how we'd want it:

screenshot-20230213-134507

airbyte-webapp/src/locales/en.json
```diff
@@ -281,6 +281,13 @@
   "sources.emptySchema": "Schema is empty",
   "sources.noSources": "Source list is empty",
   "sources.sourceSettings": "Source Settings",
+  "sources.documentationPanel.tabs.setupGuide": "Setup Guide",
+  "sources.documentationPanel.tabs.schema": "Schema",
+  "sources.documentationPanel.tabs.erd": "ERD (Entities Relationship Diagram)",
+  "sources.request.button.schema": "Request schema for this source",
+  "sources.request.button.erd": "Request ERD for this source",
+  "sources.request.prioritize": "We're still building this feature - if you'd like us to prioritize this source.",
```
timroes, line 289: I think the wording here is a bit off (grammar wise), but will refer to @nataliekwong as a native speaker.

"We're still building this feature - if you'd like us to prioritize this source." feels like language wise something is off?

I'd suggest something like: "We're still building this feature for you. Let us know if you're interested in {ERD diagrams|Schemas} for this source."?

airbyte-webapp/src/locales/en.json
```diff
@@ -281,6 +281,13 @@
   "sources.emptySchema": "Schema is empty",
   "sources.noSources": "Source list is empty",
   "sources.sourceSettings": "Source Settings",
+  "sources.documentationPanel.tabs.setupGuide": "Setup Guide",
+  "sources.documentationPanel.tabs.schema": "Schema",
+  "sources.documentationPanel.tabs.erd": "ERD (Entities Relationship Diagram)",
```
timroes, line 286: I think ERD is usually "Entity Relationship Diagram" (singular not plural for Entity). Wikipedia writes is as "Entity-relationship diagram", but that feels pretty weird having a hyphen between 1 and 2 but not 3, and since I found a lot of places that write it as "Entity Relationship Diagram" out there, I think we can also skip that hyphen (but still should go with singular, not Entities).

timroes (changes requested):
Left some comments about code, wording and design.

One larger question/concern I'd have: This is basically a no-op feature, which we just use to gather analytics data about requirements. I feel as such it should not be in OSS and only be in Cloud, since it feels weird putting something in to our OSS product we just use for analytics purposes. @nataliekwong happy about your input there?

If we only would put it in cloud, I think we could just put it behind an LaunchDarkly experiment with default value `false` in code (I don't think we'd need to create a FeatureItem for it, since this is more a temporary state).

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/DocumentationPanel.tsx
letiescanciano: You are right, in this case it's not a huge benefit  but still I think it makes code so much easy to read. You can see in a nutshell the different branches that depends on a given variable value and also if you are forgetting any possibility. While  the standard way I feel works for smaller use cases or cases where we just create new components and we can keep cases small in 2-3 lines.

Where I see this being helpful and helping as establishing a clean pattern is managing request states. Eg: https://github.com/airbytehq/airbyte/pull/21861/files#diff-c298f10fce8256b2dc3a4f848e47ed70b7b574f0ba0bc8793e8fccb238d1d8eeR25 

I can def present it in the next chapter sync to gather opinions from the team.
This is the first time I've "added" a new pattern so wasn't really sure about the approach that I should follow.

airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/ResourceNotAvailable/index.tsx
letiescanciano: Makes sense! Will follow this approach from now on

letiescanciano:
> Left some comments about code, wording and design.
> 
> One larger question/concern I'd have: This is basically a no-op feature, which we just use to gather analytics data about requirements. I feel as such it should not be in OSS and only be in Cloud, since it feels weird putting something in to our OSS product we just use for analytics purposes. @nataliekwong happy about your input there?
> 
> If we only would put it in cloud, I think we could just put it behind an LaunchDarkly experiment with default value `false` in code (I don't think we'd need to create a FeatureItem for it, since this is more a temporary state).

My 2 cents here, I think there's no harm on having it also on OSS as it might peak interest from OSS users also, and who knows, maybe a path for them to try Cloud

nataliekwong:
> > Left some comments about code, wording and design.
> > One larger question/concern I'd have: This is basically a no-op feature, which we just use to gather analytics data about requirements. I feel as such it should not be in OSS and only be in Cloud, since it feels weird putting something in to our OSS product we just use for analytics purposes. @nataliekwong happy about your input there?
> > If we only would put it in cloud, I think we could just put it behind an LaunchDarkly experiment with default value `false` in code (I don't think we'd need to create a FeatureItem for it, since this is more a temporary state).
> 
> My 2 cents here, I think there's no harm on having it also on OSS as it might peak interest from OSS users also, and who knows, maybe a path for them to try Cloud but I defer to @nataliekwong

I think it's worth passing by the founders as to whether we should experiment in OSS before getting signals from Cloud. I plan to write up a short doc on this by Thursday.

airbyte-webapp/src/locales/en.json
timroes: cc @nataliekwong Would appreciate your input on this still, because we really don't want to trust my lousy English skills :D

nataliekwong: > I think it's worth passing by the founders as to whether we should experiment in OSS before getting signals from Cloud. I plan to write up a short doc on this by Thursday.

I chatted with Jean briefly about the open question of whether we put experiments in OSS. His feedback was that we should not experiment in OSS, but instead pass our learnings into OSS once we’ve established something works in Cloud.

In this specific example, if we see that users do readily request example data or ERDs, then that feature could move to OSS. But without validating that users would find that useful, we shouldn’t push it to OSS. @letiescanciano @timroes

nataliekwong:
> > > Left some comments about code, wording and design.
> > > One larger question/concern I'd have: This is basically a no-op feature, which we just use to gather analytics data about requirements. I feel as such it should not be in OSS and only be in Cloud, since it feels weird putting something in to our OSS product we just use for analytics purposes. @nataliekwong happy about your input there?
> > > If we only would put it in cloud, I think we could just put it behind an LaunchDarkly experiment with default value `false` in code (I don't think we'd need to create a FeatureItem for it, since this is more a temporary state).
> > 
> > 
> > My 2 cents here, I think there's no harm on having it also on OSS as it might peak interest from OSS users also, and who knows, maybe a path for them to try Cloud but I defer to @nataliekwong
> 
> I think it's worth passing by the founders as to whether we should experiment in OSS before getting signals from Cloud. I plan to write up a short doc on this by Thursday.

I chatted with Jean briefly about the open question of whether we put experiments in OSS. His feedback was that we should not experiment in OSS, but instead pass our learnings into OSS once we’ve established something works in Cloud.

In this specific example, if we see that users do readily request example data or ERDs, then that feature could move to OSS. But without validating that users would find that useful, we shouldn’t push it to OSS. @letiescanciano @timroes

airbyte-webapp/src/locales/en.json
nataliekwong: Suggestion: We're still building this feature for you. Let us know if you're interested in a {Entity Relationship Diagram (ERD) | schema} for this source.

I wonder if people actually know what these are.  I was thinking we could actually send them to an example of what one looks like, but also don't want them to leave their existing flow.  Is it possible to show an example, but greyed out / shrouded in the background so that it's clear it's not to be actually used?

PR #21857, opened 2023-01-25
🪟 🎨 Remove z-index rule from Switch component handle

edmundito:
## What
Removes the z-index from the Switch handle CSS rule. This is an old rule that may no longer be needed. Tested and could not find any situations where the Switch was rendering incorrectly.


teallarson (approved):
👍 checked behavior manually on connections list table, connection item page (title and table), behind a modal, behind an overlay, and in stream details panel that interactivity and animation are as expected.

This seems like a safe change to make.

dizel852 (approved):
LGTM 👍 
Tested locally - have not found any issues

PR #21858, opened 2023-01-25
Source Stripe: fix stream schemas

roman-yermilov-gl:
## What
https://github.com/airbytehq/oncall/issues/1364

## How
Fix stream schemas:
 - start -> start_date
 - discount.start_date -> discount.start 
   for **subscription** stream,
 - add status_transitions for **invoices** stream

### DIFF
--- a/airbyte-integrations/connectors/source-stripe/source_stripe/schemas/subscriptions.json
+++ b/airbyte-integrations/connectors/source-stripe/source_stripe/schemas/subscriptions.json
@@ -115,7 +115,7 @@
         "customer": {
           "type": ["null", "string"]
         },
-        "start_date": {
+        "start": {
           "type": ["null", "integer"]
         },
         "object": {
--- a/airbyte-integrations/connectors/source-stripe/source_stripe/schemas/subscriptions.json
+++ b/airbyte-integrations/connectors/source-stripe/source_stripe/schemas/subscriptions.json
@@ -11,7 +11,7 @@
     "livemode": {
       "type": ["null", "boolean"]
     },
-    "start": {
+    "start_date": {
       "type": ["null", "integer"]
     },
     "items": {
--- a/airbyte-config/init/src/main/resources/seed/source_definitions.yaml
+++ b/airbyte-config/init/src/main/resources/seed/source_definitions.yaml
@@ -1680,7 +1680,7 @@
 - name: Stripe
   sourceDefinitionId: e094cb9a-26de-4645-8761-65c0c425d1de
   dockerRepository: airbyte/source-stripe
-  dockerImageTag: 0.1.40
+  dockerImageTag: 1.0.0
   documentationUrl: https://docs.airbyte.com/integrations/sources/stripe
   icon: stripe.svg
   sourceType: api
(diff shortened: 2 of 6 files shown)


airbyte-integrations/connectors/source-stripe/source_stripe/schemas/subscriptions.json
```diff
@@ -11,7 +11,7 @@
     "livemode": {
       "type": ["null", "boolean"]
     },
-    "start": {
+    "start_date": {
```
evantahler, line 14: I'm interested to know why a failing test didn't catch this before?  Is it because we have yet to test this stream with expected_records?

roman-yermilov-gl: We can potentially test it with expected records if we strictly compare records. If not, then we allowed by default to have additional fields that not specified in schema. We currently have task to migrate to stripe tests to strict level (assigned to me) so I can check it during migration

mfsiega-airbyte: > We currently have task to migrate to stripe tests to strict level (assigned to me) so I can check it during migration

Does this imply that we _intend_ to have all the fields in the schema, but we just missed this one? (And the future test work will help us catch it faster?)

roman-yermilov-gl: I just have remembered that migration did not catch it properly with expected records :)

evantahler: > I just have remembered that migration did not catch it properly with expected records :)

Can you please file a bug about this then?

evantahler (approved):

lazebnyi (approved):

mfsiega-airbyte:
One question: the report here mentioned at least one other field missing (`status_transitions` in the Invoices stream). Shall we add that here? And more generally, do we have a mechanism to ensure that all fields are included? (I see a thread about testing?)

airbyte-integrations/connectors/source-stripe/source_stripe/schemas/subscriptions.json
davydov-d: @evantahler @mfsiega-airbyte I'm not sure this is actually a bug. A missing field may be not spotted by the test because we guarantee that `additionalProperties` field always equals `True`. Do you think we should override this requirement when testing if data matches its schema? if yes, I will create an issue for that.
Another thing is that we have lots of streams not filled with data so the test is just skipped. This should be fixed for GA connectors when they all have a test strictness level set to `high`

davydov-d:
Bumped a major version because of schema changes

davydov-d:
> One question: the report here mentioned at least one other field missing (`status_transitions` in the Invoices stream). Shall we add that here? And more generally, do we have a mechanism to ensure that all fields are included? (I see a thread about testing?)

@mfsiega-airbyte I added `status_transitions` field within this PR. Regarding your question about a mechanism to ensure that all fields are included, please see my comment above. I think it is possible for GA connectors when they all have high test strictness level. Additionally, we should probably report a bug for source-acceptance-tests package (but I'm not sure about this)

davydov-d:
will merge this one when @klsoper approves

PR #21861, opened 2023-01-25
🪟 🧪 [Experiment] Add discover schema in source

letiescanciano:
We want to try if showing schema information along with some sample data will increase activation.

Schema information will be fetched from the Node.js BE (hosted in Vercel). Having schemas there, will help us to test with more and add more sample data if needed without needing a release

Airtable task: https://airtable.com/appIuY0uKPVnk8TWT/tbl2SxXnUwf6fVCWS/viw9frYvld7ks7aNo/recDdVKOliH4DFrgM?blocks=hide

Demo + approach explanation: https://www.loom.com/share/c7059b4d83dd4196b35e4fdc9be09542

Env variables PR: https://github.com/airbytehq/airbyte-cloud/pull/4201

Depends on: #21853

Screen Shot 2023-01-25 at 6 32 45 PM


airbyte-webapp/src/views/Connector/ConnectorDocumentationLayout/SchemaViewer/useGetSourceSchema.ts
```diff
@@ -0,0 +1,41 @@
...
+}): Promise<JSONSchema4> => {
+  const path = `${apiUrl}/source_definition/${sourceDefinitionId}/discover_schema`;
+
+  const response = await fetch(path);
+
+  if (!response.ok) {
+    throw new Error("Schema not found");
+  }
+  return response.json();
+};
```
letiescanciano, line 21: I'm not sure if this is the best approach. I'm throwing an error here because it seems that, when using `fetch` with `react-query` getting a 404 doesn't set `{status: 'error'}`  on `useQuery` response.

Happy to see if there are cleaner ways to do this

timroes:
@letiescanciano Is this PR meant to be against another branch of yours and not master? We have quiet a lot of changes in master that will conflict with this PR, so I'm uncertain if it wouldn't be better merging this against master so you'll already need to adjust to those changes as part of this PR (and not just the one merging `leti/add-request-schemas-erds-in-source-docs` in `master`).

letiescanciano:
> @letiescanciano Is this PR meant to be against another branch of yours and not master? We have quiet a lot of changes in master that will conflict with this PR, so I'm uncertain if it wouldn't be better merging this against master so you'll already need to adjust to those changes as part of this PR (and not just the one merging `leti/add-request-schemas-erds-in-source-docs` in `master`).

This depends on https://github.com/airbytehq/airbyte/pull/21853 if you could review that one first it would be great, I will then merge that one (after your comments) and rework this based on the changes on master.
Also, you mentioned this would need som rework because of the changes in the config provider, are those already merged?

PR #21862, opened 2023-01-25
🪟 🧪 Experiment to use shorter setup guides

timroes:
## What

Adds an experiment to use shorter setup guides for some connectors, to see how this affects activation rate on those connectors.

## How

Simply add some `.inapp.md` files for those two connectors we want to test and if the flag is on, we check if the documentation we want to load is one of those two and then load the other markdown file instead. I excluded all `.inapp.md` documents in Docusaurus so they won't render into the build system. I decided to keep them still in the doc folder, so we can leverage the existing copying mechanism for those markdown files.

cc @letiescanciano


lmossman (approved):
LGTM, tested locally and it properly rendered the short setup guides for just those two connectors

PR #21868, opened 2023-01-25
🪟 🔧 Update eslint and add dependencies explicitly

timroes:
## What

This PR is in preparation to switch us to `pnpm`. It does the following three things:

* Adding all dependencies we're using explicitly to `package.json` instead of relying on them being pulled in by other packages.
* This came with an update to the `import/order` rule, which sorts relative imports slightly differently now. Thus I wanted to pull this into a separate PR, since I `--fix`ed all files in validation of this.
* Due to the new dependencies installed, `npm` (un)flattens packages in a way that a prior typing conflict is now resolved, thus removed two `@ts-expect-error` statements. (Makes the point perfectly why we should switch to `pnpm` :D)

Only the `package.json` was touched manually, the rest is all just auto fixed by ESLint.


joeykmh (approved):
Did a quick scan of the eslint fixes and it all looks fine 🙂

letiescanciano (approved):
Thanks for taking care of this!

PR #21869, opened 2023-01-25
🪟 🔧 Don't request destination/source definition in `<ConnectionInfoCard />` directly, use  `useConnectionFormService()` instead

### LINKED ISSUE #21661: Don't request destination definition in ConnectionInfoCard
The ConnectionInfoCard currently makes a call to get the [Destination Definition](https://github.com/airbytehq/airbyte/issues/21659). This is available out of the ConnectionFormService and should be retrieved from there.

Additionally we should reference the source and destination icons off of the top-level source and destination objects instead of the definitions. This is a non-change but it’s better to reference them there in case we’re able to remove the definition calls in the future.

dizel852:
## What
Resolves https://github.com/airbytehq/airbyte/issues/21661

## How
Don't request destination/source definition in `<ConnectionInfoCard />` directly, use  `useConnectionFormService()` instead


m19berger (approved):
LGTM!

airbyte-webapp/src/components/connection/ConnectionInfoCard/ConnectionInfoCard.tsx
```diff
@@ -65,9 +63,9 @@ export const ConnectionInfoCard: React.FC = () => {
         >
           <ConnectorCard
             connectionName={destination.destinationName}
-            icon={destinationDefinition?.icon}
+            icon={destDefinition?.icon}
```
edmundito, line 66: From the comment in the [issue](https://github.com/airbytehq/airbyte/issues/21661): it should be using the icon source or destination  directly instead of the definition.

edmundito (changes requested):
Code is looking good, but noticed one more requirement...

airbyte-webapp/src/components/connection/ConnectionInfoCard/ConnectionInfoCard.tsx
dizel852: Oh, thanks for pointing me out 👍 Fixed

edmundito (approved):
Tested locally. LGTM

PR #21870, opened 2023-01-25
🪟 🧹 Credits page component cleanup

teallarson:
## What
Tackles some preparation for the Billing Insights PR.  
- migrates (nearly) all components on the credits page to scss modules
  - I did not migrate UsageCell because it would have been slightly more involved and is being fully replaced during https://github.com/airbytehq/airbyte/issues/22100
- implements components from our UI library to reduce amount of custom styling
- adds types for new endpoint data

Minimal visual changes due to using our standard spacing variables.  Though one visual improvement is that the connection column now displays correctly! 

Screenshot 2023-01-31 at 2 53 15 PM

## How
✨ magic ✨ (jk, it's code.)

## Recommended reading order
Not critical, but it may make most sense to start with the `CreditsPage` and work your way down the component tree.


airbyte-webapp/src/packages/cloud/lib/domain/cloudWorkspaces/types.ts
```diff
@@ -17,6 +19,8 @@ export interface CloudWorkspace {
 
 export interface CreditConsumptionByConnector {
   connectionId: string;
+  connectionName: string;
+  status: ConnectionStatus;
```
tealjulia, line 23: New data added to this endpoint in https://github.com/airbytehq/airbyte-platform-internal/pull/4164

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/ConnectionCell.tsx
```diff
@@ -12,43 +14,30 @@ interface ConnectionCellProps {
...
-        <Icon icon={sourceIcon} />
-        {sourceDefinitionName}
-      </Connector>
-      <Connector>
-        <Arrow icon={faArrowRight} />
-        <Icon icon={destinationIcon} />
-        {destinationDefinitionName}
-      </Connector>
-    </>
+    <FlexContainer justifyContent="space-between">
```
tealjulia, line 24: would love to use a grid setup for this and other bits related to table layout/sizing... but hesitate since none of our tables use it (yet)

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/CreditsUsage.module.scss
```diff
@@ -0,0 +1,16 @@
+@use "scss/variables";
+
+.empty {
+  height: 100%;
+  padding-bottom: variables.$spacing-xl;
+}
+
+.chartWrapper {
+  width: 100%;
+  height: 260px;
```
tealjulia, line 10: we don't currently have any standardized height variables, but i wonder if we should.

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/ConnectionCell.module.scss
```diff
@@ -0,0 +1,8 @@
+.connectorItem {
+  width: calc(100% - 30px / 2);
+}
+
+.arrowItem {
+  flex-basis: 30px;
+  min-width: 30px;
```
josephkmh, line 7: Is the idea to make this flex child exactly 30px? I would use `flex: 0 0 30px` here instead of setting width. Or is there a reason to set `min-width` explicitly?

Could we also add:

```
display: flex;
align-items: center;
justify-content: center;
```
So the arrow is nicely centered?

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/ConnectionCell.tsx
```diff
@@ -12,43 +14,30 @@ interface ConnectionCellProps {
...
+        <FlexContainer direction="row">
+          <ConnectorIcon icon={sourceIcon} />
+          <Text size="lg">{sourceDefinitionName}</Text>
+        </FlexContainer>
+      </FlexItem>
+      <FlexItem className={styles.arrowItem}>
+        <ArrowRightIcon />
+      </FlexItem>
+      <FlexItem className={styles.connectorItem}>
+        <FlexContainer direction="row">
```
josephkmh, line 35: Could we `alignItems: center` here and on line 26 to visually center the icon & text?

joeykmh (approved):
Nice! Much cleaner 🙂 just left a couple CSS ideas, but approving since they are not blocking.

PR #21873, opened 2023-01-25
🪟 🧹 Round credits values on credit consumption page

teallarson:
## What
closes https://github.com/airbytehq/airbyte-cloud/issues/4121

Rounds credit data to the approriate place per [Intl standards](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat).

Screenshot 2023-01-25 at 2 31 38 PM

## How
Utilizes `<FormattedNumber />` from React-Intl in the table as well as in the bar graph component. 

## Recommended reading order
any

### DIFF
--- a/airbyte-webapp/src/components/ui/BarChart/BarChart.tsx
+++ b/airbyte-webapp/src/components/ui/BarChart/BarChart.tsx
@@ -1,4 +1,5 @@
 import React, { useMemo } from "react";
+import { FormattedNumber } from "react-intl";
 import {
   Bar,
   BarChart as BasicBarChart,
@@ -63,7 +64,12 @@ export const BarChart: React.FC<BarChartProps> = React.memo(({ data, legendLabel
         >
           <Label value={yLabel} fontSize={11} fill={chartTicksColor} fontWeight={600} position="top" offset={10} />
         </YAxis>
-        <Tooltip cursor={{ fill: chartHoverFill }} />
+        <Tooltip
+          cursor={{ fill: chartHoverFill }}
+          formatter={(value: number) => {
+            return [<FormattedNumber value={value} />, yLabel];
+          }}
+        />
         {legendLabels.map((barName, key) => (
           <Bar key={barName} dataKey={barName} fill={barChartColors[key]} />
         ))}
--- a/airbyte-webapp/src/components/ui/BarChart/BarChart.tsx
+++ b/airbyte-webapp/src/components/ui/BarChart/BarChart.tsx
@@ -67,7 +67,7 @@ export const BarChart: React.FC<BarChartProps> = React.memo(({ data, legendLabel
         <Tooltip
           cursor={{ fill: chartHoverFill }}
           formatter={(value: number) => {
-            return [<FormattedNumber value={value} />, yLabel];
+            return [<FormattedNumber value={value} maximumFractionDigits={2} minimumFractionDigits={2} />, yLabel];
           }}
         />
         {legendLabels.map((barName, key) => (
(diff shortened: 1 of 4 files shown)


airbyte-webapp/src/components/ui/BarChart/BarChart.tsx
```diff
@@ -63,7 +64,12 @@ export const BarChart: React.FC<BarChartProps> = React.memo(({ data, legendLabel
         >
           <Label value={yLabel} fontSize={11} fill={chartTicksColor} fontWeight={600} position="top" offset={10} />
         </YAxis>
-        <Tooltip cursor={{ fill: chartHoverFill }} />
+        <Tooltip
+          cursor={{ fill: chartHoverFill }}
+          formatter={(value: number) => {
+            return [<FormattedNumber value={value} />, yLabel];
```
tealjulia, line 70: This is non-intuitive.  Rechart renders this as:
`${yLabel}: ${value}`

They need to be returned in this order (`[value, name]`).  The separator is configurable (and we can skip the whole thing for a custom Tooltip if we so desire)

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx
```diff
@@ -142,7 +142,11 @@ const UsagePerConnectionTable: React.FC<UsagePerConnectionTableProps> = ({ credi
         accessor: "creditsConsumed",
         collapse: true,
         customPadding: { right: 0 },
-        Cell: ({ cell }: CellProps<FullTableProps>) => <UsageValue>{cell.value}</UsageValue>,
+        Cell: ({ cell }: CellProps<FullTableProps>) => (
+          <UsageValue>
+            <FormattedNumber value={cell.value} />
```
josephkmh, line 147: Did you consider whether to use `minimumFractionDigits`/`maximumFractionDigits` to always show a set number of decimal points? Not sure I have a strong opinion, just curious if that might look better for consistency (e.g. `0.00` instead of `0`)

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx
```diff
@@ -142,7 +142,11 @@ const UsagePerConnectionTable: React.FC<UsagePerConnectionTableProps> = ({ credi
         accessor: "creditsConsumed",
         collapse: true,
         customPadding: { right: 0 },
-        Cell: ({ cell }: CellProps<FullTableProps>) => <UsageValue>{cell.value}</UsageValue>,
+        Cell: ({ cell }: CellProps<FullTableProps>) => (
+          <UsageValue>
```
josephkmh, line 146: Could we maybe get rid of the styled component here while we're in this file? 👼

joeykmh (approved):
Nice, this has been such an eyesore for me 🙂 Only non-blocking comments from my side

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx
tealjulia: I may as well.  Had punted it due to new designs coming, but that'll make those changes easier to implement/review so 🤷‍♀️

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx
tealjulia: Agree.  I was also on the fence about it but the existing usage (credits remaining) didn't do that.  I'll swap them all out!

airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx
tealjulia: Yep!  There's still some things with alignment that I'd like to tackle, but those are in the `<Table/>` component and don't belong in this PR.

PR #21874, opened 2023-01-25
🪟🔧 use https instead of http for localhost:3000 in e2e webapp tests

lmossman:
## What
The switch to Vite caused local E2E tests to be unable to run successfully, as they were trying to use `http://localhost:3000`, instead of `https://localhost:3000`, which is now required.

## How
This PR fixes the issue by switching `http` to `https` for that baseUrl


joeykmh (approved):
✨

Issue #21866, opened 2023-01-25
[Low-Code CDK] Remove legacy factory and validation flow


maxi297:
## Tell us about the problem you're trying to solve
As part of using the migration to rely on declarative_component_schema.yaml to validate the manifest files, we wanted to make sure that the behaviour/validation of the config would be the same as before. Therefore, we kept the legacy factory [for validation purposes](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py#L192) and had a toggle switch for constructing the manifest in case of issue.

Now that we are pretty confident and that we are making changes to the API that makes the legacy and the new factory diverge, we can remove the legacy part.

## Describe the solution you’d like
Remove `DeclarativeComponentFactory`. This has a couple implications (non-exhaustive):
* Remove `construct_using_pydantic_models` switch in `ManifestDeclarativeSource.__init__`
* Remove manifest_declarative_source.generate_schema()
* Remove JsonSchemaMixin

## Additional context
This will be an enabler for #21559

PR #21878, opened 2023-01-25
Issue 21866 remove legacy factory and validation flow

maxi297:
## What
Addressed #21866 

## How
* Removing `construct_using_pydantic_models` from `ManifestDeclarativeSource` and clean things associated to it
* Remove method `ManifestDeclarativeSource.generate_schema()`
* Remove dataclasses-jsonschema dependency 

## Recommended reading order
1. First commit: clean ManifestDeclarativeSource validation
2. Second commit: remove dataclasses-jsonschema


airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -175,33 +125,18 @@ def _configure_logger_level(self, logger: logging.Logger):
...
-            raise ValidationError("Validation against auto-generated schema failed") from e
-
-        # Validates the connector manifest against the low-code component json schema
         try:
             raw_component_schema = pkgutil.get_data("airbyte_cdk", "sources/declarative/declarative_component_schema.yaml")
             declarative_component_schema = yaml.load(raw_component_schema, Loader=yaml.SafeLoader)
         except FileNotFoundError as e:
             raise FileNotFoundError(f"Failed to read manifest component json schema required for validation: {e}")
 
+        streams = self._source_config.get("streams")
```
brianjlai, line 134: Instead of this additional validation, it might be better to document in the schema YAML and rely on the validator to catch manifests that don't specify at least one stream. wdyt?
```
title: DeclarativeSource
type: object
...
properties:
  type:
    type: string
    enum: [DeclarativeSource]
  streams:
    type: array
    items:
      "$ref": "#/definitions/DeclarativeStream"
    minItems: 1
```

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -213,65 +148,6 @@ def _stream_configs(self, manifest: Mapping[str, Any]):
...
-                    return subclasses
-                else:
-                    return [field_type]
-        elif generic_type == list or generic_type == Union:
-            next_classes = []
-            for underlying_type in typing.get_args(field_type):
-                next_classes.extend(ManifestDeclarativeSource._get_next_expand_classes(underlying_type))
-            return next_classes
-        return []
-
```
brianjlai, line 274: I believe the `SchemaEncoder` class below is also only used by the method to `generate_schema()` we're deleting so that can get removed too. Just commenting here since I can't comment on unmodified lines

brianjlai (approved):
No blocking changes for approval, just a few comments here and there. Nice work!

Also, if we have sufficiently gotten rid of every `JsonSchemaMixin`, can we also remove `"dataclasses-jsonschema==2.15.1"` from `setup.py` since we don't need that library to generate the schema anymore?

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
maxi297: I think [this](https://github.com/airbytehq/airbyte/pull/20747#discussion_r1057401169) is still relevant and prevents us from doing this easily with the auto-generated schemas since [the solution provided by pydantic](https://docs.pydantic.dev/usage/schema/#unenforced-field-constraints) does not align with autogeneration.

maxi297:
> Also, if we have sufficiently gotten rid of every `JsonSchemaMixin`, can we also remove `"dataclasses-jsonschema==2.15.1"` from `setup.py` since we don't need that library to generate the schema anymore?

Yes, this was [part of the second commit](https://github.com/airbytehq/airbyte/pull/21878/commits/3c8c21a2ce39cb55cdcb1553a13b15fd7500a420#diff-4447bde589c0143e4d0c6475e78f9e945e464223a6ea052c5a7f5ee973461ba1L48)

clnoll (approved):
Love a good cleanup PR. LGTM!

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
brianjlai: ah okay yeah we can leave as is

Issue #15653, opened 2022-08-15
Improve scroll behavior of frontend


timroes:
The frontend currently has containers scrolling at weird positions, which can lead to scrollbars appearing in the middle of the page:

![screenshot-20220815-132242](https://user-images.githubusercontent.com/877229/184636368-c3a9fed6-6926-46c3-af08-b04cc54fdc41.png)

In general I think we should try to have the page (and not individual containers) being scrollable as much as possible. If we want to have sticky parts on the page, I think it's preferable using position: sticky/fixed for those, rather than having just a container/div scrollable somewhere within the page. If we have scrollable containers within the page, they should ideally be "as large as possible", e.g. in this case even with being just the container of the table scrollable it would be better if the container reaches the bottom of the page, so you can use your scrollwheel/trackpad anywhere inside the main part of the page.

Affected places (non exhaustive list):

* Horizontal scrollbar below connection table

octavia-squidington-iii:
cc @airbytehq/frontend

edmundito:
I just noticed the same issue on the settings. It looks like the bottom margin to prevent the content from overlapping banners is revealing that the scrolling is part of an `overflow-y: auto` somewhere.

![Screen Shot 2022-08-15 at 08 45 31](https://user-images.githubusercontent.com/168664/184637485-36131986-0b47-45b2-9d8c-b109ab591dff.png)

edmundito:
Here's another bad thing that could maybe be resolved with the changes here? When creating a connection and selecting a new source type, the dropdown expands the height of the contents causing the side bar to scroll:

![Screen Shot 2022-08-29 at 16 04 17](https://user-images.githubusercontent.com/168664/187288356-4978e6f8-382f-4f46-bb7c-e1cb046631c0.png)

PR #21880, opened 2023-01-25
🪟 🎨  Update replication settings page layout

### LINKED ISSUE #20969: Update connection page scrolling behavior

edmundito:
## What
Resolves airbytehq/airbyte#20969

Updates the replication settings page layout:

* Save/Cancel buttons are sticky to the bottom of the screen
* Only one scrollbar (instead of having the streams table sub-scroll)
* Keep the vertical layout when the table has a few rows
* Make headers sticky once more!
* In Cloud, the create and save buttons are pushed to the side of the intercom banner

The create connection page remains the same as before, but it now has the sticky headers. Changes to the connection create page are not part of the scope.

[Figma](https://www.figma.com/proto/liLQhcbpVHiEDW18kiXQe3/01_03_CONNECTIONS?page-id=857%3A24319&node-id=865%3A52984&viewport=124%2C-519%2C0.25&scaling=min-zoom&starting-point-node-id=865%3A52984) 

Current table:

https://user-images.githubusercontent.com/168664/219041715-ad8e7add-0047-440e-ba9a-4f9f35c0abb6.mov

Next table:

https://user-images.githubusercontent.com/168664/219041698-1387f7d4-afcb-4dde-8cee-fe6b91d691c2.mov

## How

The EditControls component is now wrapped in a custom container that makes it sticky to the bottom of the page. The container has to adjust margins to extend beyond its parent since the parent has 20px padding on the sides and button (thus, you may see some negative numbers in the CSS.)

To make the stream card have a minimum height that filled the rest of the page, many elements needed to be styled with `flex: 1` until the table could fit the size:

```
Card
  Section
    CatalogTree
      CatalogTreeBody
```

Finally, added custom classes to make the table scrollable only in the create connection page, the way it's always been. The rules are only applied if the mode is `create`.


flash1293:
@edmundito anything specific I should look out for? Not deeply familiar with the connection form

airbyte-webapp/src/components/connection/ConnectionForm/Section.module.scss
```diff
@@ -1,5 +1,11 @@
 @use "scss/variables";
 
+.flexHeight {
+  flex: 1;
+  display: flex;
+  flex-direction: column;
+}
+
```
josephkmh, line 8: Do we really want to stretch the stream table to the bottom if it only has a few rows? Looks kind of strange IMO:
image

airbyte-webapp/src/pages/connections/ConnectionPage/ConnectionPage.tsx
```diff
@@ -37,13 +38,22 @@ export const ConnectionPage: React.FC = () => {
...
 
   return (
     <ConnectionEditServiceProvider connectionId={connectionId}>
       <ResourceNotFoundErrorBoundary errorComponent={<StartOverErrorView />}>
-        <MainPageWithScroll headTitle={<ConnectionHeadTitle />} pageTitle={<ConnectionPageTitle />}>
+        <MainPageWithScroll
+          headTitle={<ConnectionHeadTitle />}
+          pageTitle={<ConnectionPageTitle />}
+          noBottomPadding={isReplicationPage}
+        >
```
josephkmh, line 56: nitpick: would be a bit more declarative and cleaner (IMO) to put `<MainPageWithScroll>` inside each route, then just set `noBottomPadding={true}` in the `ConnectionRoutePaths.Replication` route. Then all the styling logic for that route is in one place, and we don't have to do a potentially brittle check based on the current URL.

joeykmh (changes requested):
Nice! I really like this change, the `EditControls` should always be visible 👏 

My main feedback is that I think we should do any conditional styling related to Intercom via a `FeatureItem` instead of `isCloudApp()` to be more transparent.

airbyte-webapp/src/components/connection/CatalogTree/CatalogTreeBody.module.scss
```diff
@@ -1,8 +1,8 @@
 .container {
-  max-height: 600px;
-  overflow: unset;
   min-width: fit-content;
   padding-right: 15px;
 
-  --webkit-overlay: true;
+  &.fixedHeight {
+    max-height: 600px;
+  }
```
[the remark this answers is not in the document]
edmundito: Yes, potentially but when I reviewed the design with Nico it looks like we needed to think more about so I had to return the support for the table scroll until that design is ready.

airbyte-webapp/src/components/connection/ConnectionForm/Section.module.scss
edmundito: That's how we planned it out in the design.

airbyte-webapp/src/components/common/MainPageWithScroll/MainPageWithScroll.tsx
```diff
@@ -13,17 +13,30 @@ interface MainPageWithScrollProps {
...
     <div className={styles.page}>
       <div>
         {headTitle}
         {pageTitle}
       </div>
       <div className={styles.contentContainer}>
-        <div className={classNames(styles.content, { [styles.cloud]: isCloudApp() })}>{children}</div>
+        <div
+          className={classNames(styles.content, {
+            [styles.cloud]: isCloudApp(),
```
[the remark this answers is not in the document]
edmundito: I started working on this but it's not as simple because in cloud, the intercom hook depends on having a context enabled, which depends on having intercom installed, so it involves some refactoring. What do you think about having an other function such as `hasIntercom()` that detects if `window.Intercom` is defined?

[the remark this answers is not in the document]
josephkmh: I'm not sure you need the intercom hook to achieve this - just `useFeatureService()`. Something like this:

```jsx
const isIntercomEnabled = useFeature(FeatureItem.IntercomChat);

return (
        <div
          className={classNames(styles.content, {
            [styles.cloud]: isIntercomEnabled
          })}>
)
```

Then just put `FeatureItem.IntercomChat` in `defaultCloudFeatures` here: https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/hooks/services/Feature/constants.ts#L12

airbyte-webapp/src/components/connection/ConnectionForm/Section.module.scss
josephkmh: Looks a bit empty to me, but ok 🙂

airbyte-webapp/src/components/common/MainPageWithScroll/MainPageWithScroll.tsx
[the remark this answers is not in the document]
edmundito: That part is easy. If we're going to support intercom from the feature service, the app needs to be updated only to enable intercom when the feature is available. The issue is that the current cloud code assumes that intercom is always available, so some changes need to be made there to wrap around the `react-use-intercom` library.

edmundito:
@flash1293 added you to see if you had any feedback on how this was done CSS-wise.

flash1293:
Not sure I prefer the new behavior over the old one (very subjective of course):

https://user-images.githubusercontent.com/1508364/215743671-40da6b3f-6666-4bc8-b519-b27c5ef435c6.mp4

 The top corner cuts off and the bottom corner looks like it scrolls _within_ the card which IMHO is breaking the visual metaphor of the card.

If there is not enough space to fit the streams tile from the beginning it's even weirder and there are still two scrollbars:

https://user-images.githubusercontent.com/1508364/215746093-2024b8d4-e190-4ce8-9b83-d103d5a0c327.mp4

That's probably out of scope, but what about making the sticky footer a visual panel that's on top of the scrollable page like this:

https://user-images.githubusercontent.com/1508364/215749798-670a55c6-b111-4f8a-920e-bb60eac55c08.mp4


In terms of CSS alone I can't come up with a much better approach as this design is inherently not really consistent in what's the scroll container.

teallarson (commented):
I agree with Joe RE: the scroll behavior and border radii in the card feeling off.

I think this would look better without the right padding in the streams table if we are moving the scrollbar out of here.  The edge does make this section visually match the edge of the searchbar... but we don't do that on the left (because of the icons we display there probably) which feels odd.

Screenshot 2023-01-31 at 8 36 14 AM

flash1293:
Another option I'm seeing is to not have separate cards for "Transfer", "Streams" and "Activate the streams you want to sync", but having one large card with some kind of separator between the sections - this way the scrolling would become trivial (it's just the contents of the card that scroll) which is probably also easier to maintain than this special case.

edmundito:
Thanks for the feedback! There's a lot of great thinking there, and I'd like to take some of this and discuss it separately from this PR. There are several ongoing threads about how to improve this page overall. Another challenge is that create connection is out of scope for this and will remain as-is for the time being, so making modifications to the form here may also impact the create connection page since they share the same form.

@tealjulia 
> I think this would look better without the right padding in the streams table if we are moving the scrollbar out of here. 
 - Fixed

edmundito:
I followed up with Nico to consolidate several improvements that can be made to this page. This will be tracked in #22313

joeykmh (approved):
@edmundito code looks good to me, so happy to approve this PR. I am a little concerned that from this one PR we are opening two more (rather large) issues:
- https://github.com/airbytehq/airbyte/issues/21981
- https://github.com/airbytehq/airbyte/issues/22313

I think especially the changes to the scroll behavior could benefit from another round of design/UX feedback from the team. IMO we should strive for consistency in scroll behavior across all pages, so we should either continue this pattern everywhere or conform to what we're doing on other pages. Maybe we could set up a demo of this in the next tech talk?

tealjulia:
Agree with the sentiment above!  Looks good to me but scroll behavior app-wide could use an evaluation.

We do have [this](https://github.com/airbytehq/airbyte/issues/15653) old placeholder issue for scroll behaviors.

airbyte-webapp/src/components/connection/ConnectionForm/EditControls.tsx
```diff
@@ -25,24 +29,23 @@ const EditControls: React.FC<EditControlProps> = ({
...
 }) => {
+  const isButtonDisabled = hidden || isSubmitting || (!dirty && !enableControls);
   return (
-    <FlexContainer justifyContent="flex-end" alignItems="center" direction="row" gap="lg" className={styles.container}>
+    <FlexContainer
+      justifyContent="flex-end"
+      alignItems="center"
+      direction="row"
+      gap="lg"
+      className={classNames(styles.container, { [styles.cloud]: isCloudApp(), [styles.hidden]: hidden })}
```
timroes, line 41: I feel moving those buttons just when the intercom button is present feels a bit weird, and like a pattern that will cause us eventually more work trying to maintain. Are we taking the intercom button at all into consideration when making designs @Upmitt?

Would it make sense in OSS just to accept that the form buttons are not "fully right aligned" and instead keep the space for where we expect in cloud to have the intercom button?

tealjulia:
Github isn't happy with me trying to upload a screenshot, but I'm wondering two things here:
1. Will we be adding a blurred border to the bottom inside edge of the streams card to indicate that there is scrollable content?
2. Should the CatalogTreeTableRow have a padding-right for the text?  The background color looks better this way, but the text can now be flush with the right edge of the card, which seemed abrupt.

krishnaglick (approved):
LGTM!

krishnaglick:
This is failing on a snapshot diff

edmundito:
I finally have this ready for final final final review. I need to update the description, but there are of changes since we last spoke, including:

* Merges some of the other PRs related to removing the table horizontal scroll and adding a soft edge to the top of the scroll
* Makes the bottom edge a soft scroll and removes the fake card
* Add sticky headers to both tables (yes, old and new!)
* Fixes layout issues when tables are deleted

There are 18 different combinations to test this table 😓 

Pages
* Create connection
* Edit replication settings
* Deleted page replication tables

Modes
* Cloud
* OSS

New table feature
* Enabled
* Disabled
* Not Cloud

joeykmh (approved):
Changes LGTM!

PR #21881, opened 2023-01-25
Source Trello: fixed `ixUpdate` board stream datatype

### LINKED ISSUE #21480: Source Trello: ixUpdate field has wrong datatype
## Environment
- **Airbyte version**:  v0.40.28

## Current Behavior
In trello connector, there is a field called ixUpated which has the wrong data type i.e ["null", "integer"] in board stream in discovery whereas from the backend API response, ixUpdated is returning as string.

## Expected Behavior
In discovery of board stream, ixUpated should be ["null", "string"]

Yes, I am submitting the PR for this issue.

zilehuda:
## What
FIxed #21480 

## How
- Changed the data type of `ixUpdate` from `["null", "integer"]` to `["null", "string"]`


airbyte-integrations/connectors/source-trello/source_trello/schemas/boards.json
```diff
@@ -61,7 +61,7 @@
       "type": ["null", "string"]
     },
     "ixUpdate": {
-      "type": ["null", "integer"]
```
marcosmarxm, line 64: From Trello documentation:
https://developer.atlassian.com/cloud/trello/rest/api-group-boards/#api-boards-id-get

image

image

Maybe you can show us an example with string from your dataset.

marcosmarxm (changes requested):
Please check my comment.

airbyte-integrations/connectors/source-trello/source_trello/schemas/boards.json
zilehuda: @sajarin 
This endpoint is used in the stream: https://developer.atlassian.com/cloud/trello/rest/api-group-members/#api-members-id-boardstars-idstar-delete
reference: https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-trello/source_trello/source.py#L86
Seems like the Trello doc is also not upto-date.

Here you can see in the screenshot that ixUpdate is a string.

ixupdated-wrong

sajarin:
Closing because this was apparently fixed in #24045

PR #21882, opened 2023-01-25
Bmoric/controller utest

benmoriceau:
## What
Add a UTest to check the return code of our API endpoints.

## How
Create a UTEST (actually, it is more an integration test) that checks that the endpoint we're call returns the expected code. This PR only contains the HealthCheck. Other PR(s) will add the other endpoints.


airbyte-server/src/test/java/io/airbyte/server/apis/MicronautHealthCheck.java
```diff
@@ -0,0 +1,34 @@
...
+@Requires(property = "mockito.test.enabled",
+          defaultValue = StringUtils.TRUE,
+          value = StringUtils.TRUE)
+@Requires(env = {Environment.TEST})
+@SuppressWarnings("PMD.JUnitTestsShouldIncludeAssert")
+class MicronautHealthCheck extends BaseControllerTest {
+
+  @Test
+  void testHealth() throws IOException {
+    Mockito.when(healthCheckHandler.health())
```
benmoriceau, line 28: The lack of this mocking is the reason why we were runnign in a NotFound error. It was not because the route was not register but because the response was null which lead micronaut to return a notFound error.

airbyte-server/src/test/java/io/airbyte/server/apis/BaseControllerTest.java
```diff
@@ -0,0 +1,260 @@
...
+
+@Slf4j
+@MicronautTest
+@Requires(property = "mockito.test.enabled",
+          defaultValue = StringUtils.TRUE,
+          value = StringUtils.TRUE)
+@SuppressWarnings("PMD.AbstractClassWithoutAbstractMethod")
+public abstract class BaseControllerTest {
+
+  AttemptHandler attemptHandler = Mockito.mock(AttemptHandler.class);
```
jdpgrailsdev, line 45: Did you mean to make everything package scoped in this class?

jdpgrailsdev:
@benmoriceau Looks like there are some formatting issues preventing a clean PR build.

airbyte-server/src/test/java/io/airbyte/server/apis/BaseControllerTest.java
benmoriceau: Yes, It is a recommendation of pmd. don't think that the base will be use outside of the package. I need to remove the public scope.

gosusnp (approved):

airbyte-server/src/test/java/io/airbyte/server/apis/BaseControllerTest.java
```diff
@@ -0,0 +1,260 @@
...
+import org.jooq.DSLContext;
+import org.mockito.Mockito;
+
+@Slf4j
+@MicronautTest
+@Requires(property = "mockito.test.enabled",
+          defaultValue = StringUtils.TRUE,
+          value = StringUtils.TRUE)
+@SuppressWarnings("PMD.AbstractClassWithoutAbstractMethod")
+abstract class BaseControllerTest {
```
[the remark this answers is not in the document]
benmoriceau: Done

benmoriceau:
Tagged as ready to ship since it is only UTests

Issue #20401, opened 2022-12-12
Roll out AvailabilityStrategy - slowly for GA connectors, as default for others 


erohmensing:
After having implemented https://github.com/airbytehq/airbyte/pull/19977, in order to better handle 403 errors that cause stream failures to turn into sync failures, we should make sure that GA connectors have the latest CDK installed with the default HttpAvailabilityStrategy available to streams that inherit the HttpStream class. This is an important step that we can do before #17853, which is likely to take more time to implement.

P0: GA connectors that have given us problems in the past: Grab from [this document](https://airbyte.metabaseapp.com/question/1130-permission-failures-by-connector-by-quarter)

P1. Other GA connectors 

P2. Beta/Alpha connectors

Open questions
* if/when do we want to enforce that connectors have this dependency updated? Shouldn't be an issue in the future, but for now we want to get the default behavior in. When moving to GA? When moving to beta? 
   * Idea - maybe must have the default HttpAvailabilityStrategy for beta, must implement own AvailabilityStrategy for GA?
* Most of these will have build fails due to low strictness. How do we handle this, especially if potentially bumping the CDK by a lot? (depends on both dependency and when the source was built last - `docker run pip list`)

evantahler:
Grooming: 

* Publish 1 Beta connector to test that everything works
* Publish all the GA connectors with the new CDK after this work is complete
* leave the rest of the connectors unpublished.

erohmensing:
Notes on rollout: 

The edge case we weren't handling but now are appeared due to issues in pinterest and github's stream_slices methods for substreams. While I can fix those specifically before enabling the default for those sources, I do think it makes sense to turn things on gradually in case there are more strange edge cases we need to address.

Here's what I'm thinking for rollout:
* In the initial PR that introduces AvailabilityStrategy, don't set a default AvailabilityStrategy for HttpStreams.
* In a separate PR, enable the default for HttpStreams and set the default for GA connectors (and Gitlab, since we know it is already not going to work due to per-record permission issues) to None . Create issues to implement an AvailabilityStrategy (default or otherwise) for GA connectors. GL already has doing that for the 10 most impacted connectors in their Q1a plans, so those get worked on first.
* For beta and alpha connectors other than gitlab, they will get the default next time they're published - if there are issues, address them as they come (I'll be on call next week - convenient!)

erohmensing:
Sources with setup.py files (python sources) that are GA: 
```
source-amazon-ads
source-amplitude
[...]

PR #21888, opened 2023-01-25
set availabilitystrategy for GA connectors + gitlab with httpstreams explicitly to `None`

erohmensing:
## What
Now that the HttpAvailabilityStrategy is published in the CDK, we want to turn it on as default for HttpStreams. In order to prevent issues in GA connectors if there is unexpected incompatibility or issues, we want to turn on the HttpAvailability (or a substitute custom AvailabilityStrategy) 

## How
I found a [list of GA python connectors](https://github.com/airbytehq/airbyte/issues/20401#issuecomment-1401121130) by scripting the following: 
- get a list of connectors from the `airbyte-integrations/connectors` folder
- filter for sources only
- filter for sources which have `setup.py` files somewhere in their folder
- filter for python sources which are marked as `generally-available`

For each of those, if the GA python source was using `HttpStream`s and was an `AbstractSource`, I set the AvailabilityStrategy explicitly to `None` by doing the following: 
* added the `availability_strategy` to the base stream class for that source's streams, set to `None`
* bumped the version in the Dockerfile
* added a changelog entry 

There are 31 of these, + gitlab = 32, so there should be 96 changed files.

Note that this doesn't change how the connector will behave - the default for all AvailabilityStrategies at the moment is `None`. This is being done in preparation for https://github.com/airbytehq/airbyte/pull/21924, where the default will be set to `HttpAvailabilityStrategy`, which will switch it on for all alpha/beta connectors minus gitlab. 

## Pre-merge
- [x] Test all connectors
- [x] Update changelog date for all connectors
- [x] Merge in master and handle merge conflicts (connectors that have been merged in the meantime) 
- [x] Publish all connectors

## Follow up 
- [x] Create an issue for each GA connector to implement an availabilitystrategy (default or custom)
- [x] Create an issue for Gitlab to implement the httpavailabilitystrategy for all streams possible (those which don't have per-record 403s), or implement a better custom availabilitystrategy


airbyte-integrations/connectors/source-facebook-marketing/source_facebook_marketing/streams/base_streams.py
```diff
@@ -40,6 +41,10 @@ class FBMarketingStream(Stream, ABC):
     # entity prefix for `include_deleted` filter, it usually matches singular version of stream name
     entity_prefix = None
 
+    @property
+    def availability_strategy(self) -> Optional["AvailabilityStrategy"]:
```
bechurch, line 45: Hey @erohmensing !

Just reviewing for my own understanding

So the gist of this is
1. We have a new default strategy for HttpStream
2. This PR explicitly sets the `availability_strategy` to the old default of `HttpStream` for subclasses of `HttpStream`.
3. This is so that when we release the new default strategy (in another PR) we dont accidentally introduce a regression

Is that all correct?

evantahler (approved):

PR #21891, opened 2023-01-25
Low Code CDK: only allow `name` and `primary_key` on `DeclarativeStream` components

### LINKED ISSUE #21560: [Low-Code CDK] DeclarativeStream should define name and primary_key and these fields should be removed from other declarative components
**BREAKING CHANGE**

It is confusing to have duplicative `name` and `primary_key` fields across the `DeclarativeStream` `SimpleRetriever`, and `HttpRequester` components. 

## Describe the solution you would like

The `name` and `primary_key` fields should be consolidated onto the `DeclarativeStream` component which is the top most field where the fields are relevant. Child components like `SimpleRetriever` and `HttpRequester` can inherit this field.

```
events_stream:
  type: DeclarativeStream
  name: events
  primary_key: id
  retriever:
    type: SimpleRetriever
    requester:
      type: HttpRequester
      url_base: "https://test.something.io/v1/"
      http_method: "GET"
      authenticator:
        type: NoAuthentication
```

## Acceptance Criteria
- [ ] The declarative_component_schema.yaml should be updated w/ `name` and `primary_key` removed from other components that are not DeclarativeStream
- [ ] SimpleRetriever and HttpRequester still have `name` and `primary_key`, but inherited from the parent component 
- [ ] Either a migration is included removing all instances of `SimpleRetriever.name`, `SimpleRetriever.primary_key` and `HttpRequest.name`
- [ ] Documentation for the SimpleRetriever and HttpRequester has been updated w/ `name` and `primary_key` removed 
- [ ] An associated ticket to update the connector builder UI frontend has been filed

## Implementation Approach

- When initializing the DeclarativeStream we should initialize a child SimpleRetriever component using it's associated model. We should also pass the `name` and `primary_key` fields from the DeclarativeStream model via kwargs
- A common pattern that was used to reduce duplication of fields on existing connectors was to define `name` and `primary_key` under `$options`. This should now be unnecessary, but could be a slightly more complex migration. This would be considered out of scope, but maybe worth seeing what the lift would be. The other concern is that the JsonFileSchemaLoader path would need to continue to reuse `options.name`.

clnoll:
Closes https://github.com/airbytehq/airbyte/issues/21560.

## What
Avoid duplication/simplify UX by removing the `name` and `primary_key` fields from the `SimpleRetriever` and `HttpRequester` components, which can inherit it as needed from `DeclarativeStream`.

## Recommended reading order
1. First commit - updates to the CDK & documentation
2. Second commit - migration of existing manifests to remove `name` and `primary_key` from non-`DeclarativeStream` components.

- [x] `./gradlew validateSourceYamlManifest` is passing after updating the manifests
- [x] Documentation which references the component is updated as needed

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -589,15 +589,12 @@ definitions:
     type: object
     required:
       - type
-      - name
       - path
       - url_base
     properties:
       type:
         type: string
         enum: [HttpRequester]
-      name:
-        type: string
       path:
         type: string
       url_base:
@@ -984,15 +981,10 @@ definitions:
         anyOf:
           - "$ref": "#/definitions/CustomRequester"
           - "$ref": "#/definitions/HttpRequester"
-      name:
-        type: string
-        default: ""
       paginator:
         anyOf:
           - "$ref": "#/definitions/DefaultPaginator"
           - "$ref": "#/definitions/NoPagination"
-      primary_key:
-        "$ref": "#/definitions/PrimaryKey"
       stream_slicer:
         anyOf:
           - "$ref": "#/definitions/CartesianProductStreamSlicer"
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -596,10 +596,13 @@ definitions:
         type: string
         enum: [HttpRequester]
       path:
+        description: The specific endpoint to fetch data from for a resource.
         type: string
       url_base:
+        description: The root of the API source.
         type: string
       authenticator:
+        description: Authenticator component that defines how to authenticate to the source.
         anyOf:
           - "$ref": "#/definitions/ApiKeyAuthenticator"
           - "$ref": "#/definitions/BasicHttpAuthenticator"
@@ -609,11 +612,13 @@ definitions:
           - "$ref": "#/definitions/NoAuth"
           - "$ref": "#/definitions/SessionTokenAuthenticator"
       error_handler:
+        description: Error handler component that defines how to handle errors.
         anyOf:
           - "$ref": "#/definitions/DefaultErrorHandler"
           - "$ref": "#/definitions/CustomErrorHandler"
           - "$ref": "#/definitions/CompositeErrorHandler"
       http_method:
+        description: The HTTP method used to fetch data from the source (can be GET or POST).
         anyOf:
           - type: string
           - type: string
@@ -976,16 +981,20 @@ definitions:
         type: string
         enum: [SimpleRetriever]
       record_selector:
+        description: Component that describes how to extract records from a HTTP response.
         "$ref": "#/definitions/RecordSelector"
       requester:
+        description: Requester component that describes how to prepare HTTP requests to send to the source API.
         anyOf:
           - "$ref": "#/definitions/CustomRequester"
           - "$ref": "#/definitions/HttpRequester"
       paginator:
+        description: Paginator component that describes how to navigate through the API's pages.
         anyOf:
           - "$ref": "#/definitions/DefaultPaginator"
           - "$ref": "#/definitions/NoPagination"
       stream_slicer:
+        description: StreamSlicer component that describes how to partition the stream, enabling incremental syncs and checkpointing.
         anyOf:
           - "$ref": "#/definitions/CartesianProductStreamSlicer"
           - "$ref": "#/definitions/CustomStreamSlicer"
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -468,16 +468,7 @@ definitions:
         default: ""
       primary_key:
         definition: The primary key of the stream
-        anyOf:
-          - type: string
-          - type: array
-            items:
-              type: string
-          - type: array
-            items:
-              type: array
-              items:
-                type: string
+        "$ref": "#/definitions/PrimaryKey"
         default: ""
       schema_loader:
         definition: The schema loader used to retrieve the schema for the current stream
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -404,7 +404,7 @@ class ModelToComponentFactory:
         )
 
     def create_declarative_stream(self, model: DeclarativeStreamModel, config: Config, **kwargs) -> DeclarativeStream:
-        retriever = self._create_component_from_model(model=model.retriever, config=config)
+        retriever = self._create_component_from_model(model=model.retriever, config=config, name=model.name, primary_key=model.primary_key)
 
         cursor_field = self._get_cursor_field_from_stream_slicer(model)
 
@@ -681,8 +681,10 @@ class ModelToComponentFactory:
             parameters=model.parameters,
         )
 
-    def create_simple_retriever(self, model: SimpleRetrieverModel, config: Config, **kwargs) -> SimpleRetriever:
-        requester = self._create_component_from_model(model=model.requester, config=config)
+    def create_simple_retriever(
+        self, model: SimpleRetrieverModel, config: Config, *, name: str, primary_key: Optional[Union[str, List[str], List[List[str]]]]
+    ) -> SimpleRetriever:
+        requester = self._create_component_from_model(model=model.requester, config=config, name=name)
         record_selector = self._create_component_from_model(model=model.record_selector, config=config)
         url_base = model.requester.url_base if hasattr(model.requester, "url_base") else requester.get_url_base()
         paginator = (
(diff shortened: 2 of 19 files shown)


airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -507,7 +507,7 @@ def create_http_requester(self, model: HttpRequesterModel, config: Config, **kwa
         )
 
         return HttpRequester(
-            name=model.name,
+            name=model.parameters.get("name", "") if model.parameters else "",
```
girarda, line 510: does this mean the name has to be set in $parameters? This seems like a non-obvious restriction

brianjlai, line 510: I don't think that we should be making an assumption that `$parameters` will be populated with fields like `name` or `primary_key`. The main reason we see so many connector manifests where `$parameters` contained these fields was because they had to be duplicated across all multiple components. So we could have reusable interpolation pointing to one place. Ideally we should actually be deriving the `name` from the DeclarativeStream which we would want to be the single place its defined. 

We can do this by using the `kwargs` parameter while parsing the DeclarativeStream model to take `model.name` and pass it down to SimpleRetriever and HttpRequesters like such:
`retriever = self._create_component_from_model(model=model.retriever, config=config, **{"name": model.name, "primary_key": model.primary_key})`. And the child components can get them from the kwargs or set "". So that way when we are constructing components, we are actually truly referencing the stream's field and not a theoretically optional component like `$parameters`

Once we stop this field duplication that  your PR is addressing, we should be able to start constructing manifests like this:
```
New Way:
projects_stream:
  name: "projects" # we will pass this value to the requester
  primary_key: "id"
  retriever:
    $ref: "#/definitions/retriever"
    requester:
      type: HttpRequester
      url_base: "{{ config['base_url'] or 'https://app.posthog.com'}}/api/"
      http_method: "GET"
      authenticator:
        type: NoAuth
  $parameters:
    path: "projects"
  
Old Way:
projects_stream:
  retriever:
    $ref: "#/definitions/retriever"
    requester:
      type: HttpRequester
      name: "{{ parameters['name'] }}" # we reference $parameters to avoid duplicate definitions
      url_base: "{{ config['base_url'] or 'https://app.posthog.com'}}/api/"
      http_method: "GET"
      authenticator:
        type: NoAuth
  $parameters:
    name: "cohorts"
    primary_key: "id"
    path: "cohorts"
```

Hopefully that makes sense, but happy to discuss this in more detail as well

brianjlai (changes requested):
The main thing to discuss is how we are trying to reuse the the values defined on the `DeclarativeStream` model instead of on `$parameters`.

A few additional points that didn't really fit inline:

- Was this done via script or manual? Not too many usages I see so either makes sense
- Pending our discussion about how we reuse `name` and `primary_key`, we can also modify the templates used when creating a new low code connector. That would mean "promoting" the fields out of $parameters and defining them directly on the stream itself. And then reused by subcomponents. For example:
```
  customers_stream:
    name: "customers"
    primary_key: "id"
    $parameters:
      path: "/example"
```

Same for the template that we provide via the connector builder server too.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
clnoll: Makes sense @brianjlai! I'd temporarily mentally rewritten how param propagation worked and assumed that the top-level parent keys would be passed down there. Fixing it up now.

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -602,15 +602,12 @@ definitions:
...
     required:
       - type
-      - name
       - path
       - url_base
     properties:
       type:
         type: string
         enum: [HttpRequester]
-      name:
```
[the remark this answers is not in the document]
clnoll: 👍 done.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
clnoll: Changed the approach per the discussion below.

clnoll:
@brianjlai the manifest changes were done manually, as there weren't too many. I made the suggested changes if you want to rereview.

airbyte-integrations/connectors/source-posthog/source_posthog/manifest.yaml
```diff
@@ -49,8 +46,8 @@ definitions:
   projects_stream:
     $ref: "#/definitions/base_stream"
     $parameters:
-      name: "projects"
       path: "projects"
+    name: "projects"
```
brianjlai, line 50: Wanted to point out an example of where extracting `name` or (`primary_key`) from the `$parameters` block might have some unintended side effects that may break a connector. This would work for the stream, retriever, and requester components. But this is probably going to break the `schema_loader` component too because it relies on the `name` in the parameters block to build the `file_path`. 

I'm actually of the mindset that we can leave many of the connectors with the `name` and `primary_key` under the declarative stream's `$parameters` because they are technically still correct. Albeit they're not the most idiomatic, but I can see a lot of little edge cases where removing them could break something not immediately obvious.

We should still remove the fields from the requester and retriever like you are already doing because the fields no longer exist there, but I was just talking about the `$parameters` block.

brianjlai (changes requested):
I think we're still missing the propagating `name` from the retriever to the requester during object instantiation and I think its worth discussing the migration where we extract `name`/`primary_key` from the `$parameters` block because that might get tricky

And last suggestion was for future connectors that get built, we could modify the template `manifest.yaml.hbs` so the `name` and `primary_key` are defined on the stream itself instead of in `$parameters`. Then future connectors will look nicer even if we don't migrate all the existing ones.

airbyte-integrations/connectors/source-posthog/source_posthog/manifest.yaml
clnoll: Thanks for the example @brianjlai. I've reverted the change that had removed `name` and `primary_key` from `$parameters`. I've also removed `name` and `primary_key` as top-level field for these manifests, since it's redundant, but let me know if you'd like to keep it in and I'll revert the last commit, e.g. to encourage people to put `name` and `primary_key` directly on the stream.

clnoll (commented):
Thanks @brianjlai I've updated the manifests and template, and fixed the `create_http_requester` function.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -709,9 +711,9 @@ def create_simple_retriever(self, model: SimpleRetrieverModel, config: Config, *
                 parameters=model.parameters,
             )
         return SimpleRetriever(
-            name=model.name,
+            name=name,
             paginator=paginator,
-            primary_key=model.primary_key.__root__ if model.primary_key else None,
```
brianjlai, line 714: This was unrelated to your exact change, but during the review I noticed that we're not actually using the `PrimaryKey` definition in our schema for `DeclarativeStream.primary_key`. See here: https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml#L890-L902

For consistency, can we make it point to the `PrimaryKey` definition. And then we'll need to keep this ternary operator to properly resolve the value. Although maybe we can do that higher up in create_declarative_stream()` before we pass it to the children. wdyt?

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -515,7 +515,7 @@ def create_dpath_extractor(self, model: DpathExtractorModel, config: Config, **k
     def create_exponential_backoff_strategy(model: ExponentialBackoffStrategyModel, config: Config) -> ExponentialBackoffStrategy:
         return ExponentialBackoffStrategy(factor=model.factor, parameters=model.parameters, config=config)
 
-    def create_http_requester(self, model: HttpRequesterModel, config: Config, **kwargs) -> HttpRequester:
+    def create_http_requester(self, model: HttpRequesterModel, config: Config, *, name: str, **kwargs) -> HttpRequester:
```
brianjlai, line 518: Does this still work if we get rid of `**kwargs` so `create_http_requester(self, model: HttpRequesterModel, config: Config, *, name: str)`? That would be more consistent to some of the other components that receive additional parameters.

I think the named args pattern works pretty well and we don't need `**kwargs` since it's not used anywhere

brianjlai (approved):
a few small comments, but nothing that needs to block or requires another review. Nice work!

airbyte-integrations/connectors/source-posthog/source_posthog/manifest.yaml
brianjlai: sorry missed responding to this in the review. How you have it is good! Moving forward with the updated template, it should guide people to having `name` and `primary_key` at the stream level and not in `$parameters`.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
clnoll: Makes sense to me! Updated.

PR #21892, opened 2023-01-25
:window: :art: Display free tag on release stage pills for users enrolled in the free connectors program

### LINKED ISSUE #21812: Display "free" status pills for Alpha/Beta connectors to users enrolled in the Free Connector Program
Part of https://github.com/airbytehq/airbyte/issues/21074.

As of https://github.com/airbytehq/airbyte/pull/21564, `ReleaseStageBadge` accepts a prop to display a new "free" label, intended for alpha/beta connectors as viewed by a user who is enrolled in the Free Connector Program. For expediency, this PR stopped short of automatically enabling the free label within the badge component iteself (doing so broke some brittle unit test setup, which proved to be at least a medium-sized time sink); this also lets us choose when to display the free pills with a bit more granularity (i.e., e.g., we can choose to show them in page A, but not dropdown B).

Where to show free connectors (links are to Nico's Figma designs):
- [credits page connection list](https://www.figma.com/proto/wihAJg5mkp61I7lLHs0IpV/FREE-CONNECTOR-PROGRAM?node-id=1095%3A16179&scaling=min-zoom&page-id=1080%3A15370&starting-point-node-id=1095%3A17886&show-proto-sidebar=1), in the cell which shows the connection's constituent connectors
- [connection page title](https://www.figma.com/proto/wihAJg5mkp61I7lLHs0IpV/FREE-CONNECTOR-PROGRAM?node-id=1104%3A28157&scaling=min-zoom&page-id=1080%3A15370&starting-point-node-id=1096%3A18573&show-proto-sidebar=1)
- [source](https://www.figma.com/file/wihAJg5mkp61I7lLHs0IpV/FREE-CONNECTOR-PROGRAM?node-id=1104%3A28754&t=9QX6RBlZoRu3gd8W-0) and destination pages
- check with Nico about whether to show the free pills in the source/destination dropdown: not included in designs, but perhaps _the_ key place where users are presented status pills in the UI

### LINKED ISSUE #21872: Refactor useFreeConnectorProgram() so it does not fire in OSS
Right now `useFreeConnectorProgram()` is called in several places in OSS which results in a `useQuery()` that ultimately fails. Due to the implementation, this failed query produces the desired effect (FCP UI is not rendered in OSS). But ideally we should not be calling this query at all in unsupported environments. 

Instead, we should create a new `FeatureItem` that represents the FCP. This feature is permanently disabled in OSS, so it's a better fit for the `FeatureService` than the `ExperimentService` anyway.

In addition, this will avoid conflicts with the ongoing config provider refactor: https://github.com/airbytehq/airbyte/pull/21456

cc @ambirdsall

ambirdsall:
## What
Displays the free badge on release stage pills if:
1) the user is enrolled in the free connector program
2) the release stage is "alpha" or "beta"

Closes https://github.com/airbytehq/airbyte/issues/21812.

## How, and some additional considerations
- Defines a new `FeatureItem.FreeConnectorProgram` flag and adds it to `defaultCloudFeatures`. As of now, there is no corresponding `featureService.FREE_CONNECTOR_PROGRAM` flag in LaunchDarkly, since all our dynamic override needs are met by the temporary experiment flag we're using to roll the feature out. 
- Extracts a new, self-contained `FreeTag` component to encapsulate all interaction with `useFreeConnectorProgram`, and by extension the cloud-only endpoint that backs it, behind a component boundary. This means OSS builds always get to short-circuit that API via conditionals like `{ useFeature(FeatureItem.FreeConnectorProgram) && <FreeTag releaseStage={releaseStage} /> }`; closes https://github.com/airbytehq/airbyte/issues/21872
  - Another benefit is to the test code: the initial definition relied a `showFreeTag` prop, largely as a way to avoid a big, pointless rewrite of some unit test setup code. This pattern dramatically simplified the extra setup needed, making it once again reasonable to centralize the "should the free tag render for this workspace/connector?" logic within the pills.
  - It's defined as a totally self-contained component within the `FreeConnectorProgram` directory, rather than as a helper component within `ReleaseStageBadge.tsx` in order to deal with the fact that there are two separate, almost-identical implementations of release stage pills. Obvious possible follow-up refactoring there.

## But how does it look
From the connection page title, embedded in a `ReleaseStageBadge`:
image

From the new source/destination dropdown, embedded in a `StageLabel` (is there any reason not to just use a `ReleaseStageBadge` here?):
image

### DIFF
--- a/airbyte-webapp/src/components/ReleaseStageBadge/ReleaseStageBadge.tsx
+++ b/airbyte-webapp/src/components/ReleaseStageBadge/ReleaseStageBadge.tsx
@@ -1,10 +1,12 @@
 import classNames from "classnames";
-import { FormattedMessage, useIntl } from "react-intl";
+import { FormattedMessage } from "react-intl";
 
 import { GAIcon } from "components/icons/GAIcon";
 import { Tooltip } from "components/ui/Tooltip";
 
 import { ReleaseStage } from "core/request/AirbyteClient";
+import { useFeature, FeatureItem } from "hooks/services/Feature";
+import { FreeTag } from "packages/cloud/components/experiments/FreeConnectorProgram";
 
 import styles from "./ReleaseStageBadge.module.scss";
 
@@ -15,16 +17,10 @@ interface ReleaseStageBadgeProps {
    * Whether to show a detailed message via a tooltip. If not specified, will be {@code true}.
    */
   tooltip?: boolean;
-  showFreeTag?: boolean;
 }
 
-export const ReleaseStageBadge: React.FC<ReleaseStageBadgeProps> = ({
-  stage,
-  small,
-  tooltip = true,
-  showFreeTag = false,
-}) => {
-  const { formatMessage } = useIntl();
+export const ReleaseStageBadge: React.FC<ReleaseStageBadgeProps> = ({ stage, small, tooltip = true }) => {
+  const fcpEnabled = useFeature(FeatureItem.FreeConnectorProgram);
 
   if (!stage) {
     return null;
@@ -37,13 +33,10 @@ export const ReleaseStageBadge: React.FC<ReleaseStageBadgeProps> = ({
       <div
         className={classNames(styles.pill, {
           [styles["pill--small"]]: small,
-          [styles["pill--contains-tag"]]: showFreeTag,
         })}
       >
         <FormattedMessage id={`connector.releaseStage.${stage}`} />
-        {showFreeTag && (
-          <span className={styles.freeTag}>{formatMessage({ id: "freeConnectorProgram.releaseStageBadge.free" })}</span>
-        )}
+        {fcpEnabled && <FreeTag releaseStage={stage} />}
       </div>
     );
 
--- a/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/FreeTag.tsx
+++ b/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/FreeTag.tsx
@@ -1,0 +1,23 @@
+import { useIntl } from "react-intl";
+
+import { ReleaseStage } from "core/request/AirbyteClient";
+
+import styles from "./FreeTag.module.scss";
+import { useFreeConnectorProgram } from "./hooks/useFreeConnectorProgram";
+
+interface FreeTagProps {
+  releaseStage: ReleaseStage;
+}
+
+// A tag labeling a release stage pill as free. Defined here for easy reuse between the
+// two release stage pill implementations (which should likely be refactored!)
+export const FreeTag: React.FC<FreeTagProps> = ({ releaseStage }) => {
+  const { enrollmentStatusQuery } = useFreeConnectorProgram();
+  const { isEnrolled } = enrollmentStatusQuery.data || {};
+  const { formatMessage } = useIntl();
+  const freeStages: Array<ReleaseStage | undefined> = ["alpha", "beta"];
+
+  return isEnrolled && freeStages.includes(releaseStage) ? (
+    <span className={styles.freeTag}>{formatMessage({ id: "freeConnectorProgram.releaseStageBadge.free" })}</span>
+  ) : null;
+};
--- a/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/FreeTag.tsx
+++ b/airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/FreeTag.tsx
@@ -15,9 +15,8 @@ export const FreeTag: React.FC<FreeTagProps> = ({ releaseStage }) => {
   const { enrollmentStatusQuery } = useFreeConnectorProgram();
   const { isEnrolled } = enrollmentStatusQuery.data || {};
   const { formatMessage } = useIntl();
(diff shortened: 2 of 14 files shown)


airbyte-webapp/src/components/ReleaseStageBadge/ReleaseStageBadge.tsx
```diff
@@ -37,13 +33,10 @@ export const ReleaseStageBadge: React.FC<ReleaseStageBadgeProps> = ({
       <div
         className={classNames(styles.pill, {
           [styles["pill--small"]]: small,
-          [styles["pill--contains-tag"]]: showFreeTag,
```
ambirdsall, line 40: One downside of putting all code shared with OSS on the other side of a component boundary: we no longer have a good way to conditionally change the `padding-right` from `6px` to `2px` based on whether a free tag is displayed. Instead, the `FreeTag` component uses a negative `margin-right`; that's a hacky solution, though, do let me know if you have a better idea.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/FreeTag.tsx
```diff
@@ -0,0 +1,23 @@
...
+  releaseStage: ReleaseStage;
+}
+
+// A tag labeling a release stage pill as free. Defined here for easy reuse between the
+// two release stage pill implementations (which should likely be refactored!)
+export const FreeTag: React.FC<FreeTagProps> = ({ releaseStage }) => {
+  const { enrollmentStatusQuery } = useFreeConnectorProgram();
+  const { isEnrolled } = enrollmentStatusQuery.data || {};
+  const { formatMessage } = useIntl();
+  const freeStages: Array<ReleaseStage | undefined> = ["alpha", "beta"];
```
josephkmh, line 18:
```suggestion
  const freeStages: ReleaseStage[] = ["alpha", "beta"];
```

I can't imagine why we would want `undefined` in this array?

airbyte-webapp/src/views/Connector/ConnectorForm/components/FrequentlyUsedConnectors/FrequentlyUsedConnectorsCard.test.tsx
```diff
@@ -2,14 +2,17 @@ import { fireEvent, render, waitFor } from "@testing-library/react";
...
 
 import { FrequentlyUsedConnectorsCard, FrequentlyUsedConnectorsCardProps } from "./FrequentlyUsedConnectorsCard";
 
 const renderFrequentlyUsedConnectorsComponent = (props: FrequentlyUsedConnectorsCardProps) =>
   render(
     <IntlProvider locale="en" messages={en}>
-      <FrequentlyUsedConnectorsCard {...props} />
+      <FeatureService features={defaultOssFeatures}>
+        <FrequentlyUsedConnectorsCard {...props} />
+      </FeatureService>
```
josephkmh, line 15: We also have `src/test-utils/tesutils.tsx` that exports its own `render()` which wraps the test component with `IntlProvider` and `FeatureService`, just FYI!

joeykmh (approved):
Looks good! Left a few ideas for improvements, but nothing blocking 👍

airbyte-webapp/src/components/ReleaseStageBadge/ReleaseStageBadge.tsx
ambirdsall: I do like that a bit better, but I'm inclined to save that for later, when consolidating the three implementations of stage pills (there's _another_ copy-paste `StageLabel` component in `cloud/components/experiments/` , used by the sign up page; I left that untouched for now since not-yet-created accounts will never be enrolled). If the overall design of the status pills changes, as things stand, it would be way too easy to overlook one of those.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/FreeTag.tsx
josephkmh: Not sure I quite follow - the change I suggested passes my TS checker just fine (so does `["alpha", "beta"].includes(releaseStage)`, for that matter). What error are you seeing there? I don't see how the type of `isEnrolled`

airbyte-webapp/src/views/Connector/ConnectorForm/components/FrequentlyUsedConnectors/FrequentlyUsedConnectorsCard.test.tsx
ambirdsall: Ooh, good tip. I'll make that change before merging, to save the next person from the ol' "test context is different from app context" surprise failure.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/FreeTag.tsx
ambirdsall: ...none, now :flushed:. I certainly remember only adding that `| undefined` in response to some `Argument of type ReleaseStage | undefined is not assignable to parameter of type ReleaseStage` error, but now it even type-checks if I go back to the inline array of strings I tried initially. I feel like I'm taking crazy pills, but I'll take the cleaner code, I guess.

PR #21897, opened 2023-01-26
Add WellKnownTypes to the Orchestrator

gosusnp:
## What
WellKnownTypes.json is only available in the workers, we need it in the orchestrator as well for kube based environments.

## How
* Add missing dependency in the worker build
* Generate WellKnownTypes.json for the orchestrator


airbyte-container-orchestrator/build.gradle
```diff
@@ -54,6 +57,18 @@ application {
     applicationDefaultJvmArgs = ['-XX:+ExitOnOutOfMemoryError', '-XX:MaxRAMPercentage=75.0']
 }
 
+task generateWellKnownTypes() {
```
edgao, line 60: let's put a comment here saying it's copy-pasted from the worker build.gradle. At some point I want to move this generation into airbyte-protocol directly (and just have it dump the yaml into `build/whatever` so that all modules can just use it directly).

airbyte-workers/build.gradle
```diff
@@ -143,6 +143,7 @@ task generateWellKnownTypes() {
 
 tasks.named("buildDockerImage") {
     dependsOn copyGeneratedTar
+    dependsOn generateWellKnownTypes
```
edgao, line 146: not sure I understand this - buildDockerImage already depends on copyGeneratedTar, which depends on generateWellKnownTypes. So isn't this dependency redundant?

edgao (approved):
couple nitpicky comments. lgtm though

airbyte-workers/build.gradle
gosusnp: Are we copying this file in the tar?

airbyte-container-orchestrator/build.gradle
gosusnp: updated

airbyte-workers/build.gradle
gosusnp: I suspect the monorepo project is going to impact those dependency, I think it might make sense to punt on dependency clean up for now.

edgao: welp. sounds good, lgtm

PR #21898, opened 2023-01-26
🎉 🪟  [Connector Builder] Resolve manifest before converting to builder form values

lmossman:
## What
Closes https://github.com/airbytehq/airbyte/issues/21610

Now that we have an endpoint in the connector builder server to resolve all $refs and $options in a manifest, this PR makes the connector builder call that endpoint on the YAML manifest in the editor before converting it to UI.
 
## How
- Add useResolveManifest to the ConnectorBuilderApiService to make the new endpoint usable
- Change `manifestToBuilderForm.ts` to a custom hook `useManifestToBuilderForm.ts`, since it needs to use the `useResolveManifest()` hook now
- Modify `convertToBuilderFormValues` to call the resolve endpoint at the beginning of the function, and use the result in the rest of the conversion
- The above change resulted in that method becoming `async`, so update its usages and the tests to be async as well


flash1293:
This is super nice! Trying this with real manifest I found a few generic issues in our conversion code: https://docs.google.com/document/d/1694N09qm839yEtsSrGfC1ptDCickVm9cQWsSYBSiT-Q/edit?usp=sharing

I don't think we need to solve all of them and most likely on a separate PR, but in order to make the "upload a yaml file" CTA useful I think we should solve the most critical ones.

lmossman:
@flash1293 thanks for compiling this! I agree we should solve these, though I think we can merge this PR in in its current state and split the doc into tickets as you suggested 👍

flash1293:
I think something went wrong with the auto-formatting

lmossman:
Woah, yep you're right. I'll work on fixing

lmossman:
@flash1293 this should be fixed now

flash1293 (approved):
LGTM

PR #21899, opened 2023-01-26
fix(platform): use updated configuration after check during syncs

### LINKED ISSUE #20912: Refresh connector configuration between attempts / activities

pedroslopez:
## What

fixes https://github.com/airbytehq/airbyte/issues/20912

This PR builds on the changes introduced in https://github.com/airbytehq/airbyte/pull/21629. While the aforementioned PR addressed using updated state/config values _between_ attempts, this PR solves the issue of updating configuration _within an attempt_. This can happen when, for example, an oauth token is refreshed during the `check` operation that can run prior to the SyncWorkflow.

## How

This implements the second part of the solution as laid out in the [tech spec](https://docs.google.com/document/d/1WSZQg2bhNyQD4cgy-Owzejfm8mXhZSYmaqWHlAOqtAs/edit#), changing _when_ we're generating the sync input and _how_ we are generating the check inputs. 

**Before** this change, the SyncWorkflow input was being generated at the start of the ConnectionManagerWorkflow, and was then converted into Check inputs if needed. The activities look something like this:

image

(these are snippets of diagrams, the full views can be found in the tech spec)

**After** this change, we separate the check input generation into its own activity method. Then, we can move the SyncWorkflow input generation to after the check activity runs. Because the previous PR updated the SyncWorkflow input generation activity to fetch latest configs, this will take into consideration any changes that happened during check.

This ends up looking something like this:

image

The change is versioned, so ConnectionManagerWorkflows that executed before this change will still follow the old path/behavior.

## 🚨 User Impact 🚨
No breaking change. Configs updated during check will now correctly be retrieved for the sync.


airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionManagerWorkflowImpl.java
```diff
@@ -208,18 +213,28 @@ private CancellationScope generateSyncWorkflowRunnable(final ConnectionUpdaterIn
...
-        final SyncCheckConnectionFailure syncCheckConnectionFailure = checkConnections(jobInputs);
+        final SyncCheckConnectionFailure syncCheckConnectionFailure = checkConnections(getJobRunConfig(), jobInputs);
         if (syncCheckConnectionFailure.isFailed()) {
           final StandardSyncOutput checkFailureOutput = syncCheckConnectionFailure.buildFailureOutput();
           workflowState.setFailed(getFailStatus(checkFailureOutput));
           reportFailure(connectionUpdaterInput, checkFailureOutput, FailureCause.CONNECTION);
         } else {
+          if (generateCheckInputVersion >= GENERATE_CHECK_INPUT_CURRENT_VERSION) {
+            jobInputs = getJobInput();
+          }
```
pedroslopez, line 236: This is where we are moving the activity that generates the SyncWorkflow inputs to after the potential checks. Based on the workflow version, we keep the old position of the activity.

We do this so that any config updates that happen during check are freshly retrieved before running the sync.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionManagerWorkflowImpl.java
```diff
@@ -84,12 +85,16 @@
...
 import java.util.function.Function;
+import javax.annotation.Nullable;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 @SuppressWarnings("PMD.AvoidDuplicateLiterals")
 public class ConnectionManagerWorkflowImpl implements ConnectionManagerWorkflow {
 
+  private static final String GENERATE_CHECK_INPUT_TAG = "generate_check_input";
+  private static final int GENERATE_CHECK_INPUT_CURRENT_VERSION = 1;
```
evantahler, line 96: ... time to version I guess!

evantahler: If you didn't version, I think that running syncs would fail, but perhaps only before the sync stage, not during or afterwords?  If that's the case, and they retry, that's not so bad?

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivity.java
```diff
@@ -36,6 +37,18 @@ class SyncInputWithAttemptNumber {
 
   }
 
+  @Data
+  @NoArgsConstructor
+  @AllArgsConstructor
```
evantahler, line 42: [`@AllArgsConstructor`](https://projectlombok.org/api/lombok/AllArgsConstructor) - today I learned

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityImpl.java
```diff
@@ -100,58 +105,173 @@ private void saveAttemptSyncConfig(final long jobId, final int attemptNumber, fi
...
+        .withAllowedHosts(configReplacer.getAllowedHosts(sourceDefinition.getAllowedHosts(), sourceConfiguration));
+
+    if (!ConfigType.RESET_CONNECTION.equals(configType)) {
+      sourceLauncherConfig.setAllowedHosts(configReplacer.getAllowedHosts(sourceDefinition.getAllowedHosts(), sourceConfiguration));
+    }
+
+    return sourceLauncherConfig;
+  }
+
+  private IntegrationLauncherConfig getDestinationIntegrationLauncherConfig(final long jobId,
```
pedroslopez, line 132: most of the changes in this file are pulling out some common logic used for both SyncWorkflow and Check input generation.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionManagerWorkflowImpl.java
```diff
@@ -336,28 +351,64 @@ private ConnectorJobOutput getCheckResponse(final CheckConnectionInput checkInpu
...
         new JobCheckFailureInput(Long.parseLong(jobRunConfig.getJobId()), jobRunConfig.getAttemptId().intValue(), connectionId);
     final boolean isLastJobOrAttemptFailure =
         runMandatoryActivityWithOutput(jobCreationAndStatusUpdateActivity::isLastJobOrAttemptFailure, jobStateInput);
-    if (isResetJob(sourceLauncherConfig) || checkFailure.isFailed() || !isLastJobOrAttemptFailure) {
+
+    if (!isLastJobOrAttemptFailure) {
+      log.info("SOURCE CHECK: Skipped, last attempt was not a failure");
+      log.info("DESTINATION CHECK: Skipped, last attempt was not a failure");
+      return checkFailure;
+    }
```
evantahler, line 391: I think I'm confused by this conditional - if we are skipping CHECKS because the previous job/attempt didn't fail, why would we return a failure?  Or is checkFailure initialized with `isFailed(false)`

pedroslopez: This was already the case before, I just moved this higher up to be clearer / more intentional about it. 

Yes, a `SyncCheckConnectionFailure` is always returned by this method. When an actual failure happens, `checkFailure.setFailureOrigin` and `checkFailure.setFailureOutput` are called. The `SyncCheckConnectionFailure` class has a method `.isFailed()` that actually returns whether it failed or not if these are set.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionManagerWorkflowImpl.java
pedroslopez: Yeah, that sounds right. Avoiding the failure wasn't too bad here with the version though so seemed worth it to keep history compatibility. FWIW there's also a test that verifies history can be replayed that not versioning would break 😄

airbyte-workers/src/main/java/io/airbyte/workers/temporal/scheduling/ConnectionManagerWorkflowImpl.java
pedroslopez: Ok, this was definitely confusing :D 

Renamed SyncCheckConnectionFailure to SyncCheckConnectionResult to be more clear that it doesn't _always_ represent a failure.

mfsiega-airbyte:
High-level comment - I think it would be worthwhile to break this down into smaller PRs, to (1) reduce the likelihood that something goes wrong; (2) minimise the impact if something _does_ go wrong.

1. The db migration should be on its own, for rollback safety.
2. I haven't taken a super close look but I'd imagine some of the workflow/persistence/API changes can be broken into separate changes?

pedroslopez:
@mfsiega-airbyte whoops, 100% - I did separate this into a smaller set of changes but when I merged the other piece this one now includes everything. Sorting that out now.

pedroslopez:
@mfsiega-airbyte OK - this should be more manageable now

pedroslopez:
superseded by https://github.com/airbytehq/airbyte-platform/pull/91

PR #21902, opened 2023-01-26
🪟 🐛 Fixes BulkEditServiceProvider to use filtered streams instead of all streams

### LINKED ISSUE #21551: All streams are selected for bulk edit after they are filtered by keyword
## Actual
All streams are selected for bulk edit even when they are filtered down.

## Expected
Only the filtered streams will be selected for bulk edit.

## Steps to Reproduce
1. Navigate to a connection replication settings
2. In the search box ("Search stream name") type the name of a stream to narrow down the list
3. Click on the checkbox below

YatsukBogdan1:
## What
Closes #21551 

## How
The solution is pretty straightforward, I just used filtered streams to use inside `<BulkEditServiceProvider />`
https://www.loom.com/share/b02731e1759948e1b299a6ed9836ea3c

But there is a tricky moment here. What should we do in case we checked all streams with some keywords and got for example 50 streams selected, but after that, we typed more symbols to search string and got only 10 results? In this case, those 50 streams will be still selected inside `useBulkEditService` while only 10 will be visible to the user
https://www.loom.com/share/466809e75bb74157915b91a6494560b0

As you can see right now `selectedBatchNodeIds` and `selectedBatchNodes` are not the same in the current implementation. So what should we do with this?
Screenshot 2023-01-26 at 09 50 03
```
const allChecked = selectedBatchNodes.size === nodes.length;
...
const ctx: BulkEditServiceContext = {
    ...
    allChecked,
    selectedBatchNodeIds: Array.from(selectedBatchNodes).filter((node): node is string => node !== undefined),
    selectedBatchNodes: nodes.filter((n) => selectedBatchNodes.has(n.id)),
    ...
};
```


edmundito:
Found a bug while testing:

1. Filter the streams and select "all"
2. Bulk edit and hit apply
3. Unselect the streams

You will notice that there is only the streams that were filtered, but the others are gone!

dizel852:
> What should we do in case we checked all streams with some keywords and got for example 50 streams selected, but after that, we typed more symbols to search string and got only 10 results? In this case, those 50 streams will be still selected inside useBulkEditService while only 10 will be visible to the user

Hmm... Interesting use-case. Probably we need to unselect all selected streams when the user changes filter settings. 
@edmundito WDYT?

dizel852:
> Found a bug while testing:
> 
> 1. Filter the streams and select "all"
> 2. Bulk edit and hit apply
> 3. Unselect the streams
> 
> You will notice that there is only the streams that were filtered, but the others are gone!

Confirm.

dizel852 (changes requested):
Need to fix the issue

edmundito (commented):
Tested locally and works as expected. Tested:

- Editing streams before and after bulk edit
- Cancel and Save
- Editing filtered streams with bulk edit

edmundito (approved):

edmundito:
I just read the question about changing the search during bulk editing. I think the best option is to disable the search input while bulk edit is enabled.

edmundito (approved):
Tested with disabled search, LGTM

dizel852 (approved):
LGTM

PR #21911, opened 2023-01-26
🪟🐛 Connector form: Make control reflect actual value of form field

### LINKED ISSUE #21910: Connector form: Do not pre-select value in select for undefined
When a prop has a default in the spec schema but it's not defined in the actual config (which can happen if the spec changes), it is rendering the default value as selected even though this is not reflected in the actual form state:
Screenshot 2023-01-26 at 13 21 15

In this screenshot, `additionalProp` is defined as:
```
{
  title: "Additional prop",
  type: "string",
  enum: ["one", "two", "three"],
  default: "two",
}
```

As visible in the debug output, it is not actually set in the form values, but the form makes it look like "two" is selected because that's the default.

Instead it should render the select without a selecting to force the user making a choice.

flash1293:
Fixes https://github.com/airbytehq/airbyte/issues/21910

The `Control` component would automatically fill in the default value if no value is set for selects, inputs and text areas. However this can hide the actual state of the form that will be persisted in the database - for unchanging specs this doesn't matter because `setDefaultValues` in

https://github.com/airbytehq/airbyte/blob/6c2ee304aa549b22a7ec7148764927eed1542040/airbyte-webapp/src/views/Connector/ConnectorForm/useBuildForm.tsx#L25

makes sure to set the default values in the form state itself, but when the spec is changing an an actor is edited, this can get out of sync.

This PR makes sure the actual value from the form values is reflected in the UI as well.

For an additional prop in the spec like this:
```
{
  title: "Additional prop",
  type: "string",
  enum: ["one", "two", "three"],
  default: "two",
}
```

Before:
Screenshot 2023-01-26 at 13 30 23

After:
Screenshot 2023-01-26 at 13 30 04


lmossman (approved):
For all of the slack discussion, multiple github issues, and great PR description, it was hilarious to me that the actual solution was literally deleting half of a line of code 😆 

Change LGTM, tested locally and it works as expected!

flash1293:
> For all of the slack discussion, multiple github issues, and great PR description, it was hilarious to me that the actual solution was literally deleting half of a line of code 😆

My next PR will be 3000 LOC and a single sentence of description to make up for it 😉

PR #21914, opened 2023-01-26
🪟 🎉 Remove advanced mode

timroes:
## What

This PR removes the Advanced Mode we had in the UI in favor of a simple collapsible in the only place it was used at the moment.

Some discussion and input from Charles made me think we should not keep this separate mode around instead of rather trying to "in place" hide setting we would hidden behind this flag in some collapsible or similar mechanism.

screenshot-20230126-141122


joeykmh (commented):
LGTM - tested locally and works as expected. I am seeing eslint warnings even after merging in `master` just now... any idea why?

Edit: nvm, had to reinstall eslint and restart the server - thought I had already done that 😅

joeykmh (approved):

PR #21922, opened 2023-01-26
Msiega/log unexpected fields

mfsiega-airbyte:
## What
Log when we see fields that aren't in the catalog.

## How
- Make a map of stream -> fields in the catalog.
- For each record, check if any of the fields are missing from the catalog; save these.
- At the end of the sync, log any unexpected fields.


airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -620,7 +636,21 @@ private static void validateSchema(final RecordSchemaValidator recordSchemaValid
...
+    if (data.isObject()) {
+      Iterator<String> fieldNamesInRecord = data.fieldNames();
+      while (fieldNamesInRecord.hasNext()) {
+        final String fieldName = fieldNamesInRecord.next();
+        if (!fieldsInCatalog.contains(fieldName)) {
+          unexpectedFieldNames.add(fieldName);
+        }
+      }
+    } else {
+      throw new RuntimeException(String.format("Unexpected data in record: %s", data));
```
benmoriceau, line 653: is there a posibility to have a valid non object here?

malikdiarra: I mean it should not happen, but if that happen something is going to break in the code in the `if` block, better to test and throw a more specific exception.
We could use a more specific error message though "Expecting object in record data, got '%s' instead" or something like this.

malikdiarra (approved):
This looks good to me. How easy is it to test this block of code before we release it?

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
mfsiega-airbyte: (@malikdiarra you had a comment on the same block)

According to the [Airbyte Protocol](https://docs.airbyte.com/understanding-airbyte/airbyte-protocol/#airbyterecordmessage) the `data` property of a record message must be a JSON of `object` type, so in theory no.

However, on second thought I don't think we should be throwing here in any case - that will basically crash the whole sync, and that's why we have the validation step (which doesn't throw, but rather builds a list of errors).

benmoriceau (approved):

PR #21930, opened 2023-01-26
Low-Code CDK: make DatetimeStreamSlicer.step InterpolatedString

grubberr:

## What
- make DatetimeStreamSlicer.step as InterpolatedString
- we interpolate early on `__post__init`, it allows to interpolate only `{{ config }}`, but one the other hand we fail fast if provided value is not correct


airbyte-cdk/python/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.25.1
+Low-Code CDK: make DatetimeStreamSlicer.step as InterpolatedString
```
maxi297, line 4: As part of the new release process, the version is updated in the new GitHub workflow. You can check https://airbytehq-team.slack.com/archives/C02UF50V9HA/p1674655636577679?thread_ts=1674654710.391809&cid=C02UF50V9HA for more information. 

An example of the issue is this: you have [another PR](https://github.com/airbytehq/airbyte/pull/21956) at the same time that has the same release version. This will clash i.e. one of the release will fail. To avoid this, we should always release on master. This allows to have one source of truth for the version and not multiple (one per branch)

grubberr: fixed

maxi297 (commented):
I don't see any issue with this MR. However, I'm curious in what cases this could be useful. Can you provide an example?

grubberr:
Hello,
I am working on rewriting source-Twilio from old to low-code CDK
Here is an example:
https://github.com/airbytehq/airbyte/pull/21446/files#diff-5b7ed3cb893f1b04c3c4ce89ac5b7abb7de4e9d12d9663a81f612c3f75bfdebdR49

maxi297:
Thanks a lot for the example @grubberr ! The more I know, the better I feel haha

maxi297 (approved):
Just adding a comment to approve. Good job!

PR #21931, opened 2023-01-26
🪟 🔧 Prevent @use renaming in SASS

timroes:
## What

Prevents renaming of `@use` statements in SASS as discussed in our chapter sync.

Also removes all the currently existing renames in SASS files.


joeykmh (approved):
Tested locally, looks good! 🥇 

One thing I noticed while testing, which is rather related to vite (I think) than stylelint: after fixing a stylelint error,  it seems like you have to restart vite to get stylelint to run again. tsc and eslint seem to re-check upon saving, but stylelint does not.

timroes:
@josephkmh Thanks for pointing this out. I'll have a look at that next week.

PR #21936, opened 2023-01-26
Use discoverActivityOptions for refresh schema activity

alovew:
Use DiscoveryActivityOptions for Refresh Schema Activity to increase scheduleToCloseTimeout to 2 hours

### DIFF
--- a/airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
+++ b/airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
@@ -64,8 +64,10 @@ public class SyncWorkflowImpl implements SyncWorkflow {
   private NormalizationSummaryCheckActivity normalizationSummaryCheckActivity;
   @TemporalActivityStub(activityOptionsBeanName = "shortActivityOptions")
   private WebhookOperationActivity webhookOperationActivity;
-  @TemporalActivityStub(activityOptionsBeanName = "discoveryActivityOptions")
+  @TemporalActivityStub(activityOptionsBeanName = "shortActivityOptions")
   private RefreshSchemaActivity refreshSchemaActivity;
+  @TemporalActivityStub(activityOptionsBeanName = "discoveryActivityOptions")
+  private RefreshSchemaActivity refreshSchemaActivityNew;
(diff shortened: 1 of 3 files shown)


benmoriceau (approved):

malikdiarra (approved):

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
```diff
@@ -88,7 +90,9 @@ public StandardSyncOutput run(final JobRunConfig jobRunConfig,
         Workflow.getVersion(AUTO_DETECT_SCHEMA_TAG, Workflow.DEFAULT_VERSION,
             AUTO_DETECT_SCHEMA_VERSION);
 
-    if (autoDetectSchemaVersion >= AUTO_DETECT_SCHEMA_VERSION) {
+    final int discoveryOptionsVersion = Workflow.getVersion(DISCOVERY_OPTIONS_TAG, Workflow.DEFAULT_VERSION, DISCOVER_OPTIONS_VERSION);
```
alovew, line 93: @benmoriceau I added a new version here, but still seeing test failures

benmoriceau: I think that we should have 2 `private RefreshSchemaActivity refreshSchemaActivity;` with the different options. Then you'll need to use the new version to call one or the other activity.

alovew: Just tried this, but I'm still seeing that test failure

alovew: just pushed up my changes - is that what you meant?

benmoriceau: The issue is in the stack trace of the failing test, The bean is not properly defined, I'm trying to make a fix.

benmoriceau: I made a fix for the workflow replayer, another test need to be updated.

PR #21949, opened 2023-01-26
pass connection_id to replication activity apm tags

cgardens:
## What
* I am trying to create a datadog dashboard to measure the throughput of our canary workspace.

## How
* Include connection id in the replication activity APM tags, so I can use it as a filter in datadog.
* Added `connection_id` as one of the keys that gets passed in via `StandardSyncInput`.

### DIFF
--- a/airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/TemporalClient.java
+++ b/airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/TemporalClient.java
@@ -400,7 +400,9 @@ public class TemporalClient {
         .withState(config.getState())
         .withResourceRequirements(config.getResourceRequirements())
         .withSourceResourceRequirements(config.getSourceResourceRequirements())
-        .withDestinationResourceRequirements(config.getDestinationResourceRequirements());
+        .withDestinationResourceRequirements(config.getDestinationResourceRequirements())
+        .withConnectionId(connectionId)
+        .withWorkspaceId(config.getWorkspaceId());
 
     return execute(jobRunConfig,
         () -> getWorkflowStub(SyncWorkflow.class, TemporalJobType.SYNC).run(
--- a/airbyte-config/config-models/src/main/resources/types/StandardSyncInput.yaml
+++ b/airbyte-config/config-models/src/main/resources/types/StandardSyncInput.yaml
@@ -70,3 +70,7 @@ properties:
     description: The id of the workspace associated with this sync
     type: string
     format: uuid
+  connectionId:
+    description: The id of the connection associated with this sync
+    type: string
+    format: uuid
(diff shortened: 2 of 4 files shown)


airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/TemporalClient.java
```diff
@@ -400,7 +400,9 @@ public TemporalResponse<StandardSyncOutput> submitSync(final long jobId, final i
         .withState(config.getState())
         .withResourceRequirements(config.getResourceRequirements())
         .withSourceResourceRequirements(config.getSourceResourceRequirements())
-        .withDestinationResourceRequirements(config.getDestinationResourceRequirements());
+        .withDestinationResourceRequirements(config.getDestinationResourceRequirements())
+        .withConnectionId(connectionId)
+        .withWorkspaceId(config.getWorkspaceId());
```
cgardens, line 405: I noticed workspace_id wasn't passed in here. But is passed in other places where we build this struct. @mfsiega-airbyte was that intentional?

airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/TemporalClient.java
```diff
@@ -400,7 +400,9 @@ public TemporalResponse<StandardSyncOutput> submitSync(final long jobId, final i
         .withState(config.getState())
```
cgardens, line 400: I noticed source_id and destination_id wasn't passed in here. But is passed in other places where we build this struct. @pedroslopez was that intentional?

pedroslopez: Not intentional. I didn't know submitSync was a thing when working on the original PR that introduced them. 

I've actually been recently trying to understand where this submitSync method comes into play... when does this get executed? From what I've seen we're always using the inputs as generated by the GenerateInputActivity in actual sync jobs.

jdpgrailsdev (approved):
:shipit:

airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/TemporalClient.java
mfsiega-airbyte: Indeed not intentional - but same question as Pedro.

airbyte-commons-temporal/src/main/java/io/airbyte/commons/temporal/TemporalClient.java
cgardens: tbh, i've lost track of it as well. @benmoriceau is this something you understand well?

PR #21950, opened 2023-01-26
Support config sub-values for `allowedHosts`

### LINKED ISSUE #21913: Consider DB proxies for AllowedHosts
How can these (optional) connections be represented?

evantahler:
Following https://github.com/airbytehq/airbyte/pull/21676 and closes https://github.com/airbytehq/airbyte/issues/21913, this PR adds support for allowedHosts that come from sub-keys in the connector's configuration.  

For example, configs for postgres sources which use SSH tunnels might look like:

```json
{
  "ssl": false,
  "host": "real-server-ip.com",
  "port": 5432,
  "schemas": [ "public" ],
  "database": "datawarehouse",
  "password": { "_secret": "..." },
  "ssl_mode": { "mode": "disable" },
  "username": "db-username",
  "tunnel_method": {
    "tunnel_host": "1.2.3.4",
    "tunnel_port": 22,
    "tunnel_user": "ssh-username",
    "tunnel_method": "SSH_PASSWORD_AUTH",
    "tunnel_user_password": { "_secret": "..." }
  },
}
```

Which means allowedHosts should support both `host` and `tunnel_method.tunnel_host`.

### DIFF
--- a/airbyte-config/init/src/main/resources/seed/source_definitions.yaml
+++ b/airbyte-config/init/src/main/resources/seed/source_definitions.yaml
@@ -1349,6 +1349,7 @@
   allowedHosts:
     hosts:
       - "${host}"
+      - "${tunnel_method.tunnel_host}"
 - name: Postmark App
   sourceDefinitionId: cde75ca1-1e28-4a0f-85bb-90c546de9f1f
   dockerRepository: airbyte/source-postmarkapp
--- a/airbyte-workers/src/main/java/io/airbyte/workers/utils/ConfigReplacer.java
+++ b/airbyte-workers/src/main/java/io/airbyte/workers/utils/ConfigReplacer.java
@@ -35,18 +35,34 @@ public class ConfigReplacer {
    * for a connector has a single depth.
    */
   public AllowedHosts getAllowedHosts(AllowedHosts allowedHosts, JsonNode config) throws IOException {
-    if (allowedHosts == null || allowedHosts.getHosts() == null) {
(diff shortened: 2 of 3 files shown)


airbyte-config/init/src/main/resources/seed/source_definitions.yaml
```diff
@@ -1349,6 +1349,7 @@
   allowedHosts:
     hosts:
       - "${host}"
+      - "${tunnel_method.tunnel_host}"
```
evantahler, line 1352: cc @bleonard / @prateekmukhedkar - this should be the complete set of needs for source-postgres!

git-phu (approved):
Ah ok, makes sense!

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
prateekmukhedkar: @evantahler I am working on adding `allowedHosts` to the remaining database sources([link](https://github.com/airbytehq/airbyte/issues/22372)). Reading through this PR, it appears that `${tunnel_method.tunnel_host}` is only required for those sources that support connection via SSH tunnels. (for example as of now we don't support connecting via ssh for source-redshift). And for all such sources I will only include `${host}` sub-value under `allowedHosts`. Does that sound correct?

evantahler: Yep!

PR #21951, opened 2023-01-27
Add SubscriptionUsage stream to Orb Source

pmossman:
## What
This is a WIP. Fair warning - this is my first real connector development so I'm definitely looking for feedback! I'm also not a python developer by any means so please give feedback on the actual python code here as well 😅 

The SubscriptionUsage API Endpoint is not currently available as a stream in the Orb Source. This PR is adding a new stream for this endpoint.

I highly recommend reading the docs page for this endpoint, as this endpoint works very differently compared to other Orb API endpoints: https://docs.withorb.com/docs/orb-docs/api-reference/operations/get-a-subscription-usage

As this is my first real connector development that I've done, I don't have a great gauge for how hairy and complex the logic here is. It feels complicated, but perhaps this is par for the course when it comes to building out incremental streams against endpoints that don't offer straight-forward pagination?

I still need to add unit tests and integration tests, but I want to put up a PR draft to start collecting feedback now that I have something that seems to sort of work when I run it in Airbyte OSS 😅 

Thanks for taking a look!

## How
There are a few key behaviors of this endpoint that inform my implementation approach for this stream:
- This endpoint is NOT paginated in a traditional sense - instead, usage quantities for a subscription are returned over a time window that is specified by the caller. 
     - If I want to incrementally fetch `subscription usage since yesterday`, for example, I need to pass a `timeframe_start` param with yesterday's date, and `timeframe_end` parameter with the current date. Furthermore, if I want to fetch usage broken down for each day in my time window, I need to specify `granularity: day`, otherwise I'll receive a single quantity representing usage over the entire time window.
     - I noticed that our Slack source behaves in a similar way, so I modeled my approach off of [this section of the Slack source](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-slack/source_slack/source.py#L150-L237)
     - I decided to hardcode `granularity: day` inside the stream because breaking down usage into single-day records makes reasoning around state much easier, and I think lends itself to an incremental stream nicely.
- It is very useful to break down subscription usage by a particular trait, like an Airbyte job_id for example. Luckily, the usage endpoint supports a `group_by` parameter which will return usage grouped by the event key you pass it.
    - A caveat here is that to use a `group_by` param, the caller must also specify a particular `billing_metric_id`. To work around this constraint, I implemented stream slicing to detect when a `group_by` is present, and yield a slice for every billing_metric_id present in the subscription's plan in this case. 

Also I made a small change to the subscriptions stream while I was working on this - I added the external_customer_id to subscription records because I know this will help our analytics team out, it isn't used or needed to implement the new subscription_usage stream.

## Recommended reading order
1. source.py
5. the rest


airbyte-integrations/connectors/source-orb/source_orb/source.py
```diff
@@ -177,6 +180,261 @@ def transform_record(self, subscription_record):
...
+        return None
+    elif isinstance(time, pendulum.DateTime):
+        return time
+    elif isinstance(time, str):
+        return pendulum.parse(time)
+    else:
+        raise TypeError(f"Cannot convert input of type {type(time)} to DateTime")
+
+
+def to_utc_isoformat(time) -> str:
```
girarda, line 195: nit: this function returns `Optional[str]`

airbyte-integrations/connectors/source-orb/source_orb/source.py
```diff
@@ -177,6 +180,261 @@ def transform_record(self, subscription_record):
...
+    """
+    one_day = pendulum.duration(days=1)
+    end_date = end_date or pendulum.now()
+
+    # Each stream_slice contains the beginning and ending timestamp for a 24 hour period
+    chunk_start_date = start_date
+    while chunk_start_date < end_date:
+        chunk_end_date = min(chunk_start_date + one_day, end_date)
+        yield pendulum.period(chunk_start_date, chunk_end_date)
+        chunk_start_date = chunk_end_date
```
girarda, line 219: the function needs to `yield from []`. Otherwise it'll return `None` if `chunk_start_date >= end_date`

pmossman:
test connector=source-orb

airbyte-integrations/connectors/source-orb/source_orb/source.py
```diff
@@ -177,6 +180,261 @@ def transform_record(self, subscription_record):
...
+        """
+        schema = super().get_json_schema()
+        if self.subscription_usage_grouping_key:
+            schema[self.subscription_usage_grouping_key] = {"type": "string"}
+
+        return schema
+
+    def stream_slices(self, stream_state: Mapping[str, Any] = None, **kwargs) -> Iterable[Optional[Mapping[str, Any]]]:
+        """
+        This stream is sliced per `subscription_id` and day, as well as `billable_metric_id`
```
girarda, line 383: the logic feels complex because the stream is sliced on 3 dimensions. That's a fun!

airbyte-integrations/connectors/source-orb/source_orb/source.py
```diff
@@ -177,6 +180,257 @@ def transform_record(self, subscription_record):
...
+    def transform_record(self, parent_record, sub_record, subscription_id):
+        self.logger.debug("incoming subscription usage record: %s", parent_record)
+
+        # Merge the parent record with the sub record
+        record = {**parent_record, **sub_record}
+
+        # Add the subscription ID to the record
+        record["subscription_id"] = subscription_id
+
+        # Un-nest billable_metric -> name,id into billable_metric_name and billable_metric_id
```
girarda, line 286: Why do you un-nest the billable metric in the source instead of letting the normalization handle it?

airbyte-integrations/connectors/source-orb/source_orb/source.py
```diff
@@ -177,6 +180,257 @@ def transform_record(self, subscription_record):
...
+        # If a grouping key is present, it should be included in the primary key
+        if self.subscription_usage_grouping_key:
+            key.append(self.subscription_usage_grouping_key)
+
+        return key
+
+    def parse_response(self, response: requests.Response, **kwargs) -> Iterable[Mapping]:
+        # use a regex capture group to extract the subscription ID from the request URL
+        # so that we can include it in the returned record
+        subscription_id = re.search("/subscriptions/(.*)/usage", response.request.url).group(1)
```
girarda, line 255: you should be able to get the `subscription_id` from the `stream_slice` if you replace the `**kwargs` argument with the [same signature](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/streams/http/http.py#L203-L210) as the `HttpStream` (the parent class)

(you could also get the `stream_slice` from `kwargs['stream_slice']` but **kwargs are tricky because they obfuscate the interface

airbyte-integrations/connectors/source-orb/source_orb/source.py
pmossman: I wanted to use billable_metric_id as part of the primary key, and felt like it would be clearer to un-nest everything at the record level so that the primary key path and record schema was easy to work with. Is there a good reason or best practice for keeping it nested that I should try to follow instead?

girarda: nope, I asked out of curiosity. We leave that decision to the connector developer. This is totally fine

airbyte-integrations/connectors/source-orb/source_orb/source.py
pmossman: Ahh perfect, I was looking for a way to access the stream_slice in parse_response but was confused after using the same signature as the IncrementalOrbStream class and didn't think to check the HttpStream class! This is much cleaner

pmossman:
^ The integration test failed because we don't have any subscription usage data in the test account that the tests call. I am following up with Orb to see if they can add some test data to the account. We have an API key for the account, but I don't see login credentials in lastpass, so I'm assuming Orb has access on their end and can help us add some fake data

girarda:
@YowanR do you know where we can get the credentials for the orb test account? can whoever has them upload them to lastpass?

pmossman:
@YowanR @girarda I got in touch with a dev at Orb who has account access, and he will send me credentials soon that I will add to LastPass

docs/integrations/sources/orb.md
```diff
@@ -52,6 +54,7 @@ an Orb Account and API Key.
 
 | Version | Date | Pull Request | Subject |
 | --- | --- | --- | --- |
+| 0.1.5 | 2023-02-02 | [21951](https://github.com/airbytehq/airbyte/pull/21951) | Add SubscriptionUsage stream
```
girarda, line 57: nit: this should be a minor update since we're adding a new stream

airbyte-integrations/connectors/source-orb/source_orb/source.py
```diff
@@ -177,6 +179,268 @@ def transform_record(self, subscription_record):
...
+        start_date: Optional[pendulum.DateTime] = None,
+        end_date: Optional[pendulum.DateTime] = None,
+        **kwargs,
+    ):
+
+        super().__init__(**kwargs)
+        self.subscription_usage_grouping_key = subscription_usage_grouping_key
+        self.plan_id = plan_id
+        # default to 30 days ago if start_date is unspecified
+        self.start_date = start_date if start_date else pendulum.now().subtract(days=30)
```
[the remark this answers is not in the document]
girarda: I would also recommend setting the default value in the spec instead of the code to make it visible to the users

girarda (changes requested):
the code looks good but I think the dynamic start date can lead to data loss

pmossman:
@girarda I made `start_date` a required field and set the new version to `1.0.0` to reflect the breaking change. Let me know if there's anything else I should address!

pmossman:
@girarda sorry for how noisy this PR has been, I finally figured out the GSM secret changes I needed to make to get the tests all passing now that start_date is required. Should be good for review now!

docs/integrations/sources/orb.md
```diff
@@ -13,20 +13,22 @@ This Source is capable of syncing the following core resources, each of which ha
...
-The Orb API does not allow querying objects based on an `updated_at` time. Therefore, this connector uses the `created_at` field to query for new data since the last sync.
+The Orb API does not allow querying objects based on an `updated_at` time. Therefore, this connector uses the `created_at` field (or the `timeframe_start` field in the Subscription Usage stream) to query for new data since the last sync.
 
 In order to capture data that has been updated after creation, please run a periodic Full Refresh.
 
 ### Features
 
 | Feature | Supported? |
-| :--- | :--- |
+| :--- | :--- |`
```
girarda, line 31: is this intentional?

girarda (approved):
:shipit:

PR #21957, opened 2023-01-27
Source Amplitude: handle null values and ampty strings in datetimefields

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1347

## How
If a datetime field contains a null value or empty string - do not try to convert it, just leave it as is


davydov-d:
I'm gonna publish this change without running tests because our previous test account has been deactivated in GSM (probably due to real/live data in it) and our new account produces an error during a test run.
I have run the tests locally using previous creds:
image

bazarnov (approved):

davydov-d:
/publish connector=connectors/source-amplitude run-tests=false

PR #21917, opened 2023-01-26
QA engine: add adoption metrics to the QA report

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/21721
We want to have adoption metrics per connector in the QA report:
- number_of_connections: 
- number_of_users
- sync_success_rate
- total_syncs_count
- failed_syncs_count
 - succeeded_syncs_count

We implemented the logic to fetch this data in a previous PR. 
This PR adds this data to the QA report dataframe **but does not persist it to GCS as we consider this data private**.

PR #21961, opened 2023-01-27
qa-engine: define cloud eligibility

### LINKED ISSUE #21919: Cloud availablity updater: assess eligibility of a connector to cloud.
Filter the QA report to spot which connectors are meeting the following requirements:
* `is_on_cloud: false`
* `latest_build_is_successful: true`
* `is_appropriate_for_cloud_use: true`
* `documentation_is_available: true`

alafanechere:
## What
Closes #21919

## How
1. Add a `is_eligible_for_cloud` column to the QA report dataframe
2. Add a `report_generation_datetime` column to the QA report dataframe: might be useful in the future when we'll archive the QA reports
3. Compute eligibility with  the `is_eligible_for_cloud`  function that checks the truthiness of the following conditions:
* Is on cloud: false
* Latest build is successful: true
* Documentation is available: true
* Is appropriate for cloud use: true
4. Expose a `get_connectors_eligible_for_cloud` function which returns a generator of eligible connectors for cloud.


tools/ci_connector_ops/setup.py
```diff
@@ -23,7 +23,7 @@
 
 
 setup(
-    version="0.1.6",
+    version="0.1.8",
```
alafanechere, line 26: Anticipating the bump of https://github.com/airbytehq/airbyte/pull/21917

tools/ci_connector_ops/ci_connector_ops/qa_engine/validations.py
```diff
@@ -20,6 +29,14 @@ def url_is_reachable(url: str) -> bool:
 def is_appropriate_for_cloud_use(definition_id: str) -> bool:
     return definition_id not in INAPPROPRIATE_FOR_CLOUD_USE_CONNECTORS
 
+def is_eligible_for_cloud(connector_qa_data: pd.Series) -> bool:
+  if connector_qa_data["is_on_cloud"]:
+    return False
```
erohmensing, line 34: Is there a case when something that wasn't on cloud before is now eligible for cloud because one of the previously passing checks now passes? Maybe I don't have the correct understanding of `is_on_cloud`

erohmensing (approved):
🛳 it!

tools/ci_connector_ops/ci_connector_ops/qa_engine/validations.py
alafanechere: No this flag is not dynamic.

We compute it by joining OSS and Cloud catalog: if the connector definition + version is not in the cloud catalog `is_on_cloud = false`

erohmensing: How does a connector become `is_on_cloud` (and therefore `is_eligible_for_cloud`) for the first time? By being manually added to the cloud catalog?

alafanechere: if `is_eligible_for_cloud is True` then this connector will be returned by the `get_connectors_eligible_for_cloud` function.
A future logic will create add this connector to cloud by:
- adding its definition id to the definitions mask on airbyte cloud
- open a PR with this change.
I started this work in https://github.com/airbytehq/airbyte/pull/21976

erohmensing: As clarified - this is used solely to determine whether an OSS-only connector is eligible to be added to cloud. It doesn't apply to connectors that are already in cloud and determining whether they're still eligible (we don't currently implement any removal from cloud based on broken builds or anything like that). I was missing that `is_on_cloud` has to be `False`, and was thinking the opposite! 

LGTM 👍🏻

PR #21962, opened 2023-01-27
🎉 🐛 Source Airtable: cast native Airtable Types to JSONSchema Types

bazarnov:
## What
Resolving: https://github.com/airbytehq/alpha-beta-issues/issues/299

## How
- refactored the `schema_helpers.py` to provide the functionality to cast data types automatically

## 🚨 User Impact 🚨

@erica-airbyte 
This is the BREAKING CHANGE. 

But what should customer do? :
After the update, the customers should `Refresh Schema` and reset the data to pick up the correct schema types and avoid issues on the destination side. Unfortunately. 

Reason for the breaking change:
The reason for the breaking change is this one: https://github.com/airbytehq/alpha-beta-issues/issues/299

## Pre-merge Checklist

- [x] Code reviews completed
- [x] Documentation updated
    - [x] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>


airbyte-integrations/connectors/source-airtable/source_airtable/schema_helpers.py
```diff
@@ -3,35 +3,115 @@
...
+                complex_type = deepcopy(COMPLEX_AIRTABLE_TYPES.get(original_type))
+                # process arrays with values
+                if complex_type == SchemaTypes.array_with_any:
+                    complex_type["items"] = deepcopy(SIMPLE_AIRTABLE_TYPES.get(field_type))
+                properties.update(**{name: complex_type})
+            elif original_type in SIMPLE_AIRTABLE_TYPES.keys():
+                properties.update(**{name: deepcopy(SIMPLE_AIRTABLE_TYPES.get(field_type))})
+            else:
+                # for all unknown types - cast to string
+                properties.update(**{name: SchemaTypes.string})
```
erohmensing, line 106: Nice! Maybe add a comment like `Airtable may add more field types in the future and don't consider it a breaking change` so it's clear to whoever sees this next that unknown types are not under our control, we've accounted for everything we could

airbyte-integrations/connectors/source-airtable/source_airtable/schema_helpers.py
```diff
@@ -3,35 +3,115 @@
...
+            exec_type: str = options.get("result", {}).get("type") if options else None
+            field_type: str = exec_type if exec_type else original_type
+
+            # choose the JsonSchema Type for known Airtable Types
+            if original_type in COMPLEX_AIRTABLE_TYPES.keys():
+                complex_type = deepcopy(COMPLEX_AIRTABLE_TYPES.get(original_type))
+                # process arrays with values
+                if complex_type == SchemaTypes.array_with_any:
+                    complex_type["items"] = deepcopy(SIMPLE_AIRTABLE_TYPES.get(field_type))
+                properties.update(**{name: complex_type})
```
erohmensing, line 101: I'm a little muddy on how the `result` is working - but I am wondering if we can do anything to enforce the type of other `options` returned by some complex types. I'm looking for example at [formula](https://airtable.com/developers/web/api/field-model#formula) and handling the boolean in the options. Maybe its all going to the same column in a nested format so it doesn't matter anyway?

airbyte-integrations/connectors/source-airtable/integration_tests/expected_records.jsonl
```diff
@@ -1,42 +1,42 @@
...
...esque gravida euismod massa, eget egestas sem facilisis gravida. Suspendisse quis mauris eget velit faucibus aliquam. Mauris porttitor urna lorem, eget tincidunt ex faucibus et. Nam viverra nibh quis turpis scelerisque, et condimentum nibh vulputate. Vivamus eu dolor posuere, finibus ligula vel, finibus erat. Sed hendrerit luctus erat, eu finibus ante blandit ut. Sed id mi ullamcorper, cursus urna nec, molestie lorem. In vulputate tempor nulla in laoreet. Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Sed a risus risus. Praesent neque lorem, fringilla vitae rhoncus a, semper eu augue.", "single_select_(colors)": "1", "single_select_(no_colors)": "1", "url": "airbyte.io", "user_(multiple)": ["{'id': 'usrr0cUCbVccLxU7x', 'email': 'integration-test@airbyte.io', 'name': 'Augustin Lafanechere'}"], "created_(with_time)": "2022-12-01T20:27:46.000Z", "created_(w/o_time)": "2022-12-01", "created_by": "{'id': 'usrr0cUCbVccLxU7x', 'email': 'integration-test@airbyte.io', 'name': 'Augustin Lafanechere'}", "last_modified_by": "{'id': 'usrr0cUCbVccLxU7x', 'email': 'integration-test@airbyte.io', 'name': 'Augustin Lafanechere'}"}, "emitted_at": 1674743463539}
```
erohmensing, line 12: The diff is turning out strange, but changes to the records for which the entire record is diffed instead of just some values look to have changed in a similar manner 👍🏻

docs/integrations/sources/airtable.md
```diff
@@ -54,10 +54,46 @@ The `Enterprise` level accounts are not supported yet.
 
 ## Data type map
```
erohmensing, line 55: Worth mentioning somewhere that all the fields are nullable?

airbyte-integrations/connectors/source-airtable/acceptance-test-config.yml
```diff
@@ -19,14 +19,21 @@ acceptance_tests:
   discovery:
     tests:
       - config_path: "secrets/config.json"
+        # the static `string` schema is replaced with casted to JSONSchema
+        backward_compatibility_tests_config:
+          disable_for_version: "1.0.2"
```
erohmensing, line 24: This is more of a clarity question for @alafanechere - `Disable the backward compatibility test for a specific version (expects a version following semantic versioning).` I'd expect that to mean that for version V we don't check to see if its backward compatible with version V-1 where V-1 means the version released before V. In that case it'd be setting it to `2.0.0` here. We actually set it for the previous version?

erohmensing (commented):
Looks legit to me - just a few questions in the review. TCS might want to reach out to users to let them know of the breaking change, in case they have pipelines consuming the data in string format - have you all aligned on any follow-up necessary?

airbyte-integrations/connectors/source-airtable/source_airtable/schema_helpers.py
bazarnov: Done.

airbyte-integrations/connectors/source-airtable/source_airtable/schema_helpers.py
bazarnov: The `COMPLEX` type is something that I've `overthought` to give the explanation of  the next structure due  to Airtable Types:
- COMPLEX could be one of the:
or `[ string. string, ... ] `
or `[ [list], [list] ] `
or `[ [string, string, etc], [string, string], etc ]`

It's basically the Array with some other `SIMPLE` type inside. It circles back and forth to get the complex type first,  then looks for the simple type that the complex type contains inside. All the information is stored inside the response from the schema API request, in the `options` > `result` object that points to the exact type that column handles. The path is always the same for all  complex  types, if something is wrong it will not break but will be cast as a `string`

bazarnov: Also, here we have a deal with `array_with_any` type, so we handle them specifically here:
```python
# process arrays with values
 if complex_type == SchemaTypes.array_with_any:
    complex_type["items"] = deepcopy(SIMPLE_AIRTABLE_TYPES.get(field_type))
```

Once there are some `array_with_any` and  we have the  `object` inside, instead  of `array`, it will be handled  like:
```python
# process arrays with values
   if complex_type == SchemaTypes.array_with_any_object:
      complex_type["properties"] = deepcopy(SIMPLE_AIRTABLE_TYPES.get(field_type)) // assuming we are dealing with ready to go field names 
 ```

File: airbyte-integrations/connectors/source-airtable/integration_tests/expected_records.jsonl (hunk 3 above)
Comment (bazarnov) [reply airbyte-integrations/connectors/source-airtable/integration_tests/expected_records.jsonl:12]: I guess it's a matter of formating the `.jsonl`, and the long record inside is just not formated as all above.

File: docs/integrations/sources/airtable.md (hunk 4 above)
Comment (bazarnov) [reply docs/integrations/sources/airtable.md:55]: Done.

File: airbyte-integrations/connectors/source-airtable/acceptance-test-config.yml (hunk 5 above)
Comment (bazarnov) [reply airbyte-integrations/connectors/source-airtable/acceptance-test-config.yml:24]: Yes, the `1.0.2` is the latest for now, it's published to our Dockerhub and contains OAuth implementation,  but it's not currently released to Cloud, we've pinned the version to be `1.0.1` to allow the server components to update before we release the connector with OAuth to Cloud. However, the `2.0.0` will include all the previous changes + this big `casting` functionality, which could be treated as a `Breaking Change` and `Major` release. All subsequent versioning will be smooth on 2.0.x and semantic versioning is considered. But thanks for the question!

**Comment by bazarnov** [comment]
> Looks legit to me - just a few questions in the review. TCS might want to reach out to users to let them know of the breaking change, in case they have pipelines consuming the data in string format - have you all aligned on any follow-up necessary?

We could safely push this one into Master because:
- we hold the Cloud version on `1.0.1` until `1.0.2` server-side components are updated for the OAuth part and merged into Cloud master and deployed.
- for OSS folks - we can safely release the `2.0.0` and give them a shot to try the new functionality.
- once the Cloud part is ready, we just unpin the version for `source-airtable` and have both `OAuth` + `dynamic schema discovery` + `cast to JsonSchema types` at the same time, and it will look like a single Major update for the Users to minimize the customer's effort.

File: airbyte-integrations/connectors/source-airtable/acceptance-test-config.yml (hunk 5 above)
Comment (erohmensing) [reply airbyte-integrations/connectors/source-airtable/acceptance-test-config.yml:24]: Cool, yup just wanted to check - code references the previous connector version 
`if previous_connector_version == inputs.backward_compatibility_tests_config.disable_for_version:` so this does make sense - just not how I interpreted the parameter.

File: docs/integrations/sources/airtable.md [hunk 6]
```diff
@@ -67,6 +103,7 @@ See information about rate limits [here](https://airtable.com/developers/web/api
 
 | Version | Date       | Pull Request                                             | Subject                                                |
 |:--------|:-----------|:---------------------------------------------------------|:-------------------------------------------------------|
+| 2.0.0   | 2023-01-27 | [21962](https://github.com/airbytehq/airbyte/pull/21962) | Added casting of native Airtable data types to JsonSchema types
```
erohmensing, line 106: If OSS users need to also reset the schema and refresh their data, I think we need to indicate that here. I'm not sure if we have any documentation already about implications to end users of semver major version changes

erohmensing (approved):
> once the Cloud part is ready, we just unpin the version for source-airtable and have both OAuth + dynamic schema discovery + cast to JsonSchema types at the same time, and it will look like a single Major update for the Users to minimize the customer's effort.

Thank you for the thorough overview! This sounds like a good idea to bunch a lot of changes together

> IMPORTANT: we would need to reach out to the existing Customers and announce that a big update is coming and they should be ready to Reset the data and Refresh the Schema.

👍🏻 please make sure TCS is aware of that on the original issue

docs/integrations/sources/airtable.md
bazarnov: Not the best idea to put these descriptions inside of the changelog, but generally speaking, I agree. We should have a dedicated place or notifications that could guide the users through obstacles with upgrades/updates. 
I think we have enough information for the `set up guide` already, the rest - should be part of general documentation at least. There is nothing special about this one, the same rules as applied to the other connectors while upgrading and the changes are breaking.

sherifnada (approved):

docs/integrations/sources/airtable.md
erohmensing: Hm, that's a good point. I'm not sure what the solution is to make sure that OSS users that don't get proactive outreach are aware

bazarnov:
@erica-airbyte The server-side changes for Airtable OAuth are deployed to the Cloud according to the latest release and Cloud version. Are we good to `unpin` the `source-airtable` version to pick up the latest `2.0.0`, meaning we should reach out to the Customers before this goes live?

erica-airbyte:
@bazarnov Yes, we need to reach out to customer if this change will cause their syncs to start failing. Will this change have any downstream affect to their destination data?  (ie. stream name changing, field name changing, etc.)

bazarnov:
> @bazarnov Yes, we need to reach out to customer if this change will cause their syncs to start failing. Will this change have any downstream affect to their destination data?  (ie. stream name changing, field name changing, etc.) 

@erica-airbyte The changes touches only the data types, the names and the whole structure are the remains, for the customers who is already updated to 1.0.1 and more, if updating from 0.1.3 - the stream names are also changed.

erica-airbyte:
Ok, thanks @bazarnov we will send this outreach tomorrow to customers. 

Can we plan to have this update go out on Thursday?

erica-airbyte:
Comms have been sent. @davydov-d please deploy this tomorrow

davydov-d:
@erica-airbyte done

davydov-d:
@bazarnov as you were the one to work on this PR, are you aware of what could go wrong on the screenshots above?

erica-airbyte:
cc @bazarnov can you take a look at this real quick. Customer are not able to get around this.

TBernstein4:
On a test workspace I changed the Auth Method to OAuth and gave access to all workspaces. Is there a chance that this update is not compatible with Users that are using API keys?

bazarnov:
@erica-airbyte @pedroslopez I've just tested using `prod` - all works as expected. 

@TBernstein4 does the customer proceed with [these instructions ](https://airtable.com/developers/web/guides/personal-access-tokens#creating-a-token)to create a new `PAT` with all necessary scopes and permissions?
Also, could you please share the workspace link with this customer so we can examine the input config?

TBernstein4:
I will ask them about the PAT. I will mention you in the oncall issue that has the effected workspace links.

Issue #21926, opened 2023-01-26
Low-Code Beta CDK in Connector Builder UI


maxi297:
## Tell us about the problem you're trying to solve
We are working on bringing the Low-Code CDK to beta. Since the Connector Builder heavily depends on this lib and would be block from further CDK changes if we do nothing, it should be updated as part of the initiative. 

## Describe the solution you’d like
Here are the issues that have been worked on. Some are already part of the `low_code_cdk_to_beta` branch.

Low Code Issue | Low Code PR | Merged in low_code_cdk_to_beta? | Fixed in FE?
-- | -- | -- | --
https://github.com/airbytehq/airbyte/issues/19410 | https://github.com/airbytehq/airbyte/pull/21403 | ✅ | ✅
https://github.com/airbytehq/airbyte/issues/19408 | https://github.com/airbytehq/airbyte/pull/21632 | ✅ | no change needed
https://github.com/airbytehq/airbyte/issues/19409 | https://github.com/airbytehq/airbyte/pull/21434 | ✅ | no change needed
https://github.com/airbytehq/airbyte/issues/19698 | https://github.com/airbytehq/airbyte/pull/21832 | ✅ | no change needed
https://github.com/airbytehq/airbyte/issues/21866 | https://github.com/airbytehq/airbyte/pull/21878 | ✅ | no change needed
https://github.com/airbytehq/airbyte/issues/21559 | https://github.com/airbytehq/airbyte/pull/21823 | ✅ | ✅
https://github.com/airbytehq/airbyte/issues/21560 | https://github.com/airbytehq/airbyte/pull/21891 | ✅  
https://github.com/airbytehq/airbyte/issues/21561 | https://github.com/airbytehq/airbyte/pull/22294 | ✅ | no change needed
https://github.com/airbytehq/airbyte/issues/21576 | https://github.com/airbytehq/airbyte/pull/21990 | ✅ | ✅
https://github.com/airbytehq/airbyte/issues/21239 | https://github.com/airbytehq/airbyte/pull/22120 | ✅
https://github.com/airbytehq/airbyte/issues/19961 | https://github.com/airbytehq/airbyte/pull/22225 | ✅ 
https://github.com/airbytehq/airbyte/issues/20792 https://github.com/airbytehq/airbyte/issues/21095 | https://github.com/airbytehq/airbyte/pull/22398 | ✅ 

Basic outline of the work to be done:
* Backend
[...]

PR #21974, opened 2023-01-27
[ISSUE #21926] setup connector-builder-server to allow for frontend local tests

maxi297:
## What
Set up Connector Builder Server to enable frontend work needed to support Low-Code CDK to Beta.

## How
* Add local airbyte-cdk dependency to connector builder server: pip install -e $ROOT_DIR/airbyte/airbyte-cdk/python/ before running the uvicorn command or the tests
* Fix the code
* Fix the tests


airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
```diff
@@ -733,15 +733,15 @@ def test_resolve_manifest():
                 "requester": {
                     "path": "/v3/marketing/lists",
                     "authenticator": {"type": "BearerAuthenticator", "api_token": "{{ config.apikey }}"},
-                    "request_parameters": {"page_size": 10},
+                    "request_parameters": {"page_size": "10"},
```
clnoll, line 736: Out of curiosity, when did this change?

clnoll (approved):
LGTM @maxi297 just one (nonblocking) q.

airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
maxi297: This is a very interesting question. This change has been induced by https://github.com/airbytehq/airbyte/issues/19410. This was a little bit surprising to me because before this issue, I would have expected something like:
```
"requester": {
  "request_options_provider": {
    "request_parameters": {"page_size": "10"},
<...>
```

However, the `request_options_provider` seems to already be missing as if https://github.com/airbytehq/airbyte/issues/19410 was already done. So:
* Before updating the version: `requester. request_parameters` was ignored since it is not part of the schema
* After updating the version: `requester. request_parameters` started to be parsed (since we've removed the `request_option_parameters`. Since `request_parameters` expect a `Union[str, Mapping[str, str]]` and 10 was an integer and not a string

PR #21979, opened 2023-01-27
🪟 🎨 🐛 Update load more button style and margins in connection status page

edmundito:
## What
Updates the load button style and margins in the connection status page.

Before:
image

After:
image


teallarson (approved):
👍 looks good, did not test locally

PR #21988, opened 2023-01-27
Remove unused methods

benmoriceau:
## What
Remove unused methods that we using a variable name attemptId which was actually an attemptNumber.


airbyte-workers/src/test/java/io/airbyte/workers/temporal/scheduling/activities/JobCreationAndStatusUpdateActivityTest.java
```diff
@@ -196,42 +192,6 @@ void createResetJob() throws JsonValidationException, ConfigNotFoundException, I
       Assertions.assertThat(output.getJobId()).isEqualTo(JOB_ID);
     }
 
-    @Test
-    @DisplayName("Test attempt creation")
-    void createAttempt() throws IOException {
```
benmoriceau, line 201: This test can be removed because there is already one that call the other method.

colesnodgrass (approved):

cgardens (approved):

PR #21989, opened 2023-01-27
`AlwaysAllowedHosts`

### LINKED ISSUE #21952: The Platform can provide a static list of always-allowed hosts
e.g. all connectors can reach data dog

evantahler:
Closes https://github.com/airbytehq/airbyte/issues/21952

This PR gives the platform the ability to inject allowedHosts which are always valid for all connectors.  This includes endpoints we would use for logging or telemetry.

### DIFF
--- a/airbyte-config/config-models/src/main/java/io/airbyte/config/constants/AlwaysAllowedHosts.java
+++ b/airbyte-config/config-models/src/main/java/io/airbyte/config/constants/AlwaysAllowedHosts.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2022 Airbyte, Inc., all rights reserved.
+ */
+
+package io.airbyte.config.constants;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class AlwaysAllowedHosts {
+
+  private final List<String> hosts = new ArrayList<>();
+
+  public AlwaysAllowedHosts() {
+    // DataDog. See https://docs.datadoghq.com/agent/proxy/?tab=linux and change the location tabs
+    hosts.add("*.datadoghq.com");
+    hosts.add("*.datadoghq.eu");
+
+    // Sentry.  See https://docs.sentry.io/api/ for more information
+    hosts.add("*.sentry.io");
+  }
+
+  public List<String> getHosts() {
+    return hosts;
+  }
+
+}
--- a/airbyte-config/config-models/src/main/java/io/airbyte/config/constants/AlwaysAllowedHosts.java
+++ b/airbyte-config/config-models/src/main/java/io/airbyte/config/constants/AlwaysAllowedHosts.java
@@ -4,21 +4,17 @@
 
 package io.airbyte.config.constants;
 
-import java.util.ArrayList;
 import java.util.List;
 
 public class AlwaysAllowedHosts {
 
-  private final List<String> hosts = new ArrayList<>();
+  private final List<String> hosts = List.of(
+      // DataDog. See https://docs.datadoghq.com/agent/proxy/?tab=linux and change the location tabs
+      "*.datadoghq.com",
+      "*.datadoghq.eu",
 
-  public AlwaysAllowedHosts() {
-    // DataDog. See https://docs.datadoghq.com/agent/proxy/?tab=linux and change the location tabs
-    hosts.add("*.datadoghq.com");
-    hosts.add("*.datadoghq.eu");
-
-    // Sentry. See https://docs.sentry.io/api/ for more information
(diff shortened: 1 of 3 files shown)


evantahler:
@git-phu This is where we would add endpoints for Google Cloud logging and similar services.  I'm not sure what the full set of endpoints is, but this is a start.

airbyte-config/config-models/src/main/java/io/airbyte/config/constants/AlwaysAllowedHosts.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2022 Airbyte, Inc., all rights reserved.
+ */
+
+package io.airbyte.config.constants;
```
evantahler, line 5: Reviewers: Is this the right place to put something like this? 🤷 
This information will need to be used by the workers and possibly the orchestrator, so `airbyte-config` seemed like the right package.

airbyte-workers/src/main/java/io/airbyte/workers/utils/ConfigReplacer.java
```diff
@@ -89,6 +92,8 @@ public AllowedHosts getAllowedHosts(AllowedHosts allowedHosts, JsonNode config)
           "All allowedHosts values are un-replaced.  Check this connector's configuration or actor definition - " + allowedHosts.getHosts());
```
bechurch, line 92: What does it mean to be unreplaced? Is it that we just counldnt parse it and thus likely dont know how to interpret it at the firewall level?

bnchrch (commented):
All looks good to me, but again first week so Ill hold off approving. :)

airbyte-workers/src/main/java/io/airbyte/workers/utils/ConfigReplacer.java
evantahler: I'm worried about drift between the allowedHosts (`${host}`) and the config/spec of the connector (used to be `host` but was recently moved to `hostname`).  In that world, the sync would fail because access to the new `hostname` would be prevented.  At least this log message gives us a pointed that none of the allowedHost interpolations succeeded... maybe check that.

airbyte-config/config-models/src/main/java/io/airbyte/config/constants/AlwaysAllowedHosts.java
erohmensing: Looking at [Cole's chart](https://airbytehq-team.slack.com/archives/C04J3DJ0ADQ/p1674233851270909?thread_ts=1674233657.797829&cid=C04J3DJ0ADQ) of our current dependencies, seems like a reasonable spot if the orchestrator needs to get in on the action, otherwise `commons-worker` would probably be good?

erohmensing (approved):

airbyte-config/config-models/src/main/java/io/airbyte/config/constants/AlwaysAllowedHosts.java
colesnodgrass: `2022` so last year...

This feels like something that would be a better fit in a configuration file than hard-coded.

airbyte-config/config-models/src/main/java/io/airbyte/config/constants/AlwaysAllowedHosts.java
```diff
@@ -0,0 +1,27 @@
...
+    hosts.add("*.datadoghq.com");
+    hosts.add("*.datadoghq.eu");
+
+    // Sentry. See https://docs.sentry.io/api/ for more information
+    hosts.add("*.sentry.io");
+  }
+
+  public List<String> getHosts() {
+    return hosts;
+  }
```
colesnodgrass, line 25: This allows for the caller to change the `hosts` list.

airbyte-config/config-models/src/main/java/io/airbyte/config/constants/AlwaysAllowedHosts.java
evantahler: I have a suspicion that this list will become more complex than somethign we can statically represent... like... maybe based on which dataplane we are running on.

airbyte-config/config-models/src/main/java/io/airbyte/config/constants/AlwaysAllowedHosts.java
evantahler: Good call! udpated in 9845ef2ce49

Issue #21576, opened 2023-01-19
[Low-Code CDK] Rename DpathExtractor.field_pointer to field_path


brianjlai:
## Tell us about the problem you're trying to solve
**BREAKING CHANGE**

`field_pointer` is not quite the right descriptor for what the value of this field for the DpathExtractor component. It should instead be `field_path` because it is a path that is to be traversed to extract a value.

## Describe the solution you would like

DpathExtractor should have a `field_path` which is the traversal path to retrieve a value.

```
record_selector:
  type: RecordSelector
  extractor:
    type: DpathExtractor
    field_path: ["path", "to", "something"]
```

## Acceptance Criteria
- [ ] The declarative_component_schema.yaml should be updated w/ `field_pointer` rename to `field_path` for the DpathExtractor component
- [ ] Either a migration is included renaming all instances of `DpathExtractor.field_pointer`
- [ ] Documentation for the DpathExtractor has been updated
- [ ] An associated ticket to update the connector builder UI frontend has been filed

## Implementation Approach

- There are a lot of usage of `field_pointer`, it is convenient that according to the schema, there are no other usages of the name `field_pointer` (there is a `field_pointers` on the RemoveFields component) so a flat grep/sed could do the trick.

PR #21690, opened 2023-01-21
Low Code CDK: DpathExtractor extend

artem1205:
## What
Parse responses with nested lists (see tests)
Resolving https://github.com/airbytehq/airbyte/issues/20225

Complex response structure connectors:

1. [ChargeBee](https://github.com/airbytehq/airbyte/blob/96a2a18fb75bdd426aff265f253852fdf19a8e02/airbyte-integrations/connectors/source-chargebee/source_chargebee/components.py#L17-L57)
2. [Facebook Pages](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-facebook-pages/source_facebook_pages/components.py#L55-L97)
3. [Monday](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-monday/source_monday/dpath_string_extractor.py)

## How
Use dpath.util.values in DpathExtractor

## Recommended reading order
1. `y.python`

## 🚨 User Impact 🚨
Are there any breaking changes? What is the end result perceived by the user? If yes, please merge this PR with the 🚨🚨 emoji so changelog authors can further highlight this if needed.

## Pre-merge Checklist
Expand the relevant checklist and delete the others.

<details><summary><strong>Updating a connector</strong></summary>

### Community member or Airbyter

- [ ] Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
- [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`
    - [ ] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - [ ] Changelog updated in `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
[...]

PR #21990, opened 2023-01-27
Issue #21576 rename dpathextractor fieldpointer

maxi297:
## What
Addresses #21576 

## How
* Rename DpathExtractor.field_pointer to field_path in declarative_component_schema and regenerate files
* Update DpathExtractor.field_pointer to field_path (not the model but the actual component we use)
* Fixes sources (some source-monday and source-pocket were re-implementing DpathExtractor)
* Update Connector Builder Server
* Migration (see https://github.com/airbytehq/airbyte/pull/21997)

## Recommended reading order
2. `airbyte-cdk/python`
3. `airbyte-integrations`
4. `airbyte-connector-builder-server`

## 🚨 User Impact 🚨
This is a breaking change that is part of the Low-Code CDK to Beta


airbyte-connector-builder-server/connector_builder/generated/models/dpath_extractor.py
```diff
@@ -19,13 +19,13 @@ class DpathExtractor(BaseModel):
...
     DpathExtractor - a model defined in OpenAPI
 
-        field_pointer: The field_pointer of this DpathExtractor.
+        field_path: The field_path of this DpathExtractor.
         config: The config of this DpathExtractor.
         decoder: The decoder of this DpathExtractor [Optional].
     """
 
-    field_pointer: List[AnyOfInterpolatedStringstring] = Field(alias="field_pointer")
+    field_path: List[AnyOfInterpolatedStringstring] = Field(alias="field_path")
```
maxi297, line 27: This file was autogenerated in December 2022. However, I can't find what is the process to re-generate this. I see [this](https://github.com/airbytehq/airbyte/blob/ca8cdc4/airbyte-connector-builder-server/src/main/openapi/openapi.yaml#L222) has been commented out but I don’t see any alternative solution provided. Also, the file used seems to be the previous json file and not `declarative_component_schema.yaml`. Based on the commit/issues associated, it seems there was a lot of struggle with code generation. Therefore, I have a couple of questions:
* Do we need code generation for the manifest in the Connector Server Builder?
* If so, should we clarify the state of the code generation in our README.md?
* If not, should we clean `openapi.yaml` file?

airbyte-integrations/connectors/source-facebook-pages/source_facebook_pages/components.py
```diff
@@ -83,12 +83,12 @@ class NestedDpathExtractor(DpathExtractor):
 
     def extract_records(self, response: requests.Response) -> List[Record]:
         response_body = self.decoder.decode(response)
-        if len(self.field_pointer) == 0:
```
maxi297, line 86: `NestedDpathExtractor` seems to be something that would be interesting in the airbyte-cdk per se. If we agree, I could create a Continuous Improvement ticket to migrate it (or make DpathExtractor support the same functionality)

Note: To be clear, I haven't checked much the difference and I'm not clear on what the gap between `DpathExtractor` and `NestedDpathExtractor` is

airbyte-integrations/connectors/source-facebook-pages/unit_tests/test_facebook_authenticator.py
```diff
@@ -19,13 +19,13 @@ def test_facebook_url_params(req_mock):
         "access_token": "initial_token",
         "page_id": "pageID"
     }
-    options = config
+    parameters = config
```
maxi297, line 22: The changes from `options` to `parameters` was part of https://github.com/airbytehq/airbyte/issues/19408 but some sources were not updated. I updated the sources that were impacted by my changes

airbyte-integrations/connectors/source-facebook-pages/source_facebook_pages/components.py
sherifnada: I think we can deprecate the nested one: https://github.com/airbytehq/airbyte/pull/21690

airbyte-cdk/python/bin/migrate_field_point_to_field_path.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
sherifnada, line 1: request: let's make sure we remove these scripts after beta migration is done

sherifnada (approved):

clnoll (approved):
LGTM @maxi297!

airbyte-integrations/connectors/source-facebook-pages/source_facebook_pages/components.py
maxi297: Confirmed with artem1205 that GL will remove this eventually

PR #21995, opened 2023-01-27
[Connector Builder] Set default value for end_datetime

### LINKED ISSUE #21572: [Connector Builder] Set default values for end datetime
## Context

It is very common for users to use `{{ now_utc() }}` as the value for the `end_datetime` field in order to keep their data fully up-to-date.

Until https://github.com/airbytehq/airbyte/issues/21571 is complete (which makes this field optional), the connector builder should automatically set this as the value of the end_datetime field, as this will be the value users usually want anyway and it will also teach them that this macro exists.

Similarly, we should set a sensible default for the `start_datetime` field so that users don't need to configure that unless they want to use a specific value.

## Acceptance Criteria
- `end_datetime` is prefilled with a value of `{{ now_utc() }}` when the datetime stream slicer is selected in the builder UI
~- `start_datetime` is prefilled with a value of now_utc minus 1 year when the datetime stream slicer is selected in the builder UI~
  - Removing this as it will result in confusing UX for either connector developer or user

Open question:
~- Is now - 1 year the right time range to use as the default for start_datetime? Or would a month or a day be more sensible?~
- Several connectors are also setting the format along with `now_utc()`, e.g. [here](https://github.com/airbytehq/airbyte/blob/59ecaeaf1557b231a2c5f1b27cd5e692c78af9fb/airbyte-integrations/connectors/source-posthog/source_posthog/posthog.yaml#L131) `datetime: "{{ now_utc().strftime('%Y-%m-%dT%H:%M:%S%z') }}"`. Can the CDK automatically determine the right format to use if end_datetime is not set?
  - This is likely not actually necessary to do. @lmossman to create ticket to validate this
- Some connectors use `today_utc()` instead of `now_utc()`, e.g. [here](https://github.com/airbytehq/airbyte/blob/59ecaeaf1557b231a2c5f1b27cd5e692c78af9fb/airbyte-integrations/connectors/source-callrail/source_callrail/callrail.yaml#L26). Can the CDK automatically determine if `today_` should be used instead of `now_`?

lmossman:
## What
Resolves https://github.com/airbytehq/airbyte/issues/21572

## How
Sets the default value of `end_datetime` to `{{ now_utc() }}` using the existing BuilderOneOf default value mechanism.

### DIFF
--- a/airbyte-webapp/src/components/connectorBuilder/Builder/StreamSlicerSection.tsx
+++ b/airbyte-webapp/src/components/connectorBuilder/Builder/StreamSlicerSection.tsx
@@ -82,7 +82,7 @@ export const StreamSlicerSection: React.FC<StreamSlicerSectionProps> = ({ stream
       label: "Datetime",
       typeValue: "DatetimeStreamSlicer",
       default: {
-        datetime_format: "",
+        datetime_format: "%Y-%m-%d %H:%M:%S.%f+00:00",
         start_datetime: "",
         end_datetime: "{{ now_utc() }}",
         step: "",
--- a/airbyte-webapp/src/components/connectorBuilder/Builder/StreamSlicerSection.tsx
+++ b/airbyte-webapp/src/components/connectorBuilder/Builder/StreamSlicerSection.tsx
@@ -84,7 +84,7 @@ export const StreamSlicerSection: React.FC<StreamSlicerSectionProps> = ({ stream
       default: {
         datetime_format: "",
         start_datetime: "",
-        end_datetime: "",
+        end_datetime: "{{ now_utc() }}",
         step: "",
         cursor_field: "",
       },


flash1293:
This behaves as discussed but I'm just wondering... this only works if the user _also_ sets the date format on the slicer correctly which is what is used for the end datetime, right? Should we set it to the format `now_utc()` generates as well? 
Screenshot 2023-01-30 at 17 52 34

Otherwise it's actually quite hard to figure out what to do about it - I had to look through existing manifests to figure out that it's `%Y-%m-%d %H:%M:%S.%f+00:00`

lmossman:
@flash1293 I think your suggestion makes sense; setting the format will help people who don't want to change anything here, and will also give an example of a format that can be used.

flash1293 (approved):
LGTM

PR #21998, opened 2023-01-27
🪟Auto detect schema: Remove backdrop on non-breaking change, add toast on no refresh

edmundito:
## What
This updates a couple of areas based on feedback on the feature:
* Removes the backdrop that blocks the connection from being edited, at least for now. Users are still warned that the schema has changed and they will now be allowed to take the risk of updating the replication settings
* When the schema is refreshed, and there are no updates, it will notify the user with a toast. Additionally, the status of the schema update is updated.

## How
* Removes logic
* For showing the toast, the connection is now updated with the refreshed schema change in case the state is no longer the previous state, so the auto-detect schema warnings can disappear after a refresh.


dizel852 (approved):
LGTM 👍 
Tested locally - works as expected

teallarson (approved):
Tested locally 👍

PR #21759, opened 2023-01-23
🪟 🐛 Long source and destination names make the connection list table scroll horizontally

m19berger:
## What

Closes [#21274](https://github.com/airbytehq/airbyte/issues/21274)

## How

- Added responsive columns
- Created new cell components
- Removed styled-components
- Refactored Table component

_Note: This PR is a temporary solution that fixes some bugs for this moment before we can start using the new Table component_

## Loom
https://www.loom.com/share/a442f62c9dd1406691b7e4af19d89c40

PR #22002, opened 2023-01-27
🪟 🐛 Fix loading state of "Launch" button in connection table

### LINKED ISSUE #20729: No feedback after launching the sync manually from `Connections` tab
## Current Behavior
There is no feedback after launching sync manually.
https://www.loom.com/share/09a4ce52fdd94f40bf5562807378c8f6

## Expected Behavior
The button should change the state to "In progress"

joeykmh:
## What
Fixes the loading state of the manual sync button on the connection table.

Closes https://github.com/airbytehq/airbyte/issues/20729

Before, there was no feedback for the user when clicking "Launch" on a manual sync, requiring the user to refresh the page to see that the sync started:

https://user-images.githubusercontent.com/7550957/215260933-7d0e6a53-e9d8-4a5a-9279-782ed1580569.mov

Now there is a loading state, the sync changes to "In progress" and its status is updated every 60 seconds (in this video it updates after 10 seconds to illustrate):

https://user-images.githubusercontent.com/7550957/215263313-47892172-b303-45a9-9448-a3f7a59413af.mov

There is also error handling if the manual sync request fails:

https://user-images.githubusercontent.com/7550957/215263328-68396b03-2514-43d5-989a-fa4d9e662b72.mov

## How
- Refactors the code a bit to avoid prop drilling and repeated definition of the `onSync` callback
- Adds a loading state from the mutation
- Refreshes/invalidates the connection list after the manual sync is triggered
- Adds a 60-second refresh interval to the connection table, so a user doesn't have to refresh the page to see the sync 
status update
- Changed the button text from "Launch" to "Start sync", which matches what we do on the connection page (we don't use "Launch" anywhere else in the app).

## Recommended reading order
1. `StatusCellControl.tsx` and `useConnectionHook.tsx` have the main changes to logic
2. The rest


airbyte-webapp/src/hooks/services/useConnectionHook.tsx
```diff
@@ -239,9 +267,12 @@ export const useRemoveConnectionsFromList = (): ((connectionIds: string[]) => vo
...
   const workspace = useCurrentWorkspace();
   const service = useWebConnectionService();
+  const REFETCH_CONNECTION_LIST_INTERVAL = 60_000;
 
-  return useSuspenseQuery(connectionsKeys.lists(payload.destinationId ?? payload.sourceId), () =>
-    service.list({ ...payload, workspaceId: workspace.workspaceId })
+  return useSuspenseQuery(
+    connectionsKeys.lists(payload.destinationId ?? payload.sourceId),
+    () => service.list({ ...payload, workspaceId: workspace.workspaceId }),
+    { refetchInterval: REFETCH_CONNECTION_LIST_INTERVAL }
```
josephkmh, line 275: I don't see any reason against re-fetching the connections every 60 seconds. That way we can show updated progress in the connection table. p99 latency for this endpoint is under 100ms, so I don't think it's a performance problem.

airbyte-webapp/src/components/EntityTable/components/StatusCellControl.tsx
```diff
@@ -15,30 +15,28 @@ interface StatusCellControlProps {
...
+  const { mutateAsync: syncConnection, isLoading: isSyncStarting } = useSyncConnection();
+
+  const onRunManualSync = (event: React.SyntheticEvent) => {
+    event.stopPropagation();
+
+    const connection = connections.find((c) => c.connectionId === id);
+    if (connection) {
+      syncConnection(connection);
+    }
+  };
```
josephkmh, line 39: This function was basically constructed in three places before (`SourceConnectionTable`, `DestinationConnectionTable` and `ConnectionTable`) and drilled down as prop called `onSync`. Instead we can just create the function here and remove all that intermediate code.

edmundito:
From watching the video, the only feedback I have is to make sure that the Enabled column is fixed so that when the button is pressed it doesn't reflow the table. However, #21759 will do some table cleanup.

josephkmh:
> From watching the video, the only feedback I have is to make sure that the Enabled column is fixed so that when the button is pressed it doesn't reflow the table. However, https://github.com/airbytehq/airbyte/pull/21759 will do some table cleanup.

Yeah there is some reflow of the table, but I think it's rather due to the "Last sync" column (see video below at 9-12 seconds). I'd rather keep that in a separate PR (probably the one you linked?) since the behavior of the columns should be fixed holistically.


https://user-images.githubusercontent.com/7550957/215524936-2c54133c-2a06-4bf1-a420-d3afd7b06e5f.mov

airbyte-webapp/src/components/EntityTable/components/StatusCellControl.module.scss
```diff
@@ -0,0 +1,5 @@
+.inProgressLabel {
+  height: 32px; /** Needs to match our Button height to avoid jumping around */
```
edmundito, line 2: Should we make this a variable? I've also had another occasion where I needed the button height

airbyte-webapp/src/hooks/services/useConnectionHook.tsx
```diff
@@ -74,20 +79,43 @@ export function useConnectionService() {
...
+      onError: (error: Error) => {
+        trackError(error);
+        notificationService.registerNotification({
+          id: `tables.startSyncError.${error.message}`,
+          text: `${formatMessage({ id: "connection.startSyncError" })}: ${error.message}`,
+          type: ToastType.ERROR,
+        });
+      },
+      onSuccess: async () => {
+        webConnectionService
```
edmundito, line 113: We may need to `await` here. I noticed that the button goes into a non-loading state for the time the list is loading.

edmundito (changes requested):
Code looking good, tested locally and noticed one detail.

airbyte-webapp/src/hooks/services/useConnectionHook.tsx
josephkmh: good point! fixed ✅

airbyte-webapp/src/components/EntityTable/components/StatusCellControl.module.scss
josephkmh: The reason I didn't is that IMO we should not really need to pass around the height of the button. The table row should have an appropriate (larger) height so that it's not reflowing depending on the content. But of course that's a little out of scope. On second thought, a variable is at least better than a comment like this, so I'll add the button heights to `_variables.scss`

airbyte-webapp/src/scss/_variables.scss
```diff
@@ -37,3 +37,7 @@ $width-modal-sm: 492px;
 $width-modal-md: 585px;
 $width-modal-lg: 940px;
 $width-modal-xl: 1008px;
+
+$buttonHeightXs: 32px;
```
edmundito, line 41: Nit: the convention for var names in SCSS is `with-dashes`

edmundito (approved):
Re-tested and looks good. Just maybe one nit that should be fixed to stay consistent, but no need for re-approval.

airbyte-webapp/src/scss/_variables.scss
josephkmh: Thanks! Fixed that

PR #20973, opened 2023-01-03
🐛 Source Marketo: fix encoding error for Lead sync

CyprienBarbault:
## What
This is a fix for an issue I reported -> https://github.com/airbytehq/airbyte/issues/20641

When syncing Marketo lead, it appears that those with name containing non-latin character would make the sync fail.  In particular we have a Lead whose name is Vietnamese (cf https://en.wikipedia.org/wiki/Vietnamese_name for exemple names)

In fact, the response from the marketo API was being fall-backed to the default encoding `ISO-8859-1` instead of `utf-8`  (cf https://github.com/psf/requests/issues/5445#issuecomment-661654120)

## How
Forcing the encoding to `utf-8` fixed the issue

## Recommended reading order
1. `source.py`

## 🚨 User Impact 🚨
Are there any breaking changes? What is the end result perceived by the user? If yes, please merge this PR with the 🚨🚨 emoji so changelog authors can further highlight this if needed.

I'm not sure but I don't think so

## Pre-merge Checklist

<summary><strong>Updating a connector</strong></summary>

### Community member or Airbyter

- [x] Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
- [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`
    - [ ] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - [ ] Changelog updated in `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
- [x] PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests)

### Airbyter

[...]

PR #22015, opened 2023-01-27
Turn off default httpavailabilitystrategy for source-marketo (GA)

erohmensing:
See https://github.com/airbytehq/airbyte/pull/21888 for description


evantahler (approved):

erohmensing:
Same failure as in [#connector-health](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1674824837998149)

evantahler:
Then I think it's probably OK to merge

erohmensing:
Thanks, will do. Double-checking all others for which this is the case and posting those as well!

erohmensing:
Apparently this version was published 11 days ago without a changelog, coming back to look into it

erohmensing:
It was released off of this PR that was never merged: https://github.com/airbytehq/airbyte/pull/20973. Seems to have been published by a contributor while skipping the tests

erohmensing:
If I try to push a new version of the connector, I'd essentially be undoing the changes from this PR for the latest version of the connector published. That would affect OSS users who upgraded to this version because they were seeing the issue linked in that PR. 

However, cloud is still seeing 0.1.11, since the cloud catalog was never updated, since the PR was merged. Therefore cloud users should see no effects of going from 0.1.11 to 0.1.13. Users who want 0.1.12 in OSS can still use that version as they see fit, and then when that PR is merged, the changes will be reflected in 0.1.14 (or whatever version is next when it goes in) for both cloud and OSS users. Does that sound good @evantahler? 

I suppose I'd add the changelog entry from 0.1.12 too, and make it clear that 0.1.13 reverts those changes

evantahler:
> However, cloud is still seeing 0.1.11, since the cloud catalog was never updated, since the PR was merged. Therefore cloud users should see no effects of going from 0.1.11 to 0.1.13. Users who want 0.1.12 in OSS can still use that version as they see fit, and then when that PR is merged, the changes will be reflected in 0.1.14 (or whatever version is next when it goes in) for both cloud and OSS users. Does that sound good @evantahler?

Yep! Although, it's a shame we got into this position.

Alternatively, you could merge in https://github.com/airbytehq/airbyte/pull/20973 and then merge this after?

erohmensing:
I would, except for the fact that the acceptance tests went from consistently succeeding to consistently failing. I think that PR needs some more work before it goes in 😕 will run tests on that branch

erohmensing:
Actually, now that I think about it, it seems weird that the connector would start failing tests on the #connector-health check since they're built off of master. So maybe that is actually unrelated

erohmensing:
Merged the contributor PR - someone had already released a newer version. Releasing this version should re-introduce the contributor's changes lost in the 1.0.0 version as well as add my changes.

PR #22035, opened 2023-01-27
🪟 🔧 Add typescript version to VSCode settings

joeykmh:
## What
Adds a link to the `airbyte-webapp` typescript module so that VSCode can correctly suggest a matching TS version to use for its checker.


lmossman (approved):
Change LGTM, tested locally and it fixed the issue

PR #22036, opened 2023-01-27
Add a feature flag for a failure because of an heartbeat failed check

benmoriceau:
## What
Add a feature flag that will then be used to protect jobs from failing because of a missed heartbeat.


configs/flags.yml
```diff
@@ -1,3 +1,5 @@
 flags:
   - name: performance.backgroundJsonSchemaValidation
     enabled: true
+  - name: heartbeat.fail_sync
```
colesnodgrass, line 4: The docs are up yet, but the existing feature-flags are all camelCased.  Could we camelCase this one for consistency?

colesnodgrass (approved):
- camelCase the flag for consistency with other flags
- create this flag in LD (if you haven't yet)

configs/flags.yml
benmoriceau: Done

jdpgrailsdev (approved):
:shipit:

colesnodgrass:
@benmoriceau heads-up that the `flags.yml` file was moved to the root directory in #22158

benmoriceau:
@colesnodgrass could you review the flag in LD? https://app.launchdarkly.com/default/production/features/heartbeat.failSync/targeting It is the first one, I'm creating and I would like to make sure that I didn't do a mistake.

PR #22048, opened 2023-01-28
Add version check on normalization

gosusnp:
## What

Moving forward, the platform is ensuring that the normalization always sees data types v1.
Add a check to make sure we do not run an older version of normalization with the newer data types as it would lead to the columns being typed to the most generic type. (JSONB for postgresql for example)

## How

Add a check on normalization version and fail with an explicit message.


edgao (approved):

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
```diff
@@ -152,6 +164,18 @@ public NormalizationInput generateNormalizationInput(final StandardSyncInput syn
         .withResourceRequirements(normalizationResourceRequirements);
   }
 
+  static boolean normalizationSupportsV1DataTypes(final IntegrationLauncherConfig destinationLauncherConfig) {
```
benmoriceau, line 167: Nit: Add a `@VisibleForTesting`

benmoriceau (approved):

PR #22053, opened 2023-01-28
🪟 🔧 Switch to `pnpm` for package managing

### LINKED ISSUE #21482: Update local webapp development instructions
### Tell us about the documentation you'd like us to add or update.
[Local webapp development.](https://docs.airbyte.com/contributing-to-airbyte/developing-locally#develop-on-airbyte-webapp)
Add `nvm install` command prior to `npm install`

### If applicable, add links to the relevant docs that should be updated
- [Develop on airbyte-webapp](https://docs.airbyte.com/contributing-to-airbyte/developing-locally#develop-on-airbyte-webapp)

timroes:
## What

Switches from using `npm` to `pnpm` for package managing in `airbyte-webapp`.

To use it locally you need to install the exact matching `pnpm` version:

```bash
# <version> must be the exact version from package.json > engines.pnpm
npm install -g pnpm@<version>
# Using it
pnpm i
pnpm start
# ...
```

Closes https://github.com/airbytehq/airbyte/issues/21482

## How

* Replace the `package-lock.json` with the `pnpm-lock.yaml` file.
* Switch Gradle tasks over to use `pnpm` instead of `npm`, also add an additional task that runs before `pnpmInstall` that validates we only have the `pnpm-lock.yml` file and no other lock file (to prevent accidental committing of a new `package-lock.json` file).
* `pnpm` has a patch mechanism built in, so I could remove the nasty `react-virtualize` patch code from Vite and instead use the proper `pnpm` patch mechanism for it.
* Couple of minor required ESLint and TS changes, that were caused by `pnpm` pulling in slightly newer versions.
* Due to differences in some libraries version they also came with new SVGs, so I needed to udpate several snapshot tests.


airbyte-webapp/build.gradle
```diff
@@ -1,28 +1,46 @@
 plugins {
     id "base"
-    id "com.github.node-gradle.node" version "3.3.0"
+    id "com.github.node-gradle.node" version "3.4.0"
```
timroes, line 3: :information_source: Required to have `PnpmTask` available.

airbyte-webapp/package.json
```diff
@@ -3,17 +3,18 @@
...
   },
   "scripts": {
     "prepare": "cd .. && husky install airbyte-webapp/.husky",
-    "prestart": "npm run generate-client",
-    "start": "node -r ./scripts/dev-overwrites.js ./node_modules/.bin/vite",
-    "prestart:cloud": "npm run generate-client",
-    "start:cloud": "AB_ENV=${AB_ENV-frontend-dev} node -r ./scripts/environment.js -r ./scripts/dev-overwrites.js ./node_modules/.bin/vite",
-    "prebuild": "npm run generate-client",
+    "prestart": "pnpm run generate-client",
+    "start": "NODE_OPTIONS='-r ./scripts/dev-overwrites.js' vite",
```
timroes, line 12: :information_source: With `pnpm` the `node_modules/.bin/vite` script is actually a wrapping Bash script not the `js` file of `vite` directly anymore, so we can't call it with `node` anymore, so using `NODE_OPTIONS` to pass down the `-r` parameter to the `node` process that will execute `vite`.

airbyte-webapp/src/components/connectorBuilder/useManifestToBuilderForm.ts
```diff
@@ -203,18 +200,18 @@ function manifestStreamSlicerToBuilder(
   }
 
   if (manifestStreamSlicer.type === "DatetimeStreamSlicer") {
-    const datetimeStreamSlicer = manifestStreamSlicer as DatetimeStreamSlicer;
+    const datetimeStreamSlicer = manifestStreamSlicer;
```
timroes, line 203: :information_source: I believe `pnpm` pulled now in a newer version of `orval` which generates the types slightly better, so this isn't needed anymore.

airbyte-webapp/src/components/ui/Markdown/remarkAdmonitionsPlugin.ts
```diff
@@ -11,7 +10,7 @@ import styles from "./admonitions.module.scss";
 const SUPPORTED_ADMONITION_NAMES: Readonly<string[]> = ["note", "tip", "info", "caution", "warning", "danger"];
 const SUPPORTED_NODE_TYPES: Readonly<string[]> = ["textDirective", "leafDirective", "containerDirective"];
 
-export const remarkAdmonitionsPlugin: Plugin<[], Root> = () => (tree) => {
+export const remarkAdmonitionsPlugin: Plugin<[]> = () => (tree) => {
```
timroes, line 13: :information_source: Typings works very fine with the `Root` so no need to pull in `mdast` (which is also the deprecated version of remark) for that.

airbyte-webapp/src/components/ui/NextTable/NextTable.tsx
```diff
@@ -26,7 +27,7 @@ export const NextTable = <TData,>({ columns, data, onClickRow, className }: Prop
             {headerGroup.headers.map((header) => (
               <th
                 colSpan={header.colSpan}
-                className={classNames(styles.th, header.column.columnDef.meta?.thClassName)}
+                className={classNames(styles.th, (header.column.columnDef.meta as ColumnMeta | undefined)?.thClassName)}
```
timroes, line 30: :information_source: The type overwrite via `declare` doesn't work with `pnpm` anymore (I don't fully understand why, yet), so moving it into an in place cast instead.

airbyte-webapp/.eslintrc.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
     "plugin:jsx-a11y/recommended",
     "plugin:@airbyte/recommended",
   ],
-  plugins: ["react", "@typescript-eslint", "prettier", "unused-imports", "css-modules", "jsx-a11y", "@airbyte"],
```
timroes, line 12: :information_source: This is with `pnpm` now conflicting with the `react` plugin already being pulled in by `eslint-config-react-app` which we extend above. Since that plugin pulls in the `react` plugin also no need for us to pull it in explicitally.

airbyte-webapp/.npmrc
```diff
@@ -1 +1,2 @@
 engine-strict=true
+enable-pre-post-scripts=true
```
timroes, line 2: :information_source: Required for `pnpm` to execute the `prestart` etc scripts. It doesn't do so by default, but we're using the heavily for Orval compilation.

airbyte-webapp/package.json
```diff
@@ -112,7 +112,6 @@
     "@types/json-schema": "^7.0.11",
     "@types/lodash": "^4.14.182",
     "@types/node": "^17.0.40",
-    "@types/query-string": "^6.3.0",
```
timroes, line 115: :information_source: Removed some deprecated package (this was a stub by now, `query-string` include the types itself).

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -126,6 +126,7 @@ definitions:
       - stream_slicers
     properties:
       type:
+        type: string
```
timroes, line 129: :information_source: We'll always need a `type: string` beside the single element `enum`, to make sure Orval can generate the actual type properly, and not render it as `unknown`. cc @flash1293 @lmossman

airbyte-webapp/package.json
```diff
@@ -68,7 +69,7 @@
     "react-paginate": "^8.1.3",
     "react-query": "^3.39.1",
     "react-reflex": "^4.0.9",
-    "react-router-dom": "^6.3.0",
+    "react-router-dom": "6.3.0",
```
timroes, line 72: :information_source: Need to fix that version for now, due to `useBlocker`. This uses unstable code that's no longer there in the most recent react-router-dom version. We should rewrite that code to the newest `unstable_useBlocker` code, but not as part of this PR :D

airbyte-webapp/src/components/connection/CatalogTree/next/__snapshots__/BulkEditPanel.test.tsx.snap
```diff
@@ -98,7 +98,7 @@ exports[`<BulkEditPanel /> should render 1`] = `
                     xmlns="http://www.w3.org/2000/svg"
                   >
                     <path
-                      d="M310.6 246.6l-127.1 128C176.4 380.9 168.2 384 160 384s-16.38-3.125-22.63-9.375l-127.1-128C.2244 237.5-2.516 223.7 2.438 211.8S19.07 192 32 192h255.1c12.94 0 24.62 7.781 29.58 19.75S319.8 237.5 310.6 246.6z"
+                      d="M137.4 374.6c12.5 12.5 32.8 12.5 45.3 0l128-128c9.2-9.2 11.9-22.9 6.9-34.9s-16.6-19.8-29.6-19.8L32 192c-12.9 0-24.6 7.8-29.6 19.8s-2.2 25.7 6.9 34.9l128 128z"
```
timroes, line 101: :information_source: Snaptshots updated due to newer versions of those libraries.

evantahler:
One thing we did at $previous-oss-company was add a [`preinstall`](https://github.com/grouparoo/grouparoo/blob/main/package.json#L21) script which [yelled at you](https://github.com/grouparoo/grouparoo/blob/main/bin/preinstall) if you weren't using pnpm.  Folks more-or-less expect that `yarn install` and `npm install` (1) to both work and (2) work equivalently.  We wanted to make it really clear that we _require_ pnpm, and it's not just a suggestion.

timroes:
@evantahler I considered that, the problem is, it's really "slow"... like adding 2-3 seconds. I wanted to avoid that, given that one of the reasons for `pnpm` was the performance benefit (besides more reliable package resolution). That's why I added the gradle build script, so while in theory you can screw up locally (if you don't read the docs), it anyway won't pass CI (in case you didn't notice that you accidentally created a new lock file which didn't exist beforehand). So we'll still require `pnpm` for installation, but unless we see many people stumbling across this, rather prefer not adding that overhead of executing `npx` in `preinstall`.

airbyte-webapp/package.json
```diff
@@ -3,17 +3,18 @@
...
   },
   "scripts": {
     "prepare": "cd .. && husky install airbyte-webapp/.husky",
-    "prestart": "npm run generate-client",
-    "start": "node -r ./scripts/dev-overwrites.js ./node_modules/.bin/vite",
-    "prestart:cloud": "npm run generate-client",
-    "start:cloud": "AB_ENV=${AB_ENV-frontend-dev} node -r ./scripts/environment.js -r ./scripts/dev-overwrites.js ./node_modules/.bin/vite",
-    "prebuild": "npm run generate-client",
+    "prestart": "pnpm run generate-client",
+    "start": "NODE_OPTIONS='-r ./scripts/dev-overwrites.js' vite",
```
lmossman, line 12: We should update the airbyte-webapp/README.md, since it still contains `npm` commands

lmossman (commented):
Left some small comments and questions, but overall this seems to work well. Tried clicking around both the Airbyte webapp and the Connector Builder and both seemed to be working as expected.

I'm not going to approve just yet since I had some questions and I'd like some others to take a look at this as well, but let me know if you'd like me to take another look and approve if necessary!

airbyte-webapp/package.json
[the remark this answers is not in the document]
timroes: This should be the same as it's currently. This wasn't introduced by `pnpm` but by Vite. They appear multiple times, because Vite forks of workers to run ESLint, TypeScript compiler and Stylelint in, which all run through that same setup code, so you'll see it 4 times. I can take a look (in a different PR), if there's an easy way I can make sure I only log that output if I'm in the "main" process of Vite.

timroes: Very good call. I've actually rewritten the README now completely to have less outdated information and just point to our official documentation for development information, so we don't have this duplicated across different places.

airbyte-webapp/package.json
timroes: I did explicitally NOT mean to update the version. The newer version breaks our `useBlocker` hook, since that was build using internal APIs that are not available in newer versions anymore. So I set a fixed version for now (since `pnpm` would have pulled in a newer one). I've created https://github.com/airbytehq/airbyte/issues/22305 to track rewriting this code and updating.

airbyte-webapp/package.json
```diff
@@ -3,17 +3,18 @@
   "version": "0.40.32",
   "private": true,
   "engines": {
-    "node": "16.18.1"
+    "node": "16.18.1",
+    "pnpm": "7.26.2"
   },
   "scripts": {
```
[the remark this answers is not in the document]
timroes: There is actually no difference in using `npm start` or `pnpm start`. The only difference is for managing libraries, why we make sure we enforce on CI that no one does that wrong and switched accidentally package managers again. Locally I considered (see also Evan's and my comments on this PR) adding a check, but the problem is that it adds "seconds" to the command, which given that `pnpm i` might often only take 2s is a bit frustrating so I decided to get rid of it. Also in addition, it won't really save you much time, since the `preinstall` command actually runs... after install since npm 7 (yeah I know very intuitive), so you'd anyway need to go through the full install first before we print you a warning that you used the wrong command (which you also will see by suddenly a new lock file appearing). And for any of the run commands it's not really needed to check, since there is no difference, so I felt we don't need to add any local checks for this. But please feel free to share opposing opinions.

airbyte-webapp/package.json
```diff
@@ -162,9 +166,9 @@
     "timezone-mock": "^1.3.4",
     "tmpl": "^1.0.5",
     "ts-node": "^10.8.1",
-    "typescript": "^4.7.3",
+    "typescript": "^4.9.5",
```
josephkmh, line 169: 🥳

joeykmh (commented):
Nice! I spun it up and everything I tested seemed to be working properly:
- OSS mode, clicking around & changing source code
- Cloud mode, clicking around & changing source code
- Storybook
- Local build command

Could we also update `airbyte-webapp-e2e-tests/README.md` to reference `pnpm` commands? I don't think we need necessarily to run cypress with `pnpm`, but there are some explicit instructions there about installing and running `airbyte-webapp` with `npm`.

Small note: we'll also want to update the frontend notion page to replace `npm` commands with `pnpm` once this is merged

airbyte-webapp/package.json
lmossman: Ah okay, I had just misunderstood this then. That makes sense!

airbyte-webapp/package.json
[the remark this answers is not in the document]
lmossman: Makes sense to me! Thanks for explaining - I think the approach you've used here sounds good

lmossman (approved):

docs/contributing-to-airbyte/developing-locally.md
```diff
@@ -174,12 +174,26 @@ BASIC_AUTH_USERNAME="" BASIC_AUTH_PASSWORD="" docker compose up
...
 cd airbyte-webapp
-npm install
-npm start
+nvm install
+```
+
+- Install the `pnpm` package manager in the required version:
+
+```bash
+# <version> must be the exact version from airbyte-webapp/package.json > engines.pnpm
```
flash1293, line 188: Not a blocker for this PR, but it looks like pnpm ships with a node version manager too, should we just use that one? One tool less

https://pnpm.io/cli/env

Didn't look into it in detail though

flash1293 (approved):
Tested and everything still seems to work, lgtm.

Should we switch over the e2e test package as well for consistency reasons?

docs/contributing-to-airbyte/developing-locally.md
timroes: I have looked a bit into it, but currently didn't saw much of a benefit over keep using `nvm` (also we anyway need `npm` first to install `pnpm`, so we either need to guide to `nvm` or a local node installation in our docs).

timroes:
> Should we switch over the e2e test package as well for consistency reasons?

Given we barely update anything in there and don't have much problems with `npm` there, I don't see a reason why... I also don't see a reason why not :shrug: (but would prefer doing it in a separate PR :D)

docs/contributing-to-airbyte/developing-locally.md
flash1293: You can install pnpm without npm, right? https://pnpm.io/installation

flash1293:
>  I don't see a reason why... I also don't see a reason why not

A user needs to make sure to call either npm or pnpm, depending on whether they are working in e2e or regular webapp (and keep both of them around and up to date)

timroes:
> A user needs to make sure to call either npm or pnpm, depending on whether they are working in e2e or regular webapp (and keep both of them around and up to date)

Problem is, we already have some other modules (the doc system) using yarn anyway, so I feel it's not that bad having `npm` for e2e tests, but I'll change it in a separate PR after this is merged.

flash1293:
> Problem is, we already have some other modules (the doc system) using yarn anyway, so I feel it's not that bad having npm for e2e tests, but I'll change it in a separate PR after this is merged.

Good thing you mention this, that was a little annoying as well :) There's a lot of different technologies in this repo, would be nice to consolidate where possible.

airbyte-webapp/package.json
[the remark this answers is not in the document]
timroes: I've opened https://github.com/airbytehq/airbyte/pull/22409 which will improve this, as requested by you.

PR #22056, opened 2023-01-28
🪟 🔧 Upgrade to Storybook 7

timroes:
## What

This upgraded Storybook to the new Storybook 7 Beta. Since we're currently blocked on the switch to `pnpm` on Storybook, I already updated now during the Beta phase. This also now makes sure to use Vite for building storybook instead of webpack.

Please regard the inline comments for additional details.


airbyte-webapp/.storybook/withProvider.tsx
```diff
@@ -38,11 +39,13 @@ export const withProviders = (getStory) => (
           <MemoryRouter>
             <IntlProvider messages={messages} locale={"en"}>
               <ThemeProvider theme={theme}>
-                <ConfigServiceProvider defaultConfig={defaultConfig} providers={[]}>
+                <ConfigServiceProvider config={config}>
```
timroes, line 42: :information_source: This is kinda "funny" that this worked beforehand. So it seems that Storybook never did typecheck those files, until Storybook 7 (or use using a `main.ts` instead of `main.js`).

airbyte-webapp/package.json
```diff
@@ -154,7 +152,7 @@
     "orval": "^6.11.0-alpha.10",
     "prettier": "^2.6.2",
     "react-select-event": "^5.5.0",
-    "storybook-addon-mock": "^2.4.1",
```
timroes, line 157: :information_source: This isn't compatible with Storybook 7 anymore. I've checked the only place it was used `ConnectorForm` and it seems this endpoint is no longer called during storybook (I assume this was resolved due to some refactoring we did in this form), so we no longer need to mock any data in storybook.

airbyte-webapp/src/index.tsx
```diff
@@ -7,7 +7,7 @@ import { loadDatadog } from "utils/datadog";
 import { loadOsano } from "utils/dataPrivacy";
 import { loadSentry } from "utils/sentry";
 
-import "./globals";
+import "./dayjs-setup";
```
timroes, line 10: :information_source: This is required, due to some missmatch in `require` paths, we can't have this file named be `globals` in `src` directly :shrug: No one knows exactly why the pathes are missmatched, but this is the easiest workaround for now.

airbyte-webapp/src/views/Connector/ConnectorForm/index.stories.tsx
```diff
@@ -42,10 +32,9 @@ export default {
     onSubmit: async (v) => console.log(v),
     availableServices: [TempConnector],
   },
-  decorators: [withMock],
 } as ComponentMeta<typeof ConnectorForm>;
 
-const Template: ComponentStory<typeof ConnectorForm> = (args) => {
+const Template: ComponentStory<typeof ConnectorForm> = (args: React.ComponentProps<typeof ConnectorForm>) => {
```
timroes, line 37: :information_source: Given those files are typechecked now, this is now needed.

.github/workflows/gradle.yml
```diff
@@ -431,6 +431,8 @@ jobs:
 
       - name: Build :airbyte-webapp
         uses: Wandalen/wretry.action@master
+        env:
```
timroes, line 434: :information_source: Got some CI failure on out heap out of space, so seems Storybook 7 might need a bit more memory than we had so far.

airbyte-webapp/.storybook/withProvider.tsx
```diff
@@ -38,11 +39,13 @@ export const withProviders = (getStory) => (
...
                   <DocumentationPanelProvider>
-                    <FeatureService features={[]}>
-                      {getStory()}
-                    </FeatureService>
+                    <AppMonitoringServiceProvider>
```
timroes, line 44: :information_source: This broke the `ConnectorForm` which I think didn't work on `master` anymore.

airbyte-webapp/build.gradle
```diff
@@ -71,6 +71,10 @@ task buildStorybook(type: NpmTask) {
     inputs.dir 'src'
 
     outputs.dir 'build/storybook'
+
+    environment = [
+        'NODE_OPTIONS': '--max_old_space_size=4096'
```
timroes, line 76: :information_source: It seems storybooks compile process with Vite requires more memory. I saw CI failing with an heap out of memory error on this, so giving it more space here.

edmundito (approved):
Changes look good. Tested locally and in chromatic

.github/workflows/gradle.yml
```diff
@@ -397,8 +397,6 @@ jobs:
     steps:
       - name: Checkout Airbyte
         uses: actions/checkout@v3
-        with:
```
timroes, line 400: :information_source: I've talked to some Chromatic people, who confirmed that this should only be needed if we do screenshot comparison, which we're not using in chromatic.

PR #22061, opened 2023-01-30
🪟🎉 Connector builder: Improve testing panel UX on invalid configuration

### LINKED ISSUE #20948: Connector builder: Lag between local streams and API stream listing not handled gracefully
The left hand side stream listing is based on the local state of the form, but the dropdown on the right side is powered by the response from the stream listing API interpolating values and so on. This difference is often invisible, but there are two cases where it's showing:
* Adding a new stream - the stream test panel "switches" to the new stream, but the API didn't reflect it back yet (gets fixed after the response returns from the server):

https://user-images.githubusercontent.com/1508364/210238508-0194c687-e5c5-446e-bb3e-a058c95c0b46.mp4

* Copying a stream with errors (will not get fixed without user input)

https://user-images.githubusercontent.com/1508364/210238574-badef04d-aa71-4464-afad-2f880c7c5368.mp4

There are a few ways how to fix this, I think we should do the following:
* Make the request for the stream listing async with an explicit loading state
* Base the dropdown on the local state so it's always in sync and the user can always switch
* For the test data and the url, look up the state from the response from the listing API - if not available yet but in progress, render "loading..." or something like this for the URL of the stream
* If the URL can't be shown because the manifest is out of sync because the form values are invalid, show "Please fix form values..." or something  for the URL of the stream

flash1293:
## What

Fixes #20948 when working on that, I noticed a few other small issues that bothered me before, I tried to stabilize that as well.

Fixes a bunch of small inconsistencies in the testing panel:
* Field would not get touched on blur (`airbyte-webapp/src/components/connectorBuilder/Builder/BuilderField.tsx`) - add back the blur
* Stream selector didn't show new stream (instead renders undefined) if it got added in invalid state (airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/StreamSelector.tsx) - use the form state instead of the response from the listing call if the listing response is not available
* Instead of only rendering the error message on failing listing call, always show the stream selection and test button:
  * Keeps the UI stable
  * Use the UI state if the listing response is not available
  * Always keep the Testing button in view to jump to the first error

The last one is especially important while editing the yaml - in this situation it's expected to jump quickly between error state and valid state and this PR makes sure the UI is not jumping around:

https://user-images.githubusercontent.com/1508364/215772345-3281e478-2fe7-4c0c-b163-1d62ee854e33.mp4

This is how it looked before (multiple jumps):

https://user-images.githubusercontent.com/1508364/215772592-c3531280-8b8d-42e4-afe2-a90f9f04f54f.mp4

## How

* Expose stream list loading state and form values invalid state from the context so the testing panel can react to it
* Use the stream list from the form values if available and mix in names from the listing response
* Move the listing error message into the stream tester component so it's rendered right below the testing button

There is still a small delay before a newly created stream is shown in the testing panel - this is due to the debounce of moving the form values up to the builder state. I can't think of a good way to get rid of this at the moment, I hope that after the migration to react-hook-form we won't need this debounce anymore.


lmossman (approved):
Changes LGTM, tested locally and it works as expected. This UX is a lot better!

PR #22069, opened 2023-01-30
[docs] update CDK Tutorial: Python HTTP

maxi297:
## What
The Exchange Rate API has changed a little bit. To make sure our tutorial is up-to-date and crystal, I've updated it to:
* Take into account the new URL
* Update the authentication method
* Ensure we use `self.base` properly
* Align `docs/` and `airbyte-cdk/python/docs/`


sergio-ropero (commented):
Awesome job! Thanks for updating it ^^
I added minor changes, not so important to be honest.

sergio-ropero (approved):

PR #22072, opened 2023-01-30
🪟 🧹 Clean up NotificationService imports

edmundito:
## What
Simplifies the export/imports for NotificationService

## How
Ensures that there is only one way to import the NotificationService and removes the default export.


timroes (approved):
Code LGTM, have not tested locally

PR #22081, opened 2023-01-30
🪟 🔧  [Connector Builder] add back middlewares and request correct url

lmossman:
## What
Fix the URL that is being requested for the connector builder server to be on the same domain, so that basic auth works on prod.

Also add back in middlewares since they will be needed when we add this to cloud

This was accidentally removed during the config provider refactor: https://github.com/airbytehq/airbyte/pull/21456/files#diff-3bf74e0deb63a2c070f4f8935d1bb61ddd16f0bb4e1b47e6bd8af55d85f24650L14


timroes (approved):
Code LGTM, did not test locally (cc @josephkmh we forgot to read the `connectorBuilderApiUrl` from `window`, which is required for OSS to work with the config provider refactoring).

PR #22087, opened 2023-01-30
🪟 🔧 Only retry end2end test using cypress not workflows

timroes:
## What

Use only Cypress logic to retry tests, since retryign the whole workflow causes more problems and is rather slow.


ambirdsall (approved):
Looks great, thanks for writing this so quickly.

PR #22107, opened 2023-01-30
🪟 🐛 Fix issue with opener window closed

timroes:
## What

This fixes a weird edge-casey issue (that I've seen some users run into), when starting an OAuth flow for a connector and then closing the original Airbyte tab, but keep the OAuth flow popup open and finish it. Currently it will bring you back to the webapp after the flow, and we'll try to call `window.opener.postMessage`, but `window.opener` (the original tab) is now closed so we'll crash the popup window also with a "Ooops something went wrong" screen.

Changing this, so in case the original tab was closed and you still finish the OAuth flow, we'll just close this, and basically ignore it, since the user aborted setting up the connector by closing that tab.

### DIFF
--- a/airbyte-webapp/src/hooks/services/useConnectorAuth.tsx
+++ b/airbyte-webapp/src/hooks/services/useConnectorAuth.tsx
@@ -210,7 +210,7 @@ export function useResolveNavigate(): void {
   const query = useQuery();
 
   useEffectOnce(() => {
-    window.opener.postMessage(query);
+    window.opener?.postMessage(query);
     window.close();
   });
 }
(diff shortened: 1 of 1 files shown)


lmossman (approved):
Changes LGTM, tested locally and it works as expected.

Though, I think it would be ideal if closing the original Airbyte tab would just cause the OAuth popup to automatically close as well. Is that approach also simple to achieve (I assume not)? If so I would vote to go with that behavior; if not then I think what we have here is fine too.

timroes:
@lmossman I agree this is indeed better. I added this to the PR as well. I'd still leave the other code for safty reasons, to not crash in case that still happens (since closing event handler can be a bit tricky and potentially not called by the browser).

PR #22118, opened 2023-01-30
Handle logging non-JSON-serializable classes in stream slices

erohmensing:
## What
*It helps to add screenshots if it affects the frontend.*
Avoid issues like [these](https://github.com/airbytehq/airbyte/actions/runs/4028599066/jobs/6925677028#step:11:1805) when trying to log stream slices which contain values that are not serializable. Note that in the case of the linked action, the `streamslice` class is not a `Mapping`, so it's not the most valid case here. However, when I implemented that as a TypedDict instead, I ran into the same issue for `datetime` formats, and I think they're a valid case we should support, hence that being the test case. 

## How
*Describe the solution*
When dumping, if a class is not json serializable, dump it as a string.

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
@@ -243,7 +243,8 @@ class AbstractSource(Source, ABC):
             has_slices = True
             if logger.isEnabledFor(logging.DEBUG):
                 yield AirbyteMessage(
-                    type=MessageType.LOG, log=AirbyteLogMessage(level=Level.INFO, message=f"{self.SLICE_LOG_PREFIX}{json.dumps(_slice, default=str)}")
+                    type=MessageType.LOG,
+                    log=AirbyteLogMessage(level=Level.INFO, message=f"{self.SLICE_LOG_PREFIX}{json.dumps(_slice, default=str)}"),
                 )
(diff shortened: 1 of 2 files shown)


maxi297 (approved):
This seems like a good idea. Out of curiosity, are there slices today where the value is not JSON serializable?

maxi297:
Note that you will probably need to run `SUB_BUILD=CONNECTORS_BASE ./gradlew format --scan --info --stacktrace`

erohmensing:
Yes! The run I linked is from a connector which defines streamslices as a StreamSlice dataclass (not a dictionary) with the following definition: 

```
@dataclass
class StreamSlice:
    start_date: DateTime
    end_date: DateTime
```

obviously, this goes against typing to be using a data class instead of a dict or other mapping format - it should still probably be updated to be a TypedDict. However, if you do that, then you still run into the error because `start_date` and `end_date` are `pendulum.datetime.DateTime`s, which are not json serializable.

I didn't notice they came from pendulum, not the `datetime` library itself, but when I wrote the failing test, `datetime.date` also failed for the same reason - TIL! But since `stream_slices` are `Iterable[Optional[Mapping[str, Any]]]`, there could be datetimes or other custom objects in that `Any`.

erohmensing:
And thanks, will format and add changelog info etc. so that I can release it :)

PR #22120, opened 2023-01-31
Low code cdk to beta remove checkpoint interval

### LINKED ISSUE #21239: [Low-Code CDK] Remove checkpoint interval from low-code schema
Allowing the connector developers to configure the checkpoint interval adds complexity for very little value. Better to let the framework decide on when to checkpoint

clnoll:
Closes https://github.com/airbytehq/airbyte/issues/21239.

Allowing the connector developers to configure the checkpoint interval adds complexity for very little value. This PR removes the option from the component schema and will let the framework decide on when to checkpoint.

No manifests were identified with `checkpoint_interval.`


sherifnada (approved):
are there any docs that need to be updated? also we should update the changelog

clnoll:
@sherifnada I've updated the docs.

This is going to be merged into the `low_code_cdk_to_beta` branch and released with the other breaking changes, so the changelog will be updated then.

PR #22122, opened 2023-01-31
🪟 🎉  Connector Builder Landing Page

### LINKED ISSUE #21490: Connector Builder Landing Page
## Context

Implement the landing page for the connector builder, as shown in this design: https://www.figma.com/proto/F1ljvrJerLoVbNvtejBWid/LOW-CODE-CONNECTOR-BUILDER?page-id=3404%3A119487&node-id=3426%3A123290&viewport=-6573%2C-494%2C2&scaling=min-zoom&starting-point-node-id=3426%3A123290

**Note**: For now, the middle tile (Load an existing connector) and its corresponding page will not be added. This will be added once the "managing multiple connectors" project is complete.

## Acceptance Criteria

A landing page for the connector builder exists with two tiles:
- Import a YAML manifest, which opens the builder in the YAML view containing the imported yaml
- Start from scratch, which for now will just open a blank connector builder

Open question: should we always show this page when going to /connector-builder right now? Or should it only be shown if they don't have anything in local storage?
- Grooming notes: Go with the latter option here to get us incrementally there

lmossman:
## What
Resolves https://github.com/airbytehq/airbyte/issues/21490

Adds a landing page for the connector builder, which is shown if the user doesn't have any connector builder changes saved to browser local storage.

Screen Shot 2023-01-31 at 5 49 54 PM

To return to this page after making changes, click the Reset All button in the bottom-left corner of the Builder.

## How
As described in [the issue](https://github.com/airbytehq/airbyte/issues/21490), this PR only adds tiles for importing a YAML and starting from scratch. Start from scratch just opens a blank connector builder UI as normal.

The Import a YAML button only allows users to select a single `.yaml` file from their computer to upload. A few outcomes can occur here:
1. The uploaded file contains invalid YAML. In this case, we just show a Toast with the YAML validation error message and line number, which is auto-dismissed if they click the import button again.
2. The uploaded file contains valid YAML, but it can't be converted to Builder UI. In this case, we open the YAML in the YAML editor view of the connector builder.
3. The uploaded file contains valid YAML and can be converted to Builder UI. In this case, we open the Builder UI with the values filled in from the YAML.
    - In this case, the Connector Name is set to the titleCase version of the file name (minus the extension), e.g. `poke_api.yaml` would result in the connector name set to `Poke Api`


airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/ConfigMenu.tsx
```diff
@@ -62,7 +62,7 @@ export const ConfigMenu: React.FC<ConfigMenuProps> = ({ className, testInputJson
               onClick={() => setIsOpen(true)}
               disabled={
                 !jsonManifest.spec ||
-                Object.keys(jsonManifest.spec.connection_specification.properties || {}).length === 0
+                Object.keys(jsonManifest.spec.connection_specification?.properties || {}).length === 0
```
lmossman, line 65: Small bugfix - the page would crash if the `spec` did not contain `connection_specification`, and this change prevents that crash from occurring

airbyte-webapp/src/components/connectorBuilder/types.ts
```diff
@@ -99,7 +99,7 @@ export interface BuilderStream {
 
 export const DEFAULT_BUILDER_FORM_VALUES: BuilderFormValues = {
   global: {
-    connectorName: "",
+    connectorName: "Untitled",
```
lmossman, line 102: We want this eventually anyway, and this prevents us from showing a `Required` error on Reset All

airbyte-webapp/src/components/connectorBuilder/useManifestToBuilderForm.ts
```diff
@@ -60,7 +61,7 @@ export const convertToBuilderFormValues = async (
   }
   const resolvedManifest = resolveResult.manifest as ConnectorManifest;
 
-  const builderFormValues = DEFAULT_BUILDER_FORM_VALUES;
+  const builderFormValues = cloneDeep(DEFAULT_BUILDER_FORM_VALUES);
```
lmossman, line 64: Fixes a bug where the lines below this one were actually updating the `DEFAULT_BUILDER_FORM_VALUES` constant's fields, causing issues like `Reset All` not behaving properly

airbyte-webapp/src/pages/ConnectorBuilderPage/LandingPage.tsx
```diff
@@ -0,0 +1,194 @@
...
+              const file = uploadEvent.target.files?.[0];
+              const reader = new FileReader();
+              reader.onload = (readerEvent) => {
+                handleYamlUpload(readerEvent.target?.result as string);
+              };
+              if (file) {
+                reader.readAsText(file);
+              }
+            }}
+          />
```
lmossman, line 131: Open to suggestions here, but this seemed like the simplest and most standard way to handle file uploads without using a library. And I couldn't find any libraries that were well supported and lightweight

airbyte-webapp/src/pages/ConnectorBuilderPage/ConnectorBuilderPage.tsx
```diff
@@ -47,6 +50,11 @@ const ConnectorBuilderPageInner: React.FC = React.memo(() => {
         validationSchema={builderFormValidationSchema}
       >
         {(props) => {
+          if (showLandingPage) {
+            return (
+              <LandingPage hideLandingPage={hideLandingPage} switchToUI={switchToUI} switchToYaml={switchToYaml} />
```
flash1293, line 55: As this is part of the regular `ConnectorBuilderPage`, it initializes the form and test state services and tries to do a first fetch which fails with a 400.

I wonder whether we should split out the landing page into its own page with its own path - it seems easier to understand in the code (no "page in a page") and we can better control what side effects are triggered and avoid state leaking from one connector to another in case the user switches (also I think we won't get around that for having multiple builder projects anyway). 

This is what I could imagine as interim state:
* `connector-builder/edit` shows the actual connector builder UI (redirects to `connector-builder` on reset)
* `connector-builder` shows the landing page (for now redirects automatically to `connector-builder/edit` if there is a non-default state in localstorage)

I think we can have a separate service for the landing page wiring up with local storage (and later on with the APIs to save and load connectors).


Doesn't absolutely need to happen on this PR as this is a meaningful increment, but it seems like a better step towards the targeted feature scope - your call how to go with this.

airbyte-webapp/src/pages/ConnectorBuilderPage/LandingPage.tsx
```diff
@@ -0,0 +1,199 @@
...
+          <FormattedMessage id="connectorBuilder.landingPage.prompt" />
+        </Heading>
+        <FlexContainer direction="row" gap="2xl">
+          <input
+            type="file"
+            accept=".yml,.yaml"
+            ref={fileInputRef}
+            style={{ display: "none" }}
+            onChange={(uploadEvent) => {
+              setImportYamlLoading(true);
```
flash1293, line 123: nit: Why not moving all of this into the callback? Isolates a bit better the raw API access

airbyte-webapp/src/pages/ConnectorBuilderPage/LandingPage.tsx
```diff
@@ -0,0 +1,199 @@
...
+        </FlexContainer>
+        <Heading as="h1" size="lg">
+          <FormattedMessage id="connectorBuilder.landingPage.prompt" />
+        </Heading>
+        <FlexContainer direction="row" gap="2xl">
+          <input
+            type="file"
+            accept=".yml,.yaml"
+            ref={fileInputRef}
+            style={{ display: "none" }}
```
flash1293, line 121: nit: the prop `hidden` should work as well here and is slightly more "correct" (not a big deal though)

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.tsx
```diff
@@ -73,6 +73,7 @@ export const BuilderSidebar: React.FC<BuilderSidebarProps> = React.memo(({ class
       onSubmit: () => {
         setValues(DEFAULT_BUILDER_FORM_VALUES);
```
flash1293, line 74: Doesn't need to be fixed on this PR, but it seems like this doesn't work reliably - after hidden "Reset all", then reloading the page I'm getting back to the previous state (there is probably another render-cycle of some kind putting it back). A separate page with a separate context for state management (see comment above) might help with this automatically.

airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderStateService.tsx
```diff
@@ -100,6 +95,11 @@ export const ConnectorBuilderFormStateProvider: React.FC<React.PropsWithChildren
   const manifestRef = useRef(derivedJsonManifest);
   manifestRef.current = derivedJsonManifest;
 
+  const [showLandingPage, setShowLandingPage] = useState(
```
flash1293, line 98: nit: `isEqual` isn't super cheap, would be nice to wrap it into an initializer function

flash1293 (commented):
Looks mostly good besides some nits and one question about the code structure. Works like a charm!

airbyte-webapp/src/pages/ConnectorBuilderPage/ConnectorBuilderPage.tsx
lmossman: I like your suggestions here - I went ahead and added those changes now. I moved the local storage handling to its own ConnectorBuilderLocalStorage service/context, and moved the landing page into its own page / updated the routes and used redirects to go between.

I also found and fixed the issue with Reset All not always working properly - this was because it wasn't updating local storage properly, so I made sure it does that before navigating back to the landing page.

airbyte-webapp/src/pages/ConnectorBuilderPage/LandingPage.tsx
lmossman: I've moved this into the callback. The callback is now pretty long but I agree that keeping the logic together is probably better encapsulation

airbyte-webapp/src/components/connectorBuilder/Builder/BuilderSidebar.tsx
lmossman: See earlier comment, after the refactor and making this properly update local storage, this should now be fixed.

airbyte-webapp/src/services/connectorBuilder/ConnectorBuilderStateService.tsx
lmossman: Moved this into the landing page component since we now only need to do this once on page load. I put it in a useEffect which should only run a single time since I'm using refs for the local storage values

airbyte-webapp/src/pages/connectorBuilder/ConnectorBuilderLandingPage/ConnectorBuilderLandingPage.tsx
```diff
@@ -0,0 +1,214 @@
...
+      <FlexContainer direction="row" gap="2xl">
+        <input type="file" accept=".yml,.yaml" ref={fileInputRef} onChange={handleYamlUpload} hidden />
+        <Tile
+          image={<ImportYamlImage />}
+          title="connectorBuilder.landingPage.importYaml.title"
+          description="connectorBuilder.landingPage.importYaml.description"
+          buttonText="connectorBuilder.landingPage.importYaml.button"
+          buttonProps={{ isLoading: importYamlLoading }}
+          onClick={() => {
+            unregisterNotificationById(YAML_UPLOAD_ERROR_ID);
```
flash1293, line 151: We should always unregister this when navigating away from the page, otherwise the user could try to upload a yaml, give up and go with the "from scratch" option:
Screenshot 2023-02-02 at 15 35 27

Can be done via return value of an effect hook
```

useEffect(() => {
  return () => unregisterNotificationById(YAML_UPLOAD_ERROR_ID);
}, [unregisterNotificationById]);
```

airbyte-webapp/src/pages/connectorBuilder/ConnectorBuilderLandingPage/ConnectorBuilderLandingPage.tsx
```diff
@@ -0,0 +1,214 @@
...
+            convertedFormValues = await convertToBuilderFormValues(json, DEFAULT_BUILDER_FORM_VALUES);
+          } catch (e) {
+            setStoredEditorView("yaml");
+            setStoredManifest(json);
+            navigate(ConnectorBuilderRoutePaths.Edit);
+            return;
+          }
+
+          if (fileName) {
+            convertedFormValues.global.connectorName = startCase(fileName.split(".")[0]);
```
flash1293, line 104: This used to work really well until https://github.com/airbytehq/airbyte/pull/21771 - now it will default to "Manifest" which is not super helpful. I think the next best thing is the url base, but no strong opinion on this.

My suggestion:
* If the file name is not "manifest", use the current logic
* If we can turn into form values, use the base url

airbyte-webapp/src/pages/connectorBuilder/ConnectorBuilderLandingPage/ConnectorBuilderLandingPage.tsx
```diff
@@ -0,0 +1,214 @@
...
+            return;
+          }
+
+          let convertedFormValues;
+          try {
+            convertedFormValues = await convertToBuilderFormValues(json, DEFAULT_BUILDER_FORM_VALUES);
+          } catch (e) {
+            setStoredEditorView("yaml");
+            setStoredManifest(json);
+            navigate(ConnectorBuilderRoutePaths.Edit);
```
flash1293, line 99: Probably out of scope for this PR, but I think it would be very nice to have telemetry for the landing page as well. Wdyt?

flash1293 (approved):
Looks great, just left to nits but LGTM otherwise. No need for another round of review

airbyte-webapp/src/pages/connectorBuilder/ConnectorBuilderLandingPage/ConnectorBuilderLandingPage.tsx
lmossman: Good catch! Fixed

airbyte-webapp/src/pages/connectorBuilder/ConnectorBuilderLandingPage/ConnectorBuilderLandingPage.tsx
lmossman: I will put up a separate PR for that 👍

PR #22135, opened 2023-01-31
🪟 🐛 Enable segment by default

joeykmh:
## What
Sets our config `segment.enabled` setting to `true` by default.

## How
In https://github.com/airbytehq/airbyte/pull/21456 we refactored how config is set, and mistakenly made `segment.enabled` dependent on whether `window.TRACKING_STRATEGY === "segment"`. In previous config, `segment.enabled` was set to `true` by default, so we should keep that behavior.


timroes (approved):
Code LGTM

letiescanciano (approved):
Tested locally.Looks good to me!

PR #22137, opened 2023-01-31
🪟 🔧 Add isCorporate to Airbyte.UI.User.Create

letiescanciano:
We use this event to track conversions across paid channels such as Google Ads.
Since we are only interested in corporate signups, we only want to track a successful conversion in those cases.
Previously, we only added the `isCorporate` flag on `.identify()` but we need this info also in `.track('Airbyte.UI.User.Create')`

Context: https://docs.google.com/document/d/13JbXLgAEVQmMoCBOPdDzAKpzwvqGclMcGSdA3bQ5KYQ/edit?disco=AAAAofb14Pg

### DIFF
--- a/airbyte-webapp/src/packages/cloud/services/auth/AuthService.tsx
+++ b/airbyte-webapp/src/packages/cloud/services/auth/AuthService.tsx
@@ -188,9 +188,8 @@ export const AuthenticationProvider: React.FC<React.PropsWithChildren<unknown>> 
       hasPasswordLogin(): boolean {
         return !!state.providers?.includes("password");
       },
-      hasCorporateEmail(email?: string): boolean {
-        const userEmail = email ?? state.currentUser?.email;
-        return !FREE_EMAIL_SERVICE_PROVIDERS.some((provider) => userEmail?.endsWith(`@${provider}`));
+      hasCorporateEmail(email: string | undefined = state.currentUser?.email): boolean {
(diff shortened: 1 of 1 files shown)


airbyte-webapp/src/packages/cloud/services/auth/AuthService.tsx
```diff
@@ -187,8 +188,9 @@ export const AuthenticationProvider: React.FC<React.PropsWithChildren<unknown>>
       hasPasswordLogin(): boolean {
         return !!state.providers?.includes("password");
       },
-      hasCorporateEmail(): boolean {
-        return !FREE_EMAIL_SERVICE_PROVIDERS.some((provider) => state.currentUser?.email.endsWith(`@${provider}`));
+      hasCorporateEmail(email?: string): boolean {
+        const userEmail = email ?? state.currentUser?.email;
+        return !FREE_EMAIL_SERVICE_PROVIDERS.some((provider) => userEmail?.endsWith(`@${provider}`));
```
timroes, line 193: Code style suggestion: adding this as the default value for the parameter saves us another variable for it.

```suggestion
      hasCorporateEmail(email: string | undefined = state.currentUser?.email): boolean {
        return !FREE_EMAIL_SERVICE_PROVIDERS.some((provider) => email?.endsWith(`@${provider}`));
```

timroes (approved):
Code LGTM, have not tested locally.

PR #22141, opened 2023-01-31
Remove `checkpoint_interval` from source-prestashop manifest

clnoll:
Follow-up from https://github.com/airbytehq/airbyte/pull/22120, removing `checkpoint_interval` from source-prestashop's manifest.


brianjlai (approved):
lgtm! It is interesting to note that while this was set to 500, it's not immediately clear based on the python CDK implementation, why this was added for `source-prestashop`

PR #22145, opened 2023-01-31
Source Freshdesk: add availability strategy

roman-yermilov-gl:
## What
*Add availability strategy*

### DIFF
--- a/airbyte-config/init/src/main/resources/seed/source_definitions.yaml
+++ b/airbyte-config/init/src/main/resources/seed/source_definitions.yaml
@@ -568,7 +568,7 @@
 - name: Freshdesk
   sourceDefinitionId: ec4b9503-13cb-48ab-a4ab-6ade4be46567
   dockerRepository: airbyte/source-freshdesk
-  dockerImageTag: 3.0.0
+  dockerImageTag: 3.0.2
   documentationUrl: https://docs.airbyte.com/integrations/sources/freshdesk
   icon: freshdesk.svg
   sourceType: api
--- a/airbyte-config/init/src/main/resources/seed/source_definitions.yaml
+++ b/airbyte-config/init/src/main/resources/seed/source_definitions.yaml
@@ -573,6 +573,9 @@
   icon: freshdesk.svg
   sourceType: api
   releaseStage: generally_available
+  allowedHosts:
+    hosts:
+      - "*.freshdesk.com"
 - name: Freshsales
   sourceDefinitionId: eca08d79-7b92-4065-b7f3-79c14836ebe7
   dockerRepository: airbyte/source-freshsales
--- a/airbyte-integrations/connectors/source-freshdesk/source_freshdesk/source.py
+++ b/airbyte-integrations/connectors/source-freshdesk/source_freshdesk/source.py
@@ -3,11 +3,10 @@
 #
 
 import logging
-from typing import Any, List, Mapping, Optional, Tuple
-from urllib.parse import urljoin
+from typing import Any, List, Mapping
 
-import requests
 from airbyte_cdk.sources import AbstractSource
+from airbyte_cdk.sources.declarative.checks import CheckStream
 from airbyte_cdk.sources.streams import Stream
 from requests.auth import HTTPBasicAuth
 from source_freshdesk.streams import (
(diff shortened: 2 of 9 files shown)


lazebnyi (approved):

airbyte-integrations/connectors/source-freshdesk/source_freshdesk/source.py
```diff
@@ -52,53 +51,50 @@ def __init__(self, api_key: str) -> None:
 
 
 class SourceFreshdesk(AbstractSource):
+    @staticmethod
+    def _get_stream_kwargs(config: Mapping[str, Any]) -> dict:
+        return {"authenticator": FreshdeskAuth(config["api_key"]), "config": config}
+
     def check_connection(self, logger: logging.Logger, config: Mapping[str, Any]) -> Tuple[bool, Optional[Any]]:
```
erohmensing, line 58: For this, it looks like we want to use `CheckStream` - it uses AvailabilityStrategy under the hood, but has the benefit of making sure the stream you're checking is in the catalog (you don't need to instantiate it, just provide the name) and it will handle the try/excepting. I think the following should about do it: 
```
def check_connection(self, logger: logging.Logger, config: Mapping[str, Any]) -> Tuple[bool, Optional[Any]]:
    check_stream = CheckStream(stream_names=["settings"])
    return check_stream.check_connection(self, logger, config)
```

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
```diff
@@ -565,6 +565,9 @@
   icon: freshdesk.svg
   sourceType: api
   releaseStage: generally_available
+  allowedHosts:
+    hosts:
+      - "${domain}"
```
erohmensing, line 570: Please add an explicit list of allowed hosts

airbyte-integrations/connectors/source-freshdesk/integration_tests/expected_records.jsonl
```diff
@@ -1,132 +1,47 @@
...
+{"stream": "agents", "data": {"available": false, "occasional": true, "id": 67021120885, "ticket_scope": 1, "created_at": "2021-02-11T22:21:43Z", "updated_at": "2021-12-01T00:09:08Z", "last_active_at": null, "available_since": null, "type": "support_agent", "contact": {"active": false, "email": "test_agent_8@test.com", "job_title": null, "language": "en", "last_login_at": null, "mobile": null, "name": "Test Agent 8", "phone": null, "time_zone": "Magadan", "created_at": "2021-02-11T22:21:43Z", "updated_at": "2021-02-11T22:21:43Z"}, "signature": "<div dir=\"ltr\"><p><br></p>\n</div>"}, "emitted_at": 1675116877912}
```
erohmensing, line 1: It looks like these expected records now have `signature`s that are not null - what are these? Asking because I don't see it in the changelog

erohmensing (changes requested):

erohmensing:
There is also custom handling for 403s in the [should_retry](https://github.com/airbytehq/airbyte/blob/991789976eb6b0da9b997e7a2b52caad319261e3/airbyte-integrations/connectors/source-freshdesk/source_freshdesk/streams.py#L59-L64) method - I think we can remove that now! 

We want to look for these and remove them as we implement AvailabilityStrategy, but I've shortlisted all the ones I noticed [here](https://docs.google.com/document/d/1ghkeIhlfqx7yf6fi3sv3nbpEk7BUpIrVlPqkBmrXR0U/edit) for easier reference

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
erohmensing: Actually, this might be okay - here is the definition of how we have the user provide their domain: 
```
      "domain": {
        "type": "string",
        "description": "Freshdesk domain",
        "title": "Domain",
        "examples": ["myaccount.freshdesk.com"],
        "pattern": "^[a-zA-Z0-9._-]*\\.freshdesk\\.com$"
      },
 ```
 
 So it seems similar to enforcing `*.freshdesk.com`, but restricted to what the user put in.

File: airbyte-integrations/connectors/source-freshdesk/integration_tests/expected_records.jsonl (hunk 3 above)
Comment (roman-yermilov-gl) [reply airbyte-integrations/connectors/source-freshdesk/integration_tests/expected_records.jsonl:1]: Some deleted `agents` streams are also have `signatures`. I don't understand what was the question

File: airbyte-integrations/connectors/source-freshdesk/source_freshdesk/source.py (hunk 1 above)
Comment (roman-yermilov-gl) [reply airbyte-integrations/connectors/source-freshdesk/source_freshdesk/source.py]: Done. I removed `HTTPError` exception from try/except. But decided to leave as is the processing of other exceptions not related to HTTP

File: airbyte-config/init/src/main/resources/seed/source_definitions.yaml (hunk 2 above)
Comment (roman-yermilov-gl) [reply airbyte-config/init/src/main/resources/seed/source_definitions.yaml]: changed to `*.freshdesk.com`. Now I think it more safe then to put whatever a client provide us with

**Comment by roman-yermilov-gl** [comment]
> There is also custom handling for 403s in the [should_retry](https://github.com/airbytehq/airbyte/blob/991789976eb6b0da9b997e7a2b52caad319261e3/airbyte-integrations/connectors/source-freshdesk/source_freshdesk/streams.py#L59-L64) method - I think we can remove that now!
> 
> We want to look for these and remove them as we implement AvailabilityStrategy, but I've shortlisted all the ones I noticed [here](https://docs.google.com/document/d/1ghkeIhlfqx7yf6fi3sv3nbpEk7BUpIrVlPqkBmrXR0U/edit) for easier reference

Removed

File: airbyte-integrations/connectors/source-freshdesk/source_freshdesk/streams.py [hunk 4]
```diff
@@ -48,8 +49,8 @@ def url_base(self) -> str:
         return parse.urljoin(f"https://{self.domain.rstrip('/')}", "/api/v2/")
 
     @property
-    def availability_strategy(self) -> Optional["AvailabilityStrategy"]:
-        return None
+    def availability_strategy(self) -> Optional[AvailabilityStrategy]:
+        return HttpAvailabilityStrategy()
```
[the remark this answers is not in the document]
roman-yermilov-gl: Now I use custom strategy, so we need to keep the property

airbyte-integrations/connectors/source-freshdesk/source_freshdesk/source.py
```diff
@@ -52,53 +51,45 @@ def __init__(self, api_key: str) -> None:
...
-            body = error.response.json()
-            error_msg = f"{body.get('code')}: {body.get('message')}"
+            check_stream = CheckStream(stream_names=["settings"], options={})
+            return check_stream.check_connection(self, logger, config)
         except Exception as error:
-            alive = False
-            error_msg = repr(error)
-
-        return alive, error_msg
+            return False, repr(error)
```
erohmensing, line 63: This (and the `try:`) is not necessary - [CheckStream will handle this part](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/declarative/checks/check_stream.py#L42-L50)

erohmensing (approved):
LGTM - thank you!

airbyte-integrations/connectors/source-freshdesk/source_freshdesk/source.py
roman-yermilov-gl: We have a `test_check_connection_empty_config` unittest that runs `check_connection` method on a `Source`. In this test we pass a wrong config value (config is missing `api_key`) to the Source and we are failing without try/except block

roman-yermilov-gl: It happen right before try/except in cdk method. This is failing row:
https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/declarative/checks/check_stream.py#L32

erohmensing: Got it - looks like something CheckStream should handle, I can look into it in the CDK

erohmensing: Looks like this is specific to the source because we need the `config` to create the authenticator in the `streams` method. So I think it makes sense to keep the try/excepting the way that it is 👍🏻

PR #22146, opened 2023-01-31
[FIX] Change quotes to fix test-command.yml

bnchrch:
## What
The if statement in the qa-engine check is failing as github actions prefers single quotes

https://github.com/airbytehq/airbyte/actions/runs/4055079918
https://github.com/actions/runner/issues/866

## How
This changes to single quotes


erohmensing (approved):
Been there before 😅

PR #22151, opened 2023-01-31
API to complete OAuth and return secret

JonsSpaghetti:
## What
New endpoint which extends the current complete_oauth endpoint to allow returning secret IDs from GSM for token data.

We're doing this work to enable usage of the current oauth flow via the API for the Powered by Airbyte headless use case. This also has the added benefit of opening up the option for the frontend to use these endpoints 

See [this doc](https://docs.google.com/document/d/1bwN-w_w5Fn6ILjTllg5HgCOvH5UFplge9S2DNeMAmRU/edit?pli=1#heading=h.agrxfsjj7zsy) for more information on PbA.
See [this doc](https://docs.google.com/document/d/1l2MfuYxqp-uycF9lVQxcPabhrQRAlmr34WnfM2YkGec/edit#heading=h.4rdqq2yi5sqg) for more information on the flow and what decisions we made.

## How
Use existing secret coordinate saving code to save our tokens and then return the secret ID to be reused at a later time/date.

## 🚨 User Impact 🚨
No user impact/breaking changes.

## Pre-merge Checklist

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>

</details>


airbyte-api/src/main/openapi/config.yaml
```diff
@@ -1723,6 +1723,31 @@ paths:
...
+                $ref: "#/components/schemas/SecretId"
+          description: Successful operation
+        "404":
+          $ref: "#/components/responses/NotFoundResponse"
+        "422":
+          $ref: "#/components/responses/InvalidInputResponse"
+      operationId: completeSourceOAuthReturnSecret
+      summary: >-
+        Given a source def ID generate an access/refresh token and save as a secret. Return the secret
+        ID.
```
evantahler, line 1750: Some destinations have oAuth as well

evantahler (commented):
At an architectural level, I think having an API for this is 👍.  

However, you should know that the connectors org is going to try very hard to remove the whole directory of [oAuth flows](https://github.com/airbytehq/airbyte/tree/master/airbyte-oauth/src/main/java/io/airbyte/oauth/flows) in Q2.  The oAuth information currently in the platform belongs in the connector to make deploying and versioning this much easier.

I'm pretty confident that an API like this will continue to work, and we can pre-cache the connector information like we do for connector SPECs to keep this kind of thing fast, but that's all TBD (there's a brainstorm on this topic next week!)

airbyte-api/src/main/openapi/config.yaml
JonsSpaghetti: Yup, we recognize that there is destination oauth as well but that doesn't serve the primary use case we need this endpoint for (Powered by Airbyte headless flow) so this is the one we what's to get in for q1. I think at some point we do plan to have a similar endpoint but for destinations.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/SecretsRepositoryWriter.java
```diff
@@ -280,4 +273,21 @@ private Optional<StandardWorkspace> getWorkspaceIfExists(final UUID workspaceId,
     }
   }
 
+  /**
+   * No frills, given a coordinate, just store the payload
+   */
+  public SecretCoordinate storeOAuthSecret(final UUID workspaceId, final String payload) {
```
sherifnada, line 279: Ideally this class, which is a generic utility for writing secrets used by all of the platform (whereas OAuth feels more like a specific application feature that shouldn't proliferate across horizontal utilities), doesn't have any special conception or knowledge of OAuth. Could we instead inject the coordinate base prefix from the calling context, since that does have knowledge of OAuth?

airbyte-api/src/main/openapi/config.yaml
```diff
@@ -1723,6 +1723,31 @@ paths:
           $ref: "#/components/responses/NotFoundResponse"
         "422":
           $ref: "#/components/responses/InvalidInputResponse"
+  /v1/source_oauths/complete_oauth_return_secret:
```
sherifnada, line 1726: just curious, why is this a new endpoint instead of a flag on the `complete_oauth` endpoint?

sherifnada (commented):
@JonsSpaghetti can you help me understand how the consumer of this method will use the secretID returned? I wasn't able to follow it from [this doc](https://docs.google.com/document/d/1l2MfuYxqp-uycF9lVQxcPabhrQRAlmr34WnfM2YkGec/edit#)

airbyte-api/src/main/openapi/config.yaml
JonsSpaghetti: I don't have a good reason, happy to turn this into a flag - will work on that.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/SecretsRepositoryWriter.java
JonsSpaghetti: Definitely we could. Will move that to the calling context.

JonsSpaghetti:
> @JonsSpaghetti can you help me understand how the consumer of this method will use the secretID returned? I wasn't able to follow it from [this doc](https://docs.google.com/document/d/1l2MfuYxqp-uycF9lVQxcPabhrQRAlmr34WnfM2YkGec/edit#)

The secret ID returned will be used for source/destination creation (initially just source creation). The intent is to provide the caller something that they can use to tell us what secrets to use for creating the source without us actually passing raw access/refresh tokens around.

airbyte-api/src/main/openapi/config.yaml
```diff
@@ -4734,6 +4734,9 @@ components:
           additionalProperties: true # Oauth parameters like code, state, etc.. will be different per API so we don't specify them in advance
         oAuthInputConfiguration:
           $ref: "#/components/schemas/OAuthInputConfiguration"
+        returnSecretCoordinate:
```
sherifnada, line 4737: would recommend setting the default to false here to avoid impacting existing flows

sherifnada: we will probably also need to update the return type signature

sherifnada (commented):
**!! Conditional approval !!**

LGTM, thanks for making the changes!

Approving conditional on two things: 
1. would a source created using a coordinate instead of an actual refresh token value work correctly? or would we "double nest" the secret in GSM by accident? If the latter please fix that as part of this PR, otherwise it will break this endpoint for consumers who use it in this way. 
2. Please make sure someone from the platform team approves. I don't own this part of the codebase -- just leaving a review for context.

airbyte-api/src/main/openapi/config.yaml
JonsSpaghetti: Good point, default updated.

I believe updating the return type signature isn't necessary because we return a Map<String, Object> and we return the same when returning a secret ID. The API spec reflect this (since the return of OAuth can be varied, it's just a generic Map).

W/ regards to the first conditional, nobody is using coordinates instead of refresh tokens currently during source creation (as it's not a supported flow) and so there should be no consumers who would expect to be able to do this. If someone used the new flag to get a secret ID through completing OAuth and then attempted to post it to connections in an unsupported way they should fail validation, but there's nothing that would prevent them from giving us a secret ID instead of access/refresh tokens if they really wanted to. However, that generally amounts to them just passing us an invalid access token/refresh token and they shouldn't expect the source to get set up correctly anyway. In current state there should be no nesting of secrets incorrectly.

This  endpoint is just a building block of a new flow in which we complete OAuth, receive a secret ID, and later use that secret ID during source creation instead of passing the raw tokens. The usage of a secret ID during source creation instead of credentials of any kind is not yet developed, but will be shortly.

JonsSpaghetti:
@davinchia I've heard through the grapevine that you're the best person to review this from the platform side. Any chance I can get you to take a look?

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
```diff
@@ -361,4 +362,27 @@ JsonNode getOAuthInputConfiguration(final JsonNode hydratedSourceConnectionConfi
     return Jsons.jsonNode(result);
   }
 
+  public Map<String, Object> writeOAuthSecret(final UUID workspaceId, final Map<String, Object> payload) {
+
+    try {
+      final String payloadString = Jackson.getObjectMapper().writeValueAsString(payload);
+      final SecretCoordinate secretCoordinate = secretsRepositoryWriter.storeSecret(generateOAuthSecretCoordinate(workspaceId), payloadString);
+      return formatOAuthSecretResponse(secretCoordinate);
```
pedroslopez, line 370: Two questions:

1. It seems like we're writing the full OAuth response, stringified, in one coordinate. On the usual creation flow, each secret gets its own coordinate. Should we follow the same approach to make things easier when it comes to creating the source with these credentials? Note that some sources return multiple secret (and non-secret) values as part of the complete_oauth response. A good example is any source with single-use refresh tokens, like GitLab, which returns access_token (secret), refresh_token (secret) and a token_expiry_date (not secret)
2. Do we want to strip out any masked instance-wide params? (I believe they would be coming through here as asterisks (`******`) )

pedroslopez (commented):
Nice! This is also a big part of https://github.com/airbytehq/airbyte-platform-internal/issues/3781

Generally looks good, but some questions on how we're storing the secrets

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
JonsSpaghetti: Might make sense to have a quick chat with you about this since I think it might be pretty illuminating for me, but generically I think part of the answer to both of your points above is to potentially port some of the frontend code to backend.

It seems like frontend has the connector specification handy and when receiving the OAuth response, uses the specification to determine what properties from the response it needs to hold onto/use in source creation.

My inclination is to snag the spec on the backend, follow the same logic that the frontend follows and save the necessary OAuth params as a single secret.

When we create the source, we would fetch the OAuth params via the secret coordinate and then re-save in the way you describe. In our discussions of this Evan (I think?) thought that it was perfectly fine for us to create new secrets as part of source/destination creation and not use the secrets created by this process to directly create sources/destinations.

Does the above seem reasonable?

JonsSpaghetti: Chatted w/ @pedroslopez about this.  Pedro - please add anything I missed!

Where we landed was:
1. As soon as reasonably possible, we should consider adding a check operation to this process to ensure people can't create/update sources/destinations with configurations that validate but can't actually be used to connect to the source/destination. As such, we might benefit from bundling check into source create/update at least externally.
    - This has the added benefit of potentially allowing us to handle the case where single use refresh tokens are consumed by check and then need to be updated in the config prior to source creation since doing both in a single operation will allow us to make the necessary updates prior to saving the source config to the db.
2. When we write source creation we'll need to make sure we separate that out from the connector config and that we port the frontend logic for complete_oauth response -> discrete fields in source config to the backend so that we don't have to update source specifications
3. As part of the above, in the future frontend will want to be able to run check operations and can't do that unless we have the ability for secret ID -> complete_oauth response -> discrete fields in source config to happen prior to the check. We need to figure out how we'll support that on the frontend.

pedroslopez: Generally +1 to the message, just adding that the logic for  "given this secret cordinate with all my oauth params, hydrate the configuration with the values in the right spots" should be re-usable since we will need this in create, update, and the various check_connection implementations (there is also a check_connection_for_update)

davinchia: Drive-by:

What does `consider adding a check operation to this process` mean? What is this check process and where does it live?

JonsSpaghetti: The same check operation that we run prior to creating a source, I guess what pedro calls `check_connection`. Right now as we were thinking about allowing creation of a source, we weren't planning on forcing a check and were initially considering just allowing JSON schema validation to be sufficient, however it seems like that is not in fact sufficient and we should also be running a `check_connection` to help prevent consumers from creating a source with valid spec which can't actually connect.

Pedro also made me aware that we don't run another check unless we get a failed sync which might be confusing/less than ideal user experience for these consumers which is another reason to get the check to happen prior to creating the source and potentially allowing syncs which would fail.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
```diff
@@ -90,24 +99,20 @@ public OAuthConsentRead getSourceOAuthConsent(final SourceOauthConsentRequest so
       if (sourceOauthConsentRequest.getSourceId() == null) {
         oAuthInputConfigurationForConsent = sourceOauthConsentRequest.getoAuthInputConfiguration();
       } else {
-        final SourceConnection hydratedSourceConnection =
-            secretsRepositoryReader.getSourceConnectionWithSecrets(sourceOauthConsentRequest.getSourceId());
+        final SourceConnection hydratedSourceConnection = secretsRepositoryReader
```
davinchia, line 102: aside: this code is due for a rewrite to get rid of all the nested blocks.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
davinchia: makes sense. Did the discussion cover how to integrate check_connection into the server?

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/OAuthHandlerTest.java
```diff
@@ -48,7 +50,8 @@ public void init() {
     trackingClient = mock(TrackingClient.class);
     httpClient = Mockito.mock(HttpClient.class);
     secretsRepositoryReader = mock(SecretsRepositoryReader.class);
-    handler = new OAuthHandler(configRepository, httpClient, trackingClient, secretsRepositoryReader);
+    secretsRepositoryWriter = mock(SecretsRepositoryWriter.class);
+    handler = new OAuthHandler(configRepository, httpClient, trackingClient, secretsRepositoryReader, secretsRepositoryWriter);
```
davinchia, line 54: are we going to add some test cases around this change?

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
JonsSpaghetti: Loosely from my first post:
> As such, we might benefit from bundling check into source create/update at least externally (for public API consumers).
>    
>    * This has the added benefit of potentially allowing us to handle the internal use case where single use refresh tokens are consumed by check and then need to be updated in the config prior to source creation since doing both in a single operation will allow us to make the necessary updates prior to saving the source config to the db.
(edited for clarity)
I think there's probably more complexity around it than that, but if that's an approach that makes sense then I can at least start thinking in that direction.

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/OAuthHandlerTest.java
JonsSpaghetti: Yeah planning to add some as soon as we solidify on what else needs to be added to the PR.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
davinchia: Ok. Writing the entire response to secret storage seems fine for now. 

Unless, we are going to redo the FE/normal connector flows, I suggest hosting all the checking logic within the public-api server for practical purposes and folding this back into the platform once we know what PBA users like/expect.

Can we leave a javadoc on `writeOAuthSecret` summarising ^? What about renaming from `writeOAuthSecret` to `writeOAuthResponseSecret` to better capture what this is actually doing?

davinchia (commented):
Thanks @JonsSpaghetti. General approach seems ok to me.

I would want to look at tests before approving.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
pedroslopez: @davinchia we do want to do this relatively soon for the regular FE flow - see https://github.com/airbytehq/airbyte-platform-internal/issues/3781

JonsSpaghetti: I'm definitely +1 on renaming and writing a javadoc to detail what's going on here. Also up for discussing where we can host the logic based on plans for the FE flow and what people generally feel makes the most sense.

davinchia: good to know @pedroslopez . do you also know who is driving that?

pedroslopez: Perhaps @timroes would have some thoughts on how this can be re-used on the frontend, specially with how we would be storing things in a single blob and would not be "expanding it" frontend-side. From chatting w/ Jon and the team's envisioning the create to just pass in the secret coordinate, it seems like it would be simpler on the FE side since it no longer has to inject the oauth pieces to the right spots in the config, but worth tagging for awareness.

JonsSpaghetti:
@davinchia requested your re-review. After doing manual testing I feel pretty good that this is going to work for us. I added a very basic unit test to ensure that the old behavior is always the default and that we don't call the new code if the flag isn't included. I feel slightly more comfortable with this given that I don't think I changed any logic in the old flow and really only added a completely separate conditional branch to the endpoint handler.

Longer term I think we envision creating a mock authorization server so that we can do a full acceptance test which isn't feasible right now because once we need to talk to the authorization server, we can't mock interaction with the HttpClient and the auth server (rightfully) rejects us since we're using garbage credentials for the test.

If you're ok with that plus the manual testing then this is hopefully good to go, but will wait for you to 👀

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
```diff
@@ -294,14 +289,14 @@ public void setDestinationInstancewideOauthParams(final SetInstancewideDestinati
   private JsonNode getOAuthInputConfigurationForConsent(final ConnectorSpecification spec,
                                                         final JsonNode hydratedSourceConnectionConfiguration,
                                                         final JsonNode oAuthInputConfiguration) {
-    final Map<String, String> fieldsToGet =
-        buildJsonPathFromOAuthFlowInitParameters(OAuthPathExtractor.extractOauthConfigurationPaths(
+    final Map<String, String> fieldsToGet = buildJsonPathFromOAuthFlowInitParameters(
```
davinchia, line 292: there are formatting changes in this file that suggest to me you have a different line width/format set up in your Intellij. It should be set to 150 and using the style guide as noted in our public docs. Can you check that?

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
```diff
@@ -216,48 +214,45 @@ public Map<String, Object> completeSourceOAuth(final CompleteSourceOauthRequest
...
-          oAuthInputConfigurationForConsent,
-          spec.getAdvancedAuth().getOauthConfigSpecification());
+          completeDestinationOAuthRequest.getQueryParams(), completeDestinationOAuthRequest.getRedirectUrl(),
+          oAuthInputConfigurationForConsent, spec.getAdvancedAuth().getOauthConfigSpecification());
     } else {
-      // deprecated but this path is kept for connectors that don't define OAuth Spec yet
-      result = oAuthFlowImplementation.completeDestinationOAuth(
-          completeDestinationOAuthRequest.getWorkspaceId(),
+      // deprecated but this path is kept for connectors that don't define OAuth Spec
+      // yet
```
davinchia, line 252: for example, this shouldn't be moving

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
```diff
@@ -294,14 +289,14 @@ public void setDestinationInstancewideOauthParams(final SetInstancewideDestinati
...
                                                         final JsonNode hydratedSourceConnectionConfiguration,
                                                         final JsonNode oAuthInputConfiguration) {
-    final Map<String, String> fieldsToGet =
-        buildJsonPathFromOAuthFlowInitParameters(OAuthPathExtractor.extractOauthConfigurationPaths(
+    final Map<String, String> fieldsToGet = buildJsonPathFromOAuthFlowInitParameters(
+        OAuthPathExtractor.extractOauthConfigurationPaths(
             spec.getAdvancedAuth().getOauthConfigSpecification().getOauthUserInputFromConnectorConfigSpecification()));
 
-    final JsonNode oAuthInputConfigurationFromDB = getOAuthInputConfiguration(hydratedSourceConnectionConfiguration, fieldsToGet);
+    final JsonNode oAuthInputConfigurationFromDB = getOAuthInputConfiguration(hydratedSourceConnectionConfiguration,
```
davinchia, line 296: or this too

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
davinchia: somewhat distracting from the main changeset, which is lines 364 onwards

JonsSpaghetti: Yeah apologies, I thought I had that set up, will check it now

JonsSpaghetti: You're definitely right, I missed that step in setup. Just updated it.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
```diff
@@ -361,4 +361,51 @@ JsonNode getOAuthInputConfiguration(final JsonNode hydratedSourceConnectionConfi
...
+    try {
+      final String payloadString = Jackson.getObjectMapper().writeValueAsString(payload);
+      final SecretCoordinate secretCoordinate = secretsRepositoryWriter.storeSecret(generateOAuthSecretCoordinate(workspaceId), payloadString);
+      return formatOAuthSecretResponse(secretCoordinate);
+    } catch (final JsonProcessingException e) {
+      throw new RuntimeException("Json object could not be written to string.", e);
+    }
+  }
+
+  public Map<String, Object> formatOAuthSecretResponse(final SecretCoordinate secretCoordinate){
```
davinchia, line 388: - this can be a `private` function since it's not used outside the class.
- also can probably fold this into the calling function as it's one line for now.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
```diff
@@ -361,4 +361,51 @@ JsonNode getOAuthInputConfiguration(final JsonNode hydratedSourceConnectionConfi
...
+
+  /**
+   * Generate our OAuthSecretCoordinate for storage For now we're treating these as always v1.
+   */
+  private SecretCoordinate generateOAuthSecretCoordinate(final UUID workspaceId) {
+    final String coordinateBase = SecretsHelpers.getCoordinatorBase("airbyte_oauth_workspace_", workspaceId, UUID::randomUUID);
+    return new SecretCoordinate(coordinateBase, 1);
+  }
+
+  public Map<String, Object> completeSourceOAuthHandleReturnSecret(final CompleteSourceOauthRequest completeSourceOauthRequest)
```
davinchia, line 400: I see we are only doing this for the source - what about the destination? will we do that in a follow up PR? Or do we not need this for the destination for now?

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
```diff
@@ -168,13 +171,13 @@ public OAuthConsentRead getDestinationOAuthConsent(final DestinationOauthConsent
 
   public Map<String, Object> completeSourceOAuth(final CompleteSourceOauthRequest completeSourceOauthRequest)
```
davinchia, line 172: i think this signature can now change to package-protected with the @VisibleForTesting annotation to indicate this is should not be used outside this class except for tests.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
davinchia: nit: I would move this method to be before the `completeSourceOAuth` method on line 172. In general, I try to place the methods so a reader scrolling from top to bottom will encounter methods in order of their importance as they go about reading the class.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
```diff
@@ -361,4 +361,51 @@ JsonNode getOAuthInputConfiguration(final JsonNode hydratedSourceConnectionConfi
...
+      throw new RuntimeException("Json object could not be written to string.", e);
+    }
+  }
+
+  public Map<String, Object> formatOAuthSecretResponse(final SecretCoordinate secretCoordinate){
+    return Map.of("secretId", secretCoordinate.getFullCoordinate());
+  }
+
+  /**
+   * Generate our OAuthSecretCoordinate for storage For now we're treating these as always v1.
```
davinchia, line 393:
```suggestion
   * Generate OAuthSecretCoordinates. Always assume V1 and do not support secret updates.
```

davinchia: I think that's what this comment means.

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/OAuthHandlerTest.java
```diff
@@ -172,15 +182,15 @@ void testBuildJsonPathFromOAuthFlowInitParameters() {
...
-        {
-          "field1": "1",
-          "field2": "2",
-          "field3": {
-            "field3_1": "3_1",
-            "field3_2": "3_2"
-          }
-        }
-        """);
+            {
```
davinchia, line 185: would also double check if this is related to the formatting bits.

davinchia (commented):
Getting there @JonsSpaghetti !

Couple more comments/questions for understanding/readability. 

One major question about whether we also want this for destinations, or only for sources.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
JonsSpaghetti: This makes sense, folded it back into the calling function.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
JonsSpaghetti: Don't need this for destination for now (at least for PbA use case) but PE is planning to add it for destinations in the future.

Appreciate the comment, will move the method and will keep that in mind as it's a good rule to follow.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
JonsSpaghetti: Makes sense, updated.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
JonsSpaghetti: Way better way to phrase, updated.

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/OAuthHandlerTest.java
```diff
@@ -204,32 +214,65 @@ void testGetOAuthInputConfiguration() {
...
+    verify(handlerSpy, never()).writeOAuthResponseSecret(any(), any());
+
+    completeSourceOauthRequest.returnSecretCoordinate(true);
+
+    handlerSpy.completeSourceOAuthHandleReturnSecret(completeSourceOauthRequest);
+
+    verify(handlerSpy, times(2)).completeSourceOAuth(completeSourceOauthRequest);
+    verify(handlerSpy).writeOAuthResponseSecret(any(), any());
+
+    completeSourceOauthRequest.returnSecretCoordinate(false);
```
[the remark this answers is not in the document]
JonsSpaghetti: I think this is taken care of as part of the first request - I can add some comments to the test so it's clearer what exactly is being tested. Agree 100% with the concern that we want to be absolutely certain this isn't just going to break the existing flow.

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/OAuthHandlerTest.java
JonsSpaghetti: It was indeed

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
JonsSpaghetti: reformatted things as well to match the original as closely as possible with 1 or 2 exceptions where I thought it made sense.

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/OAuthHandlerTest.java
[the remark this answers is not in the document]
davinchia: great, thanks

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/OAuthHandler.java
```diff
@@ -292,16 +311,15 @@ public void setDestinationInstancewideOauthParams(final SetInstancewideDestinati
   }
 
   private JsonNode getOAuthInputConfigurationForConsent(final ConnectorSpecification spec,
-                                                        final JsonNode hydratedSourceConnectionConfiguration,
-                                                        final JsonNode oAuthInputConfiguration) {
+      final JsonNode hydratedSourceConnectionConfiguration,
+      final JsonNode oAuthInputConfiguration) {
```
davinchia, line 315: nit: i think the previous format was correct

davinchia (approved):
Looks great! :shipit: @JonsSpaghetti !

jackonawalk:
I'm not sure if this is related, but it looks like it still isn't possible to create a source w/o credentials when a secretId is provided. This effectively breaks oauth for headless powered by airbyte use cases.
See: https://github.com/airbytehq/airbyte/discussions/48607

PR #22153, opened 2023-01-31
copy publishing task from root project

colesnodgrass:
## What
- Kotlin subprojects don't appear to have a way to call a function defined in the root-project
- as a quick-fix, redefine the publishing task in this Kotlin project
- medium-term-fix (not part of this PR); one of the following
    - convert the `getPublishArtifactsTask` dynamic task to work in the same way that the docker-image task works
    - create an `airbyte` gradle plugin that exposes all this functionality via configuration block


jdpgrailsdev (approved):
:shipit:

davinchia (approved):

PR #22154, opened 2023-01-31
🪟 🐛 Fix list connections call

joeykmh:
## What
Fixes a bug where after running a manual sync, the first request to update the connection list was sending an incorrect payload.


airbyte-webapp/src/hooks/services/useConnectionHook.tsx
```diff
@@ -84,7 +85,7 @@ export const useSyncConnection = () => {
   const queryClient = useQueryClient();
   const analyticsService = useAnalyticsService();
   const notificationService = useNotificationService();
-  const workspaceId = useCurrentWorkspace();
+  const workspaceId = useCurrentWorkspaceId();
```
josephkmh, line 88: This was the main mistake - we should be using only the workspace ID here.

edmundito (approved):

PR #22165, opened 2023-01-31
Normalization: handle non-object top-level schemas; treat binary data as string

edgao:
old schema
```json
{type: [object, string], properties: {...}}
```

new schema
```json
{oneOf: [{type: object, properties: {...}}, {ref: string}]}
```

normalization doesn't like this.

---

also, treat binary data as string for now - this is failing in a mysql->bigquery and mssql->snowflake sync.


airbyte-integrations/bases/base-normalization/normalization/transform_catalog/catalog_processor.py
```diff
@@ -135,7 +135,15 @@ def build_stream_processor(
...
 
             message = f"'json_schema'.'properties' are not defined for stream {stream_name}"
-            properties = get_field(get_field(stream_config, "json_schema", message), "properties", message)
+            schema = get_field(stream_config, "json_schema", f"'json_schema' is not defined for stream {stream_name}")
+            if "properties" in schema:
+                properties = get_field(schema, "properties", message)
+            elif "oneOf" in schema:
+                options = filter(lambda option: "properties" in option, schema["oneOf"])
+                if len(options) == 0:
+                    raise KeyError(f"Stream {stream_name} does not have any properties")
```
grishick, line 144: Wouldn't this break for 
`{oneOf: [{type: object, properties: {...}}, {ref: string}]}` because `{ref: string}` does not have `properties` ?

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/catalog_processor.py
```diff
@@ -135,7 +135,15 @@ def build_stream_processor(
...
+                # If there are multiple oneOf options, just pick the first one - we don't really support oneOf to begin with
```
grishick, line 145: Ehm... is there an issue that tracks supporting `oneOf` or disabling `oneOf` for sources? Aren't we always risking failing normalization when a source declares schema with `oneOf`?

edgao: normalization, even prior to protocol v1, was making the assumption that the top-level schema is a plain `{type: object}`. So, yes but I don't want to fix that right now.

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/catalog_processor.py
edgao: in this case, `schema["oneOf"]` would be `[{type: object, properties: {...}}, {ref: string}]`

so the `filter` call would return `[{type: object, properties: {...}}]`

and then `len(options)` would be 1

I think?

gosusnp (approved):
The change about binary looks reasonable to me. About the oneOf, I'd defer to Greg or someone with more experience with json schemas / data types.

edgao:
I'm going to publish this, I think the worst case is no worse than current behavior.

grishick (approved):

octavia-approvington:
What are we doing again? <br/> merge and squash

octavia-approvington (approved):

PR #22176, opened 2023-01-31
Icons passed to Button component must be wrapped in a div for proper styling

### LINKED ISSUE #18581: Icons passed to Button component must be wrapped in a div for proper styling
When providing both an `icon` to the `Button` component, and a label (as a child), the Button component does not always successfully apply the desired class names to the icon inside of the button.

However, if the icon is wrapped in a `<div>` before being passed to the Button, the styles are properly applied.

For example, we are currently doing this div-wrapping hack in the `ConnectionStatusTab` component: https://github.com/airbytehq/airbyte/blob/57068e3f2941698b9c44129b3527b646e5d4703d/airbyte-webapp/src/pages/ConnectionPage/pages/ConnectionItemPage/ConnectionStatusTab.tsx#L178-L182

To demonstrate what this does, here is what the button and its styles look like currently on prod, with the icon wrapped in a div:
Screen Shot 2022-10-27 at 5 43 54 PM

And here is what the button and styles look like when the icon is NOT wrapped in a div:
Screen Shot 2022-10-27 at 5 42 25 PM
As you can see, the `Button_buttonIcon_...`, `Button_positionLeft_...`, and `Button_withLabel_...` styles are not being applied to the `<svg>` element, which messes up the formatting 

This may be happening because we are using React.cloneElement to create the button icon here: https://github.com/airbytehq/airbyte/blob/a6974a7ba93b7f1c62f17b37ca45a0bfcb456061/airbyte-webapp/src/components/ui/Button/Button.tsx#L57-L63
And, it may only happen with specific icon elements like `<svg>`, which may have trouble having those styles applied.

One solution could be to always wrap the cloned element in a `<div>`, and apply the icon styles to that div instead, though this may cause us to end up with unnecessary `<div>` nesting.

m19berger:
## What
Closes [#18581](https://github.com/airbytehq/airbyte/issues/18581)

## How
Added property className for Rotate Icon component to support `class` attribute

image

### DIFF
--- a/airbyte-webapp/src/components/icons/RotateIcon.tsx
+++ b/airbyte-webapp/src/components/icons/RotateIcon.tsx
@@ -4,10 +4,11 @@ interface Props {
   color?: string;
   width?: string;
   height?: string;
+  className?: string;
 }
 
-export const RotateIcon = ({ color = theme.greyColor20, width = "20", height = "20" }: Props) => (
-  <svg width={width} height={height} viewBox="0 0 20 20" fill="none">
+export const RotateIcon = ({ color = theme.greyColor20, width = "20", height = "20", className }: Props) => (
+  <svg className={className} width={width} height={height} viewBox="0 0 20 20" fill="none">
     <path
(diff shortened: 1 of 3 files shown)


matter-q:
As I see it, we're not actually passing a simple `<svg>` tag, we're passing a React component that doesn't have the described `className` notation, so that's why `<div>` wrapper is a working solution. 

What do you think @lmossman , @tealjulia, @edmundito  ?

dizel852:
Hmm.. seems like this issue is part of https://github.com/airbytehq/airbyte/issues/20133.
Since atm, we don't have a consistent `<Icon />` component to handle all the available icons.

dizel852 (approved):
Tested locally - LGTM 👍 

Passing to @edmundito

edmundito (approved):

PR #22188, opened 2023-02-01
Normalization: handle nested objects correctly

edgao:
We're currently not prioritizing objects/arrays correctly in some cases. E.g. v0 schema:
```json
{"type": ["string", "object"], "properties": {}}
```

which becomes v1 schema:
```json
{
  "oneOf": [
    {"$ref": "....String"},
    {"type": "object", "properties": {}}
  ]
}
```

is being interpreted as a string, rather than object.


etsybaev:
Tried to run tests locally, but got lots of fails :( 

/home/ievgeniit/Downloads/logs_normalization

Selection_264

etsybaev:
:airbyte-integrations:bases:base-normalization:mypyCheck task seems to fail

evantahler:
@edgao I'm going though stale PRs assigned to the team - what's the status of this one?

edgao:
this was for protocol v1; closing

PR #22190, opened 2023-02-01
Add a repository method that determines if a connection has an alpha or beta connector

pmossman:
## What
In order to determine whether a particular connection should be allowed to continue running even in a delinquent workspace with negative credits, we need a simple way to check if a connection has an alpha or beta connector. This check should be performant so that we can perform it in API call validations and in our billing cron.

## How
Write a new configRepository method that joins the connection with actor -> actor_definition and uses the definition's release stage to return a boolean.

Cloud will call this new method when validating free connector usage in the API and cron.


mfsiega-airbyte (approved):
LGTM!

PR #22194, opened 2023-02-01
🪟 🔧 Remove utmStorage util

letiescanciano:
This method was only effective when we redirect users directly to cloud.airbyte.io/signup?utm_source..... but wasn't working when users first landed in airbyte.com with UTM params and then end up signing up.

We are managing UTM params with cross domain cookies in the different tools we use to measure attribution (GA and Hubspot) so we don't need this anymore.


timroes (approved):
Code LGTM, have not tested locally.

PR #22198, opened 2023-02-01
🪟 🔧 Switch to the new airbyte-cloud repo name

timroes:
## What

Fixes the path to load our environment variables from to represent the new repository name `airbyte-platform-internal`. We'll need to locally rename this for everyone developing once this is merged.


joeykmh (approved):
Good catch!

PR #22200, opened 2023-02-01
🪟 ⏪ Revert show text overflow on hover PR + remove its usage

dizel852:
## What
Revert https://github.com/airbytehq/airbyte/pull/21253

**Reason:** 
"This introduces a XSS attack, we need to reimplement that in a save way (with DOM node copying instead of innerHTML and dangerouslySetInnerHTML)" © @timroes


edmundito:
As discussed, this is not needed and the fix will be done separately.

dizel852:
Closed as obsolete

PR #22201, opened 2023-02-01
🪟 🔧 Prevent usage of dangerouslySetInnerHTML

timroes:
## What

Adds the `react/no-danger` linting rule to prevent any usage of `dangerouslySetInnerHTML`. The very very few places it makes sense to use this, can ignore the rule in line with a description of why we know the passed in HTML is safe.


joeykmh (approved):

dizel852 (approved):
LGTM 👍

PR #22204, opened 2023-02-01
Source Google-Search-Console: updated expected records

darynaishchenko:
updated expected records: removed fast-changing data


lazebnyi (approved):

darynaishchenko:
fixed in other pr

PR #22214, opened 2023-02-01
🪟 🐛 Fix ResizablePanel component

timroes:
## What

Fixes a bug that the resizable panel component doesn't render anything, if you first set `hideSecondPanel` to `true` and then to `false`. You can currently seeing this on master by clicking in an empty workspace on a source icon in the empty connection page, which will bring you do the setup for that connector, but nothing will render, due to this issue.

## How

`react-reflex` has apparently a problem when the second panel is first not rendered and then rendered and ends up rendering it with 0 width. The only thing to get it reworking was fully rerendering the component (by changing a `key` on the `ReflexContainer`), which worked to fix the issue, but also added a couple of new issues. So after syncing with Teal, we decided to change this to handle the hiding in CSS only, which also fixes the issue and is was less buggy.

We just hide the content of the 2nd panel, so that this component doesn't render unnecessarily.

### DIFF
--- a/airbyte-webapp/src/components/ui/ResizablePanels/ResizablePanels.module.scss
+++ b/airbyte-webapp/src/components/ui/ResizablePanels/ResizablePanels.module.scss
@@ -71,8 +71,17 @@
   height: 3px;
 }
 
+.fullWidth {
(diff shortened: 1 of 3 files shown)


lmossman (approved):
Change LGTM and seems to fix the issue from my local testing

PR #22226, opened 2023-02-01
📖  Reference new build system docs

bnchrch:
## What
We have new documentation for the build reports process.

## How
This adds references to the documentation from each file that composes the system.

### DIFF
--- a/tools/bin/build_report.py
+++ b/tools/bin/build_report.py
@@ -3,7 +3,12 @@
 #
 
 """
+Report Connector Build Status to Slack
+
 All invocations of this script must be run from the Airbyte repository root.
+
+BEFORE RUNNING THIS SCRIPT:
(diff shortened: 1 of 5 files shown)


tools/bin/build_report.py
```diff
@@ -3,8 +3,13 @@
 #
 
 """
+Report Connector Build Status to Slack
+
 All invocations of this script must be run from the Airbyte repository root.
 
+BEFORE RUNNING THIS SCRIPT:
+- Ensure you have read the documentation on how this system works: https://internal-docs.airbyte.io/Generated-Reports/Build-Status-Reports
```
evantahler, line 11: 👍 I think it's fine to reference internal docs in the OSS repo.  This part of the codebase is tooling.

evantahler (approved):

PR #22227, opened 2023-02-01
🪟 🐛 Hot-fix exception in streams table when stream does not have direct properties

edmundito:
## What
Hot-fix an issue in the streams table that crashed the whole page when the properties in the stream were not of the expected shape. Still investigating the root cause and next steps, so this is a temporary fix to prevent it from impacting users.

However, the fields cannot be expanded in the table in streams where the problem occurs. This will be fixed separately.

This is occurring with the Stripe source with `plans` stream in that plans is currently `{ oneOf: [{ properties: ... }, 'WellKnownTypes/String' }]` when the code can only support `{ properties: }` 

## How
Returns an empty object so that the code can take the length correctly. Ensures that the table renders well without the expandable fields.

### DIFF
--- a/airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogTree/CatalogSection.tsx
@@ -59,7 +59,10 @@ const CatalogSectionInner: React.FC<CatalogSectionInnerProps> = ({
     return traversedFields.sort(naturalComparatorBy((field) => field.cleanedName));
   }, [stream?.jsonSchema, stream?.name]);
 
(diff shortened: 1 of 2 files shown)


joeykmh (approved):

timroes:
@edmundito Have we done the investigation for the root cause on this yet? Do we have an issue to track this? (I can't find anything linked here in the PR or the corresponding oncall issue)

josephkmh:
@edmundito @timroes I created an issue here: https://github.com/airbytehq/airbyte/issues/22490

slack thread context: https://airbytehq-team.slack.com/archives/C02U46QK5C3/p1675778896991099?thread_ts=1675276422.305699&cid=C02U46QK5C3

PR #22229, opened 2023-02-01
🪟 🐛 Add cloudPublicApiUrl into webapp env

joeykmh:
## What
During the [config refactor](https://github.com/airbytehq/airbyte/pull/21456) we removed the default `cloudPublicApiUrl` value, but did not add a value to `.env`. This PR adds the value to `.env` so that `REACT_APP_CLOUD_API_URL` is defined during the build step.


terencecho (approved):

terencecho:
I'm curious why we didn't notice any error messages since your previous PR contained a block to throw an error if that var was missing.

josephkmh:
> I'm curious why we didn't notice any error messages since your previous PR contained a block to throw an error if that var was missing.

@terencecho the errors are only seen at run-time in the browser. You can see some of the datadog RUM errors logged here:

https://app.datadoghq.com/rum/explorer?query=%40type%3Aerror%20%40error.type%3AMissingConfigError&cols=&viz=stream_issues&from_ts=1675106648792&to_ts=1675279448792&live=true

We currently don't have any alerts set up for this, but I will do that now.

PR #22230, opened 2023-02-01
Fix missing secrets prefix while calling GH secret

Mykyta-Serbynevskyi:
Fix missing `secrets` prefix while calling GH secret


octavia-approvington:

octavia-approvington (approved):

PR #22280, opened 2023-02-02
Perform catalog diff on get request

alovew:
Resolves https://github.com/airbytehq/oncall/issues/1413

Instead of checking whether there is a new source catalog id for a connection's source, perform a complete catalog diff to determine whether there are schema changes.

There is a bug where schema refreshes result in new actor_catalogs even when there are no schema changes present. This means users' connection pages tell them there are schema changes, and when they refresh their schemas, there are no changes to apply.

This PR changes the Connection Get endpoint so instead of checking whether there is a new actor_catalog available, we perform a full Catalog Diff between the connection's existing catalog and the most recently detected source catalog.

### DIFF
--- a/airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/WebBackendConnectionsHandler.java
+++ b/airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/WebBackendConnectionsHandler.java
@@ -397,25 +397,13 @@ public class WebBackendConnectionsHandler {
     } else if (catalogUsedToMakeConfiguredCatalog.isPresent()) {
       // reconstructs a full picture of the full schema at the time the catalog was configured.
       syncCatalog = updateSchemaWithOriginalDiscoveredCatalog(configuredCatalog, catalogUsedToMakeConfiguredCatalog.get());
-      Optional<ActorCatalog> mostRecentActorCatalog = configRepository.getMostRecentActorCatalogForSource(connection.getSourceId());
-      // Get the diff between existing connection catalog and most recent fetched catalog
-      if (mostRecentActorCatalog.isPresent()) {
-        final io.airbyte.protocol.models.AirbyteCatalog mostRecentAirbyteCatalog =
-            Jsons.object(mostRecentActorCatalog.get().getCatalog(), io.airbyte.protocol.models.AirbyteCatalog.class);
-        final StandardSourceDefinition sourceDefinition = configRepository.getSourceDefinitionFromSource(connection.getSourceId());
-        diff =
-            connectionsHandler.getDiff(catalogUsedToMakeConfiguredCatalog.get(), CatalogConverter.toApi(mostRecentAirbyteCatalog, sourceDefinition),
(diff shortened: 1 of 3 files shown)


airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/WebBackendConnectionsHandler.java
```diff
@@ -266,8 +263,25 @@ static private WebBackendConnectionListItem buildWebBackendConnectionListItem(
...
+    }
+
+    if (!diff.getTransforms().isEmpty()) {
+      return SchemaChange.NON_BREAKING;
+    }
+
+    return SchemaChange.NO_CHANGE;
+  }
+
+  static SchemaChange getSchemaChangeForList(final ConnectionRead connectionRead,
```
colesnodgrass, line 282: Should this be `@VisibleForTesting` as well or private?

colesnodgrass (approved):
Changes seem fine. 

Do any tests need to be added/modified  as a result of this change? I'm guessing no as these methods all appear to be non-public.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/WebBackendConnectionsHandler.java
alovew: This method is still here to use the old method for the connection list page. I didn't want to calculate a catalog diff for every connection in a workspace since it's kind of an expensive operation, and it's less impactful to see an incorrect schema change notification on the list page vs the connection page itself.

gosusnp (commented):
Very partial review, I didn't get a chance to go very far, just sharing the two points I had noted so far, will come back to it tomorrow (in a few hours)

gosusnp (commented):
Rest look good to me, most pressing question in my mind is whether the `getSchemaChange` can let conflicting information get through.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/WebBackendConnectionsHandler.java
```diff
@@ -266,8 +263,25 @@ static private WebBackendConnectionListItem buildWebBackendConnectionListItem(
...
   static SchemaChange getSchemaChange(
                                       final ConnectionRead connectionRead,
-                                      final Optional<UUID> currentSourceCatalogId,
-                                      final Optional<ActorCatalogFetchEvent> mostRecentFetchEvent) {
+                                      final CatalogDiff diff) {
+    if (connectionRead == null || diff == null) {
+      return SchemaChange.NO_CHANGE;
+    }
+
+    if (connectionRead.getBreakingChange() != null && connectionRead.getBreakingChange()) {
```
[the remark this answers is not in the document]
alovew: yeah, good call, I'll update that

mfsiega-airbyte (approved):
Apart from Jimmy's comment LGTM - if you haven't, I would suggest some manual testing to verify that it all works as intended.

PR #22289, opened 2023-02-02
:window: :bug: Poll backend for Free Connector Program enrollment success

### LINKED ISSUE #22302: Validate free connector program enrollment with backend before showing success/error UI notifications
After a user goes through the Stripe checkout page, Stripe simultaneously redirects them to the `successUrl` we provide and sends a request to a webhook of ours. For a small number of users, this webhook is not successfully processed, preventing us from recognizing them as enrolled and billing accordingly; we should not show any confirmation UI until we have [verified with the Airbyte cloud backend](https://github.com/airbytehq/airbyte-platform-internal/blob/7460a379998bebe72759545cff5bf5c8cf84d2a4/cloud-api/src/main/openapi/config.yaml#L1097-L1106) that the webhook has been processed correctly. The initial implementation didn't account for this, and triggered the confirmation UI as soon as the user navigated to the `successUrl`; this needs to be updated. Lots of context in [this slack thread](https://airbytehq-team.slack.com/archives/C049XDAS19C/p1675089571810259).

So:
- when a user navigates to a page whose render invokes `useFreeConnectorProgram()` with the `fcpEnrollmentSuccess` query param, start polling the FCP program info endpoint
- if polled return value ever says `hasPaymentAccountSaved` is `true`, stop polling, hide enrollment banners, and show success UI
- if some timeout is reached, show an error popup instead

One weakness of this approach: it cannot detect errors in a scenario where users try to update the payment account of record for a workspace. We don't currently have any UI to manage payment accounts, but we'll need it: it's a table stakes feature for actually doing any billing via the saved payment accounts.

ambirdsall:
## What
Closes https://github.com/airbytehq/airbyte/issues/22302.

- when a user navigates to a page whose render invokes `useFreeConnectorProgram()` with the `fcpEnrollmentSuccess` query param, starts polling the FCP program info endpoint
- if polled return value ever says hasPaymentAccountSaved is true, stop polling, hide enrollment banners, and show success UI
- if 10s elapses with no enrollment, show an error popup instead

## How
Adds a general-purpose `pollUntil` function, then uses that to poll the endpoint.

## Recommended reading order
1. `useFreeConnectorProgram.ts`
2. `pollUntil.ts` or `pollUntil.test.ts`, depending on how you feel about rxjs, followed by the other

### DIFF
--- a/airbyte-webapp/src/core/request/pollUntil.ts
+++ b/airbyte-webapp/src/core/request/pollUntil.ts
@@ -5,16 +5,16 @@ import { timer, delay, from, concatMap, takeWhile, last, raceWith, lastValueFrom
 export function pollUntil<ResponseType>(
   apiFn: () => Promise<ResponseType>,
   condition: (res: ResponseType) => boolean,
-  options: { interval: number; maxTimeout?: number }
+  options: { intervalMs: number; maxTimeoutMs?: number }
 ) {
-  const { interval, maxTimeout } = options;
-  const poll$ = timer(0, interval).pipe(
-    concatMap((_) => from(apiFn())),
+  const { intervalMs, maxTimeoutMs } = options;
+  const poll$ = timer(0, intervalMs).pipe(
+    concatMap(() => from(apiFn())),
     takeWhile((result) => !condition(result), true),
     last()
   );
 
-  const timeout$ = maxTimeout ? from([false]).pipe(delay(maxTimeout)) : NEVER;
+  const timeout$ = maxTimeoutMs ? from([false]).pipe(delay(maxTimeoutMs)) : NEVER;
 
   return lastValueFrom(poll$.pipe(raceWith(timeout$)));
 }
--- a/airbyte-webapp/src/core/request/pollUntil.ts
+++ b/airbyte-webapp/src/core/request/pollUntil.ts
@@ -1,0 +1,20 @@
+import { timer, delay, from, concatMap, takeWhile, last, raceWith, lastValueFrom, NEVER } from "rxjs";
+
+// Known issues:
+// - the case where `apiFn` returns `false` and `condition(false) === true` is impossible to distinguish from a timeout
+export function pollUntil<ResponseType>(
+  apiFn: () => Promise<ResponseType>,
+  condition: (res: ResponseType) => boolean,
+  interval: number,
+  maxTimeout?: number
+) {
+  const poll$ = timer(0, interval).pipe(
+    concatMap((_) => from(apiFn())),
+    takeWhile((result) => !condition(result), true),
+    last()
+  );
+
+  const timeout$ = maxTimeout ? from([false]).pipe(delay(maxTimeout)) : NEVER;
+
+  return lastValueFrom(poll$.pipe(raceWith(timeout$)));
+}
--- a/airbyte-webapp/src/core/request/pollUntil.ts
+++ b/airbyte-webapp/src/core/request/pollUntil.ts
@@ -5,9 +5,9 @@ import { timer, delay, from, concatMap, takeWhile, last, raceWith, lastValueFrom
 export function pollUntil<ResponseType>(
   apiFn: () => Promise<ResponseType>,
   condition: (res: ResponseType) => boolean,
-  interval: number,
-  maxTimeout?: number
+  options: { interval: number; maxTimeout?: number }
 ) {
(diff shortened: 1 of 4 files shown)


airbyte-webapp/src/packages/cloud/locales/en.json
```diff
@@ -180,6 +180,7 @@
   "freeConnectorProgram.enrollNow": "Enroll now!",
   "freeConnectorProgram.enroll.description": "Enroll in the <b>Free Connector Program</b> to use Alpha and Beta connectors for <b>free</b>.",
   "freeConnectorProgram.enroll.success": "Successfully enrolled in the Free Connector Program",
+  "freeConnectorProgram.enroll.failure": "Unable to verify that payment details were saved successfully",
```
ambirdsall, line 183: I did wonder if we would like any verbiage about reaching out to support? The user and their browser are faced with a pretty opaque error state. Thoughts, @andyjih?

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/hooks/useFreeConnectorProgram.ts
```diff
@@ -30,16 +32,32 @@ export const useFreeConnectorProgram = () => {
...
-      setUserDidEnroll(true);
-      registerNotification({
-        id: "fcp/enrolled",
-        text: formatMessage({ id: "freeConnectorProgram.enroll.success" }),
-        type: ToastType.SUCCESS,
+      pollUntil(
+        () => webBackendGetFreeConnectorProgramInfoForWorkspace({ workspaceId }, requestOptions),
+        ({ hasPaymentAccountSaved }) => hasPaymentAccountSaved,
+        { interval: 1000, maxTimeout: 10000 }
+      ).then((maybeFcpInfo) => {
```
josephkmh, line 44: Should we also `catch` this promise in case the api call throws an error?

airbyte-webapp/src/core/request/pollUntil.ts
```diff
@@ -0,0 +1,31 @@
...
+ *   with something like `.then(bool => ({ hopefullyUsefulFieldName: bool }))`.
+ */
+export function pollUntil<ResponseType>(
+  apiFn: () => Promise<ResponseType>,
+  conditionFn: (res: ResponseType) => boolean,
+  options: { interval: number; maxTimeout?: number }
+): Promise<ResponseType | false> {
+  const { interval, maxTimeout } = options;
+  const poll$ = timer(0, interval).pipe(
+    concatMap((_) => from(apiFn())), // combine all emitted values into a single stream
```
josephkmh, line 23: nit:
```suggestion
    concatMap(() => from(apiFn())), // combine all emitted values into a single stream
```

joeykmh (approved):
Nice utility! I imagine this could be useful for short polling in other instances 👍

airbyte-webapp/src/packages/cloud/locales/en.json
andyjih: @ambirdsall Good call! Here's a pass: 

```suggestion
  "freeConnectorProgram.enroll.failure": "Unable to verify that payment details were saved successfully. Please contact support for additional help.",
```

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/hooks/useFreeConnectorProgram.ts
```diff
@@ -30,16 +32,32 @@ export const useFreeConnectorProgram = () => {
...
   useEffectOnce(() => {
     if (searchParams.has(STRIPE_SUCCESS_QUERY)) {
       // Remove the stripe parameter from the URL
-      setSearchParams({}, { replace: true });
-      setUserDidEnroll(true);
-      registerNotification({
-        id: "fcp/enrolled",
-        text: formatMessage({ id: "freeConnectorProgram.enroll.success" }),
-        type: ToastType.SUCCESS,
+      pollUntil(
```
timroes, line 40: We have pretty similar code in https://github.com/airbytehq/airbyte/blob/6a7410676941e26613592dd48c52a9f1cde33e29/airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/RemainingCredits.tsx#L73-L91 Maybe it's worth switching that also over to use the `pollUntil` utility.

PR #22291, opened 2023-02-02
Add emitted_at SAT test (22240)

### LINKED ISSUE #22240: [SAT] Assert that record `emitted_at` values increase between subsequent runs
We've recently encountered bugs with normalization when normalizing deeply nested fields that don't have cursors or primary keys for incremental syncs.  Because we can't tell if a new nested object is new or not, we end up creating ever-more rows.  To solve this, the normalization logic will change to only re-normalize object with an `emitted_at` > than then existing records (https://github.com/airbytehq/airbyte/pull/22030).

This is where SAT comes in.  We currently have no guarantees that a records's emitted_at property really is the second as which the source sent the record to the platform, which now has logical meaning.  We know the CDK does the right thing (using `datetime.now()`), but we need to ensure that the resolution of this timestamp is (1) in seconds and (2) really does change between syncs.

Add a new SAT test or extend an existing one so that:
* a READ is preformed using working CONFIG, CATALOG, and STATE such that at least 1 record is produced
* store the max values for emitted at
* sleep for 1 second
* run the READ again, with the same CONFIG, CATALOG, and STATE.
* assert that the same records are produced but with different `emitted_at`s, and the second emitted_at is always greater than the previous run's `max(emitted_at)`

Optional:
* assert that the values for emitted at are +/- 5 seconds from the SAT's internal notion of what time it is

bnchrch:
## What
We've recently encountered bugs with normalization when normalizing deeply nested fields that don't have cursors or primary keys for incremental syncs. Because we can't tell if a new nested object is new or not, we end up creating ever-more rows. 

This adds a test for that case

Closes #22240

## How
Add a new SAT test or extend an existing one so that:

1. a READ is preformed using working CONFIG, CATALOG, and STATE such that at least 1 record is produced
2. store the max value for emitted at
3. sleep for 1 second
4. run the READ again, with the same CONFIG, CATALOG, and STATE.
5. assert that the same records are produced but with different emitted_ats, and the second emitted_ats are always greater than the previous run's max(emitted_at)


evantahler (commented):
I'm testing some connectors ^ ... let's see how they do

airbyte-integrations/bases/source-acceptance-test/unit_tests/test_core.py
```diff
@@ -329,6 +329,94 @@ def test_read(schema, record, expectation):
...
+            [
+                AirbyteMessage(type=Type.RECORD, record=AirbyteRecordMessage(stream="test_stream", data={"aa": 23}, emitted_at=111)),
+                AirbyteMessage(type=Type.RECORD, record=AirbyteRecordMessage(stream="test_stream", data={"aa": 24}, emitted_at=111)),
+            ],
+            [
+                AirbyteMessage(type=Type.RECORD, record=AirbyteRecordMessage(stream="test_stream", data={"aa": 23}, emitted_at=111)),
+                AirbyteMessage(type=Type.RECORD, record=AirbyteRecordMessage(stream="test_stream", data={"aa": 24}, emitted_at=112)),
+            ],
+            pytest.raises(AssertionError, match="emitted_at should increase on subsequent runs")
+        ),
```
evantahler, line 370: 👍 this is the main assertion

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
```diff
@@ -826,6 +828,38 @@ def test_read(
...
+        second_read_reords_emitted_at = [message.record.emitted_at for message in second_read_records]
+
+        min_emitted_at = min(second_read_reords_emitted_at)
+
+        assert max_emitted_at < min_emitted_at, "emitted_at should increase on subsequent runs"
+
+        # Compare two lists, ignoring order of elements.
+        diff = DeepDiff(first_read_records_data, second_read_records_data, ignore_order=True, report_repetition=True)
+
+        assert diff == {}, f"records should be the same on subsequent runs. Diff: {diff}"
```
alafanechere, line 861: I think this "two subsequent runs" logic is already happening [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/tests/test_full_refresh.py#L39). I would suggest adding your new assertions in `TestFullRefresh.test_sequentials_reads` to avoid code duplication and limit the number of READ attempts. Let me know if you think otherwise.

alafanechere (changes requested):

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
bechurch: > limit the number of READ attempts

I think thats a good reason to change. Ill put up a commit shortly

bechurch: Actually @alafanechere .

How much of an impact do reads have to run time?

Should we always been trying to optomize for a low number of reads to run the SAT suite?

alafanechere: It highly depends on the connector under test but its definitely one of the most time-consuming step of the test.

> Should we always been trying to optomize for a low number of reads to run the SAT suite?

I'd say we shall indeed optimize to lower the number of reads and implement caching whenever possible.

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_full_refresh.py
```diff
@@ -78,3 +81,27 @@ def test_sequential_reads(
...
+        ignored_fields = getattr(inputs, "ignored_fields") or {}
+        configured_catalog = full_refresh_only_catalog(configured_catalog)
+        output_1 = docker_runner.call_read(connector_config, configured_catalog)
+        records_1 = [message.record for message in output_1 if message.type == Type.RECORD]
+
+        # sleep for 1 second to ensure that the emitted_at timestamp is different
+        time.sleep(1)
+
+        output_2 = docker_runner.call_read(connector_config, configured_catalog)
+        records_2 = [message.record for message in output_2 if message.type == Type.RECORD]
```
alafanechere, line 102: nit: We could create class scoped fixtures for outputs/records 1 and 2.
We could then re-use these fixtures in differents tests instead of calling the `self.assert_*` functions from `test_sequential_reads`.

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_full_refresh.py
```diff
@@ -36,24 +37,26 @@ def primary_keys_only(record, pks):
...
+        assert first_read_records_data, "At least one record should be read using provided catalog"
+
+        first_read_records_emitted_at = [record.emitted_at for record in first_read_records]
+        max_emitted_at = max(first_read_records_emitted_at)
+
+        second_read_reords_emitted_at = [record.emitted_at for record in second_read_records]
+
+        min_emitted_at = min(second_read_reords_emitted_at)
+
+        assert max_emitted_at < min_emitted_at, "emitted_at should increase on subsequent runs"
```
alafanechere, line 51:
```suggestion
        assert max_emitted_at_first_read < min_emitted_at_second_read, "emitted_at should increase on subsequent runs"
```
I find it a bit confusing to read `max < min`

alafanechere (approved):
Thanks for the changes, only minor suggestions 😄

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_full_refresh.py
bechurch: great suggestion

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_full_refresh.py
bechurch: I opted to not do this as it complicates the down stream unittests.

If we used a fixture for holding onto the data from read 1 and read 2.

The same fixture would not be available when we call the test explicitly in our unit tests.

[semi-relavent stack overflow](https://stackoverflow.com/questions/65547878/pytest-how-to-call-a-function-with-fixture-argument-directly)

[Where we explicitly call the test](https://github.com/airbytehq/airbyte/blob/a2c42d628ce76f1b21ca0d40ab0e082077b7db5d/airbyte-integrations/bases/connector-acceptance-test/unit_tests/test_test_full_refresh.py#L283) (note the arguments which when run against a connector are typically fixtures)

Issue #21561, opened 2023-01-18
[Low-Code CDK] Remove stream_cursor_field from DeclarativeStream


brianjlai:
## Tell us about the problem you're trying to solve
**BREAKING CHANGE**

We redundantly define the cursor field for incremental streams in multiple places on both the `DeclarativeStream.stream_cursor_field` and stream_slicer's `cursor_field`. Similar to other issues, these fields should be the same for a given stream so there is no need to define them twice.

## Describe the solution you would like

Because incremental syncs always require a stream slicer, the `cursor_field` should be defined on that component. The DeclarativeStream component can use the `DeclarativeStream.retriever.stream_slicer` model to get the `cursor_field`.

```
events_stream:
  type: DeclarativeStream
  name: events
  primary_key: id
  retriever:
    type: SimpleRetriever
    stream_slicer:
      type: DatetimeStreamSlicer
      start_datetime:
        datetime: "{{ config['start_date'] }}"
        datetime_format: "%Y-%m-%dT%H:%M:%S%z"
      end_datetime:
        datetime: "{{ now_utc().strftime('%Y-%m-%dT%H:%M:%S%z') }}"
        datetime_format: "%Y-%m-%dT%H:%M:%S%z"
      datetime_format: "%Y-%m-%dT%H:%M:%S.%f%z"
      cursor_granularity: "PT0.000001S"
      step: "P1D"
      cursor_field: timestamp
    requester:
      type: HttpRequester
      url_base: "https://test.something.io/v1/"
      http_method: "GET"
      authenticator:
        type: NoAuthentication
```

## Acceptance Criteria
- [ ] The declarative_component_schema.yaml should be updated w/ `stream_cursor_field` removed from DeclarativeStream
- [ ] DeclarativeStream `stream_cursor_field` is derived from the stream slicer
- [ ] Either a migration is included removing all instances of `DeclarativeStream.stream_cursor_field`
- [ ] Documentation for the DeclarativeStream has been updated
- [ ] An associated ticket to update the connector builder UI frontend has been filed

## Implementation Approach

- We had originally named the field `DeclarativeStream.stream_cursor_field` (which is redundant), because of a collision w/ the `cursor_field()` property method. We can investigate renaming, but this all under the hood so not a big deal.
- It is possible for a stream's retriever to not specify a stream slicer, such as in the case of a stream only supporting full refresh. So in the case that there is no stream slicer, we should just return `[]` like we do in the defaults.

[...]

Issue #19961, opened 2022-12-01
[Low-Code CDK] Separate date time iteration from stream slice iteration within the SimpleRetriever component


brianjlai:
## Tell us about the problem you're trying to solve
**BREAKING CHANGE**

Stream slicers are a difficult concept to reason about for new connector developers to grasp because they cover two related but different concepts: incremental syncs, and partitioning the records. For the `SimpleRetriever` component, the `stream_slicer` field allows for various types of slicers the most common of which include the DatetimeStreamSlicer and the SubstreamSlicer.

## Tell us the solution you would like

In order to clearly delineate components that are used to implement incremental syncs, the `SimpleRetriever` should have a new field called `datetime_range` which accepts a DatetimeStreamSlicer component which will continue to function as it currently does.

In order to clearly delineate components that perform partitioning and iteration based on a heuristic, we will rename the existing `stream_slicer` on the `SimpleRetriever` component to `iterator`. This field will accept `CartesianProductStreamSlicer`, `SingleSlice`, `Substreamslicer`, and `CustomStreamSlicer`.

New Manifest:
```
retriever:
    type: SimpleRetriever
    name: "{{ options['name'] }}"
    primary_key: "{{ options['primary_key'] }}"
    datetime_range:
      type: DatetimeStreamSlicer
      cursor_field: updated_at
      start_datetime: 12-12-2022
      end_datetime: 01-13-2022
      ...
      step: 1D
    iterator:
      type: SubstreamSlicer
      parent_stream_configs
        - stream: "#/definitions/messages_stream"
           parent_key: id
           stream_slice_field: id
```

## Open Questions

- Should we rename DatetimeStreamSlicer to DatetimeRangePartition?
- Do we foresee a need to allow multiple DatetimeStreamSlicers per requester? (We don't currently have connectors that need the cartesian product of multiple stream slicers)

## Acceptance Criteria

- [ ] The declarative_component_schema.yaml should be updated with the decouples `iterator` and `datetime_range`
- [ ] Models have been regenerated
- [ ] Retrievers that have both `iterator` and `datetime_range` defined use the cartesian product of both slicers
- [ ] Existing connectors are migrated so that datetime slicers are under the `datetime_range` field and other slicers are `iterator`. Stream slicer is a frequently used field, so a migration script to rename is likely needed.
- [ ] Documentation is updated to reflect the changes.

## Implementation Approach

A few points of interest during implementation.

[...]

PR #22294, opened 2023-02-02
[Low-Code CDK] Rremove stream_cursor_field from stream and derive it from stream_slicer

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/21561

## What
For connectors that implement incremental syncs, we currently require that the cursor be specified in two location. `stream_cursor_field` on DeclarativeStream and `cursor_field` on various Stream Slicers. This is verbose and often confusing as it is not made clear that a cursors are only used for incremental syncs which are implemented by way of a stream slicer.

## How
This PR deprecates the `stream_cursor_field` field on a DeclarativeStream. Under the hood while constructing declarative components, a Stream will derive the cursor value by looking at -> Stream -> Retriever -> Slicer -> Cursor Field. The schema is modified and relevant constructors.

Where this starts to get a little bit tricky is in how some of the existing connectors were using the `stream_cursor_value` in some unique ways. This wasn't quite a one size fits all solution so I did not use a migration script. I will annotate some comments to explain my thinking on how I modified some of the connectors now that the field has been deprecated on the stream.

## Recommended reading order

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -406,6 +406,8 @@ class ModelToComponentFactory:
     def create_declarative_stream(self, model: DeclarativeStreamModel, config: Config, **kwargs) -> DeclarativeStream:
         retriever = self._create_component_from_model(model=model.retriever, config=config)
 
+        cursor_field = self._get_cursor_field_from_stream_slicer(model)
+
         if model.schema_loader:
             schema_loader = self._create_component_from_model(model=model.schema_loader, config=config)
         else:
@@ -423,11 +425,20 @@ class ModelToComponentFactory:
             primary_key=model.primary_key,
             retriever=retriever,
             schema_loader=schema_loader,
-            stream_cursor_field=model.stream_cursor_field or [],
+            stream_cursor_field=cursor_field or [],
             transformations=transformations,
             config=config,
             parameters={},
         )
+
+    @staticmethod
+    def _get_cursor_field_from_stream_slicer(model: DeclarativeStreamModel) -> Optional[Union[str, List[str]]]:
+        if not model.retriever or not model.retriever.stream_slicer:
+            return None
+        elif hasattr(model.retriever.stream_slicer, "cursor_field"):
+            return model.retriever.stream_slicer.cursor_field
+        else:
+            return None
 
     def create_default_error_handler(self, model: DefaultErrorHandlerModel, config: Config, **kwargs) -> DefaultErrorHandler:
         backoff_strategies = []
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -437,6 +437,22 @@ class ModelToComponentFactory:
             return None
         elif hasattr(model.retriever.stream_slicer, "cursor_field"):
             return model.retriever.stream_slicer.cursor_field
+        elif isinstance(model.retriever.stream_slicer, CartesianProductStreamSlicerModel):
+            # TODO: Remove once we redesign stream slicers
+            # This condition is a temporary hack to derive the cursor_field for streams that use a CartesianProductStreamSlicer that
+            # contains at least one cursor_field implying that it is an incremental stream. We have an upcoming PR that aims to decouple
+            # incremental and iteration concepts and once we support that, we can remove this logic and reference the cursor field stored
+            # in model.incremental_sync.cursor_field.
+            for slicer in model.retriever.stream_slicer.stream_slicers:
+                if hasattr(slicer, "cursor_field"):
+                    return slicer.cursor_field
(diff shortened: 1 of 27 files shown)


airbyte-integrations/connectors/source-callrail/source_callrail/manifest.yaml
```diff
@@ -26,7 +26,6 @@ definitions:
       datetime: "{{ today_utc() }}"
       datetime_format: "%Y-%m-%d"
     step: "#/definitions/step"
-    cursor_field: "{{ parameters.stream_cursor_field }}"
```
brianjlai, line 29: This was a somewhat common case where cursor_field was different between streams, rather than rely on an interpolated reference to the stream’s `stream_cursor_field`, instead we can redefine the `$parameters` block to specify `cursor_field` which is then automatically propagated to the stream slicer when creating the components

airbyte-integrations/connectors/source-woocommerce/source_woocommerce/manifest.yaml
```diff
@@ -36,7 +36,7 @@ definitions:
     end_time_option:
       field_name: modified_before
       inject_into: request_parameter
-    cursor_field: "{{ parameters['stream_cursor_field'] }}"
+    cursor_field: "date_modified_gmt"
```
brianjlai, line 39: Some streams used the same `cursor_field` but used interpolation to reference `stream_cursor_field`, I removed all instances and instead defined it directly the stream slicer where it should belong

airbyte-integrations/connectors/source-gocardless/source_gocardless/manifest.yaml
```diff
@@ -49,7 +49,6 @@ streams:
   - type: DeclarativeStream
     $parameters:
       name: "payments"
-      stream_cursor_field: "created_at"
```
brianjlai, line 52: Some cases where the stream_cursor_field was defined but unneeded by a full refresh only streams

airbyte-integrations/connectors/source-greenhouse/source_greenhouse/manifest.yaml
```diff
@@ -57,7 +57,6 @@ definitions:
         $ref: "#/definitions/requester"
   base_incremental_stream:
     $ref: "#/definitions/base_stream"
-    stream_cursor_field: "updated_at"
```
brianjlai, line 60: Greenhouse is an example of a custom date time stream slicer where the cursor_field is an explicit field on the custom slicer. We can still extract it from there and store it on the stream

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -423,12 +425,37 @@ def create_declarative_stream(self, model: DeclarativeStreamModel, config: Confi
...
             parameters={},
         )
 
+    @staticmethod
+    def _get_cursor_field_from_stream_slicer(model: DeclarativeStreamModel) -> Optional[Union[str, List[str]]]:
+        if not model.retriever or not model.retriever.stream_slicer:
+            return None
+        elif hasattr(model.retriever.stream_slicer, "cursor_field"):
+            return model.retriever.stream_slicer.cursor_field
+        elif isinstance(model.retriever.stream_slicer, CartesianProductStreamSlicerModel):
```
brianjlai, line 440: This is where things get just a little bit crazy specifically for Posthog and Zenloop because they use either a CartesianProductStreamSlicer or a custom CartesianProductStreamSlicer so we cannot easily derive what the `cursor_field` should be since there are multiple underlying slicers.

This is a temporary hack where we look for the first slicer we can find that has a cursor_field and use that for the current stream. This isn't a long term solution as the comments mention. Rather, once we separate the `incremental` and `iterable` concepts we can then directly reference the incremental implementation's cursor field and get rid of the existing conditions here.

maxi297: Death to this method (... eventually)!

maxi297 (approved):
LGTM! Eager to unblock the clean up with #19961

brianjlai:
@grubberr 
Yes that is correct. I tried to account for this using a temporary workaround here: https://github.com/airbytehq/airbyte/pull/22294/files#r1093992449

This was a case I noticed during the migration. In the case where the cartesian stream slicer is defined, I opt to get the first occurrence of a cursor_field when it exists. So far, I have only seen cartesian product stream slicers where only one slicer in the list has a `cursor_field` and in that scenario we return the first one we find. Does that sound okay to you?

grubberr:
@brianjlai yes I have 2 cases: https://github.com/airbytehq/airbyte/pull/20960 https://github.com/airbytehq/airbyte/pull/21446
for both it has to be ok.

Thank you

brianjlai:
Yes, I believe both should work with this approach. Both of your new connectors have custom slicer implementations that inherit `CartesianProductStreamSlicer` and both have a child DatetimeStreamSlicer which should allow the stream to derive the `cursor_field` which is passed from the `$options`. There is another existing connector `source-posthog` that uses a custom CartesianProductStreamSlicer that I will double check works w/ these changes that should behave similarly to yours.

brianjlai:
@grubberr 
I verified that for `source-posthog` that the stream's underlying field `stream_cursor_field` will be populated w/ the value of the date slicer's `cursor_value` even though it is underneath a cartesian.

PR #22301, opened 2023-02-02
Connector acceptance test: Fix discovered catalog caching for different configs

grubberr:
## What

- for different configs keep different cached values of discovered catalog
- always reset cache dict if `cache_discovered_catalog=True` (in previous implementation we overwrite on top of old cache - which is not good)


evantahler (approved):
👍 pending the connector tests above ^

grubberr:

PR #22303, opened 2023-02-02
Source Sentry: turn on default availability strategy

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1383

## How
Use default availability strategy not to fail syncs when 403 error is risen


lazebnyi (approved):

airbyte-integrations/connectors/source-sentry/source_sentry/streams.py
```diff
@@ -26,10 +25,6 @@ def __init__(self, hostname: str, **kwargs):
     def url_base(self) -> str:
         return self._url_base
 
-    @property
-    def availability_strategy(self) -> Optional["AvailabilityStrategy"]:
```
roman-yermilov-gl, line 30: As I understand from oncall comments it needs `availability_strategy` to be turned on by default for this connector. But if to just remove the property from the stream, then nothing to be called by default. It must be specified explicitly for something to happen

roman-yermilov-gl (changes requested):

airbyte-integrations/connectors/source-sentry/source_sentry/streams.py
davydov-d: The base class defines this property, so I just removed the override

roman-yermilov-gl: Right, they add it to the base class yesterday

roman-yermilov-gl (approved):

pedroslopez (approved):

PR #22304, opened 2023-02-02
🪟 🎨🔧 Remove oauth position ff

letiescanciano:
We validated that having oauth signup methods on top led to better conversion rate.

This PR just removes the feature flag we used and set the ouath on top of the signup form.

I also took the chance to change the position in the login page for consistency 

Screen Shot 2023-02-02 at 10 41 22 AM


juweins (commented):
LGTM! It is much more convenient to have the SSO buttons on top.

timroes (approved):
Code LGTM, aside from one minor left over that needs to be removed. Tested locally works as expected on login and signup. Please don't forget to delete the flag in LaunchDarkly after merging.

PR #22308, opened 2023-02-02
Source Zendesk Support: add availability strategy

### LINKED ISSUE #21788: Source Zendesk Support: Enable AvailabilityStrategy

roman-yermilov-gl:
## What
*Add availability strategy. Fix tests: fix expected records, add missing streams to configured catalog, handle error for tickets stream when state is abnormal*


lazebnyi (approved):

erohmensing (changes requested):
I think the availabilitystrategy looks good. However can we split these into separate PRs? There is a lot going on here, and it'd be good to be able to easily revert an enhancement (adding availabilitystrategy) without also reverting fixing expected records, fixing configured catalog, etc

roman-yermilov-gl:
> I think the availabilitystrategy looks good. However can we split these into separate PRs? There is a lot going on here, and it'd be good to be able to easily revert an enhancement (adding availabilitystrategy) without also reverting fixing expected records, fixing configured catalog, etc

Done: https://github.com/airbytehq/airbyte/pull/22483

airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/streams.py
```diff
@@ -118,8 +122,8 @@ def __init__(self, subdomain: str, start_date: str, ignore_pagination: bool = Fa
         self._ignore_pagination = ignore_pagination
 
     @property
-    def availability_strategy(self) -> Optional["AvailabilityStrategy"]:
-        return None
+    def availability_strategy(self) -> Optional[AvailabilityStrategy]:
+        return HttpAvailabilityStrategy()
```
erohmensing, line 126: Just wondering, why not define all streams availabilitystrategy as the `ZendeskSupportAvailabilityStrategy` here instead of in `SourceZendeskSupportStream`?

erohmensing (approved):

airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/streams.py
roman-yermilov-gl: We have two types of streams. One of them requires custom `ZendeskSupportAvailabilityStrategy`. So I put `ZendeskSupportAvailabilityStrategy` to the stream definition which need customization

erohmensing: Got it!

PR #22320, opened 2023-02-02
Add API source to tracking data

### LINKED ISSUE #22218: Track source of API requests in Segment
The server should for every Segment `track` event that's send out in the context of a HTTP requests happening against it, check if the request has a `X-Airbyte-Analytic-Source` header, and if so attaches the content of this header to a property called `airbyte_source` in the event.

Notes from the discussion with Jonathan: We should be able to get access to the headers potentially by inecting `ServerRequestContext`.

cc @jdpgrailsdev 

Related issues:
* https://github.com/airbytehq/airbyte/issues/22217
* https://github.com/airbytehq/airbyte/issues/22215
* https://github.com/airbytehq/airbyte/issues/23680

jdpgrailsdev:
## What
* Track source of API requests in Segment
* fix airbytehq/airbyte-internal-issues#1344

## How
* Check if the current HTTP request contains the API source header.  If so, add it to the tracked data

## Recommended reading order
1. `SegmentTrackingClient.java`

## Tests

* Added unit test to verify behavior
* All other tests pass
* Project builds locally


airbyte-analytics/src/main/java/io/airbyte/analytics/SegmentTrackingClient.java
```diff
@@ -138,4 +146,15 @@ public void track(@Nullable final UUID workspaceId, final String action, final M
         .properties(mapCopy));
   }
 
+  private Optional<String> getAirbyteSource() {
+    final Optional<HttpRequest<Object>> currentRequest = ServerRequestContext.currentRequest();
+    if (currentRequest.isPresent()) {
+      if (currentRequest.get().getHeaders().contains(AIRBYTE_ANALYTIC_SOURCE_HEADER)) {
+        return Optional.of(currentRequest.get().getHeaders().get(AIRBYTE_ANALYTIC_SOURCE_HEADER));
+      }
```
colesnodgrass, line 154: can this be reduced to an `Optional.ofNullable` call?

colesnodgrass (approved):

airbyte-analytics/src/main/java/io/airbyte/analytics/SegmentTrackingClient.java
jdpgrailsdev: Sure

timroes:
The header will also be needed to added to the `ACCESS_CONTROL_ALLOW_HEADERS` headers in our CorsFilter, so we can send this header cross domain to the backend. This shouldn't be required for production strictly spoken, since we're usually run on the same domain, but during development we usually develop against a cloud backend, so it would be nice if we could have that header allowed for CORS.

benmoriceau (approved):

jdpgrailsdev:
> ACCESS_CONTROL_ALLOW_HEADERS

@timroes The custom `CorsFilter` will go away once we convert Airbyte Cloud's wrapped server to Micronaut.  I will add the header to it for now so it works in the meantime.

PR #22332, opened 2023-02-02
use PAT instead of GITHUB_TOKEN for test_reporter action

c-p-b:
## What
By default this action uses GITHUB_TOKEN which has a pretty low rate limit. Sometimes it fails for exceeding that rate limit

## How
use env.PAT which is derived from find a pat script instead


supertopher (approved):
looks fly to me.

edmundito (approved):

perangel:
Testing a comment

evantahler:
@cpdeethree can we close this PR?

cpdeethree:
I think so, it looks like after monorepo we stopped hitting the rate limit for this

Issue #22331, opened 2023-02-02
Destination S3: Wrong header on Step 2


juweins:
### Tell us about the documentation you'd like us to add or update.
There is a wrong header for step 2 in destination S3.

### If applicable, add links to the relevant docs that should be updated
- [Source S3](https://docs.airbyte.com/integrations/destinations/s3)

### Screenshot addressing this issue

<img width="683" alt="image" src="https://user-images.githubusercontent.com/26652109/216405114-38e01dd7-b25c-43e9-adb7-9b1406adba78.png">

Edit: Its on Destination S3, not Source.

juweins:
I will work on this and create a PR shortly.

juweins:
Closed with PR #22333

PR #22333, opened 2023-02-02
Fixed: Header for Destination S3: Wrong header on Step 2 #22331

juweins:
## What
There is a mistake in the outline due to a wrong header type.

This PR addresses #22331 

## How
degraded header type for step 2 to match expected outline.

## 🚨 User Impact 🚨
minor

## Pre-merge Checklist

Edit: Added Issue Number for tracking


marcosmarxm (approved):
Thanks @juweins

Issue #22242, opened 2023-02-01
Cannot create a custom connector without documentation URL


girarda:
## Current Behavior
*We cannot create a custom component if its spec's documentationURL isn't set

## Expected Behavior
* Create a connector even if the documentationURL is not set

## Logs
```
2023-02-01 14:01:07 java.lang.NullPointerException: Cannot invoke "java.net.URI.toString()" because the return value of "io.airbyte.protocol.models.ConnectorSpecification.getDocumentationUrl()" is null
2023-02-01 14:01:07     at io.airbyte.commons.server.handlers.SchedulerHandler.getSourceDefinitionSpecification(SchedulerHandler.java:343) ~[io.airbyte-airbyte-commons-server-0.40.32.jar:?]
2023-02-01 14:01:07     at io.airbyte.server.apis.SourceDefinitionSpecificationApiController.lambda$getSourceDefinitionSpecification$0(SourceDefinitionSpecificationApiController.java:35) ~[io.airbyte-airbyte-server-0.40.32.jar:?]
2023-02-01 14:01:07     at io.airbyte.server.apis.ApiHelper.execute(ApiHelper.java:18) ~[io.airbyte-airbyte-server-0.40.32.jar:?]
2023-02-01 14:01:07     at io.airbyte.server.apis.SourceDefinitionSpecificationApiController.getSourceDefinitionSpecification(SourceDefinitionSpecificationApiController.java:35) ~[io.airbyte-airbyte-server-0.40.32.jar:?]
2023-02-01 14:01:07     at io.airbyte.server.apis.$SourceDefinitionSpecificationApiController$Definition$Exec.dispatch(Unknown Source) ~[io.airbyte-airbyte-server-0.40.32.jar:?]
2023-02-01 14:01:07     at io.micronaut.context.AbstractExecutableMethodsDefinition$DispatchedExecutableMethod.invoke(AbstractExecutableMethodsDefinition.java:378) ~[micronaut-inject-3.8.3.jar:3.8.3]
2023-02-01 14:01:07     at io.micronaut.context.DefaultBeanContext$4.invoke(DefaultBeanContext.java:594) ~[micronaut-inject-3.8.3.jar:3.8.3]
2023-02-01 14:01:07     at io.micronaut.web.router.AbstractRouteMatch.execute(AbstractRouteMatch.java:303) ~[micronaut-router-3.8.3.jar:3.8.3]
[...]

PR #22340, opened 2023-02-02
[Connector Builder] Set placeholder documentation_url for builder-generated spec

lmossman:
## What
See slack context: https://airbytehq-team.slack.com/archives/C03TP6W8081/p1675365814016559?thread_ts=1675198798.481999&cid=C03TP6W8081

Manifests generated by the connector builder currently do not pass integration tests and cannot be used, because they have an empty documentation_url.

This issue platform issue will handle fixing the airbyte-server to not fail in this case: https://github.com/airbytehq/airbyte/issues/22242
But until that is complete, we should have an intermediate solution here to fix the problem in the meantime.

## How
This PR addresses the issue by setting `documentation_url` in the manifest produce by the connector builder to the placeholder documentation url we have used in other connectors so far.

This can be removed once https://github.com/airbytehq/airbyte/issues/22242 is complete

### DIFF
--- a/airbyte-webapp/src/components/connectorBuilder/types.ts
+++ b/airbyte-webapp/src/components/connectorBuilder/types.ts
@@ -636,7 +636,7 @@ export const convertToManifest = (values: BuilderFormValues): ConnectorManifest 
 
   const spec: Spec = {
     connection_specification: specSchema,
-    documentation_url: "https://docsurl.com",
+    documentation_url: "https://example.org",
     type: "Spec",
   };
 
--- a/airbyte-webapp/src/components/connectorBuilder/types.ts
+++ b/airbyte-webapp/src/components/connectorBuilder/types.ts
@@ -636,7 +636,7 @@ export const convertToManifest = (values: BuilderFormValues): ConnectorManifest 
 
   const spec: Spec = {
     connection_specification: specSchema,
-    documentation_url: "",
+    documentation_url: "https://docsurl.com",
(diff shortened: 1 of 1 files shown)


airbyte-webapp/src/components/connectorBuilder/types.ts
```diff
@@ -636,7 +636,7 @@ export const convertToManifest = (values: BuilderFormValues): ConnectorManifest
 
   const spec: Spec = {
     connection_specification: specSchema,
-    documentation_url: "",
+    documentation_url: "https://docsurl.com",
```
flash1293, line 639: nit: What about using `https://example.org/` for this? Its official purpose is to be used as an example

flash1293 (approved):
Makes sense, thanks for fixing. I'm wondering whether we should have an integration test for this path (configure something, download yaml, then make it work), but maybe it's not worth it as the end to end flow in the UI will cover it will enough once that's implemented

airbyte-webapp/src/components/connectorBuilder/types.ts
lmossman: I like that suggestion, will update

lmossman:
Yeah I think an E2E test to do that whole flow in the UI makes sense to add once we have the full flow implemented

lmossman:
Yeah I think an E2E test to do that whole flow in the UI makes sense to add once we have the full flow implemented

PR #22344, opened 2023-02-02
Remove manual docker upgrades from workflows

timroes:
## What

Removes the manual docker upgrades in the workflow files. We should always extend the AMI image that's used for our CI runners when we need to make adjustments to packages, and not put it into the workflow files. This also is flaky since Gradle parallelizes tasks and we're manually messing with lock files here of apt.

Since our AMI used for CI runners already has the newest docker version as of today, removing those upgraded should be fine.


gosusnp (approved):

git-phu (approved):

supertopher (approved):
this can only succeed

PR #22348, opened 2023-02-02
Only run orval when the api specs change

ambirdsall:
## What
Wraps `npm run generate-client` usage in lifecycle scripts with `make`, which will skip running the code-generation if (and only if) all generated output files exist and have a modified timestamp more recent than that of their respective input files.

In practice, this change completely eliminates the multi-second lag from rebuilding the api code when nothing has changed (which is most of the time), which makes (re)starting a vite dev server in particular feel much closer to instantaneous.

## Try it out!
The `prestart`, `pretest`, and `prestart:cloud` scripts all use `make` to only run `npm run generate-client` at need, speeding up `npm start`, `npm test`, and `npm run start:cloud`. There's also a temporary test commit adding `js:start`, `js:test`, and `js:start:cloud` scripts for head-to-head comparisons. This trick helps, too:
```diff
-    "prestart": "make generate-client",
+    "prestart": "time make generate-client",

-    "prejs:start": "npm run generate-client",
+    "prejs:start": "time npm run generate-client",
```

## one new potential problem
Because the `Makefile` maintains its own graph of input/output file dependencies, it needs to stay in sync with `orval.config.ts`. I thought it made sense to add a few lines to the `pre-commit` git hook that will fail with an error message if any `target` file defined in `orval.config.ts` is not a dependency of `make generate-client`. It seems to work as expected in my manual testing, but I would appreciate extra eyes and fingers poking at this to make sure it's something we can rely on.

## a bit more on the time savings
I measured the code generation step on an M1 at ≈2.68s; a normal-fast computer like my thinkpad (x86 processor, same amount of RAM, also SSD), is just about 2s slower yet:
```
┌ ~/c/airbyte/airbyte-webapp (alex/only-rebuild-api-client-on-changes|✔) 🌓 17fca0ddeb5
└➣ time npm run generate-client

> airbyte-webapp@0.40.32 generate-client
> orval

🍻 Start orval v6.11.0-alpha.10 - A swagger client generator for typescript
🎉 api - Your OpenAPI spec has been converted into ready to use orval!
🎉 connectorBuilder - Your OpenAPI spec has been converted into ready to use orval!
🎉 connectorManifest - Your OpenAPI spec has been converted into ready to use orval!
npm run generate-client  2.68s user 0.36s system 107% cpu 2.833 total
```

Here's skipping the code generation because it's already up-to-date:
```
┌ ~/c/airbyte/airbyte-webapp (alex/only-rebuild-api-client-on-changes|✔) 🌓 17fca0ddeb5
└➣ time make generate-client
make: Nothing to be done for `generate-client'.
make generate-client  0.00s user 0.01s system 59% cpu 0.023 total
```
Analysis: not doing things is faster.

When one of the openAPI yaml spec files has a more recent modified timestamp, `make generate-client` runs the npm script exactly once*, with negligible overhead:

\* Okay, it's once _per openApi spec/client pair_, which means there's a rare edge case where a single npm script run's startup will be a bit slower with this update. Even then, we only have 3 openAPI specs, which means even in the worst case scenario it will only takes 4 combined `npm start`/`npm test` invocations for the amortized runtime under `make` to be faster.
```
┌ ~/c/airbyte/airbyte-webapp (alex/only-rebuild-api-client-on-changes|✔) 🌓 17fca0ddeb5
└➣ touch ../airbyte-api/src/main/openapi/config.yaml

┌ ~/c/airbyte/airbyte-webapp (alex/only-rebuild-api-client-on-changes|✔) 🌓 17fca0ddeb5
└➣ time make generate-client
echo building AirbyteClient.ts
building AirbyteClient.ts
npm run generate-client

> airbyte-webapp@0.40.32 generate-client
> orval

🍻 Start orval v6.11.0-alpha.10 - A swagger client generator for typescript
🎉 api - Your OpenAPI spec has been converted into ready to use orval!
🎉 connectorBuilder - Your OpenAPI spec has been converted into ready to use orval!
🎉 connectorManifest - Your OpenAPI spec has been converted into ready to use orval!
make generate-client  2.63s user 0.33s system 125% cpu 2.349 total

┌ ~/c/airbyte/airbyte-webapp (alex/only-rebuild-api-client-on-changes|✔) 🌓 17fca0ddeb5
└➣ time make generate-client
make: Nothing to be done for `generate-client'.
make generate-client  0.01s user 0.01s system 63% cpu 0.021 total
```

## 🚨 User Impact 🚨
This does add a new tool as a dev dependency to the frontend build, though not to the repo as a whole:
```
┌ ~/c/airbyte (alex/only-rebuild-api-client-on-changes|✔) 🌓 17fca0ddeb5
└➣ find . -name '[Mm]akefile' | grep -vE 'node_modules/|\.gradle/'
./airbyte-integrations/connectors/source-google-analytics-data-api/Makefile
./airbyte-cdk/python/reference_docs/Makefile
./airbyte-webapp/Makefile
```

`make` is near-universally preinstalled on unixy operating systems, though, and basically universally available; and our `package.json` scripts already use windows-incompatible shell commands anyway. If our build system handles it, it should be a completely transparent addition to existing development workflows.


ambirdsall:
Closing in favor of https://github.com/airbytehq/airbyte/pull/23128

PR #22351, opened 2023-02-03
🪟 🔧 [Connector Builder] Add telemetry for landing page events

lmossman:
## What
Adds telemetry for actions taken in the new connector builder landing page

## How
Moves the `Connector Builder Start` event from ConnectorBuilderEditPage to ConnectorBuilderLandingPage, replacing that event in the Edit page with a `ConnectorBuilderEdit` event, since the landing page is the new "starting point".

Tracks events for each of the cases that the user can end up in when importing:
- Start from scratch
- Invalid YAML uploaded
- UI-incompatible YAML imported
- UI-compatible YAML imported


flash1293 (approved):
LGTM, tested with Amplitude and works as expecte

PR #22353, opened 2023-02-03
Catch empty state in incremental SAT

### LINKED ISSUE #21863: [SAT] - Test that STATE emitted during run matches provided sample states
From https://github.com/airbytehq/oncall/issues/1317 which was solved by https://github.com/airbytehq/airbyte/pull/21864

We don't test that syncs /actually emit/ a state message.  We should do that. 
We should also confirm that the format (e.g. the keys in the state JSON) match the state messages which are provided as inputs to SAT. 

To discuss:
* do all or only some properties of a state message need to match?
* can we determine based on the catalog (and assuming per-stream-state) which state messages we expect?

bnchrch:
## What
Improves the incremental SAT test to catch the case in which the STATE message yields no cursor value

closes #21863 

## Whats left todo

- [x] Add unit tests
- [x] Break test into own test case
- [x] Discuss concerns with team

## Concerns
@airbytehq/connector-operations I have a few areas im unsure on with these tests that I want to discuss before making any large changes

### 1. `records_with_state` can emit an empty list even if called with many records
One of the issues here is that the `records_with_state` function uses the `continue` function if no `state_value` is found for a record.

If this happens for all records then BOOM empty list.

In the previous context of the `test_two_sequential_reads` test that is considered a pass.

But to me that seems like an error, if no state can be infered from any record using the cursor, then they havent given us a passing test config.

Is that right?

**Impact if so: changing the SAT to account for it may cause previously passing connectors to fail**

### 2. In `test_two_sequential_reads` we do not ensure that we get data on the second read
We should make sure that the test case can test for two reads that result in data. 

Is that right?

**Impact if so: changing the SAT to account for it WILL cause previously passing connectors to fail. Latest version of sentry for example**

### DIFF
--- a/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py
+++ b/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py
@@ -83,7 +83,7 @@ def configured_catalog_for_incremental_fixture(configured_catalog) -> Configured
     return catalog
 
 
-def records_with_state(records, state, stream_mapping, state_cursor_paths) -> Iterable[Tuple[Any, Any]]:
+def records_with_state(records, state, stream_mapping, state_cursor_paths) -> Iterable[Tuple[Any, Any, Any]]:
     """Iterate over records and return cursor value with corresponding cursor value from state"""
     for record in records:
         stream_name = record.record.stream
@@ -180,7 +180,14 @@ class TestIncremental(BaseTest):
             latest_state = states_1[-1].state.data
             state_input = states_1[-1].state.data
 
-        for record_value, state_value, stream_name in records_with_state(records_1, latest_state, stream_mapping, cursor_paths):
+
+        parsed_records_1 = records_with_state(records_1, latest_state, stream_mapping, cursor_paths);
+
+        # This catches the case of a connector that emits an invalid state that is not compatible with the schema
+        # See https://github.com/airbytehq/airbyte/issues/21863 to understand more
+        assert parsed_records_1, "Should produce at least one record with state"
+
+        for record_value, state_value, stream_name in parsed_records_1:
             assert (
                 record_value <= state_value
             ), f"First incremental sync should produce records younger or equal to cursor value from the state. Stream: {stream_name}"
@@ -188,16 +195,24 @@ class TestIncremental(BaseTest):
         output = docker_runner.call_read_with_state(connector_config, configured_catalog_for_incremental, state=state_input)
         records_2 = filter_output(output, type_=Type.RECORD)
 
+
+        assert records_2, "Should produce at least one record on subsequent read"
+
+        parsed_records_2 = records_with_state(records_1, latest_state, stream_mapping, cursor_paths);
+        assert parsed_records_2, "Should produce at least one record with state"
+
+
         for record_value, state_value, stream_name in records_with_state(records_2, latest_state, stream_mapping, cursor_paths):
             assert compare_cursor_with_threshold(
                 record_value, state_value, threshold_days
             ), f"Second incremental sync should produce records older or equal to cursor value from the state. Stream: {stream_name}"
 
+
     def test_read_sequential_slices(
         self, inputs: IncrementalConfig, connector_config, configured_catalog_for_incremental, cursor_paths, docker_runner: ConnectorRunner
     ):
         """
-        Incremental test that makes calls the read method without a state checkpoint. Then we partition the results by stream and
+        Incremental test that makes calls to the read method without a state checkpoint. Then we partition the results by stream and
         slice checkpoints.
         Then we make additional read method calls using the state message and verify the correctness of the
         messages in the response.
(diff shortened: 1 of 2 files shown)


erohmensing:
Caveat: I'm also unsure, I haven't done much on incremental tests yet. That being said, my interpretation: 

> If this happens for all records then BOOM empty list.
In the previous context of the test_two_sequential_reads test that is considered a pass.
But to me that seems like an error, if no state can be infered from any record using the cursor, then they havent given us a passing test config.
Is that right?

I think that sounds right - if the cursor_path supplied in the `acceptance-test-config.yml` doesn't actually point to anything, that sounds like something that should fail and be updated 🤔 

> In test_two_sequential_reads we do not ensure that we get data on the second read
We should make sure that the test case can test for two reads that result in data.
Is that right?

Reading brian's message, it sounds to me like ensuring we don't get data on the second read was intentional - basically saying if we do the sync and output the state message that we already processed all the data, if the state message was correct, on a second run, we shouldn't have more data to process. I think we would have to artificially inject new data after the first sync in order to have new cursor values (e.g. `updated_at`) to process in the second sync?

bechurch:
> Reading brian's message, it sounds to me like ensuring we don't get data on the second read was intentional - basically saying if we do the sync and output the state message that we already processed all the data, if the state message was correct, on a second run, we shouldn't have more data to process. I think we would have to artificially inject new data after the first sync in order to have newly updated cursor values (e.g. updated_at) to process in the second sync?

Ok so that makes sense that we want to test the case `second read of same data with new cursor yields no records`

Should we explicitly assert that?

Currently in the [code](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py#L248-L254): 
```py
output = docker_runner.call_read_with_state(connector_config, configured_catalog_for_incremental, state=state_input)
            records = filter_output(output, type_=Type.RECORD)

            for record_value, state_value, stream_name in records_with_state(records, complete_state, stream_mapping, cursor_paths):
                assert compare_cursor_with_threshold(
                    record_value, state_value, threshold_days
                ), f"Second incremental sync should produce records older or equal to cursor value from the state. Stream: {stream_name}"
```

`records_with_state` on second read can, and in the case of sentry, does return None.

But sometimes it does not, and then we test for the case `Second incremental sync should produce records older or equal to cursor value from the state`

My worry here is by not having test cases explicit for 
- cursor from records applied to same records yields None
- Second read leads to records ahead of cursor from first read

We end up only ever testing one case but not the other.

bechurch:
Just going to ping @evantahler and @pedroslopez to get your thoughts.

pedroslopez:
>  records_with_state on second read can, and in the case of sentry, does return None.
> 
> But sometimes it does not.
> 
> And only then we test for the case Second incremental sync should produce records older or equal to cursor value from the state

This is intentionally structured this way because the implementation is kind of left up to the connector in how it wants to use state. Specifically, in whether > or >= is used when retrieving records from the API.

An example is, say you save state with the date 2020-01-01. Since this date isn't very granular, the implementation would likely use a "greater than or equal to" to retrieve the records so that we wouldn't miss any new record that may have been added on that date. This means that we may also get repeated records, but that's ok since we have "at least once" deliverability. If there wasn't any data at this point, or the connector implements the state using ">", then we would not get any records.

evantahler:
I'm having trouble parsing that python code (I think the test might be backwards? Is that the new state or starting state that's being compared?) but I think the `None` case is weird. If the CDK were stronger-typed, I think READ should always produce at least `[]` rather than `None`, perhaps?

I also think that Airbyte is a "deliver at least once" system, and we should always be choosing to re-send a record if it's the safer thing to do.  So, I think that all sources probably should be using `>=` cursor comparisons, and the test should always also produce at least one record from the previous read?  I guess that's not a hard rule if the source can be sure that cursors are high-resolution and un-changing.  There's another test with "abnormally large state" that should produce no records, where the cursor is far higher than what would be reasonable (e.g. `updated_at=9999-01-01`)

bechurch:
@pedroslopez 

I realized I had a typo in the message you quoted. (Sorry!)

Instead of `records_with_state` I meant `records = filter_output(output, type_=Type.RECORD)`

The second read from `test_incremental_two_sequential_reads` for sentry currently returns no records. Even before we filter for state.

(I updated the previous comment to correct this)

Does that change your answer at all?

pedroslopez:
@bechurch the filtering I'm referring to happens on the source side or by the API it calls, not something in our testing code. It could still come back with or without records for the reasons I mentioned (no new records have been emitted / using > in the source would lead to no records, >= would lead to repeated records)

bechurch:
@pedroslopez Ok that makes sense.

I'm still stuck on one thing though.

Don't we want to have connectors give us enough data to be able to test in the context of "sequential reads, that both return data"?

bechurch:
FYI from my wonderful call with Pedro

1. I'm going to keep this PR focused around empty state during first pass
2. Im going to add an issue to grooming around testing the return the of second read for data as its likely a useful test case we aren't currently supporting
3. We likely want to split these cases out into separate named tests

bechurch:
@connector-operations This is ready for review again.

alafanechere (approved):
Nice, could you please run a couple of `/test` and the connectors that have this problem?

bechurch:
/test connector=connectors/source-sentry

Note: This test will fail.... but not for the test ive added

PR #22354, opened 2023-02-03
JDBC cli tool

edgao:
usage - probably needs some tweaking. Put a dest-postgres config in `secrets/conf.json`.
```java
./gradlew :airbyte-int:conn:destination-postgres:build -x test
cd airbyte-integrations/connectors/destination-postgres
cd build/distributions
tar -xf destination-postgres-0.40.32.tar
cd -
java -cp 'build/distributions/destination-postgres-0.40.32/lib/*' io.airbyte.integrations.destination.postgres.PostgresCli "$(cat secrets/conf.json)" schema1 table1 schema2 table2
```

### DIFF
--- a/airbyte-integrations/connectors/destination-postgres/src/main/java/io/airbyte/integrations/destination/postgres/PostgresCli.java
+++ b/airbyte-integrations/connectors/destination-postgres/src/main/java/io/airbyte/integrations/destination/postgres/PostgresCli.java
@@ -10,7 +10,7 @@ import java.util.ArrayList;
 import java.util.List;
 import org.apache.commons.lang3.tuple.Pair;
 
-public class PostgresCil {
+public class PostgresCli {
 
   public static void main(final String[] args) throws Exception {
     if (args.length < 3 || args.length % 2 != 1) {
@@ -20,7 +20,7 @@ public class PostgresCil {
 
     final String configStr = args[0];
     final List<Pair<String, String>> tables = new ArrayList<>();
-    for (int i = 1; i < args.length; i++){
+    for (int i = 1; i < args.length; i += 2){
       tables.add(Pair.of(args[i], args[i + 1]));
     }
 
@@ -43,13 +43,11 @@ public class PostgresCil {
             }
           }
 
+          System.out.println("All tables found to exist; dropping them now.");
           // If they all exist, drop them
           for (final Pair<String, String> tableIdentifier : tables) {
-            final ResultSet tableList = dbm.getTables(null, tableIdentifier.getLeft(), tableIdentifier.getRight(), null);
-            if (!tableList.next()) {
-              ops.dropTableIfExists(database, tableIdentifier.getLeft(), tableIdentifier.getRight());
-              System.exit(1);
-            }
+            System.out.println("Attempting to drop " + tableIdentifier);
+            ops.dropTableIfExists(database, tableIdentifier.getLeft(), tableIdentifier.getRight());
           }
         }
     );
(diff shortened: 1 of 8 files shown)


airbyte-integrations/connectors/destination-postgres/src/main/java/io/airbyte/integrations/destination/postgres/PostgresCli.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.List;
+import org.apache.commons.lang3.tuple.Pair;
+
+public class PostgresCli {
+
+  public static void main(final String[] args) throws Exception {
+    if (args.length < 3 || args.length % 2 != 1) {
+      System.out.println("Usage: java -jar whatever.jar <postgres-config> <schema1> <table1> [<schema2> <table2> ...]");
+      System.exit(1);
+    }
```
jbfbell, line 19: Not gonna nitpick too much but just want to make sure I understand the logic here
argument 0 - config
then all arguments should be in pairs, with at least 3 args (config, schema, table)
so it should be >= 3 and if >3 an odd number?
Logic is fine just making sure I'm reading that correctly

airbyte-integrations/connectors/destination-postgres/src/main/java/io/airbyte/integrations/destination/postgres/PostgresCli.java
```diff
@@ -0,0 +1,55 @@
...
+
+    final String configStr = args[0];
+    final List<Pair<String, String>> tables = new ArrayList<>();
+    for (int i = 1; i < args.length; i += 2){
+      tables.add(Pair.of(args[i], args[i + 1]));
+    }
```
jbfbell, line 25: again just gonna state the logic here
this loops through the args, skipping the config arg and shouldn't have an index out of bounds error because
a) we've insured the list is odd
b) you're incrementing by 2 so the last iteration is length - 2 or second to last (breaking out on length or last index + 1)

airbyte-integrations/connectors/destination-postgres/src/main/java/io/airbyte/integrations/destination/postgres/PostgresCli.java
```diff
@@ -0,0 +1,55 @@
...
+              System.out.println("Table does not exist: " + tableIdentifier);
+              System.exit(1);
+            }
+          }
+
+          System.out.println("All tables found to exist; dropping them now.");
+          // If they all exist, drop them
+          for (final Pair<String, String> tableIdentifier : tables) {
+            System.out.println("Attempting to drop " + tableIdentifier);
+            ops.dropTableIfExists(database, tableIdentifier.getLeft(), tableIdentifier.getRight());
```
jbfbell, line 50: non blocking, just curious `dropTablesIfExists` _sounds_ like it would check the existence for you? if it doesn't we should throw something in the backlog about renaming it

jbfbell (approved):

airbyte-integrations/connectors/destination-postgres/src/main/java/io/airbyte/integrations/destination/postgres/PostgresCli.java
edgao: yep, that's exactly the intent

airbyte-integrations/connectors/destination-postgres/src/main/java/io/airbyte/integrations/destination/postgres/PostgresCli.java
edgao: it does (or rather, it does a `drop table if exists`). Mostly I just want to make sure we only drop if the destination data is in the state that we expect, i.e. this is a (weak) safety measure

grishick (approved):

PR #22358, opened 2023-02-03
source-snowflake: use a safer method for parsing a BigInteger cursor value

rodireich:
## What
Setting a cursor value in the sql statement that is performed during incremental sync is unsafe for a cursor of bigint value which is given in a scientific notation (e.g "3E+1").
This closes [alpha-beta-issues/#738](https://github.com/airbytehq/alpha-beta-issues/issues/783)

## How
Make parsing of bigint string compatible with scientific as well as regular notations.

## 🚨 User Impact 🚨
This should fix failing syncs that run into bad values. no backward compatibility issues expected.


airbyte-db/db-lib/src/main/java/io/airbyte/db/jdbc/AbstractJdbcCompatibleSourceOperations.java
```diff
@@ -217,7 +217,7 @@ protected void setInteger(final PreparedStatement preparedStatement, final int p
   }
 
   protected void setBigInteger(final PreparedStatement preparedStatement, final int parameterIndex, final String value) throws SQLException {
-    preparedStatement.setLong(parameterIndex, Long.parseLong(value));
+    preparedStatement.setLong(parameterIndex, new BigDecimal(value).toBigInteger().longValue());
```
subodh1810, line 220: Since this change is in `AbstractJdbcCompatibleSourceOperations.java`, will it be safe to assume that its correct for all the sources that use this class?

rodireich: That's the assumption - that by sending it through BigDecimal it will be safer in case of incoming string in notation other than plain integer

rodireich:
CAT investigated in #23098

akashkulk (approved):
Thanks. Could you just link the original issue?

octavia-approvington:
Bernice doesnt think this is a good time </br>Approval bot is currently not enabled.  </br>Managers and staff engineers can still merge your code for security fixes and non-prod changes.

PR #22367, opened 2023-02-03
🪟 Fix vulnerability in new streams table cell tooltip code

### LINKED ISSUE #22196: Rewrite tooltip in Streams table code
Currently https://github.com/airbytehq/airbyte/pull/21253 introduced a XSS vulnerability that allows injecting bad code into the streams table in https://github.com/airbytehq/airbyte/blob/09421ce197337cb6469ff516f68a0acc2f593389/airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableCell.tsx#L30. Because we read the content of any `Text` node and set it's `innerHtml` via `dangerouslySetInnerHtml` to the tooltip after decodeUriComponents, you can inject code into the DOM, if e.g. a field is named `%3Cimg onerror='alert("should not happen")' src="reallynotgood" /%3E`, which will execute JS once a user hovers over the field.

We luckily don't have the new table design roled out to any user yet, thus this is not critical enough to need to be reverted, but this should be fixed asap still.

The correct implementation here would be to simply clone the original text node and attach it to the tooltip.

edmundito:
## What
Resolves #22196 

Please make sure that the text in the tablecell tooltip is safely rendered.

## How
Swaps the div with TextWithHTML, which securely render potentially dangerous text. Also, it does some trivial code cleanup.

### DIFF
--- a/airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableCell.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableCell.tsx
@@ -2,6 +2,7 @@ import classNames from "classnames";
 import React, { useCallback, useEffect, useRef, useState } from "react";
 import { useWindowSize } from "react-use";
 
+import { TextWithHTML } from "components/ui/TextWithHTML";
 import { Tooltip } from "components/ui/Tooltip";
 
 import styles from "./CatalogTreeTableCell.module.scss";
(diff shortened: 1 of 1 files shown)


krishnaglick:
I think [this](https://github.com/airbytehq/airbyte/pull/22647) PR handles this issue and more, and we can close this one.

edmundito:
Correct. This preceded that PR, so I will close.

PR #22369, opened 2023-02-03
🪟 🐛  Allow Free Connector Program syncs in workspace with no credits

joeykmh:
## What
Closes https://github.com/airbytehq/airbyte-internal-issues/issues/3829

Allows users to enable and manually sync connections irrespective of their workspaces `creditStatus` or `remainingCredits`. This avoids the issue where workspaces with 0 credits cannot start free syncs. This solution was discussed here: https://airbytehq-team.slack.com/archives/C049XDAS19C/p1675775965163619

## How
Instead of conditionally enabling pieces of the UI, this PR actually enables all manual sync buttons and "enable connection" switches. The backend will potentially reject these requests if the workspace credits do not suffice:

https://user-images.githubusercontent.com/7550957/217348791-fbf7bcb7-69e5-41bb-8e20-510a7fd9ed57.mov

I've also removed `FeatureItem.AllowSync`, since I removed the only usage of the feature.


airbyte-webapp/src/core/domain/catalog/traverseSchemaToField.test.ts
```diff
@@ -2,7 +2,7 @@ import { AirbyteJSONSchema } from "core/jsonSchema/types";
 
 import { traverseSchemaToField } from "./traverseSchemaToField";
 
-describe(`${traverseSchemaToField}`, () => {
+describe(`${traverseSchemaToField.name}`, () => {
```
josephkmh, line 5: Unrelated to this PR, but prevents the stringified function from being logged during unit tests 😄

airbyte-webapp/src/packages/cloud/locales/en.json
```diff
@@ -150,7 +150,7 @@
   "credits.emailVerification.resendConfirmation": "We send you a new verification link.",
   "credits.emailVerification.resend": "Send verification link again",
   "credits.lowBalance": "Your credit balance is low. You need to add more credits to prevent your syncs from being stopped.",
-  "credits.zeroBalance": "All your syncs have been stopped because your credit balance is 0. Add more credits to enable your syncs again.",
+  "credits.zeroBalance": "All your paid syncs have been stopped because your credit balance is 0. Add more credits to enable your paid syncs again.",
```
josephkmh, line 153: @andyjih any thoughts on this wording? I want to make it clear that free connector program syncs will continue.

image

airbyte-webapp/src/hooks/services/useConnectionHook.tsx
```diff
@@ -230,12 +230,23 @@ const useUpdateConnection = () => {
...
       updateConnection({ connectionId, status: enable ? ConnectionStatus.active : ConnectionStatus.inactive }),
     {
+      onError: (error: Error) => {
+        trackError(error);
+        registerNotification({
+          id: `tables.updateFailed.${error.message}`,
+          text: `${formatMessage({ id: "connection.updateFailed" })}: ${error.message}`,
+          type: ToastType.ERROR,
+        });
+      },
```
josephkmh, line 249: Displays an error toast if enabling/disabling the connection failed (e.g. due to credit balance)

https://user-images.githubusercontent.com/7550957/217348029-9af939eb-f6ba-41c7-931d-b6d959b7fb34.mov

airbyte-webapp/src/packages/cloud/locales/en.json
josephkmh: Discussed here: https://airbytehq-team.slack.com/archives/C049XDAS19C/p1675775965163619

airbyte-webapp/src/components/EntityTable/components/StatusCell.tsx
```diff
@@ -1,30 +1,30 @@
...
 
 export const StatusCell: React.FC<StatusCellProps> = ({
   enabled,
   isManual,
   id,
   isSyncing,
-  allowSync,
   schemaChange,
   hasBreakingChange,
+  connection,
```
tealjulia, line 27: Now that this is receiving/passing along the entire connection object we can probably weed out some of these other props I would think.

tealjulia: Also, it looks like these are a `ConnectionListItem` not `WebBackendConnectionRead`.  Maybe naming it `connectionListItem` would be more clear that we don't have an entire connection object here?

airbyte-webapp/src/components/EntityTable/components/StatusCellControl.tsx
```diff
@@ -57,7 +56,7 @@ export const StatusCellControl: React.FC<StatusCellControlProps> = ({
         <Switch
           checked={enabled}
           onChange={onSwitchChange}
-          disabled={!allowSync || hasBreakingChange}
```
tealjulia, line 60: This came from LaunchDarkly, right?  Was it currently targeting anyone?

airbyte-webapp/src/components/EntityTable/utils.tsx
```diff
@@ -98,6 +98,7 @@ export const getConnectionTableData = (
     lastSyncStatus: getConnectionSyncStatus(connection.status, connection.latestSyncJobStatus),
     connectorIcon: type === "destination" ? connection.source.icon : connection.destination.icon,
     entityIcon: type === "destination" ? connection.destination.icon : connection.source.icon,
+    connection,
```
tealjulia, line 101: Is this pulling out and renaming a bunch of the properties from each ConnectionListItem, and then also returning a nested copy of the entire original ConnectionListItem object? 

Should we just pull out the additional property(/ies) that we want?

teallarson (commented):
Testing locally, but had a few questions along the way.

airbyte-webapp/src/components/EntityTable/utils.tsx
josephkmh: Yeah so the issue is that we actually need (almost) the full `WebBackendConnectionListItem` (or `WebBackendConnectionRead`) to call `syncConnection(connection)` in `StatusCellControl.tsx`. I would rather not pull out and drill through the additional 6 properties, instead just pass through `connection`.

Agreed that it's weird that we are pulling out and renaming a bunch of properties here. I was also surprised to come across that. Probably the better option would be to just pass through the `connection` object and do all this renaming closer to the consumer. I did not want to go into that refactoring in this PR though!

airbyte-webapp/src/components/EntityTable/components/StatusCell.tsx
josephkmh: > Now that this is receiving/passing along the entire connection object we can probably weed out some of these other props I would think.

True, I think we can reduce quite a few of them! I would prefer to address this in a follow-up PR if you don't mind. Filed an issue here: https://github.com/airbytehq/airbyte/issues/22600

> Also, it looks like these are a ConnectionListItem not WebBackendConnectionRead. Maybe naming it connectionListItem would be more clear that we don't have an entire connection object here?

Hm, I didn't think about that. I generally rely on typescript to tell me what the actual underlying type is. Agreed that people probably would assume it's a `WebBackendConnectionRead` by the variable name, but I would say that both `WebBackendConnectionListItem` and `WebBackendConnectionRead` describe a connection.

teallarson (commented):
It looks like this is all the usages of `FeatureItem.AllowSync`.  Does that mean this should go away?: https://github.com/airbytehq/airbyte/blob/a6990bc1958c28fcc4718676f8fc5f9710cdb108/airbyte-webapp/src/packages/cloud/cloudRoutes.tsx#L48

josephkmh:
> It looks like this is all the usages of FeatureItem.AllowSync. Does that mean this should go away?:

That's a really good point, and yes! I think there is no reason to keep it around. This PR is the appropriate place to get rid of it too - I'll add that.

airbyte-webapp/src/packages/cloud/cloudRoutes.tsx
```diff
@@ -40,21 +37,7 @@ const CloudSettingsPage = React.lazy(() => import("./views/settings/CloudSetting
...
   const workspace = useCurrentWorkspace();
-  const cloudWorkspace = useGetCloudWorkspace(workspace.workspaceId);
-
-  useEffect(() => {
-    const outOfCredits =
-      cloudWorkspace.creditStatus === CreditStatus.NEGATIVE_BEYOND_GRACE_PERIOD ||
-      cloudWorkspace.creditStatus === CreditStatus.NEGATIVE_MAX_THRESHOLD;
-    // If the workspace is out of credits it doesn't allow creation of new connections
-    // or syncing existing connections.
-    setWorkspaceFeatures(outOfCredits ? ({ [FeatureItem.AllowSync]: false } as FeatureSet) : []);
```
ambirdsall, line 53: Enforcing the policy by dynamically toggling an app-wide feature setting feels like a really hacky use of the feature system anyway; not sad to see it go.

airbyte-webapp/src/packages/cloud/components/experiments/FreeConnectorProgram/lib/model.ts
```diff
@@ -0,0 +1,3 @@
+import { ReleaseStage } from "core/request/AirbyteClient";
+
+export const freeReleaseStages: ReleaseStage[] = ["alpha", "beta"];
```
ambirdsall, line 3: Extracting this from the free tag component file is 100% the right move, thanks. Extremely optional suggestion for the API:
```suggestion
export const isFreeReleaseStage = (stage: ReleaseStage): boolean => ["alpha", "beta"].includes(stage);
```

ambirdsall (commented):
:ship:

ambirdsall (approved):

teallarson (approved):
tested locally, lgtm

Issue #1338, opened 2020-12-15
Job Status not rendered properly.


cgardens:
## Tell us about the problem you're trying to solve
* When we added attempts to the UI some more values were added in the enum for JobStatus. I think they didn't get accounted for in the UI yet.

e.g. "pending 0"
![Screen Shot 2020-12-15 at 2 03 52 PM](https://user-images.githubusercontent.com/9092207/102278383-96499a80-3ede-11eb-941e-5fe5e7772515.png)

## Describe the solution you’d like
* these are the values that are currently in the enum. each of them should display in the UI.
```
  PENDING("pending"),
  
  RUNNING("running"),
  
  INCOMPLETE("incomplete"),
  
  FAILED("failed"),
  
  SUCCEEDED("succeeded"),
  
  CANCELLED("cancelled");
```

PR #22371, opened 2023-02-03
fix: Add check on Postgres source connection for missing privileges

sergio-ropero:
## What
Adding a new CDC connection with a user without the required privileges doesn't throw a error when creating it but the sync process will fail.
See: [#1338](https://github.com/airbytehq/oncall/issues/1338)

## How
Add a new condition for the `check` process so the connection is not created.

Missing tests


airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -314,6 +314,24 @@ public List<CheckedConsumer<JdbcDatabase, Exception>> getCheckOperations(final J
       checkOperations.add(database -> {
         PostgresUtils.checkFirstRecordWaitTime(config);
       });
+
+      checkOperations.add(database -> {
+        final String userName = config.get("username").toString();
+
+        final List<JsonNode> userPrivileges = database.queryJsons(connection -> {
+          final PreparedStatement ps = connection.prepareStatement("SELECT userepl FROM pg_user WHERE usename = ?");
```
subodh1810, line 322: can we query the `pg_user` table with the permissions that we ask for?

sergio-ropero: Yes we can.
In order to test this I just created a new user without any privilege (neither REPLICATE or table access), and the user is able to query it.
I think that by default, all users can query that table.

subodh1810 (approved):

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -314,6 +314,24 @@ public List<CheckedConsumer<JdbcDatabase, Exception>> getCheckOperations(final J
...
+          ps.setString(1, userName);
+          LOGGER.info("Verifying required privileges for user: {}", userName);
+          return ps;
+        }, sourceOperations::rowToJson);
+
+        if (!userPrivileges.get(0).get("userepl").asBoolean()) {
+          throw new RuntimeException(
```
akashkulk, line 329: Can we throw a ``ConfigErrorException`` here instead? That way, even if this failure occurs during ``read()`` or ``discover()`` the user will see the same message and it will be attributed towards a configuration error

sergio-ropero: Makes total sense! Thanks for pointing this :)

sergio-ropero: I will also modify the other ones in `getCheckOperations`, so we make sure all the errors are marked same way.
My doubt is... is really going to to throw `ConfigErrorException` when running `read` or `discover`? AFAIK, they are different commands and they are out of the `check` function (unless `check` is called always before)

akashkulk: It will protect against future changes to that codepath. 

In addition, currently for the Postgres connector ``check()`` is called from ``read()`` : https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java#L324

akashkulk: Thanks for fixing them for the other check operations!

airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/CdcPostgresSourceTest.java
```diff
@@ -166,6 +166,17 @@ private static DSLContext getDslContext(final JsonNode config) {
...
 
+  private void revokeReplicationPermission() {
+    executeQuery("ALTER USER " + container.getUsername() + " NOREPLICATION;");
+  }
+
+  @Test
+  void testCheckWithoutPrivileges() throws Exception {
+    revokeReplicationPermission();
+    final AirbyteConnectionStatus status = source.check(config);
+    assertEquals(status.getStatus(), AirbyteConnectionStatus.Status.FAILED);
```
akashkulk, line 177: super small nit: Can you also test the exact error message here as well?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
sergio-ropero: Awesome! Good to know, so at least we are checking in each sync if there has been any change in the user that doesn't allow to move forward with it (Thanks for looking for it ^^)

airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/CdcPostgresSourceTest.java
sergio-ropero: Done ✅

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -314,6 +317,22 @@ public List<CheckedConsumer<JdbcDatabase, Exception>> getCheckOperations(final J
       checkOperations.add(database -> {
         PostgresUtils.checkFirstRecordWaitTime(config);
       });
+
```
akashkulk, line 320: Small nit : Can you add a quick comment on what this is checking for in CDC?

akashkulk (approved):
Looks good!! Feel free to merge

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
akashkulk: Ideally...we should be removing this at some point, since the platform runs `check` before a sync. @subodh1810 had mention the check is required for CDC mode in the read as well (not sure exactly why) - but either way outside of the scope of this PR

sergio-ropero:

octavia-approvington:

octavia-approvington (approved):

adam-bloom:
FYI - I'm not entirely clear this works with AWS RDS Postgres (at least Aurora Postgres). We currently do not grant replication to the role that our airbyte user is in, but rather, grant it the `rds_replication` role. Replication works correctly, but it does not set the flags that this change checks for, as that role grants slightly fewer permissions.

subodh1810:
Hey @adam-bloom we realised what you metnioned and have changed the logic to not check for permissions but instead try to establish a replication connection to make sure that the user has all the right permissions. Please follow this PR https://github.com/airbytehq/airbyte/pull/23138

PR #22374, opened 2023-02-03
Fix naming of functions for pod sweeper

terencecho:
## What
pod sweeper was changed to sweeping job pods instead of worker pods through https://github.com/airbytehq/airbyte/pull/21829, but the method names were not changed.

## 🚨 User Impact 🚨
Change is only for code readability, no impact on behavior.


c-p-b (approved):

cpdeethree:
Thanks terence!

PR #21403, opened 2023-01-13
[ISSUE #19410] remove request_options_provider from the …

maxi297:
…manifest

## What
Solves #19410 

## How
Remove it from the declarative_component_schema.yaml, re-generate the files, update the factory by still using InterpolatedRequestOptionProvider and update the sources.

## Recommended reading order
1. airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml: remove from manifest
2. airbyte-cdk/python/airbyte_cdk/sources/declarative/config_component_schema.json: remove from json schema
3. airbyte-cdk/python/airbyte_cdk/sources/declarative/models/declarative_component_schema.py: auto-generated files
4. airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py: cleanup factory
5. all the sources

## 🚨 User Impact 🚨
This is a breaking change as described by #19410 

TLDR:
* source-datascope: The request_options_provider was written but at the incorrect location rendering it useless
* source-monday: This source uses something deprecated and we'll need to find out how we can migrate/test it

A `X` indicates that I validated InterpolatedRequestOptionsProvider is created with at least one parameter that isn't `None`. A `~` indicates that there was something special to be looked at for this connector which prevented me from performing the migration. The assumptions is that I could not test completely the API when `NO SECRETS`. Some other errors were documented as well which could affect the completeness of the tests

X source-alpha-vantage/source_alpha_vantage/alpha_vantage.yaml
X source-ashby/source_ashby/ashby.yaml
&nbsp;&nbsp;NO SECRETS
X source-braze/source_braze/braze.yaml
X source-breezometer/source_breezometer/breezometer.yaml
X source-callrail/source_callrail/callrail.yaml
X source-chartmogul/source_chartmogul/chartmogul.yaml
X source-close-com/source_close_com/close_com.yaml
X source-coin-api/source_coin_api/coin_api.yaml
X source-coingecko-coins/source_coingecko_coins/coingecko_coins.yaml
X source-coinmarketcap/source_coinmarketcap/coinmarketcap.yaml
X source-convertkit/source_convertkit/convertkit.yaml
X source-datascope/source_datascope/datascope.yaml
&nbsp;&nbsp;request_options_provider is under retriever and not requester. It was removed for now
X source-emailoctopus/source_emailoctopus/emailoctopus.yaml
X source-facebook-pages/source_facebook_pages/facebook_pages.yaml
X source-gnews/source_gnews/gnews.yaml
[...]

PR #22375, opened 2023-02-03
🪟 🔧 Connector Builder frontend fixes for low_code_cdk_to_beta

lmossman:
## What
This PR address the breaking frontend changes added to low_code_cdk_to_beta so far by fixing them in the connector builder frontend.

The only breaking changes merged into low_code_cdk_to_beta so far which had an effect on the frontend were the following:
https://github.com/airbytehq/airbyte/pull/21403 (removing `request_options_provider` and moving its children up to the `requester` level)
https://github.com/airbytehq/airbyte/pull/21823 (removing `url_base` from paginator)
https://github.com/airbytehq/airbyte/pull/21990 (renaming `field_pointer` to `field_path`)

## How
This PR fixes the above breaking changes by adjusting both the YAML -> UI and the UI -> YAML conversion logic to account for the changes.
One nice thing about how we've structured the code here is that those are the only places that need to be adjusted. We can leave the frontend form structure completely untouched for these smaller changes.

In order to properly **test** these changes, this requires several manual steps:
1. `cd` into the `airbyte/airbyte-connector-builder-server` directory and follow the steps in the airbyte-connector-builder-server's [Getting started section of the README](https://github.com/airbytehq/airbyte-platform/blob/main/airbyte-connector-builder-server/README.md#getting-started), but don't yet run the `uvicorn` command. 
3. Before running the uvicorn command, execute the following to pull in the local CDK code: `pip uninstall airbyte_cdk && pip install -e $AIRBYTE_ROOT/airbyte-cdk/python/` (replacing `$AIRBYTE_ROOT` with the path to your `airbyte` repo directory)
4. Run the uvicorn command in the readme (`uvicorn connector_builder.entrypoint:app --host 0.0.0.0 --port 8080`)
5. Apply the following diff in order to point the dev webapp to the locally running uvicorn connector builder server, rather than the dockerized one:
    ```
    diff --git a/airbyte-webapp/src/config/config.ts b/airbyte-webapp/src/config/config.ts
    index f0843d0c46c..0f9b238e44c 100644
    --- a/airbyte-webapp/src/config/config.ts
    +++ b/airbyte-webapp/src/config/config.ts
    @@ -7,6 +7,7 @@ export const config: AirbyteWebappConfig = {
       },
       apiUrl: window.API_URL ?? process.env.REACT_APP_API_URL ?? `http://${window.location.hostname}:8001/api`,
       connectorBuilderApiUrl:
    +    "http://0.0.0.0:8080" ??
         window.CONNECTOR_BUILDER_API_URL ??
         process.env.REACT_APP_CONNECTOR_BUILDER_API_URL ??
         `http://${window.location.hostname}:8003`,
    ```
6. Allow `Insecure content` in your browser for localhost:3000. Find instructions for this [here](https://experienceleague.adobe.com/docs/target/using/experiences/vec/troubleshoot-composer/mixed-content.html?lang=en). This is required because our `localhost:3000` runs on HTTPS, whereas uvicorn runs on HTTP, and by default chrome does not allow an HTTPS page to make requests to an HTTP url.
7. Open `https://localhost:3000/connector-builder` and try it out!

### DIFF
--- a/airbyte-connector-builder-server/setup.py
+++ b/airbyte-connector-builder-server/setup.py
@@ -41,7 +41,7 @@ setup(
     },
     packages=find_packages(exclude=("unit_tests", "integration_tests", "docs")),
     package_data={},
-    install_requires=["airbyte-cdk==0.25", "fastapi", "uvicorn"],
+    install_requires=["airbyte-cdk==0.27", "fastapi", "uvicorn"],
     python_requires=">=3.9.11",
     extras_require={
         "tests": [
--- a/airbyte-webapp/src/components/connectorBuilder/types.ts
+++ b/airbyte-webapp/src/components/connectorBuilder/types.ts
@@ -582,19 +582,16 @@ function builderStreamToDeclarativeSteam(
         url_base: values.global?.urlBase,
         path: stream.urlPath,
         http_method: stream.httpMethod,
-        request_options_provider: {
-          type: "InterpolatedRequestOptionsProvider",
-          request_parameters: Object.fromEntries(stream.requestOptions.requestParameters),
-          request_headers: Object.fromEntries(stream.requestOptions.requestHeaders),
-          request_body_json: Object.fromEntries(stream.requestOptions.requestBody),
-        },
+        request_parameters: Object.fromEntries(stream.requestOptions.requestParameters),
+        request_headers: Object.fromEntries(stream.requestOptions.requestHeaders),
+        request_body_json: Object.fromEntries(stream.requestOptions.requestBody),
         authenticator: builderAuthenticatorToManifest(values.global),
       },
       record_selector: {
         type: "RecordSelector",
         extractor: {
           type: "DpathExtractor",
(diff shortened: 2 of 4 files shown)


flash1293 (approved):
Tested and works as expected, thank you a lot for the detailed testing descriptions.

It seems like existing connectors are [adjusted already](https://github.com/airbytehq/airbyte/pull/21403), however for the `request_options_provider` thing I think one slipped through (FYI @maxi297): https://github.com/airbytehq/airbyte/blob/db98026cdc26529fc027bf7ff9b39435c2444037/airbyte-integrations/connectors/source-dremio/source_dremio/manifest.yaml#L10 I guess we need to take a pass over those anyway before merging in the main low code beta branch.

lmossman:
I talked with Maxime about the above - he is aware and confirmed that they will be rerunning the migrations before merging in the low code branch to catch cases like these of connectors that are younger than the branch

PR #22376, opened 2023-02-03
Download latest oss connector catalog on build and use as reference

c-p-b:
## What
This is part of a series of changes (#22653) to refactor platform to no longer reference the YAML files directly.

## How
This PR introduces two changes:
1. During gradle build, go get the latest connector catalog
2. Update `LocalDefinitionsProvider` to use the local oss_catalog instead of the YAML files

## Changes planed in subsequent PRS
- [x] Replace AirbyteGithubStore with a RemoteDefinition implementation ([PR #23055](https://github.com/airbytehq/airbyte/pull/23055))
- [ ]  Ensure we hard fail in cloud on remote definitions not being present
- [ ]  Update to use new subdomain based bucket urls instead of direct GCS


evantahler:
I *think* this PR should also remove this [class](https://github.com/airbytehq/airbyte/blob/master/airbyte-commons-server/src/main/java/io/airbyte/commons/server/services/AirbyteGithubStore.java#L31) and swap any uses of it with this new one?

Also, since I found myself searching for `raw.github` in our codebase, here are some other things that probably need to change:
* https://github.com/airbytehq/airbyte/blob/master/docs/reference/api/rapidoc-api-docs.html#L8
* https://github.com/airbytehq/airbyte/blob/master/docs/contributing-to-airbyte/code-style.md?plain=1#L8
* https://github.com/airbytehq/airbyte/blob/master/spotbugs-exclude-filter-file.xml#L5
* https://github.com/airbytehq/airbyte/blob/master/octavia-cli/README.md?plain=1#L93
* https://github.com/airbytehq/airbyte/blob/master/tools/openapi2jsonschema/examples/airbyte.local/openapi.yaml#L35

I think the Octavia CLI install process is the most interesting one...

airbyte-config/init/build.gradle
```diff
@@ -80,4 +90,5 @@ class IconValidationTask extends DefaultTask {
 }
 
 task validateIcons(type: IconValidationTask)
+check.dependsOn downloadCombinedConnectorCatalog
```
cpdeethree, line 93: TODO: for local development, maybe we want to do something else than running this every time before check

airbyte-config/init/src/main/java/io/airbyte/config/init/CombinedConnectorCatalogDownloader.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class CombinedConnectorCatalogDownloader {
+
+  public static void main(final String[] args) throws Exception {
+    final Path outputRoot = Path.of(args[0]);
+    final String outputFileName = args[1];
+
+    // TODO (ben) - get url from config
+    final RemoteDefinitionsProvider remoteDefinitionsProvider =
+        new RemoteDefinitionsProvider("https://storage.googleapis.com/prod-airbyte-cloud-connector-metadata-service/oss_catalog.json", 10000);
```
cpdeethree, line 20: TODO: Apply https://github.com/airbytehq/airbyte-platform-internal/pull/4449 so we can use the "better" url

airbyte-config/specs/build.gradle
```diff
@@ -44,6 +44,7 @@ task generateCombinedConnectorCatalog(type: JavaExec, dependsOn: generateSeedCon
     args 'oss_catalog.json'
 }
 
+// TODO (ben): we need to ensure this is called separately in deploy-oss-catalog.yml so as not to overwrite the catalog downloaded by cloud
```
bnchrch, line 47: @airbytehq/connector-operations Checking assumptions here

 Am I right in thinking that on the move of platform into platform-internal the below line which triggers `generateCombinedConnectorCatalog` when `processResources` is run should be removed? and then referenced explicitly in the deploy-oss-catalog?

airbyte-config/init/build.gradle
```diff
@@ -32,10 +32,18 @@ tasks.named("buildDockerImage") {
...
+    mainClass = 'io.airbyte.config.init.CombinedConnectorCatalogDownloader'
+}
+
+
 /**
  * Simple task that checks if all icons in the seed definition files exist as well as
  * that no icon in the icons folder is unused.
  */
+// TODO (ben): Refactor this in a future PR
 class IconValidationTask extends DefaultTask {
```
bnchrch, line 47: @airbytehq/connector-operations In preparation for the monorepo switch, do we know who is responsible for validation of icons? Its the OSS side correct?

bnchrch:
@airbytehq/connector-operations putting this up for review as I ensure everything is working appropriately.

airbyte-config/config-models/src/main/java/io/airbyte/config/CatalogDefinitionsConfig.java
```diff
@@ -0,0 +1,32 @@
...
+import java.nio.file.Path;
+
+public class CatalogDefinitionsConfig {
+
+  private static final String RESOURCE_DIRECTORY_PATH = "src/main/resources/";
+  private static final String LOCAL_CONNECTOR_CATALOG_FILE_NAME = "oss_catalog.json";
+  private static final String SEED_SUBDIRECTORY = "seed/";
+
+  private static final String REMOTE_OSS_CATALOG_URL =
+      "https://storage.googleapis.com/prod-airbyte-cloud-connector-metadata-service/oss_catalog.json";
```
evantahler, line 16: Do we have our pretty url yet @cpdeethree?

airbyte-config/init/build.gradle
evantahler: I think since (at the bottom) we've got `check.dependsOn validateIcons`, I believe that both the OSS and Cloud builds will run this task

airbyte-config/config-models/src/main/java/io/airbyte/config/CatalogDefinitionsConfig.java
cpdeethree: working on getting this in now! Should be ready before merge

airbyte-config/init/src/main/java/io/airbyte/config/init/LocalDefinitionsProvider.java
```diff
@@ -28,41 +21,48 @@
...
-        StandardDestinationDefinition.class);
+  public Map<UUID, StandardDestinationDefinition> getDestinationDefinitionsMap() {
+    final CombinedConnectorCatalog catalog = getLocalDefinitionCatalog();
+    return catalog.getDestinations().stream().collect(
+        Collectors.toMap(
+            StandardDestinationDefinition::getDestinationDefinitionId,
+            destination -> destination.withProtocolVersion(
+                AirbyteProtocolVersion.getWithDefault(
+                    destination.getSpec() != null
+                        ? destination.getSpec().getProtocolVersion()
```
jbfbell, line 58: The sextuple indent makes me think this might be worth breaking out into its own function. One stylistic nitpik you can feel free to ignore would be to use `Optional` again here
```
destination -> Optional.ofNullable(destination.getSpec()).map(spec::getProtocolVersion).orElse(null)
```

airbyte-config/init/src/main/java/io/airbyte/config/init/LocalDefinitionsProvider.java
```diff
@@ -28,41 +21,48 @@
...
 
-  public static final Class<?> DEFAULT_SEED_DEFINITION_RESOURCE_CLASS = SeedType.class;
-
-  private final static String PROTOCOL_VERSION = "protocol_version";
-  private final static String SPEC = "spec";
+  private static final String LOCAL_CONNECTOR_CATALOG_PATH = CatalogDefinitionsConfig.getLocalConnectorCatalogPath();
 
-  private Map<UUID, StandardSourceDefinition> sourceDefinitions;
-  private Map<UUID, StandardDestinationDefinition> destinationDefinitions;
+  public CombinedConnectorCatalog getLocalDefinitionCatalog() {
```
jbfbell, line 30: Honest question because I have no idea - Are these new methods covered by existing tests?

jbfbell (approved):

airbyte-config/init/src/main/java/io/airbyte/config/init/LocalDefinitionsProvider.java
bnchrch: Good question Joe! They are covered by the previous LocalDefinitionsProvider tests! 

The reason this is so seemless is the parsing logic was taken from RemoteDefinitionsProvider, and both LocalDefinitionsProvider and RemoteDefinitionsProvider implement the same interface

airbyte-config/config-models/src/main/java/io/airbyte/config/CatalogDefinitionsConfig.java
cpdeethree: For reference, here's the PR adding this. You can see how I've created the other environment url's too for later use e.g. dev-1-connectors.airbyte.com

airbyte-config/init/build.gradle
pedroslopez: If this is going to run on the platform side, we shouldn't reference any definition yamls since they will not be available. Should probably use the JSON catalog to get which icons we need to check.

This is in a weird spot right now because the images are bundled with the platform - this should obviously move to be a sole concern of the connectors side later. For now, I think it makes sense for this to run platform side.

airbyte-config/init/src/main/java/io/airbyte/config/init/LocalDefinitionsProvider.java
```diff
@@ -28,41 +21,48 @@
...
- * This provider contains all definitions according to the local yaml files.
+ * This provider contains all definitions according to the local catalog json files.
  */
 final public class LocalDefinitionsProvider implements DefinitionsProvider {
 
-  public static final Class<?> DEFAULT_SEED_DEFINITION_RESOURCE_CLASS = SeedType.class;
-
-  private final static String PROTOCOL_VERSION = "protocol_version";
-  private final static String SPEC = "spec";
+  private static final String LOCAL_CONNECTOR_CATALOG_PATH = CatalogDefinitionsConfig.getLocalConnectorCatalogPath();
```
pedroslopez, line 28: How would we potentially override this on the Cloud side to point to a different catalog? Should we keep a constructor param for overriding the path?

airbyte-config/specs/build.gradle
pedroslopez: Yes. We should not be generating connector catalogs on the platform side. Cloud currently does a similar thing: the catalog generation is its own task that is called directly and has nothing to do with the build.

pedroslopez (commented):
Overall looks good - but some questions + I think we need to do something about the icons

airbyte-config/init/build.gradle
bnchrch: Ok great.

Just refactored it into a task that uses the LocalProvider which references the local copy of oss_catalog from gcs

airbyte-config/init/src/main/java/io/airbyte/config/init/LocalDefinitionsProvider.java
bnchrch: I could move to an env variable to mirror the remote catalog override. Thoughts?

airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
```diff
@@ -47,6 +47,7 @@ public class EnvConfigs implements Configs {
   public static final String AIRBYTE_API_AUTH_HEADER_VALUE = "AIRBYTE_API_AUTH_HEADER_VALUE";
   public static final String WORKER_ENVIRONMENT = "WORKER_ENVIRONMENT";
   public static final String SPEC_CACHE_BUCKET = "SPEC_CACHE_BUCKET";
+  public static final String LOCAL_CATALOG_PATH = "LOCAL_CATALOG_PATH";
```
pedroslopez, line 50: nit: to make consistent with `REMOTE_CONNECTOR_CATALOG_URL` and not overload "catalog" too much

```suggestion
  public static final String LOCAL_CONNECTOR_CATALOG_PATH = "LOCAL_CONNECTOR_CATALOG_PATH";
```

bnchrch:
@evantahler @pedroslopez This is back up for review with the following changes
1. The `IconCheckTask` was refactored to java class that makes use of the `LocalDefinitionsProvider`
2. the `generateCatalog` task was removed as a dependency from `processResources` and now the `deploy-oss-catalog` github action calls it explicitly
3. You can now override the LocalDefinitionProvider path with an env variable. This should help when developing and testing changes to the catalog locally.

airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
```diff
@@ -344,6 +345,10 @@ public String getSpecCacheBucket() {
     return getEnvOrDefault(SPEC_CACHE_BUCKET, DEFAULT_SPEC_CACHE_BUCKET);
   }
 
+  public Optional<String> getLocalCatalogPath() {
+    return Optional.ofNullable(getEnv(LOCAL_CATALOG_PATH));
+  }
```
pedroslopez, line 350: Someone on the platform team can correct me, but I don't think we should be adding new things to EnvConfigs as it should be on its way out, replaced by the micronaut configuration.

pedroslopez (commented):
👍🏻 main concern is with using EnvConfigs, which is [on its way out](https://airbytehq-team.slack.com/archives/C04LENE16SY/p1675795054760429?thread_ts=1675795019.718409&cid=C04LENE16SY)

Note that the LocalDefinitionsProvider is used on cloud when deploying locally and in acceptance tests. We'll need to make sure the env var is set by default on cloud and pointing to the right spot.

airbyte-config/config-models/src/main/java/io/airbyte/config/EnvConfigs.java
bnchrch: so the reason I went this route is as I understand micronaut config is not available at the gradle task level. Was I wrong there?

bnchrch: Context here, after a conversation with Pedro, and some digging on my side we're going to leave this as EnvConfig until after the monorepo switch. Once complete there's likely some improvements to be done about how LocalDefinitions is loaded, which should open up the possibility of using micronauts value decorators

bnchrch:
@cpdeethree @evantahler @pedroslopez 
 
 Tests passed on the last build 🎉 
 
 Back up and ready for review.
 
 _p.s. I just made a variable rename so you'll notice the tests are running again_

airbyte-config/init/src/main/java/io/airbyte/config/init/CombinedConnectorCatalogDownloader.java
cpdeethree: Looks like there is a problem with tfstate, if we can't get that sorted we should just merge this API as-is with the understanding that we will bring in the new url shortly thereafter.

.github/workflows/deploy-oss-catalog.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           distribution: "zulu"
           java-version: "17"
       - name: Generate catalog
-        run: SUB_BUILD=PLATFORM ./gradlew :airbyte-config:init:processResources
+        run: SUB_BUILD=PLATFORM ./gradlew :airbyte-config:init:generateOssConnectorCatalog
```
cpdeethree, line 32: Much better name

airbyte-config/config-models/src/main/java/io/airbyte/config/CatalogDefinitionsConfig.java
cpdeethree: We should continue without the fancy url, there is a problem with the tfstate on that PR, we are trying to get it resolved but it shouldn't block this url.

c-p-b (commented):
@bnchrch LGTM

PR #22377, opened 2023-02-03
only compute diff if the schema discovery actually succeeded

mfsiega-airbyte:
## What
Prevent NPE when we try to diff schemas after unsuccessful schema discovery.

## How
Previously, we always tried to compute a diff. Now we only compute a diff if the discovery job was successful.

## Testing
Added unit test coverage for the unhappy path here.


jdpgrailsdev (approved):
:shipit:

alovew (approved):

Issue #21831, opened 2023-01-25
Tech Spec for mitigation of record growth in nested tables


grishick:
In incremental/dedup mode, nested tables can grow at a polynomial rate, which increases with each level of nesting and each subsequent run of normalization. 
Fixing deduplication of nested tables may be a bigger project, but to begin, we need to fix the bug that causes nested tables to grow so fast.

A couple of approaches have been discussed so far:
 - replace `>=` with `>` (see [here](https://github.com/airbytehq/airbyte/pull/7162#discussion_r734723202) a discussion of why we are using `>=` instead of `>`). This may not be safe, because `airbyte_emitted_at` is generated by source connectors and there is no guarantee that it will be unique per job
 - add another field to dedup on that is guaranteed to grow monotonically (e.g. job ID added by platform or destination connector)
 - move adding of `airbyte_emitted_at` from source to platform or destination

grishick:
Timeboxing this to 13 points

edgao:
spec is out for review; linking for posterity https://docs.google.com/document/d/1Ep6_jojbi5FdHEI-c1W_DrOCylmQJFeU1xvKM6oK-rc/edit#heading=h.fa72b39y2m99

PR #22381, opened 2023-02-04
Normalization: Use strict > comparison in incremental mode

### LINKED ISSUE #19207: Normalization creates duplicate rows in the properties table (Source Hubspot)
## Environment
- **Airbyte version**:  0.40.18
- **OS Version / Instance**: Ubuntu on AWS EC2
- **Deployment**: Docker
- **Source Connector and version**: Hubspot (v0.2.2)
- **Destination Connector and version**: Snowflake (v0.4.38)
- **Step where error happened**: Sync Job (Incremental | Deduped + history) 

## Current Behavior
Running the sync for the company stream, I get no duplicates in `companies` and `companies_scd` but I do get duplicate rows for multiple records in `companies_properties`. I've also tried resetting the data.
## Expected Behavior
Every record in companies should have just one record in `companies_properties` 
## Logs

Nothing is failing.
[e7d621ab_9b7c_45c6_b54f_b69ccfa777de_logs_1744_txt.txt](https://github.com/airbytehq/airbyte/files/9973103/e7d621ab_9b7c_45c6_b54f_b69ccfa777de_logs_1744_txt.txt)

## Steps to Reproduce
1. Create a connection from Hubspot to Snowflake
2. Set companies to incremental | deduped + history

Tried to look into the connector's code myself but I couldn't understand which part is related.

edgao:
## What
As described in https://github.com/airbytehq/airbyte/issues/21831, deeply-nested, sparsely-populated objects may see unexpectedly high amounts of duplicate records after normalization.

## How
Switch the incremental query to use `>` instead of `>=` on `_airbyte_emitted_at`. We're adding SATs to require sources to emit monotonically-increasing timestamps across syncs, which will make this safe to do (https://github.com/airbytehq/airbyte/issues/22240).

A new test will run three syncs, verifying the number of records in the destination DB after each sync. `messages.txt` is the first sync's messages, for consistency with our other normalization tests. This test case fails with the old `>=` behavior (in the second sync, since the first sync behaves identically).

## Recommended reading order

Destinations folks, please review: (normalization changes)

1. incremental.sql
2. `data_input/*` - this directory defines the two streams that we'll use for testing, along with what messages we'll push into the raw destination table across three syncs
    1. replace_identifiers.json is boilerplate for tests; it's useful for when different destinations have different output tablenames (e.g. different max tablename length, different restrictions on valid characters, etc)
4. `dbt_test_config` - for each sync, there's a `syncX_expectations` folder, which contains a file defining the expected number of records, and a `syncX_assertions` folder, which defines the assertions for those expected record counts. (the assertions are very simple)
5. `test_normalization.py` - this is the scaffolding to actually run the tests

## 🚨 User Impact 🚨
This should be fine, as long as the user has a well-behaved source connector.

## Pre-merge Checklist

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful


etsybaev (approved):

grishick (changes requested):
Let's practice feature-flagging and put this behind a feature flag as well as test what happens when this change gets rolled back.

airbyte-commons/src/main/java/io/airbyte/commons/features/FeatureFlagHelper.java
```diff
@@ -4,31 +4,40 @@
...
+                                           final UUID workspaceId) {
+    final String workspaceIdsString = flagRetriever.apply(featureFlags);
     final Set<UUID> workspaceIds = new HashSet<>();
     if (workspaceIdsString != null && !workspaceIdsString.isEmpty()) {
       for (final String id : workspaceIdsString.split(",")) {
         try {
           workspaceIds.add(UUID.fromString(id));
         } catch (final IllegalArgumentException e) {
-          log.warn("Malformed workspace id for field selection: {}", id);
+          log.warn("Malformed workspace id: {}", id);
```
gosusnp, line 36: Can we pass the context (field selection or strict comparison), I think that would be helpful if it came to debugging which var has an issue.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivity.java
```diff
@@ -12,14 +12,16 @@
...
 
 @ActivityInterface
 public interface NormalizationActivity {
 
   @ActivityMethod
   NormalizationSummary normalize(JobRunConfig jobRunConfig,
                                  IntegrationLauncherConfig destinationLauncherConfig,
-                                 NormalizationInput input);
+                                 NormalizationInput input,
+                                 final UUID workspaceId);
```
gosusnp, line 24: Can we add `workspaceId` to the `NormalizationInput` instead? This would line up better with how we are tracking labels for traces in the replication activity.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
```diff
@@ -62,13 +64,16 @@ public class NormalizationActivityImpl implements NormalizationActivity {
...
+  private final FeatureFlags featureFlags;
   private final Integer serverPort;
   private final AirbyteConfigValidator airbyteConfigValidator;
   private final TemporalUtils temporalUtils;
   private final ResourceRequirements normalizationResourceRequirements;
   private final AirbyteApiClient airbyteApiClient;
 
-  private final static Version MINIMAL_VERSION_FOR_DATATYPES_V1 = new Version("0.3.0");
+  private static final Version MINIMAL_VERSION_FOR_DATATYPES_V1 = new Version("0.3.0");
+  private static final String STRICT_COMPARISON_IMAGE_TAG = "0.4.0";
```
gosusnp, line 75: Explicitly tagging versions here feels a bit fragile. This feels like a short term fix, do we have another solution in mind?

gosusnp (commented):
@colesnodgrass, could you take a look at the feature flag usage, I want to make sure it plays well with what you were working on.

@edgao, I didn't look at the normalization part. A few comments about the form, on the change itself, let's chat, I think I am missing a bit of context here.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
gosusnp: Also, I think this value should be read from the config to be easier to override (ie not having to go through full build cycle to change the version here)

airbyte-commons/src/main/java/io/airbyte/commons/features/FeatureFlagHelper.java
edgao: done

edgao: I think https://github.com/airbytehq/airbyte/pull/22381/commits/35ba27e702083f97895c7b744177a427bd8a90c6 accomplishes this, hopefully I didn't miss something

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
edgao: can you say a bit more about the config? Not familiar with what you're referring to

gosusnp: The value here being hardcoded means that to iterate on the tag we want to use when the feature flag is active would require merging another change in OSS.
Having it as env variables allow to patch the config we use to deploy and avoid the slowest part of the dev cycle.

Generally, configs for the worker should go in https://github.com/airbytehq/airbyte/blob/master/airbyte-workers/src/main/resources/application.yml. The `${CONTAINER_ORCHESTRATOR_ENABLED:false}` notation is `${env_var:default}`.
This being a feature flag and the current constraints, it's probably fine to directly read the value from the featureflag environment.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivity.java
gosusnp: Thanks, do we have a test for `NormalizationInput.workspaceId` being null? This will happen for the running syncs at the time of the deploy, if not handled properly, we'll have to manually clean up temporal errors.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
```diff
@@ -156,28 +169,48 @@ public NormalizationSummary normalize(final JobRunConfig jobRunConfig,
...
     return new NormalizationInput()
         .withDestinationConfiguration(syncInput.getDestinationConfiguration())
         .withCatalog(syncOutput.getOutputCatalog())
-        .withResourceRequirements(normalizationResourceRequirements);
+        .withResourceRequirements(normalizationResourceRequirements)
+        .withWorkspaceId(syncInput.getWorkspaceId());
   }
 
   @VisibleForTesting
   static boolean normalizationSupportsV1DataTypes(final IntegrationLauncherConfig destinationLauncherConfig) {
```
gosusnp, line 192: Can you add a test for the version of the protocol we would pick with the tags related to this change? Let's make sure we are covered here and don't run into conflict with the recent changes until it gets cleaned up.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
edgao: added flags for the new version (STRICT_COMPARISON_NORMALIZATION_TAG) and the version from which it was branched (NON_STRICT_COMPARISON_NORMALIZATION_TAG).

what do you think is better behavior if someone pushes a newer version of normalization (e.g. 0.2.26) without also updating the strict comparison version?
* use 0.2.26 (silently deactivate feature flag)
* use strict_comparison (silently use an older version of normalization)
* fail the sync
* other?

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
edgao: done

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
gosusnp: My expectations in pseudo code would be:
```
if (shouldUseStrictComparisonForWorkspace(workspaceId)) {
  normalizationTag = FeatureFlag.getNormalizationTagForStrictComparision();
}
// the else clause should be implicitly keep using the version we defined in the connector catalog
// Same, the versioning logic should be applied on the tag without specific checks.
```

I think I'll look into cleaning up the datatypes v1 switch for the time being, so as long at we use data types v0 for anything that isn't in the `0.3.x` branch, we should be good.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivity.java
edgao: done

airbyte-commons/src/main/java/io/airbyte/commons/features/EnvVariableFeatureFlags.java
```diff
@@ -64,6 +68,21 @@ public String fieldSelectionWorkspaces() {
...
+    return getEnvOrDefault(STRICT_COMPARISON_NORMALIZATION_WORKSPACES, "", (arg) -> arg);
+  }
+
+  @Override
+  public String strictComparisonNormalizationTag() {
+    return getEnvOrDefault(STRICT_COMPARISON_NORMALIZATION_TAG, "strict_comparison", (arg) -> arg);
+  }
+
+  @Override
+  public String nonStrictComparisonNormalizationTag() {
```
gosusnp, line 82: Should be the last comment about env vars, I think this (+ the chain of constants and the env propagation) is no longer needed.

airbyte-integrations/bases/base-normalization/integration_tests/test_normalization.py
```diff
@@ -332,6 +389,31 @@ def setup_dbt_incremental_test(destination_type: DestinationType, test_resource_
     )
 
 
+def setup_dbt_sparse_nested_streams_test(destination_type: DestinationType, test_resource_name: str, test_root_dir: str, sync_number: int):
+    """
+    Prepare the data (copy) for the models for dbt test.
+    """
+    replace_identifiers = os.path.join("resources", test_resource_name, "data_input", "replace_identifiers.json")
```
grishick, line 396: since `replace_identifiers.json` does not have any replacements, you dan skip it altogether, because `copy_test_files` will ignore it if the file does not exist:
```
if os.path.exists(replace_identifiers):
...
```

grishick (approved):
LGTM. Conditional on CI checks and integration tests passing ;)

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
```diff
@@ -156,28 +167,49 @@ public NormalizationSummary normalize(final JobRunConfig jobRunConfig,
         () -> context);
   }
 
+  @VisibleForTesting
+  static void activateStrictNormalizationComparisonIfPossible(final IntegrationLauncherConfig destinationLauncherConfig,
+                                                              final FeatureFlags featureFlags) {
+    // Strict comparison was branched from normalization 0.2.25, so we shouldn't apply it if we're
+    // trying to use a newer version of normalization
+    if (featureFlags.nonStrictComparisonNormalizationTag().equals(getNormalizationImageTag(destinationLauncherConfig))) {
```
gosusnp: I would remove this check. I think the idea here is to have a feature flag to control which version to use. This checks looks like an extra hurdle that could trip us when setting the override. For example, a normalization bump would disable the feature flag override.

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
edgao: cool. pushed https://github.com/airbytehq/airbyte/pull/22381/commits/e470fd133a5902b3bdf8d2afc60736359860d60b, which I think accomplishes that.

gosusnp (commented):
From a platform pov, looks good to me if we remove the `NON_STRICT_COMPARISON_NORMALIZATION_TAG` code paths.

edgao:
> integration tests

the good news is that they pass locally!

the bad news is they only pass locally :( apparently CI is doing something weird - `distutils.errors.DistutilsFileError: cannot copy tree '../dbt-project-template': not a directory`

gosusnp (approved):

edgao:
publishing with run-tests=false because I have a successful test run , and the only diff since then is removing debug statements

edgao:
publishing with run-tests=false because I have a successful test run , and the only diff since then is removing debug statements

airbyte-integrations/bases/base-normalization/Dockerfile
```diff
@@ -30,5 +30,5 @@ ENTRYPOINT ["/airbyte/entrypoint.sh"]
 
 # 0.3.x is tombstoned.
 # The next minor bump should go directly to 0.4.0
-LABEL io.airbyte.version=0.2.25
+LABEL io.airbyte.version=strict_comparison
```
grishick, line 33: is the idea behind this label that when we are ready to roll this out to 100% of workspaces, we will change "strict_comparison" label to the next version (0.2.26 or whatever it happens to be) and remove the feature flags?

edgao: exactly - bump to new version + update destination_definitions for global rollout.

Issue #21830, opened 2023-01-25
connector builder FE should consume a versioned artifact of declarative_component_schema.yaml


cpdeethree:
## Tell us about the problem you're trying to solve
Currently the platform frontend build needs a file called declarative_component_schema.yaml to build. This lives in the connector sdk subfolder. This hard dependency should be updated to consume a pinned version of that schema instead.

## Describe the solution you’d like
A version of the file already gets published as part of the airbyte-cdk pypi package. We could simply update the frontend build to download a pinned version of that pypi package, extract the tar, and go grab the yaml file in the subfolder - rather than referencing the hardcoded file

## Describe the alternative you’ve considered or used
We also consider putting this file in the "shared" code that is shared between connectors and platform. However, this approach should be discouraged because we are in the middle of decoupling and removing as many hard dependencies as possible in shared code.

## Additional context
This was discovered as part of a larger effort of refactoring the split between airbyte platform and airbyte connectors. However, Alex G. noted it probably should be done under any circumstance because currently it's possible that the server can use a different version from the webapp.

timroes:
Cc @flash1293 @lmossman for information

flash1293:
Agree with the described solution! I took a quick look at this and it should be really simple to add the loading to the existing `generate-client` script in the webapp package.json loading it in a pre-determined location like this:
```
curl -L https://pypi.python.org/packages/source/a/airbyte-cdk/airbyte-cdk-${version}.tar.gz | tar -xzO airbyte-cdk-${version}/airbyte_cdk/sources/declarative/declarative_component_schema.yaml > build/lowcode_manifest.yaml
```

Ideally we also find a way to always keep this in sync with the version in the connector builder server: https://github.com/airbytehq/airbyte/blob/ca8cdc40aae629945e08d465871352235f5d53b5/airbyte-connector-builder-server/setup.py#L44

flash1293:
Grooming note:
* Make sure this version is in sync with what the connector builder server does

cpdeethree:
Thank you guys for the quick turnaround on this!

PR #22391, opened 2023-02-05
🪟🔧 Connector builder: Do not rely on local cdk anymore

flash1293:
## What

Related to #21830 

Downloads the CDK in the specified version as part of the frontend build to build typescript types instead of relying on CDK files in a folder as part of the same repo.

This is not fixing the linked issue as it's not ensuring connector builder server and webapp use the same version of the CDK, it's done to unblock the repo split in the short term.

## How

Do a curl request to fetch the tar ball of the cdk, then extract the declarative manifest schema from it. Also add a way to overwrite the used version and location of the manifest schema via environment variables to allow local development.


airbyte-webapp/scripts/load-declarative-schema.sh
```diff
@@ -0,0 +1,26 @@
+set -e
+mkdir -p build
+
+# Make sure this is aligned with the CDK version of the connector builder server
+DEFAULT_CDK_VERSION="0.25.0"
```
lmossman, line 5: Just confirming, is the plan to address keeping these aligned automatically in a follow-up PR, since in the description you said this PR does not fully close that issue?

Do you have an idea of how to solve that?

docs/contributing-to-airbyte/developing-locally.md
```diff
@@ -198,6 +198,15 @@ pnpm start
 
 - Happy Hacking!
 
+#### Using a custom version of the CDK declarative manifest schema for the connector builder UI
+
+When working on the connector builder UI and doing changes to the CDK and the webapp at the same time, you can start the dev server with `CDK_MANIFEST_PATH` or `CDK_VERSION` environment variables set to have the correct Typescript types built. If `CDK_VERSION` is set, it's loading the specified version of the CDK from pypi instead of the default one, if `CDK_MANIFEST_PATH` is set, it's copying the schema file locally.
+
+For example:
+```
+CDK_MANIFEST_PATH=../../airbyte/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_componenpnt_schema.yaml pnpm start
```
lmossman, line 207:
```suggestion
CDK_MANIFEST_PATH=../../airbyte/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml pnpm start
```

airbyte-webapp/scripts/load-declarative-schema.sh
```diff
@@ -0,0 +1,26 @@
+set -e
```
lmossman, line 1: Having a comment at the top of this script explaining why we have it may be helpful for anyone coming here from just seeing the script used in `generate-client` in package.json

lmossman (approved):
One small nit about adding a comment but up to you on if you think it is necessary to add. 

Otherwise this LGTM, tested locally and it works as expected

PR #22394, opened 2023-02-05
Changed docs to differentiate between docker compose standalone and plugin

JulianRommel:
## What & How
I changed the local deployment instructions to better differentiate between docker-compose (standalone version) and docker compose (plugin). The docs aren't concise on this currently. I also changed the link to the Docker Engine installation guide, as this includes the Docker Compose plugin. Changed "tested on Ubuntu 20.04" to "... Ubuntu 22.04" and fixed typo on line 54(thanks @juweins). 

## 🚨 User Impact 🚨
Should help users who haven't used docker before to find the correct version.

### DIFF
--- a/docs/deploying-airbyte/local-deployment.md
+++ b/docs/deploying-airbyte/local-deployment.md
@@ -1,13 +1,13 @@
 # Local Deployment
 
 :::info
-These instructions have been tested on MacOS, Windows 10 and Ubuntu 20.04.
+These instructions have been tested on MacOS, Windows 10 and Ubuntu 22.04.
 
 :::
 
 ## Setup & launch Airbyte
 
-- Install `Docker Engine` and the `Docker Compose plugin` on your workstation \(see [plugin instructions](https://docs.docker.com/engine/install/)\). Alternatively you can install the `docker-compose` standalone version \(see [standalone instructions](https://docs.docker.com/compose/install/other/)\).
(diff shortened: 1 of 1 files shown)


juweins (changes requested):
Hey @JulianRommel ! Thank you for your first Contribution! Welcome 🎉

I reviewed your PR and I have got two questions for you:
1. Regarding your proposal: Is there really a need for instructions on docker engine (standalone)?

I would say no, since this adds complexity to the quick start and may cause some confusion (docker compose vs. -) Furthermore, user of the standalone version will be aware of using the (deprecated) hyphen command.

2. I found a typo on line 54: there is a backslash \ following the WSL_ which causes it to display in the documentation. It would be great if you can include this in you PR.

3. Please make sure to sign the CLA, so your PR can be further processed.

Thoughts?

JulianRommel:
Hey @juweins and thank you so much for your feedback! You are probably right about the standalone version. I will change it and also include a fix for the typo on line 54.

juweins (commented):
Thanks for removing the odd character in line 54.

**From my side, this LGTM!** The changes are reasonable and provide value. @sh4sh 

BTW:
My previous comment should initiate a little discussion and not terminate your proposals. You can bring in your thoughts, too! 👍🏽

Make sure to sign the CLA from your previous commit. I don't know what happened here.

JulianRommel:
@juweins that's weird, I have signed the CLA but it seems to be buggy. Any ideas how to fix this? I already removed it and signed it again...

marcosmarxm (approved):
Thanks @JulianRommel

PR #22395, opened 2023-02-05
Connector acceptance test: Fix oneof check

flash1293:
Noticed a small issue in the connector acceptance tests - if `oneOf`s are nested, the test fails because it's trying to join a list made out of strings _and_ ints. This PR makes sure this works and also adjusts the types.


alafanechere (approved):
:shipit:

flash1293:

PR #22396, opened 2023-02-06
🪟🐛 Connector form: Fix squeezed status icon

flash1293:
If the space is tight, the status icon in the connector form can be shown in a squeezed way. This PR makes sure it's always shown properly by preventing it to shrink in a flex context:
Screenshot 2023-02-05 at 16 47 45
Screenshot 2023-02-05 at 16 48 54

I know this is changing a styled component without converting it and I started this work but then realized there's a bit more work to that because there are so many special cases and doing it right would require quite a bit of refactoring. I will create a separate issue for that and tackle it as a filler once I find the time


lmossman (approved):
LGTM, tested locally

PR #22397, opened 2023-02-06
Bmoric/fail connection if heartbeat failure

benmoriceau:
## What
This is checking than the source is pushing data or not. If the source haven't emitted any message within the configured acceptable delay, the connection will be failed.

The beat of the heartbeat monitor is done in the `DefaultAirbyteSource.java`. This allow to include the logs message as a message relevant for the timeout. Then another class `HeartbeatTimeoutChaperon` is use to monitor that the timeout is up-to-date. It will check every minute if the source is still beating which means that it has send a message in the last X seconds (X is a configurable value).
The heartbeat timeout is a wrapper of the execution. It will start 2 threads, one to run the timeout check, one to run the runnable provided to it. It will then wait until one of the thread terminate. If the timeout check thread is the first to finish, it means that the source is not responsive which will make the connection to fail with a dedicated failure type.

There is a feature flag which prevent making a connection to fail, if the flag is false, the chaperone will log that the timeout check failed won't return so the sync won't be failed. There is another PR that will add datadog monitoring when that happens.

## Recommended reading order
1. `HeartbeatTimeoutChaperone.java` => wrapper to check the heartbeat
2. `DefaultReplicationWorker.java` => Call the chaperone and create a custom error messages
3. `ReplicationJobOrchestrator.java` and `ReplicationActivityImpl.java` => Load the DB configured timeout value through the api
4. `FailureHelper.java` => Error message that will be displayed to the user.


airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -117,18 +119,20 @@ public DefaultReplicationWorker(final String jobId,
...
+                                  final boolean fieldSelectionEnabled,
+                                  final HeartbeatTimeoutChaperone srcHeartbeatTimeoutChaperone) {
     this.jobId = jobId;
     this.attempt = attempt;
     this.source = source;
     this.mapper = mapper;
     this.destination = destination;
     this.messageTracker = messageTracker;
-    this.executors = Executors.newFixedThreadPool(2);
+    this.executors = Executors.newFixedThreadPool(4);
```
gosusnp, line 130: Should we add a dedicated a threadpool in the HeartbeatTimeoutChaperone? Given that in this context we're looking at fixed expected number of long running threads, feels like injection the threadpool is more error prone.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
```diff
@@ -0,0 +1,130 @@
...
+
+  private boolean hasFailed = false;
+
+  public HeartbeatTimeoutChaperone(final HeartbeatMonitor heartbeatMonitor,
+                                   final Duration timeoutCheckDuration,
+                                   final FeatureFlagClient featureFlagClient,
+                                   final UUID workspaceId) {
+    this.timeoutCheckDuration = timeoutCheckDuration;
+
+    LOGGER.info("Starting source heartbeat check. Will check every {} minutes.", timeoutCheckDuration.toMinutes());
```
gosusnp, line 51: Wouldn't this log be more accurate in `runWithHeartbeatThread`?

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
```diff
@@ -0,0 +1,130 @@
...
+
+  public Runnable runWithHeartbeatThread(final Runnable runnable, final ExecutorService executorService) {
+    return () -> {
+      final CompletableFuture<Void> runnableFuture = CompletableFuture.runAsync(runnable, executorService);
+      final CompletableFuture<Void> heartbeatFuture = CompletableFuture.runAsync(this::monitor, executorService);
+
+      try {
+        CompletableFuture.anyOf(runnableFuture, heartbeatFuture).get();
+      } catch (final InterruptedException e) {
+        LOGGER.error("Heartbeat chaperone thread was interrupted.", e);
```
gosusnp, line 66: If runnableFuture end up throwing an interrupted exception, I think we'd end up logging the wrong information here.
Should we reuse the same logic as l80 instead?

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
```diff
@@ -0,0 +1,130 @@
...
+    };
+  }
+
+  @SuppressWarnings("BusyWait")
+  private void monitor() {
+    while (true) {
+      try {
+        sleep(timeoutCheckDuration.toMillis());
+      } catch (final InterruptedException e) {
+        LOGGER.info("Heartbeat thread has been interrupted (this is expected; the heartbeat was healthy the whole time).");
```
gosusnp, line 103: The message looks confusing to me, what about a simple `"stopping the heartbeat monitor"`?

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
```diff
@@ -0,0 +1,130 @@
...
+        return;
+      }
+
+      // if not beating, return. otherwise, if it is beating or heartbeat hasn't started, continue.
```
gosusnp, line 107: Is it worth moving the "heartbeat hasn't started" to another loop before this one? I think we may eventually have a config on how much time do we give to a connector before the first message.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
```diff
@@ -0,0 +1,130 @@
...
+      if (!heartbeatMonitor.isBeating().orElse(true)) {
+        if (!hasFailed) {
+          LOGGER.error("Source has stopped heart beating.");
+          if (featureFlagClient.enabled(ShouldFailSyncIfHeartbeatFailure.INSTANCE, new Workspace(workspaceId))) {
```
gosusnp, line 111: This looks a bit confusing, I'd expect the heartbeatMonitor to always fail the heartbeats fails. I'd have the feature flag in the caller (`runWithHeartbeatThread`?) to decide if we want to abort or not.

airbyte-commons-worker/src/test/java/io/airbyte/workers/internal/HeartBeatTimeoutChaperoneTest.java
```diff
@@ -0,0 +1,85 @@
...
+        }, executorService).run());
+  }
+
+  @Test
+  void testNotFailingHeartbeat() {
+    flags.put("heartbeat.failSync", true);
+    when(heartbeatMonitor.isBeating()).thenReturn(Optional.of(true));
+    assertDoesNotThrow(() -> heartbeatTimeoutChaperone.runWithHeartbeatThread(() -> {
+      try {
+        Thread.sleep(10);
```
gosusnp, line 65: Any chance we could mock the time here? having tests that relies on 10ms might end up being flaky.

airbyte-commons-worker/src/test/java/io/airbyte/workers/general/DefaultReplicationWorkerTest.java
```diff
@@ -166,7 +175,9 @@ void test() throws Exception {
         messageTracker,
         recordSchemaValidator,
         workerMetricReporter,
-        connectorConfigUpdater, false);
+        connectorConfigUpdater,
```
gosusnp, line 178: OPT: this constructor feels very redundant, is it worth abstracting to ease c/p?

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteSource.java
```diff
@@ -61,23 +60,19 @@ public class DefaultAirbyteSource implements AirbyteSource {
   private Integer exitValue = null;
   private final boolean featureFlagLogConnectorMsgs;
 
-  public DefaultAirbyteSource(final IntegrationLauncher integrationLauncher, final FeatureFlags featureFlags) {
-    this(integrationLauncher, new DefaultAirbyteStreamFactory(CONTAINER_LOG_MDC_BUILDER), new DefaultProtocolSerializer(), featureFlags);
+  public DefaultAirbyteSource(final IntegrationLauncher integrationLauncher,
```
gosusnp, line 63: It might be easier to leave a copy of the old constructor. With the monorepo, I think we'll end up duplicating the `DefaultAirbyteSource` for the connector tests and eventually reimplementing a specific one. This could limit the scope of this PR.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
```diff
@@ -0,0 +1,130 @@
...
+        throw new HeartbeatTimeoutException(String.format("Heartbeat has stopped. Heartbeat freshness threshold: %s Actual heartbeat age: %s",
+            heartbeatMonitor.getHeartbeatFreshnessThreshold(),
+            heartbeatMonitor.getTimeSinceLastBeat().orElse(null)));
+      }
+    };
+  }
+
+  @SuppressWarnings("BusyWait")
+  private void monitor() {
+    while (true) {
```
gosusnp, line 99: would having a flag here make sense simpler? The monitor thread would terminate normally, any exception thrown would always be an error.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -216,7 +220,7 @@ private void replicate(final Path jobRoot,
           });
 
       final CompletableFuture<?> readSrcAndWriteDstThread = CompletableFuture.runAsync(
-          readFromSrcAndWriteToDstRunnable(
+          srcHeartbeatTimeoutChaperone.runWithHeartbeatThread(readFromSrcAndWriteToDstRunnable(
```
gosusnp, line 223: I am wondering if we have a way around this extra nesting. In the current logic, we're injecting the heartbeat chaperone in the middle which adds another level of complexity. I am wondering if a slightly more flat structure would improve the readability here.

gosusnp (commented):
The change makes sense, but my first impression is that it is adding quite a lot of complexity to the code. I am wondering we have other alternatives, happy to chat.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
benmoriceau: Yes, that makes senses, done.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
benmoriceau: yes, done

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
benmoriceau: I'm donb't think so. An interruption exception doesn't mean that there was a timeout. I wonder if that might happen when we cancel a sync which is not a timeout error.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
benmoriceau: Done

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
benmoriceau: I am not sure if it worth it since now we're using the logs as a beat it seems to be unlikely to not have a log message. Could we test it first? I may add a datadog monitor for it if it helps.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
benmoriceau: the goal is to be able to monitor the potential forced failure. I just want to feature flag the action of making the connection to fail. The monitoring is added in another PR which is currently in draft: https://github.com/airbytehq/airbyte/pull/22584/files#diff-64e01615c3f046cc79b2df9bd1e55b088345ccb4ad23290d57aa466317d2ec2eR119 Does it explain why the feature flag is not on the call to monitor?

airbyte-commons-worker/src/test/java/io/airbyte/workers/internal/HeartBeatTimeoutChaperoneTest.java
benmoriceau: I am not sure about what could be a good strategy here. The issue is that the monitoring check is running on a different thread, I am just trying to have a main process that is running for "long enough".  I think that one solution would be to have a test implementation of monitor that will return on some condition. I'll try something.

benmoriceau: I have done something let me know if this looks better.

airbyte-commons-worker/src/test/java/io/airbyte/workers/general/DefaultReplicationWorkerTest.java
benmoriceau: Done

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteSource.java
benmoriceau: Ok, but it will need to have an heartbeatMonitor still. I'm not sure why keeping the old ctor will help/

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
benmoriceau: Is https://github.com/airbytehq/airbyte/pull/22397/files#r1107879305 answer that question?

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
benmoriceau: I am not sure how to change that to make it flatter.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
gosusnp: Got it

benmoriceau:
I've addressed most of the comments. On test is failing and I'm working on fixing it but the rest of the PR should be ok

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -238,6 +242,8 @@ private void replicate(final Path jobRoot,
                 replicationRunnableFailureRef.set(FailureHelper.sourceFailure(ex, Long.valueOf(jobId), attempt));
               } else if (ex.getCause() instanceof DestinationException) {
                 replicationRunnableFailureRef.set(FailureHelper.destinationFailure(ex, Long.valueOf(jobId), attempt));
+              } else if (ex.getCause() instanceof HeartbeatTimeoutChaperone.HeartbeatTimeoutException) {
```
jdpgrailsdev, line 245: Nit:  it feels like this if/else block is due to be refactored into a method that returns the FailureHelper.  It would allow us to write a test to verify the behavior by itself and explain what it is doing.

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/FailureHelper.java
```diff
@@ -113,6 +113,13 @@ public static FailureReason sourceFailure(final Throwable t, final Long jobId, f
         .withExternalMessage("Something went wrong within the source connector");
   }
 
+  public static FailureReason sourceHeartbeatFailure(final Throwable t, final Long jobId, final Integer attemptNumber) {
+    return connectorCommandFailure(t, jobId, attemptNumber, ConnectorCommand.READ)
+        .withFailureOrigin(FailureOrigin.SOURCE)
+        .withFailureType(FailureType.HEARTBEAT_TIMEOUT)
+        .withExternalMessage("The source was unresponsive");
```
jdpgrailsdev, line 120: Nit:  avoid passive tense -> use "is" instead of "was".

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
```diff
@@ -0,0 +1,149 @@
...
+                                   final UUID workspaceId) {
+    this.timeoutCheckDuration = timeoutCheckDuration;
+
+    this.heartbeatMonitor = heartbeatMonitor;
+    this.featureFlagClient = featureFlagClient;
+    this.workspaceId = workspaceId;
+    this.customMonitor = Optional.empty();
+  }
+
+  @VisibleForTesting
```
jdpgrailsdev, line 61: Nit:  This annotation can be removed, as the constructor is public.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
```diff
@@ -0,0 +1,149 @@
...
+        runnableFuture.cancel(true);
+      }
+
+      if (runnableFuture.isDone() && !heartbeatFuture.isDone()) {
+        heartbeatFuture.cancel(true);
+      }
+
+      if (heartbeatTimedOut) {
+        throw new HeartbeatTimeoutException(String.format("Heartbeat has stopped. Heartbeat freshness threshold: %s Actual heartbeat age: %s",
+            heartbeatMonitor.getHeartbeatFreshnessThreshold(),
```
jdpgrailsdev, line 109: I think that we will want some additional formatting on these `Duration` objects here, as the toString of duration uses the ISO-8601 notation (e.g. `PT8H6M12.345S`).  I'm not sure how readable that is to end users, especially non-technical ones.  I think it would probably make more sense to standardize on a unit of measure (seconds, perhaps) and just log the number of seconds.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
benmoriceau: Done

airbyte-commons-worker/src/main/java/io/airbyte/workers/helper/FailureHelper.java
benmoriceau: done

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
benmoriceau: I move the ctor to default scope instead.

airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/HeartbeatTimeoutChaperone.java
benmoriceau: Done (the exception creation has been moved)

benmoriceau:
@gosusnp I have updated the implementation based on what we discussed offline, I had to make some modification compared to what we discussed earlier.

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -215,8 +220,8 @@ private void replicate(final Path jobRoot,
             }
           });
 
-      final CompletableFuture<?> readSrcAndWriteDstThread = CompletableFuture.runAsync(
-          readFromSrcAndWriteToDstRunnable(
+      final CompletableFuture readSrcAndWriteDstThread =
+          srcHeartbeatTimeoutChaperone.runWithHeartbeatThread(CompletableFuture.runAsync(readFromSrcAndWriteToDstRunnable(
```
gosusnp, line 224: Is this the part that differed from what we discussed? If so, I am curious to understand why we still need to wrap the current future.

gosusnp (commented):
Thanks for the changes, just want to understand part about the future wrapping.

benmoriceau:
> Thanks for the changes, just want to understand part about the future wrapping.

@gosusnp I am in a the process of moving the PR. I'll copy paste your 2 comments in the new one and we can follow the discussion there.

Issue #20792, opened 2022-12-22
RequestOption should be split in two different structs


girarda:
## Current Behavior
RequestOptions are used to defined where to set a request parameter.

The struct has two fields:
```
    inject_into: RequestOptionType
    field_name: Optional[str] = None
```
Inject into defines how to update the request. It can either update a request param, a request header, body, or the path.

The field_name cannot be set when inject_into is path. This is done by making field_name optional and raising an exception if inject_into == path and field_name is set.

This is a problem because using a path and field_name seems like a valid manifest, but will cause a failure at runtime.

## Expected Behavior
- modifying the request path should be done using a separate struct.
- path should not be a valid RequestOptionType value

Example update schema (better name and description TBD)
```
  RequestModifier:
    anyOf:
      - "$ref": "#/definitions/RequestOption"
      - "$ref": "#/definitions/RequestPath"
  RequestOption:
    description: Describes an option to set on a request
    type: object
    required:
      - type
      - inject_into
      - field_name
    properties:
      type:
        type: string
        enum: [ RequestOption ]
      inject_into:
        enum:
          - request_parameter
          - header
          - body_data
          - body_json
      field_name:
        type: string
  RequestPath:
    description:
    type: object
    required:
      - type
      - path
    properties:
      type:
        type: string
        enum: [ RequestPath ]
      path:
        type: string
```

Issue #21095, opened 2023-01-06
[Low-Code CDK] Fix SubstreamSlicer parameter assignment and throw errors when RequestOption is path.


brianjlai:
## What is the problem

Originally identified from work on another PR here: https://github.com/airbytehq/airbyte/pull/20861#discussion_r1062971659

From the substream slicer, we currently allow for developers to attempt to inject values into the request path. This is quite misleading because under the hood we don't actually do anything with the path because stream slicers don't have access to the path. They are built onto the retriever whereas path is managed by requesters. We should just throw an error if they attempt to do that. We have no existing connectors that try to do this

We are also incorrectly assigning the stream_state_value for the slices being created by the substream slicer.

todo add an example.

## What is the desired solution

Throw errors when path is configured. We should be properly assigning to the correct key in the stream slice based on the parent stream config's request option (not the stream_state_value which is meant to indicate where to pull state data from a record)

## Acceptance Criteria

- An error should be thrown when a parent_stream_config's `request_option`, `inject_into` is set to path
- During `stream_slices()`, the resulting slice should store the `stream_state_value` under a key based on the `parent_stream_config.request_option.field_name`

PR #22398, opened 2023-02-06
[Low-Code CDK] Separate request path from RequestOption component

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/20792 https://github.com/airbytehq/airbyte/issues/21095

## What
This PR addresses a variety of the issues brought up in the linked tickets.

- The usability of the RequestOptions component is not very intuitive because request options are interspersed with request path injection so it is not explicitly clear what combinations will fail and run time. By splitting these components apart we can make the schema more clear and make it more obvious how to configure these components
- Fixes a bug with how the substream slicer injects request options

## How
This introduces a breaking change where any usage of the path RequestOption has been moved into its own dedicated component RequestPath. With this new component, we can tear out a lot of the internal validation used during component initialization because we can catch it ahead of time while parsing manifests into Pydantic models. This is reflected in how the schema has changed.

Fixes a small bug where we were accidentally injecting request options using the `stream_slice_field` instead of the request option on outbound requests. The only reason this wasn't caught was the only usage of this was on `source-braze` where both fields were the same.

I manually replaced `inject_into: path` with the component definition since there were not too many uses. And I did some manual replacement via pycharm to explicitly mark the `type: RequestOption` because a default value makes a lot less sense for a paginator where there are two options now. 

## Recommended reading order
1. `request_path.py`
2. `request_option.py`
3. `declarative_component_schema.yaml`
4. `default_paginatory.py`
5. `datetime_stream_slicer.py`
6. `substream_slicer.py`
7. `list_stream_slicer.py`


airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/request_option.py
```diff
@@ -25,21 +24,10 @@ class RequestOption:
...
         inject_into (RequestOptionType): Describes where in the HTTP request to inject the parameter
-        field_name (Optional[str]): Describes the name of the parameter to inject. None if option_type == path. Required otherwise.
     """
 
+    field_name: str
     inject_into: RequestOptionType
     parameters: InitVar[Mapping[str, Any]]
-    field_name: Optional[str] = None
-
-    def __post_init__(self, parameters: Mapping[str, Any]):
```
brianjlai, line 36: In the original implementation, only path RequestOption could skip `field_name`. Now that its been extracted, field_name is required and we can get rid of this entire validation

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/manifest_component_transformer.py
```diff
@@ -31,7 +31,6 @@
     # DefaultPaginator
     "DefaultPaginator.decoder": "JsonDecoder",
     "DefaultPaginator.page_size_option": "RequestOption",
-    "DefaultPaginator.page_token_option": "RequestOption",
```
brianjlai, line 34: getting rid of a default type because this could be a `RequestPath` now

airbyte-cdk/python/unit_tests/sources/declarative/requesters/paginators/test_request_option.py
```diff
@@ -9,15 +9,9 @@
 @pytest.mark.parametrize(
     "test_name, option_type, field_name, should_raise",
     [
-        ("test_limit_path_no_field_name", RequestOptionType.path, None, False),
```
clnoll, line 12: Looks like none of these `should_raise` anymore so the test body can be updated.

That said, it's not immediately obvious to me why the non-`.path` cases are being deleted.

clnoll (approved):
Nice @brianjlai! Just one question, not a blocker.

airbyte-cdk/python/unit_tests/sources/declarative/requesters/paginators/test_request_option.py
brianjlai: ah yes good point i'll clean this test up

PR #22400, opened 2023-02-06
CDK: improve `day_delta` macro and MinMaxDatetime component

grubberr:

## What

- `day_delta` - add 2-nd optional parameter `format`
- `MinMaxDatetime` - skip min or max `datetime` if result of evaluation is empty string


brianjlai (commented):
So the basic functionality is that when a provided datetime is earlier, then we set it to the boundary. When developing a connector, was there a limitation of the `MinMaxDatetime` component that prevented you from using instead of the macro? And if there was, maybe we could try to fold the functionality into the existing component instead of a new jinja macro

grubberr:
@brianjlai 

oops it seems I have missed that mix/max feature of datetime param
let me re-implement

grubberr:
@brianjlai 

I have removed that new macro `adjust_datetime` which I  introduced.
But I have changed current macro `day_delta` and changed current component `MinMaxDatetime` 
can you please look ?

airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/min_max_datetime.py
```diff
@@ -62,11 +62,15 @@ def get_datetime(self, config, **additional_options) -> dt.datetime:
         time = self._parser.parse(str(self.datetime.eval(config, **additional_options)), datetime_format, self.timezone)
 
         if self.min_datetime:
-            min_time = self._parser.parse(str(self.min_datetime.eval(config, **additional_options)), datetime_format, self.timezone)
-            time = max(time, min_time)
+            min_time = str(self.min_datetime.eval(config, **additional_options))
```
brianjlai, line 65: can you add tests to cover this new condition in `test_min_max_datetime.py`

brianjlai (approved):
changes look good! non-blocking change, but just a note to add some new tests for this case

brianjlai (approved):
thanks for adding the tests!

Issue #21702, opened 2023-01-23
Password indicator for (re)setting passwords


timroes:
We now have a way in Firebase to [enforce password policies](https://cloud.google.com/identity-platform/docs/password-policy#enable). We should leverage that feature, but first build a checker in the UI, that helps understand what needs to be part of the password. I'd suggest we enable lowercase, numeric, non alphanumeric character, a minimum length of 12.

I'd imagine the user sees an indicator for those below the password field when enter, like a todo list of requirements that need to be fulfilled:

```
✅ contains a lowercase character
❌ contains a number
❌ contains a special character
```

After we have released that we can enable the password policy in firebase on Prod, to make sure we're validating this server side as well.

We should test which error code we're getting from Firebase in case a password doesn't match the policy and make sure, that we also handle that error with a nicer message, in case our indicator has a gap in how it tests against firebase ever.

We can with that also remove the length checks we have in `GoogleAuthService.tsx`.

mfsiega-airbyte:
Refining notes:
- We need to clarify exactly what the password requirement is
- If we have a requirement, it's probably driven by some compliance thing so we should check with @git-phu if we have a deadline as well
- Possibly check with Nico for design input
- Loom on how to use this in Firebase: https://www.loom.com/share/3f0d7a0cc4e24fb7a40461ab655b8dab

PR #22402, opened 2023-02-06
🪟 🧪 [Experiment] Simplify signup left side

letiescanciano:
We want to experiment if simplifying the left side on the sign up page and dividing the signup methods, increase conversion of corporate emails on the page. How?
- Making Oauth the default method
- Email/password method is still available but user will see an error for personal emails. They can still sign up though.

Airtable details:
https://airtable.com/appIuY0uKPVnk8TWT/tbl2SxXnUwf6fVCWS/viw9frYvld7ks7aNo/recnFzE4HBB8RP1uY?blocks=hide
Demo: https://www.loom.com/share/9b706682d89845b1bf2455a1f3e1520d


airbyte-webapp/src/packages/cloud/views/auth/SignupPage/components/SignupForm.tsx
```diff
@@ -75,40 +81,78 @@ export const CompanyNameField: React.FC = () => {
...
+              <Text
+                size="sm"
+                className={classNames({
+                  [styles.error]: Boolean(meta.error) && meta.touched,
+                })}
+              >
+                <FormattedMessage id="signup.password.minLength" />
+              </Text>
+            </FlexContainer>
+          )}
```
timroes, line 154: We already have an issue out at https://github.com/airbytehq/airbyte/issues/21702 to improve that error handling. I think we don't need to add the error message improvements before that. But if we do so, it should imho not behind an experimentation flag, but just be added to the UI.

airbyte-webapp/src/packages/cloud/views/auth/SignupPage/components/SimpleLeftSide/check.svg
```diff
@@ -0,0 +1,9 @@
+<svg width="18" height="20" viewBox="0 0 18 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
timroes, line 1: :thinking: Is this icon used somewhere? I can't find any reference to this in your PR?

letiescanciano: Leftover sorry!

airbyte-webapp/src/packages/cloud/locales/en.json
```diff
@@ -158,6 +158,12 @@
   "firebase.auth.error.default": "Confirmation email cannot be sent. Please try again later.",
 
   "signup.password.minLength": "Password should be at least 12 characters",
+  "signup.details.noCreditCard": "No credit card required",
+  "signup.details.instantSetup": "Instant setup",
+  "signup.details.freeTrial": "14-day free trial",
+  "signup.title": "Create your Airbyte account",
+  "signup.method.email": "Sign up with email",
+  "signup.method.oauth": "Sign up with Google or Github",
```
timroes, line 166:
```suggestion
  "signup.method.email": "Sign up using email",
  "signup.method.oauth": "Sign up using Google or Github",
```

I think "sign up with x" sounds slightly off in English, but cc @nataliekwong for her opinion on this.

airbyte-webapp/src/packages/cloud/views/auth/SignupPage/components/SignupForm.tsx
```diff
@@ -75,40 +81,78 @@ export const CompanyNameField: React.FC = () => {
...
-          label={label || <FormattedMessage id="login.yourEmail" />}
-          placeholder={formatMessage({
-            id: "login.yourEmail.placeholder",
-          })}
-          type="text"
-          error={!!meta.error && meta.touched}
-          message={meta.touched && meta.error && formatMessage({ id: meta.error })}
-        />
+        <>
+          {" "}
```
timroes, line 100: Why do we need this space here?

airbyte-webapp/src/packages/cloud/views/auth/SignupPage/components/SignupForm.tsx
timroes: I think for things we clearly know are better (show users form validation) there's no purpose in putting them behind an experiment, and will just clutter the code and thus maintainability. But I won't block this PR on this opinion, but a more general thought for you and @nataliekwong to discuss moving forward.

airbyte-webapp/src/packages/cloud/views/auth/SignupPage/components/SimpleLeftSide/SimpleLeftSide.module.scss
```diff
@@ -0,0 +1,15 @@
...
+}
+
+.detailTextContainer {
+  min-width: fit-content;
+  color: colors.$dark-blue;
+}
+
+.detailsContainer {
+  overflow: visible;
+}
```
timroes, line 15: Trying to understand why we need this? I tried to remove it and I can't see any difference (no matter the width of the window). Could you clarify what use-case this is for?

airbyte-webapp/src/packages/cloud/views/auth/SignupPage/components/SimpleLeftSide/SimpleLeftSide.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const SimpleLeftSide: React.FC = () => {
+  const [showOauth, setShowOauth] = useState(true);
+  return (
+    <FlexContainer direction="column" gap="xl">
+      <HeadTitle titles={[{ id: "login.signup" }]} />
+      <Heading as="h1" centered>
+        <FormattedMessage id="signup.title" />
+      </Heading>
+
+      <FlexContainer className={styles.detailsContainer} alignItems="center">
```
timroes, line 33: I think we might want to have a justify center here as well, so the bullet points are centered above the buttons and not as currently left aligned and extending to the right.

airbyte-webapp/src/packages/cloud/views/auth/SignupPage/components/SimpleLeftSide/SimpleLeftSide.tsx
```diff
@@ -0,0 +1,64 @@
...
+const Detail: React.FC<React.PropsWithChildren<unknown>> = ({ children }) => {
+  return (
+    <FlexContainer gap="sm" alignItems="center" className={styles.detailTextContainer}>
+      <FontAwesomeIcon icon={faCheckCircle} className={styles.checkIcon} />
+      {children}
+    </FlexContainer>
+  );
+};
+export const SimpleLeftSide: React.FC = () => {
+  const [showOauth, setShowOauth] = useState(true);
```
timroes, line 25: :broom: Once we clean this up, it would be nicer to manage this via an URL route instead. Currently when switching between those the back button of your browser won't go back, since we only manage this state in JS, which already got me multiple times. I think for the final implementation this shoudl be separate routes, so your back button works as expected when switching between them.

timroes (changes requested):
Overall looks good. I've left some code cleanup and wording suggestions to discuss.

airbyte-webapp/src/packages/cloud/views/auth/SignupPage/components/SignupForm.tsx
letiescanciano: Formatter leftover

airbyte-webapp/src/packages/cloud/views/auth/SignupPage/components/SignupForm.tsx
letiescanciano: I agree on having things that we know are better for users shouldn't usually live under an experiment or FF, but in this case, my reasoning was:
- I have the FF anyway
- it's not going to live long in the code 
But I will remove it along with other changes.

airbyte-webapp/src/packages/cloud/views/auth/SignupPage/components/SimpleLeftSide/SimpleLeftSide.tsx
letiescanciano: I'm curious, if it's the same action (signup) but just two different flows, in which cases do you think it make more sense to have two routes?
I didn't think about the back button to be honest, as you can just use the elements in the page to navigate, but I'm talking from my ignorance in this matter!

airbyte-webapp/src/packages/cloud/views/auth/SignupPage/components/SimpleLeftSide/SimpleLeftSide.module.scss
letiescanciano: Sure! Details were being cutted off, but it seems that is working now, sorry!

timroes (approved):
Code LGTM, tested locally seems to work (despite the "navigating back" issue, which we can fix once we end the experiment).

PR #22407, opened 2023-02-06
Use shallow clone in the getting started example

kfirgollan:
## What
The AirByte repository got pretty big. This results in rather long clone time for getting started, which is mostly redundant. By using the `--depth` option the clone becomes much faster.

## How
Minor update in the main README file.
Add `--depth 1` for the clone example.

## 🚨 User Impact 🚨
In case of users without experience with git changing branches can be harder after clone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Explore our [demo app](https://demo.airbyte.io/).
 You can run Airbyte locally with Docker.
 
 ```bash
-git clone https://github.com/airbytehq/airbyte.git
+git clone --depth 1 https://github.com/airbytehq/airbyte.git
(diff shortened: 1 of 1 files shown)


juweins (approved):
Hey @kfirgollan ! Thank you for this addition to increase user experience! Sometimes, there is only little change needed for improvement!

Welcome to the Contributor of Airbyte! 🎉

Regarding this PR: LGTM!

Does it make sense to add a note clarifying why we use depth 1? Or does that speak for itself (From a new/juniors' perspective)

@sajarin it's your turn

marcosmarxm (approved):

PR #22408, opened 2023-02-06
migrate column selection flag helper to kotlin and add new exclude list

mfsiega-airbyte:
## What
Add a new column selection exclude list.

NOTE: for now, we have three (!) flags: a general flag (env variable), an explicit include-list (env variable), and an explicit exclude-list (launch darkly). The order of precedence is (1) include-list; (2) exclude-list; (3) general.

Once the exclude-list is deployed and in place, we'll rely solely on that and remove the other env-variable flags. 

## How
- Migrate the flag helper to the feature flag package (instead of commons).
- Add support to use the flag client (aka launch darkly in deployment).
- Expand the tests a bit.


mfsiega-airbyte:
A note here - this is using a kind of crazy setup with three backend flags (plus a frontend flag which isn't involved in this PR, and switches on users today).

An alternative - just start reading the existing frontend launchdarkly flag. Then add the workspaces we want to exclude as `False` targeting. This will also smooth the path for the frontend to start passing workspace context instead of users, which is the intention. Does this sound saner? @colesnodgrass @josephkmh

josephkmh:
The approach of reusing the existing frontend flag sounds saner to me, are there downsides? If it requires us to change how we consume the feature flags on the frontend, that could be a blocker. Our migration from user-based to workspace-based contexts will involve a major SDK version upgrade plus a rewrite of our implementation.

pmossman:
@mfsiega-airbyte going through older PRs, is this still worth reviewing/doing?

mfsiega-airbyte:
Nope! I'll close it.

PR #22409, opened 2023-02-06
🪟 🔧 Improve Output when overwriting experiments

timroes:
## What

Improve console output when experiments are overwritten:

* Makes sure it's only output once (in the main thread, not all the worker threads spawned by `vite-plugin-checker`).
* Styles the output a bit nicer:

screenshot-20230206-145137


lmossman (approved):
Changes look good, output looks clean from local testing and works as expected!

Issue #18267, opened 2022-10-20
First E2E test for connection streams table redesign


edmundito:
Create initial e2e test for new connection table during the connection create state. This is the first pass for supporting the new table in end to end testing

* Test should try to create a connection, set some values in the table, and then delete the table.
* Provide list of items that were not covered in the test so that it can be broken down into different Github issues.

### Tasks
```[tasklist]
- [x]  Add base test ( https://github.com/airbytehq/airbyte/pull/22412):
    - [x] Create Source and Destination via API
    - [x] Populate DB
    - [x] Implement flow: New connection => Existing Source => Existing destination => New Connections Stream table
    - [x] Check page routing(how app redirects and where we are after actions)
    - [x] Set Replication Frequency to `Manual`
    - [x] Set up connection - check API response (partially)
    - [x] Clean up: remove the connection, source, and destination
- [ ] Test specifically for Streams Table functionality (https://github.com/airbytehq/airbyte/pull/22582):
    - [x] Check Source and Destination connector icons
    - [x] Check table columns names
    - [x] Check how many rows are displayed(actual/expected amount). _Need to prepare appropriate test data_
    - [x] Check table filtering
    - [x] Check table scrolling
```

- [ ] Check table empty state (empty source/empty filter results) (**blocked** by https://github.com/airbytehq/airbyte/issues/22505) (TODO: Split into its own issue)
    

TODO: Split each root task into other github issues
- [ ] Test stream:
    - [ ] Turn off the sync stream (stream should have a red background and minus icon on the right)
    - [ ] _Stream select (will be covered in bulk edit tests)_
    - [ ] Check stream namespace, stream name, selected sync mode, destination namespace, destination stream name
    - [ ] Check the stream details panel opening by clicking on stream(row)
    - [ ] Stream should have all available sync modes
- [ ] Test stream sync modes(source and destination should support all available). Check that some sync modes have required fields(cursor and PK) :
    - [ ] Incremental Dedup + History: Courser defined and PK defined
    - [ ] Incremental Dedup + History: Select Cursor value(only one can be selected)  and PK defined
    - [ ] Incremental Dedup + History: Select Cursor value(only one can be selected)  and select PK (can be selected multiple) 
    - [ ] Incremental Dedup + History: Courser defined  and select PK (can be selected multiple)
    - [ ] Incremental + Append: cursor is required
    - [ ] Full refresh + Overwrite: cursor and PK are not supported
[...]

PR #22412, opened 2023-02-06
🪟 🚦 Add base E2E test for new stream table (part 1)

dizel852:
## What
Related to https://github.com/airbytehq/airbyte/issues/18267

## How
The test skips user flow: create source and destination connectors via UI.
We create them via API request and then start creating a new connection flow from choosing the existing ones.

Areas covered in the base test:
1. Selecting Source and Destination connectors from existing ones
2. Page routing (at what page we currently are after actions)
3. Changing replication frequency (only "Manual" options)
4. Connection set up - partial check API response(connection name and replication frequency)


airbyte-webapp-e2e-tests/cypress/pages/newConnection.ts
```diff
@@ -0,0 +1,25 @@
+type ConnectorType = "source" | "destination";
```
edmundito, line 1: File should be called `newConnectionPage`

airbyte-webapp-e2e-tests/cypress/integration/streamTable.spec.ts
```diff
@@ -0,0 +1,131 @@
+import { initialSetupCompleted } from "commands/workspaces";
```
edmundito, line 1: This file should go in `integration/connection/` (we can move the other files later)

airbyte-webapp/src/components/connection/ConnectionOnboarding/ConnectionOnboarding.tsx
```diff
@@ -241,7 +241,7 @@ export const ConnectionOnboarding: React.FC<ConnectionOnboardingProps> = ({ onCr
         </div>
       </div>
       <div className={styles.footer}>
-        <Button onClick={() => onCreate()} size="lg" data-id="new-connection">
+        <Button onClick={() => onCreate()} size="lg" data-id="new-connection" data-testid="new-connection-button">
```
edmundito, line 244: is `data-id` still important? What is it being used for currently?

edmundito (commented):
Looks good, just had a couple of nits and a question.

airbyte-webapp-e2e-tests/cypress/pages/newConnection.ts
dizel852: Renamed

airbyte-webapp/src/components/connection/ConnectionOnboarding/ConnectionOnboarding.tsx
dizel852: I've double-checked, seems like we don't use it at all, so I removed it.

airbyte-webapp-e2e-tests/cypress/integration/streamTable.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  waitForDiscoverSchemaRequest,
+  waitForGetSourceDefinitionsRequest,
+  waitForGetSourcesListRequest,
+} from "commands/interceptors";
+import { Connection, Destination, Source } from "commands/api/types";
+import { selectSchedule } from "pages/replicationPage";
+import { runDbQuery } from "commands/db/db";
+import { createUsersTableQuery, dropUsersTableQuery } from "commands/db/queries";
+
+// TODO: Disable before merge
```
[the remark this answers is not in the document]
dizel852: oh, probably because of "base" tests 😄

edmundito (approved):

PR #22415, opened 2023-02-06
Update pod-sweeper to allow multi-ns deploy

xpuska513:
## What
* Update pod-sweeper chart to allow namespace value override


charts/airbyte-pod-sweeper/values.yaml
```diff
@@ -224,4 +224,8 @@ timeToDeletePods:
   completed: 120
   error: 1440
 
-replicaCount: 1
\ No newline at end of file
+replicaCount: 1
+
+
+## namespace [string] namespace to deploy pod-sweeper at
+namespace: ""
```
Mykyta-Serbynevskyi, line 231: nit: newline

Mykyta-Serbynevskyi (approved):

PR #22422, opened 2023-02-06
Sync cdk manifest between webapp and connector builder server

### LINKED ISSUE #21830: connector builder FE should consume a versioned artifact of declarative_component_schema.yaml

flash1293:
Fixes https://github.com/airbytehq/airbyte/issues/21830 by making sure the connector builder server and the webapp always use the same CDK version.

This is done by a central file `CDK_VERSION` which is read by both the connector builder server setup and the webapp setup.

Also makes sure the webapp is built when the cdk version changes (adds a forgotten dependency on the openapi spec of the connector builder server as well)


lmossman (approved):
These changes make sense to me, tested locally and it works as expected 👍

PR #22433, opened 2023-02-06
Create Icon component

m19berger:
## What
Related to [#20133](https://github.com/airbytehq/airbyte/issues/20133)

## How
Created icon component

_In the future, the component should support properties `color` and `size`_

Note: Previous version of icons `CrossIcon`, `MoonIcon`, and `PauseIcon` as React Component supported the `<title>` tag, now it should be added separately

image

## Storybook
Currently, a demo version of the icon component is available in the storybook

image


airbyte-webapp/src/components/ui/Icon/types.ts
```diff
@@ -0,0 +1,58 @@
...
+export interface IconProps {
+  type: IconType;
+  className?: string;
+  height?: number | string;
+  width?: number | string;
+  // We need to consider the implementation of a mechanism with predefined colors and sizes
+  // https://github.com/airbytehq/airbyte/issues/20133
+}
+
+export const Icons: Record<IconType, React.FC<React.SVGProps<SVGSVGElement>>> = {
```
timroes, line 43: I'd suggest moving this to `Icon.tsx` instead. This is strictly spoken not really a type (so might be a bit misplaced in types.ts), also we don't really need this anywhere outside the `Icon.tsx` file.

airbyte-webapp/src/components/ui/Icon/index.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Icon } from "./Icon";
+
+export default {
+  title: "Ui/Icon",
+  component: Icon,
+} as ComponentMeta<typeof Icon>;
+
+const Template: ComponentStory<typeof Icon> = (args) => <Icon {...args} />;
+
+export const ArrowRightIcon = Template.bind({});
```
timroes, line 12: I think we should just have one story for `Icon` (instead of having an individual story per specific `type` property here), since that's really the component a dev would consume in their components, and just have a control in the story for the type, so it reflects more how a dev actually will use that component later on.

airbyte-webapp/src/components/ui/Icon/types.ts
```diff
@@ -0,0 +1,58 @@
...
+  | "play"
+  | "plus"
+  | "rotate";
+
+export interface IconProps {
+  type: IconType;
+  className?: string;
+  height?: number | string;
+  width?: number | string;
+  // We need to consider the implementation of a mechanism with predefined colors and sizes
```
timroes, line 39: I think we should at least build in the `size` property for now, instead of adding a `height`/`width`, that we'd then later once that icon is already used need to change in all consuming places. I think the color could potentially be part of a separate PR.

matter-q: @timroes, should it be something similar to other components, for instance, property `large` (height 25px/width 25px), `medium` (height 15px/width 15px)...?

timroes: Yes, I'd suggest we use the size abbreviations we use in other places as well (`sm`, `md`, `lg`, `xl`, ...). I don't have a clear idea what they might match to. @Upmitt @edmundito do you know if we have standardized icon sizes in our designs usually that we can match them to?

airbyte-webapp/src/components/ui/Icon/types.tsx
```diff
@@ -0,0 +1,58 @@
...
+export interface IconProps {
+  type: IconType;
+  className?: string;
+  height?: number | string;
+  width?: number | string;
+  // We need to consider the implementation of a mechanism with predefined colors and sizes
+  // https://github.com/airbytehq/airbyte/issues/20133
+}
+
+export const Icons: Record<IconType, React.FC<React.SVGProps<SVGSVGElement>>> = {
```
timroes, line 43: If we `as const` this object we should not require a check `Icons[type]` in `Icon.tsx`, since TS can guarantee at that point that we have for all icons an actual SVG in this object.

dizel852 (approved):
LGTM 👍 
Tested locally in storybook

airbyte-webapp/src/components/ui/Icon/types.ts
edmundito: We have not defined the icon sizes yet, will follow up about this but this should not block the PR.

krishnaglick:
@timroes Do we want to migrate this work? I can just bundle it into my color & style for icons PR.

timroes:
If you already working on the colors and styles, I'm happy to have both of them into one PR in airbyte-platform and closing this.

krishnaglick:
Closing, superceded by https://github.com/airbytehq/airbyte-platform/pull/111

PR #22464, opened 2023-02-07
remove unused test utils

colesnodgrass:
## What
- cleanup test-utils code
- remove unused code
- move code to be closer to usage

## How
- remove unused annotations
  - `@IntegrationTest`
  - `@SlowIntegrationTest` 
- rename pakage `io.airbyte.test.airbyte_test_container` -> `io.airbyte.test.container`
- remove unused `CockroachDBContainerHelper`
- remove `public` modifier from `GKEPostgresConfig`
- move `PostgresSQLContainerHelper` functions to `PostgresDestintationTest`
- replace `SchemaTableNamePair` class with `record`


airbyte-test-utils/src/main/java/io/airbyte/test/annotations/IntegrationTest.java
```diff
@@ -1,21 +0,0 @@
-/*
```
benmoriceau, line 1: We need to remove the related gradle task as well. 
https://github.com/airbytehq/airbyte/blob/master/build.gradle#L374
and https://github.com/airbytehq/airbyte/blob/master/build.gradle#L387
as well as the tags name: https://github.com/airbytehq/airbyte/blob/master/build.gradle#L341

benmoriceau (commented):
we should remove more (in the build.yml file)

benmoriceau (approved):

PR #22469, opened 2023-02-07
Additional Headers for RabbitMQ Destination connector

efimmatytsin:

Routing messages queues in a rabbit can be configured depending on the value in the argument (header). Also, the spring framework and many frameworks in java use headers to store service information, such as the name of the class for serialization / deserialization. Therefore, there was a desire to have a header configuration for sending messages to the queue.

Introduce a configurable parameter within the spec, logic to retrieve the user-configured parameter, additional headers stored like json object and appends to existing headers.

## Tests proof

image


juweins (commented):
Hey @efimmatytsin ! Thanks for this useful addition! A warm welcome from my side to the Airbyte Contriutors! 🎉

So far, your code addition makes sense to me. I could verify that the unit & integration tests are successful from my side. 

However, I just spin up the connector with your changes and a rabbitMQ server to make some smoke testing on different header types. I will come back on this later.

efimmatytsin:
Hello @juweins ! Thank you!

juweins (approved):
After testing with few different settings (apart from the included integration test), i got familiar with your addition. From my side: **LGTM!** 👍🏽

We should definitely add more integration tests to this connector, especially when it comes to AMQP header types. From my perception, this is the real usecase for your implementation.
However, I do not have experience with that. (yet)

Anyway, @marcosmarxm its your turn! 👀

image

efimmatytsin:
Hello! any updates?

PR #22471, opened 2023-02-07
Bmoric/right error for refresh

benmoriceau:
## What
Set a specific error if the refresh schema activity fails.

This PR set a meaningful error message, in order to avoid having an hard to understand root cause in case of a timeout error.

## File Order
- SyncWorkflowImpl (and test) => Catch all errors and return the right type.
- SyncOutputProvider => Create the syncoutput with an error.
- config.yml/failureReason.yml => add a new failure type specific to the refersh_schema. This might need to be in the failure origin instead. Happy to revisit it.


jdpgrailsdev (approved):
:shipit:

PR #22480, opened 2023-02-07
🪟 🎉 Show notification if Airbyte got updated

### LINKED ISSUE #19767: Show refresh notification when backend is outdated
We ever so often see users running into the following problem: they have their Airbyte webapp opened in the browser for very long time. We update the backend during that time. When they now try to navigate and load a JS/CSS chunk they haven't loaded beforehand this chunk no longer exists on the server, and loading the chunk will just fail, which usually just causes weird errors in the UI. We should show a toast notification to the user if the backend got updated and they need to refresh the page.

For that we could leverage the already existing `v1/health` call. This could return as part of it's response a custom hash for that server version, and we compare it against the one the frontend code still knows. If they differ show a toast notification asking to refrsh the page.

Why not send the version as part of the response. Sending out versions that the server is, is usually flagged during SOC2 audits as giving out too much information about the system. While we anyway have the version already in the FE, this would kind of not add any new information, but we should still try to not that route so we can later remove versions altogether from the (cloud) product. Any uniquely generated hash would be enough.

timroes:
## What

Closes https://github.com/airbytehq/airbyte/issues/19767

Show an update notification once the airbyte instance got updated and a user still has a webapp tab open.

screenshot-20230207-170141

## How

Every time Vite runs and loads our `build-info` plugin (i.e. on every build) we'll just generate a new random UUID and write this out to `/public/buildInfo.json` as well as making it available during the build as `process.env.BUILD_HASH`.

The build check than periodically every minute query that file without using a browser cache, and in case you kept the webpage open (and thus you still have the old `process.env.BUILD_HASH`, but the backend was updated, we now see a missmatch between those versions and show the notification. We also check every time the window gets refocused (which might happen if the browser tab was in the background and refocused), but make sure to throttle the overall calls to once every 10 seconds, which would be more than enough).

Why exactly those values? I don't have any really strong opinion about them, but I wanted to make sure the notification isn't coming "too late" after the user might already run into the error that some assets might not load for them anymore, since they're gone, and also wanted to make sure focusing works, so in case you had it in the background for very long (in which case some browser might stop your interval), you still get a pretty immediate check if it's still up to date.

I also added `data-testid` to all notification as part of that, so we can easier filter for those and create watched elements in fullstory.

## How to test this locally?

To test this locally, you can do the following:

```sh
pnpm build
npx http-server build/app
# Open the URL (usually localhost:8080) in your browser and keep it open
pnpm build
# Go back to the page and it should show that notification now after refocusing the tab or some interval
```


airbyte-webapp/src/hooks/services/Notification/NotificationService.tsx
```diff
@@ -35,6 +35,9 @@ export const NotificationService = React.memo(({ children }: { children: React.R
         <Toast
```
josephkmh, line 35: Not related to your changes, but the `.actionButton` style in `Toast` has a margin top that makes it unaligned: 
image

Can we remove that?

After:

image

airbyte-webapp/src/hooks/services/useBuildUpdateCheck.ts
```diff
@@ -0,0 +1,51 @@
...
+import { useIntl } from "react-intl";
+import { fromEvent, interval, merge, throttleTime } from "rxjs";
+
+import { useNotificationService } from "./Notification";
+
+interface BuildInfo {
+  build: string;
+}
+
+const INTERVAL = 60 * 1000;
```
josephkmh, line 11: I'm a fan of this syntax, but take it or leave it 😃 

```suggestion
const INTERVAL = 60_000;
```

airbyte-webapp/src/hooks/services/useBuildUpdateCheck.ts
```diff
@@ -0,0 +1,51 @@
...
+              id: "webapp-updated",
+              text: formatMessage({ id: "notifications.airbyteUpdated" }),
+              nonClosable: true,
+              actionBtnText: formatMessage({ id: "notifications.airbyteUpdated.reload" }),
+              onAction: () => window.location.reload(),
+            });
+          }
+        } catch (e) {
+          // We ignore all errors from the build update check, since it's an optional check to
+          // inform the user. We don't want to treat failed requests here as errors.
```
josephkmh, line 43: Not even worth logging to datadog? No downside IMO, and then we have some data if something is going wrong.

airbyte-webapp/src/hooks/services/useBuildUpdateCheck.ts
```diff
@@ -0,0 +1,51 @@
...
+  const { registerNotification } = useNotificationService();
+  useEffect(() => {
+    // Trigger the check every ${INTERVAL} milliseconds or whenever the window regains focus again
+    const subscription = merge(interval(INTERVAL), fromEvent(window, "focus"))
+      // Throttle it to maximum once every 10 seconds
+      .pipe(throttleTime(10 * 1000))
+      .subscribe(async () => {
+        try {
+          // Fetch the buildInfo.json file without using any browser cache
+          const buildInfoResp = await fetch("/buildInfo.json", { cache: "no-store" });
```
josephkmh, line 29: Just double checking - our nginx config doesn't have any server-side caching set up, right? This will be a route we'll want to avoid caching on the CDN if possible.

airbyte-webapp/src/components/ui/Toast/Toast.tsx
```diff
@@ -22,6 +22,7 @@ export interface ToastProps {
   onAction?: () => void;
   actionBtnText?: string;
   onClose?: () => void;
+  "data-testid"?: string;
```
josephkmh, line 25: AFAICT this isn't being tested, so what's the `data-testid` for?

joeykmh (approved):
Looks good, just left some non-blocking comments. Tested locally ✨

airbyte-webapp/src/hooks/services/useBuildUpdateCheck.ts
timroes: It has by default. In general I wanted to make this independant from whatever the infrastructure might have (miss)configured. So I thought forcing this here, makes more sense?

airbyte-webapp/src/hooks/services/useBuildUpdateCheck.ts
timroes: I thought logging this to datadog might just give us more errors there, that we don't really ever care about, but I'm also not against logging it, so will add it.

airbyte-webapp/src/hooks/services/useBuildUpdateCheck.ts
```diff
@@ -0,0 +1,55 @@
...
+export const useBuildUpdateCheck = () => {
+  const { formatMessage } = useIntl();
+  const { registerNotification } = useNotificationService();
+  const { trackError } = useAppMonitoringService();
+
+  useEffect(() => {
+    // Trigger the check every ${INTERVAL} milliseconds or whenever the window regains focus again
+    const subscription = merge(interval(INTERVAL), fromEvent(window, "focus"))
+      // Throttle it to maximum once every 10 seconds
+      .pipe(throttleTime(10 * 1000))
```
ambirdsall, line 28: Nice clean way to avoid bunching up requests if the focus event comes right before the next interval; I briefly wondered if some `switchMap` shenanigans would be needed to avoid that scenario (and whether it would actually be a problem, given our current scale and the CDN deployment discussions), but this is much better.

nit: I do think that it would be better to be consistent with how the time windows are defined; so `10 * 1000` should become `10_000` and either both or neither number should be extracted to a named binding (no strong preference which):
```ts
// either this...
const INTERVAL = 60_000;
const MINIMUM_WAIT_BEFORE_REFETCH = 10_000;

merge(interval(INTERVAL), fromEvent(window, "focus"))
  .pipe(throttleTime(MINIMUM_WAIT_BEFORE_REFETCH))

// ...or this
merge(interval(60_000), fromEvent(window, "focus"))
  .pipe(throttleTime(10_000))
```

ambirdsall (commented):
Looks good to me, only issue I saw was a trivial/non-blocking style thing. Testing locally now (insert xkcd about waiting for code to compile here); assuming everything is in order, I'll approve shortly

airbyte-webapp/src/hooks/services/useBuildUpdateCheck.ts
```diff
@@ -0,0 +1,55 @@
...
+      .subscribe(async () => {
+        try {
+          // Fetch the buildInfo.json file without using any browser cache
+          const buildInfoResp = await fetch("/buildInfo.json", { cache: "no-store" });
```
ambirdsall, line 32: nit: if you parse the response as json in a `then`, there's no need to bind any intermediate variable for the raw response. I'm not big on mixing async/await and then syntaxes randomly, but here I think it makes the intent clearer by reducing "fetch some json from a file" to a single step.
```ts
const buildInfo: BuildInfo = await fetch("/buildInfo.json", { cache: "no-store" }).then((res) => res.json());
```

ambirdsall (approved):
Tested as working, only trivial style suggestions: LGTM :rocket: :guitar: :fire: :sparkles:

PR #22500, opened 2023-02-07
Source S3: speed up discovery

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1470

## How
When `discover` is called, the connector reads all the files one by one and tries to infer the schema of each of them and then merge them into a single final schema. In case there are too many files, it takes much time. This PR introduces a threaded approach to reading the files.

It used to take the connector mentioned in the oncall issue about 1 second to read each of the 3.3k files they have in the bucket + 60-90 seconds to fetch the list of files. Now the whole `discover` takes about 6.5 minutes


airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/stream.py
```diff
@@ -52,6 +55,7 @@ def fileformatparser_map(self) -> Mapping[str, type]:
     ab_file_name_col = "_ab_source_file_url"
     airbyte_columns = [ab_additional_col, ab_last_mod_col, ab_file_name_col]
     datetime_format_string = "%Y-%m-%dT%H:%M:%S%z"
+    parallel_tasks_size = 256
```
davydov-d, line 58: 128 results in 6 min 50 seconds
256 results in 6 min 30 seconds
512 ~ almost the same

davydov-d:
:point_up: the failing test is not related, I'm working on it

bazarnov (approved):

PR #22511, opened 2023-02-07
Fix bumpversion config

### LINKED ISSUE #19674: Bumpversion blindly updates versions
We currently use a `bumpversion` script/config to run version updates across the codebase. What `bumpversion` does is though just finding a semver in the given files and set them to the new versions. Several files that we have `bumpversion` running over (https://github.com/airbytehq/airbyte/blob/master/.bumpversion.cfg) are though also containing other semvers than our own, e.g. we run it over the `package.json` of the webapp, which contains all the webapps dependencies.

I've validated, that if any of the dependencies we're using has the same version than Airbyte has, it would simply also set those versions up (which it should definitely not do). This problem is not just for the `package.json` but also other files in that list contain not just our version number.

That this hasn't crashed yet, is likely because we're already at a rather high non-stable minor (0.40.xx) which not a lot of tools have. As soon as we'd set our version to 1.0.0 (or also beforehand if we are unlicky and add a tool in our version range), we can't run OSS releases anymore, because the bumpversion script would break those files. We need to find a way to bump versions, that really only our versions in those files are updated and not every semver, that is the same as ours.

timroes:
## What

Fixes https://github.com/airbytehq/airbyte/issues/19674

Change the `bumpconfig` configuration file to not blindly update all versions that might occasionally be the same as ours in some files. This could have broken in case e.g. any dependency in the webapp would have had the same version that we have, on any OSS release.

I went through all the files and for all files I felt there is a danger we might update arbitrary other versions, that are in that file, I added a clear `search`/`replace` config that now only updates the version we actually want to update.

I've also removed the `package-lock.json`, since it doesn't exist anymore since our change to `pnpm` and the `pnpm-lock.yml` file doesn't have our own version printed in it (in contrast to the npm lock file).

You can test this locally by making sure to put the same version as our current one in each of the files with a `search`/`replace` and check that using the `bumpversion` tool will only bump the correct version. That's how I validated it locally.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -22,9 +22,9 @@ serialize =
 
 [bumpversion:file:airbyte-server/Dockerfile]
 
-[bumpversion:file:airbyte-webapp/package-lock.json]
-
 [bumpversion:file:airbyte-webapp/package.json]
+search = "version": "{current_version}"
+replace = "version": "{new_version}"
 
 [bumpversion:file:airbyte-workers/Dockerfile]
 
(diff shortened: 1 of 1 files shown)


supertopher (approved):
This change seems like it will make bump version less greedy.
That sounds great.

PR #22514, opened 2023-02-07
Add normalization strict incremental feature flag

gosusnp:
## What
* Add a feature flag to conditionally enable a different normalization tag for specific workspaces.

Extracted the backend changes from https://github.com/airbytehq/airbyte/pull/22381

## How
*Describe the solution*


benmoriceau (commented):
I think that we should at least add this feature flag in the feature flag module in addition of the old way. The old feature flag is deprecated.

benmoriceau (approved):

edgao (approved):

Issue #18187, opened 2022-10-19
[Epic] Connection streams table redesign


edmundito:
```[tasklist]
- [x] #19142
- [x] #19092
- [x] #18188
- [x] #18242
- [x] #18243 
- [x] #18186
- [x] #17823 
- [x] #17937 
- [x] #20372
- [x] #19918
- ~#19308~
- [x] #20196
- [x] #20450 
- [x] #20451
- [x] #21338
- [x] #20452
- [x] #20454
- [x] #20455
- [x] #20584
- [ ] #18267 
- [ ] https://github.com/airbytehq/airbyte/issues/20969
- [ ] https://github.com/airbytehq/airbyte/issues/22196
- [ ] https://github.com/airbytehq/airbyte/issues/22429
- [ ] #22624
- [ ] #22625 
- [ ] #22626 
- [ ] #21772
- [ ] https://github.com/airbytehq/airbyte/issues/24163
```

josephkmh:
@edmundito does it make sense to add this column selection issue to your epic as well? https://github.com/airbytehq/airbyte/issues/20584

edmundito:
It's live! Future improvements will be handled outside of the epic as individual enhancements or new epics.

PR #22524, opened 2023-02-07
🪟 🎨 Add soft edge to main view scroll content

edmundito:
## What
Related to #18187 

Based on a design discussion, the main scroll layout now adds a soft edge to the top of the scroll area so the content is not cut off drastically:

Screen Shot 2023-02-07 at 16 12 49

It also supports sticky table headers, and situations where the page is a table (for the connections, sources, and destinations list) where the soft edge can be disabled.

Demo:

https://user-images.githubusercontent.com/168664/217366982-c4561c1f-e95b-4524-bada-6f241bf55dda.mov

## How

The solution applies a floating div to achieve the effect. It sits between the margins and paddings of the header and the content of the `MainPageWithScroll` component. The component creates enough padding to hide the effect until the user starts scrolling.


teallarson (approved):
lgtm.  I am not recalling the specifics for this discussion -- do we want this on the bottom of the scrolling container as well?

edmundito:
@tealjulia this change is isolated from #21880 because it impacts all the pages so that we have a consistent top scroll style. The bottom will be updated in #21880 once this is merged.

tealjulia:
Got it!  I meant on pages like `/settings/sources` where there is a long scrolling container.  I don't know that it's needed and certainly isn't blocking.

Issue #21933, opened 2023-01-26
Make connector builder server build incremental


sherifnada:
## problem
`SUB_BUILD=PLATFORM ./gradlew :airbyte-connector-builder-server:build` does the same build steps every time. Even if no code changed, running that command twice runs the exact same build steps. Gradle easily supports incremental build steps. 

## Solution clues
Specifying incremental is most likely only a matter of specifying the input files. See [this link](https://docs.gradle.org/current/userguide/performance.html#enable_incremental_build_for_custom_tasks). 

## Acceptance Criteria
- [ ] Running `SUB_BUILD=PLATFORM ./gradlew :airbyte-connector-builder-server:build` twice in a row should not run the build twice if no input files changed. The second build should be fast/`UP-TO-DATE`
- [ ] If we modify a source file or a dependency (or anything "upstream" changes), the whole thing should rebuild.

maxi297:
Migrated to https://github.com/airbytehq/airbyte-platform/pull/110

maxi297:
I wanted to close the PR, not the issue. Oops!!!

maxi297:
The build hasn't been made fully incremental. However, partially incremental builds are 4s so it's good enough for now. See https://github.com/airbytehq/airbyte/pull/22548#issuecomment-1435118108 for more information

sherifnada:
@maxi297 I think we can close this issue for now. Once we move to the java server it should be incremental "by default"

sherifnada:
I didn't realize this PR was not merged, apologies. Reopening

PR #22548, opened 2023-02-08
Connector builder server: upgrade OpenAPI generator and make generateOpenApiPythonServer task incremental

sherifnada:
## What
Building towards #21933

* makes the OpenAPI generate task incremental 
* upgrades the OpenAPI generator to receive bugfixes that make this task incremental 
* As part of upgrading the OpenAPI generator plugin the format of the generated files changes. This PR commits those generated files. 

## How
*Describe the solution*

#### Why did we need to upgrade generator? 
Previously if you changed templates the task would be considered `UP-TO-DATE` which was a bug. I wasn't able to find where in the [OpenAPI generator changelog](https://github.com/OpenAPITools/openapi-generator/releases) this bugfix was applied exactly but I was able to confirm via manual testing that this upgrade fixes the issues we faced with incremental builds working incorrectly on this task (see appendix)

#### The issue that was happening before this PR
`generateOpenApiPythonServer` task was not running incrementally. When you turned on incremental builds for the task (by removing the `upToDateWhen{false}` block), it skipped running the task when it shouldn't have (which is likely why we added the `upToDateWhen{false}` block in the first place). 

To reproduce this:
1. use version 5.3.1 of the openAPI generator plugin i.e; `id "org.openapi.generator" version "5.3.1"`
1. Run `gw -i :airbyte-connector-builder-server:generateOpenApiPythonServer` and notice that task was not marked as `UP-TO-DATE`
2. run the same task again, it is marked as up-to-date
3. change the templates in `openapi/templates/api_interfaces.mustache` file in any way then re-run the task. Observe that it is _still_ being shown as `UP-TO-DATE`
4. change the OpenAPI spec in some way, then re-run the task. Observe that the task is no longer `UP-TO-DATE` and is properly being re-run now

Now upgrade the OpenAPI generator plugin to 6.3.0 and observe that none of the bugs above are happening. 

So it seems like 6.3.0 solves an issue around incremental builds.


airbyte-connector-builder-server/build.gradle
```diff
@@ -11,17 +11,15 @@ airbytePythonDocker {
...
+    // TODO this openapi file doesn't need to be nested in 3 empty folders
     inputSpec = "$rootDir.absolutePath/airbyte-connector-builder-server/src/main/openapi/openapi.yaml"
-    outputDir = generatedCodeDir
-
     generatorName = "python-fastapi"
     configFile = "$projectDir/openapi/generator_config.yaml"
     templateDir = "$projectDir/openapi/templates"
     packageName = "connector_builder.generated"
 
+    outputDir = generatedCodeDir
```
sherifnada, line 22: moved here from line 18, doesn't actually change behavior

airbyte-connector-builder-server/build.gradle
```diff
@@ -11,17 +11,15 @@ airbytePythonDocker {
 }
 
 task generateOpenApiPythonServer(type: GenerateTask) {
-    outputs.upToDateWhen { false }
```
sherifnada, line 14: this line previously turned off incrementality for this task

buildSrc/src/main/groovy/airbyte-docker.gradle
```diff
@@ -174,6 +174,7 @@ class AirbyteDockerPlugin implements Plugin<Project> {
 
     static def createTask(Project project, String taskName, String dockerFile) {
         if (project.file(dockerFile).exists()) {
+            // TODO -- why does this exclude venv files?
```
maxi297, line 177: Not an expert and I don't have a lot of context but this seems to assume that `.venv` is never copied by the Dockerfile and that there will be an `pip install` command in the Dockerfile to make the virtual environment available in the docker image/container (which seems like an assumption that isn't too far-fetched). 

What I'm missing is that we've done this assumption in order to solve a problem and I don't get what the problem was. My best guess is performance but I doubt this would be very costly not to ignore .venv

buildSrc/src/main/groovy/airbyte-python-docker.gradle
```diff
@@ -5,30 +5,25 @@ import org.gradle.api.tasks.Exec
...
-        def extension = project.extensions.create('airbytePythonDocker', AirbytePythonDockerConfiguration)
 
         project.task('airbytePythonDockerApply', type: Exec) {
             /*
             Install the dependencies and run the tests from a docker container
              */
-            commandLine "docker", "run", "-v", "${project.projectDir.getAbsolutePath()}:/home", "--entrypoint", "/bin/bash", "python:3.9-slim", "-c", "chmod +x /home/run_tests.sh && /home/run_tests.sh /home"
-        }
 
-        project.task('blackFormat', type: Exec) {
```
maxi297, line 27: Removed because not used

buildSrc/src/main/groovy/airbyte-python-docker.gradle
```diff
@@ -5,30 +5,25 @@ import org.gradle.api.tasks.Exec
...
-            /*
-            Run formatter and static analysis from a docker container
-             */
-            commandLine "docker", "run", "-v", "${project.projectDir.getAbsolutePath()}:/home", "--entrypoint", "/bin/bash", "python:3.9-slim", "-c", "chmod +x /home/run_tests.sh && /home/run_format.sh /home"
+            // This doesn't take into account dependencies if you don't have a lock file. We strongly recommend using a
+            // lock file in other to ensure that given no file changes and two "pip install", you have the same output
+            inputs.dir project.projectDir.getAbsolutePath()
+            outputs.cacheIf { true }
```
maxi297, line 23: To be confirmed if needed

maxi297: Based on https://docs.gradle.com/enterprise/tutorials/caching/#introduction_build_avoidance, it seems like it could still be useful as "build avoidance" even though there are no outputs. If locally or remotely a cache entry has the same input, it would avoid us having to re-run the tests

airbyte-connector-builder-server/requirements.txt
```diff
@@ -0,0 +1,109 @@
+#
```
sherifnada, line 1: having this lockfile is awesome

sherifnada (commented):
This build is not being run incrementally. If i run the build twice (` SUB_BUILD=PLATFORM gw :airbyte-connector-builder-server:build`) I don't see an UP-TO-DATE on the task. The build runs twice (specifically `airbyteDockerPythonApply` and `airbyteDocker` tasks are taking a long time)

Do you know what's the problem there?

maxi297:
>  If i run the build twice ( SUB_BUILD=PLATFORM gw :airbyte-connector-builder-server:build) I don't see an UP-TO-DATE on the task

I was calling `airbyteDockerPythonApply` directly hence the difference. Basically, build generates .pyc files with . I added exclusions on byte-compiled, distribution and packaging files. Thanks for pointing that out! 

> and airbyteDocker tasks are taking a long time
This has been identified in https://docs.google.com/document/d/1PClBeVbdcbhIIUKKNf7Tkg4y4rM8eLnaqODiN8cZWd4/edit# as a possible improvement. It seems like the logic https://github.com/airbytehq/airbyte/blob/master/buildSrc/src/main/groovy/airbyte-docker.gradle#L133 always return false. I'll investigate the specific case locally for this story (and maybe timebox just in case)

maxi297:
> and airbyteDocker tasks are taking a long time

I’ll timebox my investigation here. The best lead I have so far is that there are two tasks building docker: :airbyte-connector-builder-server:buildDockerImage and :airbyte-connector-builder-server:airbyteDocker. However, they seem to generate slightly different outputs (in the picture, airbyteDocker is on the left and buildDockerImage is on the right). The outcome is that none of these will ever be up-to-date because they don’t have the same SHA and therefore the same output. The difference in terms of output seems to be caused by buildkit

maxi297:
Migrated to https://github.com/airbytehq/airbyte-platform/pull/110

PR #22554, opened 2023-02-08
🪟 🧹 Remove experimentation flags for OAuth

timroes:
## What

Remove the OAuth experimentation flags from the code. We have this fully rolled out by now, and we can't even use them as a killswitch really, because users wouldn't be able to log in properly anymore, so talked to Natalie, and we decided to remove those flags fully.


joeykmh (approved):
LGMT, did not test locally

PR #22556, opened 2023-02-08
🪟 🎨 Use switches instead of checkboxes for column selection

### LINKED ISSUE #22429: New table should use switches instead of checkboxes
Currently the new table design uses checkboxes for allowing users to enable/disable which fields to sync:

screenshot-20230206-214536

We should switch them to Switches (no pun intended), since we use checkbox in the connection page otherwise for bulk edit, and using the very same pattern suddenly on the same page for enableing/disabling (which we do on a stream level with switches) wouldn't be great usability.

At this point we could potentially also remove the word "Sync" from the table header, which is currently causing a bit weird spacing between the checkbox and the title, that looks like there might be some content that could go there (while there isn't; it's just because of the word "Sync" needing more space in the header).

joeykmh:
## What
Closes https://github.com/airbytehq/airbyte/issues/22429

Changes the column selection UI in the new stream table from using checkboxes to using switches:

image


airbyte-webapp/src/components/ui/Text/Text.tsx
```diff
@@ -45,9 +48,10 @@ export const Text: React.FC<React.PropsWithChildren<TextProps>> = React.memo(
     className: classNameProp,
     size = "md",
     inverseColor = false,
+    color = "default",
```
josephkmh, line 51: I think we need to improve our Text component by accepting a `color` prop: https://github.com/airbytehq/airbyte/issues/22557

I'd follow up further refactoring in a separate PR

airbyte-webapp/src/locales/en.json
```diff
@@ -226,7 +226,8 @@
   "form.selected": "Selected",
   "form.notSelected": "Not selected",
   "form.nullable": "Nullable",
-  "form.field.sync": "Sync",
+  "form.stream.sync": "Sync Stream",
```
edmundito, line 229:
```suggestion
  "form.stream.sync": "Sync stream",
```

airbyte-webapp/src/locales/en.json
```diff
@@ -226,7 +226,8 @@
...
+  "form.field.sync": "Sync field",
```
edmundito, line 230: Is this value still being used?

airbyte-webapp/src/components/connection/CatalogTree/next/StreamDetailsPanel/StreamFieldsTable/StreamFieldsTable.tsx
```diff
@@ -118,26 +118,29 @@ export const StreamFieldsTable: React.FC<StreamFieldsTableProps> = ({
...
-              header: () => (
-                <FlexContainer gap="md">
-                  <CheckBox
-                    checkboxSize="sm"
-                    indeterminate={config?.fieldSelectionEnabled && !!config?.selectedFields?.length}
-                    checked={!config?.fieldSelectionEnabled}
-                    onChange={toggleAllFieldsSelected}
-                    disabled={mode === "readonly"}
-                  />
-                  <FormattedMessage id="form.field.sync" />
```
edmundito, line 134: Looks like this string is no longer needed?

edmundito (commented):
Overall it looks good.

I noticed that the switch does not animate when I enable and disable the column. Would you know why that is? Do you see the same thing?

teallarson (approved):
Tested this locally with the feature flag both on and off (since some of the layout was adjusted).  Lgtm.  Left a few small comments.  It also looks like the animation may not be happening on the Switch?  I don't think any of it is blocking.

airbyte-webapp/src/components/connection/CatalogTree/next/StreamDetailsPanel/StreamFieldsTable/SyncFieldCell/SyncFieldCell.tsx
```diff
@@ -58,20 +60,20 @@ export const SyncFieldCell: React.FC<SyncFieldCellProps> = ({
   }, [isCursor, isChildFieldCursor, isPrimaryKey, isChildFieldPrimaryKey, isNestedField, field.path]);
 
   return (
-    <>
+    <FlexContainer alignItems="center">
       {!showTooltip && (
```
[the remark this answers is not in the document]
josephkmh: I think the distinction here is that selection could be disabled in readonly mode, but we wouldn't show a tooltip. So that's why just `isDisabled` or `selectionDisabled` didn't quite make sense to me!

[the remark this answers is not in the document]
tealjulia: Oh, yep. That makes a lot of sense now, thanks!

josephkmh:
@edmundito @tealjulia yes the switch animation is not working 😢 I think this is because of how our table is rendering - basically any changes to any field re-renders the entire formik form, which in turn re-renders the table columns (because they depend on `syncCatalog.streams`), which essentially makes these components un-memo-able, because they depend on tons of values that in turn depend on `syncCatalog.streams`. My understanding is that it's all due to the fact that the entire sync catalog is stuffed into a single Formik `Field`, which makes the whole connection form very unstable. Refactoring that is of course outside the scope here, just wanted to share what I think the root cause is!

edmundito (approved):
🚢

airbyte-webapp/src/components/connection/CatalogTree/next/StreamDetailsPanel/StreamFieldsTable/StreamFieldsTable.tsx
```diff
@@ -212,6 +204,7 @@ export const StreamFieldsTable: React.FC<StreamFieldsTableProps> = ({
         },
       }),
     ],
+    // eslint-disable-next-line
```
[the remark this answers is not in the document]
josephkmh: Thanks for catching that! It's actually not needed, just leftover when I was playing with memoization trying to get the switch animation to work 😄 I've removed it now

timroes:
Superseded by https://github.com/airbytehq/airbyte-platform/pull/80

PR #22565, opened 2023-02-08
🪟 🎨 Fix Configuration settings page layout

edmundito:
## What
Updates the Configuration setting page layout to be consistent with other pages.

## How
* Card extends to full width as other settings pages
* Removes extra top margin


dizel852 (approved):
LGTM 👍 
Tested locally - UI is consistent with other pages

PR #22577, opened 2023-02-08
use launchdarkly feature flagging for column selection

mfsiega-airbyte:
## What
Use LaunchDarkly feature-flagging for column selection.

## How
Check whether column selection should be enabled for the workspace against the feature flag client (LaunchDarkly).

```
if yes -> do it
if no -> fall back on pre-existing env variable configs
```

This PR will always return `no` for now, so this is effectively a no-op. Then we can update the rules in LaunchDarkly as we see fit. Once we're happy with those, we can stop reading (and eventually remove) the various environment variables that are in use today.

## Testing
To validate this change, I did the following: (1) deployed to dev-3; (2) disabled column selection per the various old env variable flags; (3) enabled in LD; (4) verified via logs that column selection was being applied in the backend.


airbyte-container-orchestrator/src/main/java/io/airbyte/container_orchestrator/orchestrator/ReplicationJobOrchestrator.java
```diff
@@ -158,6 +161,12 @@ public Optional<String> runJob() throws Exception {
         sourceLauncherConfig.getDockerImage());
 
     log.info("Setting up replication worker...");
+    final UUID workspaceId = syncInput.getWorkspaceId();
```
colesnodgrass, line 164: Any possibility that `syncInput.getWorkspaceId();` returns null?

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/ReplicationActivityImpl.java
```diff
@@ -313,6 +315,13 @@ private CheckedSupplier<Worker<StandardSyncInput, ReplicationOutput>, Exception>
       final MetricClient metricClient = MetricClientFactory.getMetricClient();
       final WorkerMetricReporter metricReporter = new WorkerMetricReporter(metricClient, sourceLauncherConfig.getDockerImage());
 
+      final UUID workspaceId = syncInput.getWorkspaceId();
```
colesnodgrass, line 318: Any possibility that `syncInput.getWorkspaceId();` returns null?

colesnodgrass (approved):
👍🏻 as long as `syncInput.getWorkspaceId()` doesn't return `null`

airbyte-container-orchestrator/src/main/java/io/airbyte/container_orchestrator/orchestrator/ReplicationJobOrchestrator.java
mfsiega-airbyte: Yes indeed, will check for it in the next revision here.

joeykmh (approved):
Looks good to me. Also the feature flagging looks much less complicated with the new client - nice!

mfsiega-airbyte:
Some notes:
- I planned to add unit test coverage here, but discovered that these files have no unit test suites whatsoever. We should change that, but I don't want to block column selection on it.
- To validate this change, I did the following: (1) deployed to dev-3; (2) disabled column selection per the various old env variable flags; (3) enabled in LD; (4) verified via logs that column selection was being applied in the backend.

All in all I'm very confident that the change is safe, and also pretty confident that it works as intended, so I'll go ahead here.

PR #22582, opened 2023-02-08
🪟 🚦Add base E2E test for new stream table (part 2)

dizel852:
## What
Related to https://github.com/airbytehq/airbyte/issues/18267

## How
Add more tests:

- Check Source and Destination connector icons
- Check table columns names
- Check how many rows are displayed(actual/expected amount). Need to prepare appropriate test data
- Check table filtering
- Check table scrolling


airbyte-webapp-e2e-tests/cypress/integration/connection/streamTable.spec.ts
```diff
@@ -103,6 +116,37 @@ describe("New stream table - new connection set up ", () => {
     selectSchedule("Manual");
   });
 
+  it("should check check connector icons and titles in table", () => {
+    checkConnectorIconAndTitle("source");
```
krishnaglick, line 120: Is this just checking that they have a value, but not what that value is?

krishnaglick: Looking below, yes. Not sure if it's worth confirming the icon and name are what we expect since that should be fed to us by the DB.

krishnaglick (approved):
Had one comment but LGTM

airbyte-webapp-e2e-tests/cypress/integration/connection/streamTable.spec.ts
dizel852: We just check that icons and titles are displayed, but we don't check what exact icons are displayed.
Should I remove the test?

krishnaglick: Ehh, can't hurt. Well, I guess it can be another test in our already slow test suite but I don't think we should remove it.
Ignore my ramble :)

PR #22586, opened 2023-02-08
[Abandoned] : `AbstractJdbcSource` refactoring : Move SSL out of abstract class

akashkulk:
Some refactoring related to `AbstractJdbcSource` : 
- Move SSL helper methods out of `AbstractJdbcSource` into `JdbcSSLConnectionUtil`
- Move duplicate SSL constants out of `PostgresSource` into `PostgresSslConnectionUtil`

In general, part of a larger refactoring where to detangle the abstract classes
- Static methods should be preferred over abstract methods, that need to be overriden
- Deeply nested abstract class methods make it harder to reason about code flow. Removing as much as possible


airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/util/PostgresSslConnectionUtils.java
```diff
@@ -5,6 +5,7 @@
 package io.airbyte.integrations.util;
```
rodireich, line 5: I think the location of the file is confusing - is this a postgres source only? is this a base class?
Most of the functionality is not used in source

akashkulk: It's not great, but it can be part of a future refactor. There are a lot of constants here that are actually used in PostgresSource. (PARAM_MODE, PARAM_SSL, PARAM_SSL_MODE etc. So it's a good first place to put this

rodireich: Consts are the same, it's just the non of the functionality in this file should be used. 
I'm OK with moving this here for now to simplify but it will need probably another cleanup

rodireich (approved):

PR #22591, opened 2023-02-08
[ISSUE #19961] rename SingleSlice to SinglePartitionRouter

maxi297:
## What
This is a partial PR for https://github.com/airbytehq/airbyte/issues/19961 . It covers the renaming of SingleSlice to SinglePartitionRouter


brianjlai (commented):
given that this component is not used in any existing manifests, we can remove this component from the language entirely and just use it as a default when a partition router is not specified

brianjlai (approved):
(over the shoulder review w/ maxime) lgtm!

PR #22597, opened 2023-02-08
Source Salesforce: handle too many properties

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1403

## How
Make multiple requests for a page, each for a limited set of properties, then merge results by a primary key.


airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -135,10 +177,69 @@ def read_records(
...
+                    # (is allowed when properties length does not exceed the maximum value)
+                    # so there would be a single iteration, therefore we may and should yield records immediately
+                    yield from chunk_page_records
+                    break
+                chunk_page_records = {record[self.primary_key]: record for record in chunk_page_records}
+
+                for record_id, record in chunk_page_records.items():
+                    if record_id not in records:
+                        records[record_id] = record
+                    records[record_id].update(record)
```
grubberr, line 211: on high traffic streams there is a probability where you will get here only one part of record

davydov-d: you mean the data may be changed over time between two http calls? well yes, but I dont know if it can be solved in some other way..

davydov-d: or you mean something different?

davydov-d: Updated the code: log a warning and skip inconsistent records for now

grubberr (approved):

lazebnyi (approved):

PR #22609, opened 2023-02-08
Make connector acceptance test gradle plugin work for destinations

flash1293:
When trying to use the connector acceptance test plugin for destinations, it didn't execute correctly (see https://github.com/airbytehq/airbyte/pull/22317 ) This change makes the gradle plugin call the acceptance test in the same way the `acceptance-test-docker.sh` file from the connector template does it.

I don't notice any difference in what's actually run as part of the test, but maybe I'm missing something here.


alafanechere (approved):
It looks like a reasonable change to me but it's not very clear why it's solving the import error p problem for destinations.

buildSrc/src/main/groovy/airbyte-connector-acceptance-test.gradle
```diff
@@ -19,7 +19,7 @@ class AirbyteConnectorAcceptanceTestPlugin implements Plugin<Project> {
                         '-w', "$targetMountDirectory",
                         '-e', "AIRBYTE_SAT_CONNECTOR_DIR=${project.projectDir.absolutePath}",
                         'airbyte/connector-acceptance-test:dev',
-                        '-p', 'integration_tests.acceptance',
+                        '--acceptance-test-config', targetMountDirectory,
```
sherifnada, line 22: are you sure this isn't a red herring? I don't follow why this fixes the issue

flash1293: For example with https://github.com/airbytehq/airbyte/pull/22393 the acceptance tests run correctly when called this way but fail with “could not find module integration_tests.acceptance” with the old call.

I’m not sure why though, that’s why I pinged you on this.

sherifnada: Ah I think I see the problem. 

`-p integration_tets.acceptance` tells the running pytest process (CAT in this case) to look at a file `integration_tests/acceptance.py` to configure tests. 

In this situation the working directory is always set to the connector's directory. Python connectors always have a `integration_tests/acceptance.py` file, so this works for them. Even the java connectors which implement CAT also have that directory with an `acceptance.py` file. 

That directory is important because it allows the user to configure setup/teardown actions e.g: create a database and tear it down. 

I think the correct way to handle this would be to create an `integration_tests/acceptance.py` directory for the connector in question, or maybe conditionally add this flag depending on whether that directory exists. 

I think the `--acceptance-test-config` arg is unnecessary. By default SAT looks for an `acceptance-test-config.yaml` in the working directory (which is always set to the connector dir) so I think we can remove this arg.

sherifnada (changes requested):

flash1293:
> conditionally add this flag depending on whether that directory exists

I went with this approach for now, running a test here https://github.com/airbytehq/airbyte/pull/22393#issuecomment-1434452492 to see whether it works

flash1293:
@sherifnada OK, this seems to work. On the other PR where I configured the s3 destination to run acceptance tests I made the same changes as on this PR and the acceptance tests run correctly (showing the real problem with oneOf-usage): https://github.com/airbytehq/airbyte/pull/22393#issuecomment-1434452492

Running tests on a source still works correctly: https://github.com/airbytehq/airbyte/pull/22393#issuecomment-1434454139

To make sure the module is still picked up correctly when it's available, I broke the setup method here: https://github.com/airbytehq/airbyte/pull/22393/commits/31d9c9b22131a0e1415e5e57b00c48ebc7cfc2ab

and as expected the tests failed with this error message: https://github.com/airbytehq/airbyte/pull/22393#issuecomment-1434530935

alafanechere:
> I think the correct way to handle this would be to create an integration_tests/acceptance.py directory for the connector in question, or maybe conditionally add this flag depending on whether that directory exists.

Oh yes you're right, this is what we had to do to make the acceptance run on java sources.

sherifnada (approved):
FYI @clnoll in case this overlaps with your changes

PR #22610, opened 2023-02-08
🪟 🎨 🐛 Fix long name overflow and field path names in Catalog Diff modal

### LINKED ISSUE #22493: [Bug] Catalog diff is not showing diff correctly for nested fields
There are a couple of things that are happening.

1. The catalog diff is not showing the full path of the field, causing some confusion where a field is removed from one path and the same name appears on the other.
3. The root field is appearing as `nullobject -> nullobject` but it should not at all since each subfield should appear in the diff.

# Steps to reproduce
1. Use a source that supports nested fields (probably an API)
2. Remove a field from a stream and add it to a subfield in another stream
3. Refresh the catalog and look at the diff

## Expected
The diff shows the removal of the root field and addition of the nested field with the dot notation. It does not show the diff on the root object.

## Actual
Looks like the field in the root has been removed and re-added. The root field shows as a type change `nullobject -> nullobject`

## Context
[Slack Message](https://airbytehq-team.slack.com/archives/C042ABK94FR/p1675720099651229?thread_ts=1675720099.651229&cid=C042ABK94FR)

edmundito:
## What
Related to #22493

This change mainly fixes a few of the issues in the catalog diff modal

* Long names for namespace, stream name, or field name overflowing to the next column. Now, these fields will be truncated with an ellipsis
* Field names were not showing the full path with dot notation, which made it confusing. For example, instead of showing that "user.name" and "company.name" were added to the stream, it would show the last part of the path so you would see what seems to be a duplicate "name" entry.
* Some of the column names were misaligned, and now they are not

Before:
image

After:
Screen Shot 2023-02-08 at 16 52 26

## How
* Lots of CSS work
* Added mixin for text overflow issues
* Updated storybook to include long names as examples

### DIFF
--- a/airbyte-webapp/src/components/connection/CatalogDiffModal/DiffAccordionHeader.module.scss
+++ b/airbyte-webapp/src/components/connection/CatalogDiffModal/DiffAccordionHeader.module.scss
@@ -1,5 +1,6 @@
 @forward "./StreamRow.module.scss";
 @forward "./DiffSection.module.scss";
+@use "scss/colors";
 @use "scss/variables";
 
 .row {
@@ -8,6 +9,10 @@
 
 .namespace {
   padding-left: variables.$spacing-sm;
+}
+
+.grey {
+  color: colors.$grey-300;
 }
 
 .name {
--- a/airbyte-webapp/src/components/connection/CatalogDiffModal/DiffAccordionHeader.module.scss
+++ b/airbyte-webapp/src/components/connection/CatalogDiffModal/DiffAccordionHeader.module.scss
@@ -9,3 +9,7 @@
 .namespace {
   padding-left: variables.$spacing-sm;
 }
+
+.name {
+  width: 290px;
+}
--- a/airbyte-webapp/src/components/connection/CatalogDiffModal/DiffAccordionHeader.tsx
+++ b/airbyte-webapp/src/components/connection/CatalogDiffModal/DiffAccordionHeader.tsx
@@ -25,21 +25,24 @@ export const DiffAccordionHeader: React.FC<DiffAccordionHeaderProps> = ({
   newCount,
   changedCount,
 }) => {
-  const nameCellStyle = classnames(styles.nameCell, styles.row);
-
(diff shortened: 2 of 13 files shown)


krishnaglick:
Looks like there's some prettier issues failing the build.

dizel852:
Changes look great - tested locally on some test data and storybook, just a few comments:

I assume we can show a stream name longer:
Screenshot at Feb 09 22-18-53
Screenshot at Feb 09 22-22-23

We don't apply text ellipsis on field names with changes, I guess it should be aligned with "Data type" column name: 
image
Storybook:
image

krishnaglick (approved):
LGTM, pending Vlad's feedback!

krishnaglick:
Failing on what looks like a linting issue!

dizel852 (approved):
LGTM 👍 
Re-tested locally

image

edmundito:
@dizel852 Hmm I haven't changed anything since your comment... how did the overflow get fixed? 🤔

edmundito:
Followed up with @dizel852 and looks like some of the issues mentioned were resolved (bad pull?)

>  I assume we can show a stream name longer:
 Screenshot at Feb 09 22-18-53

I just pushed one more change to make this possible along with a small change to align the namespace with the value, and should be ready to go.

edmundito:
Whoops, made a couple of fixes that needed to be changed:

1. The diff modal shows a `object -> object` field change that has been very confusing. I made an update to exclude fields that have a same type change. For `object -> object`, it just shows the properties as added or removed
2. I noticed that when there is no namespace it was showing a blank. This changes it so that it shows "No namespace"
3. Small adjustment to a heading margin

Before:
Screen Shot 2023-02-16 at 10 28 45


After:
image

krishnaglick (approved):
Only thing that stands out to me is the modal can have horizontal scroll which seems like poor UX. But that's in storybook so...
Edit: Nevermind this is fine

dizel852 (approved):

timroes:
Superseded by https://github.com/airbytehq/airbyte-platform/pull/87

PR #22613, opened 2023-02-08
implement `PerfBackgroundJsonValidation` feature-flag

colesnodgrass:
## What
- add implementation of `PerfBackgroundJsonValidation` feature-flag
- convert `RecordSchemaValidationException` to `RuntimeException`
- add test for background-validation
- move `validationErrors` updating from `DefaultReplicationWorker.validateSchema` to `RecordSchemaValidator.validateSchema`
    - as the validation now happens in a background thread, the `RecordSchemaValidator` needs to update the `validationErrors` map
    - this isn't good code, as we're communicating via a `HashMap` being passed around via method parameters, but this code was already doing this higher up... I just pushed it down one more layer
- change `HashMap<AirbyteStreamNameNamespacePair, ImmutablePair<Set<String>, Integer>> validationErrors` to `ConcurrentHashMap`
    - as the `RecordSchemaValidator` could be validating in a different thread and is potentially updating the `validationErrors` while `validationErrors` is also being read in a different thread

## How
- moved `PerfBackgroundJsonValidation` check out of the `RecordSchemaValidator` to higher in the stack
    - this flag only needs to be checked once per job, not once per record


airbyte-featureflag/src/main/kotlin/Client.kt
```diff
@@ -118,7 +118,7 @@ class LaunchDarklyClient(private val client: LDClient) : FeatureFlagClient {
  *
  * @param [values] is a map of [Flag.key] to enabled/disabled status.
  */
-class TestClient @JvmOverloads constructor(val values: Map<String, Boolean> = mapOf()) : FeatureFlagClient {
+class TestClient @JvmOverloads constructor(val values: Map<String, Boolean> = mutableMapOf()) : FeatureFlagClient {
```
davinchia, line 121: what is this change for?

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -128,6 +130,7 @@ public DefaultReplicationWorker(final String jobId,
     this.recordSchemaValidator = recordSchemaValidator;
     this.metricReporter = metricReporter;
     this.connectorConfigUpdater = connectorConfigUpdater;
+    this.featureFlagClient = featureFlagClient;
```
davinchia, line 133: This doesn't seemed used. Are we planning to use this?

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
```diff
@@ -336,7 +339,7 @@ private static Runnable readFromSrcAndWriteToDstRunnable(final AirbyteSource sou
       MDC.setContextMap(mdc);
       LOGGER.info("Replication thread started.");
       long recordsRead = 0L;
-      final Map<AirbyteStreamNameNamespacePair, ImmutablePair<Set<String>, Integer>> validationErrors = new HashMap<>();
+      final ConcurrentHashMap<AirbyteStreamNameNamespacePair, ImmutablePair<Set<String>, Integer>> validationErrors = new ConcurrentHashMap<>();
```
davinchia, line 342: nit: add a comment explaining why this needs to be a concurrent hash map

airbyte-commons-worker/src/main/java/io/airbyte/workers/RecordSchemaValidator.java
```diff
@@ -52,31 +62,55 @@ public RecordSchemaValidator(final FeatureFlagClient featureFlagClient,
...
-    } else {
-      log.debug("workspace id is null");
-    }
+    });
+  }
+
+  /**
+   * @throws RecordSchemaValidationException
+   */
+  private void doValidateSchema(final AirbyteRecordMessage message, final AirbyteStreamNameNamespacePair messageStream) {
```
davinchia, line 111: nit: instead of `messageStream` -> `airbyteStream`?

`messageStream` keeps on making me think this is a stream of messages.

airbyte-commons-worker/src/main/java/io/airbyte/workers/RecordSchemaValidator.java
```diff
@@ -52,31 +62,55 @@ public RecordSchemaValidator(final FeatureFlagClient featureFlagClient,
...
+        handleException(e, messageStream, validationErrors);
+      }
+    }
+  }
 
-    if (workspaceId != null) {
-      if (featureFlagClient.enabled(PerfBackgroundJsonValidation.INSTANCE, new Workspace(workspaceId))) {
-        log.debug("feature flag enabled for workspace {}", workspaceId);
+  private void handleException(final RecordSchemaValidationException e,
+                               final AirbyteStreamNameNamespacePair messageStream,
```
davinchia, line 95: same comment here on better naming for `messageStream`

airbyte-commons-worker/src/main/java/io/airbyte/workers/RecordSchemaValidator.java
```diff
@@ -52,31 +62,55 @@ public RecordSchemaValidator(final FeatureFlagClient featureFlagClient,
       ((ObjectNode) schema).put("$schema", "http://json-schema.org/draft-07/schema#");
       validator.initializeSchemaValidator(stream.toString(), schema);
     }
-
   }
 
   /**
    * Takes an AirbyteRecordMessage and uses the JsonSchemaValidator to validate that its data conforms
    * to the stream's schema If it does not, this method throws a RecordSchemaValidationException
```
davinchia, line 69: nit: update this javadoc, we are no longer throwing a `RecordSchemaValidtionException`

airbyte-commons-worker/src/test/java/io/airbyte/workers/RecordSchemaValidatorTest.java
```diff
@@ -26,28 +28,57 @@ class RecordSchemaValidatorTest {
...
 
   @Test
-  void testValidateValidSchema() throws Exception {
-    final var featureFlagClient = new TestClient();
-    final var recordSchemaValidator = new RecordSchemaValidator(featureFlagClient, syncInput.getWorkspaceId(),
-        WorkerUtils.mapStreamNamesToSchemas(syncInput));
-    recordSchemaValidator.validateSchema(VALID_RECORD.getRecord(), AirbyteStreamNameNamespacePair.fromRecordMessage(VALID_RECORD.getRecord()));
+  void testValidateValidSchema() {
+    final var recordSchemaValidator = new RecordSchemaValidator(WorkerUtils.mapStreamNamesToSchemas(syncInput), false);
+    for (var i = 0; i < 10; i++) {
```
davinchia, line 43: nit: would be clearer if `i` were named something to reflect the validation threshold.

airbyte-commons-worker/src/test/java/io/airbyte/workers/RecordSchemaValidatorTest.java
```diff
@@ -26,28 +28,57 @@ class RecordSchemaValidatorTest {
...
+        INVALID_RECORD.getRecord(),
+        AirbyteStreamNameNamespacePair.fromRecordMessage(INVALID_RECORD.getRecord()),
+        validationErrors);
+    assertEquals(1, validationErrors.size());
+  }
+
+  @Test
+  void testValidateValidSchemaWithBackgroundValidation() {
+    final var recordSchemaValidator = new RecordSchemaValidator(WorkerUtils.mapStreamNamesToSchemas(syncInput), true);
+    for (var i = 0; i < 10; i++) {
```
davinchia, line 64: same thought here about i

airbyte-commons-worker/src/test/java/io/airbyte/workers/RecordSchemaValidatorTest.java
davinchia: curious, why do 10 times? If the record value isn't changing each time, seems not necessary to check 10 times.

davinchia (commented):
Looks great @colesnodgrass !

General comments for readability. Meat of logic looks sane.

Most of my comments aren't blocking. I do want to confirm one thing before I approve:

Today the behaviour on a validation error is to continue the sync, including the stream with validation errors. After the sync is finished, we print the validation errors out. Can I confirm this is identical to the prior behaviour?

airbyte-featureflag/src/main/kotlin/Client.kt
colesnodgrass: This allows for easier testing with the feature flag `TestClient`.  Maps in kotlin default to immutable maps, you have to explicitly label them as `mutable`.

The use case for this test client is to allow you to modify the map within a test to verify behavior:
```java
@Test
public void testSomething() {
  final var ffClient = new TestClient();
  // test scenario A
  ffClient.values.add("flagA", true);
  // test scenario B
}
```

airbyte-commons-worker/src/main/java/io/airbyte/workers/general/DefaultReplicationWorker.java
colesnodgrass: I had removed this, but it causes a conflict with `origin/master`, so I left it in to resolve that conflict.  Will remove it again.

colesnodgrass:
@davinchia addressed pr feedback, also marking this PR as `ready for review`

airbyte-commons-worker/src/main/java/io/airbyte/workers/RecordSchemaValidator.java
```diff
@@ -52,34 +62,56 @@ public RecordSchemaValidator(final FeatureFlagClient featureFlagClient,
...
-
   }
 
   /**
    * Takes an AirbyteRecordMessage and uses the JsonSchemaValidator to validate that its data conforms
-   * to the stream's schema If it does not, this method throws a RecordSchemaValidationException
-   *
-   * @param message
-   * @throws RecordSchemaValidationException
+   * to the stream's schema If it does not, an error is added to the validationErrors map.
```
davinchia, line 69:
```suggestion
   * to the stream's schema. If it does not, an error is added to the validationErrors map.
```

davinchia (approved):
Aside from the build - looks great!

PR #22633, opened 2023-02-09
🪟 🎨 BEM SASS naming convention styleguide & example

joeykmh:
## What
Adds a section in our frontend styleguide about using BEM notation in CSS modules. I've also converted an example component, `<Text>` to illustrate what this would look like in practice.


airbyte-webapp/src/components/ui/Text/text.module.scss
```diff
@@ -5,40 +5,40 @@
...
+    font-size: 11px;
   }
-}
 
-.sm,
-.md {
-  font-weight: 400;
-  line-height: 1.2em;
-}
+  &--sm,
```
letiescanciano, line 12: I always struggle to differentiate `element` and `modifier`. This `Text` component  will only have `modifiers`, any `element` example that you could think of? :)

josephkmh: In React I think `__element` tends to just be a different component in many cases, so it's likely we won't use them much. In the case of `<Text>` it's kind of hard to think of an example. But here's a fictional example:

```html
<div class="infoBox infoBox--highlighted">
  <span class="infoBox__icon"><svg ... /></span>
  <p class="infoBox__text infoBox__text--error">Hello!</p>
</div>
```
```scss
.infoBox {
  border: 1px solid black;
  
  &--highlighted {
    border-color: yellow;
  }

  &__icon {
    width: 10px;
    height: 10px;
  }
  
  &__text {
    color: black;

      &--error {
        color: red;
      }
  }
}
```

So **block** and **element** are both referring to separate HTML elements. The block is the main component, and it can have many elements that make it up (they can also be nested).

**Modifiers** you can think of like variants. They are not useful by themselves, only when used to modify a block or an element from it's normal styling depending on some dynamic logic (e.g. React props).

letiescanciano: Thanks a lot for this!! Totally get it now :)

PR #22635, opened 2023-02-09
fix: inject oauth params when generating sync input

pedroslopez:
## What

When moving the configuration fetching to happen at the attempt level in https://github.com/airbytehq/airbyte/pull/21629 we stopped injecting OAuth parameters into the configuration. This led to errors due to not having the right params during syncs. 

## How

- Properly inject oauth prams.


alafanechere (approved):

mfsiega-airbyte:
(Left a comment along these lines in Slack, but for visibility)

What testing have we done to make sure this is safe?

pedroslopez:
> (Left a comment along these lines in Slack, but for visibility)
> 
> What testing have we done to make sure this is safe?

@mfsiega-airbyte I've added tests here to verify that we're injecting oauth params, but currently deploying this to dev-2 to check it out in cloud. Will merge once I confirm there.

mfsiega-airbyte:
> @mfsiega-airbyte I've added tests here to verify that we're injecting oauth params, but currently deploying this to dev-2 to check it out in cloud. Will merge once I confirm there.

Did we not do this for the original PR?

pedroslopez:
> > @mfsiega-airbyte I've added tests here to verify that we're injecting oauth params, but currently deploying this to dev-2 to check it out in cloud. Will merge once I confirm there.
> 
> Did we not do this for the original PR?

I did not do this with an OAuth connector for the original PR.

evantahler (approved):

airbyte-workers/src/test/java/io/airbyte/workers/temporal/scheduling/activities/GenerateInputActivityTest.java
```diff
@@ -225,6 +239,8 @@ void testGetResetSyncWorkflowInput() throws IOException, ApiException {
         .withDestinationConfiguration(DESTINATION_CONFIGURATION)
         .withState(STATE);
 
+    verify(oAuthConfigSupplier).injectDestinationOAuthParameters(DESTINATION_DEFINITION_ID, WORKSPACE_ID, DESTINATION_CONFIGURATION);
```
evantahler, line 242: nit: could this test ensure that a secret was decrypted?

pedroslopez: Updated the test to actually check that we're using the config returned by the injection method. Internals of how the injection works left up to unit tests for the OAuthConfigSupplier.

PR #22636, opened 2023-02-09
Source Salesforce: turn on default HttpAvailabilityStrategy

### LINKED ISSUE #21785: Source Salesforce: Enable AvailabilityStrategy

grubberr:

## What
Issue: https://github.com/airbytehq/airbyte/issues/21785 https://github.com/airbytehq/airbyte-internal-issues/issues/1353

- rest full_refresh  - turned **ON** default HttpAvailabilityStrategy
- rest incremental - turned **ON** default HttpAvailabilityStrategy
- bulk full_refresh - turned **OFF** default HttpAvailabilityStrategy
- bulk incremental - turned **OFF** default HttpAvailabilityStrategy

Because it can slow down "bulk" streams because bulk stream run an async job and wait the result

## Pre-merge Checklist

- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`

### Airbyter

- [ ] Build is successful

</details>


erohmensing (commented):
@lazebnyi To be clear, for bulk jobs, if we get a 403 that isn't about request limits, we are skill skipping, and not erroring out (although we log an error) - is that correct? Will other streams sync after this? 

I'm looking at:
```
                elif error.response.status_code == codes.FORBIDDEN and error_code != "REQUEST_LIMIT_EXCEEDED":
                    self.logger.error(
                        f"Cannot receive data for stream '{self.name}' ,"
                        f"sobject options: {self.sobject_options}, error message: '{error_message}'"
                    )
```
[here](https://github.com/airbytehq/airbyte/pull/22636/files?diff=unified&w=0#diff-0572edeedd561901f8110c0f88b48efde87d1ed91247c0eb189c7229da12090dR312)

If so, I think that might be bulk streams' approach to availability...

grubberr:
@lazebnyi @erohmensing 
sorry writing here and there I have put some thoughts in issue
https://github.com/airbytehq/airbyte/issues/21785#issuecomment-1433077081

docs/integrations/sources/salesforce.md
```diff
@@ -129,6 +129,7 @@ Now that you have set up the Salesforce source connector, check out the followin
 
 | Version | Date       | Pull Request                                             | Subject                                                                                                                          |
 |:--------|:-----------|:---------------------------------------------------------|:---------------------------------------------------------------------------------------------------------------------------------|
+| 2.0.3   | 2023-02-15 | [22636](https://github.com/airbytehq/airbyte/pull/22636) | Turn on default HttpAvailabilityStrategy for RestSalesforceStream                                                                |
```
erohmensing, line 132:
```suggestion
| 2.0.3   | 2023-02-15 | [22636](https://github.com/airbytehq/airbyte/pull/22636) | Turn on default HttpAvailabilityStrategy for all streams that are not of class BulkSalesforceStream                                               |
```

erohmensing (commented):
I think this makes sense - we enable it for all the streams that aren't bulk streams, because it is too slow to check the bulk stream by pulling all of the data before... pulling all of the data 😄 and we already have error handling for bulk streams that skips only that stream on a 403, and does not fail the sync (is that correct?)


Can we remove the error handling and the associated comment around `FORBIDDEN` errors in the `read_records` method of the `SalesforceStream` base class? If we check at the beginning of the stream it shouldn't be necessary to also except it in the reading itself. 

I revised the changelog just a bit in order to make it more clear what is being opted in vs. out of the availability strategy (because it wasn't clear to me with less context that there are only the RestSalesforceStream and BulkSalesforceStream types)

lazebnyi (approved):
Once Ella's requested changes are implemented, I'm good with this PR.

airbyte-integrations/connectors/source-salesforce/source_salesforce/availability_strategy.py
```diff
@@ -0,0 +1,34 @@
...
+          1. Sobjects for which the user, after setting up the data using Airbyte, restricted access,
+             and we will receive 403 HTTP errors.
+          2. There are streams that do not allow you to make a sample using Salesforce `query` or `queryAll`.
+             And since we use a dynamic method of generating streams for Salesforce connector - at the stage of discover,
+             we cannot filter out these streams, so we catch them at the stage of reading data.
+        """
+        error_data = error.response.json()[0]
+        if error.response.status_code in [codes.FORBIDDEN, codes.BAD_REQUEST]:
+            error_code = error_data.get("errorCode", "")
+            if error_code != "REQUEST_LIMIT_EXCEEDED" or error_code == "INVALID_TYPE_FOR_OPERATION":
```
erohmensing, line 32: I think this just evaluates to `if error_code != "REQUEST_LIMIT_EXCEEDED"`, since `error_code == "INVALID_TYPE_FOR_OPERATION"` is included in that 🤔 

Is what we want here essentially "If Forbidden, or Bad Request with the specific INVALID_TYPE_FOR_OPERATION error, the stream is unavailable"?

This isn't where we want to handle rate limiting, as we could run into that at any point in the stream. We used to do it in `read_records` although I'm not 100% sure it belongs there vs in a customized [should_retry method](https://github.com/airbytehq/airbyte/blob/2e099acc52aeddc172c5ee66ed98426c69449a4a/airbyte-cdk/python/airbyte_cdk/sources/streams/http/http.py#L227-L238)

erohmensing: Edit, @grubberr pointed out that we handle rate limit exceeded by stopping the sync [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-salesforce/source_salesforce/source.py#L144) :)

airbyte-integrations/connectors/source-salesforce/source_salesforce/availability_strategy.py
```diff
@@ -0,0 +1,34 @@
...
+                return False, f"Cannot receive data for stream '{stream.name}', error message: '{error_data.get('message')}'"
+        return True, None
```
erohmensing, line 34:
```suggestion
        raise e
```
If we hit an error we don't expect, we still want to raise it. That is how the [default availability strategy](https://github.com/airbytehq/airbyte/blob/1c482ee3393a5d6d4e6283095c605669015e5419/airbyte-cdk/python/airbyte_cdk/sources/streams/http/availability_strategy.py#L80-L81) handles it

grubberr: good catch, updated!

airbyte-integrations/connectors/source-salesforce/source_salesforce/availability_strategy.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        There are several types of Salesforce sobjects that require additional processing:
+          1. Sobjects for which the user, after setting up the data using Airbyte, restricted access,
+             and we will receive 403 HTTP errors.
+          2. There are streams that do not allow you to make a sample using Salesforce `query` or `queryAll`.
+             And since we use a dynamic method of generating streams for Salesforce connector - at the stage of discover,
+             we cannot filter out these streams, so we catch them at the stage of reading data.
+        """
+        error_data = error.response.json()[0]
+        if error.response.status_code in [codes.FORBIDDEN, codes.BAD_REQUEST]:
```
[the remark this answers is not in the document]
grubberr: yes I agree, updated!

airbyte-integrations/connectors/source-salesforce/source_salesforce/availability_strategy.py
grubberr: 1. About the `if` condition
   yes, to be honest, this second part of the `if` condition is useless `error_code == "INVALID_TYPE_FOR_OPERATION"`
   I just copy-pasted this `if` block from the `read_records` method and I just decided not to change it.

2. About rate-limiting as I understand we discussed it in slack?

erohmensing: > About rate-limiting as I understand we discussed it in slack

yup! looks handled in read_stream 👍🏻

airbyte-integrations/connectors/source-salesforce/source_salesforce/availability_strategy.py
```diff
@@ -0,0 +1,35 @@
...
+    def handle_http_error(
+        self, stream: Stream, logger: logging.Logger, source: Optional["Source"], error: HTTPError
+    ) -> Tuple[bool, Optional[str]]:
+        """
+        There are several types of Salesforce sobjects that require additional processing:
+          1. Sobjects for which the user, after setting up the data using Airbyte, restricted access,
+             and we will receive 403 HTTP errors.
+          2. There are streams that do not allow you to make a sample using Salesforce `query` or `queryAll`.
+             And since we use a dynamic method of generating streams for Salesforce connector - at the stage of discover,
+             we cannot filter out these streams, so we catch them at the stage of reading data.
```
erohmensing, line 27:
```suggestion
             we cannot filter out these streams, so we check for them before reading from the streams.
```

airbyte-integrations/connectors/source-salesforce/source_salesforce/availability_strategy.py
erohmensing: > About the if condition

I think it's worth refactoring, but separately is fine in order to avoid introducing unexpected behavior here. But I do think we might be excepting too much with regards to BAD_REQUESTs that aren't due to INVALID_TYPE_FOR_OPERATION.

erohmensing (approved):

PR #22638, opened 2023-02-09
🪟 🎉 Fetch LaunchDarkly flags & experiments scoped to a workspace

joeykmh:
## What
This PR allows us to use workspace-scoped flags on the frontend

## How
- Updates our SDK to version 3.1.0 (not strictly necessary for this PR, but opens the door to use contexts later)
- Adjusts our `mapUserToLDUser()` method to accept an optional `workspaceId`
- Changes our `useCurrentWorkspaceId()` hook to work outside `<Route>` elements (useful for other situations as well!)
- Adjusts `useEffect()` in `LDExperimentService` that detects location changes and re-identifies the user


airbyte-webapp/src/packages/cloud/views/DefaultView.tsx
```diff
@@ -1,7 +1,6 @@
 import { Navigate } from "react-router-dom";
 
 import { RoutePaths } from "../../../pages/routePaths";
-import { CloudRoutes } from "../cloudRoutePaths";
```
timroes, line 4: Can the `CloudRoutes.SelectWorkspace` be deleted from its file now? Seems we didn't do that in this PR.

timroes (approved):
Code LGTM, have seen and debugged together with Joey, so saw the functionality working as expected and allowing us to trigger in LaunchDarkly as we'd expect. Small comment, but other than that looks good for me.

airbyte-webapp/src/packages/cloud/views/DefaultView.tsx
josephkmh: You mean removing `SelectWorkspace` from the `CloudRoutes` enum entirely? Yes, we can do that - addressed here: 47a1e15174336a27f43065ffebf7e965b3415c8f

PR #22646, opened 2023-02-09
🪟 🔧 Add TS CSS modules plugin

joeykmh:
## What
Adds `typescript-plugin-css-modules` so that we get a nice TS-server powered autocomplete & checking of our CSS module files 🎉


timroes (approved):
This is a nice extension during development. (Just for posterity: this does not type check when we actually build). `pnpm audit` showed that this library pulls in some vulnerable new dependencies, so I made sure we're fixing them to their patched versions and pushed it to the PR. Otherwise looks good to me

PR #22667, opened 2023-02-09
🪟 🐛 Add space between job is in progress and schema changes detected indicator in connections table

### LINKED ISSUE #22629: [Bug] Missing space between job is in progress and schema changes detected indicator in connections table
Image

dizel852:
## What
Resolves https://github.com/airbytehq/airbyte/issues/22629

## How
Add `gap` for `flex` `<StatusCell />` container

Before:
Screenshot at Feb 09 20-29-58

After fix:
Screenshot at Feb 09 20-31-02

### DIFF
--- a/airbyte-webapp/src/components/EntityTable/components/StatusCell.module.scss
+++ b/airbyte-webapp/src/components/EntityTable/components/StatusCell.module.scss
@@ -1,7 +1,9 @@
+@use "scss/variables";
+
 .container {
   display: flex;
   flex-direction: row;
   align-content: center;
   justify-content: space-between;
-  width: 100px;
+  gap: variables.$spacing-md;
 }
--- a/airbyte-webapp/src/components/EntityTable/components/StatusCell.module.scss
+++ b/airbyte-webapp/src/components/EntityTable/components/StatusCell.module.scss
@@ -5,5 +5,5 @@
   flex-direction: row;
   align-content: center;
   justify-content: space-between;
-  gap: variables.$spacing-md;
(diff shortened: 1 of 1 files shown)


edmundito:
@dizel852 Looking a lot better but I was wondering if one more thing can be done. When pressing the "sync now" button next to a connection with a breaking/non breaking change indicator will lead the whole table to resize because the column width has changed.

Is it possible to assign a fixed width to that column that can fit all the states so that the table doesn't resize when there's a state change?

dizel852:
> @dizel852 Looking a lot better but I was wondering if one more thing can be done. When pressing the "sync now" button next to a connection with a breaking/non breaking change indicator will lead the whole table to resize because the column width has changed.
> 
> Is it possible to assign a fixed width to that column that can fit all the states so that the table doesn't resize when there's a state change?

Sure! Will update the fix

airbyte-webapp/src/components/EntityTable/components/StatusCell.module.scss
```diff
@@ -1,7 +1,9 @@
+@use "scss/variables";
```
krishnaglick, line 1: This can be removed.

krishnaglick (approved):

airbyte-webapp/src/components/EntityTable/components/StatusCell.module.scss
dizel852: Oh, yes, missed that. Thanks!

PR #22669, opened 2023-02-09
Use Micronaut with airbyte-server-wrapped

jdpgrailsdev:
## What
* Support conversion of the airbyte-server-wrapped Cloud module to Micronaut

## How
* Remove `@Requires` rule from controllers so they will be used in both OSS and Cloud
* Remove `airbyte-commons-worker` dependency from `airbyte-server` by moving shared code to new `airbyte-commons-converters` module

## Recommended reading order
1. `*ApiController.java`

## Testing

* All tests pass
* Project builds locally
* Tested changes in Cloud locally

**N.B.** This will not be merged until the corresponding Cloud PR has been tested in a dev environment.  At the point of merging, this will temporarily break Cloud if Cloud is updated to that OSS commit or later without the corresponding change being merged on the Cloud side.


airbyte-commons-converters/src/main/java/io/airbyte/converters/CatalogClientConverters.java
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) 2023 Airbyte, Inc., all rights reserved.
  */
 
-package io.airbyte.workers.helper;
+package io.airbyte.converters;
```
benmoriceau, line 5: Nit: All the other commons have `commons` in their package names, should we keep this pattern for this new module?

airbyte-server/src/main/java/io/airbyte/server/apis/ConnectionApiController.java
```diff
@@ -31,9 +30,7 @@
 import io.micronaut.security.rules.SecurityRule;
 
 @Controller("/api/v1/connections")
-@Context()
-@Requires(property = "airbyte.deployment-mode",
-          value = "OSS")
+@Context
```
benmoriceau, line 33: Should we remove`@Context` or add it to the other controller as well? I'm not sure that it has any impact on cloud. It was just an annotation I added after noticing slowness on my local test.

airbyte-server/src/main/java/io/airbyte/server/apis/NotificationsApiController.java
```diff
@@ -13,23 +13,20 @@
...
-@Requires(property = "airbyte.deployment-mode",
-          value = "OSS")
+@Controller("/api/v1/notifications")
 @Secured(SecurityRule.IS_AUTHENTICATED)
 public class NotificationsApiController implements NotificationsApi {
 
   public NotificationsApiController() {}
 
-  @Post
+  @Post("/try")
```
benmoriceau, line 29: Should we do the same for the other controllers (like the openApi one)

airbyte-server/src/main/java/io/airbyte/server/config/ApplicationBeanFactory.java
```diff
@@ -109,6 +110,11 @@ public AirbyteProtocolVersionRange airbyteProtocolVersionRange(
     return new AirbyteProtocolVersionRange(new Version(minVersion), new Version(maxVersion));
   }
 
+  @Singleton
```
benmoriceau, line 113: Why is it needed in the server? It feel strange to have a need for a client in the server. Is the server doing request to the server?

benmoriceau (approved):
I have nit comments that shouldn't block merging that.

airbyte-commons-converters/src/main/java/io/airbyte/converters/CatalogClientConverters.java
jdpgrailsdev: Probably.  I will do a rename.

airbyte-server/src/main/java/io/airbyte/server/apis/ConnectionApiController.java
jdpgrailsdev: I think that we should do a pass on this after we get these changes in.  I'm hesitant to remove anything until we get everything working and have a baseline.

airbyte-server/src/main/java/io/airbyte/server/config/ApplicationBeanFactory.java
jdpgrailsdev: It is used by the `OAuthHandler`, presumably to fetch OAuth tokens.

PR #22670, opened 2023-02-09
[ISSUE #19961] update code to newest CDK version and clean autogenera…

maxi297:
## What
This is a partial PR for https://github.com/airbytehq/airbyte/issues/19961 . It covers the cleaning of the generated files in the connector-builder-server since those files were not aligned anymore with the current model and that these files are not used


brianjlai (approved):
begone useless manifest!

PR #22678, opened 2023-02-09
Add source timeout configuration column

benmoriceau:
## What
Create a Migration to create the new column that will allow to configure the source heartbeat timeout.

Tested localy.


jdpgrailsdev (approved):
:shipit:

PR #22832, opened 2023-02-10
Fix `source-github` data types

evantahler:
1. If the stream has `created_at`, it can't be null
1. If the stream has `updated_at`, it can't be null
2. `workflow_runs.run_started_at` is a datetime


alex-gron (approved):
Thanks Evan! 🚀

grubberr (approved):

pedroslopez (approved):
LGTM but this is technically breaking since we're narrowing the type (nullable, not nullable). This should be ok as long as github has really been setting these.

I would expect needing to specify the breaking change test skip after that test run finishes up.

evantahler:
> I would expect needing to specify the breaking change test skip after that test run finishes up.

Good catch!

evantahler:
The `/test` failures appear to be timeouts... I guess I'll keep trying

evantahler:
@grubberr do you have ay suggestions about how to get these tests to pass?  Feel free to modify this PR if you have any ideas...

evantahler:
@grubberr! Ping! Do you have any suggestions on how to get these tests pasing?

grubberr:
BTW as I see this PR https://github.com/airbytehq/airbyte/pull/24523 do the same and more than current PR

PR #22834, opened 2023-02-10
🪟 🐛 Fix clearing source-defined cursor or primary key when there's a valid update

edmundito:
## What
This fixes a scenario where a stream has an updated source-defined cursor and primary key, but the web app would clear it when the new field used in the source-defined cursor or primary key was added during the stream refresh. Now it checks to ensure that it removes the breaking changes.

Here's a scenario where this would occur:

* Previous cursor: ['created_at']
* New cursor: [ 'updated_at']
* Diff: Remove field 'updated_at', Add field 'created_at'
* Resulting cursor: []

Because the diff is adding 'created_at' and the code was not checking to see if this was important, it would set the cursor to an empty array.

## How
* The catalogDiff already signals which fields in the change are breaking. Before determining in the primary key or cursor need to be cleared in `clearBreakingFieldChanges`, the `breaking` flag filters the list of updates.
* Updated unit tests to support a scenario where the source-defined key or cursor would be updated with newly added fields.
* Code cleanup around `calculateInitialCatalog.`

### DIFF
--- a/airbyte-webapp/src/components/connection/ConnectionForm/calculateInitialCatalog.ts
+++ b/airbyte-webapp/src/components/connection/ConnectionForm/calculateInitialCatalog.ts
@@ -37,7 +37,6 @@ const clearBreakingFieldChanges = (nodeStream: SyncSchemaStream, breakingChanges
       .map((update) => update.fieldName);
 
     // if there is a primary key in the config, and any of its field paths were removed, we'll be clearing it
-    // The source-defined primary key already been updated by verifySourceDefinedProperties
     if (
       !!primaryKey?.length &&
       primaryKey?.some((primaryKeyPath) => breakingFieldPaths.some((path) => isEqual(primaryKeyPath, path)))
@@ -46,7 +45,6 @@ const clearBreakingFieldChanges = (nodeStream: SyncSchemaStream, breakingChanges
     }
 
     // if there is a cursor field, and any of its field path was removed, we'll be clearing it
-    // The source-defined cursor has already been updated by verifySourceDefinedProperties
     if (!!cursorField?.length && breakingFieldPaths.some((path) => isEqual(path, cursorField))) {
       clearCursorField = true;
     }
--- a/airbyte-webapp/src/components/connection/ConnectionForm/calculateInitialCatalog.ts
+++ b/airbyte-webapp/src/components/connection/ConnectionForm/calculateInitialCatalog.ts
@@ -21,42 +21,49 @@ const clearBreakingFieldChanges = (nodeStream: SyncSchemaStream, breakingChanges
     return nodeStream;
   }
 
+  const { primaryKey, cursorField } = nodeStream.config;
+
   let clearPrimaryKey = false;
   let clearCursorField = false;
 
   for (const streamTransformation of breakingChangesByStream) {
-    // get all of the removed field paths for this transformation
-    const removedFieldPaths = streamTransformation.updateStream?.map((update) => update.fieldName);
-
-    if (!removedFieldPaths?.length) {
+    if (!streamTransformation.updateStream || !streamTransformation.updateStream?.length) {
       continue;
     }
 
+    // get all of the removed field paths for this transformation
+    const breakingFieldPaths = streamTransformation.updateStream
+      .filter(({ breaking }) => breaking)
+      .map((update) => update.fieldName);
+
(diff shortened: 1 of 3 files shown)


airbyte-webapp/src/core/domain/catalog/traverseSchemaToField.test.ts
```diff
@@ -2,7 +2,7 @@ import { AirbyteJSONSchema } from "core/jsonSchema/types";
 
 import { traverseSchemaToField } from "./traverseSchemaToField";
 
-describe(`${traverseSchemaToField}`, () => {
+describe(`${traverseSchemaToField.name}`, () => {
```
edmundito, line 5: I noticed that the test name for this was spitting out the function definition 😆

airbyte-webapp/src/components/connection/ConnectionForm/calculateInitialCatalog.ts
```diff
@@ -21,42 +21,49 @@ const clearBreakingFieldChanges = (nodeStream: SyncSchemaStream, breakingChanges
...
-    const removedFieldPaths = streamTransformation.updateStream?.map((update) => update.fieldName);
-
-    if (!removedFieldPaths?.length) {
+    if (!streamTransformation.updateStream || !streamTransformation.updateStream?.length) {
       continue;
     }
 
+    // get all of the removed field paths for this transformation
+    const breakingFieldPaths = streamTransformation.updateStream
+      .filter(({ breaking }) => breaking)
```
edmundito, line 36: This is the fix. The rest of the code is cleanup.

airbyte-webapp/src/components/connection/ConnectionForm/calculateInitialCatalog.ts
```diff
@@ -21,42 +21,49 @@ const clearBreakingFieldChanges = (nodeStream: SyncSchemaStream, breakingChanges
...
+      .map((update) => update.fieldName);
+
     // if there is a primary key in the config, and any of its field paths were removed, we'll be clearing it
+    // The source-defined primary key already been updated by verifySourceDefinedProperties
```
tealjulia, line 40: is it worth clarifying this by saying `"The source-defined primary key, if one exists..."`

airbyte-webapp/src/components/connection/ConnectionForm/calculateInitialCatalog.ts
```diff
@@ -21,42 +21,49 @@ const clearBreakingFieldChanges = (nodeStream: SyncSchemaStream, breakingChanges
...
     ) {
       clearPrimaryKey = true;
     }
 
     // if there is a cursor field, and any of its field path was removed, we'll be clearing it
-    if (
-      !!nodeStream.config?.cursorField?.length &&
-      removedFieldPaths.some((removedPath) => isEqual(removedPath, nodeStream?.config?.cursorField))
-    ) {
+    // The source-defined cursor has already been updated by verifySourceDefinedProperties
```
tealjulia, line 49: Same wondering as the comment above.

airbyte-webapp/src/components/connection/ConnectionForm/calculateInitialCatalog.ts
krishnaglick: It looks like an empty `breakingFieldPaths` does not trigger a continue when it did before.
But below it wouldn't set either of the flags. Is this cleanup just removing an early exit of the loop?

teallarson (approved):
lgtm, tested locally with a replaced primary key and a partially replaced composite primary key.  one question on comment wording but not blocking.

airbyte-webapp/src/components/connection/ConnectionForm/calculateInitialCatalog.ts
edmundito: I did the cleanup before the fix, but I moved it up because the map is only necessary if the list has anything to change:

```
streamTransformation.updateStream?.length === streamTransformation.updateStream?.map((update) => update.fieldName).length
```

krishnaglick (approved):

airbyte-webapp/src/components/connection/ConnectionForm/calculateInitialCatalog.ts
edmundito: I will remove these. These were from an older solution that originally involved checking source-defined here, but with the check on breaking flag, it should do the right thing.

PR #22840, opened 2023-02-10
Reduce input generation payload

benmoriceau:
## What
Use the minimum needed payload in the generation of the sync input. It is still an activity in order to avoid changing the injection of the resourcesRequirement.


airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
```diff
@@ -186,7 +188,13 @@ public StandardSyncOutput run(final JobRunConfig jobRunConfig,
   private NormalizationInput generateNormalizationInput(final StandardSyncInput syncInput,
                                                         final StandardSyncOutput syncOutput) {
 
-    return normalizationActivity.generateNormalizationInput(syncInput, syncOutput);
+    final int version = Workflow.getVersion(REMOVE_ACTIVITY_FOR_NORM_INPUT, Workflow.DEFAULT_VERSION, REMOVE_ACTIVITY_FOR_NORM_INPUT_CURRENT_VERSION);
+    if (version == Workflow.DEFAULT_VERSION) {
+      return normalizationActivity.generateNormalizationInput(syncInput, syncOutput);
+    } else {
+      return normalizationActivity.generateNormalizationInputWithMinimumPayload(syncInput.getDestinationConfiguration(), syncInput.getCatalog(),
```
pmossman, line 195: instead of 'syncInput.getCatalog()', should this instead be `syncOutput.getOutputCatalog()`? The existing `normalizationActivity.generateNormalizationInput` implementation uses the output catalog instead of the input catalog, so I wanted to call this out

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/SyncWorkflowImpl.java
```diff
@@ -49,6 +49,8 @@ public class SyncWorkflowImpl implements SyncWorkflow {
   private static final int NORMALIZATION_SUMMARY_CHECK_CURRENT_VERSION = 1;
   private static final String AUTO_DETECT_SCHEMA_TAG = "auto_detect_schema";
   private static final int AUTO_DETECT_SCHEMA_VERSION = 2;
+  private static final String REMOVE_ACTIVITY_FOR_NORM_INPUT = "remove_activity_for_norm_input";
```
pmossman, line 52: this version name is a bit confusing now that you're no longer deciding to remove the activity. Maybe `use_minimal_norm_input` is a better version name now?

pmossman (approved):
@benmoriceau I called out a few potential changes you might want to make, but generally looks good to me!

benmoriceau:
There is an issue with normalization when I run on my local

benmoriceau:
@pmossman I have activated the auto merge, let me know if it should go before the code freeze

airbyte-workers/src/main/java/io/airbyte/workers/temporal/sync/NormalizationActivityImpl.java
```diff
@@ -176,6 +178,11 @@ public NormalizationSummary normalize(final JobRunConfig jobRunConfig,
 
   @Trace(operationName = ACTIVITY_TRACE_OPERATION_NAME)
   @Override
+  @Deprecated(forRemoval = true)
+  /**
+   * This activity is deprecated. It is using a big payload which is not needed, it has been replace
```
cgardens, line 183: what is big in `SyncInput`? Isn't catalog the big object?

benmoriceau: It is likely one of the biggest object. However the Sync Input also have unused object by the normalization such as State that can be big. The goal is to resolve the blob size issue face in https://airbytehq-team.slack.com/archives/C04NTJ82RC4/p1676046492611669?thread_ts=1676042334.857199&cid=C04NTJ82RC4. This is a short term fix which aims to unstuck a connection that is currently stuck.

benmoriceau: It also avoid having the catalog store in the SyncInput

davinchia:
I retested this after merging in https://github.com/airbytehq/airbyte/pull/23039 just in case.

octavia-approvington:
Lets do it! <br/> giddddy-up

octavia-approvington (approved):

PR #22846, opened 2023-02-10
Default CH ssl to true and fix the failure if ssl property is missing

### LINKED ISSUE #22845: Clickhouse normalization fails with clickhouse-strict-encrypt connector
There is a bug in Clickhouse normalization. It assumes that its configuration object includes "ssl" property, but "ssl" property is removed from the configuration object by clickhouse-strict-encrypt variant. Error message looks like this:
```
2023-02-01 02:49:51 normalization >   File "/usr/local/lib/python3.10/site-packages/normalization/transform_config/transform.py", line 327, in transform_clickhouse
2023-02-01 02:49:51 normalization >     "secure": config["ssl"],
```

grishick:
fixes https://github.com/airbytehq/airbyte/issues/22845

## What
Clickhouse-strict-encrypt removes "ssl" property from the connector's spec. This change fixes the error in normalization that fails when it does not find "ssl" property in the config and makes Clickhouse normalization use SSL by default when the property is missing.


airbyte-integrations/bases/base-normalization/normalization/transform_config/transform.py
```diff
@@ -325,10 +325,17 @@ def transform_clickhouse(config: Dict[str, Any]):
...
         }
         if "password" in config:
             dbt_config["password"] = config["password"]
+
+        # ssl is an optional configuration and is not present in strict-encrypt config
+        # if ssl option is not present in the config - default to True
+        if "ssl" in config:
+            dbt_config["secure"] = config["ssl"]
+        else:
+            dbt_config["secure"] = True
```
jbfbell, line 337: Logic is fine here, one liner would be:
```python
dbt_config["secure"] = config.get("ssl", True)
```

jbfbell (approved):

ryankfu (approved):
LGTM

airbyte-integrations/bases/base-normalization/normalization/transform_config/transform.py
ryankfu: `+1` on the default value one-liner

grishick: clearly, I don't code enough in Python :)

PR #22854, opened 2023-02-11
QA checks: Validate the connector code does not contain `http://` URL 

### LINKED ISSUE #21717: Automated QA checks: Validate the connector code does not contain `http://` URL
We want to make sure no connector is calling non SSL URL (`http://`) .

* Implement the `check_connector_https_url_only ` function [here](https://github.com/airbytehq/airbyte/blob/fbff8aacbe57476a7328325f6ffc228880b5d372/tools/ci_connector_ops/ci_connector_ops/qa_checks.py)

A `grep` like this could do the job (or python equivalent): 
```bash
grep -r -i --include \*.py --include \*.java --exclude-dir .venv  --exclude-dir unit_tests --exclude-dir generator "http://" .
```

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/21717
We want to make sure no connector uses plain HTTP:// urls.

## How
* Implement a grep style qa check to:
  *  read all readable files in a connector
  * ignore a couple of files with patterns (`.venv`, `test`)
  * ignore json schema HTTP urls.
 
## 🚨 User Impact 🚨
This new QA check is likely to make fail the build of a couple of connectors:

Failing alpha connectors:
	- source-activecampaign
	- source-shortio
	- source-statuspage
	- source-surveycto
	- source-tmdb
	- source-visma-economic
	- source-weatherstack
	- source-zuora
	- destination-doris
	- destination-iceberg
	- destination-google-sheets

Failing generally_available connectors:
	- source-github
	- source-google-search-console
	- source-google-sheets
	- source-marketo
	- source-stripe
	- destination-s3

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/qa_checks.py
+++ b/tools/ci_connector_ops/ci_connector_ops/qa_checks.py
@@ -98,6 +98,7 @@ def read_all_files_in_directory(directory: Path, ignored_directories: Optional[S
 
 
 IGNORED_DIRECTORIES_FOR_HTTPS_CHECKS = {".venv", "tests", "unit_tests", "integration_tests", "build", "source-file"}
+IGNORED_URLS_PREFIX = {"http://json-schema.org", "http://localhost"}
 
 
 def check_connector_https_url_only(connector: Connector) -> bool:
@@ -111,10 +112,11 @@ def check_connector_https_url_only(connector: Connector) -> bool:
     """
     files_with_http_url = set()
     for filename, line in read_all_files_in_directory(connector.code_directory, IGNORED_DIRECTORIES_FOR_HTTPS_CHECKS):
-        if "http://json-schema.org" or "http://localhost" in line:
+        if any([prefix in line.lower() for prefix in IGNORED_URLS_PREFIX]):
             continue
         if "http://" in line.lower():
             files_with_http_url.add(str(filename))
+
     if files_with_http_url:
         files_with_http_url = "\n\t- ".join(files_with_http_url)
         print(f"The following files have http:// URLs:\n\t- {files_with_http_url}")
--- a/tools/ci_connector_ops/ci_connector_ops/qa_checks.py
+++ b/tools/ci_connector_ops/ci_connector_ops/qa_checks.py
@@ -85,10 +85,16 @@ def check_connector_icon_is_available(connector: Connector) -> bool:
     return connector.icon_path.exists()
 
 
-def read_all_files_in_directory(directory: Path, ignored_directories: Optional[Set[str]] = None) -> Iterable[Tuple[str, str]]:
+def read_all_files_in_directory(
+    directory: Path, ignored_directories: Optional[Set[str]] = None, ignored_filename_patterns: Optional[Set[str]] = None
+) -> Iterable[Tuple[str, str]]:
     ignored_directories = ignored_directories if ignored_directories is not None else {}
+    ignored_filename_patterns = ignored_filename_patterns if ignored_filename_patterns is not None else {}
+
     for path in directory.rglob("*"):
-        ignore = any([ignored_directory in path.parts for ignored_directory in ignored_directories])
+        ignore_directory = any([ignored_directory in path.parts for ignored_directory in ignored_directories])
+        ignore_filename = any([path.match(ignored_filename_pattern) for ignored_filename_pattern in ignored_filename_patterns])
+        ignore = ignore_directory or ignore_filename
         if path.is_file() and not ignore:
             try:
                 for line in open(path, "r"):
@@ -97,7 +103,8 @@ def read_all_files_in_directory(directory: Path, ignored_directories: Optional[S
                 continue
 
 
-IGNORED_DIRECTORIES_FOR_HTTPS_CHECKS = {".venv", "tests", "unit_tests", "integration_tests", "build", "source-file"}
+IGNORED_DIRECTORIES_FOR_HTTPS_CHECKS = {".venv", "tests", "unit_tests", "integration_tests", "build", "source-file", ".pytest_cache"}
+IGNORED_FILENAME_PATTERN_FOR_HTTPS_CHECKS = {"*Test.java"}
 IGNORED_URLS_PREFIX = {"http://json-schema.org", "http://localhost"}
 
 
@@ -111,7 +118,9 @@ def check_connector_https_url_only(connector: Connector) -> bool:
         bool: Wether the connector code contains only https url.
     """
     files_with_http_url = set()
-    for filename, line in read_all_files_in_directory(connector.code_directory, IGNORED_DIRECTORIES_FOR_HTTPS_CHECKS):
+    for filename, line in read_all_files_in_directory(
+        connector.code_directory, IGNORED_DIRECTORIES_FOR_HTTPS_CHECKS, IGNORED_FILENAME_PATTERN_FOR_HTTPS_CHECKS
+    ):
         if any([prefix in line.lower() for prefix in IGNORED_URLS_PREFIX]):
             continue
         if "http://" in line.lower():
--- a/tools/ci_connector_ops/ci_connector_ops/qa_checks.py
+++ b/tools/ci_connector_ops/ci_connector_ops/qa_checks.py
@@ -103,7 +103,16 @@ def read_all_files_in_directory(
                 continue
 
 
-IGNORED_DIRECTORIES_FOR_HTTPS_CHECKS = {".venv", "tests", "unit_tests", "integration_tests", "build", "source-file", ".pytest_cache"}
+IGNORED_DIRECTORIES_FOR_HTTPS_CHECKS = {
+    ".venv",
+    "tests",
+    "unit_tests",
+    "integration_tests",
+    "build",
+    "source-file",
+    ".pytest_cache",
+    "acceptance_tests_logs",
+}
 IGNORED_FILENAME_PATTERN_FOR_HTTPS_CHECKS = {"*Test.java"}
 IGNORED_URLS_PREFIX = {"http://json-schema.org", "http://localhost"}
 
--- a/tools/ci_connector_ops/ci_connector_ops/qa_checks.py
+++ b/tools/ci_connector_ops/ci_connector_ops/qa_checks.py
@@ -100,6 +100,7 @@ def read_all_files_in_directory(
                 for line in open(path, "r"):
                     yield path, line
             except UnicodeDecodeError:
+                print(f"{path} could not be decoded as it is not UTF8.")
                 continue
 
 
@@ -112,9 +113,28 @@ IGNORED_DIRECTORIES_FOR_HTTPS_CHECKS = {
     "source-file",
     ".pytest_cache",
     "acceptance_tests_logs",
+    ".hypothesis",
 }
-IGNORED_FILENAME_PATTERN_FOR_HTTPS_CHECKS = {"*Test.java"}
+
+IGNORED_FILENAME_PATTERN_FOR_HTTPS_CHECKS = {"*Test.java", "*.pyc", "*.gz"}
 IGNORED_URLS_PREFIX = {"http://json-schema.org", "http://localhost"}
+
+
+def is_comment(line: str, file_path: Path):
+    language_comments = {
+        ".py": "#",
+        ".yml": "#",
+        ".yaml": "#",
+        ".java": "//",
+        ".md": "<!--",
+    }
+
+    denote_comment = language_comments.get(file_path.suffix)
+    if not denote_comment:
+        return False
+
+    trimmed_line = line.lstrip()
+    return trimmed_line.startswith(denote_comment)
 
 
 def check_connector_https_url_only(connector: Connector) -> bool:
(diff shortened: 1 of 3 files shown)


tools/ci_connector_ops/tests/test_qa_checks.py
```diff
@@ -6,87 +6,75 @@
...
-    (utils.Connector("source-faker"), True),
-    (utils.Connector("source-foobar"), False),
-])
+@pytest.mark.parametrize(
+    "connector, expect_exists",
+    [
+        (utils.Connector("source-faker"), True),
+        (utils.Connector("source-foobar"), False),
+    ],
+)
```
alafanechere, line 18: Lots of changes from `black` - my precommit was on vacation in the last couple of weeks :( But it's back!

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
```diff
@@ -93,9 +105,19 @@ def check_connector_https_url_only(connector: Connector) -> bool:
     Returns:
         bool: Wether the connector code contains only https url.
     """
-    # TODO implement
+    files_with_http_url = set()
+    for filename, line in read_all_files_in_directory(connector.code_directory, path_patterns_to_ignore):
+        if "http://json-schema.org/draft-07/schema#" in line:
```
evantahler, line 110: Ha! JSONschema is unsafe.  Maybe 

```suggestion
        if "http://json-schema.org/" in line:
```

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
```diff
@@ -93,9 +105,19 @@ def check_connector_https_url_only(connector: Connector) -> bool:
...
+            continue
+        if "http://" in line:
+            files_with_http_url.add(str(filename))
```
evantahler, line 113: Yeah, I guess this all we need to check! 
Have you seen any false positives other than JSONschema?

evantahler (approved):

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
evantahler: A malicious person could do `"h"+"t"+"t"+"P"+":"`, and that is a known risk that I'm comfortable with, given the firewalls we will be adding.

bnchrch (commented):
I wanted to ask What's the intended goal of this particular check?

If it's to ensure we are more secure I have a few concerns. 

1. This could fail at the mention of http in a comment not logical use. 
2. In request or uri libraries sometimes the parameter to call using http might be used in an argument. Or obscured behind a flag. Both which potentially won't be caught with a line check
3. Not all code is checked in. Libraries are one example of code a developer can bring in which makes an http call. 

Is this something we could enforce at the allowed hosts level? Parse the options to ensure they define only https urls?

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
alafanechere: > Have you seen any false positives other than JSONschema?

I've seen a couple in test folders, expected records etc. This is why I ignored any path having `test` in it.

alafanechere:
> I wanted to ask What's the intended goal of this particular check?

@bnchrch sorry for the low context.
The idea of this check is not to prevent malicious use of HTTP, it's more an approach to check that a new connector is using https to request its source service. Something like guaranteeing that `url_base` is `https://`.
There are multiple false positives that we could fix along the way.  
I'm going to run these checks on a couple of GA connectors to make sure I'm not causing their build to fail.

The firewall project with `allowed_host` metadata + vulnerability scans will be a good guarantee that no shady traffic happens...

alafanechere:
@grishick do you think there's a similar `grep` like approach that could be taken for JDBC connectors?

alafanechere:
I had to make minor changes to the following GA connectors so that I don't break their builds:
- source-github
- source-google-sheets
- destination-s3

airbyte-integrations/connectors/source-google-search-console/source_google_search_console/streams.py
```diff
@@ -239,13 +239,13 @@ def get_updated_state(
 
         {
           "stream": {
-            "http://domain1.com": {
+            "https://domain1.com": {
```
alafanechere, line 242: Not bumping version: it's a docstring

airbyte-integrations/connectors/source-marketo/source_marketo/source.py
```diff
@@ -411,7 +411,7 @@ def get_json_schema(self) -> Mapping[str, Any]:
 class ActivityTypes(MarketoStream):
     """
     Return list of all activity types.
-    API Docs: http://developers.marketo.com/rest-api/lead-database/activities/#describe
+    API Docs: https://developers.marketo.com/rest-api/lead-database/activities/#describe
```
alafanechere, line 414: Not bumping version: it's a docstring

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
```diff
@@ -21,36 +23,32 @@ def check_documentation_file_exists(connector: Connector) -> bool:
...
     """Documentation guidelines are defined here https://hackmd.io/Bz75cgATSbm7DjrAqgl4rw"""
     follows_guidelines = True
     with open(connector.documentation_file_path) as f:
-        doc_lines = [l.lower() for l in f.read().splitlines()]
+        doc_lines = [line.lower() for line in f.read().splitlines()]
     if not doc_lines[0].startswith("# "):
         print("The connector name is not used as the main header in the documentation.")
         follows_guidelines = False
-    if connector.definition: # We usually don't have a definition if the connector is not published.
+    if connector.definition:  # We usually don't have a definition if the connector is not published.
```
jbfbell, line 35: super nitpik (not blocking): while this pr adds an additional space which is required for pep 8 inline comments, not using inline comments is preferred

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
```diff
@@ -84,6 +83,29 @@ def check_connector_icon_is_available(connector: Connector) -> bool:
     """
     return connector.icon_path.exists()
 
+
+def read_all_files_in_directory(directory: Path, path_patterns_to_ignore: Optional[Set[str]] = None) -> Iterable[Tuple[str, str]]:
+    path_patterns_to_ignore = path_patterns_to_ignore if path_patterns_to_ignore is not None else {}
+    for path in directory.rglob("*"):
+        ignore = bool([pattern for pattern in path_patterns_to_ignore if pattern in str(path)])
```
jbfbell, line 90: This is a little fragile because the `in` matches too much. In the`PATH_PATTERNS_TO_IGNORE_FOR_HTTPS_CHECKS` set you have `test` which would match `contest.py` or `protest.yml`. [Pathlib has a match method](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.match) which takes a regex which could be helpful here. Something like:
```python
ignore = any([path.match(pattern) for pattern in path_patterns_to_ignore])
```

bnchrch:
@alafanechere thanks for the explanation!

I wanted to ask if you had looked at any static analysis tool to fit this case for us?

For example https://github.com/PyCQA/bandit can scan to ensure all requests are done over ssl. 

This potentially could cover our use case and more.


Food for thought. If this pr needs to go in now, perhaps it's a future enhancement.

alafanechere:
I identified the list of connectors for which this check will fail.
I created an issue for GA connectors: https://github.com/airbytehq/airbyte/issues/22878
I'll update their code in separate PRs to make their test pass.

alafanechere:
@bnchrch and @jbfbell , I'd appreciate an additional review. 
I made some changes to make the "ignore file path" matching more robust.

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
```diff
@@ -93,9 +126,22 @@ def check_connector_https_url_only(connector: Connector) -> bool:
     Returns:
         bool: Wether the connector code contains only https url.
     """
-    # TODO implement
+    files_with_http_url = set()
+    for filename, line in read_all_files_in_directory(
```
bnchrch, line 130: @alafanechere I think a small improvement here would be to omit lines that are comments.

```python
def is_comment(line, file_path):
    language_comments = {
        ".py": "#",
        ".yml": "#",
        ".yaml": "#",
        ".java": "//",
        ".md": "<!--",
    }

    denote_comment = language_comments.get(file_path)
    if not denote_comment:
        return False

    trimmed_line = line.lstrip()    
    return trimmed_line.startswith(denote_comment)
```

bnchrch (approved):
I left one small improvement comment. But other than that looks good to me.

I had one question, did we consider turning this into a custom pylint check?

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
```diff
@@ -21,36 +23,33 @@ def check_documentation_file_exists(connector: Connector) -> bool:
...
     return connector.documentation_file_path.exists()
 
+
 def check_documentation_follows_guidelines(connector: Connector) -> bool:
     """Documentation guidelines are defined here https://hackmd.io/Bz75cgATSbm7DjrAqgl4rw"""
     follows_guidelines = True
     with open(connector.documentation_file_path) as f:
-        doc_lines = [l.lower() for l in f.read().splitlines()]
+        doc_lines = [line.lower() for line in f.read().splitlines()]
     if not doc_lines[0].startswith("# "):
```
jbfbell, line 32: This is probably fine, technically a tab is valid too after the `#`.

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
```diff
@@ -84,6 +84,39 @@ def check_connector_icon_is_available(connector: Connector) -> bool:
...
+    for path in directory.rglob("*"):
+        ignore_directory = any([ignored_directory in path.parts for ignored_directory in ignored_directories])
+        ignore_filename = any([path.match(ignored_filename_pattern) for ignored_filename_pattern in ignored_filename_patterns])
+        ignore = ignore_directory or ignore_filename
+        if path.is_file() and not ignore:
+            try:
+                for line in open(path, "r"):
+                    yield path, line
+            except UnicodeDecodeError:
+                continue
```
jbfbell, line 103: Its a good idea to at least log something here

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
```diff
@@ -93,9 +126,22 @@ def check_connector_https_url_only(connector: Connector) -> bool:
     Returns:
         bool: Wether the connector code contains only https url.
     """
-    # TODO implement
+    files_with_http_url = set()
+    for filename, line in read_all_files_in_directory(
+        connector.code_directory, IGNORED_DIRECTORIES_FOR_HTTPS_CHECKS, IGNORED_FILENAME_PATTERN_FOR_HTTPS_CHECKS
+    ):
+        if any([prefix in line.lower() for prefix in IGNORED_URLS_PREFIX]):
```
jbfbell, line 133: This will ignore a line has two (or more) urls where at least one url should be ignored, regardless of whether the other urls should be ignored.
```
This is ok http://localhost:8000 but this is not http://airbyte.io
```

bnchrch (changes requested):
Hey @alafanechere 

I wanted to reopen this after seeing the google sheets PR
https://github.com/airbytehq/airbyte/pull/22884#discussion_r1104740165

My main issue is that this method of finding http is a bit broad and leaky as it can catch either comments that include HTTP or sources that are trying to check their underlying data for HTTP.

This makes me want to ask how necessary is this for going in before we have a tool in place like snyk or bandit?

Rules in either of these are purpose built be able to catch HTTP calls at the library level.

If that's too heavy handed an alternative to bandit/snyk could be a custom pylint rule would be able to catch HTTP references, omit comments, and allow source authors to opt out using an inline ignore flag.

alafanechere:
@bnchrch the idea behind this PR is indeed in the linting realm: enforcing best practices, not checking actual network calls.
I don't think we're using pylint in the codebase, and it would fall short for java connectors. 
I do understand you're concerned about this new check.
IMO it could be removed once we have snyk and firewalls with "allowed_hosts` metadata.
In the meantime, I think it still provides value to ensure we're not releasing alpha connectors to the cloud that call HTTP base URLs.

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
alafanechere: Good catch, I'll try to work around this.

tools/ci_connector_ops/ci_connector_ops/qa_checks.py
alafanechere: Will do.

bnchrch (approved):
@alafanechere fair enough, It is a good alpha connectors check! 😄 

Approved.

Would you mind adding a TODO referencing the allowed_hosts or Snyk issue for later?

alafanechere:
> Would you mind adding a TODO referencing the allowed_hosts or Snyk issue for later?

Done here https://github.com/airbytehq/airbyte/blob/0100a4875da953c7ba831f89fc6f8bb4735aed18/tools/ci_connector_ops/ci_connector_ops/qa_checks.py#L188

octavia-approvington:
It's time <br/> fine lets go

octavia-approvington (approved):

PR #22856, opened 2023-02-11
Pre-check only suggested streams for new connections

### LINKED ISSUE #22851: A/B Test to disable suggesting all streams for new connections
After https://github.com/airbytehq/airbyte/pull/21577, we now have the ability for sources to suggest only those important streams for users setting up new connectors. Today, if a connector has not implemented `suggestedStreams`, all streams are selected by default.  We want to set up a test for a small group of users which swaps this behavior - what if no streams were suggested by default?

**Positive Hypothesis:**
* First sync success % likely goes up.  We are syncing less streams, and only the 'well-tested' popular streams
* First sync duraiton likely goes down.  If we sync less streams, it takes less time
* Users think more critically about the data they want to move, leading to more investment in the resulting destination's dataset
* We gain data about which streams matter, so that we can then populate `suggestedStreams` for more connectors

**Negative Hypothesis:**
* Adds additional friction for the first sync for most sources (as most sources don't yet have suggested streams)
* Some sources, e.g. databases, without a static list of streams, will likely never have any streams suggested by default in this new world.

I *think* we can do this test entirely in the front end.  

**The work to be done is:**
1. Set up the Launch Darkly feature flag and test group
3. Fire segment events for those users within the test group to store which streams they chose for each connector
    * We don't want data about users not in the test group
4. Report on this data in Metabase:
    * Did sync success rate go up or down for the test group?
    * Did sync duration go up or down for the test group?
    * Did we learn what streams are popular for the connectors and can we populate more `suggestedStreams`

evantahler:
Closes https://github.com/airbytehq/airbyte/issues/22851 (see <-- for description)
By default, if there are no SugestedStreams for the source, new connections will have no streams checked.

### DIFF
--- a/airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/helpers/CatalogConverter.java
+++ b/airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/helpers/CatalogConverter.java
@@ -138,14 +138,13 @@ public class CatalogConverter {
         .cursorField(stream.getDefaultCursorField())
         .destinationSyncMode(io.airbyte.api.model.generated.DestinationSyncMode.APPEND)
         .primaryKey(stream.getSourceDefinedPrimaryKey())
-        .selected(!suggestingStreams)
-        .suggested(true);
+        .selected(false)
+        .suggested(false);
 
     if (suggestingStreams) {
       if (suggestedStreams.contains(stream.getName())) {
         result.setSelected(true);
-      } else {
-        result.setSuggested(false);
+        result.setSuggested(true);
       }
     }
 


evantahler:
⚠️  Do not merge me until you get the ok from @michel-tricot and @YowanR!

pedroslopez (approved):
Code-wise looks good, though it appears there are some tests that may be expecting all the streams to be selected by default

```
expected: <[SchemaTableNamePair[schemaName=output_namespace_public, tableName=output_table_id_and_name], SchemaTableNamePair[schemaName=output_namespace_public, tableName=_airbyte_raw_output_table_id_and_name]]> but was: <[]>
```

timroes:
I'm a bit concerned merging this. We (unfortunately) allow at the moment in the FE to save a connection without any stream enabled for syncing. Having streams disabled by default, I fear will lead to a larger amount of users just clicking through without noticing and will create connections without any streams in it, which will just cause overhead and even more cause confusion to users why no data eventually was synced.

I filed https://github.com/airbytehq/airbyte/issues/22944 to track treating no streams as an error case, but I don't think we should merge this here, until we've fixed the FE side of this.

evantahler:
Closing in favor of A/B testing this in the front-end https://github.com/airbytehq/airbyte/issues/22851#issuecomment-1432414085

PR #22378, opened 2023-02-03
:window: :art: Connector builder navigation links

lmossman:
## What
Resolves https://github.com/airbytehq/airbyte/issues/21288

This PR adds several navigation links to the connector builder page in the airbyte webapp. Here are some screenshots of the new additions:
Screen Shot 2023-02-08 at 10 09 29 PM
Screen Shot 2023-02-08 at 10 09 42 PM
Screen Shot 2023-02-08 at 10 09 51 PM
Screen Shot 2023-02-08 at 10 09 18 PM

All of the above additions are hidden behind the `"connectorBuilder.showNavigationLinks"` experiment feature flag, which defaults to `false`. Used an Experiment instead of a Feature for this because we eventually want to roll this out to all OSS and Cloud users, following the guidelines here: https://www.notion.so/Feature-Flags-and-Experimentation-cd052d01ebed42e5b53047fe032348e7

## How
I went through and added comments explaining the various changes. See the linked ticket for a list of all the main things that were addressed in this PR.

**Note**: to test this in OSS (which is the only place it really makes sense to test, because the `/connector-builder` route is only enabled in OSS), you will need to create a file under `airbyte-webapp/` named `.experiments.json`, containing the following:
```
{
  "connectorBuilder.showNavigationLinks": true
}
```
This enables the experiment in OSS, so you can see all of the navigation links. If you don't do this, the only change you should see from prod is that the airbyte sidebar is shown next to the connector builder when going to that route.

This experiment will eventually be controlled LaunchDarkly as well, but that isn't important at the moment because we won't be enabling this for anyone in cloud until the connector builder route is enabled for cloud.

## Recommended reading order
1. `BuilderPrompt.tsx`
2. `SourceForm.tsx`
3. `ConnectorServiceTypeControl.tsx`
4. `SideBar.tsx`
5. `CreateConnector.tsx`
6. The rest

## 🚨 User Impact 🚨
Users for which

PR #22881, opened 2023-02-13
🪟 🎉 Link component

### LINKED ISSUE #18825: Create Link component
We currently don't have a fully working `Link` component in the app. We should create a component `Link`, that does the following things:

* Applies our regular link styling, that we want (underlining, color, etc.)
* Accepts either a `to`, `href` or `onClick` property, if
  * `to` is specified rendering a react-router link that will lead to the `to` page
  * `href` is specifies, treat it as an external link and render a `a` element. Also set `noopener noreferer` if it's an external domain not under our control
  * `onClick` render `button` element (that just looks like a link), triggering the `onClick`.
* Should support `openInNewWindow` to set `target="_blank"` for the `to` and `href` case.

joeykmh:
## What
Closes https://github.com/airbytehq/airbyte/issues/18825

- Adds two new components:
   - `<Link>`, which accepts a `to: string` property for internal `react-router`-based links
   - `<ExternalLink>`, which accepts a `href: string` property for external links
- Adds a `no-restricted-imports` rule so that developers cannot `import { Link } from "react-router-dom"`.
- Replaces the usage of `components/common/Link` and `react-router-dom/Link` with the new components
    - ~~To avoid too many merge conflicts, I will remove `components/common/Link` and its two remaining references in `Sidebar` once https://github.com/airbytehq/airbyte/pull/21227#pullrequestreview-1291952384 is merged~~ addressed in 9bb49b6d58ce02d65672887dc6127c9a23974cb3 

## Considerations
I considered creating a single component for both internal and external linking, but decided against it because:
- It's convoluted to get TypeScript to work with discriminated unions (e.g. `Link: React.FC<ChildrenWithProps<InternalLinkProps | ExternalLinkProps>>` gets ugly) 
- It didn't save much duplicate code anyway (~10 lines) and created indirection
- Any future changes to how we handle internal & external links are much easier to implement this way


airbyte-webapp/src/components/ui/Link/Link.module.scss
```diff
@@ -0,0 +1,15 @@
...
+
+.link {
+  color: colors.$grey-300;
+  text-underline-offset: 2px;
+
+  &--primary {
+    color: colors.$blue-400;
+    text-decoration: none;
+
+    &:hover {
```
timroes, line 11: Wouldn't we want any hover effect for normal links as well? I think we could make them slightly darker on hover?

airbyte-webapp/src/components/ui/Link/Link.tsx
```diff
@@ -0,0 +1,36 @@
...
+  to,
+  opensInNewTab = false,
+  variant = "default",
+  ...props
+}) => {
+  return (
+    <ReactRouterLink
+      {...props}
+      className={getLinkClassNames({ className, variant })}
+      rel={opensInNewTab ? "noopener noreferrer" : undefined}
```
timroes, line 29: I don't think we need this here, since we'd always have internal links only in this component. (If we decide to keep, I think the same as my comment above applies).

timroes (approved):
Code LGTM (aside some minor comments I left). I checked Storybook and checked a couple of pages the component is used and all seems good to me.

teallarson (approved):
LGTM -- looked at code and at Storybook

+1 to Tim's questions

airbyte-webapp/src/components/ui/Link/Link.module.scss
josephkmh: I agree, but since @Upmitt is working on standardizing text colors & inline link styling, I would prefer to wait for that before changing the behavior. At the moment, most of our "default" light gray links don't have hover effects.

airbyte-webapp/src/components/ui/Link/ExternalLink.tsx
```diff
@@ -0,0 +1,29 @@
...
+  href,
+  variant,
+  ...props
+}) => {
+  return (
+    <a
+      {...props}
+      className={getLinkClassNames({ className, variant })}
+      href={href}
+      rel={opensInNewTab ? "noopener noreferrer" : undefined}
```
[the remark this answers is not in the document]
josephkmh: Yes good point, let's set it to `noreferrer` by default. This was a mistake/holdover from when I had a single component handling both internal/external links. Much easier this way 🙂  addressed in https://github.com/airbytehq/airbyte/pull/22881/commits/61e192743f181a8549fbb92d2cb463a8103afcff

airbyte-webapp/src/components/ui/Link/Link.tsx
josephkmh: Agreed, addressed in https://github.com/airbytehq/airbyte/pull/22881/commits/73704a4b477e376e51790b28162da90b05e0deeb

airbyte-webapp/src/components/ui/Link/ExternalLink.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+  title: "Ui/ExternalLink",
+  component: ExternalLink,
+} as ComponentMeta<typeof ExternalLink>;
+
+const Template: ComponentStory<typeof ExternalLink> = (args) => <ExternalLink {...args} />;
+
+export const Primary = Template.bind({});
+Primary.args = {
+  children: "External link",
+  href: "https://google.com",
```
edmundito, line 17: 😉 
```suggestion
  href: "https://airbyte.com",
```

edmundito (commented):
Code ✅  Will defer approval

edmundito:
One error in CI
```
/actions-runner/_work/airbyte/airbyte/airbyte-webapp/src/components/ui/DropdownMenu/DropdownMenu.tsx
  5:10  error  'Link' import from 'react-router-dom' is restricted. Please use `import { Link, ExternalLink } from "components/ui/Link";` instead  no-restricted-imports```

josephkmh:
Well at least the linting rule is working 😅 

Looks like these changes conflict with https://github.com/airbytehq/airbyte/pull/22378 so I'll have to find some time to rework `DropdownMenu` to use our new `Link`/`ExternalLink` components.

timroes:
Superseded by https://github.com/airbytehq/airbyte-platform/pull/89

Issue #22878, opened 2023-02-13
Remove `http://` urls from GA connectors


alafanechere:
https://github.com/airbytehq/airbyte/pull/22854 added a QA check to scan for `http://` URLs in connector code. 
We want to encourage https:// only URL in connector code.

The following GA connectors use `http://` URLs in their code: 
- source-github
- source-google-search-console
- source-google-sheets
- source-marketo
- source-stripe
- destination-s3

alafanechere:
✅ source-marketo
✅ source-google-sheets
✅ source-google-search-console

🔨 [source-stripe](https://github.com/airbytehq/airbyte/pull/22886): can't merge because build is not passing on master)
🔨 [source-github](https://github.com/airbytehq/airbyte/pull/22882): can't merge because build is not passing on master)
🔨 [destination-s3](https://github.com/airbytehq/airbyte/pull/22887): Can't merge because of a publish error

PR #22884, opened 2023-02-13
source-google-sheets: do not consume http spreadsheets

alafanechere:
## What
https://github.com/airbytehq/airbyte/issues/22878
Everything is in the title.
We want the new QA check (https://github.com/airbytehq/airbyte/pull/22854) to pass for source-google-sheets.
This connector allowed the consumption of `http` sheets.


airbyte-integrations/connectors/source-google-sheets/unit_tests/test_helpers.py
```diff
@@ -236,15 +236,11 @@ def test_get_spreadsheet_id(self):
         result = Helpers.get_spreadsheet_id(test_url)
         self.assertEqual("18vWlVH8BfjGa-gwYGdV1BjcPP9re66xI8uJK25dtY9Q", result)
 
-        test_url = "http://docs.google.com/spreadsheets/d/18vWlVH8BfjGegwY_GdV1BjcPP9re_6xI8uJ-25dtY9Q/"
+        test_url = "https://docs.google.com/spreadsheets/d/18vWlVH8BfjGegwY_GdV1BjcPP9re_6xI8uJ-25dtY9Q/"
```
bnchrch, line 239: Good catch here

airbyte-integrations/connectors/source-google-sheets/google_sheets_source/helpers.py
```diff
@@ -199,7 +199,7 @@ def row_contains_relevant_data(cell_values: List[str], relevant_indices: Iterabl
 
     @staticmethod
     def get_spreadsheet_id(id_or_url: str) -> str:
-        if re.match(r"(http://)|(https://)", id_or_url):
+        if re.match(r"(https://)", id_or_url):
```
bnchrch, line 202: @alafanechere I think this is an example where us scanning the source code falls down.

For example, like here, a source "MAY" need to check the contents of datapoint in a source for HTTP. We shouldn't prevent them from doing that.

bnchrch (commented):
Not approving just yet. Going to continue this convo is the main PR

sherifnada (approved):

airbyte-integrations/connectors/source-google-sheets/google_sheets_source/helpers.py
alafanechere: I don't really get why a source may need to check the contents of an HTTP data point and why we shouldn't prevent them from doing that.

bnchrch: Fair question! and a fair point. Let me give an example.

_**Warning Im going down a "might happen" path. Feel free to call out as this wont happen.**_

So say for example I have a connector for a product that can be self hosted. Think matterport, grafana, strapi, your own service.

In the `spec` for the connector I ask for a required field of `apiUrl`.

Because I'm self hosted I know that a customer could have airbyte-oss and my-self-hosted-product on the same machine. Meaning that the value of `apiUrl` could be either `http://localhost:3333/api` or `https://org.my-self-hosted-product.com/api`

I may look for that http definition to either Fail and say http is not supported for this connector unless its localhost

But I dont want to take this too far with "What ifs". So Ill throw an approve on this since you know the product better than I do, and merge if you think this is a non-issue 🙂

bnchrch (approved):

airbyte-integrations/connectors/source-google-sheets/google_sheets_source/helpers.py
alafanechere: I agree that this scenario can happen. What you mention exists for the `source-file` connector, which I ignored from the QA check. For google sheets this is unlikely to happen. I'd prefer us to be aware of which connector behaves like you mention and ignore them in the check if needed.

alafanechere:
/publish connector=connectors/source-google-sheets run-tests=false

octavia-approvington:

octavia-approvington (approved):

PR #22937, opened 2023-02-13
connector-acceptance-tests: support custom environment variables

alafanechere:
## What
After investigation with @rodireich I discovered that some of our java connectors require custom environment variables to behave as we expect them to behave in tests.

This PR allow passing custom environment variable in `acceptance-test-config.yaml` that will be passed to the container (connector) under test.

## How
* Expose a new `custom_environment_variables` configuration field in the acceptance test configuration model.
* Store this field value in a session fixture
* Use this new value to create instantiate the docker runner (in a fixture too)
* Call the docker run client with this new  field value 

## Recommended reading order
1. https://github.com/airbytehq/airbyte/blob/27ee556c2662372c9fe24e073d2eadd2808cc155/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/config.py#L203
2. https://github.com/airbytehq/airbyte/blob/27ee556c2662372c9fe24e073d2eadd2808cc155/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/conftest.py#L59
3. https://github.com/airbytehq/airbyte/blob/27ee556c2662372c9fe24e073d2eadd2808cc155/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/conftest.py#L148
4. https://github.com/airbytehq/airbyte/blob/27ee556c2662372c9fe24e073d2eadd2808cc155/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/utils/connector_runner.py#L104

## 🚨 User Impact 🚨
This should not break any existing configuration as the new field is optional and defaults to empty dict.

PR #21864, opened 2023-01-25
Source Sentry: add state persistence

roman-yermilov-gl:
## What
https://github.com/airbytehq/oncall/issues/1317

## How
Add state persistence for incremental streams

PR #22943, opened 2023-02-13
Acceptance Test: Fix states for `LEGACY` incremental syncs

sergio-ropero:
## What
The tests were failing for `LEGACY` states in the incremental mode.
Tested with Postgres Source.

## How
To fix the error in the acceptance test I had to rework on the `records_with_state` function, responsible of getting the record and state value.

The schema for `LEGACY` states was a bit different and I had to include new logic to fit them. Also, for Incremental states I have faced the error in with the value in the state is string instead of integer, so I had to include a casting function to sort it out. So far I have only included casting to `int`.

These are examples on the state data in the messages:
 ### LEGACY non CDC state data
```
{'cdc': False, 'streams': [{'stream_name': 'id_and_name', 'stream_namespace': 'public', 'cursor_field': ['id'], 'cursor': '3', 'cursor_record_count': 1}]}
```

### LEGACY CDC state data
```
{'cdc': False, 'cdc_state': {'state': {'{"schema":null,"payload":["postgres",{"server":"postgres"}]}': '{"transaction_id":null,"lsn":2101199320,"txId":1055370,"ts_usec":1676304204421970}'}}, 'streams': [{'stream_name': 'id_and_name', 'stream_namespace': 'public', 'cursor_field': ['id']}]}
```

Also, modified the `seed.sql` for Postgres as the tables for CDC **MUST** have a primary key.

## NOTES
When working in this issue, I realise that CDC state messages are adding `cdc: False` in the state message.
Not investigated if this only impact `LEGACY` state messages or it's happening to new state messages.

### DIFF
--- a/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py
+++ b/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py
@@ -121,7 +121,6 @@ def records_with_state(records, state, stream_mapping, state_cursor_paths) -> It
                     try:
                         for stream_state in state["streams"]:
                             if stream_state["stream_name"] == stream_name:
-                                print({stream_state["cursor_field"][0]: stream_state["cursor"]})
                                 state_value = cursor_field.parse(
                                     record={stream_state["cursor_field"][0]: stream_state["cursor"]}
                                 )
--- a/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py
+++ b/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py
@@ -84,6 +84,7 @@ def configured_catalog_for_incremental_fixture(configured_catalog) -> Configured
 
 
 def records_with_state(records, state, stream_mapping, state_cursor_paths) -> Iterable[Tuple[Any, Any, Any]]:
+    state_value = None
     """Iterate over records and return cursor value with corresponding cursor value from state"""
     for record in records:
         stream_name = record.record.stream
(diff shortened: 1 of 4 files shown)


rodireich:
@sergio-ropero : Augustin is adding environment variable support to SAT.
See #22937 
This was for an unrelated failing test in STREAM mode that is currently disabled in master.
I don't know if it has direct impact on this issue or fix. But FYI

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py
```diff
@@ -102,7 +103,29 @@ def records_with_state(records, state, stream_mapping, state_cursor_paths) -> It
                 # try second time as an absolute path in state file (i.e. bookmarks -> stream_name -> column -> value)
                 state_value = cursor_field.parse(record=state, path=state_cursor_paths[stream_name])
             except KeyError:
-                continue
+                # LEGACY states in incremental mode
+                if "cdc_state" in state.keys():
```
bnchrch, line 107: I'm not sure this is the right place to address the CDC state.

My guiding principle here is: The Airbyte protocol and a result the CAT's should not be aware of how an individual connector consumes state.

As I understand `cdc_state` is specific to postgres's implementation (or database implementations). Which violates the above principle.

As an alternative, `source-postgres` should own this if statement.

I believe `source-sentry` [has implemented a similar method](https://github.com/airbytehq/airbyte/pull/21864/files#diff-350b5fdafcf9ae2b90ed4ab49c4d8d323826a10f16054941f5c812da040f7712R98) to handle legacy issues by overriding cursor value in certain situations.

bnchrch (changes requested):
Hey Sergio, thanks for this change.

I left a recommended change in this review.

Though, I will admit I'm a bit shaky on the details of which test was failing and how.

Could you also post the failing test message and a reason why you think it should pass?

akashkulk:
I was looking into the state object and ideally 

> @sergio-ropero : Augustin is adding environment variable support to SAT. See #22937 This was for an unrelated failing test in STREAM mode that is currently disabled in master. I don't know if it has direct impact on this issue or fix. But FYI

+ 1. Maybe worth seeing if this fixes the given issue. Based on a conversation with Subodh - we really shouldn't have the ``LEGACY`` state anywhere and perhaps by overriding the environment variables, we will get an expected non-LEGACY state

sergio-ropero:
Looks like the issue is going to be resolved with this pr:
https://github.com/airbytehq/airbyte/pull/22937

In case it not, I will reopen it.

PR #22957, opened 2023-02-14
Bmoric/fetch timeout actor configuration

benmoriceau:
## What
* Update the SourceDefinition accessor in order to add the new column seconds_between_messages to the result. This is then propagate to the API which will be use in the worker.
* The default value is 10800 (60 * 60 * 3 = 3 hours in seconds) which is the default value for OSS
* The value is not added to the config on purpose. It is needed on `ServerApp.java` which will be remove soon.

## Recommended reading order
1. `ConfigPersistence.java` => fetch the new column
2. `DatabaseBeanFactory.java` => Inject the default value


gosusnp (approved):
LGTM, one question, it looks like we do not want to persist the timeout if not explicitly defined. In this case, would it be easier to apply the default where we use the value instead of whenever we read it?

benmoriceau:
> instead

The issue with that is that the default changes between OSS and cloud. I would prefer to use an env in order to define it, especially since it might change.

PR #22974, opened 2023-02-14
[Low-Code CDK] Add script to run low code unit tests and address issues with a few connectors

brianjlai:
## What
While going through the validation checklist for low-code to beta,  some manifests had a few one-off issues that should be addressed to incorporate the changes to the language. Before we push a release to Beta, we should merge this into the branch to retain compatibility

## How
In addition to some fixes, I also added a script to run any low-code connectors that have unit tests to validate they are also compatible with the changes.

## Recommended reading order
1. `low-code-unit-tests.sh`
2. connector changes


airbyte-integrations/connectors/source-greenhouse/source_greenhouse/manifest.yaml
```diff
@@ -125,13 +127,14 @@ definitions:
...
         $ref: "#/definitions/requester"
         path: "applications/{{ stream_slice.parent_id }}/demographics/answers"
-      partition_router:
-        class_name: source_greenhouse.components.GreenHouseSubstreamSlicer
-        parent_stream: "#/definitions/applications_stream"
-        request_cursor_field: "updated_after"
-        stream_slice_field: "parent_id"
-        cursor_field: "updated_at"
-        parent_key: "id"
+    incremental_sync:
```
brianjlai, line 130: after inspecting this connector a bit, this really should be an incremental_sync custom component instead of a custom partition_router because it has a `cursor_field`. In the existing code, we would inadvertently only perform full refreshes since we didn't seen `incremental_sync` component and therefore assumed it was only full refresh.

This also is more in line w/ our ideal structure where `incremental_sync` dictates the sync capabilities.

airbyte-integrations/connectors/source-posthog/source_posthog/manifest.yaml
```diff
@@ -103,7 +103,7 @@ definitions:
       type: CustomIncrementalSync
       class_name: source_posthog.components.EventsCartesianProductStreamSlicer
       cursor_field: timestamp
-      partition_routers:
+      stream_slicers:
```
brianjlai, line 106: This one is a bit gross, but because it overrides the CartesianProductStreamSlicer (which shouldn't even be in the language anymore), our underlying implementation detail (namely the `stream_slicers` field) bleeds into the code a little bit.

airbyte-integrations/connectors/source-braze/source_braze/datetime_incremental_sync.py
```diff
@@ -38,8 +38,8 @@ def _partition_daterange(self, start, end, step: datetime.timedelta):
         """
         Puts a step to each stream slice. `step` is a difference between start/end date in days.
         """
-        get_start_time = operator.itemgetter(self.stream_slice_field_start.eval(self.config))
-        get_end_time = operator.itemgetter(self.stream_slice_field_end.eval(self.config))
+        get_start_time = operator.itemgetter(self.partition_field_start.eval(self.config))
```
brianjlai, line 41: Fixes the component because the `DatetimeBasedCursor` this was originally implementing had changed a bit.

airbyte-integrations/connectors/source-gnews/unit_tests/test_wait_until_midnight_backoff.py
```diff
@@ -25,6 +25,6 @@
 def test_wait_until_midnight(test_datetime, test_name, now_utc, expected_backoff_time):
     test_datetime.utcnow = Mock(return_value=datetime.strptime(now_utc, DATE_FORMAT))
     response_mock = Mock()
-    backoff_stratery = WaitUntilMidnightBackoffStrategy(options={}, config={})
+    backoff_stratery = WaitUntilMidnightBackoffStrategy(parameters={}, config={})
```
maxi297, line 28: I seems like there was another connector like this (see [the commit I added in the branch](https://github.com/airbytehq/airbyte/pull/22974/commits/66e02098036d92efc463309b189ded773d857443))

brianjlai: Wait that is a really weird use of `CheckStream`, `source-freshdesk` isn't even a low-code connector so I had to do a double take about what we're doing. For some reason, on a python connector, the latest change swapped out the check logic for the declarative component. That seems like a very questionable decision, granted this change is minor so we can leave it in, but this really shouldn't have even been here in the first place

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_stream.py
```diff
@@ -39,11 +40,11 @@ class DeclarativeStream(Stream):
     _name: str = field(init=False, repr=False, default="")
     _primary_key: str = field(init=False, repr=False, default="")
     _schema_loader: SchemaLoader = field(init=False, repr=False, default=None)
-    stream_cursor_field: Optional[Union[List[str], str]] = None
+    stream_cursor_field: Optional[Union[InterpolatedString, str]] = None
```
brianjlai, line 43: This was an interesting gap. We used to hard code this to not support interpolation just a simple or nested cursor field. But this didn't align with the slicer's `cursor_field`. No connectors used a nested cursor field so far, but we can always extend this out as new functionality if we need it.

airbyte-integrations/connectors/source-gnews/unit_tests/test_wait_until_midnight_backoff.py
maxi297: Maybe `source-freshdesk` is not declarative but it uses `CheckStream` which is in the `airbyte_cdk.sources.declarative` package. I guess the issue is not exactly with our changes on the declarative language but more with how `source-freshdesk` is implemented

brianjlai: So this shouldn't have passed review in my opinion. A non-low code connector shouldn't be even using declarative components.

airbyte-integrations/connectors/source-square/source_square/manifest.yaml
```diff
@@ -27,9 +27,12 @@ definitions:
       bearer: "#/definitions/bearer_authenticator"
       oauth: "#/definitions/oauth_authenticator"
 
-    request_headers:
-      Square-Version: "2022-10-19"
-      Content-Type: "application/json"
+  #    Uncomment this block later. request_headers used to accidentally get overridden but this was fixed in beta release
+  #    However, because expected_records were originally generated with the buggy flow missing these, we need to
+  #    regenerate the expected records. Temporarily commenting it out to retain the old behavior until we do it.
```
brianjlai, line 32: fun stuff here... but it would appear there isn't a trivial way to generate new expected_records. The fix and path is clear to resolve this, but I don't really want to hold up the release of this on something like regenerating the expected records

brianjlai:
I messed up the merge from the dev branch this is off of, just gonna close this one out in favor of a clean PR here: https://github.com/airbytehq/airbyte/pull/23123

PR #22996, opened 2023-02-14
SAT ignored fields key for basic read

### LINKED ISSUE #21737: [SAT]  ignored_fields key for basic read test
Sometimes sources change data in records for each read, like a `last_refresh_time ` column.

See https://github.com/airbytehq/airbyte/pull/21419, https://github.com/airbytehq/airbyte/pull/21593 and https://github.com/airbytehq/airbyte/pull/21775 for examples

roman-yermilov-gl:
## What
*Add `ignored_fields` section for basic read*

## How
- *added new section to config*
- *added bypass reason for empty fields*
- *handled ignored fields in basic read and full refresh tests*
- *updated unittests and added new ones*
- *made configs backward compatible*


YowanR:
@roman-yermilov-gl @lazebnyi Looks like this is important for Stripe. @airbytehq/connector-operations are you able to help prioritize this review?

airbyte-integrations/connectors/source-faker/acceptance-test-config.yml
```diff
@@ -30,8 +30,10 @@ acceptance_tests:
         configured_catalog_path: integration_tests/configured_catalog.json
         ignored_fields:
           users:
-            - created_at
-            - updated_at
+            - name: created_at
+              bypass_reason: "fast changing data"
```
erohmensing, line 34: Why would this be fast changing?

airbyte-integrations/connectors/source-google-pagespeed-insights/acceptance-test-config.yml
```diff
@@ -35,4 +35,8 @@ acceptance_tests:
       - config_path: "secrets/config.json"
         configured_catalog_path: "integration_tests/configured_catalog.json"
         ignored_fields:
-          pagespeed: ["lighthouseResult", "analysisUTCTimestamp"] # This differs between two runs, as they are newly generated per run
+          pagespeed:
+            - name: lighthouseResult
+              bypass_reason: "This differs between two runs, as they are newly generated per run"
+            - name: analysisUTCTimestamp
+              bypass_reason: "This differs between two runs, as they are newly generated per run"
```
erohmensing, line 42: This seems legit for `analysisUTCTimestamp`, why is it the case for `lighthouseResult`?

airbyte-integrations/connectors/source-zoom/acceptance-test-config.yml
```diff
@@ -37,8 +37,10 @@ tests:
...
-        "meetings":
-          - "start_url"
-        "webinars":
-          - "start_url"          
+        meetings:
+          - name: start_url
+            bypass_reason: "fast changing data"
+        webinars:
+          - name: start_url
+            bypass_reason: "fast changing data"
```
erohmensing, line 45: Does our test account have fast-changing list of meetings and webinars?

erohmensing (commented):
@YowanR I'm concerned about us putting "fast-changing data" as the reason for each ignored column that we don't currently have a commented reason for. Once it gets merged, I don't forsee someone coming back to check them, so I'd like to get it right the first time. Id like some way to confirm that that's the reason. For some of them (parameters like last accessed at, and data like instagram feeds/stories which are populated by some algorithm we don't control, I think it makes sense, but I think there are others that are less clear. 

This is only a review of adding the bypass reasons as a result, I can review the implementation code separately

airbyte-integrations/connectors/source-stripe/acceptance-test-config.yml
```diff
@@ -44,6 +44,8 @@ tests:
...
       configured_catalog_path: "integration_tests/connected_account_configured_catalog.json"
       ignored_fields:
-        "invoices":
-          - invoice_pdf
-          - hosted_invoice_url
+        invoices:
+          - name: invoice_pdf
+            bypass_reason: "fast changing data"
+          - name: hosted_invoice_url
+            bypass_reason: "fast changing data"
```
evantahler, line 51:
```suggestion
        invoices:
          - name: invoice_pdf
            bypass_reason: "URL changes upon each request for privacy/security"
          - name: hosted_invoice_url
            bypass_reason: "URL changes upon each request for privacy/security"
```

airbyte-integrations/connectors/source-instagram/acceptance-test-config.yml
```diff
@@ -39,21 +39,35 @@ acceptance_tests:
...
-            - profile_views
-            - reach
-            - text_message_clicks
-            - website_clicks
-            - impressions_week
-            - reach_week
-            - impressions_days_28
-            - reach_days_28
-            - online_followers
+          user_insights:
```
evantahler, line 42: Which user are we using here?  If this is a fake/test user, this data shouldn't change.  If this is a "real" user, then this information would be changing, but then is inappropriate for this test (not a sandbox)

airbyte-integrations/connectors/source-azure-table/acceptance-test-config.yml
```diff
@@ -24,4 +24,6 @@ tests:
     - config_path: "secrets/config.json"
       configured_catalog_path: "integration_tests/configured_catalog.json"
       ignored_fields:
-        "AirbyteTest": ["record"]
+        AirbyteTest:
+          - name: record
+            bypass_reason: "fast changing data"
```
evantahler, line 29: If we really are using a static dataset, this should't be changing?

airbyte-integrations/connectors/source-clickup-api/acceptance-test-config.yml
```diff
@@ -17,7 +17,11 @@ acceptance_tests:
     - config_path: secrets/config.json
       configured_catalog_path: integration_tests/configured_catalog.json
       ignored_fields:
-        "team": ["members", "last_active"]
+        team:
+          - name: members
```
evantahler, line 21: why would our list of team members be changing?

airbyte-integrations/connectors/source-coinmarketcap/acceptance-test-config.yml
```diff
@@ -40,15 +40,26 @@ acceptance_tests:
         timeout_seconds: 1800
         ignored_fields:
           categories:
-            - "avg_price_change"
-            - "market_cap"
-            - "market_cap_change"
-            - "volume"
-            - "volume_change"
+            - name: avg_price_change
```
evantahler, line 43: This source uses real market data, and that's probably always changing... so 👍

airbyte-integrations/connectors/source-dockerhub/acceptance-test-config.yml
```diff
@@ -21,4 +21,8 @@ tests:
       configured_catalog_path: "integration_tests/configured_catalog.json"
       # testing sequentially for same results can fail because of pull counts increasing for an image between runs
       ignored_fields:
-        "docker_hub": ["pull_count", "last_updated"]
+        docker_hub:
+          - name: pull_count
+            bypass_reason: "fast changing data"
+          - name: last_updated
+            bypass_reason: "fast changing data"
```
evantahler, line 28: This seems to imply we are testing against a real docker image that someone's using?

airbyte-integrations/connectors/source-shopify/acceptance-test-config.yml
```diff
@@ -36,4 +36,10 @@ tests:
...
       timeout_seconds: 7200
       ignored_fields:
-        "products_graph_ql": ["onlineStorePreviewUrl", "updatedAt", "totalInventory"]
+        products_graph_ql:
+          - name: onlineStorePreviewUrl
+            bypass_reason: "fast changing data"
+          - name: updatedAt
+            bypass_reason: "fast changing data"
+          - name: totalInventory
+            bypass_reason: "fast changing data"
```
evantahler, line 45: What causes these properties to change?  Isn't our test storefront/products static?  Does SAT "make a purchase" and change the inventory?

airbyte-integrations/connectors/source-stripe/acceptance-test-config.yml
evantahler: For streams that really have changing data, we need to add better reasons to explain why

evantahler (changes requested):
👍 on the SAT change

Like @erohmensing, I'm also worried about how many properties we are skipping here.  I added comments below on the streams I think which shouldn't have changing data... and perhaps this means that we aren't using good sandboxes.

For the remaining bypassed properties, please add more detail to why that specific column is changing and why.  I added an example of this to Stripe.

lazebnyi:
@evantahler @erohmensing @roman-yermilov-gl I'm not sure if we should overwrite the old version format to `ignore_fields` during a full upgrade for all connectors. This change can affect many connectors, which can be in different operational states. And we'll get stuck fixing those connectors or in try to add bypass reasons connectors. 

My proposal is add new version only for high strictness SAT. WDYT?

roman-yermilov-gl:
@lazebnyi @evantahler @erohmensing 
As I see from the source code we are really not overriding old configs. If I override configs, then I also must migrate those configs to new version, otherwise we will have a mix of the old config version with new style `ignored_fields`. It is not allowed (and logically incorrect) to have such a mix of versions in one config so it is better to leave configs as is and make the feature backward compatible. PR is updated

evantahler:
removing my review in favor of the other folks on the Connector Operations tea

erohmensing:
> I'm not sure if we should overwrite the old version format to ignore_fields during a full upgrade for all connectors. This change can affect many connectors, which can be in different operational states. And we'll get stuck fixing those connectors or in try to add bypass reasons. My proposal is add new version only for high strictness SAT. WDYT?

> As I see from the source code we are really not overriding old configs. If I override configs, then I also must migrate those configs to new version, otherwise we will have a mix of the old config version with new style ignored_fields. It is not allowed (and logically incorrect) to have such a mix of versions in one config so it is better to leave configs as is and make the feature backward compatible. PR is updated

@lazebnyi @roman-yermilov-gl As far as I understand it, the proposal is to add the new `ignored_fields` format to multiple types of tests only in the new format, while keeping the old way for the old configs that have not been migrated. Please let me know if my understanding is correct: 
* Connectors that currently use `ignored_fields` with an unmigrated config will be unaffected
* When those connectors try to change their format to the new one (most likely during a switch to high strictness), they will have to implement bypass reasons for their `ignored_fields` (is this enforced in code, or something we should manually enforce?)

In any case, I do think this is a better approach than trying to migrate all at once and this PR being blocked. Let's get this in to unblock the cases that need it and connector ops can be on the look out for configs that don't yet have bypass_reasons when they should (similar to how we now to `allowedHosts`). Maybe we can migrate one connector in this PR so that we can both test the changes and use it as an example?

roman-yermilov-gl:
> > I'm not sure if we should overwrite the old version format to ignore_fields during a full upgrade for all connectors. This change can affect many connectors, which can be in different operational states. And we'll get stuck fixing those connectors or in try to add bypass reasons. My proposal is add new version only for high strictness SAT. WDYT?
> 
> > As I see from the source code we are really not overriding old configs. If I override configs, then I also must migrate those configs to new version, otherwise we will have a mix of the old config version with new style ignored_fields. It is not allowed (and logically incorrect) to have such a mix of versions in one config so it is better to leave configs as is and make the feature backward compatible. PR is updated
> 
> @lazebnyi @roman-yermilov-gl As far as I understand it, the proposal is to add the new `ignored_fields` format to multiple types of tests only in the new format, while keeping the old way for the old configs that have not been migrated. Please let me know if my understanding is correct:
> 
> * Connectors that currently use `ignored_fields` with an unmigrated config will be unaffected
> * When those connectors try to change their format to the new one (most likely during a switch to high strictness), they will have to implement bypass reasons for their `ignored_fields` (is this enforced in code, or something we should manually enforce?)
> 
> In any case, I do think this is a better approach than trying to migrate all at once and this PR being blocked. Let's go ahead and work to get this in get this in to unblock the cases that need it and connector ops can be on the look out for configs that don't yet have bypass_reasons for their ignored_fields when they should (similar to how we now to `allowedHosts`). Maybe we can migrate one connector in this PR so that we can both test the changes and use it as an example?

**- Connectors that currently use ignored_fields with an unmigrated config will be unaffected** - yes

**- When those connectors try to change their format to the new one (most likely during a switch to high strictness), they will have to implement bypass reasons for their ignored_fields (is this enforced in code, or something we should manually enforce?)** - yes, for now there is a special function that reformats a config from old version to new one but it is not called for new configs, so it is required for connector to have `bypass_reason` for `ignored_fields`

**Maybe we can migrate one connector in this PR so that we can both test the changes and use it as an example?** - I tested this feature on the Stripe connector so I already have it there. I can update Stripe PR after this one will be merged

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_full_refresh.py
```diff
@@ -84,13 +84,12 @@ def assert_two_sequential_reads_produce_same_or_subset_records(
 
     def test_sequential_reads(
         self,
-        inputs: ConnectionTestConfig,
         connector_config: SecretDict,
         configured_catalog: ConfiguredAirbyteCatalog,
+        ignored_fields: Optional[Mapping[str, List[IgnoredFieldsConfiguration]]],
```
erohmensing, line 89:
```suggestion
        ignored_fields: Optional[Mapping[str, List[IgnoredFieldsConfiguration]]] = {},
```
Do we need to add a default either here or in the method? If it's optional I wonder what will happen if its not declared, since we used to set a default [here](https://github.com/airbytehq/airbyte/pull/22996/files?diff=unified&w=0#diff-82c6340d6fd6cc2a860e7a6080d0eabf830e4e4504a3a10b6f5ccf08818be126L93)

airbyte-integrations/bases/connector-acceptance-test/unit_tests/test_core.py
```diff
@@ -277,30 +278,94 @@ def test_configured_catalog_fixture(mocker, test_strictness_level, configured_ca
...
-    "schema, record, expectation",
+    "schema, ignored_fields, expect_records_config, record, expected_records_by_stream, expectation",
     [
-        ({"type": "object"}, {"aa": 23}, does_not_raise()),
-        ({"type": "object"}, {}, does_not_raise()),
+        (
+            {"type": "object"},
+            {},
+            ExpectedRecordsConfig(path="foobar"),
+            {"aa": 23}, {}, does_not_raise()
```
erohmensing, line 287:
```suggestion
            {"aa": 23}, 
            {}, 
            does_not_raise()
```
Nit: just for clarity/consistency

airbyte-integrations/bases/connector-acceptance-test/unit_tests/test_core.py
```diff
@@ -277,30 +278,94 @@ def test_configured_catalog_fixture(mocker, test_strictness_level, configured_ca
...
         # Recharge shop stream case
         (
             {"type": "object", "properties": {"shop": {"type": ["null", "object"]}, "store": {"type": ["null", "object"]}}},
+            {},
+            ExpectedRecordsConfig(path="foobar"),
             {"shop": {"a": "23"}, "store": {"b": "23"}},
+            {},
+            does_not_raise(),
+        ),
+        # Fail when expected and actual records are not equal
```
erohmensing, line 330: These are great for understanding the behavior 👍🏻

airbyte-integrations/bases/connector-acceptance-test/unit_tests/test_core.py
```diff
@@ -277,30 +278,94 @@ def test_configured_catalog_fixture(mocker, test_strictness_level, configured_ca
...
+            does_not_raise(),
+        ),
+        # Fail when expected and actual records are not equal and exact_order=True
+        (
+            {"type": "object"},
+            {},
+            ExpectedRecordsConfig(extra_fields=False, exact_order=True, extra_records=True, path="foobar"),
+            {"constant_field": "must equal", "fast_changing_field": [{"field": 2}]},
+            {"test_stream": [{"constant_field": "must equal", "fast_changing_field": [{"field": 1}]}]},
+            pytest.raises(AssertionError, match="Stream test_stream: Mismatch of record order or values"),
```
erohmensing, line 355: I'm not sure if I understand why this fails for the exact order one. Shouldn't it fail because the values don't match? An exact order test would be if we had multiple records output for the stream, which match the initial ones but are out of order

erohmensing (changes requested):
This is looking pretty good! For reference, what I was concerned about was with regards to how we were approaching enforcing bypass reasons (or not). Looks like we have taken Serhii's suggestion of only doing so for high strictness, which I think makes sense. Then we can fix those that are already high strictness and don't have bypass reasons (as they'll start failing), and handle the others as we migrate them to high strictness. This is similar to how we approach bypass reasons for expected_records. 

Most of my comments are just questions or suggestions - the current blocking items are 
* Change to a minor bump instead of a patch
* Implement unit tests for the fixture 
* Making sure we continue to handle ignored_fields not being set in test_sequential_reads

airbyte-integrations/bases/connector-acceptance-test/unit_tests/test_core.py
roman-yermilov-gl: I need this test to ensure that my code works in `if exact_order:` branch

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
```diff
@@ -885,10 +896,14 @@ def compare_records(
                     break
                 if extra_fields:
                     r2 = TestBasicRead.remove_extra_fields(r2, r1)
+                if ignored_fields:
+                    delete_fields(r1, ignored_fields)
+                    delete_fields(r2, ignored_fields)
```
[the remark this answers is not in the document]
roman-yermilov-gl: I just applied an action that I only needed

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/conftest.py
```diff
@@ -199,6 +200,22 @@ def empty_streams_fixture(inputs, test_strictness_level) -> Set[EmptyStreamConfi
     return empty_streams
 
 
+@pytest.fixture(name="ignored_fields")
+def ignored_fields_fixture(inputs, test_strictness_level) -> Optional[Mapping[str, List[IgnoredFieldsConfiguration]]]:
```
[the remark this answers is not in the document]
roman-yermilov-gl: done

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/config.py
```diff
@@ -110,6 +110,16 @@ def __hash__(self):  # make it hashable
...
 
 
+class IgnoredFieldsConfiguration(BaseConfig):
+    name: str
+    bypass_reason: Optional[str] = Field(default=None, description="Reason why this field is considered ignored.")
+
+
+ignored_fields: Optional[Mapping[str, List[IgnoredFieldsConfiguration]]] = Field(
+    description="For each stream, list of fields path ignoring in sequential reads test"
+)
```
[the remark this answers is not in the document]
roman-yermilov-gl: It has the same field structure for both test configs, so I thought I can declare it one time

airbyte-integrations/bases/connector-acceptance-test/Dockerfile
```diff
@@ -33,7 +33,7 @@ COPY pytest.ini setup.py ./
 COPY connector_acceptance_test ./connector_acceptance_test
 RUN pip install .
 
-LABEL io.airbyte.version=0.6.1
+LABEL io.airbyte.version=0.6.2
```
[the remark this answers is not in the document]
roman-yermilov-gl: done

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_full_refresh.py
roman-yermilov-gl: I assigned default value in the fixture

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/utils/compare.py
```diff
@@ -87,12 +95,7 @@ def make_hashable(obj, exclude_fields: List[str] = None) -> str:
     """
     if isinstance(obj, Mapping):
         # If value is Mapping, some fields can be excluded
-        exclude_fields = exclude_fields or []
-        for field in exclude_fields:
-            try:
-                dpath.util.delete(obj, field)
-            except dpath.exceptions.PathNotFound:
-                pass
+        delete_fields(obj, exclude_fields or [])
```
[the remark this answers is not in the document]
roman-yermilov-gl: done

airbyte-integrations/bases/connector-acceptance-test/unit_tests/test_core.py
roman-yermilov-gl: done

docs/connector-development/testing-connectors/connector-acceptance-tests-reference.md
```diff
@@ -143,23 +143,25 @@ These backward compatibility tests can be bypassed by changing the value of the
...
+| `empty_streams`                           | array of objects | \[\]                                        | List of streams that might be empty with a `bypass_reason`                                                    |
+| `empty_streams[0].name`                   | string           |                                             | Name of the empty stream                                                                                      |
+| `empty_streams[0].bypass_reason`          | string           | None                                        | Reason why this stream is empty                                                                               |
+| `ignored_fields[stream][0].name`          | string           |                                             | Name of the ignored field                                                                                      |
+| `ignored_fields[stream][0].bypass_reason` | string           | None                                        | Reason why this field is ignored                                                                               |
```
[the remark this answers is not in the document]
roman-yermilov-gl: done

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_full_refresh.py
erohmensing: Thanks, I saw that - you were a step ahead of me while I was reviewing 😄

airbyte-integrations/bases/connector-acceptance-test/unit_tests/test_core.py
erohmensing: Got it - I misinterpreted this as a new test for the exact_order behavior

airbyte-integrations/bases/connector-acceptance-test/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.6.2
```
erohmensing, line 3:
```suggestion
## 0.7.0
```

erohmensing (approved):
Ready to see this in action in Stripe! Updated the changelog, should be ready for publishing

evantahler (approved):

roman-yermilov-gl:

roman-yermilov-gl:
Publish was successful except updating publish comment in PR: `Error: API rate limit exceeded for installation ID 6913263.`

PR #22068, opened 2023-01-30
🪟 🧹 Optimize`destination` over fetching - pass `destination` through react-router outlet context

dizel852:
## What
Resolves https://github.com/airbytehq/airbyte/issues/21667

## How
Fetch the `destination` object once and pass it through the react-router `<Outlet />` context

PR #20845, opened 2022-12-22
🪟 🧹 Reorganize connections pages

edmundito:
## What
Reorganize the new connections pages with better names, less folder nesting, and unified routing component

## How
The reasons behind this change were to ensure that all the routes were defined under a single file (ConnectionRoutes.tsx), simplify the folder structure, and clean up any areas (names, confusing imports)

The Changes include:
* Moves the pages from `ConnectionPage` to `connections/` to match route
* Replace `Tab` with `Page` for all pages
* Update route syntax to use the new react-router `<Outlet />` component
* Cleans up an issue with the job logs

## Recommended reading order
1. `pages/connections/ConnectionsRoutes.tsx`

PR #23002, opened 2023-02-14
Refactor Source Item Page to match Destination Item Page, don't load source twice

### LINKED ISSUE #21663: Refactor Source Item Pages
Refactor the source item & settings page to more closely match the architecture of the destination item and settings pages.

We can load the Source top-level and provide it in the route context instead of prop-drilling. This doesn’t get us anything but enables the refactor.
`<Outlet context={{ source }} />`
`const { source } = useOutletContext<{ source: SourceRead }>();`

We’re loading all the connections for a source, but we only need the list on the Source Item Page. The source settings page uses the connection list but only for the delete call. If we can fold the connections list call into the delete source call we can isolate it to just the sources item page.

The source components themselves should be refactored to match the file structure of destination.

krishnaglick:
## What
Resolves https://github.com/airbytehq/airbyte/issues/21663
Converts the source item page to be more like destination item page.

## How
The source item page had its routes defined in the page instead of in the routes. I've pulled the routes up and now use outlet to render the appropriate page. Additionally I've leveraged the outlet context so we aren't loading the source on both pages.


airbyte-webapp/src/packages/cloud/cloudRoutes.tsx
```diff
@@ -57,11 +67,19 @@ const MainRoutes: React.FC = () => {
           <Route path={DestinationPaths.NewDestination} element={<CreateDestinationPage />} />
           <Route path={DestinationPaths.NewConnection} element={<CreateConnectionPage />} />
           <Route path={DestinationPaths.Root} element={<DestinationItemPage />}>
-            <Route path={DestinationPaths.Settings} element={<DestinationSettingsPage />} />
             <Route index element={<DestinationOverviewPage />} />
+            <Route path={DestinationPaths.Settings} element={<DestinationSettingsPage />} />
```
krishnaglick, line 71: Moved this so the index route is at the "top" of its section.

airbyte-webapp/src/packages/cloud/cloudRoutes.tsx
```diff
@@ -57,11 +67,19 @@ const MainRoutes: React.FC = () => {
...
+          </Route>
+        </Route>
+        <Route path={RoutePaths.Source}>
+          <Route index element={<AllSourcesPage />} />
+          <Route path={SourcePaths.NewSource} element={<CreateSourcePage />} />
+          <Route path={SourcePaths.NewConnection} element={<CreateConnectionPage />} />
+          <Route path={SourcePaths.Root} element={<SourceItemPage />}>
+            <Route index element={<SourceOverviewPage />} />
+            <Route path={SourcePaths.Settings} element={<SourceSettingsPage />} />
```
krishnaglick, line 80: Now using source paths.

airbyte-webapp/src/packages/cloud/cloudRoutes.tsx
```diff
@@ -17,22 +17,32 @@ import { CompleteOauthRequest } from "views/CompleteOauthRequest";
...
 import { LDExperimentServiceProvider } from "./services/thirdParty/launchdarkly";
 import { VerifyEmailAction } from "./views/FirebaseActionRoute";
-import { RoutePaths, DestinationPaths } from "../../pages/routePaths";
+import { RoutePaths, DestinationPaths, SourcePaths } from "../../pages/routePaths";
 
 const MainView = React.lazy(() => import("packages/cloud/views/layout/MainView"));
 const WorkspacesPage = React.lazy(() => import("packages/cloud/views/workspaces"));
 const Auth = React.lazy(() => import("packages/cloud/views/auth"));
 const CreditsPage = React.lazy(() => import("packages/cloud/views/credits"));
+const SpeakeasyRedirectPage = React.lazy(() => import("pages/SpeakeasyRedirectPage"));
```
krishnaglick, line 26: Moved for better grouping.

airbyte-webapp/src/pages/SourcesPage/pages/SourceItemPage/SourceItemPage.tsx
```diff
@@ -34,10 +24,6 @@ const SourceItemPage: React.FC = () => {
   );
 
   const source = useGetSource(params.id || "");
-  const sourceDefinition = useSourceDefinition(source.sourceDefinitionId);
-
-  // We load only connections attached to this source to be shown in the connections grid
-  const { connections } = useConnectionList({ sourceId: [source.sourceId] });
```
krishnaglick, line 40: All this was moved into the SourceOverviewPage. Should improve the speed of the settings page!

airbyte-webapp/src/pages/SourcesPage/pages/SourceItemPage/components/SourceOverviewPage.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+import { SourceRead } from "core/request/AirbyteClient";
+import { useConnectionList } from "hooks/services/useConnectionHook";
+import { useDestinationList } from "hooks/services/useDestinationHook";
+import { RoutePaths } from "pages/routePaths";
+import { useSourceDefinition } from "services/connector/SourceDefinitionService";
+
+import SourceConnectionTable from "./SourceConnectionTable";
+
+const SourceOverviewPage = () => {
```
krishnaglick, line 16: The overview logic from the source item page is in here now.

airbyte-webapp/src/pages/routes.tsx
```diff
@@ -53,7 +58,15 @@ const MainViewRoutes: React.FC = () => {
...
-          <Route path={`${RoutePaths.Source}/*`} element={<SourcesPage />} />
+          <Route path={RoutePaths.Source}>
+            <Route index element={<AllSourcesPage />} />
+            <Route path={SourcePaths.NewSource} element={<CreateSourcePage />} />
+            <Route path={SourcePaths.NewConnection} element={<CreateConnectionPage />} />
+            <Route path={SourcePaths.Root} element={<SourceItemPage />}>
+              <Route index element={<SourceOverviewPage />} />
+              <Route path={SourcePaths.Settings} element={<SourceSettingsPage />} />
+            </Route>
+          </Route>
```
krishnaglick, line 69: Same work as cloud routes

krishnaglick:
@edmundito Moved the pages to match the destination item pages!

edmundito:
@krishnaglick the structure of the page should be the following. The root folder should match the root path name, no `pages` subfolders, no nested component subfolders.

```
pages/
   source/
      AllSourcesPage/
      SourceItemPage/
      SourceOverviewPage/
      SourceSettingsPage/
```

airbyte-webapp/src/pages/source/SourceOverviewPage/sourceOverviewContext.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export const useSetupSourceOverviewContext: (id: string) => SourceOverviewContext = (id: string) => {
+  const source = useGetSource(id);
+  const sourceDefinition = useSourceDefinition(source.sourceDefinitionId);
+  // We load only connections attached to this source to be shown in the connections grid
+  const { connections } = useConnectionList({ sourceId: [source.sourceId] });
+
+  return { source, sourceDefinition, connections };
+};
```
krishnaglick, line 25: Since there's 3 items here referenced in 3 places, thought consolidating it would be useful.

krishnaglick:
@edmundito Things should be good w/ the file structure, and made a few more improvements around the place!

edmundito (approved):
Discussed issues with KC over Slack and re-tested all routes. LGTM.

airbyte-webapp/src/pages/connections/CreateConnectionPage/ConnectionCreateSourceForm.tsx
```diff
@@ -3,7 +3,7 @@ import { useLocation, useNavigate } from "react-router-dom";
 
 import { ConnectionConfiguration } from "core/domain/connection";
 import { useCreateSource } from "hooks/services/useSourceHook";
-import { SourceForm } from "pages/SourcesPage/pages/CreateSourcePage/components/SourceForm";
+import { SourceForm } from "pages/source/CreateSourcePage/SourceForm";
```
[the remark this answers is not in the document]
krishnaglick: https://github.com/airbytehq/airbyte/issues/23113

airbyte-webapp/src/pages/source/SourceOverviewPage/sourceOverviewContext.ts
josephkmh: I would actually rather be in favor of just calling `useGetSource()`, `useSourceDefinition()` and `useConnectionList()` in the two routes (it's only 2 places, right? ). I think creating a hook just to avoid writing 3 lines in each route adds more indirection. It actually creates more code and you also have to follow `useSourceOverviewContext()` > `SourceItemPage` > `useSetupSourceOverviewContext()` to find the underlying API hooks. IMO this will make it harder to add or remove more data-fetching later.

Performance-wise I don't think there's much to gain here, right? `react-query` will cache responses and deduplicate requests for us, so we shouldn't worry about calling e.g. `useConnectionList()` too many times! If we see too many network requests, we should tweak our cache/stale/revalidation times in `react-query`.

airbyte-webapp/src/pages/routes.tsx
josephkmh: @edmundito didn't you want to keep these nested routes in a separate file? https://github.com/airbytehq/airbyte/pull/20845#discussion_r1061328629

I don't really feel strongly either way, just as long as we're consistent. I have it on my to-do list to refactor the destination paths like you did in https://github.com/airbytehq/airbyte/pull/20845, so that's why I'm wondering how we want to proceed here

joeykmh (commented):
Nice work! My only concern is that we're creating a bit of unnecessary indirection with `useSetupSourceOverviewContext()`

airbyte-webapp/src/pages/source/SourceOverviewPage/sourceOverviewContext.ts
krishnaglick: react-query doesn't actually cache the result. If you navigate to the pages today it doubles up on the request.
The intent of this ticket was to consolidate the requests in the higher level component so it's not fetched twice. The page movement and re-org was just a useful result of that goal. I can rename it or something, but I felt it was worth consolidating into two hooks so the data had a consistent contract.

dizel852 (approved):
LGTM 👍 
Tested locally

airbyte-webapp/src/pages/source/SourceOverviewPage/sourceOverviewContext.ts
josephkmh: > react-query doesn't actually cache the result. If you navigate to the pages today it doubles up on the request.

I'm pretty sure it _does_ cache the result (or it should, that's kind of the main benefit of react-query) and you're just seeing duplicate requests because the `staleTime` is set to 0 by default. So the page is not suspended due to the second request, that just happens in the background. We could accomplish the same here by setting `staleTime` to something more reasonable (maybe 2 seconds) to avoid duplicate requests.

krishnaglick: I'm not sure I agree. I think this pattern gets the job done and it's an easy "Go-to Definition" click to see what's going on.
[DestinationItemPage](https://github.com/airbytehq/airbyte/blob/adc2c30ca3c6e9bb92c29ce6432c2644c8a5e1af/airbyte-webapp/src/pages/destination/DestinationItemPage/DestinationItemPage.tsx#L50) does this too, though with less data.

I think we may want to add a `staleTime` to the whole app. I assumed we had one, but that would be another ticket.

josephkmh: Yep, both patterns definitely work to avoid extra network requests. The recent PR to do the same thing in [DestinationItemPage](https://github.com/airbytehq/airbyte/pull/22068) is the same deal - we can either work around calling `useGetDestination()` multiple times, or we could adjust the `staleTime` of that query. In my opinion we should work with `react-query` rather than around it - feel free to do as you see fit here, I just wanted to suggest an alternative solution 🙂

airbyte-webapp/src/pages/source/SourceOverviewPage/sourceOverviewContext.ts
```diff
@@ -0,0 +1,25 @@
...
+  source: SourceRead;
+  sourceDefinition: SourceDefinitionRead;
+  connections: WebBackendConnectionListItem[];
+}
+
+export const useSourceOverviewContext = () => {
+  return useOutletContext<SourceOverviewContext>();
+};
+
+export const useSetupSourceOverviewContext: (id: string) => SourceOverviewContext = (id: string) => {
```
timroes, line 18: :art: Code style nitpick: We most often specify the return type on the arrow function, so it reads a bit shorter.

```suggestion
export const useSetupSourceOverviewContext = (id: string): SourceOverviewContext => {
```

airbyte-webapp/src/pages/source/SourceOverviewPage/sourceOverviewContext.ts
timroes: I'd personally also feel that creating new context for just sharing responses of a query might be a bit overhead and causing us an "context inflation", while we have react-query, which can handle this for us, and make the consuming component looks nicer, because it's clearer that it gets data from an API. I think though this might be an overall discussion, and Joey wanted to present something abour react-query soon, which I hope we can use as a discussion for clarifying which way we'd prefer in the future.

airbyte-webapp/src/pages/source/SourceOverviewPage/sourceOverviewContext.ts
krishnaglick: Did this locally incase of lint issue.

timroes:
Superseded by https://github.com/airbytehq/airbyte-platform/pull/92

PR #23014, opened 2023-02-14
source-postgres connector-acceptance-tests : Fix saved state json

rodireich:
## What
In order to allow the `future_state` step of Connector Acceptance Test, the saved state needs to match the expected format consumed by source-postgres.
With this change, alongside the environment variable enabling `USE_STREAM_CAPABLE_STATE`(#22937), all CAT test in source-postgres should go through

## How
Add to saved json state record all the expected fields.


alafanechere:
@rodireich I do get the expected behavior when I run `docker run --rm -e USE_STREAM_CAPABLE_STATE=true -v $(pwd)/secrets:/secrets -v $(pwd)/integration_tests:/integration_tests airbyte/source-postgres:1.0.44 read  --config /secrets/config.json --catalog /integration_tests/incremental_configured_catalog.json --state /integration_tests/abnormal_state.json`.

I'll need to investigate why acceptance tests are still failing.

rodireich:
@alafanechere where do you see a failure? The platform build below here is showing all green (I thought this was the indication).

alafanechere:
@rodireich in the test I ran above. 
I suspect the error is related to the fact that the type of the states values are `string` and the `integer` in records.

alafanechere:
@rodireich just to move forward on this PR: would you mind only keeping the state file changes but reverting the `acceptance-test-config.yml` change? Then we can merge and wait for #23098 to be fixed for re-enabling the bypassed tests.

alafanechere (approved):
I think we can merge it.
We did not fix the bug of https://github.com/airbytehq/airbyte/issues/23098 but we have an `abnormal_state.json` with the valid format.
(this file is only used in the tests we disabled)

rodireich:
Thanks @alafanechere 
the problem in #23098 seems like more of the same variety - unexpected state structure.
Not sure why there's a variety as the existing structure - even on this PR, works with CDK and python connectors.

rodireich:
Sidenote: I was unable to disable `incremental:` altogether using `bypass_reason`.

PR #23030, opened 2023-02-14
only run review requirements on source connectors

c-p-b:
## What
I think this is misconfigured to run on EVERY pull request review event instead of just pull request review events on certain paths. Need to verify that is the case.

## How
only fire on pull request review events for certain paths


bnchrch (approved):
Looks like this should solve it.

octavia-approvington:

octavia-approvington (approved):

PR #23031, opened 2023-02-14
🪟 🎨 Credits Page Usage Table Redesign Pt. 1

teallarson:
## What
closes https://github.com/airbytehq/airbyte-internal-issues/issues/3797

- Adds a feature flag to use going forward with the redesign.
- Adds additional columns to the Credits Usage Table on the credits page. (NOTE: Schedule data is being added to the API, but I added a static column for it to get the styling right for now)
- Also renames it to Billing page.

Screenshot 2023-02-16 at 3 55 20 PM

## Recommended reading order
1. `UsagePerConnectionTable.*`
2. others


airbyte-webapp/src/packages/cloud/views/credits/CreditsPage/components/UsagePerConnectionTable.tsx
```diff
@@ -105,6 +112,160 @@ const UsagePerConnectionTable: React.FC<UsagePerConnectionTableProps> = ({ credi
...
+                isAscending={sortOrder === SortOrderEnum.ASC}
+              >
+                <FormattedMessage id="credits.connection" />
+              </SortableTableHeader>
+            ),
+            meta: {
+              thClassName: classNames(styles.header, styles.light),
+              responsive: true,
+            },
+            cell: (props) => (
```
tealjulia, line 150: This (and the source + destination cells) should have a Tooltip if they have an ellipsis.

We have a pattern [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-webapp/src/components/connection/CatalogTree/next/CatalogTreeTableCell.tsx), which makes me wonder if this is something we want to extract for more generic use.  I am leaning yes?

Issue #17095, opened 2022-09-23
Add a "soft reset" feature


edgao:
## Tell us about the problem you're trying to solve
There might be cases where the raw table becomes out of sync with the final, normalized tables. For example:
* A bugged connector pushed data into the wrong location (e.g. https://github.com/airbytehq/airbyte/pull/17054) and the user is able to manually copy the `_raw` data back into where it's supposed to be located
* A normalization bug causes the final tables to be incorrect
* User accidentally deletes data from the normalized tables

In these cases, it would be nice to restore the normalized tables without doing a full reset. This allows us to skip the source -> destination phase, and rely instead on the `_raw` tables.

This also solves cases where dbt can't do normal schema evolutions, e.g. https://github.com/airbytehq/airbyte/issues/16668 (Snowflake restricts what types `ALTER COLUMN` can change).

## Describe the solution you’d like
Add a feature to do a `dbt run --full-refresh` in normalization, and add a button in  the UI to trigger it. Name needs some workshopping, maybe "dbt reset", "partial reset", "dbt rebuild"...

some discussion in https://airbytehq-team.slack.com/archives/C03C4AVJWG4/p1663941602536999

## Describe the alternative you’ve considered or used
Full connection reset. This incurs additional compute+time because it requires fetching the data out of the source and re-writing it into the destination. That's unnecessary, because the `_raw` tables already have that information.

## Are you willing to submit a PR?
sure :)

grishick:
troubleshooting the first attempt to rollback v1 protocol would have been easier if we had this feature

evantahler:
Linking https://github.com/airbytehq/airbyte/issues/17579 and https://github.com/airbytehq/airbyte/issues/17095

PR #23039, opened 2023-02-15
Add WorkspaceId to Job Reset configs

gosusnp:
## What
* We are not setting the workspaceId for resets, this may prevent some feature flags to work as expected.

## How
* Pass the workspaceId for resets.


benmoriceau (approved):

benmoriceau:
LGTM.

edgao (approved):
is there somewhere reasonable we can add a `Preconditions.checkNotNull(config.getWorkspaceId())`? Mostly thinking of if we ever want to add more jobs types (e.g. [soft reset](https://github.com/airbytehq/airbyte/issues/17095)) we might end up in the same position

gosusnp:
@edgao, I don't think we have an obvious place at the moment that would be catching all the cases, but it did came up in some discussions. We're expending on feature flagging, one idea is to align the context we use for tracing and feature flagging, we may be able to figure something out once we start that work.

edgao (approved):
makes sense, lgtm

octavia-approvington:
Lets do it! <br/> giddddy-up

octavia-approvington (approved):

PR #23045, opened 2023-02-15
Strict comparison normalization: Use launchdarkly feature flag

edgao:
## What
flag defined in https://app.launchdarkly.com/default/production/features/normalization.strictComparisonNormalizationEnabled

image tag continues to be an env var though, since it needs to be an arbitrary string

## How
if workspace is in either the env var, _or_ the feature flag, then use the strict_comparison normalization image. This will make the transition smoother (i.e. anyone already in the env var continues to work, and then we can add more workspaces via launchdarkly)

## Recommended reading order
yes, I recommend you read the files in order

## 🚨 User Impact 🚨
none

### DIFF
--- a/airbyte-featureflag/src/main/kotlin/Flags.kt
+++ b/airbyte-featureflag/src/main/kotlin/Flags.kt
@@ -19,6 +19,8 @@ object ApplyFieldSelection : EnvVar(envVar = "APPLY_FIELD_SELECTION")
 
 object PerfBackgroundJsonValidation : Temporary(key = "performance.backgroundJsonSchemaValidation")
 
+object StrictComparisonNormalizationEnabled : Temporary(key = "normalization.strictComparisonNormalizationEnabled")
+
 object FieldSelectionEnabled : Temporary(key="connection.columnSelection")
 
 // NOTE: this is deprecated in favor of FieldSelectionEnabled and will be removed once that flag is fully deployed.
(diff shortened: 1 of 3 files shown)


airbyte-featureflag/src/main/kotlin/Flags.kt
```diff
@@ -19,6 +19,8 @@ object ApplyFieldSelection : EnvVar(envVar = "APPLY_FIELD_SELECTION")
 
 object PerfBackgroundJsonValidation : Temporary(key = "performance.backgroundJsonSchemaValidation")
 
+object StrictComparisonNormalizationEnabled : Temporary(key = "normalization.strictComparisonNormalizationEnabled")
```
benmoriceau, line 22: I think that you're missing an update in the file flags.yml in order to be able to use this flag in Open source. If the plan is to remove the other flag, we should add it.

benmoriceau (approved):

airbyte-featureflag/src/main/kotlin/Flags.kt
edgao: is flags.yml just default values to handle the case where OSS users can't reach launchdarkly?

also, which other flag are you referring to? (the env var for image tag, or the env var for enabling strict_comparison)

edgao: added an entry to flags.yml, ptal just in case and then I'll mark as ready to ship

benmoriceau: yes, it is about having a configurable value in OSS.

benmoriceau (approved):

airbyte-featureflag/src/main/kotlin/Flags.kt
benmoriceau: If you want to test it locally, you'll need to add the path to the flag file in the application configuration like what is done in https://github.com/airbytehq/airbyte/pull/22397/files#diff-d0c67bf55b2b8412bf233bf144f0a4f74f60bb749faccbe0aba89c77730d12ba

edgao: tested locally with that setup, confirmed that normalization runs with `strict_comparison` after flipping `flags.yml` to true

edgao:
moved over to https://github.com/airbytehq/airbyte-platform/pull/98

Issue #22572, opened 2023-02-08
Connector builder projects: Add delete API for builder projects


flash1293:
Introduce APIs for deleting builder projects

```
POST /v1/api/connector_builder_projects/delete
{
    "workspaceId": "<workspace>",
    "builderProjectId": "<id>",
}
```

Fails when there is an actor definition id specified (this means the connector is published and can’t be deleted anymore) 400

For all apis the following constraints apply:
* Builder project doesn’t exist in workspace 404
* Name needs to be non-empty
* Draft can be “null” as well to unset it completely

## Acceptance criteria

* All APIs are functional as described above
* All APIs are unit-tested
* All introduced APIs need to be wrapped in cloud with Editor permission for workspace

Issue #22571, opened 2023-02-08
Connector builder projects: Add update API for builder projects


flash1293:
Introduce APIs for updating builder projects

```
POST /v1/api/connector_builder_projects/update
{
    "workspaceId": "<workspace>",
    "builderProjectId": "<id>",
    "builderProject": {
        "name": "My connector",
        "draftManifest": {...}
    }
}
```

Response 204 No content

* Fail with 400 if the builder project is not associated with that workspace ID
* Builder project doesn’t exist in workspace 404
* Name needs to be non-empty
* Draft can be “null” as well to unset it completely

## Acceptance criteria

* All APIs are functional as described above
* All APIs are unit-tested
* All introduced APIs need to be wrapped in cloud with Editor permission for workspace

PR #22879, opened 2023-02-13
Create builder projects table

flash1293:
Fixes https://github.com/airbytehq/airbyte/issues/22444

By introducing the `connector_builder_project` table and adding methods to read/write/delete projects to the `ConfigRepository` class, including tests for these methodes.

Not sure about the version for the platform - is this worth a minor version bump or should it be patch?

## How to test

* Run the tests introduced as part of the PR
* Build platform and start via docker, then log into internal database via `docker exec -ti airbyte-db psql -U docker -d airbyte` and inspect created table.

PR #23063, opened 2023-02-15
Connector builder api: create, update, delete

flash1293:
Based on https://github.com/airbytehq/airbyte/pull/22879

Fixes https://github.com/airbytehq/airbyte/issues/22570
Fixes https://github.com/airbytehq/airbyte/issues/22571
Fixes https://github.com/airbytehq/airbyte/issues/22572

This PR is adding the API layer for creating, updating and deleting builder projects based on the new model. This includes the following parts:
* Extend openapi spec for the server (`airbyte-api/src/main/openapi/config.yaml`)
* Implement API controller (`airbyte-server/src/main/java/io/airbyte/server/apis/ConnectorBuilderProjectApiController.java`) which is forwarding everything to the handler
* Implement handler (`airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandler.java`) which is executing the write operations with help of the `ConfigRepository`
  * Checking whether the project exists in the current workspace is centralized in a `validateWorkspace` method that check with the database
* Adjust config repository `airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java`:
  * Do not check workspace consistency (this is pare of the handler now)
  * Throw instead of returning optional for the builder project for consistency
* Add tests for the handler and adjust tests for the config repository.

Notes:
* I tried to follow established approaches whereever possible to make this code as little "suprising" as possible for a platform engineer, but maybe I misunderstood something
* In theory it would be possible in some places to collapse multiple queries into one (e.g. the validate call is already implicitly checking whether a builder project exists, but the write call is doing another lightweight exists call), however separating different concerns seemed more important here. If it turns out to be necessary, performance optimizations can be done in a later step.

Can be tested end-to-end by building the server via `SUB_BUILD=PLATFORM ./gradlew assemble`, starting via `BASIC_AUTH_USERNAME="" BASIC_AUTH_PASSWORD="" VERSION=dev docker-compose --file ./docker-compose.yaml up` and using curl to call the new endpoints while checking the state of the database by logging into the database via `docker exec -ti airbyte-db psql -U docker -d airbyte`


xiaohansong (approved):

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandler.java
```diff
@@ -0,0 +1,92 @@
...
+import io.airbyte.config.persistence.ConfigNotFoundException;
+import io.airbyte.config.persistence.ConfigRepository;
+import io.airbyte.validation.json.JsonValidationException;
+import jakarta.inject.Inject;
+import jakarta.inject.Singleton;
+import java.io.IOException;
+import java.util.UUID;
+import java.util.function.Supplier;
+
+@SuppressWarnings("PMD.AvoidCatchingNPE")
```
girarda, line 22: why suppress the warning?

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandler.java
```diff
@@ -0,0 +1,92 @@
...
+    return new ConnectorBuilderProject().withBuilderProjectId(id).withWorkspaceId(projectCreate.getWorkspaceId())
+        .withName(projectCreate.getBuilderProject().getName())
+        .withManifestDraft(new ObjectMapper().valueToTree(projectCreate.getBuilderProject().getDraftManifest()));
+  }
+
+  private ConnectorBuilderProjectIdWithWorkspaceId idResponseFromBuilderProject(final ConnectorBuilderProject project) {
+    return new ConnectorBuilderProjectIdWithWorkspaceId().workspaceId(project.getWorkspaceId()).builderProjectId(project.getBuilderProjectId());
+  }
+
+  public void validateWorkspace(final UUID projectId, final UUID workspaceId) throws ConfigNotFoundException, IOException {
```
girarda, line 61: can this method be package-protected? I don't think it should be used directly from another class

girarda (approved):

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandler.java
flash1293: Leftover, removed that.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandler.java
flash1293: You are right, made it private

PR #23065, opened 2023-02-15
[DRAFT]🪟 🎨 Update replication page and streams table layout

### LINKED ISSUE #20969: Update connection page scrolling behavior

edmundito:
## What
Resolves #20969

This change merges the changes in these PRs:

- [x] #22173
- [x] #22524
- [x] #21880

There are 12 different combinations to test these changes:

Pages

* Create connection
* Edit replication settings
* Deleted page replication tables

Apps
* Cloud
* OSS

New table feature
* Enabled
* Disabled

## How
Changes described in each PR. This was created as one feature because each had some conflicts with the other and wanted to introduce the changes to users in one pass.


edmundito:
Replaced by https://github.com/airbytehq/airbyte-platform/pull/72

PR #23070, opened 2023-02-15
🪟🔧 Improve the preview of the ConnectorForm in storybook

timroes:
## What

* Adds a new ConnectorForm/Preview story into storybook, which allows easier pasting of _just_ the `connectionSpecification` without any surrounding connector information needed as well as allowing to preview the result of the form submit in the Actions tab.
* Turn the story into CSF 3 format
* Adjusts the markdown documentation we had for this, to be way more simple and just point to our deployed Storybook.

localhost_9009__path=_story_connector-connectorform--preview


airbyte-webapp/src/views/Connector/ConnectorForm/index.stories.tsx
```diff
@@ -216,5 +215,142 @@ Oneof.args = {
...
+    formId: hidden,
+    selectedConnectorDefinition: hidden,
+    selectedConnectorDefinitionSpecification: hidden,
+    isEditMode: hidden,
+    connectorId: hidden,
+    connectionTestSuccess: hidden,
+    isTestConnectionInProgress: hidden,
+    bodyClassName: hidden,
+    headerBlock: hidden,
+    renderFooter: hidden,
```
timroes, line 353: :information_source: By default all properties of the base component (default export) will show up in storybook, so hiding all except `onSubmit` (needed for Actions to work), and the `formType` one (just to be able to render it nicely in whatever type you prefer). Also adding the `connectionSpecification` property here, which our custom render function on top will use.

flash1293 (approved):
That's awesome, thanks for fixing it. Pushed one super small fix (making sure the form type is set matching the spec), but this looks great. Very helpful tool!

timroes:
Superseded by https://github.com/airbytehq/airbyte-platform/pull/94

PR #23078, opened 2023-02-15
Hide the react query devtool button in cypress

timroes:
## What

Hides the react query devtools button while running in cypress. When developing locally you usually have a dev version running that you run cypress tests against (via `cypress:open`), which has that button visible by default. We've seen cases where the button overlaps another button and cypress can't click that unerlying button in this case. So making sure we always hide that button when running cypress cases.

### DIFF
--- a/airbyte-webapp-e2e-tests/cypress.json
+++ b/airbyte-webapp-e2e-tests/cypress.json
@@ -2,6 +2,7 @@
   "projectId": "916nvw",
   "baseUrl": "https://localhost:3000",
   "testFiles": ["**/*.spec.*"],
+  "supportFile": "cypress/support/e2e.ts",
   "viewportHeight": 800,
   "viewportWidth": 1280,
   "retries": {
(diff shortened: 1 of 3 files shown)


teallarson (approved):
Tried it out locally, lgtm.

Do we want to do the same for the Vite button that shows up when there are console errors/warnings?  Ambivalent on that one (since presumably folks should just clear up the errors/warnings 😉)

timroes:
> Do we want to do the same for the Vite button that shows up when there are console errors/warnings? Ambivalent on that one (since presumably folks should just clear up the errors/warnings wink)

I thought that we don't really need that, because it feels like you can (should?) fix your warning/errors before running the end to end tests, why I decided not to do. But happy to revise if you feel it's beneficial.

ambirdsall (approved):

timroes:
Superseded by https://github.com/airbytehq/airbyte-platform/pull/81

PR #23085, opened 2023-02-15
🪟 🎉 Add Launch Darkly context support to webapp

joeykmh:
## What
This PR adds support for Launch Darkly [Contexts](https://docs-stg.launchdarkly.com/contexts-eap/home/contexts).

Previously Launch Darkly only had a concept of "users", this has been expanded to allow any user-defined "context". Contexts can also be combined into "multi-contexts" to allow for more fine-grained targeting. This requires us to construct one object per context, and re-identify with the `LDClient.identify()` method provided by the Launch Darkly SDK.

## Recommended reading order
Top to bottom

## 🚨 User Impact 🚨
Before merging, we need to make sure that all targeting rules are properly targeting user contexts instead of user objects.

## Pre-merge Checklist
- [ ] Update all Launch Darkly flags to target contexts


airbyte-webapp/src/packages/cloud/services/thirdParty/launchdarkly/contexts.ts
```diff
@@ -0,0 +1,46 @@
...
+export function createUserContext(user: User | null, locale: string): LDSingleKindContext {
+  const kind = "user";
+
+  if (!user) {
+    return {
+      kind,
+      anonymous: true,
+      locale,
+      // @ts-expect-error LaunchDarkly's typescript types specify that key must be defined, but actually they handle a null or undefined key. It's even recommended in their docs to omit the key for anonymous users https://docs.launchdarkly.com/sdk/client-side/javascript/migration-2-to-3#understanding-changes-to-anonymous-users
+      key: undefined,
```
josephkmh, line 14: Unless I'm misreading this, this is an error in the LD SDK. It's written in JS, so it's possible this is an oversight by them, types are here: https://github.com/launchdarkly/js-sdk-common/blob/main/typings.d.ts#L349

airbyte-webapp/src/packages/cloud/services/thirdParty/launchdarkly/LDExperimentService.tsx
```diff
@@ -164,7 +140,9 @@ const LDInitializationWrapper: React.FC<React.PropsWithChildren<{ apiKey: string
 
   // Whenever the user, locale or workspaceId changes, we need to re-identify with launchdarkly
   useEffect(() => {
-    ldClient.current?.identify(mapUserToLDUser(user, locale, workspaceId || null));
+    const userContext = createUserContext(user, locale);
+    const contexts = workspaceId ? createMultiContext(userContext, createWorkspaceContext(workspaceId)) : userContext;
```
timroes, line 144: Nitpick: for easier reading: shouldn't we just always wrap it into a multi context even if it's only the user context in there, so we just sometimes fill it with user and workspace and sometimes just with workspace?

airbyte-webapp/src/packages/cloud/services/thirdParty/launchdarkly/LDExperimentService.tsx
```diff
@@ -123,7 +99,7 @@ const LDInitializationWrapper: React.FC<React.PropsWithChildren<{ apiKey: string
   };
 
   if (!ldClient.current) {
-    ldClient.current = LDClient.initialize(apiKey, mapUserToLDUser(user, locale, null));
+    ldClient.current = LDClient.initialize(apiKey, createUserContext(user, locale));
```
timroes, line 102: I think we also should set the multi context here directly, so we don't have the initial auth against LD without the workspace context?

airbyte-webapp/src/packages/cloud/services/thirdParty/launchdarkly/LDExperimentService.tsx
josephkmh: So according to the LD docs, a single context should be declared without the "multi" wrapper. But I just tested it and seemingly a multi-context containing only a single context works fine. Happy to change that if you think it increases readability.

airbyte-webapp/src/packages/cloud/services/thirdParty/launchdarkly/LDExperimentService.tsx
josephkmh: Sure, we have access to `workspaceId` here on the first render, so we can conditionally add the workspace context. Added in 9360c9c8150242c7345362d066b97c7b3206529b

airbyte-webapp/src/packages/cloud/services/thirdParty/launchdarkly/LDExperimentService.tsx
```diff
@@ -123,7 +99,12 @@ const LDInitializationWrapper: React.FC<React.PropsWithChildren<{ apiKey: string
   };
 
   if (!ldClient.current) {
-    ldClient.current = LDClient.initialize(apiKey, mapUserToLDUser(user, locale, null));
+    const userContext = createUserContext(user, locale);
+    const contexts = workspaceId
+      ? createMultiContext(userContext, createWorkspaceContext(workspaceId))
+      : createMultiContext(userContext);
+    ldClient.current = LDClient.initialize(apiKey, contexts);
+    console.log("workspaceId is", workspaceId);
```
timroes, line 107: Leftover log statement.

airbyte-webapp/src/packages/cloud/services/thirdParty/launchdarkly/LDExperimentService.tsx
```diff
@@ -123,7 +99,12 @@ const LDInitializationWrapper: React.FC<React.PropsWithChildren<{ apiKey: string
   };
 
   if (!ldClient.current) {
-    ldClient.current = LDClient.initialize(apiKey, mapUserToLDUser(user, locale, null));
+    const userContext = createUserContext(user, locale);
+    const contexts = workspaceId
+      ? createMultiContext(userContext, createWorkspaceContext(workspaceId))
+      : createMultiContext(userContext);
```
timroes, line 105: Minor code style nitpick (take or leave): Since it's the very same code as L148ff it might make sense to just create a `createContext(user, locale, workspaceId?)` that returns the final multi context for us, so we can make sure it's never diverging from each other.

timroes (approved):
Code LGTM (one minor take or leave nitpick). Tested locally and it seems targeting via user as well as workspace contexts is working as expected.

Need to make sure that we don't target anyone on user.workspace right now, since this will be gone.

josephkmh:
migrated here https://github.com/airbytehq/airbyte-platform-internal/pull/5196

PR #23091, opened 2023-02-15
Source TikTok-Marketing: added more clear message for 504 error

darynaishchenko:
added more clear message for 504 error
on-call: https://github.com/airbytehq/oncall/issues/980


bazarnov (approved):

lazebnyi (changes requested):
Please add allowedHosts.

airbyte-integrations/connectors/source-tiktok-marketing/source_tiktok_marketing/streams.py
```diff
@@ -427,7 +431,7 @@ def stream_slices(self, **kwargs) -> Iterable[Optional[Mapping[str, Any]]]:
         ids = self.get_advertiser_ids()
         start, end, step = 0, len(ids), 100
         for i in range(start, end, step):
-            yield {"advertiser_ids": ids[i: min(end, i + step)]}
+            yield {"advertiser_ids": ids[i : min(end, i + step)]}
```
maxi297, line 434: nit: this does not align with [PEP 8: E203 whitespace before ':'](https://www.flake8rules.com/rules/E203.html) even though our black formatter does not pick it up

airbyte-integrations/connectors/source-tiktok-marketing/source_tiktok_marketing/streams.py
```diff
@@ -204,10 +204,14 @@ def should_retry(self, response: requests.Response) -> bool:
         Once the rate limit is met, the server returns "code": 40100
         Docs: https://business-api.tiktok.com/marketing_api/docs?id=1701890997610497
         Retry 50002 as well - it's a server error.
+        Retry when 504 error: response doesn't consist json, so we need to handle response status code to retry.
         """
         try:
             data = response.json()
         except Exception:
+            if response.status_code == 504:
```
maxi297, line 212: I have two questions: 
* Do we expect Tiktok to maybe someday have the response body as JSON as well? If so, this code will not work anymore
* Why don't we add this as part of the HttpStream? I seems like whenever there will be a 504, this will be an error that we will want to have clear logging on. @brianjlai is there a reason why we would not put this in the CDK itself?

maxi297 (approved):
The requested changes are trivial or more of an open discussion therefore I'll approve the PR

airbyte-integrations/connectors/source-tiktok-marketing/source_tiktok_marketing/streams.py
darynaishchenko: Answering for first question, [docs](https://ads.tiktok.com/marketing_api/docs?id=1737172488964097), tik tok has declaretaled status codes in json response, there aren't any status for 504, which is html.

lazebnyi (approved):

PR #23099, opened 2023-02-15
ci: weekly builds for alpha connectors

alafanechere:
We want to run the alpha connectors builds weekly instead of nightly for cost reasons.

### DIFF
--- a/.github/workflows/connector_integration_tests_alpha.yml
+++ b/.github/workflows/connector_integration_tests_alpha.yml
@@ -1,4 +1,4 @@
-name: Connector Integration Tests
+name: Connector Integration Tests (Alpha connectors only)
 
 # Launches the connector integration tests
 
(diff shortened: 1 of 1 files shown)


c-p-b (approved):

evantahler (approved):
👍 

nit: should we make 6x workflows (alpha-sources, alpha-destinations, beta-sources, beta-destinations, etc) to be clearer about the seperation?  Would adding parallelism like this be helpful?

cpdeethree:
> 👍
> 
> nit: should we make 6x workflows (alpha-sources, alpha-destinations, beta-sources, beta-destinations, etc) to be clearer about the seperation? Would adding parallelism like this be helpful?

Augustin, Ben and I have been having ongoing conversations about how these should be structured. I think the visibility aspect of this dictates how we want to do it along with performance. We will be chatting again on Tuesday of next week and perhaps at that point we will have a clearer answer

octavia-approvington:
It's time <br/> fine lets go

octavia-approvington (approved):

alafanechere:
> Would adding parallelism like this be helpful?
Technically these tests are already all dispatched in parallel. 
Logically separating source/release stage could be interesting but I don't know why for yet 😄

PR #23111, opened 2023-02-15
check certain tasks to see if they already exist before registering them

colesnodgrass:
## What
- inline `publish-repositories.gradle`
    - the `rootDir` path causes an issue when building from cloud
    - saw no benefit to having this maven configuration block existing in a separate file  
- fix airbytehq/airbyte-internal-issues#3885
    - as these tasks will already exist when building built by cloud, check if they already exist before adding them again

## How
- check for task already existing with the same name for tasks `copyDocker` and `listAllDependencies`


build.gradle
```diff
@@ -196,9 +203,12 @@ def Task getPublishArtifactsTask(String buildVersion, Project subproject) {
 allprojects {
     apply plugin: 'com.bmuschko.docker-remote-api'
 
-    task copyDocker(type: Sync) {
-        from "${project.projectDir}/Dockerfile"
```
cpdeethree, line 200: So if the behavior of the tasks different from OSS to cloud, wouldn't one override the other in this approach? I worry that if that were to happen (even though these two tasks are currently identical) you might end up in a situation where the wrong task was used unexpectedly.

Should we instead ensure that there are no duplicate names (maybe by just letting the build fail) and rename the OSS or cloud tasks and update the builds where they get called??

I know at least listAllDependencies isn't called by any builds (I wrote that one ;) )

colesnodgrass: > So if the behavior of the tasks different from OSS to cloud, wouldn't one override the other in this approach?

When building `oss`, the `oss` one will be used. When building `cloud`, the `cloud` one will be used (as `cloud` will load first and only `oss` has this check).  This is the behavior we want.

Maybe they should have different names, but that is something we can address after the merge of `oss` into `cloud`. I'm trying to keep breaking changes to a minimum prior to the merge.

build.gradle
```diff
@@ -196,9 +203,12 @@ def Task getPublishArtifactsTask(String buildVersion, Project subproject) {
...
-        into "build/docker/"
+    // check if the copyDocker task already exists as it may have already been added by cloud
+    if (tasks.findByName('copyDocker') == null) {
```
jdpgrailsdev, line 207: Groovy nit:  you can just do `!tasks.findByName(...)` without the `== null` bit.

jdpgrailsdev (approved):
:shipit:

c-p-b (approved):

davinchia (approved):

build.gradle
davinchia: can you add a comment explaining this nuance? will be easier to misunderstand otherwise.

colesnodgrass: @davinchia I'm planning on doing this better (something along the lines of what Conor recommended) after the code has been merged.  So this nuance should only be around for about a week (famous last words, I know).

colesnodgrass:
No longer needed.

Issue #21967, opened 2023-01-27
Upgrade Debezium version to 2.1


subodh1810:
- [x] Postgres 
The version would allow us to get rid of the custom jar file that we have since the [property](https://debezium.io/documentation/reference/stable/connectors/postgresql.html#postgresql-property-flush-lsn-source) is now in GA.
There is a bug that has been bothering us in version 1.9.6 (the version that we are using) which has been fixed in the later versions. See [this](https://debezium.zulipchat.com/#narrow/stream/302529-users/topic/Postgres.20stuck.20on.20.22Searching.20for.20WAL.20resume.20position.22)

<img width="892" alt="Screenshot 2023-02-01 at 11 21 10 PM" src="https://user-images.githubusercontent.com/28527634/216123190-5e01b749-ed25-41ab-8952-6fc2a66ea2d1.png">
[debezium-issue-1](https://issues.redhat.com/browse/DBZ-6026)
and 
[debezium issue 2](https://issues.redhat.com/browse/DBZ-6012)

- [x] MySQL
- [x] MSSQL

subodh1810:
We should go through the changelogs
https://debezium.io/releases/2.1/release-notes
and 
https://debezium.io/releases/2.0/release-notes

to make sure that there are no breaking changes

PR #23112, opened 2023-02-15
DB sources cdc : upgrade Debezium version to 2.1.2

subodh1810:
Issue : https://github.com/airbytehq/airbyte/issues/21967

There are few breaking changes in this PR:
1. The support for wal2json plugin for Postgres CDC is removed
2. Debezium state format has changed but I have added logic to make the old state compatible with the new state. 

Reading order:
The PR shows so many files changed cause I have renamed the debezium module. First take a look at the debezium module changes and then individual connector changes.


airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIterator.java
```diff
@@ -203,4 +200,35 @@ private void throwExceptionIfSnapshotNotFinished() {
...
+      }
+
+      final Class<? extends ChangeEvent> eventClass = heartbeatEvent.getClass();
+      final Field f;
+      if (heartbeatEventSourceField.containsKey(eventClass)) {
+        f = heartbeatEventSourceField.get(eventClass);
+      } else {
+        f = eventClass.getDeclaredField("sourceRecord");
+        f.setAccessible(true);
+        heartbeatEventSourceField.put(eventClass, f);
```
sergio-ropero, line 217: I still don't fully understand what `heartbeatEventSourceField` is, how is intended to be used, ...
I see it's a property for the class `DebeziumRecordIterator` but we don't have any description of it.
That would help to review/modify the code in the future.

airbyte-integrations/connectors/source-mssql/src/main/java/io/airbyte/integrations/source/mssql/MssqlCdcHelper.java
```diff
@@ -36,7 +36,7 @@ public enum ReplicationMethod {
    * The default "SNAPSHOT" mode can prevent other (non-Airbyte) transactions from updating table rows
    * while we snapshot. References:
    * https://docs.microsoft.com/en-us/sql/t-sql/statements/set-transaction-isolation-level-transact-sql?view=sql-server-ver15
-   * https://debezium.io/documentation/reference/1.4/connectors/sqlserver.html#sqlserver-property-snapshot-isolation-mode
+   * https://debezium.io/documentation/reference/stable/connectors/sqlserver.html#sqlserver-property-snapshot-isolation-mode
```
sergio-ropero, line 39: As discussed it's better to include version `2.1.2` in the docs, so if Debezium release another stable version, it does not mismatch the docs with the imported.
Change for all the references :)

airbyte-integrations/connectors/source-postgres/src/main/resources/spec.json
```diff
@@ -246,8 +246,8 @@
               "plugin": {
                 "type": "string",
                 "title": "Plugin",
-                "description": "A logical decoding plugin installed on the PostgreSQL server. The `pgoutput` plugin is used by default. The `wal2json` plugin is deprecated and will soon be removed so it's not recommended to use. Read more about <a href=\"https://docs.airbyte.com/integrations/sources/postgres#step-2-select-a-replication-plugin\">selecting replication plugins</a>.",
-                "enum": ["pgoutput", "wal2json"],
+                "description": "A logical decoding plugin installed on the PostgreSQL server.",
+                "enum": ["pgoutput"],
                 "default": "pgoutput",
                 "order": 2
               },
```
sergio-ropero, line 253: Should we just remove this?
So far there is only one option, and showing it to the users would be confusing.
```suggestion
```

airbyte-integrations/connectors/source-postgres/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/AbstractPostgresSourceDatatypeTest.java
```diff
@@ -293,26 +293,26 @@ protected void initTests() {
...
+//            // https://github.com/airbytehq/airbyte/issues/7338
+//            /* "'-92233720368547758.08'", "'92233720368547758.07'" */)
+//            .addExpectedValues(
+//                null,
+//                // Double#toString method is necessary here because sometimes the output
+//                // has unexpected decimals, e.g. Double.toString(-1000) is -1000.0
+//                "999.99", "1001.01", Double.toString(-1000),
+//                "999.99", "1001.01", Double.toString(-1000)
+//            /* "-92233720368547758.08", "92233720368547758.07" */)
+//            .build());
```
sergio-ropero, line 316: Just remove it, we have control version to recover if needed.
```suggestion
```

airbyte-integrations/connectors/source-postgres/src/test-integration/resources/expected_spec.json
```diff
@@ -246,8 +246,8 @@
               "plugin": {
                 "type": "string",
                 "title": "Plugin",
-                "description": "A logical decoding plugin installed on the PostgreSQL server. The `pgoutput` plugin is used by default. The `wal2json` plugin is deprecated and will soon be removed so it's not recommended to use. Read more about <a href=\"https://docs.airbyte.com/integrations/sources/postgres#step-2-select-a-replication-plugin\">selecting replication plugins</a>.",
-                "enum": ["pgoutput", "wal2json"],
+                "description": "A logical decoding plugin installed on the PostgreSQL server.",
+                "enum": ["pgoutput"],
                 "default": "pgoutput",
                 "order": 2
               },
```
sergio-ropero, line 253: Same than before, adding options to the user to not being able to chose... not good 🤣 
```suggestion
```

airbyte-integrations/connectors/source-postgres/src/test-integration/resources/expected_strict_encrypt_spec.json
```diff
@@ -226,8 +226,8 @@
               "plugin": {
                 "type": "string",
                 "title": "Plugin",
-                "description": "A logical decoding plugin installed on the PostgreSQL server. The `pgoutput` plugin is used by default. The `wal2json` plugin is deprecated and will soon be removed so it's not recommended to use. Read more about <a href=\"https://docs.airbyte.com/integrations/sources/postgres#step-2-select-a-replication-plugin\">selecting replication plugins</a>.",
-                "enum": ["pgoutput", "wal2json"],
+                "description": "A logical decoding plugin installed on the PostgreSQL server.",
+                "enum": ["pgoutput"],
                 "default": "pgoutput",
                 "order": 2
               },
```
sergio-ropero, line 233: Same
```suggestion
```

airbyte-integrations/bases/debezium/src/test/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIteratorTest.java
```diff
@@ -0,0 +1,57 @@
...
+import java.time.Duration;
+import java.util.Collections;
+import java.util.concurrent.LinkedBlockingQueue;
+import org.apache.kafka.connect.source.SourceRecord;
+import org.junit.jupiter.api.Test;
+
+public class DebeziumRecordIteratorTest {
+
+  @Test
+  public void getHeartbeatPositionTest() {
```
sergio-ropero, line 22: We should include more tests for this function:
- what if is null (or not heartbeat event?)
- if the size > 1
- ...
For me that function is quite complex (which makes hard to test it as well).

sergio-ropero (changes requested):
Hard work here man! 🚀 

I have seen than in some tests you have added primary key in the tables. Could you modify also file `airbyte-integrations/connectors/source-postgres/integration_tests/seed.sql` and include the primary key in the table? 🙏

airbyte-integrations/connectors/source-mssql/src/main/java/io/airbyte/integrations/source/mssql/MssqlCdcHelper.java
subodh1810: done

airbyte-integrations/connectors/source-postgres/src/main/resources/spec.json
subodh1810: I was thinking the same but while testing I realised that the front end is smart enough to not display the option if there is only 1 option available and default is present so I let it be for now since I didn't want to change a lot. In future once this version is running in prod safely then we can get rid of it. 
Screenshot 2023-02-27 at 10 20 47 PM

airbyte-integrations/connectors/source-postgres/src/test-integration/resources/expected_spec.json
subodh1810: See https://github.com/airbytehq/airbyte/pull/23112#discussion_r1119018382

airbyte-integrations/connectors/source-mssql/src/test-integration/java/io/airbyte/integrations/source/mssql/MssqlSourceAcceptanceTest.java
```diff
@@ -101,14 +105,14 @@ protected JsonNode getState() {
   }
 
   private static DSLContext getDslContext(final JsonNode config) {
-    return DSLContextFactory.create(
+    return DSLContextFactory.create(DataSourceFactory.create(
```
sergio-ropero, line 108: Same, can we reuse `getDataSource `?

airbyte-integrations/connectors/source-mssql/src/test-integration/java/io/airbyte/integrations/source/mssql/MssqlSourceDatatypeTest.java
```diff
@@ -41,19 +44,20 @@ protected Database setupDatabase() throws Exception {
...
     config = Jsons.clone(configWithoutDbName);
     ((ObjectNode) config).put(JdbcUtils.DATABASE_KEY, DB_NAME);
+    ((ObjectNode) config).put("is_test", true);
 
     return database;
   }
 
   private static DSLContext getDslContext(final JsonNode config) {
-    return DSLContextFactory.create(
+    return DSLContextFactory.create(DataSourceFactory.create(
```
sergio-ropero, line 53: Same about `getDataSource`.

airbyte-integrations/connectors/source-postgres/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/AbstractPostgresSourceDatatypeTest.java
```diff
@@ -612,6 +592,29 @@ protected void addJsonbArrayTest() {
             .build());
   }
 
+  protected void addMoneyTest() {
```
sergio-ropero, line 595: can we call it `currency`? hehe

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/AirbyteFileOffsetBackingStore.java
```diff
@@ -59,14 +60,30 @@ public Map<String, String> read() {
   public void persist(final JsonNode cdcState) {
     final Map<String, String> mapAsString =
         cdcState != null ? Jsons.object(cdcState, Map.class) : Collections.emptyMap();
-    final Map<ByteBuffer, ByteBuffer> mappedAsStrings = mapAsString.entrySet().stream().collect(Collectors.toMap(
+
+    final Map<String, String> updatedMap = updateStateForDebezium2_1(mapAsString);
```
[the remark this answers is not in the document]
sergio-ropero: Can we just call this function if the state is not Debezium 2.1?

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIterator.java
```diff
@@ -203,4 +200,35 @@ private void throwExceptionIfSnapshotNotFinished() {
     }
   }
 
+  @VisibleForTesting
+  protected Long getHeartbeatPosition(final ChangeEvent<String, String> heartbeatEvent) {
+    try {
+      if (heartbeatEvent == null) {
+        return null;
+      }
+
```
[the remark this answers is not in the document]
sergio-ropero: 1. In programming it's highly recommended to use _**early return**_ (like explained [here](https://medium.com/swlh/return-early-pattern-3d18a41bba8) as example) as it helps with maintenance and readability. Also it's recommended to only use the `try` in the pieces of code that can raise an exception and not the hole code.

2. this function returns `Long` that's why I mentioned that `-1L` will work and makes it standard.

airbyte-integrations/connectors/source-postgres/src/main/resources/spec.json
sergio-ropero: Nice!
Btw, if there is no more plugins and we don't support them, I would just remove it, less code is better code.

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIterator.java
sergio-ropero: Could you add a description to the variable `heartbeatEventSourceField`?

sergio-ropero (changes requested):

airbyte-integrations/connectors/source-postgres/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/AbstractPostgresSourceDatatypeTest.java
subodh1810: I just use the sql data type name for the test and in DB its called `money`

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/AirbyteFileOffsetBackingStore.java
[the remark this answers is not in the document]
subodh1810: I did avoid the mutation if its already 2.1 state as seen in the method but the problem is that I need to look into the contents of the map and do some logic on top of it to determine if it is. 
```

      if (i == 0 && i1 == key.length() - 1) {
        // The state is Debezium 2.1 compatible. No need to change anything.
        return mapAsString;
      }
```

airbyte-integrations/connectors/source-postgres/src/main/resources/spec.json
subodh1810: If we remove this from the spec then the config of all the existing connectors would needed to be migrated as well which would increase the scope. I will create an issue to clean this up as part of refactoring work

akashkulk:
Some small nits : 

- Can you update the description to include what the breaking changes will be? i.e. why we would need the feature flag
- Can you add a recommended reading order/summary of important changes as this is touching ~75 files

Just want to make sure I understand it if I have to look back at it :)

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/AirbyteFileOffsetBackingStore.java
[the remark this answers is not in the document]
sergio-ropero: Ok... if you say so, for me it's not easy to read, something like this would be much easier:
```
map = get the map
if (isOldVersion()
  transformTo2-1
...
```
With the logic you have in-place, you are creating object that will not be used after all.

sergio-ropero (approved):

subodh1810:
/publish connector=connectors/source-mssql-strict-encrypt run-tests=false

ifun248:
After upgrading Debezium version to 2.1.2, it brings a bug for MySQL Source 2.0.0+ which make the limit.blogSize.error value haven't an impact.

For example at  MySQL Source 2.0.0+,  throwing error i.a.w.t.s.ReplicationActivityImpl(lambda$replicate$3):222 - Sync output exceeds the max temporal message size of 2097152, actual is 6297529.
But at MySQL Source 1.0.21-, there is no such error about max temporal message size.

The detail bug has reported at https://github.com/airbytehq/airbyte/issues/24650. And hope can fix this problem soon!

PR #23120, opened 2023-02-15
cloud-availability-updater: batch connector deployment

alafanechere:
## What
This PR has three goals for the Cloud Availability Updater that the QA engine runs:
- Replace the *per-connector* PR creation with a daily PR creation batching changes to definitions/cloud catalog files for all eligible connectors
- Grow the list of inappropriate for cloud use connectors: I tried to manually spot those missing strict encrypt variants.
- List the eligible connector excluded because of version pinning / explicit exclusion in mask
- Make the created PR body more readable by using markdown tables.
- Increase the unit test coverage

An example of a PR generated by the updated *Cloud Availability Updater* can be found [here](https://github.com/airbytehq/airbyte-platform-internal/pull/4577)

NB: I we want to switch back to per connector PR we'd just have to revert to version 0.1.13


tools/ci_connector_ops/tests/test_qa_engine/test_cloud_availability_updater.py
```diff
@@ -230,20 +179,122 @@ def test_set_git_identity(mocker):
     assert repo == mock_repo
 
 
-def test_deploy_eligible_connectors_to_cloud_repo(mocker, tmp_path):
+def test_get_authenticated_repo_url(mocker):
```
alafanechere, line 182: The new test below are missings unit tests of previously existing functions.

evantahler (approved):

PR #23123, opened 2023-02-15
[Low-Code CDK] Add script to run low code unit tests and address issues with a few connectors

brianjlai:
Remaking the PR since I sort of messed it up w/ the merge from master instead of the dev branch. Reposting below:

## What
While going through the validation checklist for low-code to beta,  some manifests had a few one-off issues that should be addressed to incorporate the changes to the language. Before we push a release to Beta, we should merge this into the branch to retain compatibility

## How
In addition to some fixes, I also added a script to run any low-code connectors that have unit tests to validate they are also compatible with the changes.

## Recommended reading order
1. `low-code-unit-tests.sh`
2. connector changes

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_stream.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_stream.py
@@ -6,6 +6,7 @@ from dataclasses import InitVar, dataclass, field
 from typing import Any, Iterable, List, Mapping, MutableMapping, Optional, Union
 
 from airbyte_cdk.models import SyncMode
+from airbyte_cdk.sources.declarative.interpolation import InterpolatedString
 from airbyte_cdk.sources.declarative.retrievers.retriever import Retriever
 from airbyte_cdk.sources.declarative.schema import DefaultSchemaLoader
 from airbyte_cdk.sources.declarative.schema.schema_loader import SchemaLoader
@@ -25,7 +26,7 @@ class DeclarativeStream(Stream):
         schema_loader (SchemaLoader): The schema loader
         retriever (Retriever): The retriever
         config (Config): The user-provided configuration as specified by the source's spec
-        stream_cursor_field (Optional[List[str]]): The cursor field
+        stream_cursor_field (Optional[Union[InterpolatedString, str]]): The cursor field
         transformations (List[RecordTransformation]): A list of transformations to be applied to each output record in the
         stream. Transformations are applied in the order in which they are defined.
     """
@@ -39,11 +40,11 @@ class DeclarativeStream(Stream):
     _name: str = field(init=False, repr=False, default="")
     _primary_key: str = field(init=False, repr=False, default="")
     _schema_loader: SchemaLoader = field(init=False, repr=False, default=None)
-    stream_cursor_field: Optional[Union[List[str], str]] = None
+    stream_cursor_field: Optional[Union[InterpolatedString, str]] = None
     transformations: List[RecordTransformation] = None
 
     def __post_init__(self, parameters: Mapping[str, Any]):
-        self.stream_cursor_field = self.stream_cursor_field or []
+        self.stream_cursor_field = InterpolatedString.create(self.stream_cursor_field, parameters=parameters)
         self.transformations = self.transformations or []
         self._schema_loader = self.schema_loader if self.schema_loader else DefaultSchemaLoader(config=self.config, parameters=parameters)
 
@@ -86,7 +87,8 @@ class DeclarativeStream(Stream):
         Override to return the default cursor field used by this stream e.g: an API entity might always use created_at as the cursor field.
         :return: The name of the field used as a cursor. If the cursor is nested, return an array consisting of the path to the cursor.
         """
-        return self.stream_cursor_field
+        cursor = self.stream_cursor_field.eval(self.config)
+        return cursor if cursor else []
 
     def read_records(
         self,
--- a/airbyte-cdk/python/bin/low-code-unit-tests.sh
+++ b/airbyte-cdk/python/bin/low-code-unit-tests.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+[ -z "$ROOT_DIR" ] && exit 1
+
+CONNECTORS_DIR=$ROOT_DIR/airbyte-integrations/connectors
+CDK_DIR=$ROOT_DIR/airbyte-cdk/python/
+
+for directory in $CONNECTORS_DIR/source-* ; do
+  MANIFEST_DIRECTORY=$(basename $directory | tr - _)
+  SOURCE_NAME=${MANIFEST_DIRECTORY#source_}
+  if test -f "$directory/$MANIFEST_DIRECTORY/manifest.yaml"; then
+    cd $directory
+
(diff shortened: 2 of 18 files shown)


airbyte-integrations/connectors/source-greenhouse/source_greenhouse/manifest.yaml
```diff
@@ -125,13 +127,14 @@ definitions:
...
         $ref: "#/definitions/requester"
         path: "applications/{{ stream_slice.parent_id }}/demographics/answers"
-      partition_router:
-        class_name: source_greenhouse.components.GreenHouseSubstreamSlicer
-        parent_stream: "#/definitions/applications_stream"
-        request_cursor_field: "updated_after"
-        stream_slice_field: "parent_id"
-        cursor_field: "updated_at"
-        parent_key: "id"
+    incremental_sync:
```
brianjlai, line 130: after inspecting this connector a bit, this really should be an incremental_sync custom component instead of a custom partition_router because it has a cursor_field. In the existing code, we would inadvertently only perform full refreshes since we didn't seen incremental_sync component and therefore assumed it was only full refresh.

This also is more in line w/ our ideal structure where incremental_sync dictates the sync capabilities.

airbyte-integrations/connectors/source-posthog/source_posthog/manifest.yaml
```diff
@@ -103,7 +103,7 @@ definitions:
       type: CustomIncrementalSync
       class_name: source_posthog.components.EventsCartesianProductStreamSlicer
       cursor_field: timestamp
-      partition_routers:
+      stream_slicers:
```
brianjlai, line 106: This one is a bit gross, but because it overrides the CartesianProductStreamSlicer (which shouldn't even be in the language anymore), our underlying implementation detail (namely the stream_slicers field) bleeds into the code a little bit.

airbyte-integrations/connectors/source-braze/source_braze/datetime_incremental_sync.py
```diff
@@ -38,8 +38,8 @@ def _partition_daterange(self, start, end, step: datetime.timedelta):
         """
         Puts a step to each stream slice. `step` is a difference between start/end date in days.
         """
-        get_start_time = operator.itemgetter(self.stream_slice_field_start.eval(self.config))
-        get_end_time = operator.itemgetter(self.stream_slice_field_end.eval(self.config))
+        get_start_time = operator.itemgetter(self.partition_field_start.eval(self.config))
```
brianjlai, line 41: Fixes the component because the DatetimeBasedCursor this was originally implementing had changed a bit.

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_stream.py
```diff
@@ -39,11 +40,11 @@ class DeclarativeStream(Stream):
     _name: str = field(init=False, repr=False, default="")
     _primary_key: str = field(init=False, repr=False, default="")
     _schema_loader: SchemaLoader = field(init=False, repr=False, default=None)
-    stream_cursor_field: Optional[Union[List[str], str]] = None
+    stream_cursor_field: Optional[Union[InterpolatedString, str]] = None
```
brianjlai, line 43: This was an interesting gap. We used to hard code this to not support interpolation just a simple or nested cursor field. But this didn't align with the slicer's cursor_field. No connectors used a nested cursor field so far, but we can always extend this out as new functionality if we need it.

airbyte-integrations/connectors/source-square/source_square/manifest.yaml
```diff
@@ -27,9 +27,12 @@ definitions:
       bearer: "#/definitions/bearer_authenticator"
       oauth: "#/definitions/oauth_authenticator"
 
-    request_headers:
-      Square-Version: "2022-10-19"
-      Content-Type: "application/json"
+  #    Uncomment this block later. request_headers used to accidentally get overridden but this was fixed in beta release
+  #    However, because expected_records were originally generated with the buggy flow missing these, we need to
+  #    regenerate the expected records. Temporarily commenting it out to retain the old behavior until we do it.
```
brianjlai, line 32: fun stuff here... but it would appear there isn't a trivial way to generate new expected_records. The fix and path is clear to resolve this, but I don't really want to hold up the release of this on something like regenerating the expected records

airbyte-cdk/python/bin/low-code-unit-tests.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
girarda, line 1: nit: `set -e` in case we ever use this script from another script

airbyte-integrations/connectors/source-freshdesk/source_freshdesk/source.py
```diff
@@ -57,7 +57,7 @@ def _get_stream_kwargs(config: Mapping[str, Any]) -> dict:
 
     def check_connection(self, logger: logging.Logger, config: Mapping[str, Any]):
         try:
-            check_stream = CheckStream(stream_names=["settings"], options={})
+            check_stream = CheckStream(stream_names=["settings"], parameters={})
```
girarda, line 60: 🔪

airbyte-integrations/connectors/source-greenhouse/source_greenhouse/manifest.yaml
girarda: agreed

airbyte-integrations/connectors/source-posthog/source_posthog/manifest.yaml
girarda: I might have to take it back one day, but I don't hate that stream_slicers can still be customized if needed

airbyte-integrations/connectors/source-square/source_square/manifest.yaml
girarda: can you create an issue if this won't be resolved it's merged in?

airbyte-integrations/connectors/source-woocommerce/source_woocommerce/manifest.yaml
```diff
@@ -156,6 +156,7 @@ definitions:
       end_time_option:
         field_name: before
         inject_into: request_parameter
+      cursor_field: "date_created_gmt"
```
girarda, line 159: was this just not used?

girarda (approved):
🚢

airbyte-integrations/connectors/source-square/source_square/manifest.yaml
brianjlai: will do!

airbyte-integrations/connectors/source-woocommerce/source_woocommerce/manifest.yaml
brianjlai: It is used in the manifest, which was why it triggered a CAT failure, but it was a mistake on my end that I accidentally deleted `date_created_gmt` cursor. I thought product reviews stream was using the same cursor as the other streams (`date_modified_gmt`) so it didn't need an override, but it wasn't so thats why I added this back

PR #23128, opened 2023-02-16
Only generate api client code as needed

ambirdsall:
## What
Replaces the bare `orval` call in `package.json["scripts"]["generate-client"]` with a new node script that only (re)generates API client files that are missing or out of date (as determined by [the files' `mtime`](https://www.gnu.org/software/coreutils/manual/html_node/File-timestamps.html)). This significantly speeds up startup time for scripts that rely on `generate-client`, like `start`, `start:cloud`, and `test`.

### DIFF
--- a/airbyte-webapp/package.json
+++ b/airbyte-webapp/package.json
@@ -8,13 +8,13 @@
   },
   "scripts": {
     "prepare": "cd .. && husky install airbyte-webapp/.husky",
-    "prestart": "pnpm run generate-client",
+    "prestart": "TS_NODE_TRANSPILE_ONLY=true pnpm run generate-client",
     "start": "NODE_OPTIONS='-r ./scripts/dev-overwrites.js' vite",
-    "prestart:cloud": "pnpm run generate-client",
+    "prestart:cloud": "TS_NODE_TRANSPILE_ONLY=true pnpm run generate-client",
     "start:cloud": "AB_ENV=${AB_ENV-frontend-dev} NODE_OPTIONS='-r ./scripts/environment.js -r ./scripts/dev-overwrites.js' vite",
     "prebuild": "pnpm run generate-client",
     "build": "vite build",
-    "pretest": "pnpm run generate-client",
+    "pretest": "TS_NODE_TRANSPILE_ONLY=true pnpm run generate-client",
     "test": "jest --watch",
     "test:ci": "jest --watchAll=false --silent",
     "test:coverage": "jest --coverage --watchAll=false",
@@ -25,7 +25,7 @@
     "stylelint": "stylelint 'src/**/*.{css,scss}'",
     "stylelint-check": "stylelint-config-prettier-scss-check",
     "license-check": "node ./scripts/license-check.js",
-    "generate-client": "./scripts/load-declarative-schema.sh && orval",
+    "generate-client": "./scripts/load-declarative-schema.sh && ts-node --skip-project ./scripts/regenerate-api-clients.ts",
     "validate-links": "ts-node --skip-project ./scripts/validate-links.ts"
   },
   "dependencies": {
--- a/airbyte-webapp/package.json
+++ b/airbyte-webapp/package.json
@@ -180,27 +180,13 @@
     "minimatch": "^3.0.5"
   },
   "lint-staged": {
-    "src/**/*.{js,jsx,ts,tsx}": [
-      "eslint --fix"
-    ],
-    "src/**/*.{css,scss,md,json}": [
-      "prettier --write"
-    ],
-    "{public,src}/**/*.{css,scss}": [
-      "stylelint --fix"
-    ]
+    "src/**/*.{js,jsx,ts,tsx}": ["eslint --fix"],
+    "src/**/*.{css,scss,md,json}": ["prettier --write"],
+    "{public,src}/**/*.{css,scss}": ["stylelint --fix"]
   },
   "browserslist": {
-    "production": [
-      ">0.2%",
-      "not dead",
-      "not op_mini all"
-    ],
-    "development": [
-      "last 1 chrome version",
-      "last 1 firefox version",
-      "last 1 safari version"
-    ]
+    "production": [">0.2%", "not dead", "not op_mini all"],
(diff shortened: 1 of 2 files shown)


airbyte-webapp/scripts/regenerate-api-clients.js
```diff
@@ -0,0 +1,12 @@
+const { statSync } = [require("fs");
+const { resolve } = require("path");
+const conf = require("../orval.config.js");
+
+const lastTouched = (filepath) => statSync(resolve(filepath)).mtime;
+
+for (const [, api] of Object.entries(conf)) {
+  if (lastTouched(api.input) > lastTouched(api.output.target)) {
+    // no need to read and parse the orval lib if all files are up to date
```
ambirdsall, line 9: This ended up being a useful optimization:
```
# `require("orval")` outside of if block
0.71 real         0.63 user         0.15 sys

# `require("orval")` inside of if block
0.29 real         0.26 user         0.03 sys

File: airbyte-webapp/package.json [hunk 2]
```diff
@@ -180,27 +180,13 @@
...
-    ],
-    "src/**/*.{css,scss,md,json}": [
-      "prettier --write"
-    ],
-    "{public,src}/**/*.{css,scss}": [
-      "stylelint --fix"
-    ]
+    "src/**/*.{js,jsx,ts,tsx}": ["eslint --fix"],
+    "src/**/*.{css,scss,md,json}": ["prettier --write"],
+    "{public,src}/**/*.{css,scss}": ["stylelint --fix"]
```
krishnaglick, line 185: Curious why this changed. Was linting not run here previously?

airbyte-webapp/scripts/regenerate-api-clients.ts
```diff
@@ -0,0 +1,13 @@
+import { statSync, existsSync } from "fs";
+import { resolve } from "path";
+
+import conf from "../orval.config";
+
+const lastTouched = (filepath: string) => statSync(resolve(filepath)).mtime;
```
krishnaglick, line 6: Did you test what happens if the file doesn't exist?

airbyte-webapp/package.json
ambirdsall: Apparently not! This file is outside of `src/`, so `lint-staged` won't ensure linting; I guess just have a more aggressive auto-format setup in my editor than most. It'll be nice to get the linted version merged, though: I've skipped adding these lines to a bunch of commits already.

airbyte-webapp/scripts/regenerate-api-clients.ts
ambirdsall: It breaks, as you presumably guessed. I added the `existsSync` check below, on line 9, after seeing the build step fail in CI without it (I thought that would happen, and it did! Hooray for validating guesses.)

airbyte-webapp/package.json
```diff
@@ -8,13 +8,13 @@
   },
   "scripts": {
     "prepare": "cd .. && husky install airbyte-webapp/.husky",
-    "prestart": "pnpm run generate-client",
+    "prestart": "TS_NODE_TRANSPILE_ONLY=true pnpm run generate-client",
     "start": "NODE_OPTIONS='-r ./scripts/dev-overwrites.js' vite",
-    "prestart:cloud": "pnpm run generate-client",
+    "prestart:cloud": "TS_NODE_TRANSPILE_ONLY=true pnpm run generate-client",
     "start:cloud": "AB_ENV=${AB_ENV-frontend-dev} NODE_OPTIONS='-r ./scripts/environment.js -r ./scripts/dev-overwrites.js' vite",
     "prebuild": "pnpm run generate-client",
```
ambirdsall, line 15: `prebuild` runs the generation script with type checking; the other, more development-focused commands which run it (like `prestart`) skip type checking with `TS_NODE_TRANSPILE_ONLY=true` to optimize for execution speed.

airbyte-webapp/package.json
timroes: I think there might be some more issues with the setup. Even if I run prettier over package.json manually, it will end up with the previous format not the one your editor ended up with. Can we check if there's anything in your setup that might format this with another formatter than prettier?

airbyte-webapp/scripts/regenerate-api-clients.ts
```diff
@@ -0,0 +1,17 @@
...
+
+for (const api of Object.values(conf)) {
+  if (!existsSync(api.output.target) || lastTouched(api.input) > lastTouched(api.output.target)) {
+    // no need to read and parse the orval lib if all files are up to date
+    require("orval").generate(api);
+  } else {
+    console.log(`🎉 ${chalk.green(basename(api.output.target))} is up-to-date`);
```
timroes, line 15: @ambirdsall Added some output if a file is up-to-date (mirroring the output of Orval), to make sure it's still visible that something was checked.

airbyte-webapp/scripts/regenerate-api-clients.ts
```diff
@@ -0,0 +1,17 @@
+import { statSync, existsSync } from "fs";
+import { resolve, basename } from "path";
+
+import chalk from "chalk";
+
+import conf from "../orval.config";
+
+const lastTouched = (filepath: string) => statSync(resolve(filepath)).mtime;
+
+for (const api of Object.values(conf)) {
```
timroes, line 10: `Object.entries` -> `Object.values`, since we never cared abour the key.

airbyte-webapp/scripts/regenerate-api-clients.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (!existsSync(api.output.target) || lastTouched(api.input) > lastTouched(api.output.target)) {
```
timroes, line 11: Could we still add a `process.env.CI ||` to this check, so we regenerate always for sure, even if it's already up to date (since 5s overhead isn't that bad on CI, i'd rather make sure we're 100% always running on fresh files in every build).

ambirdsall:
Replaced by https://github.com/airbytehq/airbyte-platform-internal/pull/4892

PR #23136, opened 2023-02-16
Source gnews remove jsonschemamixin

clnoll:
Removes `JsonSchemaMixin` from source-gnews' `WaitUntilMidnightBackoffStrategy`, which is no longer necessary and was causing CAT failures.


clnoll:
Looks like this was already fixed.

PR #23138, opened 2023-02-16
fix: Permission check failing when privilege is at ROLE level

sergio-ropero:
# What
The replication can be only done when the user has `SUPERUSER` or `REPLICATION` privileges. 
```
psql: error: FATAL:  must be superuser or replication role to start walsender
```
In the documentation we assumed that using a `role` with any of the above privileges attached to a user, grants these attributes to the user, [but it's not right](https://www.postgresql.org/docs/current/role-attributes.html#:~:text=initiating%20replication,REPLICATION%20LOGIN).

# Solution
In order to fix the check function we will try directly to connect to the database doing a replication connection and verify that it's correct.

Added tests for all cases:
- Access with SUPERUSER privilege
- Access with REPLICATION privilege
- NO access (no privileges)

# FROM HERE IS NOT RELEVANT ANYMORE! (But can be interesting to keep)
This commit fixes the issue when permission is granted at ROLE level instead of USER level.

Missing revoke privileges in the tests.

### Set up
The user doesn't have replication privilege at user level, neither for the role attached (`secondary_role`). But the role `secondary_role` is attached to `airbyte_replication` role, so at the end the user has access to replication thanks to last link.
```
CREATE USER airbyte_ro PASSWORD 'XXXXX';
CREATE ROLE airbyte_replication REPLICATION LOGIN;
CREATE ROLE secondary_role LOGIN;
GRANT secondary_role TO airbyte_ro;
GRANT secondary_role TO airbyte_replication;
```
### User Table
We see it doesn't have replication permission:
```
select usesysid, usename, userepl from pg_user where usename = 'airbyte_ro';
```
Screenshot 2023-02-16 at 16 54 56

## Assigned roles
As well without replication permission.
```
select pgr.rolname, rolreplication  from pg_catalog.pg_roles pgr
join pg_catalog.pg_auth_members pgam on pgam.roleid = pgr.oid 
join pg_catalog.pg_user pgu on pgam.member = pgu.usesysid 
where usename ='airbyte_ro'; 
```
Screenshot 2023-02-16 at 16 56 19

### Query to check permission (by any subrole).
If it return at least one record, the user has privilege to Postgres replication:
```
WITH RECURSIVE cte AS (
   SELECT usesysid  FROM pg_catalog.pg_user WHERE usename = 'airbyte_ro'
   UNION ALL
   SELECT m.roleid
   FROM   cte
   JOIN   pg_auth_members m ON m.member = cte.usesysid
   )
SELECT rolname FROM cte
JOIN pg_roles ON usesysid = oid
WHERE rolreplication;
```
Screenshot 2023-02-16 at 17 01 36


airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -322,14 +322,34 @@ public List<CheckedConsumer<JdbcDatabase, Exception>> getCheckOperations(final J
...
-          LOGGER.info("Verifying required privileges for user: {}", userName);
+          LOGGER.info("Verifying required privileges at user level for user: {}", userName);
           return ps;
         }, sourceOperations::rowToJson);
 
-        if (!userPrivileges.get(0).get("userepl").asBoolean()) {
+        // Verify if the user has privilege at role level
+        final List<JsonNode> rolePrivileges = database.queryJsons(connection -> {
+          final PreparedStatement ps = connection.prepareStatement("""
+              WITH RECURSIVE cte AS (
```
rodireich, line 336: So looking recursively what roles the assigned roles have is what is needed?

rodireich (approved):

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
sergio-ropero: Yes, basically if we have:
granted roleA (No replication) to user
granted roleB (No replication) to roleA
granted roleC (Replication) to roleB

The query will output roleC, so we are sure that a subrole has access.
I added the test to cover that case (only one lvl BTW)

sergio-ropero: Adding extra comment in the function

sergio-ropero:
@rodireich @subodh1810 @bleonard 
The failures in the `/test` command are related to the acceptance test that we know about:
```
FAILED test_incremental.py::TestIncremental::test_two_sequential_reads[inputs0]
FAILED test_incremental.py::TestIncremental::test_two_sequential_reads[inputs1]
```
Can I just move forward and publish skipping the tests?
Also, once merged, I will change the mask in Airbyte cloud to get >= latest version (1.0.46 if I'm not wrong).
Thanks all!

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -318,18 +318,18 @@ public List<CheckedConsumer<JdbcDatabase, Exception>> getCheckOperations(final J
...
         final List<JsonNode> userPrivileges = database.queryJsons(connection -> {
-          final PreparedStatement ps = connection.prepareStatement("SELECT userepl FROM pg_user WHERE usename = ?");
+          final PreparedStatement ps = connection.prepareStatement("SELECT userepl or usesuper AS has_access FROM pg_user WHERE usename = ?");
           ps.setString(1, userName);
           LOGGER.info("Verifying required privileges for user: {}", userName);
           return ps;
         }, sourceOperations::rowToJson);
 
-        if (!userPrivileges.get(0).get("userepl").asBoolean()) {
+        if (!userPrivileges.get(0).get("has_access").asBoolean()) {
```
rodireich, line 332: Do we know whether the cloud users that had the problem with 1.0.45 will have this permission?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -318,18 +318,18 @@ public List<CheckedConsumer<JdbcDatabase, Exception>> getCheckOperations(final J
         PostgresUtils.checkFirstRecordWaitTime(config);
       });
 
-      // Verify that the db user has required privilege to perform replication.
+      // Verify that the db user has required privilege to perform replication. The user has to be super user of to has replication access.
```
rodireich, line 321: super user of db?

sergio-ropero: My bad .. I will modify the comment.
Basically the db user needs to be superuser or to have replication privilege.

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
sergio-ropero: I have asked to the user to understand how they set up. Based on that I will adapt the PR, but this should be enough (supposedly)

subodh1810 (commented):
The more I think about the permission checks, the more I feel that we should just instead establish a replication connection to the source DB and run the `IDENTIFY_SYSTEMS` query to make sure that the connection is valid. If the user has right permissions it will work otherwise it would fail.

evantahler:
There's a lot going on here... perhaps a rebase/recreate from master might be needed after the monorepo changes

sergio-ropero:
> 

Yeah... I mess up a bit because this branch was created from master and being updated from there frequently. Then I rebased if from this integration branch (`dbdw-srcs-integration-code-freeze-feb23`) which allows us to run the tests for postgres.
In that rebase process I had some headaches because of master commits that were not in this branch, but I sorted it. I think it's fine now:
Screenshot 2023-02-21 at 17 56 45

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/PostgresDebeziumStateUtil.java
```diff
@@ -111,8 +90,8 @@ public void commitLSNToPostgresDatabase(final JsonNode jdbcConfig,
 
     final LogSequenceNumber logSequenceNumber = LogSequenceNumber.valueOf(savedOffset.getAsLong());
 
-    try (final BaseConnection pgConnection = ((BaseConnection) connection(jdbcConfig))) {
-      validateReplicationConnection(pgConnection);
+    try {
+      final BaseConnection pgConnection = (BaseConnection) PostgresDebeziumConnection.createConnection(jdbcConfig);
```
subodh1810, line 94: this should be in a try block otherwise the connection will not be closed

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -318,21 +320,14 @@ public List<CheckedConsumer<JdbcDatabase, Exception>> getCheckOperations(final J
...
-        final List<JsonNode> userPrivileges = database.queryJsons(connection -> {
-          final PreparedStatement ps = connection.prepareStatement("SELECT userepl FROM pg_user WHERE usename = ?");
-          ps.setString(1, userName);
-          LOGGER.info("Verifying required privileges for user: {}", userName);
-          return ps;
-        }, sourceOperations::rowToJson);
-
-        if (!userPrivileges.get(0).get("userepl").asBoolean()) {
+        try {
+          PostgresDebeziumConnection.createConnection(config);
```
subodh1810, line 327: We need to close the connection as well so we can put this in try with resources block

subodh1810 (changes requested):
We are not closing the replication connection which can cause issues.

sergio-ropero:
Closing and creating a brand new PR

Issue #22573, opened 2023-02-08
Connector builder projects: Add fetch with manifest API for builder projects


flash1293:
Introduce APIs for reading a builder project with the manifest draft

```
POST /v1/api/connector_builder_projects/get_with_manifest
{
    "builderProjectId": "<id>",
    "workspaceId": "<workspace>"
}
```

Response
```
{
    "builderProject": {
       "name": "..:",
       "builderProjectId": "<id>",
       "actorDefinitionId": "<id>", // might be undefined
       "hasDraft": true,
       "activeDeclarativeManifestVersion": 1, // might be undefined
    },
    "declarativeManifest": {
       "isDraft": true,
       "manifest": {...}
    } 
}
```

For all apis the following constraints apply:
* Builder project doesn’t exist in workspace 404
* Name needs to be non-empty
* Draft can be “null” as well to unset it completely

## Acceptance criteria

* All APIs are functional as described above
* All APIs are unit-tested
* All introduced APIs need to be wrapped in cloud with Editor permission for workspace

Issue #22445, opened 2023-02-06
Connector builder projects: Add list API for builder projects


flash1293:
Introduce API for listing builder projects of the current workspace

```
POST /v1/api/connector_builder_projects/list
{
    "workspaceId": "<workspace>"
}
```

Response
```
{
  "builderProjects": [
    {
        "builderProjectId": "<id>",
        "name": "My draft-only connector“,
        "hasDraft": false,
    },
    ...
  ]
}
```

If possible do not fetch the actual draft from the database into java memory but only a boolean column whether it is set to null or not.

For all apis the following constraints apply:
* Builder project doesn’t exist in workspace 404
* Name needs to be non-empty
* Draft can be “null” as well to unset it completely

## Acceptance criteria

* All APIs are functional as described above
* All APIs are unit-tested
* All introduced APIs need to be wrapped in cloud with Editor permission for workspace

lmossman:
Grooming notes:
- Estimate for just one endpoint: 3
- @flash1293 to split up into one issue per endpoint

PR #23143, opened 2023-02-16
Connector builder projects: List and get APIs

flash1293:
## What

Based on https://github.com/airbytehq/airbyte/pull/23063

Fixes https://github.com/airbytehq/airbyte/issues/22445
Fixes https://github.com/airbytehq/airbyte/issues/22573

By adding APIs to list all builder projects of a workspace as well as fetching a single project with its manifest.

## How

* Extend open api spec: `airbyte-api/src/main/openapi/config.yaml`
* Extend the builder project fetching with a calculated field `hasDraft` (`airbyte-config/config-models/src/main/resources/types/ConnectorBuilderProject.yaml`, `airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/DbConverter.java`, `airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java`) which is used for the listing to know whether there is a draft without actually fetching it.
* Fix an existing problem of not storing a string `"null"` instead of actually no value if there is no manifest draft defined (also adding tests for this) - this happened because JSON serialization will turn a Java `null` into a JSON string `"null"`. Added a branch to https://github.com/airbytehq/airbyte/pull/23143/files#diff-2d29b293778e290157be96e22294c2df1fdfa7c72a48961d03396776446deeddR1663 to fix this

For testing and general notes see description of https://github.com/airbytehq/airbyte/pull/23063


airbyte-config/config-models/src/main/resources/types/ConnectorBuilderProject.yaml
```diff
@@ -21,6 +21,8 @@ properties:
   actorDefinitionId:
     type: string
     format: uuid
+  hasDraft:
```
girarda, line 24: why are we adding a flag instead of checking if `manifestDraft == null`?
is there a risk the two fields could get out of sync?

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandlerTest.java
```diff
@@ -97,6 +108,26 @@ void testUpdateConnectorBuilderProject() throws IOException, JsonValidationExcep
...
+    when(configRepository.getConnectorBuilderProject(project.getBuilderProjectId(), false)).thenReturn(project);
+
+    final ExistingConnectorBuilderProjectWithWorkspaceId update = new ExistingConnectorBuilderProjectWithWorkspaceId()
+        .builderProject(new ConnectorBuilderProjectDetails().name(project.getName()))
+        .workspaceId(workspaceId).builderProjectId(project.getBuilderProjectId());
+
+    connectorBuilderProjectsHandler.updateConnectorBuilderProject(update);
+
+    verify(uuidSupplier, never()).get();
+    verify(configRepository, atMostOnce())
```
girarda, line 126: why `atMostOnce()` instead of exactly once?

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandlerTest.java
```diff
@@ -97,6 +108,26 @@ void testUpdateConnectorBuilderProject() throws IOException, JsonValidationExcep
             project);
   }
 
+  @Test
+  @DisplayName("updateConnectorBuilderProject should update an existing project removing the draft")
+  void testUpdateConnectorBuilderProjectWipeDraft() throws IOException, ConfigNotFoundException {
+    final ConnectorBuilderProject project = generateBuilderProject();
+    project.setManifestDraft(null);
```
girarda, line 115: is the test's DisplayName right? It looks like the manifest is set to null before running the update

airbyte-server/src/main/java/io/airbyte/server/apis/ConnectorBuilderProjectApiController.java
```diff
@@ -52,6 +55,24 @@ public void deleteProject(final ConnectorBuilderProjectIdWithWorkspaceId connect
...
+  @Secured({EDITOR})
+  @SecuredWorkspace
+  public ConnectorBuilderProjectRead getProject(final ConnectorBuilderProjectIdWithWorkspaceId connectorBuilderProjectIdWithWorkspaceId) {
+    return ApiHelper.execute(() -> connectorBuilderProjectsHandler.getBuilderProjectWithManifest(connectorBuilderProjectIdWithWorkspaceId));
+  }
+
+  @Override
+  @Post(uri = "/list")
+  @Status(HttpStatus.OK)
+  @Secured({EDITOR})
```
girarda, line 70: what is the editor {EDITOR} role?

girarda (changes requested):
the code looks good but I'm rejecting with a request for more information. will approve when the questions are answered.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandler.java
```diff
@@ -4,20 +4,27 @@
...
-import io.airbyte.validation.json.JsonValidationException;
 import jakarta.inject.Inject;
 import jakarta.inject.Singleton;
 import java.io.IOException;
+import java.util.Map;
 import java.util.UUID;
 import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 @SuppressWarnings("PMD.AvoidCatchingNPE")
```
xiaohansong, line 29: why is this needed?

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -1602,12 +1603,15 @@ public boolean getConnectionHasAlphaOrBetaConnector(final UUID connectionId) thr
     return countResult > 0;
   }
 
+  private static final List<Field<?>> baseConnectorBuilderProjectColumns =
```
xiaohansong, line 1606: nit: should move to the top of the file and use CAPITAL_CASE format for static final object.

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
```diff
@@ -1627,8 +1631,7 @@ public Stream<ConnectorBuilderProject> getConnectorBuilderProjectsByWorkspace(fi
 
     return database
```
xiaohansong, line 1632: should we add a @NotNull annotation for the input `workspaceId` to make sure it's not misused?

airbyte-config/config-models/src/main/resources/types/ConnectorBuilderProject.yaml
flash1293: The flag is only in the POJO, not in the database - it's to avoid loading the full draft from database into Java memory just to check whether it's null

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandlerTest.java
flash1293: should be exactly once, fixed

airbyte-commons-server/src/test/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandlerTest.java
flash1293: It is correct, the order of operations is a little confusing - it is set to `null` for the validation by the end of the file whether the project passed to `writeBuilderProject` doesn't have a draft. I rearranged the code to make this clearer.

airbyte-server/src/main/java/io/airbyte/server/apis/ConnectorBuilderProjectApiController.java
flash1293: This is the permission level for cloud - I checked with Jonathan that editor is the right one for us - it's the role a regular user gets when logged in in the frontend for their own workspace. We can expand privileges later on but this is the important one.

airbyte-commons-server/src/main/java/io/airbyte/commons/server/handlers/ConnectorBuilderProjectsHandler.java
flash1293: it's not, removed

airbyte-config/config-persistence/src/main/java/io/airbyte/config/persistence/ConfigRepository.java
flash1293: done

PR #23152, opened 2023-02-16
Source Intercom: updated test_switch_to_standard_endpoint and test_companies_scroll

darynaishchenko:
updated test_switch_to_standard_endpoint and test_companies_scroll: changed len(records) and total time
added TypeTransformer for stream companies


bazarnov:
@lazebnyi @darynaishchenko 
FYI, these tests fixed are omitted along with the Intercom (Low Code) release, since there is no need to test switch between 2 endpoints, because it creates throttling on the connector's side. Low Code's version uses the scroll endpoint effectively as the only option to fetch the companies.

bazarnov (approved):

lazebnyi (changes requested):
Please add allowedHosts

lazebnyi (approved):

PR #23170, opened 2023-02-16
Destination BigQuery (+denormalized): Fix build

edgao:
our dependencies are a complete mess, but we should be ok:
* previously we were using google-cloud-storage 2.2.2, which pulled in google-api-client 1.x
* now we're on google-cloud-storage 2.17.2, which declares google-api-client 2.x
* destinations bigquery (+denorm) now force that back to 1.31.5, because google-cloud-bigquery:1.133.1 requires google-api-client 1.x
* but google-cloud-storage 2.17 is ok with google-api-client 1.x OR 2.x, as long as it's >= 1.31.1

also some notes about the absolute state of this dependency tree:
* bigquery depends on `com.google.cloud:google-cloud-bigquery:1.122.2`
* also depends on airbyte-db:db-lib -> `com.google.cloud:google-cloud-bigquery:1.133.1`
* also depends on `com.google.cloud:google-cloud-storage:2.4.5` 
* protocol-models -> `com.google.cloud:google-cloud-storage:2.17.2` -> `com.google.api-client:google-api-client:2.1.2` (this comes from deps.toml)
* destination-jdbc -> `com.google.cloud:google-cloud-storage:1.113.16`

submitted https://github.com/airbytehq/airbyte-internal-issues/issues/1439 to make this better; I don't have a well-informed opinion on what the right path forward is.


airbyte-integrations/connectors/destination-bigquery-denormalized/build.gradle
```diff
@@ -31,3 +31,9 @@ dependencies {
 
     implementation files(project(':airbyte-integrations:bases:base-java').airbyteDocker.outputs)
 }
+
+configurations.all {
+  resolutionStrategy {
+    force 'com.google.api-client:google-api-client:1.31.5'
```
ryankfu, line 37: Was this the version of `google-api-client` we were using before the updated versions or the version most compatible with all versions `2.17.X` and `2.22.X`

ryankfu (approved):
LGTM, should we also run `base-normalization` integration tests as well since this was originally a blocker for that?

airbyte-integrations/connectors/destination-bigquery-denormalized/build.gradle
edgao: just what we had previously. I didn't want to spend even more time validating different versions to find the latest compatible one

PR #23173, opened 2023-02-16
🪟 🚦 Add e2e tests to confirm that schema changes detected UI is gone after refresh

edmundito:
## What
This adds a couple of tests to verify that after being notified of non-breaking or breaking changes, and the schema is refreshed that leads to no changes, the UI is cleared and the "no changes" toast is displayed.

## How
* New test cases added
* Moved autoDetectSchema spec as it is intended to be inside the connection subfolder


dizel852 (approved):
LGTM 👍 
Have not tested locally - seems CI test run is ok

PR #23182, opened 2023-02-16
:window: :tada: Handle existing service tokens

### LINKED ISSUE #23018: Allow deleting (or at least removing) dbt Cloud service token from workspace

ambirdsall:
## There are a few issues with the MVP dbt Cloud settings page
- it gives no visual indication if a service token has already been saved, which is suboptimal
- while you can replace an existing token with a different one, there is currently no way to remove the integration altogether, which is table stakes functionality for this feature going forwards
- speaking of issues: this PR closes https://github.com/airbytehq/airbyte/issues/23018 and https://github.com/airbytehq/airbyte/issues/23017

## So what are you going to do about it, huh?
- in a workspace with no dbt Cloud integration, there are no apparent changes: users see the text input for entering a service token and a submit button
- in a workspace with an existing integration, users see a pair of buttons: one to replace the current service token and one to remove it
- clicking the "replace service token" button replaces the button group with the service token text input, but with an additional "cancel" button to restore the button group
- clicking the "delete service token" button does what it says on the tin; when the deletion request successfully returns, the page rerenders in its "no existing integration" state

## A bit more about deletion
Although there is [an open issue about it](https://github.com/airbytehq/airbyte-internal-issues/issues/6028), there's not currently a way to properly delete the service token in the backend. What we can do is use a PATCH-type update to set the workspace's webhook operator configs to a list devoid of their dbt Cloud token.

## How about some pictures tho
Great idea.

dbtcloud--managing-tokens_360

### DIFF
--- a/airbyte-webapp/src/packages/cloud/views/settings/integrations/DbtCloudSettingsView.tsx
+++ b/airbyte-webapp/src/packages/cloud/views/settings/integrations/DbtCloudSettingsView.tsx
@@ -1,3 +1,6 @@
+import { faRecycle, faTrash } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import classNames from "classnames";
 import { Field, FieldProps, Form, Formik } from "formik";
 import React, { useState } from "react";
 import { FormattedMessage, useIntl } from "react-intl";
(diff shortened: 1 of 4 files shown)


airbyte-webapp/src/packages/cloud/views/settings/integrations/DbtCloudSettingsView.tsx
```diff
@@ -6,73 +9,125 @@ import { LabeledInput } from "components/LabeledInput";
...
-import { useSubmitDbtCloudIntegrationConfig } from "packages/cloud/services/dbtCloud";
+import { useDbtCloudServiceToken } from "packages/cloud/services/dbtCloud";
 import { SettingsCard } from "pages/SettingsPage/pages/SettingsComponents";
 import { links } from "utils/links";
 
 import styles from "./DbtCloudSettingsView.module.scss";
 
+const cleanedErrorMessage = (e: Error): string => e.message.replace("Internal Server Error: ", "");
+
 export const DbtCloudSettingsView: React.FC = () => {
```
ambirdsall, line 20: This is pretty significantly restructured:
- large swaths of the returned TSX expression are now conditional on whether there's an existing token and/or whether there user is actively entering a new one
- so I extracted two main subtrees to named variables: `ServiceTokenForm` for entering a token and `ReplaceOrDeleteToken` for managing existing tokens.
- Because those subtrees share so much state from the parent components' hooks, they're defined inside its scope. Having to scroll past them to see the final return value, which is as simple and high-level as I could easily make it, is not ideal for readability, though.

PR #23187, opened 2023-02-17
sunset kustomize

cgardens:
## What
* We are sunsetting the kustomize kube setup in favor of the helm approach.

## How
* Remove kustomize
* Remove any related scripts

## To Do
- [ ] `acceptance_test_kube_gke.sh` uses kustomize. need to switch it to use helm. Update: or maybe we don't care? this hasn't past in ages. we should have this test, but probably create an issue to bring it back with helm later? No point in blocking the sunset on it though, it's already busted.
Screen Shot 2023-02-16 at 8 56 41 PM

## Recommended reading order
1. `x.java`
2. `y.python`

## 🚨 User Impact 🚨
* We are sunsetting helm. Docs will be updated to reflect how to keep it running for people already using it. Otherwise, new users should use helm. Docs will happen in a separate PR.


cgardens:
@davinchia , do you have any context on `acceptance_test_kube_gke.sh`? It looks like it runs daily but hasn't succeeded in ages. It uses kustomize which we are deprecating. I'm inclined to just delete it. What I'm not sure of is whether this is a test suite that we should create a story to bring back using helm.

davinchia:
@cgardens we have duplicate GKE tests in Cloud. This one was written when we first got on GKE and discovered small differences among all the Kube implementations, and forgotten among everything. Let's delete it for now.

davinchia (approved):
Thanks @cgardens , this looks good to me!

We also discussed announcing this to the community. What's the plan there? Should we involve Shawn?

PR #23188, opened 2023-02-17
[connector builder] Return string format to allow for connector read() that return non-object responses

### LINKED ISSUE #20764: Connector builder server: Error if body is a json-encoded array
The connector builder server only responds with
```
"Could not perform read with with error: [ErrorWrapper(exc=DictError(), loc=('body',))]"
```

if the body of the response is a an array instead of ab object. The same manifest works fine if run directly via python.

Example manifest:
```
version: 0.1.0
type: DeclarativeSource
check:
  type: CheckStream
  stream_names: []
streams:
  - name: accounts
    type: DeclarativeStream
    retriever:
      type: SimpleRetriever
      name: accounts
      requester:
        name: accounts
        url_base: https://api.timelyapp.com/1.1
        path: /accounts
        authenticator:
          type: BearerAuthenticator
          api_token: '{{ config[''api_key''] }}'
      record_selector:
        type: RecordSelector
        extractor:
          type: DpathExtractor
          field_pointer: []
      paginator:
        type: NoPagination
spec:
  type: Spec
  documentation_url: http://example.org
  connection_specification:
    $schema: http://json-schema.org/draft-07/schema
    type: object
    required:
      - api_key
    properties:
      api_key:
        type: string
        title: API Key
        airbyte_secret: true
    additionalProperties: true
```

## Solution
The server can return the body as a string and let the webapp convert it to json.
```
    HttpResponse:
      type: object
      required:
        - status
      properties:
       <...>
        body:
          type: string
          description: String representation of the body of the HTTP response, if present
       <..>
```

### Alternative
* Change the type of HttpResponse.body to anyOf(object, array, string, number, boolean)
* This approach has additional complexity because the OpenAPI generator for Python does not support anyOf

## Acceptance criteria
- The connector builder can return responses where the body is a any valid json data type (string, number, object, array, boolean, null)
- [ ] pair with frontend eng @lmossman likely to finish frontend work

## Out of scope
- Support for non-json data types

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/20764

## What
When we first implement request/response in the connector builder server, we operated under the assumption that responses would always be in an object format. However, in many cases, APIs return responses as a list or possibly even a string. In these cases we would fail the request with:
```
"Could not perform read with with error: [ErrorWrapper(exc=DictError(), loc=('body',))]"
```

## How
We've decided to let the frontend code handle the parsing of the response into it's relevant class due to limitations of the openapi spec not allowing for `anyOf`. This PR changes the openapi spec to reflect that responses will come in the form of a string. This also performs the small change to no longer parse the response body into json object and pass the escaped string w/o any modification.

The result should look like:
```
{
    "status": 200,
    "body": "[{\"id\":937971,\"name\":\"Pliancy - Staging\",\"color\":\"313d4b\",\"currency\":{\"id\":\"usd\",\"name\":\"United States Dollar\",\"iso_code\":\"USD\",\"symbol\":\"$\",\"symbol_first\":true},\"from\":\"Web\",\"max_users\":0,\"seats\":11,\"max_projects\":0,\"plan_id\":221,\"plan_name\":\"Unlimited\",\"next_charge\":\"2023-02-18\",\"industry\":\"\",\"memory_retention_days\":0,\"num_users\":10,\"num_projects\":55,\"active_projects_count\":55,\"total_projects_count\":95,\"status\":\"active\",\"beta\":false,\"azure_ad_enabled\":false,\"expired\":false,\"trial\":false,\"days_to_end_trial\":0}]",
    "headers": {
        "something": "here"
    }
}
```

## Recommended reading order
1. `openapi.yaml`
2. `default_api.py`

### DIFF
--- a/airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/PageDisplay.tsx
+++ b/airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/PageDisplay.tsx
@@ -40,7 +40,16 @@ export const PageDisplay: React.FC<PageDisplayProps> = ({ page, className, infer
 
   const formattedRecords = useMemo(() => formatJson(page.records), [page.records]);
   const formattedRequest = useMemo(() => formatJson(page.request), [page.request]);
-  const formattedResponse = useMemo(() => formatJson(page.response), [page.response]);
+  const formattedResponse = useMemo(() => {
+    if (!page.response) {
+      return "";
+    }
+    const unpackedBodyResponse = {
+      ...page.response,
+      body: page.response.body ? JSON.parse(page.response.body) : undefined,
+    };
+    return formatJson(unpackedBodyResponse);
+  }, [page.response]);
   const formattedSchema = useMemo(() => inferredSchema && formatJson(inferredSchema, true), [inferredSchema]);
 
   let defaultTabIndex = 0;
--- a/airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/PageDisplay.tsx
+++ b/airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/PageDisplay.tsx
@@ -41,12 +41,21 @@ export const PageDisplay: React.FC<PageDisplayProps> = ({ page, className, infer
   const formattedRecords = useMemo(() => formatJson(page.records), [page.records]);
   const formattedRequest = useMemo(() => formatJson(page.request), [page.request]);
   const formattedResponse = useMemo(() => {
-    if (!page.response) {
+    if (!page.response || !page.response.body) {
       return "";
     }
+    let parsedBody: unknown;
+    try {
+      // body is a string containing JSON most of the time, but not always.
+      // Attempt to parse and fall back to the raw string if unsuccessfull.
+      parsedBody = JSON.parse(page.response.body);
+    } catch {
+      parsedBody = page.response.body;
+    }
+
     const unpackedBodyResponse = {
       ...page.response,
-      body: page.response.body ? JSON.parse(page.response.body) : undefined,
+      body: parsedBody,
     };
     return formatJson(unpackedBodyResponse);
   }, [page.response]);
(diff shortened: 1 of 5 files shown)


flash1293:
@brianjlai adjusted two things:
* Fix formatting issues in connector builder server that caused frontend build to fail
* Adjusted frontend to unpack the json string to render it correctly, works now even if the body is an array:
Screenshot 2023-02-17 at 11 21 36

One thing I would like to adjust though - with the new API there is no indication `body` is actually encoded JSON, could we rename this property to `bodyJsonString` or something like this to make it obvious?

brianjlai:
Yeah that should be relatively easy, and the frontend would just need to be updated in the same block you added to use `bodyJsonString` as well. The only debate would be if we ever expect responses to be non-json, like just a plain string or something. I left it flexible enough to receive a string response from the API and pass it back through to the frontend.

In my opinion, it will be more confusing later down the line for it to be called `bodyJsonString` when it can also expect a regular string, so even though it's not obvious, `body` might still be the accurate term

airbyte-webapp/src/components/connectorBuilder/StreamTestingPanel/PageDisplay.tsx
```diff
@@ -40,7 +40,16 @@ export const PageDisplay: React.FC<PageDisplayProps> = ({ page, className, infer
...
   const formattedRecords = useMemo(() => formatJson(page.records), [page.records]);
   const formattedRequest = useMemo(() => formatJson(page.request), [page.request]);
-  const formattedResponse = useMemo(() => formatJson(page.response), [page.response]);
+  const formattedResponse = useMemo(() => {
+    if (!page.response) {
+      return "";
+    }
+    const unpackedBodyResponse = {
+      ...page.response,
+      body: page.response.body ? JSON.parse(page.response.body) : undefined,
```
brianjlai, line 49: as mentioned in the other comment, the connector builder server could return a plain string if the calling API allows for it. Can we use a conditional block to test the body string is valid json first, or try/catch on the json parse. Whichever is idiomatic for typescript

flash1293: If the contract is “this is the body, it might be json, it might be not”, I agree with keeping it named body - I will add the try/catch on Monday to be future-safe

brianjlai:
Closing this one because we'll need to merge it into the split platform repo now in https://github.com/airbytehq/airbyte-platform/pull/118

Issue #20703, opened 2022-12-20
Salesforce Stream "Account" cannot be processed by REST or BULK API


koconder:
## Environment
- **Airbyte version**: 0.40.25
- **OS Version / Instance**: Kubernetes
- **Deployment**: Kubernetes via Helm Chart
- **Source Connector and version**: Salesforce (1.0.27)
- **Destination Connector and version**: Amazon S3 (0.1.26) 
- **Step where error happened**: Sync Job

## Current Behavior
When i connect to my Salesforce instance, i am unable to fetch metadata unless i exclude using connection filters where object equals "Account".

Error in pod logs:
```
Exception: Stream Account cannot be processed by REST or BULK API.
2022-12-20 08:01:18 [1;31mERROR[m i.a.w.i.DefaultAirbyteStreamFactory(internalLog):116 - Stream Account cannot be processed by REST or BULK API.
Traceback (most recent call last):
  File "/airbyte/integration_code/main.py", line 13, in <module>
    launch(source, sys.argv[1:])
  File "/usr/local/lib/python3.9/site-packages/airbyte_cdk/entrypoint.py", line 131, in launch
    for message in source_entrypoint.run(parsed_args):
  File "/usr/local/lib/python3.9/site-packages/airbyte_cdk/entrypoint.py", line 116, in run
    catalog = self.source.discover(self.logger, config)
  File "/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/abstract_source.py", line 70, in discover
    streams = [stream.as_airbyte_stream() for stream in self.streams(config=config)]
  File "/airbyte/integration_code/source_salesforce/source.py", line 110, in streams
    streams = self.generate_streams(config, stream_objects, sf)
  File "/airbyte/integration_code/source_salesforce/source.py", line 95, in generate_streams
    raise Exception(f"Stream {stream_name} cannot be processed by REST or BULK API.")
```

## Expected Behavior
Data should sync, metadata should be shown.

koconder:
Issue comes from:
* https://github.com/airbytehq/airbyte/pull/16086 @grubberr @bazarnov @edgao 
* https://github.com/airbytehq/airbyte/pull/9302 @sergei-solonitcyn @vitaliizazmic @augan-rymkhan

tagged original reviewers and committers

poolmaster:
The issue seemed to be that bulk API and REST API cannot cover tables with "many" columns and "special" datatypes.
[...]

PR #23190, opened 2023-02-17
Source Salesforce: include primary key in every chunk

davydov-d:
## What
https://github.com/airbytehq/airbyte/issues/20703
*this is a followup for the on-call issue https://github.com/airbytehq/oncall/issues/1403*

## How
If the properties are chunked - we need to include the primary key in every chunk, otherwise we won't be able to stick response chunks


airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -136,13 +136,16 @@ def request_params(
     def chunk_properties(self) -> Iterable[Mapping[str, Any]]:
         selected_properties = self.get_json_schema().get("properties", {})
 
+        def empty_props():
```
bazarnov, line 139: This should have another naming, how about `include_primary_key()`?

bazarnov (approved):

maxi297 (approved):
LGTM!

bazarnov (approved):

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
davydov-d: renamed

lazebnyi (approved):

PR #23193, opened 2023-02-17
🪟 🎉 New Loading Page

timroes:
## What

Replaces our full page loading spinner with an animated Airbyte logo instead. Also turns the `LoadingPage` component over to SASS.

https://user-images.githubusercontent.com/877229/219646832-3a8661a2-9cee-4ab6-927e-52afc26b05bf.mp4

### Notes:
* Joey and I discussed whether we want to wait for an animation cycle to finish before fading it out, but (aside technical complexity), we would in the worst case slow down page loading by 2s, which we decided is not worth it.
* We currently render the LoadingPage multiple times during page load (while we're jumping between React Suspend levels/fallbacks). While we could build some way that this is rendered in a wrapping context and all the fallbacks would basically tell the context that they still need to loading spinner, and that shows it - so that animation wouldn't jump between those loadings - we decided not to go that route (for now), due to adding a lot of additional technical complexity, for a very minor design win.


joeykmh (approved):
Looks good! I would almost say we should go the extra mile here and avoid as many stacked `<Suspense fallback={Loading} />` renders as possible (which interrupt the animation). But I don't want to block shipping this - we can improve it in iterations.

Issue #23197, opened 2023-02-17
Source S3: fix schema discovery for Avro


davydov-d:
Currently, S3 streams are discovered with a schema that does not match real data types if the file format is Avro and the data structure is nested:

{"id": 0, "fullname_and_valid": {"fullname": "cfjwIzCRTL", "valid": false}

`fullname_and_valid` here must be `object` but is discovered as a string

PR #23198, opened 2023-02-17
Source s3: fix avro discovery

davydov-d:
## What
https://github.com/airbytehq/airbyte/issues/23197

## How
Make nested avro objects `object`, not `string`

Let's assume we have record `{"id": 0, "fullname_and_valid": {"fullname": "cfjwIzCRTL", "valid": false}`
This record's schema used to be `{"id": "integer", "fullname_and_valid": "string"}`
Now it is `{"id": "integer", "fullname_and_valid": "object"}`


airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/formats/avro_parser.py
```diff
@@ -50,7 +50,7 @@ def _parse_data_type(self, data_type_mapping: dict, avro_schema: dict) -> dict:
                 schema_dict[i["name"]] = [data_type_mapping[dtype] for dtype in data_type]
             # TODO: Figure out a better way to handle nested records. Currently a nested record is returned as a string
```
maxi297, line 51: Is this solved and therefore we should remove the TODO?

maxi297 (commented):
Hey @davydov-d ! I get the specific details of the issue but not the broader context. What benefits does it have for the schema to be returning `objects` instead of `string` for nested objects? As part of the schema, should we also evaluate the "object" so that we have `nested_schema_output = {"lastname": "string", "address": {"streetaddress": "string", "city": "string }}` instead of just `nested_schema_output = {"lastname": "string", "address": "object"}`?

airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/formats/avro_parser.py
davydov-d: @maxi297 well first of all the data type of our records should match the schema that is discovered, otherwise we will have normalization issues. As for now, nested data is emitted as is - an object inside another object. At the same time we declare it to be a string.. So we have to either modify the emitted data or the stream schema. This PR is rather a short term solution, a hotfix to unblock other PRs, so I decided to modify stream schema. As a long term solution, I believe, few more things should be done:
 - improve avro to json schema convertion (so we could be more precise in description of nested objects, arrays etc.)
 - implement json to avro schema convertion
 - use the json-converted-to-avro schema when reading the file, so that, for instance, we could pass a user enforced schema and read that nested object as a string
 
 So, answering your questions - yes, we should evaluate the `object` to be more descriptive, but I believe this should be done in a separate PR. This also means that the `todo` comment is not resolved

bazarnov (approved):

airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/formats/avro_parser.py
maxi297: Perfect! Do we have an issue for this? Should it be added to the comment?

maxi297 (approved):

airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/formats/avro_parser.py
davydov-d: Sure, done!

lazebnyi (approved):

davydov-d:
Hey @erica-airbyte @jnr0790 here we have another PR with breaking changes that should be released before https://github.com/airbytehq/oncall/issues/1470.

This will affect users of S3 source working with Avro files and destinations that support normalization.
 - the reason for the breaking change is that complex data from avro files is now discovered as object/array/etc
 - nested avro structures used to be strings, now their discovered type should correspond to the actual data type
 - the data stays the same, but the schema changes, that may affect normalization
 - once the changes goes live, customers need to refresh source schema and reset the data 

Please let me know when we're good to release this one.

erica-airbyte:
@davydov-d we have no users using Avro so you can deploy whenever

jnr0790:
@davydov-d no one is using Arvo so you can go ahead and deploy

PR #23202, opened 2023-02-17
don't run airbyte ci on schedule

c-p-b:
## What
Ever since we started looking only at modified files, this workflow has been wasting minutes running on cron and doing nothing


flash1293 (approved):
🙏

PR #23205, opened 2023-02-17
🪟 🚦 Cleanup connection e2e test page objects

edmundito:
## What
This change reorganizes the connections test so that

* Focuses on the basic create connection flow
* Testing steps can be isolated
* Page objects are used in a more object-oriented way

## How
* newTable.spec: renamed to createConnection. Group tests into `describe` blocks
* Moves all connection related pageobjects to `pages/connection` and renames them to end with `PageObject` to make it more clear during omnisearch
* Splits the Replication page object into ConnectionForm and Streams Table
* Makes Streams Table page object into a base class and interface so that the two versions have the same interface.


krishnaglick (approved):
Not sure why git says the file was deleted.

krishnaglick (approved):
This is awesome stuff! Good work!

Issue #22130, opened 2023-01-31
Speed up Testing & Publishing a connector


evantahler:
The goal of this story is to speed up our `/test` and `/publish` comands. 

Say you are [publishing a python connector](https://github.com/airbytehq/airbyte/actions/runs/4051035949/jobs/6969001007).  The `/publish` task takes over 10 minutes, and most of that time is spent on sub-tasks that are not relevant to the connector.  There are some strange dependencies in our grade tasks. 

For example, running the SAT test for a connector (e.g. `./gradlew :airbyte-integrations:connectors:source-faker:integrationTest`) tests normalization (`:airbyte-integrations:bases:base-normalization:checkPython`), and builds and tests all of SAT locally, including SAT's own test suite (`:airbyte-integrations:bases:source-acceptance-test:_unitTestCoverage`).  

This issue is to think about our gradel DAG and look for improvements:
* Why is normalization involved at all?  I'll bet we build the normalization imagage in-case the connector is a destination needs normalization.  Can we be smarter and only build normalization for destinations?
* If there are no changes to the local SAT directory, can we use the `#latest` docker tag, or at the very least, skip SATs own internal tests?
* Should the publish command's `manage.sh` script use `gradle --scan` so we can better track/time these things?

sherifnada:
great minds think alike. Just created https://github.com/airbytehq/airbyte/issues/22128 which should fix some of these problems

> Should the publish command's manage.sh script use gradle --scan so we can better track/time these things?
yes -- free win!

evantahler:
> great minds think alike. Just created https://github.com/airbytehq/airbyte/issues/22128 which should fix some of these problems
> 
> > Should the publish command's manage.sh script use gradle --scan so we can better track/time these things?
> 
> yes -- free win!

Nice! I'll close this and we can keep your issue - the memes are better

alafanechere:
I'm re-opening because I think there are test specific optimization a bit different from the one mentioned in `SUB_BUILD=CONNECTOR_BASE` #22128

alafanechere:
https://github.com/airbytehq/airbyte/pull/23332 was also an improvement in speeding up connector testing.

alafanechere:
Connector /test are not testing and building connector-acceptance-test anymore. 
We should perceive a speed improvement of ~10mn on /test .

PR #23208, opened 2023-02-17
connector testing: remove superfluous tasks

alafanechere:
## What
Issue https://github.com/airbytehq/airbyte/issues/22130
This is a quick-win attempt at reducing test/build time for connectors.
Non-goals: investigate incremental build optimizations.

## How
What was wrong:
- Connector acceptance tests and base-normalization were built (and tested for CAT) on connector tests.

What I changed:
-  Use `latest` published *connector acceptance test* image on `/test` instead of building its image locally.  Developers can override this if they want to `/test` connector on their WIP CAT change by manually changing/committing [this line](https://github.com/airbytehq/airbyte/blob/ad2f39207bf0b5802476ee181205a9c1bb78559d/buildSrc/src/main/groovy/airbyte-connector-acceptance-test.gradle#L8).
- Remove a legacy *hack* that I don't understand the purpose of. It runs connector-acceptance-test and base-normalization builds on connector test.

## 🚨 User Impact 🚨
* 🎉 **Speed gain**:  11mn faster on `source-pokeapi` ([PREVIOUSLY](https://github.com/airbytehq/airbyte/actions/runs/4183241096) vs [THIS PR CHANGE](https://github.com/airbytehq/airbyte/actions/runs/4205657035)). It's a 61% decrease in test duration... from 18mn to 7mn.

TODO: test this change on CDK, Java sources and Java destinations.


sherifnada (approved):
Great find!

buildSrc/src/main/groovy/airbyte-connector-acceptance-test.gradle
```diff
@@ -1,10 +1,27 @@
...
+        
         project.task('connectorAcceptanceTest') {
+            dependsOn project.build
+            dependsOn project.airbyteDocker
+            
             doFirst {
+                if (CONNECTOR_ACCEPTANCE_TEST_IMAGE != "dev") {
+                    project.exec {
+                        def args = [
+                            'docker', 'pull', 'airbyte/connector-acceptance-test:latest'
```
cpdeethree, line 20: TODO: Perhaps highly consider using a versioned image here, since currently we can publish to this a breaking change which pulls the rug out from under our test suite. I'm ok with this going in as-is for now because the speed improvements are so dramatic

cpdeethree: Is the base image static enough that we could go ahead and pin this today and you could remember to update it when you update the base image?

c-p-b (approved):
LGTM with one suggested change, either now or a TODO for later (unless you disagree, in which point we should have the convo ;) )

buildSrc/src/main/groovy/airbyte-connector-acceptance-test.gradle
alafanechere: Yep, this is something I considered.
The image is quite dynamic but mainly maintained by an identified pool of engineer I trust enough to write unit tests and locally tests their changes before publishing.
I also like that we can quickly pin the version if the latest one has a bug.

There's also no better approach than waiting for nightly builds to run for all connectors on the latest version to assess the stability of tests. If the connector tests were reasonably fast we could automatically run the branch version of connector acceptance tests on a pool of previously succeeding connectors and then autobump the versions used on master if 🟢 . But this is not something we can currently achieve.

alafanechere:
It looks like I've broken something on the java side according to the failing big query tests

buildSrc/src/main/groovy/airbyte-connector-acceptance-test.gradle
cpdeethree: 👍 
If you agree, a TODO in the code is sufficient for me then. Otherwise feel free to leave as-is.

alafanechere:
I think the error on the bigquery connector is on master too.

buildSrc/src/main/groovy/airbyte-python.gradle
```diff
@@ -149,11 +149,6 @@ class AirbytePythonPlugin implements Plugin<Project> {
                 command = "install -r requirements.txt"
                 inputs.file('requirements.txt')
                 outputs.file('build/installedlocalreqs.txt')
-
-                // HACK: makes all integrations depend on installing requirements for bases. long term we should resolve deps and install in order.
-                if (project.getPath().startsWith(":airbyte-integrations:connectors") && !project.hasProperty("reports_folder")) {
-                    dependsOn project.rootProject.getTasksByName("airbytePythonApply", true).findAll { it.project.getPath().startsWith(":airbyte-integrations:bases") }
-                }
```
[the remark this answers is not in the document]
davinchia: Sherif knows more than me here so I want him to confirm.

My thoughts:
I don't think we can remove this without changing more dependencies. This block basically makes each connector submodule depend on all the submodules in the bases folders e.g. (base, base-java, base-java-s3), this is required for connectors whose docker file depends on a `dev` base image. See [this](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/destination-bigquery/Dockerfile#L1) for example.

To remove this, I think we would need to go through all the connector dockerfiles and make sure they either use a versioned base image so we don't have any local image expectations.

buildSrc/src/main/groovy/airbyte-connector-acceptance-test.gradle
```diff
@@ -1,10 +1,27 @@
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 
+
 class AirbyteConnectorAcceptanceTestPlugin implements Plugin<Project> {
+    
+    // Change this value to 'dev' if you want to use a locally built connector acceptance test image
+    String CONNECTOR_ACCEPTANCE_TEST_IMAGE = 'latest'
```
pedroslopez, line 8: Could this be exposed as an argument to the command somehow? I think this means when working on acceptance test changes, we won't be able to test our changes against connectors easily

evantahler: An ENVIRONMENT variable might be a good way to handle this

alafanechere: I've added an option to the `/test` command: `connector-acceptance-test-version`, which defaults to latest.
It's then passed to gradle as a project property via env var with `ORG_GRADLE_PROJECT_connectorAcceptanceTestVersion`.
So, on a Connector Acceptance Test PR, we can run `/test connector=connectors/source-pokeapi connector-acceptance-test-version=dev` to use a locally built CAT.

alafanechere: Here's an example workflow to which I've passed connector-acceptance-test-version=dev : https://github.com/airbytehq/airbyte/actions/runs/4222475489

alafanechere:
Re-requested review from @pedroslopez and @evantahler following my parameter addition to run a custom version (`dev`) of the connector acceptance test.

buildSrc/src/main/groovy/airbyte-python.gradle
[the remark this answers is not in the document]
cpdeethree: What's the risk factor here of pushing ahead and addressing issues as they come up?

This is a massive improvement that will undoubtedly save us a not-insignificant amount of time and $$ on ci bills.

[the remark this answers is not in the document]
alafanechere: @davinchia thanks for your reply.
I'm under the impression this "hack" is safe to remove because:
1. It's python only. As you mentioned, there could be more stickiness to `base` pre-installation on Java connector, but this change is python connector specific.
2. It's always formatting and installing base-normalization and connector-acceptance-test, even if these are not defined as a dependency on the connector side which I think is a superfluous task in the context of connector build.
3. We already have the guarantee that connector acceptance tests are installed in the connector because it is defined in their `requirement.txt` ([eg](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-pokeapi/requirements.txt#L1)) and installed in the `installLocalReqs` task.
I don't know exactly why we make connector-acceptance-test a test requirements dependency of connectors and why we declare this dependency both in `setup.py` and `requirements.txt`.

@sherifnada please share I you have more details on this matter 😄 🙏

buildSrc/src/main/groovy/airbyte-connector-acceptance-test.gradle
```diff
@@ -1,10 +1,35 @@
...
+    
     void apply(Project project) {
+        
         project.task('connectorAcceptanceTest') {
+            dependsOn project.build
+            dependsOn project.airbyteDocker
+            
+            if (!project.hasProperty('connectorAcceptanceTestVersion')) {
+                project.ext.connectorAcceptanceTestVersion = 'latest'
```
pedroslopez, line 17: I'm missing some gradle knowledge here - sometimes I see this being referred as `project.ext.connectorAcceptanceTestVersion` and others as `project.connectorAcceptanceTestVersion`. What's the difference? Can they be used interchangeably?

I also see `CONNECTOR_ACCEPTANCE_TEST_IMAGE` being set to `latest` at the top and then checked on line 25 - should these usages be changed to refer to `project.connectorAcceptanceTestVersion` (or the `.ext` variant?)

pedroslopez, line 17: I'm missing some gradle knowledge here - sometimes I see this being referred as `project.ext.connectorAcceptanceTestVersion` and others as `project.connectorAcceptanceTestVersion`. What's the difference? Can they be used interchangeably?

I also see `CONNECTOR_ACCEPTANCE_TEST_IMAGE` being set to `latest` at the top and then checked on line 25 - should these usages be changed to refer to `project.connectorAcceptanceTestVersion` (or the `.ext` variant?)

alafanechere: @pedroslopez 

> I'm missing some gradle knowledge here - sometimes I see this being referred as project.ext.connectorAcceptanceTestVersion 

From what I understood: project properties are usually set directly in projects not it plugins (through env var, CLI input or declared in `build.gradle`). To set a project property from a plugin you've to use `project.ext.thePropertyIwantToset`. 

> I also see CONNECTOR_ACCEPTANCE_TEST_IMAGE being set to latest at the top and then checked on line 25 - should these usages be changed to refer to project.connectorAcceptanceTestVersion 

Yep sorry I forgot to remove its usage. it's done now.

buildSrc/src/main/groovy/airbyte-python.gradle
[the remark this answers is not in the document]
davinchia: @alafanechere I missed that this is only for the python modules. If that's true, I believe your comment is correct. I double checked by spot checking a couple of python connector build setups - don't seem like they directly rely on the various connector bases. 

Have you tried removing this and build 2 - 3 python connectors? That's the best sanity check.

pedroslopez (approved):
Code for connectorAcceptanceTestVersion LGTM!

Will let others comment on the removal of the "hack"

.github/workflows/test-command.yml
```diff
@@ -111,6 +114,9 @@ jobs:
           ACTION_RUN_ID: ${{github.run_id}}
           # Oracle expects this variable to be set. Although usually present, this is not set by default on Github virtual runners.
           TZ: UTC
+          ORG_GRADLE_PROJECT_connectorAcceptanceTestVersion: ${{github.event.inputs.connector-acceptance-test-version}}
+          S3_BUILD_CACHE_ACCESS_KEY_ID: ${{ secrets.SELF_RUNNER_AWS_ACCESS_KEY_ID }}
+          S3_BUILD_CACHE_SECRET_KEY: ${{ secrets.SELF_RUNNER_AWS_SECRET_ACCESS_KEY }}
```
pedroslopez, line 119: why were these added?

alafanechere: It's an addition to allow gradle to use the remote cache stored on S3. It's something already setup on other gradle CI pipeline but was missing for this one.

PR #23231, opened 2023-02-17
[source-monday] [source-square] [source-greenhouse] [source-tempo] [source-zenloop] fix a few connectors with failing custom component unit tests + release some connectors w/ beta

brianjlai:
note this will not be merged until after the freeze. This is non-urgent because it only blocks release of a new connector, not the existing one being run. If a P0 for either Monday or Square occurs, then I will merge this in to unblock.

## What
Might have had a stale virtual environment, but a few of the connectors `source-monday` and `source-square` had custom components that were still inheriting `JsonSchemaMixin` which has been deprecated during the move to low code to beta.

## How
Removed `JsonSchemaMixin` from the relevant connectors. 

Also modified the test script because `set -e` causes the script to exit early when tests fail. I wasn't able to figure out exactly why the script passed earlier this week, but my `.venv` might have retained the old `dataclasses-jsonschema` package


girarda (approved):

airbyte-integrations/connectors/source-monday/integration_tests/expected_records.jsonl
```diff
@@ -69,7 +69,8 @@
...
+{"stream": "updates", "data": {"assets": [], "body": "<p>\ufeffTest</p>", "created_at": "2022-11-21T14:41:21Z", "creator_id": "36694549", "id": "1825302913", "item_id": "3555437747", "replies": [{"id": "1825303266", "creator_id": "36694549", "created_at": "2022-11-21T14:41:29Z", "text_body": "Test test", "updated_at": "2022-11-21T14:41:29Z", "body": "<p>\ufeffTest test</p>"}], "text_body": "Test", "updated_at": "2022-11-21T14:41:29Z"}, "emitted_at": 1676683112518}
```
brianjlai, line 72: Ran some tests around this, some of the emoji codes don't render right in `text_body`. Not sure if this was a change on monday's side. And there's a third record in the `updates` stream when I tested. Retesting this passed the tests

airbyte-cdk/python/bin/low-code-unit-tests.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-set -e
+# Ideally we'd like to have set -e, but when set -e is used, the `python -m pytest unit_tests` command can return a
+# non-zero exit code and end the whole script prematurely
+# set -e
```
maxi297, line 5: Should we just remove those lines entirely if it's just commented and we don't intend to change this? I think the nature of this script makes it not needed

maxi297 (approved):
LGTM!

airbyte-cdk/python/bin/low-code-unit-tests.sh
clnoll: Normally I'm all about getting rid of commented-out code but in this case, IMO a comment explaining why we don't want to use `set -e` is kind of nice to have because my tendency would be to add it if I was updating a script where it felt like it should be used.

clnoll (approved):
LGTM!

airbyte-integrations/connectors/source-monday/source_monday/dpath_string_extractor.py
```diff
@@ -12,11 +12,10 @@
 from airbyte_cdk.sources.declarative.extractors.record_extractor import RecordExtractor
 from airbyte_cdk.sources.declarative.interpolation.interpolated_string import InterpolatedString
 from airbyte_cdk.sources.declarative.types import Config, Record
-from dataclasses_jsonschema import JsonSchemaMixin
 
 
 @dataclass
-class DpathStringExtractor(RecordExtractor, JsonSchemaMixin):
+class DpathStringExtractor(RecordExtractor):
```
artem1205, line 18: We can now remove this class.
https://docs.airbyte.com/connector-development/config-based/understanding-the-yaml-file/record-selector#selecting-fields-nested-in-arrays

artem1205 (changes requested):

artem1205 (approved):

PR #23238, opened 2023-02-19
Remove unused deps in airbyte-commons-worker

c-p-b:
## What
Removes airbyte-commons-worker files that are unused after the migration

## How
Deletes code not needed to run destination and connector acceptance tests.


colesnodgrass (approved):
Glanced over it. See a lot of what appears to be formatting changes, curious if the code was improperly formatted before or if something else is going on.

cpdeethree:
> Glanced over it. See a lot of what appears to be formatting changes, curious if the code was improperly formatted before or if something else is going on.

I'm about to push another change - only like half of these files can be actually deleted unfortunately. But everything in the new change that I push will be deleted code. I'll ask you to do another once-over once I push again

cpdeethree:
@colesnodgrass only deleting files now.

Issue #23215, opened 2023-02-17
PR `/test` results polluting nightly connector tests


evantahler:
running `/test` on a PR also seems to update https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/builds.md. This is bad, because these results power both #connector-health and the QA engine.  Only the tests run nightly via github action should be included in these results.

PR #23244, opened 2023-02-20
test-command: change report status condition

alafanechere:
Solve https://github.com/airbytehq/airbyte/issues/23215

## What
When a /test command is dispatched from a PR the `github.ref` is always `master` because the workflow runs on the master branch (even though it checkouts the value of the `gitref` inputs and runs tests on the correct branch).
The `Reports status` step which builds [this kind of page](https://dnsgjos7lj2fu.cloudfront.net/tests/summary/connectors/source-pokeapi) is therefore executed on non-main branches, which is not what we want... We want these build pages to only gather nightly builds results from `master`. 

## How
Change the conditional logic that runs `Report status`:
- use `github.event.inputs.gitref` input variable instead of `github.ref` to check if we're on the `master` branch.

### DIFF
--- a/.github/workflows/test-command.yml
+++ b/.github/workflows/test-command.yml
@@ -155,7 +155,7 @@ jobs:
         run: |
           run-qa-checks ${{ github.event.inputs.connector }}
       - name: Report Status
-        if: github.ref == 'refs/heads/master' && always()
+        if: github.event.inputs.gitref == 'master' && always()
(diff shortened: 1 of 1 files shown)


alafanechere:
I'm confirming the new syntax works as the report status step was skipped here: https://github.com/airbytehq/airbyte/actions/runs/4221899349/jobs/7329775378

evantahler (approved):
👍 nice!
Screenshot 2023-02-20 at 11 10 04 AM

PR #23248, opened 2023-02-20
Source Typeform, Paypal-transaction, Posthog: fix builds

davydov-d:
## What
https://github.com/airbytehq/airbyte-internal-issues/issues/997

## How
- Define cursor paths for all three sources in acceptance-test-config.yml
- Create dedicated config for Typeform incremental tests including a single `form_id`
- Update Typeform so it stores state as a string - just like the cursor values in records


airbyte-integrations/connectors/source-typeform/source_typeform/source.py
```diff
@@ -145,6 +145,13 @@ def get_form_id(self, record: Mapping[str, Any]) -> Optional[str]:
         referer = record.get("metadata", {}).get("referer")
         return urlparse.urlparse(referer).path.split("/")[-1] if referer else None
 
+    def current_state_value_int(self, current_stream_state: MutableMapping[str, Any], form_id: str) -> int:
+        # state used to be stored as int, now we store it as str, so need to handle both cases
+        value = current_stream_state.get(form_id, {}).get(self.cursor_field)
+        if isinstance(value, str):
+            value = pendulum.from_format(value, self.date_format).int_timestamp
+        return value or 1
```
lazebnyi, line 153: I think better add `self.start_date.int_timestamp` instead `1` and move cursor field not exist in last record handler to get method in 150 line: 

```
value = current_stream_state.get(form_id, {}).get(self.cursor_field, self.start_date.int_timestamp)
if isinstance(value, str):
   value = pendulum.from_format(value, self.date_format).int_timestamp
return value
```

lazebnyi (changes requested):

airbyte-integrations/connectors/source-typeform/source_typeform/source.py
davydov-d: :+1: done

bazarnov (approved):

lazebnyi (approved):

PR #23258, opened 2023-02-20
remove airbyte-oauth

c-p-b:
## What
removes the airbyte-oauth module

## How
remove usages in persistence layer


airbyte-oauth/build.gradle
```diff
@@ -8,7 +8,6 @@ dependencies {
     implementation project(':airbyte-config:config-persistence')
     implementation project(':airbyte-json-validation')
     implementation libs.airbyte.protocol
-    testImplementation project(':airbyte-oauth')
```
pedroslopez, line 11: Should this file be removed entirely?

pedroslopez (approved):
👋

Do you think we should run a /test for a java destination just in case? Or does the connectors-base build passing cover it?

pedroslopez (commented):
If the build passes again, :shipit: !

cpdeethree:
> 👋
> 
> Do you think we should run a /test for a java destination just in case? Or does the connectors-base build passing cover it?

I would think the destinationacceptancetest and the connectoracceptancetest suites would cover this, or am I wrong? cc @evantahler

PR #23259, opened 2023-02-20
`source-faker` v2.0.3 to test publication

evantahler:


erohmensing (approved):
Thanks for grabbing my changelog update! Looks like conor is still using that test PR to test other things so makes sense to merge it in here

PR #23262, opened 2023-02-20
remove airbyte-metrics

c-p-b:
## What
Removes the airbyte-metrics module

## How
Given that the metrics are unlikely to be used in destination acceptance tests and connectors, we remove their usage in airbyte-commons-workers


cpdeethree:
cc @cgardens

cpdeethree:
Since persistence also relies on analytics, I think that #23263 should just supersede this pr

PR #21907, opened 2023-01-26
Source Slack: high test strictness level

davydov-d:
## What
https://github.com/airbytehq/airbyte/issues/19064

## How
- Modify abnormal state
- Add configured catalogs
- Do not join channels that not gonna be synced
- Refactor acceptance test config

PR #23314, opened 2023-02-21
resolve connector dependency workflow missing changelogs

erohmensing:
## What
Resolve the following issues: 
image

Surfaced during the monorepo project in [this PR ](https://github.com/airbytehq/airbyte/pull/23238)

## How
- Add missing changelog entry from clickhouse
- Move destination-csv's docs to the correct place where we're looking for it, update reference, remove recently added one for what I assume we thought was a missing docs page

## 🚨 User Impact 🚨
CSV docs live at a different url - hopefully project refactor managed most of the references


evantahler (approved):
Thanks @erohmensing!

erohmensing:
Can try to hack it again to test it!

erohmensing:
Well, csv and clickhouse are happy 😅

erohmensing:
@evantahler some of these are probably legit - this helped notice that slack was bumped but not published in [this pr](https://github.com/airbytehq/airbyte/pull/21907). Do we go restore that branch and publish from it? 🤔

evantahler:
> @evantahler some of these are probably legit - this helped notice that slack was bumped but not published in [this pr](https://github.com/airbytehq/airbyte/pull/21907). Do we go restore that branch and publish from it? 🤔

oof. I think a new PR is probably the way to go, because /publish needs to modify the actor definition files.

erohmensing:
True. I'm going to merge this (after reverting the temporary change) as fixes to csv and clickhouse, and then ill follow up with the urgent fixes from connectors which weren't published. Other things surfaced here seem less urgent and I have less context on (like a destination not being in the seed)

PR #23315, opened 2023-02-21
Include debug yaml in download and fix colors

evantahler:
1. `docker-compose.debug.yaml` is important for running airbyte.  We should download this file so we can actually access the DB and see temporal's UI (when needed) - `docker-compose -f docker-compose.yaml -f docker-compose.debug.yaml up`
2. Fix the colors in the console with `echo -e`

### DIFF
--- a/run-ab-platform.sh
+++ b/run-ab-platform.sh
@@ -19,27 +19,28 @@ PS4="$blue_text""${BASH_SOURCE}:${LINENO}: ""$default_text"
 Help()
 {
    # Display Help
-   echo "This Script will download the necessary files for running docker compose"
-   echo "It will also run docker compose up"
-   echo "Take Warning! These assets may become stale over time!"
+   echo -e "This Script will download the necessary files for running docker compose"
+   echo -e "It will also run docker compose up"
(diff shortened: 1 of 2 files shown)


run-ab-platform.sh
```diff
@@ -19,27 +19,28 @@ PS4="$blue_text""${BASH_SOURCE}:${LINENO}: ""$default_text"
...
+   echo -e ""
 }
 
 ########## Declare assets care about ##########
-docker_compose_yaml="docker-compose.yaml"
-            dot_env=".env"
-        dot_env_dev=".env.dev"
-              flags="flags.yml"
+      docker_compose_yaml="docker-compose.yaml"
+docker_compose_debug_yaml="docker-compose.debug.yaml"
```
evantahler, line 38: This is the main change.  everything else is just `echo -e`

cpdeethree: can you also add this to .gitignore?

cpdeethree: Disregard above, i see you did already

c-p-b (approved):

PR #23318, opened 2023-02-21
Publish connectors whose dockerfile versions were already updated 

erohmensing:
## What
The following connectors have dockerfile versions higher than on dockerhub because they weren't published, although their code was merged to master: 
* source-recharge ([PR](https://github.com/airbytehq/airbyte/pull/22473))
* source-slack ([PR](https://github.com/airbytehq/airbyte/pull/21907))

The following were published but the *_definitions.yml file wasn't generated automatically and wasn't manually updated accordingly: 
* source-tiktok-marketing ([publish comment](https://github.com/airbytehq/airbyte/pull/22309#issuecomment-1413852910))
* destination-databend ([PR](https://github.com/airbytehq/airbyte/pull/21182) - looks like they were generated, but a later force push removed them)

## How to fix
Publish the first 2 in this PR, updating the changelogs accordingly. Only changes should be changelog dates and commits made automatically via the publish workflow. 

After the first 2 are published, manually update source_definitions.yml and destination_definitons.yml to reflect the correct latest versions for databend and tiktok marketing.

### DIFF
--- a/airbyte-config/init/src/main/resources/seed/destination_definitions.yaml
+++ b/airbyte-config/init/src/main/resources/seed/destination_definitions.yaml
@@ -426,7 +426,7 @@
 - name: Databend
   destinationDefinitionId: 302e4d8e-08d3-4098-acd4-ac67ca365b88
   dockerRepository: airbyte/destination-databend
-  dockerImageTag: 0.1.0
+  dockerImageTag: 0.1.1
   icon: databend.svg
   documentationUrl: https://docs.airbyte.com/integrations/destinations/databend
   releaseStage: alpha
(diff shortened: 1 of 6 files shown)


evantahler (approved):
/publish away!

I don't really know what to do if the tests fail though...

erohmensing:
Ugh, yeah. Probably the same as last time - check if the failure is expected as of the latest #connector-health state and if so, merge without tests

erohmensing:
Both failures are the same as the ones currently on master: 
* [recharge](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1677071327556449)
* [slack](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1677071327759699) 

With no changes here, it's using the code already on master. Both test suites were passing when the code was merged, and tests have passed in nightly runs previously after they were merged, so I don't think the fact that they are failing now is related to those code changes.

erohmensing:
/publish connector=connectors/source-recharge run-tests=false

erohmensing:
I don't see any actors in the cloud DB that use destination-databend yet, so I think the deletion in the spec is safe - otherwise that would be a change we could probably have to enforce users who had selected a different protocol to use.

PR #23330, opened 2023-02-22
fix: Permission check failing when privilege is at ROLE level

sergio-ropero:
From PR: https://github.com/airbytehq/airbyte/pull/23138

# What
The replication can be only done when the user has `SUPERUSER` or `REPLICATION` privileges. 
```
psql: error: FATAL:  must be superuser or replication role to start walsender
```
In the documentation we assumed that using a `role` with any of the above privileges attached to a user, grants these attributes to the user, [but it's not right](https://www.postgresql.org/docs/current/role-attributes.html#:~:text=initiating%20replication,REPLICATION%20LOGIN).

# Solution
In order to fix the check function we will try directly to connect to the database doing a replication connection and verify that it's correct.

Added tests for all cases:
- Access with SUPERUSER privilege
- Access with REPLICATION privilege
- NO access (no privileges)

# FROM HERE IS NOT RELEVANT ANYMORE! (But can be interesting to keep)
This commit fixes the issue when permission is granted at ROLE level instead of USER level.

Missing revoke privileges in the tests.

### Set up
The user doesn't have replication privilege at user level, neither for the role attached (`secondary_role`). But the role `secondary_role` is attached to `airbyte_replication` role, so at the end the user has access to replication thanks to last link.
```
CREATE USER airbyte_ro PASSWORD 'XXXXX';
CREATE ROLE airbyte_replication REPLICATION LOGIN;
CREATE ROLE secondary_role LOGIN;
GRANT secondary_role TO airbyte_ro;
GRANT secondary_role TO airbyte_replication;
```
### User Table
We see it doesn't have replication permission:
```
select usesysid, usename, userepl from pg_user where usename = 'airbyte_ro';
```
Screenshot 2023-02-16 at 16 54 56

## Assigned roles
As well without replication permission.
```
select pgr.rolname, rolreplication  from pg_catalog.pg_roles pgr
join pg_catalog.pg_auth_members pgam on pgam.roleid = pgr.oid 
join pg_catalog.pg_user pgu on pgam.member = pgu.usesysid 
where usename ='airbyte_ro'; 
```
Screenshot 2023-02-16 at 16 56 19

### Query to check permission (by any subrole).
If it return at least one record, the user has privilege to Postgres replication:
```
WITH RECURSIVE cte AS (
   SELECT usesysid  FROM pg_catalog.pg_user WHERE usename = 'airbyte_ro'
   UNION ALL
   SELECT m.roleid
   FROM   cte
   JOIN   pg_auth_members m ON m.member = cte.usesysid
   )
SELECT rolname FROM cte
JOIN pg_roles ON usesysid = oid
WHERE rolreplication;
```
Screenshot 2023-02-16 at 17 01 36


sergio-ropero:
/publish connector=connectors/source-postgre-strict-encrypt auto-bump-version=false

sergio-ropero:

airbyte-integrations/bases/debezium-v1-9-6/src/main/java/io/airbyte/integrations/debezium/internals/PostgresReplicationConnection.java
```diff
@@ -0,0 +1,84 @@
...
+    properties.setProperty("password", jdbcConfig.has(JdbcUtils.PASSWORD_KEY) ? jdbcConfig.get(JdbcUtils.PASSWORD_KEY).asText()
+        : null);
+    properties.setProperty("assumeMinServerVersion", "9.4");
+    properties.setProperty("ApplicationName", "Airbyte Debezium Streaming");
+    properties.setProperty("replication", "database");
+    properties.setProperty("preferQueryMode", "simple"); // replication protocol only supports simple query mode
+
+    LOGGER.info("Creating a CDC connection.");
+    final String jdbcUrl = jdbcConfig.has(JdbcUtils.JDBC_URL_KEY) ? jdbcConfig.get(JdbcUtils.JDBC_URL_KEY).asText()
+        : String.format(DatabaseDriver.POSTGRESQL.getUrlFormatString(),
```
subodh1810, line 42: I found a bug out here if a connection only accepts SSL connection then we need to add the attributes related to SSL connection while constructing the URL format. Here we are not adding em. It will cause the check to fail if a server only accepts SSL connection. I tested this with a cloud sql instance which only accepts SSL connections

subodh1810 (changes requested):
PR has a bug

subodh1810:
I and @sergio-ropero discussed the issue on call and we deleted the published images for alloy db. Postgres image was not published.

subodh1810 (approved):
We tested this against a cloud sql instance with SSL parameters and this works. Lets just make sure that the build is green and tests pass before publishing/merging

sergio-ropero:
> Connector	Did it publish?	Were definitions generated?
> connectors/source-postgres	✅	❌

Manually done following guidelines

PR #23321, opened 2023-02-22
Fix `destination-exasol` build and CatalogDefinition generation

evantahler:
Fixes the `destination-exasol` build (added via https://github.com/airbytehq/airbyte/pull/21200).  Lots of stuff changed when we ripped apart our monorepo 😅 

Tested locally via 

```
./gradlew :airbyte-integrations:connectors:destination-exasol:test :airbyte-integrations:connectors:destination-exasol:integrationTest
```

PR #23332, opened 2023-02-22
python-connectors: remove explicit dependencies to connector-acceptance-test build

alafanechere:
## What
The improvement expected by https://github.com/airbytehq/airbyte/pull/23208 is not benefitting a lot of connectors because some explicitly declare a dependency to `connector-acceptance-test` build in their `build.gradle` file.

It means these connectors' build depends on `connector-acceptance-test:airbyteDocker`, which depends on `connector-acceptance-test:assemble` (which is testing connector-acceptance-test, and this is long...).

I don't think we should explicitly declare dependencies of connectors to `connector-acceptance-test:airbyteDocker`. 
https://github.com/airbytehq/airbyte/pull/23208 introduced on optional dependency to this task if a `connectorAcceptanceTestVersion` property is added to the project and if this property == "dev".

I don't have the history of why some connector's build.gradle file declared this dependency and other don't (e.g source-pokeapi don't).  I believe this dependency was once expressed in the `build.gradle.hbs` file used to generate connectors.

## How
Remove explicit dependency on `connector-acceptance-test:airbyteDocker` in `build.gradle` file of  63 connectors.


sherifnada (approved):

evantahler (approved):
Nice discovery!

if you wait until https://github.com/airbytehq/airbyte/pull/23321 is merged to master and pull that in, the failing test should be fixed

Issue #15729, opened 2022-08-17
Source Hubspot :  Handle 502 Bad Gateway Error better (Better Error message for customer)


YowanR:
## Current Behavior
When the Hubspot connectors gets a 502 bad gateway from the API, and it takes the full HTML it was seeing and exposed it as an error message. 

## Expected Behavior
We should parse through the error and display something more useful for the user.

## Additional Infomation
See image below for current error message:
![image.png](https://images.zenhubusercontent.com/62c49418a8561882ea2fd091/b34b6d42-44e6-4621-a5a2-198d7f1b6c41)

The UI portion of this bug was fixed in [#15633](https://github.com/airbytehq/airbyte/pull/15633)

bazarnov:
@YowanR do we have the link to this particular workspace? I need to check for retries to take place here.

YowanR:
Let's prioritize this in the next sprint. For now, let's focus our efforts on Q1a roadmap items, please! 🙏

bazarnov:
@lazebnyi cc FYI

davydov-d:
As Maxime [mentioned](https://github.com/airbytehq/airbyte/pull/23337#pullrequestreview-1309755721), we can not limit the length of the error message because the error should remain clear while working with the CLI. Parsing the HTML is also not worth the effort, besides it is rather an edge case.

PR #23337, opened 2023-02-22
Source Hubspot: limit error message length

davydov-d:
## What
#15729 

## How
Cut the error message: display ~100 symbols - a half from the start of the original error message and a half from the end


davydov-d:
Blocked by the expected records failing test :(

airbyte-integrations/connectors/source-hubspot/source_hubspot/streams.py
```diff
@@ -379,11 +379,17 @@ def read_records(
...
+    def error_message(self, response: requests.Response) -> str:
+        return self.parse_response_error_message(response)
+
     def parse_response_error_message(self, response: requests.Response) -> Optional[str]:
         try:
             body = response.json()
         except json.decoder.JSONDecodeError:
-            return response.text
+            message = response.text
+            if len(message) > 100:
```
maxi297, line 390: Where does this 100 characters limitation comes from? If it's a UI issue, why does the source needs to know about it?

maxi297 (commented):
I'm wary as if the fix should be at that place. This fixes source-hubspot but what about all the other sources? For me, this seems like a UI issue and not a CDK issue. If a user using command line to test the source wants to have the full message, it does not make much sense to truncate the message because the UI doesn't handle this very well

airbyte-integrations/connectors/source-hubspot/source_hubspot/streams.py
davydov-d: @maxi297 that's fair. The thing is the issue has a linked PR that fixes it on the UI and the issue is still not closed, so I thought we need to handle it here as well. Do you think we can just close the issue?

maxi297: If this change should be done in the UI because it is just UI related, yes I would close this one and push for the PR on the UI side to be closed

PR #23339, opened 2023-02-22
Update the airbyte quickstart to use `./run-ab-platform.sh`

evantahler:
After the monorepo change, the instructions to run airbuyte have changed. 

As a note, this means that Airbyte no longer runs on windows.  We will need to make a .bat script as well... (https://github.com/airbytehq/monorepo/issues/33)


Amruta-Ranade (approved):
Thanks for the fix!

PR #23365, opened 2023-02-22
Remove featureflag and notification modules

c-p-b:
## What
These are used in platform only for now


colesnodgrass (approved):
I imagine the feature-flag client will likely be extracted from the platform and deployed as a shared library soon.  But as it's not used currently (and will have some modifications on the platform side soon) 🔪

cgardens (approved):

PR #23370, opened 2023-02-23
Changes to airbyte-config should trigger base build

evantahler:
... and also remove the backend filter we no longer have


.github/workflows/gradle.yml
```diff
@@ -59,18 +58,14 @@ jobs:
         with:
           # Note, the following glob expression within a filters are ORs.
           filters: |
-            backend:
```
evantahler, line 62: No other action was consuming the "backend" variable

c-p-b (approved):

.github/workflows/gradle.yml
```diff
@@ -59,18 +58,14 @@ jobs:
...
-              - 'docker-compose*.yaml'
-              - '(charts|kube)/**'
             build:
               - '.github/**'
               - 'buildSrc/**'
               - 'tools/**'
               - '*.gradle'
               - 'deps.toml'
+              - 'airbyte-config/**'
+              - 'airbyte-featureflag/**'
```
pedroslopez, line 68: I think this one is going away soon

pedroslopez (approved):

.github/workflows/gradle.yml
cpdeethree: I already deleted featureflag yesterday

config, well, I think that story is going to be a longer conversation. It's, um, quite intertwined with both repositories right now

PR #23386, opened 2023-02-23
fix: Exclude file `connectors.md` from git diff as it's autogenerated and it should change

sergio-ropero:
## What
This [PR](https://github.com/airbytehq/airbyte/pull/23367) introduces a file that changes on every run but there is a check in the github action that compare no files has changed.
That prevent connectors changes to be released

## How
Exclude `connectors.md` file from the `git diff` command.

FYI: @evantahler


evantahler:
https://github.com/airbytehq/airbyte/pull/23372 should fix this

sergio-ropero:
> #23372 should fix this

Fair enough hahaha
Although was handy to know when the page was updated.
Closing the PR.

PR #23392, opened 2023-02-23
remove generated date from connectors.md

pedroslopez:
## What

We removed this from the generator in  https://github.com/airbytehq/airbyte/pull/23372 but forgot to also remove it from the generated file.


davydov-d (approved):

evantahler (commented):
Thank you!

PR #23399, opened 2023-02-23
Delete airbyte-cdk/python/docs directory

sherifnada:
## What
CDK docs now live in the top-level `docs/` directory. This is deleting the duplicated `airbyte-cdk/python/docs` directory.


maxi297 (approved):
LGTM

PR #23403, opened 2023-02-23
rebuild master connectors.md

evantahler:
Run `./gradlew :airbyte-config:init:processResources`.  https://github.com/airbytehq/airbyte/pull/23379 didn't run the flake test or process resources.  This was added later via https://github.com/airbytehq/airbyte/pull/23370.  I don't think this will be an issue going forward...


pedroslopez (approved):
Modifying the definitions files like what happened in that branch will not trigger the connectors base build, so we won't run the re-generation and validation.

Perhaps rather than running this on build we can separate this out into its own task and via GitHub actions trigger it whenever the yaml files are touched + autocommit to avoid this kind of thing and not have people worry about it.

PR #23408, opened 2023-02-23
Updated links in docs from .io to .com

sophia-wiley:
Replaced [cloud.airbyte.io](http://cloud.airbyte.io/) with [cloud.airbyte.com](http://cloud.airbyte.com/), [docs.airbyte.io](http://docs.airbyte.io/) with [docs.airbyte.com](http://docs.airbyte.com/), and https://airbyte.io with https://airbyte.com. I left https://internal-docs.airbyte.io as is because there is not a redirect for it yet. 

I didn't change the https://airbyte.io links in the expected_records.jsonl files because I wasn't sure what that would do.


airbyte-api/src/main/openapi/config.yaml
```diff
@@ -24,7 +24,7 @@ info:
     url: "https://opensource.org/licenses/MIT"
 externalDocs:
   description: Find out more about Airbyte
-  url: "https://airbyte.io"
```
cpdeethree, line 27: FYI, this updates the swagger url. The only place this is used in connectors repo is to feed Octavia CLI. So now Octavia CLI will use updated url. just wanted to point out the app implication here, though it is a change we want.

sophia-wiley: Okay I'll remove this

sophia-wiley:
Closing this PR in favor of [23652](https://github.com/airbytehq/airbyte/pull/23652)

PR #23411, opened 2023-02-23
Isolate our build so that the feedback loop is faster

maxi297:
## What
We want the extensibility team to have their own checks and be impacted as less as possible with the builds of other.

The build is now 4.5 times faster! This increases the feedback loop and reduce the risk of occurrence of https://github.com/airbytehq/airbyte-internal-issues/issues/1345
image

## How
By creating our own check and moving airbyte-cdk changes under this check, we will:
* trigger "Connectors Base: Build" check less often
* be the master of our own build

## Recommended reading order
1. `settings.gradle`: Gradle build changes
2. `.github/workflows/gradle.yml`: GitHub checks changes


.github/workflows/gradle.yml
```diff
@@ -150,6 +152,65 @@ jobs:
...
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: "zulu"
+          java-version: "17"
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.9"
+
```
maxi297, line 179: "Connectors Base: Build" was adding "Install Pyenv" (introduced [here](https://github.com/airbytehq/airbyte/commit/683a67c023d316cdf4a2212ba6146529e3a42d25)) but it's not so clear to me why we would need this

.github/workflows/gradle.yml
```diff
@@ -150,6 +152,65 @@ jobs:
...
+      - name: Build
+        uses: Wandalen/wretry.action@master
+        with:
+          command: SUB_BUILD=CDK ./gradlew build --scan
+          attempt_limit: 3
+          attempt_delay: 5000 # in ms
+
+      - name: Ensure no file change from code formatting
+        run: git --no-pager diff && test -z "$(git --no-pager diff)"
+
```
maxi297, line 213: At that point, "Connectors Base: Build" would generate test report (to GCS). Who uses those test reports and do we need to do it here as well? If we don't know, let's just not add this for now and wait for someone to scream

settings.gradle
```diff
@@ -71,28 +71,34 @@ if (!System.getenv().containsKey("SUB_BUILD")) {
...
-include ':airbyte-json-validation'
-include ':airbyte-test-utils'
-
-// airbyte-workers has a lot of dependencies.
-include ':airbyte-commons-worker'
-include ':airbyte-config:config-persistence' // transitively used by airbyte-workers.
-include ':airbyte-db:jooq' // transitively used by airbyte-workers.
-include ':airbyte-persistence:job-persistence' // transitively used by airbyte-workers.
+if (!System.getenv().containsKey("SUB_BUILD") || System.getenv().get("SUB_BUILD") == "CDK" || System.getenv().get("SUB_BUILD") == "ALL_CONNECTORS") {
+    include ':airbyte-commons' // this wouldn't be necessary if it wasn't from https://github.com/airbytehq/airbyte/blob/645558b74aab0b91fda1b4628b37b7095d92b4cc/build.gradle
```
maxi297, line 80: This seems very odd but I didn't wanted to investigate this since the build is still somewhat fast compared to before. Note that this seems like another improvement we could do though

maxi297 (commented):
It seems like a partial win since we don't often do changes in isolation and this will probably trigger the build for "Connectors Base: Build" when we update some sources. We will need to find a way to avoid running "Connectors Base: Build" when modifying a source that is relying on the airbyte-cdk

.github/workflows/gradle.yml
sherifnada: I think these test reports are not used by anything in the CDK. So it feels safe to ignore here.

sherifnada (changes requested):
Nice! Small surgical change. Two small requests; 
1. Could you also confirm you looked at the docs anywhere it says `SUB_BUILD=CONNECTORS_BASE` to make sure that doc is still current? 
2. Also, the PR description should mention the gains in build speed 😎

maxi297:
Also added similar logic to https://github.com/airbytehq/airbyte/pull/23415

sherifnada (approved):

PR #23422, opened 2023-02-24
Remove workers

cgardens:
## What
* Connectors code depends on workers code for running connectors in acceptance tests.
* This PR renames them to be test harnesses to avoid confusion in the future.
* Also prunes a lot of unneeded code (including all jobs and configs db migrations, envconfigs, airbyte-notifications, and worker code)


airbyte-db/db-lib/build.gradle
```diff
@@ -58,53 +58,17 @@ dependencies {
...
-    args 'jobs', 'migrate'
-}
-
-task(dumpJobsSchema, dependsOn: 'classes', type: JavaExec) {
-    mainClass = 'io.airbyte.db.instance.development.MigrationDevCenter'
-    classpath = files(configurations.migrations.files)
-    args 'jobs', 'dump_schema'
-}
-
-task copyInitSql(type: Copy) {
```
edgao, line 97: I think this + the `tasks.named("buildDockerImage")` block are still needed. Or at least, `:airbyte-db:db-lib:buildDockerImage` is failing because `/actions-runner/_work/airbyte/airbyte/airbyte-db/db-lib/build/docker` doesn't exist

edgao (commented):
lgtm assuming connector tests pass + build succeeds; I didn't really look through the full file list though (just checked for the non-deletion of a few randomly selected files). One comment about the db-lib build.gradle diff to maybe get the build back to green.

airbyte-db/db-lib/build.gradle
cgardens: thanks @edgao . do you have any idea what this docker container is used? i was having trouble finding usages.

cgardens: oh nvm. i think it is exclusively used in the platform. so probably not needed here.

edgao: ah nice, that makes sense. had thought this goes into one of the connectors base images but that wouldn't make sense (and is verifiably false anyway)

I think you also need to delete this line https://github.com/airbytehq/airbyte/blob/master/airbyte-db/db-lib/gradle.properties#L1 to remove the buildDockerImage task as well

cgardens: thanks!

cgardens:
@rodireich the postgres source integration test failed because it got an unexpected spec. i don't _think_ this is related to th changes in this PR, but never say never. do you have any context into this failure?

rodireich:
@cgardens This is the failure
Screenshot 2023-02-24 at 10 23 52 AM

rodireich:
Could you try updating your branch. 
Testing the failing PostgresSourceStrictEncryptAcceptanceTest.testGetSpec() is passing on master for me

rodireich:
`/test` seems to be [passing](https://github.com/airbytehq/airbyte/pull/23447#issuecomment-1444226725) on current master. 
It's either one of the [8 commits](https://github.com/airbytehq/airbyte/compare/cgardens/remove-workers...master) you branch is behind, or the actual code the change (though I see no reason for that).

cgardens:
* Merged this because I'm worried about bit rot. There's a lot of dangerous code left over from the split, that if people start using it will create a huge amount of tech debt, but before the PR it looked safe to use.
* It has had eyes on it by sources, destinations, and prod eng. 
* In terms of production--risk is low. 
* There is some possibility that I have removed a compile time dependency for a java source or destination. If that is the case, we will know. 
* We know source postgres and destination snowflake are good to go

PR #23434, opened 2023-02-24
Run CATs with local CDK in CI

clnoll:
## What
Updates the /test command to take an optional `local_cdk` argument that, when set, will test the connector against changes made to the CDK on the current branch.

Usage:
`/test connector=<CONNECTOR NAME> local_cdk=1`

## How
- Modifies the /test config file (test-command.yml) to accept the `local_cdk` argument, which it passes to the `tools/bin/ci_integration_test.sh` script when running connector acceptance tests.
- Modifies `ci_integration_test.sh` so that the connectors' integration tests are invoked with the `connectorAcceptanceTest.useLocalCdk` and `connectorAcceptanceTest.connectorName` gradle properties.
- The `airbyteDocker` plugin builds the docker image that's run during connector acceptance tests, so we also modify
the plugin to call the `build-connector-image-with-local-cdk.sh` script if the `connectorAcceptanceTest.useLocalCdk` property is set. 

## Testing
This PR updates to the /test command, and so needs to be tested by triggering the workflow manually so it will run against changes on this branch.

When the workflow is triggered with the `local_cdk` variable set, we can see from [CI output](https://github.com/airbytehq/airbyte/actions/runs/4285089531/jobs/7462965563) that the `build-connector-image-with-local-cdk.sh` script is run.
image

When the workflow is triggered against a version of the CDK with an exception, [the exception is printed to the CI output](https://github.com/airbytehq/airbyte/actions/runs/4284671344/jobs/7461988929).
image

## Pre-merge Checklist
- [x] Documentation updated

### DIFF
--- a/buildSrc/src/main/groovy/airbyte-docker.gradle
+++ b/buildSrc/src/main/groovy/airbyte-docker.gradle
@@ -56,10 +56,30 @@ abstract class AirbyteDockerTask extends DefaultTask {
         }
     }
 
+    def buildDockerfileWithLocalCdk(String scriptPath, String fileName) {
+        if (project.file(fileName).exists()) {
+            def tag = DockerHelpers.getDevTaggedImage(projectDir, dockerfileName)
+            project.exec {
+                environment "CONNECTOR_TAG", tag
+                environment "CONNECTOR_NAME", project.findProperty('connectorAcceptanceTest.connectorName')
+                commandLine scriptPath
+            }
+        }
+    }
+
     @TaskAction
     def dockerTask() {
-        def scriptPath = Paths.get(rootDir.absolutePath, 'tools/bin/build_image.sh').toString()
-        buildDockerfile(scriptPath, dockerfileName)
+        if (
+            project.hasProperty('connectorAcceptanceTest.useLocalCdk') &&
+            project.properties["connectorAcceptanceTest.useLocalCdk"]
+        ) {
+            def scriptPath = Paths.get(rootDir.absolutePath, 'airbyte-integrations/scripts/build-connector-image-with-local-cdk.sh').toString()
+            buildDockerfileWithLocalCdk(scriptPath, dockerfileName)
+        }
+        else {
+            def scriptPath = Paths.get(rootDir.absolutePath, 'tools/bin/build_image.sh').toString()
+            buildDockerfile(scriptPath, dockerfileName)
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


clnoll:
@sherifnada I know we'd talked about creating a separate plugin instead of modifying `airbyteDocker`, but in the end this felt like a natural place for it, and like it would have the most minimal changes. LMK what you think.

evantahler (approved):
Nice!

buildSrc/src/main/groovy/airbyte-docker.gradle
```diff
@@ -56,10 +56,30 @@ abstract class AirbyteDockerTask extends DefaultTask {
...
+            }
+        }
+    }
+
     @TaskAction
     def dockerTask() {
-        def scriptPath = Paths.get(rootDir.absolutePath, 'tools/bin/build_image.sh').toString()
-        buildDockerfile(scriptPath, dockerfileName)
+        if (
+            project.hasProperty('connectorAcceptanceTest.useLocalCdk') &&
```
sherifnada, line 73: why have these two separate checks? is the second just checking if the value is truth-y?

sherifnada (commented):
It's a bit smelly that we have two different scripts for running essentially the same docker command but with one logic fork (whether to use the local CDK or not). Does it make sense to consolidate the two?

buildSrc/src/main/groovy/airbyte-docker.gradle
clnoll: Yes the second checks to see if the value is truthy.

I've tried using `getProperty` here instead of the combination of `hasProperty` + indexing but get an exception in the [Octavia Cli: Build step](https://github.com/airbytehq/airbyte/actions/runs/4292402149/jobs/7478793758): 
```
Execution failed for task ':octavia-cli:airbyteDocker'.
> Could not get unknown property 'connectorAcceptanceTest.useLocalCdk' for project ':octavia-cli' of type org.gradle.api.Project.
```

clnoll:
@sherifnada the two scripts are actually quite different. They both ultimately call `docker build`, but `build-connector-image-with-local-CDK.sh` has a much more specific purpose and should not be called outside of a context where we know that the image is being built for a connector that uses the CDK. I personally think it would pollute `build-image.sh` to try to incorporate that logic but can see what it looks like if you're interested.

sherifnada (approved):

PR #18782, opened 2022-11-01
Initial commit

akashkulk:
Test

PR #23458, opened 2023-02-25
QA Checks for strict-encrypt connectors should use regular variants

evantahler:
When running QA Checks for strict-encrypt source, we should be checking the metadata of the original variant.  Solves problems like [this](https://github.com/airbytehq/airbyte/actions/runs/4266720151/jobs/7427589422) from https://github.com/airbytehq/airbyte/pull/18782

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/qa_checks.py
+++ b/tools/ci_connector_ops/ci_connector_ops/qa_checks.py
@@ -198,7 +198,7 @@ def run_qa_checks():
     if not connector_technical_name.startswith("source-") and not connector_technical_name.startswith("destination-"):
         print("No QA check to run as this is not a connector.")
         sys.exit(0)
-    if connector_technical_name.__contains__("-strict-encrypt"):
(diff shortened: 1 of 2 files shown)


akashkulk (approved):

bnchrch (approved):
Seems like a fair enough change to me.

@evantahler I'm sure theres a good reason but do you know why are these separate images and not perhaps a separate env variable in the connector?

evantahler:
> @evantahler I'm sure theres a good reason but do you know why are these separate images and not perhaps a separate env variable in the connector?

There's not a good reason, and that's totally what we should do.  One day we will tackle the `Spec Mutations + Remove Strict Encrypt connector variants` project ([ideas](https://docs.google.com/document/d/1t6msniiEgllqmeJtvWXZXBvtEetxFO8S2zJOBiBqTtQ/edit))

alafanechere (approved):
Nice fix, thank you! Please bump the ci_connector_ops version in setup.py :)

PR #23462, opened 2023-02-25
📚 Documentation Typo Fix in Developing On Docker

chandrasekharan98:
## What
There seems to be a [typo](https://docs.airbyte.com/contributing-to-airbyte/developing-on-docker#:~:text=version%20of%20an-,exiting%20module,-.%20First%20your%20docker) in the **New module** section of developing-on-docker.md.

The previous section talks about **Existing modules** which makes more sense that `exiting modules` is a typo for `existing modules`

image


marcosmarxm (approved):
Thanks @chandrasekharan98

PR #23468, opened 2023-02-27
Source Faker: export json files

sherifnada:
## What
This bundles all the JSONs in the `schemas/` directory with the wheel/pip package which gets built when someone runs `pip install` against this package. 

The motivation is that I'm working on a project which directly imports connectors such as this one from another python package e.g `from source_faker import SourceFaker`. But `read` and `discover` don't run properly since this these JSON files are needed for the connector to work but are not being packaged with the pip package produced by this connector. 

## Notes
This is not meaningful enough to require publishing a new connector version so I won't do it. But I'll run `/test` anyways

### DIFF
--- a/airbyte-integrations/connectors/source-faker/setup.py
+++ b/airbyte-integrations/connectors/source-faker/setup.py
@@ -19,7 +19,7 @@ setup(
     author_email="evan@airbyte.io",
     packages=find_packages(),
     install_requires=MAIN_REQUIREMENTS,
-    package_data={"": ["*.json", "schemas/*.json"]},
+    package_data={"": ["*.json", "schemas/*.json", "record_data/*.json"]},
     extras_require={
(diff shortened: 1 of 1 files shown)


airbyte-integrations/connectors/source-faker/setup.py
```diff
@@ -19,7 +19,7 @@
     author_email="evan@airbyte.io",
     packages=find_packages(),
     install_requires=MAIN_REQUIREMENTS,
-    package_data={"": ["*.json"]},
+    package_data={"": ["*.json", "schemas/*.json", "record_data/*.json"]},
```
evantahler, line 22: For all connectors which might one day need to do this... is there a globpattern we should add to bundle all data files like `["**/*.json", "**/*.yaml"]`?

evantahler (approved):
Sweet.

PR #23487, opened 2023-02-27
ci_connector_ops: POC of CI connector pipelines in python

### LINKED ISSUE #23504: ci_connector_ops: POC rewriting airbyte-python.gradle and airbyte-connector-acceptance.gradle in Python
I believe we could take back the ownership of the connector test pipeline by rewriting them with Python code. 
I'd like to tackle this by rewriting [airbyte-python.gradle](https://github.com/airbytehq/airbyte/blob/d4a225fa9f921e8697e3423571b9dc99b14f65d5/buildSrc/src/main/groovy/airbyte-python.gradle) and [airbyte-connector-acceptance-test.gradle](https://github.com/airbytehq/airbyte/blob/master/buildSrc/src/main/groovy/airbyte-connector-acceptance-test.gradle#L1) in Python.

Declaring test pipelines in Python has the following advantages:

It’s easily readable and understandable by the Airbyte engineering team.
Enhances the code discoverability and debugging experience: we’ll no longer rely on a combination of shell scripts, GitHub workflows, Gradle/Python code to get the full picture of our test pipeline. We’ll have a centralized Python project to look at. 
The workflow can be unit tested.
We can benefit from abstractions to declare reusable pipeline components and grow them with scale and flexibility. 
The test pipelines can be executed locally in the same way they’ll be run in the CI.
A significant portion of the existing test tooling is already written in Python: ci-credentials to manage connector test secrets, ci-connector-ops which bundle multiple checks (QA, allowed host, test strictness level etc.), connector acceptance tests. It means we could directly load and reuse these packages.
We can use Dagger’s SDK to benefit from free incremental builds and isolation by design and continue to consider containers Airbyte’s first class citizens. 

We’ll need to pay extra care at building reusable components (a.k.a actions) to populate the workflow. This is the key to have a lean test pipeline and make sure this project can grow with new use cases in the future.

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/23504
This is an attempt at rewriting [airbyte-python.gradle](https://github.com/airbytehq/airbyte/blob/7d7e48b2a342a328fa74c6fd11a9268e1dcdcd64/buildSrc/src/main/groovy/airbyte-python.gradle) and [airbyte-connector-acceptance-test.gradle](https://github.com/airbytehq/airbyte/blob/master/buildSrc/src/main/groovy/airbyte-connector-acceptance-test.gradle) in Python (with the use of Dagger).
The motivation behind this work can be found [here](https://docs.google.com/document/d/16aXyIR1hbN_DUYwcnaZ2mnXlwfI7C0yv2vbduhLrJZI/edit#)

I apologize for the size of the PR but I'd like to use it as a common playground for Dagger team and us to iterate on best practices and optimization they could suggest.

## How
* Declare a new `pipelines` subpackage in `ci_connector_ops`
* Expose a `connectors-ci` command that can run CI pipelines on one our multiple connectors and according to code change:
```bash
connectors-ci test-connectors --name=source-pokeapi --name=source-openweather
connectors-ci test-connectors --modified # Test only connectors modified in the current branch
```
## Recommended reading order
1. [The README !!!](https://github.com/airbytehq/airbyte/blob/augustin/poc/ci-in-python/tools/ci_connector_ops/ci_connector_ops/pipelines/README.md)
2. [The entrypoints with DAG logic](https://github.com/airbytehq/airbyte/blob/augustin/poc/ci-in-python/tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py)
3. [A context object to customize execution and hold pipeline configuration](https://github.com/airbytehq/airbyte/blob/augustin/poc/ci-in-python/tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py)
4. [The actual tests we run](https://github.com/airbytehq/airbyte/blob/augustin/poc/ci-in-python/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/tests.py)
5. [The reusable environments](https://github.com/airbytehq/airbyte/blob/augustin/poc/ci-in-python/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py)

All the other change (commits before [0431cf6](https://github.com/airbytehq/airbyte/pull/23487/commits/0431cf642f7e3c9482732ce8e3a4f5e83d617355)) are made to improve/patch our existing tooling to work in a CI context with Dagger.

 ### The target DAG
```mermaid
flowchart LR;
    AB_GIT_REPO[Airbyte Git Repo] --> MOUNT_AB[Mount Airbyte repo to container];
    AB_GIT_REPO[Airbyte Git Repo] --> MOUNT_CONN[Mount connector source code to container];
    DOWN_SECRETS[Download secrets from GSM]-->CAT
    MOUNT_AB-->QA[Run QA checks];
    MOUNT_CONN-->FORMAT[Code format checks];
    MOUNT_CONN-->INSTALL_CONN[Install connector package in container];
    INSTALL_CONN-->UNIT_TESTS[Run unit tests];
    UNIT_TESTS-->INTEGRATION_TESTS[Run integration tests];
    UNIT_TESTS-->DOCKER_BUILD[Docker build connector dev image];
    DOCKER_BUILD-->CAT[Run acceptance tests];
    CAT-->UPLOAD_SECRETS[Upload updated secrets to GSM];
    CAT-->REPORT[Build test report];
    UNIT_TESTS-->REPORT;
    INTEGRATION_TESTS-->REPORT;
    QA-->REPORT;
    FORMAT-->REPORT;
    REPORT--if in CI-->UPLOAD[Upload to S3];
```

## 🚨 User Impact 🚨
Before assessing a user impact I'd like to first get my question to Dagger teams answered:
https://github.com/airbytehq/airbyte/blob/augustin/poc/ci-in-python/tools/ci_connector_ops/ci_connector_ops/pipelines/README.md


airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/utils/connector_runner.py
```diff
@@ -44,37 +44,38 @@ def output_folder(self) -> Path:
     def input_folder(self) -> Path:
         return self._volume_base / f"run_{self._runs}" / "input"
 
-    def _prepare_volumes(self, config: Optional[Mapping], state: Optional[Mapping], catalog: Optional[ConfiguredAirbyteCatalog]):
+    def _prepare_image(self, config: Optional[Mapping], state: Optional[Mapping], catalog: Optional[ConfiguredAirbyteCatalog]):
```
alafanechere, line 47: This is a patch whose reason is detailed [here](https://github.com/airbytehq/airbyte/blob/e388dbf5205d3f86770edb868b38cc4680a7a34a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/tests.py#L119) and that I'd prefer to avoid.

pyproject.toml
```diff
@@ -42,7 +42,7 @@ extend-ignore = [
 [tool.isort]
 profile = "black"
 color_output = false
-skip_gitignore = true
+# skip_gitignore = true
```
alafanechere, line 45: This is to allow black to run in a non-git context.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/builds.py
```diff
@@ -0,0 +1,81 @@
...
+        exclude (List, optional): The list of directory or file that should not be considered in the image build. Defaults to Optional[List].
+
+    Raises:
+        Exception: Raised if an error happened during image build.
+        Exception: Raised if an error happened during image tagging.
+
+    Returns:
+        Tuple[str, str]: The built image name and tag, the built image short id.
+    """
+    # Question to Dagger team: Can build and tag an image in Dagger?
```
kpenfound, line 62: `Publish` should be what you're looking for here: https://dagger-io.readthedocs.io/en/sdk-python-v0.3.3/api.html#dagger.api.gen.Container.publish
Example using publish: https://docs.dagger.io/730264/quickstart-publish

alafanechere: I had the impression `publish` induces a `docker push` that I don't want in this scenario. I would only like to `docker build . -t airbyte/source-pokeapi:dev` to locally tag an image.

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
```diff
@@ -0,0 +1,81 @@
...
+- Centralize the CI logic for connector testing 
+- Try out Dagger.io as a promising tool that can provide parallelism and caching out of the box for CI 
+
+## Install and use
+From `airbyte` repo root:
+
+### Install
+```bash
+cd tools/ci_connector_ops
+python -m venv .venv (please use Python 3.10)
```
evantahler, line 14: How do we enforce or change our python version?

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
```diff
@@ -0,0 +1,81 @@
...
+source .venv/bin/activate
```
evantahler, line 15: Gradle sort of does all of this for us right now.  We probably want a helper bash script to set this up for users who want to run this locally, e.g. `./bin/setup_ci.sh`

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
```diff
@@ -0,0 +1,81 @@
...
+(source-pokeapi does not require GSM access)
+```bash
+connectors-ci test-connector source-pokeapi
+```
+
+### **Run the pipeline for multiple connectors**
+
+```bash
+connectors-ci test-connector source-pokeapi source-openweather
+```
```
evantahler, line 47: Nice!

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/builds.py
evantahler: Can we also publish images for multiple/varied architectures (e.g. buildx)?

tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py
```diff
@@ -0,0 +1,154 @@
...
+from ci_connector_ops.pipelines.utils import StepStatus, write_connector_secrets_to_local_storage
+from ci_connector_ops.utils import Connector, ConnectorLanguage, get_changed_connectors
+from dagger.api.gen import Client, Container
+from graphql import GraphQLError
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+async def run_connector_test_pipelines(dagger_client: Client, connector: Connector, gsm_credentials: Optional[str]):
```
evantahler, line 24: Since this is async, can we run any of the sub-tasks in parallel?

tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    config = dagger.Config(log_output=sys.stdout)
+
+    async with dagger.Connection(config) as dagger_client:
+        for connector in connectors:
+            # We scoped this POC only for python and low-code connectors
+            if connector.language in [ConnectorLanguage.PYTHON, ConnectorLanguage.LOW_CODE]:
+                await run_connector_test_pipelines(dagger_client, connector, gsm_credentials)
+            else:
+                logger.warning(f"Not running test pipeline for {connector.technical_name} as it's not a Python or Low code connector")
```
evantahler, line 91: Is this loop synchronous or do we run all tests in parallel?  What if we wanted to throttle these in batches of N?

evantahler (commented):
Very cool POC!

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/utils/connector_runner.py
```diff
@@ -44,37 +44,38 @@ def output_folder(self) -> Path:
...
         self.input_folder.mkdir(parents=True)
-        self.output_folder.mkdir(parents=True)
+
+        current_image_tag = self._image.tags[0]
+        new_image_tag = current_image_tag + "-with-secrets"
+        dockerfile = f"FROM {current_image_tag}\n RUN mkdir /data\n"
```
cpdeethree, line 52: I think we could get away from this if we got rid of dockerfiles and instead used dagger api to do this behavior, no? Though that is probably a much larger refactor than this approach

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/builds.py
alafanechere: @evantahler yes we can. the `publish` function has an optional `platform_variant` parameter: https://dagger-io.readthedocs.io/en/sdk-python-v0.3.2/api.html#dagger.api.gen.Container.publish

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
alafanechere: The package declares that it requires Python 3.10 here: https://github.com/airbytehq/airbyte/blob/1c49dc760ef31fd5c5675a1557fc42f6ba1b56e9/tools/ci_connector_ops/setup.py#L51

I'm using pyenv to switch my python version to 3.10 when I'm working on this project. Our developers should have pyenv installed as I think it's part of our core python development tooling.

This file says to pyenv to switch to 3.10 when you're working on this code directory:
https://github.com/airbytehq/airbyte/blob/1c49dc760ef31fd5c5675a1557fc42f6ba1b56e9/tools/ci_connector_ops/.python-version#L1

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
alafanechere: I think we eventually want to publish this python package to a Pypi repository so that installing the CLI is as easy as:
```bash
pip install ci_connector_ops
connectors-ci test-connectors source-pokeapi
```


If we don't want to require user to have a properly setup python environment we could also consider packaging this project in its own docker image. In this scenario running the cli would look something like:
```bash
docker run --rm -i -v /var/run/docker.sock:/var/run/docker.sock airbyte/ci_connector_ops:latest connectors-ci test-connectors source-pokeapi
```
And if we consider this command not very nice to use on a daily basis we could do what we did for `octavia-cli` to install it we an alias in users bashrc: 
```
curl -s -o- https://raw.githubusercontent.com/airbytehq/airbyte/master/octavia-cli/install.sh | bash
```

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
alafanechere: By downloading the connector catalog we can easily create other nice commands like:
```bash
connectors-ci test-all-connectors
connectors-ci test-all-ga-connectors
connectors-ci test-all-beta-connectors
connectors-ci test-all-alpha-connectors
```

The command structure is not definitive and can easily be revamped to our liking, `e.g`:
```
connectors-ci test-connectors --language python --language low-code --release-stage beta
connectors-ci test-connectors --name source-pokeapi --name source-openweather
```

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+async def check_path_in_workdir(container: Container, path: str):
+    workdir = (await container.with_exec(["pwd"]).stdout()).strip()
+    mounts = await container.mounts()
+    if workdir in mounts:
+        expected_file_path = Path(workdir[1:]) / path
+        return expected_file_path.is_file() or expected_file_path.is_dir()
+    else:
+        return False
```
helderco, line 40: You'll be able to simplify this when https://github.com/dagger/dagger/issues/3764 is merged.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/builds.py
```diff
@@ -0,0 +1,81 @@
...
+    Args:
+        dagger_client (Client): _description_
+        connector_container (Container): _description_
+        extras (List, optional): List of extra_requires to intall e.g: tests, dev, main. (pip install -e .[tests]). Defaults to None.
+
+    Returns:
+        Container: A container in which the connector python package is installed with all its dependencies
+    """
+    if await check_path_in_workdir(connector_container, "requirements.txt"):
+        requirements_txt = await connector_container.file("requirements.txt").contents()
```
helderco, line 29: Until https://github.com/dagger/dagger/issues/3764 is ready, you can still simplify the use case of needing a file's contents.

```python
async def get_file_contents(container: dagger.Container, path: str) -> str:
    try:
        return await container.file(path).contents()
    except dagger.QueryError as e:
        if "no such file or directory" not in str(e):
            # this is the hicky bit of the stopgap because 
            # this error could come from a network issue
            raise
    return None

async def install(dagger_client: Client, connector_container: Container, extras: Optional[List] = None) -> Container:
    if requirements_txt := await get_file_contents(connector_container, "requirements.txt"):
        ...
```

The advantage is it makes only one query.

helderco (commented):
Nice work 👍 🙂 

I didn't review everything because I need to stop for now and don't want to leave it pending.

alafanechere:
Thank you so much @helderco for your review and suggestions!!!

tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py
alafanechere: I changed it following @helderco suggestion. This loop is non-blocking now.

tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py
alafanechere: Yes we can :)

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
evantahler: The pip package seems great! https://github.com/airbytehq/airbyte/issues/23582

bnchrch (commented):
Love it, excited to hear your thought on how this all felt coming together.

My review is a lot of "whats this do? why this"

But overall its looking really great

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/build_contexts.py
```diff
@@ -0,0 +1,46 @@
...
+    "pytest==6.2.5",
+    "coverage[toml]==6.3.1",
+    "pytest-custom_exit_code",  # I installed this plugin to work around the problem described in tests._run_tests_in_directory
+]
+
+
+def get_build_context(dagger_client: Client, connector: Connector) -> Container:
+    """Create a Python container in which the connector source code will be mounted alongside with the pyproject.toml.
+    We install the dependency required for our testing tools: formatting, unit tests, pip etc.
+    The pyproject.toml file gathers configurations for the formatting and testing tools eventually used downstream.
```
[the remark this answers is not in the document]
alafanechere: I just followed how things are done in {airbyte-python.gradle](https://github.com/airbytehq/airbyte/blob/7d7e48b2a342a328fa74c6fd11a9268e1dcdcd64/buildSrc/src/main/groovy/airbyte-python.gradle)
I think we use this file to centralize linting and formatting rules, and to have commonly shared rules by all connectors.

tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py
```diff
@@ -0,0 +1,154 @@
...
+        write_connector_secrets_to_local_storage(connector, gsm_credentials)
+
+    acceptance_tests_container: Container = build_dev_image_client.pipeline(f"{connector.technical_name} - Acceptance tests")
+    # The connector_image_short_id  is used as a cache buster. If it changed the acceptance tests will run.
+    acceptance_tests_status: StepStatus = await tests.run_acceptance_tests(
+        acceptance_tests_container, connector, connector_image_short_id, "airbyte/connector-acceptance-test:dev"
+    )
+
+    # TODO run QA checks: this should probably be done inside a dagger container to benefit from caching?
+    # TODO Upload modified secrets to GSM
```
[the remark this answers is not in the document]
alafanechere: The connector using short-lived token are likely to emit control messages with their updated configuration with the newest value of their refresh-token. We need to upload these configuration updates to GSM. This is done [here](https://github.com/airbytehq/airbyte/blob/master/.github/workflows/test-command.yml#L103) ATM.

tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py
```diff
@@ -0,0 +1,154 @@
...
+    main_pipeline_name = f"CI test for {connector.technical_name}"
+    pipeline_logger = logging.getLogger(main_pipeline_name)
+    connector_ci_client: Client = dagger_client.pipeline(main_pipeline_name)
+    build_context_client: Client = connector_ci_client.pipeline(f"{connector.technical_name} - Build context")
+    build_context: Container = build_contexts.get_build_context(build_context_client, connector)
+    format_check_container: Container = build_context.pipeline(f"{connector.technical_name} - Format Check")
+    format_check_status: StepStatus = await tests.check_format(format_check_container)
+
+    install_container: Container = build_context.pipeline(f"{connector.technical_name} - Install connector")
+    installed_connector: Container = await builds.install(connector_ci_client, install_container, extras=["tests", "dev", "main"])
```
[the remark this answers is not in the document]
alafanechere: Thank you for the naming suggestion!

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Questions for the Dagger team (in priority order)
+
+Dear Dagger team. You should be able to execute the code we pushed with the instructions above.
+Please ignore the `Set your environment variables` step and focus on running `connectors-ci test-connectors source-pokeapi` to reproduce the problems I mention below.
+
+1. How to handle exit codes: if exit_code == 1 an exception is thrown and stops the other pipeline execution. Code context [here](https://github.com/airbytehq/airbyte/blob/7d7e48b2a342a328fa74c6fd11a9268e1dcdcd64/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/tests.py#L25)
+2. Can we make with_mounted_directory writable so that the container can write to the host FS? Code context [here](https://github.com/airbytehq/airbyte/blob/7d7e48b2a342a328fa74c6fd11a9268e1dcdcd64/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/tests.py#L119)
+3. How to get access to visualizations: We'd love to have dynamic status checks on our GitHub PRs, with links to pipeline visualization [like](https://propeller.fly.dev/runs/da68273e-48d8-4354-8d8b-efaccf2792b9). 
+4. Can we build and tag an image in Dagger?
```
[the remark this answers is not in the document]
alafanechere: The `publish` function in Dagger does tag an image but also pushes it to a remote registry. 
I'd like to tag it locally only like we do for `:dev` images. I did a hacky workaround with the docker-py lib [here](https://github.com/airbytehq/airbyte/blob/a6ccd8bc6791830b44d00da59a711836bb446f17/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/builds.py#L75) but @kpenfound might suggest a better solution.

alafanechere:
> Love it, excited to hear your thought on how this all felt coming together.

I love it so far. It feels empowering to declare these things in 🐍 . The Dagger SDK is not very constraining and gives us full flexibility on how we could structure our code. I feel I came first with a simple code structure to tackle the use-case for python connector tests but there is room for powerful abstractions that could help us support other use cases (like Java connectors) in an elegant manner. 
Once we overcome the technical questions I asked the Dagger team I'd love to brainstorm about what our ideal code architecture would look like.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/tests.py
```diff
@@ -0,0 +1,156 @@
...
+    Returns:
+        StepStatus: Failure or success status of the check.
+    """
+    formatter = (
+        connector_container.with_exec(["echo", "Running black"])
+        .with_exec(RUN_BLACK_CMD)
+        .with_exec(["echo", "Running Isort"])
+        .with_exec(RUN_ISORT_CMD)
+        .with_exec(["echo", "Running Flake"])
+        .with_exec(RUN_FLAKE_CMD)
```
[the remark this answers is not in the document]
alafanechere: Do you mean you like the declaration of commands in constants?

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
alafanechere: Thanks for the issue link. I added it in a comment.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/builds.py
alafanechere: Thank you for the suggestion making me discover the walrus operator 🙏

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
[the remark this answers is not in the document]
cpdeethree: One thing have done as another alternative is spin up a local registry and publish to that. Ideally you should be able to tag images locally but a local image registry publish is a pretty viable workaround. Curious what other alternatives Dagger team might suggest though

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/tests.py
```diff
@@ -0,0 +1,156 @@
...
+        test_directory (str): The directory in which the tests are declared.
+
+    Returns:
+        StepStatus: Failure or success status of the tests.
+    """
+    # Question to dagger team: according to https://github.com/dagger/dagger/issues/3192 :
+    # A with_exec returning a non 0 status code will throw a TransportQueryError
+    # How can we execute with_exec commands that may return non 0 status code and decide further in the pipeline how to handle these status code
+    # If I'm not mistaken these TransportQueryError will stop the global pipeline execution
+    # In other words, if this with_exec fails we can't run other tests and get their results.
```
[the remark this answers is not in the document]
alafanechere: Thank you for the suggestion. I implemented the stop-gap and referenced the related issue on your repo in the comments.

tools/ci_connector_ops/ci_connector_ops/utils.py
```diff
@@ -96,16 +118,29 @@ def icon_path(self) -> Path:
...
+    def language(self) -> ConnectorLanguage:
+        if Path(self.code_directory / self.technical_name.replace("-", "_") / "manifest.yaml").is_file():
+            return ConnectorLanguage.LOW_CODE
+        if Path(self.code_directory / "setup.py").is_file():
+            return ConnectorLanguage.PYTHON
+        with open(self.code_directory / "Dockerfile") as dockerfile:
+            if "FROM airbyte/integration-base-java" in dockerfile.read():
+                return ConnectorLanguage.JAVA
+        raise ConnectorLanguageError(f"We could not infer {self.technical_name} connector language")
+
```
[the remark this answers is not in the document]
alafanechere: @helderco I'm not very familiar with asynchronous programmation in Python. Is it problematic to block the event loop for these file access operations that should be short? According to `aiopath` doc:
> If you used pathlib instead of aiopath, tasks accessing the disk would block the event loop, and async tasks accessing the network would suspend until the event loop was unblocked.

Is it very problematic to be in this blocking situation or is it simply an optimization path?
Do you have a suggestion to share regarding this specific Connector class. How would you expose properties that we could call from a synchronous and asynchronous context? Should I create an async variant of each property? e.g.: `async def language_async`?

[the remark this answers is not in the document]
helderco: > Is it very problematic to be in this blocking situation or is it simply an optimization path?

It's an optimization that you can ignore in some cases. You need to take into consideration other code that may be running concurrently.

Let's say you're just running a single task, only calling `await` synchronously. In this sense, if nothing's running during those `await`s, you're still waiting doing nothing. In this case, blocking the event loop won't impact you.

But if you're processing multiple connectors concurrently, or running the pipeline in a context where you already have an even loop running (e.g., as a webhook in a web service somewhere), then you could notice an impact if one piece of a connector pipeline blocks all others.

You could be lucky lets say, that even concurrently processing multiple connectors at the same time, you could be mostly waiting responses from the engine at the same time, i.e., waiting for I/O. So doing a blocking `Path.open()` won't also impact you much in this case (assuming it'll be relatively fast in comparison).

It does require some bit of duplication to support both sync/async. It's my major pain point developing the Python SDK (there's examples of this all over the code because we support both). As you're not familiar with async in Python, you can try to profile later and see if you need to optimize. Then decide if you want to keep both sync/async to optimize or just keep it in sync and let it block in these cases.

Ideally though, gaining awareness while developing of what's blocking or not, can save you time later.

Another option that you can use to avoid blocking the whole event loop on blocking code is to run it in a thread. You can see an example of this in the Python SDK where the provisioning part is being done in sync.

https://github.com/dagger/dagger/blob/318cb775bed51438f6eb45fb4c96db1c2a2d39cd/sdk/python/src/dagger/engine/conn.py#L46-L54

The [Asyncer](https://asyncer.tiangolo.com) library can also help you with this.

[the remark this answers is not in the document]
alafanechere: Thank you @helderco, this is very interesting. 
I think I'll keep the utils' code as it is for the moment. It is reused in other modules with synchronous execution so I'd prefer not to introduce async/await on the Connector class.
I'll keep in my this optimization in mind when we'll execute the connector loop on a large number of connectors.
If I understand correctly the purpose of Asyncer I could `asyncify` my synchronous code calls, it would be a good approach to overcome this problem by only touching the pipeline code.

In terms of profiling, do you have a specific tool to recommend? I could give https://github.com/joerick/pyinstrument a try?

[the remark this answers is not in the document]
helderco: > In terms of profiling, do you have a specific tool to recommend? 

Not really, not specifically 🙂

I've not acquired the habit of profiling myself because I didn't use to hit performance issues. That's when you reach for it, when you want to look for bottlenecks in your code in order to speed it up. I've used the profiling that's integrated with the IDE (PyCharm), and one tool or another, but I don't remember which ones. PyInstrument looks good though. 🤩

So, since this is a POC, you can leave it to the end or skip it, if you don't mind the performance you're getting, especially if the POC has a very narrow scope. But if you're interested in optimizing your results in the end, or start thinking about scaling it up, then profiling can help there. I'm talking performance in general, not just to help spot blocking code.

I think most of the waiting you'll see in your pipeline code will be mostly waiting for a response from the engine. Just building a container and running a few execs in a single query can take a while depending on your request. Reading a small file from disk with `Path.open()` will be pretty fast in comparison. That won't block the engine if it's running a query, just Python being able to process more requests and responses. 

However, you can also easily miss some easy performance optimizations if you don't take advantage of the concurrency. If want to see another example of that in practice, take a look at these two examples for testing the FastAPI library with dagger:
- [helderco/dagger-examples/test_multiple.py](https://github.com/helderco/dagger-examples/blob/main/test_multiple.py) → ~3:12s
- [helderco/dagger-examples/test_concurrency.py](https://github.com/helderco/dagger-examples/blob/main/test_concurrency.py) → ~1:15s

`test_multiple.py` is using async dagger but in a sync way. There's no concurrency. When you use `await` in a synchronous way, it's the same as using sync Python. `test_concurrency.py` however, is taking advantage of async and running the tests concurrently, by using an [AnyIO task group](https://anyio.readthedocs.io/en/stable/tasks.html).

You can see the execution time gain from 3.12s to 1.15s (both running with clean cache). That's because instead of waiting for the first test suite to finish before requesting the second test suite to start, you're just making the requests one after the other, and then waiting for all responses before continuing. So if each version test takes longer individually, you'll see a bigger gap in execution time between both implementations.

If you want to learn a bit more about this, may I suggest this easy guide from the FastAPI documentation and the "concurrent burgers" analogy:

👉 [FastAPI - Concurrency and async / await](https://fastapi.tiangolo.com/async/#technical-details)

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/tests.py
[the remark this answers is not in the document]
evantahler: ^ I like that anyway!

evantahler (commented):
I don't really know how to review this 😓... perhaps a loom video or other demo would be a good entrypoint for reviewers?

BUT... it looks like you sped up the test by 1/2! That's a win!

alafanechere:
> I don't really know how to review this 😓...

@evantahler I'd suggest you focus on the [entry points](https://github.com/airbytehq/airbyte/blob/augustin/poc/ci-in-python/tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py#L78) and [the documented DAG ](https://github.com/airbytehq/airbyte/blob/augustin/poc/ci-in-python/tools/ci_connector_ops/ci_connector_ops/pipelines/README.md). The `[run](https://github.com/airbytehq/airbyte/blob/augustin/poc/ci-in-python/tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py#L78)` functions holds the DAG logic and I'd love to make sure its as readable and understandable as possible.

alafanechere:
@cpdeethree @evantahler  I was able to run the pipeline in GHA:
* [A `source-pokeapi` run](https://github.com/airbytehq/airbyte/actions/runs/4356572688)
* [A long running job on 3 GA sources](https://github.com/airbytehq/airbyte/actions/runs/4356557536)

The command to check only modified connectors behaves strangely - I'll investigate soon.

.github/workflows/connector_integration_test_single_dagger.yml
```diff
@@ -2,9 +2,21 @@ name: POC Connectors CI - test pipeline
...
+        description: "List of connectors to tests"
+      test_modified_connectors:
+        description: "Test only modified connectors on the current branch"
+        type: boolean
+        default: false
+  pull_request:
+    paths:
+      - "airbyte-integrations/connectors/**"
+  push:
+    paths:
```
cpdeethree, line 16: can we move this logic into the dagger pipeline?

.github/workflows/connector_integration_test_single_dagger.yml
```diff
@@ -27,30 +39,57 @@ jobs:
           aws-secret-access-key: ${{ secrets.SELF_RUNNER_AWS_SECRET_ACCESS_KEY }}
           github-token: ${{ env.PAT }}
           ec2-instance-type: "c5.2xlarge" # https://aws.amazon.com/ec2/instance-types/
-  launch_integration_tests:
+  launch_connector_tests:
     timeout-minutes: 120
-    runs-on: ${{ needs.start-connectors-integration-test-runner.outputs.label }} # run the job on the newly created runner
+    needs: start-runner
+    runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
+    env:
```
cpdeethree, line 46: eventually I want to use SOPS to handle this, but this works for now

.github/workflows/connector_integration_test_single_dagger.yml
alafanechere: Yes, but without this filter, the GitHub action will be triggered on each code change and boot a runner. 💸

.github/workflows/connector_integration_test_single_dagger.yml
```diff
@@ -27,30 +39,57 @@ jobs:
...
       - name: Install ci-connector-ops package
         run: pip install --quiet -e ./tools/ci_connector_ops
-  stop-connectors-integration-test-runner:
+      - name: Run connectors-ci for specific connectors
+        if: "${{ github.event.inputs.connectors_under_test != '' }}"
+        run: connectors-ci --is-local=False test-connectors ${{ github.event.inputs.connectors_under_test }}
+        env:
+          CI_GIT_BRANCH: ${{ steps.extract_branch.outputs.branch }}
+      - name: Run connectors-ci for modified connectors
+        if: github.event_name == 'pull_request' || github.event_name == 'push' || github.event.inputs.test_modified_connectors == 'true'
```
cpdeethree, line 79: A bit wordy, but feels necessary due to the event input.

Also, push is a bit greedy here. I think you should only run on push to master instead of just push so that we aren't spinning up a workflow every time that someone pushes an arbitrary branch that isn't part of a pull request. You can do that by modifying the "on:" push clause to only include master branch above

.github/workflows/connector_integration_test_single_dagger.yml
```diff
@@ -27,30 +39,57 @@ jobs:
...
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: extract_branch
       - name: Install Python 3.10
         uses: actions/setup-python@v4
         with:
           python-version: "3.10"
-      - name: Upgrade pip
-        run: pip install --upgrade pip
       - name: Install ci-connector-ops package
         run: pip install --quiet -e ./tools/ci_connector_ops
```
cpdeethree, line 72: Maybe this is out of scope of the POC but I would like to get out of running this in editable mode eventually

.github/workflows/connector_integration_test_single_dagger.yml
cpdeethree: On second thought, I think we should leave this here for now because it will skip spinning up the large runner when it's here. Once we shift to using EKS I think we can move it then

.github/workflows/connector_integration_test_single_dagger.yml
```diff
@@ -27,30 +39,57 @@ jobs:
...
     timeout-minutes: 120
-    runs-on: ${{ needs.start-connectors-integration-test-runner.outputs.label }} # run the job on the newly created runner
+    needs: start-runner
+    runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
+    env:
+      GCP_GSM_CREDENTIALS: ${{ secrets.GCP_GSM_CREDENTIALS }}
+      AWS_ACCESS_KEY_ID: ${{ secrets.STATUS_API_AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.STATUS_API_AWS_SECRET_ACCESS_KEY }}
+      AWS_DEFAULT_REGION: "us-east-2"
+      TEST_REPORTS_BUCKET_NAME: "airbyte-connector-build-status"
```
cpdeethree, line 51: Can we add the environment variables for gradle caching like in your other PR? Presumably this is going to invoke gradle too

.github/workflows/connector_integration_test_single_dagger.yml
cpdeethree: lol, just noticed you commented the same thing above, guess i didn't refresh before I commented

.github/workflows/connector_integration_test_single_dagger.yml
alafanechere: But how would we get the guarantee that all the connectors changed in a PR are passing tests before merge? 
Do you mean we should avoid the on-push for the moment until we get the infra with remote cache ready?

.github/workflows/connector_integration_test_single_dagger.yml
alafanechere: We can discard `-e` in this context I think.

.github/workflows/connector_integration_test_single_dagger.yml
alafanechere: I don't think I get what you mean 😄

.github/workflows/connector_integration_test_single_dagger.yml
alafanechere: Could you share what would be done differently with SOPS?

tools/ci_connector_ops/ci_connector_ops/pipelines/models.py
```diff
@@ -0,0 +1,99 @@
...
+from dataclasses import dataclass, field
+from datetime import datetime
+from enum import Enum
+from typing import List, Optional, Union
+
+from ci_connector_ops.pipelines.contexts import ConnectorTestContext
+from dagger import Client, Container
+
+
+class Step(Enum):
```
bnchrch, line 14: I like how you broke out the steps here

tools/ci_connector_ops/ci_connector_ops/pipelines/models.py
```diff
@@ -0,0 +1,99 @@
...
+    CODE_FORMAT_CHECKS = "Code format checks"
+    PACKAGE_INSTALL = "Package install"
+    UNIT_TESTS = "Unit tests"
+    INTEGRATION_TESTS = "Integration tests"
+    DOCKER_BUILD = "Docker Build"
+    ACCEPTANCE_TESTS = "Acceptance tests"
+    QA_CHECKS = "QA Checks"
+
+    def get_dagger_pipeline(self, dagger_client_or_container: Union[Client, Container]) -> Union[Client, Container]:
+        return dagger_client_or_container.pipeline(self.value)
```
bnchrch, line 24: Can you explain how a client and a container differ? and why they both have a pipeline?

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/tests.py
```diff
@@ -0,0 +1,212 @@
...
+        .with_env_variable("CACHEBUSTER", connector_under_test_image_id)
+        .with_mounted_directory("/test_input", context.get_connector_dir(exclude=["secrets", ".venv"]))
+        .with_directory("/test_input/secrets", context.secrets_dir)
+        .with_entrypoint(["python", "-m", "pytest", "-p", "connector_acceptance_test.plugin", "-r", "fEsx"])
+        .with_exec(["--acceptance-test-config", "/test_input"])
+    )
+
+    secret_dir = cat_container.directory("/test_input/secrets")
+    updated_secrets_dir = None
+    if secret_files := await secret_dir.entries():
```
bnchrch, line 163: Nit not related to a poc, only if we merge.

Why do you think about:
1. Moving the secret file directory logic to a named function?
2. Move "/test_input/secrets" to a constant

bnchrch (approved):
Honestly, theres nothing that jumps out to me.

All I have is one kudos, one nitpick and one question.

Overall my impression is
1. Its much easier to follow the logic now that its in a single modern programming language
2. From a naive view the pipeline primitives seem very composable which bodes well for reuse
3. The Dagger technology is new enough that we will be working with unfinished edges.

Other than that Im much more interested to hear yours and the infra teams opinions after the Dagger team answers your pointed questions.

.github/workflows/connector_integration_test_single_dagger.yml
cpdeethree: We would not load secrets as environment variables here, only GCP service account key. We would store the secrets encrypted as a file in the repository. We would use the GCP service account key to authenticate to GCP to access the key needed to decrypt the secrets, and then we would decrypt the secrets and use them in the workflow as part of the CI run.

evantahler:
I'm trying to run this locally, and i'm in a pickle:

```
(.venv) ➜  airbyte git:(augustin/poc/ci-in-python) ✗ connectors-ci --use-remote-secrets=False
Usage: connectors-ci [OPTIONS] COMMAND [ARGS]...
Try 'connectors-ci --help' for help.

Error: Missing command.
---
(.venv) ➜  airbyte git:(augustin/poc/ci-in-python) ✗ connectors-ci --help
Usage: connectors-ci [OPTIONS] COMMAND [ARGS]...

  A command group to gather all the connectors-ci command

Options:
  --use-remote-secrets BOOLEAN
  --is-local BOOLEAN
  --git-branch TEXT
  --git-revision TEXT
  --help                        Show this message and exit.

Commands:
  test-all-modified-connectors  Launches a CI pipeline for all the...
  test-connectors               Runs a CI pipeline the connector passed...
---
(.venv) ➜  airbyte git:(augustin/poc/ci-in-python) ✗ connectors-ci test-connectors source-pokeapi
Usage: connectors-ci [OPTIONS] COMMAND [ARGS]...
Try 'connectors-ci --help' for help.

Error: You have to set the GCP_GSM_CREDENTIALS if you want to download secrets from GSM. Set the --use-gsm-secrets option to false otherwise.
---
(.venv) ➜  airbyte git:(augustin/poc/ci-in-python) ✗ connectors-ci --use-gsm-secrets False
Usage: connectors-ci [OPTIONS] COMMAND [ARGS]...
Try 'connectors-ci --help' for help.

Error: No such option: --use-gsm-secrets Did you mean --use-remote-secrets?
```

.github/workflows/connector_integration_test_single_dagger.yml
cpdeethree: > But how would we get the guarantee that all the connectors changed in a PR are passing tests before merge?

The pull_request event handles that. Basically it will run every time you push to a branch, as long as there is an open pull request for that branch. 

restricting push to only run on master basically prevents invocation of the tests when you push to a non-master branch with no pull request open on it. But it would run as soon as you opened the pull request and also to subsequent pushes to that branch, so long as the PR is open.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/builds.py
```diff
@@ -0,0 +1,40 @@
...
+#
+
+import tempfile
+from typing import List, Optional, Tuple
+
+import docker
+from ci_connector_ops.pipelines.contexts import ConnectorTestContext
+
+
+async def build_dev_image(context: ConnectorTestContext, exclude: Optional[List] = None) -> Tuple[str, str]:
```
cpdeethree, line 12: Personal preference, I think this function name is not descriptive enough. Perhaps calling out that this is specificially used to run the test suite might be better? I also think it is not very reusable outside of this test suite, so perhaps does not necessarily belong in the "actions" folder. 

Willing to debate on these points though!

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -0,0 +1,172 @@
...
+
+from typing import List, Optional
+
+from ci_connector_ops.pipelines.contexts import ConnectorTestContext
+from ci_connector_ops.pipelines.utils import get_file_contents
+from dagger import CacheSharingMode, CacheVolume, Container, Directory, Secret
+
+PYPROJECT_TOML_FILE_PATH = "pyproject.toml"
+
+CONNECTOR_TESTING_REQUIREMENTS = [
```
cpdeethree, line 15: I'm not a huge fan of embedding these pins in a python file. Can we use a top level .env file or similar?

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/remote_storage.py
```diff
@@ -0,0 +1,34 @@
...
+    Args:
+        dagger_client (Client): The dagger client.
+        file_to_upload_path (Path): The local path to the file to upload.
+        key (str): The key that will be written on the S3 bucket.
+        bucket (str): The S3 bucket name.
+
+    Returns:
+        int: Exit code of the upload process.
+    """
+    file_to_upload: File = dagger_client.host().directory(".", include=[str(file_to_upload_path)]).file(str(file_to_upload_path))
```
cpdeethree, line 21: nice 👍

evantahler:
THIS IS VERY COOL

These are all nits:

**DX Notes**
* When downloading a connector to test, we should log "downloading image `airbyte/soruce-pokeapi:latest" or something like that to show what is happening.  Right now it looks like the process is hanging
* `ERROR:ci_connector_ops.pipelines.connectors_ci:failed to compute cache key: failed to calculate checksum of ref qu6onr5srogz207c5swoohsua::u378khn0p09a72optelo7uigb: "/secrets": not found`.  We should `mkdir -p` this directory, or check if it exists.  Pokeapi doesn't need it.
  * How would I know which directory on my machine the above error is about? I thought it was `airbyte-integrations/connectors/source-pokeapi/integration_tests/secrets` but that's not it
  * EDIT: it was `airbyte-integrations/connectors/source-pokeapi/secrets/config.json`
* It looks like we build a local SAT image.  Do we still want to do that?
  * Probably... let's make that change explicitly as part of another project
* I like the stoplight summary at the end!
  * It's hard to find specific test output within the big combined log, so this is probably a nescecity
  * If part of the build failed (e.g. my missing `secrets/config.json` file), can we still get the stoplights?  Is there an `always()` part of the DAG?
  * Is it possible to highlight some specific container output at the end of the run, e.g. the pytest reults?


Stoplight:

```
#16 docker build
INFO:CI test for source-pokeapi:
SOURCE-POKEAPI - TEST RESULTS
Package install: 🟢 — Successful
Code format checks: 🟢 — Successful
Unit tests: 🟢 — Successful
Integration tests: 🟡 - Skipped
Docker Build: 🟢 — Successful
Acceptance tests: 🟢 — Successful
QA Checks: 🟢 — Successful
Total duration: 88 seconds
```

tools/ci_connector_ops/ci_connector_ops/utils.py
```diff
@@ -148,15 +185,30 @@ def acceptance_test_config(self) -> Optional[dict]:
...
+    """Retrieve a set of Connectors that were changed in the current branch (compared to master)."""
     changed_source_connector_files = {
         file_path
         for file_path in AIRBYTE_REPO.git.diff("--name-only", DIFFED_BRANCH).split("\n")
         if file_path.startswith(SOURCE_CONNECTOR_PATH_PREFIX)
     }
     return {Connector(get_connector_name_from_path(changed_file)) for changed_file in changed_source_connector_files}
+
+
+def get_changed_connectors_between_branches(branch_a: str, branch_b: str) -> Set[Connector]:
```
cpdeethree, line 205: future work if you don't want to do it for this poc: Make this a generic diffing function (instead of connector specific), since it's definitely something we will want to do for basically every pipeline given our setup

tools/ci_connector_ops/setup.py
```diff
@@ -2,39 +2,62 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
 
+from pathlib import Path
+
 from setuptools import find_packages, setup
 
 MAIN_REQUIREMENTS = [
```
cpdeethree, line 9: same as above, i think we should promote this to a top level .env file (even at repo level!)

tools/ci_connector_ops/ci_connector_ops/utils.py
```diff
@@ -148,15 +185,30 @@ def acceptance_test_config(self) -> Optional[dict]:
...
             return None
 
+    def get_secret_manager(self, gsm_credentials: str):
+        return SecretsManager(connector_name=self.technical_name, gsm_credentials=gsm_credentials)
+
     def __repr__(self) -> str:
         return self.technical_name
 
+
 def get_changed_connectors() -> Set[Connector]:
```
cpdeethree, line 195: Same as below, this one definitely feels like it could be generic diff function

c-p-b (changes requested):
I don't think we should consider the POC complete without a test harness in place. Let's discuss tomorrow?

tools/ci_connector_ops/ci_connector_ops/pipelines/models.py
alafanechere: A client is client of Dagger API. 
You can create containers from it, access the host etc... 
A container is a dagger container. 

The Dagger SDK allows creating "pipelines" from clients and containers. 
A pipeline groups client or container in execution in a logical way, it has no impact on the underlying execution flow but will help build the DAG representation and logging structure. 

Calling `pipeline` on a container returns a container. Calling pipeline on a client returns a client.
You would call `pipeline` on the original client if you want to create a new client for a specific group of step.
You would call `pipeline` on a base container if you want to make the next interaction on this container in a specific group of step.

I created this `get_dagger_pipeline` function to make the `Step` model responsible for generating their pipelines.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/builds.py
alafanechere: Sure, I originally put it in this module because it was generic enough to build any image, but I latter changed it to build connector images only for testing. Will move to `tests`.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
alafanechere: I can create a `connector_testing_requirements.txt` file. I like to store it in a constant because it makes it easier to mock it for testing.

tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py
```diff
@@ -0,0 +1,289 @@
...
+    Args:
+        test_context (ConnectorTestContext): The initialized test context.
+
+    Returns:
+        ConnectorTestReport: The test reports holding tests results.
+    """
+    test_context = await setup(test_context)
+    qa_checks_results_future = asyncio.create_task(tests.run_qa_checks(test_context))
+    connector_source_code = await environments.with_airbyte_connector(test_context, install=False)
+    connector_under_test = await environments.with_airbyte_connector(test_context)
```
helderco, line 113: I think we can improve the readability (and maybe performance as well) on the body of this coroutine. I'll see if I can come up with something tomorrow and get back to you.

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+The main differences are that:
+- The pipeline will pull the branch under test from Airbyte's GitHub repo
+- The pipeline will upload per connector test reports to S3 
+
+
+## What does a connector pipeline run
+
+```mermaid
+flowchart LR;
```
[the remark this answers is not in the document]
helderco: After looking at the code, doesn't this need to be update?

helderco (commented):
Tried to focus more on dagger and async parts (performance). There's some places you can optimize without effort. 🚀

I didn't get to execute the pipeline at this time, just GitHub. As I say in a comment, I'll try to come up with some organizational improvements soon. 🙂

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/tests.py
```diff
@@ -0,0 +1,212 @@
...
+                "python",
+                "-m",
+                "pytest",
+                "-s",
+                test_directory,
+                "-c",
+                test_config,
+            ]
+        )
+        return StepStatus.from_exit_code(await with_exit_code(tester)), await with_stderr(tester), await with_stdout(tester)
```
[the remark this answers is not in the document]
helderco: Note that most of the waiting will be in executing a command. Getting all three of those should be quite fast due to the underlying cached command, but they're still easy pickings to put in a task group.

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/utils/connector_runner.py
jpadams: We use "dagger in dagger" for our own CI, so we can definitely explore this in the future 😃

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/builds.py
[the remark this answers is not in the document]
alafanechere: Thanks for pointing this out.
I discarded this logic as I can build the image directly on the "proxy docker host" introduced for the usage of the new `insecure_root_capabilities` in SDK 0.4.0.
I made the changes here: https://github.com/airbytehq/airbyte/pull/23487/commits/3f50fa7e01b346e586ce9ab6b2ef12b1b064d0d3#diff-579ff5a783a01ad14dcb90bf6e4fc73e7d5daaafbd189e8e1cbfc5652ccba94b

tools/ci_connector_ops/ci_connector_ops/pipelines/models.py
```diff
@@ -0,0 +1,100 @@
...
+            }
+        )
+
+    def __str__(self) -> str:
+        nice_output = f"\n{self.connector_test_context.connector.technical_name.upper()} - TEST RESULTS\n"
+        nice_output += "\n".join([str(sr) for sr in self.steps_results]) + "\n"
+        nice_output += f"Total duration: {round(self.run_duration)} seconds"
+        for failed_step in self.failed_steps:
+            nice_output += f"{failed_step.step.value} FAILURES:\n{failed_step.stderr}"
+        return nice_output
```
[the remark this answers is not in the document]
alafanechere: Thank you for the suggestion. I used it (7a026c2f4a091f0918107721e09b612d438f4a33) and generate reports like this 🤩 

Screen Shot 2023-03-09 at 13 48 30

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/utils/connector_runner.py
alafanechere: The connector-acceptance-test project does not need to be patched now that I can use the `insecure_root_capabilities` flag.

tools/ci_connector_ops/ci_connector_ops/utils.py
alafanechere: I changed the logic for a more generic approach here: https://github.com/airbytehq/airbyte/blob/b9b8474a5e063e241d0c74a133439de0ab416daf/tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py#L128

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -0,0 +1,163 @@
...
+        return await container.file(path).contents()
+    except QueryError as e:
+        if "no such file or directory" not in str(e):
+            # this is the hicky bit of the stopgap because
+            # this error could come from a network issue
+            raise
+    return None
+
+
+def write_connector_secrets_to_local_storage(connector: Connector, gsm_credentials: str):
```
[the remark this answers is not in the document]
alafanechere: My bad, this function is not used anymore. I containerized the lib managing container secrets, it exposes a CLI so I'm calling it here https://github.com/airbytehq/airbyte/blob/362679310dbb753f64cef9498c723da465c509a9/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/secrets.py#L71

tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py
```diff
@@ -0,0 +1,289 @@
...
+                if test_context.connector.language in [ConnectorLanguage.PYTHON, ConnectorLanguage.LOW_CODE]:
+                    test_context.dagger_client = dagger_client.pipeline(f"{test_context.connector.technical_name} - Test Pipeline")
+                    tg.start_soon(run, test_context)
+                    if not test_context.is_local:
+                        send_commit_status_check(
+                            test_context.git_revision,
+                            "pending",
+                            test_context.gha_workflow_run_url,
+                            f"Running tests for {test_context.connector.technical_name}",
+                        )
```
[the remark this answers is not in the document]
alafanechere: I asyncified it 6ac3a758a876dd35bf31be2ab2b2a387baf20260

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/remote_storage.py
```diff
@@ -0,0 +1,34 @@
...
+    aws_region: Secret = dagger_client.host().env_variable("AWS_DEFAULT_REGION").secret()
+    return await (
+        dagger_client.container()
+        .from_("amazon/aws-cli:latest")
+        .with_file(str(file_to_upload_path), file_to_upload)
+        .with_secret_variable("AWS_ACCESS_KEY_ID", aws_access_key_id)
+        .with_secret_variable("AWS_SECRET_ACCESS_KEY", aws_secret_access_key)
+        .with_secret_variable("AWS_DEFAULT_REGION", aws_region)
+        .with_exec(["s3", "cp", str(file_to_upload_path), f"s3://{bucket}/{key}"])
+        .exit_code()
```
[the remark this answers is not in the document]
alafanechere: I think I want to bubble it up as it's a crucial logic and it has few moving parts.

c-p-b (approved):
Per discussion on slack, unit tests to be added in a future pr

tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py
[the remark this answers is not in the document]
helderco: Fyi, `asyncify` is just doing `anyio.to_thread.run_sync` underneath:

```python
def asyncify(
    function: Callable[T_ParamSpec, T_Retval],
    *,
    cancellable: bool = False,
    limiter: Optional[anyio.CapacityLimiter] = None
) -> Callable[T_ParamSpec, Awaitable[T_Retval]]:
    async def wrapper(
        *args: T_ParamSpec.args, **kwargs: T_ParamSpec.kwargs
    ) -> T_Retval:
        partial_f = functools.partial(function, *args, **kwargs)
        return await anyio.to_thread.run_sync(
            partial_f, cancellable=cancellable, limiter=limiter
        )
    return wrapper
```

As you can see, the main point of _asyncer_ is to provide a better typing experience in the IDE. If you remove the types, that can be reduced to just:

```python
asyncify = anyio.to_thread.run_sync
```

Just saying this to put the added dependency in context. 🙂

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/remote_storage.py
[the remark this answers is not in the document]
helderco: Problem is it'll interrupt the rest of teardown (i.e., skip over running `update_commit_status_check`) and the whole test report if you don't handle it.

alafanechere:
@evantahler 
> Running the CAT container got really noisy
I think I succeed in making it a bit less noisy by changing the level of `dockerd` (there are still docker related loglines I don't really get the origin.

> If there is a SAT failure, the container exists with non-0 and then it looks like the whole workflow fails:
Fixed

tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py
```diff
@@ -0,0 +1,230 @@
...
+    """Runs a CI pipeline for a single connector.
+    A visual DAG can be found on the README.md file of the pipelines modules.
+
+    Args:
+        test_context (ConnectorTestContext): The initialized test context.
+
+    Returns:
+        ConnectorTestReport: The test reports holding tests results.
+    """
+    test_context = await setup(test_context)
```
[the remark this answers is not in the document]
alafanechere: I implemented the context manager interface in the ConnectorTestContext class. It's indeed far more elegant and perfectly fit what I needed. Thanks!!!
7f3ba4ac26a66b531607d6d689c0a42647cc7fd3

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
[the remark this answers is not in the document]
helderco: You're right, just took a closer look. I'm trying to get a better sense of the pipeline to help think how the structure could be improved. Can you confirm if the following looks correct? Note that this represents how it **could** be organized, not how it’s currently organized.


```mermaid
flowchart TB
    conn([Connector pipeline])==>qa & fmt & tu
    src[[Mount connector source code to container]]-->fmt & pkg
    fmt & tu & ti & ta & qa -.-> r([Build test report])--if in CI-->s3[[Upload to S3]]
    subgraph QA Checks
        repo[[Mount Airbyte repo to container]]-->qa[Run QA checks]
    end
    subgraph Tests
        tu[Run unit tests]==>ti[Run integration tests] & ta[Run acceptance tests]
        pkg[[Install connector package in container]]-->tu & ti
        subgraph Acceptance
            dsec[[Download secrets from GSM]] & bld[[Build connector dev image]]-->ta-->upsec[[Upload updated secrets to GSM]]
        end
    end
    subgraph Code format
        fmt[Run code format checks]
    end
```

### Notes

- This represents a single connector's pipeline;
- The [stadium-shaped](https://mermaid.js.org/syntax/flowchart.html#a-stadium-shaped-node) nodes mark beginning and end;
- The default square nodes mark the main steps;
- The [subroutine shaped](https://mermaid.js.org/syntax/flowchart.html#a-node-in-a-subroutine-shape) nodes mark dependencies (input) or sub-processes (output) and are better called from the main steps;
- The thick links highlight the (higher level) flow for the main steps;
- The normal links highlight the flow for the subroutines;
- The dotted links are “soft dependencies”, i.e., any failure in each step should be catched so it always returns a status result;
- As always, any node with a failed dependency will fail as well;
- The subgraphs represent cancel scopes, i.e., a failure inside one shouldn’t cancel others outside it;
- Cancel scopes are a good way to decide when to use task groups[^1] (for concurrency), e.g., “Tests”, “QA checks” and “Code Format”, where “Acceptance” adds its sub-routines to the parent task group;
- The subgraphs also suggest better encapsulation in separate coroutines;
- Current report steps includes “Package install” but it looks like it’s just a dependency of the test suite (unit and integration tests). Current graph also reflects this since there’s no link to “Build test report”.

[^1]: [Cancel scopes](https://anyio.readthedocs.io/en/stable/cancellation.html) are a feature of [task groups](https://anyio.readthedocs.io/en/stable/tasks.html).

alafanechere:
I'll merge now. There are still great improvements to make but I feel like they should be done incrementally. IMO the current state of this PR accomplished the goal of reimplementing airbyte-python.gradle and airbyte-connector-acceptance.gradle in Python with Dagger.
We'll observe how it behaves on connector PR after merge and probably find interesting learnings.
Here are the main improvements to make according to the review from @helderco and my personal opinion:
* Rework the concurrency implementation. We shall rely only on aniyo and not call asyncio native code. 
* Some DAG logic is not perfect: the integration tests should get secrets too. 
* The `ci_connector_ops` python package and dependencies are becoming difficult to manage because we stuffed it with different project that have very diverse needs in term of python version and and libraries.
* We need to focus a bit more on making sure each step re-execution is legit and leverage caching as much as possible. (writing secrets on every run leads to retrigger of acceptance test on every run because the secrets file timestamp changed even if there content is different).
* We might consider relying more on OOP to benefit from abstraction that can be useful to run pipelines for Python and Java connectors with common logic and reusable code.

cc @evantahler @cpdeethree @bnchrch @jpadams @kpenfound

bazarnov:
@alafanechere the https://github.com/airbytehq/airbyte/actions/runs/4391568285 test is failing but says  `success`, not sure what's going on here, need more context around the work you've done, but I think I can safely merge in the related PR: 
https://github.com/airbytehq/airbyte/pull/23822 

Am I right? or should I wait until it's fixed on your side?

Issue #22442, opened 2023-02-06
Declarative connectors runtime: Extend CDK publish process to update docker base image


flash1293:
When publishing a new version of the CDK, the docker base image implemented as part of https://github.com/airbytehq/airbyte/issues/22438 should be re-built with the new version and published as well.

## Acceptance criteria

* There is a process in place that ensures there is a version of the `airbyte/source-declarative-manifest` docker image for every released CDK version with matching version numbers
* Docker image should only publish if publish to PyPI also succeeded

## Implementation Hints
1. there is a script `tools/integrartions/manage.sh` which builds & publishes docker images for connectos. we can reuse that. 
2. we might also need to modify the CI `.github` yaml files to push that image

maxi297:
https://github.com/airbytehq/github-workflow-test-repo-base repo can be used to perform some tests

PR #23501, opened 2023-02-27
[ISSUE #22442] publish docker

maxi297:
## What
Addresses #22442 

Tests are documented here https://github.com/airbytehq/airbyte/actions/workflows/test-publish-cdk-docker-image.yml
Those tests created real "prod" Docker images which can be seen here: https://hub.docker.com/r/airbyte/source-declarative-manifest/tags (I don't think it is worth cleaning them but feel free to suggest otherwise and I'll find how I can clean them)

## How
Adding a job to the GitHub when publishing CDK manually


.github/workflows/publish-cdk-command-manually.yml
```diff
@@ -220,3 +249,90 @@ jobs:
...
+        with:
+          attempt_limit: 3
+          attempt_delay: 5000 # in ms
+          command: curl -sL https://sentry.io/get-cli/ | bash || echo "sentry cli already installed"
+      - name: Set Name and Version Environment Vars
+        run: |
+          source tools/lib/lib.sh
+          echo "IMAGE_NAME=source-declarative-manifest" >> $GITHUB_ENV
+          echo "IMAGE_VERSION=$(_get_docker_image_version airbyte-cdk/python/Dockerfile)" >> $GITHUB_ENV
+      - name: Create Sentry Release
```
maxi297, line 300: This was part of the publish pipeline but I'm not sure how it works. I've asked for more information here https://airbytehq-team.slack.com/archives/C02UF50V9HA/p1677616768240659. In the meanwhile, I validated that the release is created in Sentry but could not validate how this release gets populated with information so this might be useless

maxi297: After some discussions with Pedro, this is not necessary as Sentry will be creating the release automatically if it does not exist

.github/workflows/publish-cdk-command-manually.yml
```diff
@@ -220,3 +249,70 @@ jobs:
...
+        uses: actions/checkout@v3
+        with:
+          repository: airbytehq/airbyte
+          ref: master
+      - name: Install Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: "zulu"
+          java-version: "17"
+      - name: Install Python
```
clnoll, line 271: Is Python required? I didn't see where it was needed in the subsequent steps.

clnoll (approved):
LGTM @maxi297, just one question.

flash1293 (commented):
Two basic questions:
* What about letting sentry know we did a release? Do we need this? https://github.com/airbytehq/airbyte/blob/7953c45a093345abfed6cf03fe4455c89b336689/.github/workflows/publish-command.yml#L323-L332
* The code looks like it would take care of that, but I'm surprised there is no 0.29.0 image on hub.docker.com - will we do the first "real" release when releasing the next CDK version?

flash1293 (approved):
Caught up on Slack discussions - looks fine for me for now 👍

.github/workflows/publish-cdk-command-manually.yml
maxi297: That's a good question. `manage.sh` uses command like `./gradlew --no-daemon :airbyte-cdk:python:build -x test` (see https://github.com/airbytehq/airbyte/actions/runs/4296899471/jobs/7489230541#step:7:558) which I assumed uses Python. That being said, I haven't confirmed so if you think it shouldn't, let me know and I'll test the workflow without it

clnoll: Ah yep looks like you're right, I just overlooked that.

Issue #23391, opened 2023-02-23
Parallelise test execution on Postgres source connector 


sergio-ropero:
## Tell us about the problem you're trying to solve
Based on issue [22968](https://github.com/airbytehq/airbyte/issues/22968)

Nowadays the test for the connectors run in serial order. If we use parallelisation instead, we can reduce the total execution time of the test to the longest test in the connector, being able to reduce it around 50% in some cases (like Postgres source).
<img width="1286" alt="Screenshot 2023-02-23 at 15 43 00" src="https://user-images.githubusercontent.com/42538006/220951352-dc37fbd6-f810-4d39-9c16-f7239ee1362f.png">

## Describe the solution you’d like
Execute test in a parallel manner.

sergio-ropero:
Instead of parallelise between different tests (`test`, `integration test`, `acceptance test`) we would parallelise the execution inside of each section.
The dependency between the tests are there because of a reason and we should not just remove it.

This PR improved execution time on each of the sections.
Before:
![image](https://user-images.githubusercontent.com/42538006/222172766-f6d3fb8c-b5df-46de-ac1e-b1e943dc85c0.png)

After:
![image](https://user-images.githubusercontent.com/42538006/222172801-bcc05901-cac6-481d-96f2-25f350b1b460.png)

Issue #23390, opened 2023-02-23
Remove duplicate execution on Postgres source tests


sergio-ropero:
## Tell us about the problem you're trying to solve
Based on issue [22968](https://github.com/airbytehq/airbyte/issues/22968)

Looking on the following image, we see that some of the test are executed twice (all tests in `test` are executed in `integrationTestJava`).
<img width="1032" alt="Screenshot 2023-02-23 at 15 40 44" src="https://user-images.githubusercontent.com/42538006/220938945-c49da81c-09cf-461d-b954-6370909c5c60.png">
This ends into a waste of time and resources that slow down Postgres velocity.

## Describe the solution you’d like
Remove redundancy on the test being executed, so `integrationTest` will not contain the tests that run under `test`.

prateekmukhedkar:
@sergio-ropero so the end result would be that `integrationTest` will not contain the tests that run under `test`?

PR #23506, opened 2023-02-27
feat: Set number of forks to execute tests in parallel

sergio-ropero:
## What
Based on the issues [23390](https://github.com/airbytehq/airbyte/issues/23390), the integration test executes already executed tests.
Also based on [23391](https://github.com/airbytehq/airbyte/issues/23391), test are run in serial.
The idea is to remove redundant tests and execute them in parallel to reduce total execution time.

## How
To fix the redundancy in the integration test, we just need to assign the test classes from the integration test folder, and not append them to the already tested tests.

To execute tests in parallel, modifying the variable `maxParallelForks` is enough (default is `1`). Modified for both, `test` and `integrationTest`.
The property `org.gradle.parallel=true` is already set up in `gradle.properties` file, but without `maxParallelForks` it just use one thread.
Also to mention, that this parallelise test internally, meaning that test inside `test` or `integration tests` are running in parallel, but not mixing them. It still has the dependency.

⚠️ This change impacts all connectors (test) and all Java connectors (integration tests) ⚠️ 

## Test
Before changes:
Screenshot 2023-02-28 at 15 30 00
Screenshot 2023-02-28 at 16 04 26

After changes:
Screenshot 2023-02-28 at 15 29 03
Screenshot 2023-02-28 at 16 03 48

## Future work
There are test that are not compatible to run in parallel and they randomly fail:
```
CdcWalLogsPostgresSourceDatatypeTest > testDataTypes() FAILED

[1097](https://github.com/airbytehq/airbyte/actions/runs/4292195225/jobs/7478361953#step:11:1099) org.opentest4j.AssertionFailedError: The streamer test_2_numeric should return all expected values. Missing values: [123, null, 1.2345678901234567E9] ==> expected: <true> but was: <false>
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -354,6 +354,8 @@ subprojects { subproj ->
     }
 
     test {
+        maxParallelForks = Runtime.runtime.availableProcessors() ?: 1
+
         jacoco {
             enabled = true
             excludes = ['**/*Test*', '**/generated*']
--- a/buildSrc/src/main/groovy/airbyte-integration-test-java.gradle
+++ b/buildSrc/src/main/groovy/airbyte-integration-test-java.gradle
@@ -22,7 +22,9 @@ class AirbyteIntegrationTestJavaPlugin implements Plugin<Project> {
         }
 
         project.task('integrationTestJava', type: Test) {
-            testClassesDirs += project.sourceSets.integrationTestJava.output.classesDirs
+            mustRunAfter project.test
+
+            testClassesDirs = project.sourceSets.integrationTestJava.output.classesDirs
             classpath += project.sourceSets.integrationTestJava.runtimeClasspath
 
             useJUnitPlatform {
@@ -37,15 +39,12 @@ class AirbyteIntegrationTestJavaPlugin implements Plugin<Project> {
                 // showStandardStreams = true
             }
 
-            outputs.upToDateWhen { false }
-
             if(project.hasProperty('airbyteDocker')) {
                 dependsOn project.airbyteDocker
             }
 
             maxHeapSize = '3g'
-
-            mustRunAfter project.test
+            maxParallelForks = Runtime.runtime.availableProcessors() ?: 1
 
             // This is needed to make the destination-snowflake tests succeed - https://github.com/snowflakedb/snowflake-jdbc/issues/589#issuecomment-983944767
             jvmArgs = ["--add-opens=java.base/java.nio=ALL-UNNAMED"]
--- a/buildSrc/src/main/groovy/airbyte-integration-test-java.gradle
+++ b/buildSrc/src/main/groovy/airbyte-integration-test-java.gradle
@@ -39,6 +39,8 @@ class AirbyteIntegrationTestJavaPlugin implements Plugin<Project> {
                 // showStandardStreams = true
             }
 
+            outputs.upToDateWhen { false }
+
             if(project.hasProperty('airbyteDocker')) {
                 dependsOn project.airbyteDocker
             }


sherifnada (approved):
Nice! Great find. 

You should probably get review from someone who writes java on a regular basis for knowledge sharing reasons but this LGTM!

buildSrc/src/main/groovy/airbyte-integration-test-java.gradle
```diff
@@ -44,8 +46,7 @@ class AirbyteIntegrationTestJavaPlugin implements Plugin<Project> {
             }
 
             maxHeapSize = '3g'
```
akashkulk, line 48: Not sure what the effect of this would be, but maybe we can bump the maxHeapSize to 4g as well?

akashkulk:
Great find- this is awesome! Quick question, why are is `CdcWalLogsPostgresSourceDatatypeTest` failing? want to make sure it's not a regression - is it passing on master?

build.gradle
```diff
@@ -354,6 +354,8 @@ subprojects { subproj ->
     }
 
     test {
+        maxParallelForks = Runtime.runtime.availableProcessors() ?: 1
```
davinchia, line 357: huh, I could have sworn this was already set from a while back

davinchia: from the documentation, maybe  `maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1` is better?

otherwise, if run locally, might suck up alot of a machine's resources.

either way easy to change, so not doing the divide is also good!

davinchia (approved):
Nice catch!

build.gradle
cpdeethree: are we sure this will actually make a difference? 

@git-phu and i played with this late last year and our essential takeaway was that gradle was already basically parallelizing as best as it could overall in builds (parallel behavior is set at the top level build.gradle alread)

buildSrc/src/main/groovy/airbyte-integration-test-java.gradle
cpdeethree: I'm not sure it will actually make a difference (probably depends on your test suite but 3g is already a decent chonker of a heap), but it probably wouldn't actively hurt anything either

build.gradle
cpdeethree: OH this is interesting. So apparently there are two levels of parallelism here with gradle. One is at the task level, and one is at the test level. The one sergio is modifying is at the test level, and the one @git-phu and I played with is the task level parallelism: https://discuss.gradle.org/t/relationship-between-forkevery-maxparallelfork-and-parallel/25126

davinchia: yeah! the [smallest unit of parallelism is still the test class](https://kousenit.org/2020/11/22/simple-demo-of-gradle-parallel-tests/) so for the platform build we'll probably need to do some work splitting up the bigger test files that exists.

actually @sergio-ropero probably a good idea to add a comment explaining this since I'd imagine other devs will have the same questions as conor

cpdeethree: one other somewhat interesting implication of this is that peter and i played with the max-workers number and it didn't seem to affect much back then. But I wonder whether playing with max-workers with forks in a test would actually make a difference after sergio's changes when targeting these specific builds.

buildSrc/src/main/groovy/airbyte-integration-test-java.gradle
sergio-ropero: Based in the test, we are using only 0'5g... so it's more than enough 🤣

sergio-ropero:
> Great find- this is awesome! Quick question, why are is `CdcWalLogsPostgresSourceDatatypeTest` failing? want to make sure it's not a regression - is it passing on master?

It failed randomly and my guess is that it's colliding with another test. At the end it works in a retry. I'm opening an issue to review it and identify if it's using a shared resource and change it if required.
Is not failing in master as there is executed in serial and shared resources are not modified by other tests.

build.gradle
sergio-ropero: > from the documentation, maybe maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1 is better?
> otherwise, if run locally, might suck up alot of a machine's resources.

To be honest I thought the same... probably we can add a logic if it's local execution or remote. For me, makes sense to use as much as possible of the remote instances... they just span to execute the test, so using half of them is not good.

I have tested in local and I have not feel like getting out of resources.
Also, to be hones, even if there were nodes available, not all of them were used. I assume that at some point, only a few classes runs in parallel.
Also, tests takes around 1 minute... so I think is fine

> actually @sergio-ropero probably a good idea to add a comment explaining this since I'd imagine other devs will have the same questions as conor

Yeah... At the beginning I was thinking about parallelise both test (unit test and integration test... even acceptance test) but as I saw in the gradle configuration files, there is an explicit dependency... so I assumed that running unit before integration was intended. Also, I think that makes sense, as usually integration tests are more time consuming, and if unit tests are wrong, there is no reason to run integration tests. Open to discuss it with you all 😊

jcowanpdx:
@sergio-ropero Follow up that this is affecting at least two destination connectors: redshift and snowflake. The reason is that we have limited integration testing resources in the various cloud environments for int testing these components so now that we are running in parallel, we are using more resources there. This is causing integration tests to intermittently fail due to hitting things like connection limits. I wonder if the "divide by two" recommendation is preferred or is there an easy way to override the max parallelism for these specific connectors? @ryankfu

Issue #22372, opened 2023-02-03
Add AllowedHosts to all Database Sources


evantahler:

DevPJ9:
Hello @evantahler I would like to work on this.

bleonard:
I seem to remember manybe we had an OC issue with allowed hosts a while back. Am I crazy @evantahler ? Just want to make sure not to mess it up.

evantahler:
There was a platform-level OC issue (https://github.com/airbytehq/airbyte/pull/21923) which was due to the weirdness that is a reset-connection. Adding these values for a connector, even if they are wrong, shouldn't break anything in production.

The path to ensuring your allowedHosts entries are properly interpolating config information is to look at the logs when the containers boot.  Look for a log line that interpolates the allowedHosts properly.  This is the failure case:

```
airbyte-worker | 2023-01-26 19:42:58 ERROR i.a.w.u.ConfigReplacer(getAllowedHosts):60 - The allowedHost value, '${host}', is expecting an interpolation value from the connector's configuration, but none is present
```

More information @ https://github.com/airbytehq/airbyte/pull/21950

prateekmukhedkar:
Work done: PR reviewed for non-GA source connectors.   
Work remaining:
1. Close out AlloyDB PR https://github.com/airbytehq/airbyte/pull/23596
2. Publish all new connectors

prateekmukhedkar:
Plan to publish all database connectors early next week.

prateekmukhedkar:
Done for all DB Sources **except** MongoDB, Redshift and Azure Table DB Source because of previously failing integration tests. Moving this to blocked/hold lane until we fix these tests and get them in a publishable state. Issue created for tracking publish https://github.com/airbytehq/airbyte/issues/24107

octavia-squidington-iii:
At Airbyte, we seek to be clear about the project priorities and roadmap. This issue has not had any activity for 365 days, suggesting that it's not as critical as others. It's possible it has already been fixed. It is being marked as stale and will be closed in 20 days if there is no activity. To keep it open, please comment to let us know why it is important to you and if it is still reproducible on recent versions of Airbyte.

PR #23596, opened 2023-02-28
Adding allowedHost for Postgres AlloyDB

prateekmukhedkar:
## What
This change adds `allowedHosts` for Postgres AlloyDB source. As AlloyDB is hosted on Google Cloud, we include a wildcard on known GCP domain names in the list. Issue https://github.com/airbytehq/airbyte/issues/22372

## How
For databases, this information is dynamic and read from the connector's config (shown for Postgres here):
https://github.com/airbytehq/airbyte/blob/e6441bf6be0ef931670e94e268d07b1e1465dee4/airbyte-config/init/src/main/resources/seed/source_definitions.yaml#L1370-L1374

Lines 1370 to 1374 in [e6441bf](https://github.com/airbytehq/airbyte/commit/e6441bf6be0ef931670e94e268d07b1e1465dee4)

 allowedHosts: 
   hosts: 
     - "${host}" 
     - "${tunnel_method.tunnel_host}"

## Recommended reading order
1. `source_definitions.yaml `


connectors.md
```diff
@@ -62,7 +62,7 @@
...
-| **DynamoDB** | <img alt="DynamoDB icon" src="https://raw.githubusercontent.com/airbytehq/airbyte/master/airbyte-config/init/src/main/resources/icons/dynamodb.svg" height="30" height="30"/> | Source | airbyte/source-dynamodb:0.1.1 | alpha | [link](https://docs.airbyte.com/integrations/sources/dynamodb) | [code](https://github.com/airbytehq/airbyte/tree/master/airbyte-integrations/connectors/source-dynamodb) | <small>`50401137-8871-4c5a-abb7-1f5fda35545a`</small> |
+| **DynamoDB** | <img alt="DynamoDB icon" src="https://raw.githubusercontent.com/airbytehq/airbyte/master/airbyte-config/init/src/main/resources/icons/dynamodb.svg" height="30" height="30"/> | Source | airbyte/source-dynamodb:0.1.2 | alpha | [link](https://docs.airbyte.com/integrations/sources/dynamodb) | [code](https://github.com/airbytehq/airbyte/tree/master/airbyte-integrations/connectors/source-dynamodb) | <small>`50401137-8871-4c5a-abb7-1f5fda35545a`</small> |
```
evantahler, line 65: Why did this change?

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
```diff
@@ -50,6 +50,12 @@
   icon: alloydb.svg
   sourceType: database
   releaseStage: generally_available
+  allowedHosts:
+    hosts:
+      - "${host}"
+      - "*.cloud.google.com"
+      - "*.googleapis.com"
```
evantahler, line 57: Why would we need these in addition to the provided host or SSL tunnel?

Screenshot 2023-02-28 at 2 19 17 PM

prateekmukhedkar: The thinking is that hosts under `cloud.google.com` and `googleapis.com` are needed to access Google Services such as authentication tokens and storage. We do not explicitly create connections to these 2 domains from `source-alloydb` so I am not able to confirm if they are actually needed or not.

Reference: https://cloud.google.com/apis/docs/overview

connectors.md
prateekmukhedkar: I think this happens if there were changes in master that I yet did not pull in my branch before sending the PR. After the update from master this file no longer shows as changed so closing this comment.

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
evantahler: Granting access to all of google cloud seems a bit broad - a shady connector could be uploading things to GCS for example, which is the kind of thing we are trying to block with these firewalls.

AlloyDB seems to be using the postgres protocol, which I think means that it should only need access to the database host?

prateekmukhedkar: Agree wildcards here are not ok, the bigger question is which domains do we need to whitelist access to when using cloud hosted databases? I will setup an AlloyDB instance to try this out.

By a shady connector, you mean in general, right? AlloyDB should go through a `db-sources` team review.

subodh1810 (changes requested):

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
prateekmukhedkar: ok, I tested with an AlloyDB database verified I was able to sync without needing wildcards on GCP domains. So I have removed them and the `allowedHost` property is the same postgres source. cc @evantahler @subodh1810 

`2023-03-03 06:02:48 [32mINFO[m i.a.w.p.DockerProcessFactory(create):128 - Creating docker container = source-alloydb-read-102-0-mfvec with resources io.airbyte.config.ResourceRequirements@5fd69a9e[cpuRequest=,cpuLimit=,memoryRequest=,memoryLimit=] and _allowedHosts io.airbyte.config.AllowedHosts@14134dca[hosts=[10.41.128.7, 34.123.104.114, *.datadoghq.com, *.datadoghq.eu, *.sentry.io],additionalProperties={}]_`

evantahler (approved):

subodh1810 (commented):
Need to pull in latest master. We have already released version 2.0.0 For AlloyDB

subodh1810 (approved):

prateekmukhedkar:
Currently blocked because of the `Connectors Base: Build (pull_request)` status check failing with `Execution failed for task ':airbyte-cdk:python:checkPython` which seems unrelated to the change in this PR.

PR #23610, opened 2023-03-01
Source Salesforce: change the sequence of requests

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1571

## How
Version 2.0.1 introduced new functionality in the [PR](https://github.com/airbytehq/airbyte/pull/22597). It allowed supporting non-bulk streams with a huge amount of properties that couldn't fit in a single http request because of max length. The idea was to split all the properties of an entity into chunks and make multiple requests to get the data: one request per chunk per page. Requests were made in a sequence:

Page 1
  - chunk 1 (select `id`, `a`, `b` from `table` order by `id`)
  - chunk 2 (select `id`, `c`, `d` from `table` order by `id`)

Page 2
  - chunk 1 (select `id`, `a`, `b` from `table` order by `id`)  # offset and limit are handled by the page number
  - chunk 2 (select `id`, `c`, `d` from `table` order by `id`)  # offset and limit are handled by the page number

Then parts of the records were sticked together by the primary key value within one page. This solution did not take into account that the Salesforce API does not apply a constant page size for all the queries. So when the first query returns 2k records, second returns 200 records, we have 1.8k incomplete records.

To fix this situation we change the sequence of requests. Since now each subsequent request will be made to retrieve data from the chunk that has the fewest number of records read from independent of pagination (although we store current and next page for each chunk). After sticking the record we may yield it right after we know it contains all the properties needed. This allows us supporting streams with a huge number of properties, avoiding high memory consumption and emitting consistent data.

Let's see an example. Let's say we have 6 properties split into 3 chunks.
1 request: select `id`, `a`, `b` from `table` order by `id` -> returns [{`id`: 1, `a`: 1, `b`: 1}, {`id`: 2, `a`: 2, `b`: 2}, {`id`: 3, `a`: 3, `b`: 3}, {`id`: 4, `a`: 4, `b`: 4}]
2 request: select `id`, `c`, `d` from `table` order by `id` -> returns [{`id`: 1, `c`: 1, `d`: 1}, {`id`: 2, `c`: 2, `d`: 2}] # salesforce decided to decrease page size compared to first request in favour of keeping high performance
3 request: select `id`, `e`, `f` from `table` order by `id` -> returns [{`id`: 1, `e`: 1, `f`: 1}] # page size decreased even more

Now we have 1 (id=1) complete record that can be emitted. Next steps are: find out which chunk has the fewest records (#3) and make request to fetch next page for it. When we have the response, we'll be able to emit 1+ or more records. Repeat until we have the full set of properties for each primary key.


clnoll:
@davydov-d nice!

To make sure I understand - when we query salesforce, we'll send a request for, e.g. properties `a` and `b`, and then another request for `c` and `d`, and only paginate once all properties have been requested?

Do we do any checks to make sure we have all properties? e.g. if the page with `c` and `d` has the fewest records and we use that for pagination, do we need to do anything special to make sure we also get `a` and `b`?

davydov-d:
> @davydov-d nice!
> 
> To make sure I understand - when we query salesforce, we'll send a request for, e.g. properties `a` and `b`, and then another request for `c` and `d`, and only paginate once all properties have been requested?
> 
> Do we do any checks to make sure we have all properties? e.g. if the page with `c` and `d` has the fewest records and we use that for pagination, do we need to do anything special to make sure we also get `a` and `b`?

@clnoll each set of properties (chunk) has its own pagination. If the `c` and `d` page has the fewest records (most probably it has the smallest page size) we will read another page of that chunk until another chunk has the fewest records or there's no next page for this one. This will be repeated until all the chunks finish their pagination. It's made to kind of balance chunks so we shouldn't have to keep all the partial records in memory and wait for the first chunk to go through all its pages, then the second and so on but could emit them earlier. This [line](https://github.com/airbytehq/airbyte/pull/23610/files#diff-0572edeedd561901f8110c0f88b48efde87d1ed91247c0eb189c7229da12090dR217) itself is actually a check that the record has all the properties.

Regarding your first question - first we'll request first pages of all the chunks, then the sequence of requests will depend on the page size of each chunk returned.

brianjlai:
I looked over the code trying to functionally understand what its doing. I think it makes sense and should perform the grouping as @davydov-d clearly explained in the PR description and comments. I also don't see how this would be affecting the abnormal sync behavior so it is a little bit puzzling why that test is starting to fail all of a sudden. I will continue reviewing the code itself tomorrow too

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -164,54 +197,67 @@ def _read_pages(
...
+                    records_by_primary_key[record_id] = (partial_record, counter)
+
+        # Process what's left.
+        # Because we make multiple calls to query N records (each call to fetch X properties of all the N records),
+        # there's a chance that the number of records corresponding to the query may change between the calls.
+        # Select 'a', 'b' from table order by pk -> returns records with ids `1`, `2`
+        #   <insert smth.>
+        # Select 'c', 'd' from table order by pk -> returns records with ids `1`, `3`
+        # Then records `2` and `3` would be incomplete.
+        # This may result in data inconsistency. We skip such records for now and log a warning message.
```
brianjlai, line 252: So now that we're able to stitch records together, is this comment still relevant? Do you have a sense for what might cause us to still have partial records. I think I agree with keeping this in for visibility just in case, but this does have me curious

brianjlai (approved):
The changes look good @davydov-d , nice work on what was a really challenging implementation of the grouping logic.

I pushed a fix for the infinite loop and consolidating the chunk states into a new object to clean the code up a bit (that was originally going to be my main suggestion during the review). If you want to look over those changes I made, otherwise I think this is in a good state to publish as long as we can get a passing CAT run.

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -155,6 +172,22 @@ def empty_props_with_pk_if_present():
...
+    def _next_chunk_id(property_chunks: Mapping[int, PropertyChunk]) -> Optional[int]:
+        """
+        Figure out which chunk is going to be read next.
+        It should be the one with the least number of records read by the moment.
+        """
+        non_exhausted_chunks = {
+            # We skip chunks that have already attempted a sync before and do not have a next page
+            chunk_id: property_chunk.record_counter
+            for chunk_id, property_chunk in property_chunks.items()
+            if property_chunk.first_time or property_chunk.next_page
```
brianjlai, line 185: Previously, we had a condition that checked if `records > 0`, which we used to indicate that this was the first sync attempt so we did not need to check for the next page token. However, this had the unintended consequence of looping infinitely if a stream did not have any records because it would be tricked into thinking it was still the first one. Replacing this check with a boolean field `first_time` for each chunk partition allows us to definitively tell that this is in fact the first time

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
davydov-d: Originally this was added to catch records that might be inserted between the two api calls. I dont think it's a common case, but we'd better have it to know if the number of such records becomes unreasonably high

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
davydov-d: @brianjlai thanks for the thorough investigation and the fix, I really missed that case. Changes look good to me and now the CAT run does not produce that many api calls, so I'm going to publish and release this

PR #23620, opened 2023-03-01
[ISSUE #22442] move success message from pypi publish to docker image…

maxi297:
… publish

## What
Following first release after https://github.com/airbytehq/airbyte/pull/23501, I saw two problems:
* bumpversion didn't work properly for the Dockerfile
* we should move the "success" Slack message to the publishing of the Docker image. 

## How
Add Dockerfile to commit when bumping version
Step `publish-cdk` only reports to Slack on error now. `publish-docker-image` reports on either success or failure

### DIFF
--- a/.github/workflows/publish-cdk-command-manually.yml
+++ b/.github/workflows/publish-cdk-command-manually.yml
@@ -83,7 +83,7 @@ jobs:
       - name: Commit and Push Changes
         uses: stefanzweifel/git-auto-commit-action@v4
         with:
-          file_pattern: airbyte-cdk/python/setup.py airbyte-cdk/python/.bumpversion.cfg airbyte-cdk/python/CHANGELOG.md
+          file_pattern: airbyte-cdk/python/setup.py airbyte-cdk/python/.bumpversion.cfg airbyte-cdk/python/CHANGELOG.md airbyte-cdk/python/Dockerfile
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-cdk-command-manually.yml
```diff
@@ -194,33 +194,6 @@ jobs:
         env:
           TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
           TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
-      - name: Post success to Slack channel dev-connectors-extensibility
```
clnoll, line 197: Do we not still want to post success to Slack on `publish-cdk`?

maxi297: That's a good question. For me now with the added publishing of the Docker image, I felt that a release is not a success until both of these steps were done. 

Example: Given a workflow that is able to push to pypi and not Dockerhub, I probably want just one message and it is that it failed. In the message, there is a link to the GitHub run and I can easily see which step failed. One thing that might be troublesome in that case is that the pipeline currently does not allow to re-run the pipeline without publishing to pypi so I'll have to bump the version again which is a little bit annoying...

clnoll (approved):
LGTM!

.github/workflows/publish-cdk-command-manually.yml
clnoll: 👍 makes sense to me.

PR #23627, opened 2023-03-01
Removing unused gradle task for common build

maxi297:
## What
Fix following build failure on master associated to https://github.com/airbytehq/airbyte/pull/23411

## How
Some gradle tasks were removed during monorepo split which caused some issues


clnoll (approved):
LGTM!

PR #23646, opened 2023-03-01
Makes docker script more linux happy

supertopher:
## What
linux `which` isn't down with -s.  this changes makes the same functional check in a more agnostic way

### DIFF
--- a/run-ab-platform.sh
+++ b/run-ab-platform.sh
@@ -134,7 +134,7 @@ if test $(tput cols) -ge 64; then
 fi
 
 ########## Dependency Check ##########
-if ! which -s docker-compose; then
+if ! which docker-compose > /dev/null; then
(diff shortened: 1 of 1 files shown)


supertopher:
Screenshot 2023-03-01 at 1 13 57 PM
tested very briefly locally

git-phu (approved):

run-ab-platform.sh
```diff
@@ -134,7 +134,7 @@ if test $(tput cols) -ge 64; then
 fi
 
 ########## Dependency Check ##########
-if ! which -s docker-compose; then
+if ! which docker-compose > /dev/null; then
```
ambirdsall, line 137: This is good, but this would be even better:
```suggestion
if ! command -v docker-compose &> /dev/null; then
```
`command` is a POSIX-specified builtin command, so it's both faster (no need to launch a whole external process) and more portable.

PR #23651, opened 2023-03-01
Fixing SurveyMonkey collection survey_questions

jman0119:
Fixing SurveyMonkey collection survey_questions


sherifnada (changes requested):
@jman0119 could you fix the git diff please?

PR #23655, opened 2023-03-02
Remove ExtendedNameTrasformer

sherifnada:
## What
Removes the `ExtendedNameTransformer` class because it was basically completely unused. It also replaces all usages of it with its superclass `StandardNameTransformer`

## Recommended reading order
1. `ExtendedNameTransformer.java`
2. everything else

## 🚨 User Impact 🚨
There should be zero user impact since this refactor just removes code which was completely unused 

I think we don't need to republish since this is a refactor with no behavior change


ryankfu (approved):
Changes look good, pending passing GA connectors integration test since changes will likely not get surfaced until these connectors get published so tests should capture any non deterministic changes

PR #23656, opened 2023-03-02
Add speed benchmark to e2e test source

cgardens:
## What

Adds a speed benchmark mode to the e2e-test source. The goal here is have a _deterministic_ way of generating data for speed benchmarks. This generator can be used to produce data quickly on the fly in the source _and_ to produce files with the _same_ data to be used in other data sources (e.g. postgres and S3). This generator is what underlies our "bottleneck detector" for trying to determine where the speed bottlenecks are in our system.

In addition, it fixes the `source-e2e-test-cloud`. Looks like there was a bug introduced when trying to filter the spec for the cloud version that killed `source-e2e-test`. This fixes that. This does result in a change in the spec of `source-e2e-test-cloud` _but_ I wrote a test to confirm that configs made using the old spec are still valid against the new spec. You can check out that test in this commit: https://github.com/airbytehq/airbyte/pull/23656/commits/285a10124d61cf92f743fd9f23e368f7bb83560d . 

Finally, `source-e2e-test-cloud` 2.1.1 (the one that is currently in production now) is broken. It fails discover schema. I think this is because it's technically not outputting a valid catalog (though I'm not entirely sure why that's the case. Either way, I have tested the new version in this PR (version 2.1.1) and verified locally that discover _does_ work (and also sanity checked that 2.1.1 doesn't work locally). So This should get us back to having a working source here. 😄 

---
More details on the discover error that is now resolved....

Screen Shot 2023-03-01 at 8 26 23 PM
```
airbyte-worker                    | 2023-03-02 04:52:34 ERROR i.a.w.i.DefaultAirbyteStreamFactory(validate):134 - Validation failed: {"type":"CATALOG","catalog":{"streams":[{"name":"data_stream","json_schema":{"type":"object","properties":{"column1":{"type":"string"}}},"supported_sync_modes":[],"default_cursor_field":[],"source_defined_primary_key":[]}]}}
airbyte-worker                    | 2023-03-02 04:52:34 INFO i.a.w.i.DefaultAirbyteStreamFactory(parseJson):125 - 2023-03-02 04:52:34 INFO i.a.i.b.IntegrationRunner(runInternal):171 - Completed integration: io.airbyte.integrations.source.e2e_test.CloudTestingSources
airbyte-worker                    | 2023-03-02 04:52:34 INFO i.a.w.t.TemporalAttemptExecution(lambda$getWorkerThread$5):198 - Completing future exceptionally...
airbyte-worker                    | io.airbyte.workers.exception.WorkerException: Integration failed to output a catalog struct and did not output a failure reason
```

old catalog (not working):
```
{"type":"CATALOG","catalog":{"streams":[{"name":"data_stream","json_schema":{"type":"object","properties":{"column1":{"type":"string"}}},"supported_sync_modes":[],"default_cursor_field":[],"source_defined_primary_key":[]}]}}
```
new catalog (working):
```
{"type":"CATALOG","catalog":{"streams":[{"name":"data_stream","json_schema":{"type":"object","properties":{"column1":{"type":"string"}}},"supported_sync_modes":["full_refresh"],"default_cursor_field":[],"source_defined_primary_key":[]}]}}
```


cgardens:
@evantahler The failure in the integration test is because the icon doesn't exist. The icon doesn't exist because this source isn't in the OSS catalog, so I think it is reasonable to ignore this test. Does that make sense? Will publishing work without this test passing?

evantahler:
... what's the difference between source-e2e-test and source-e2e-test-cloud?

docs/integrations/sources/e2e-test-cloud.md
```diff
@@ -0,0 +1,34 @@
...
+|                   | random seed         | integer | no       | current time millis | The seed is used in random Json object generation. Min 0. Max 1 million.                                                                                |
+|                   | message interval    | integer | no       | 0                   | The time interval between messages in millisecond. Min 0 ms. Max 60000 ms (1 minute).                                                                   |
+
+## Changelog
+
+The OSS and Cloud variants have the same version number. The Cloud variant was initially released at version `1.0.0`.
+
+| Version | Date       | Pull request                                             | Notes                                               |
+|---------|------------|----------------------------------------------------------|-----------------------------------------------------|
+| 2.1.2   | 2023-03-01 | [23656](https://github.com/airbytehq/airbyte/pull/23656) | Fix inheritance between e2e-test and e2e-test-cloud |
```
evantahler, line 33: [nit] I view this sort of like a strict-encrypt connector. Ideally the versions would match.
```suggestion
| 2.1.4   | 2023-03-01 | [23656](https://github.com/airbytehq/airbyte/pull/23656) | Fix inheritance between e2e-test and e2e-test-cloud |
```

airbyte-integrations/connectors/source-e2e-test-cloud/Dockerfile
```diff
@@ -17,5 +17,5 @@ ENV ENABLE_SENTRY true
 
 COPY --from=build /airbyte /airbyte
 
-LABEL io.airbyte.version=2.1.1
+LABEL io.airbyte.version=2.1.2
```
evantahler, line 20:
```suggestion
LABEL io.airbyte.version=2.1.4
```

evantahler (approved):

docs/integrations/sources/e2e-test-cloud.md
cgardens: 2.1.2 is right. this is for e2e-test-cloud not e2e-test. the latter is the one a 2.1.4.

cgardens: or is your goal to just have e2e-tst-cloud and e2e-test always have the same versions?

davinchia (approved):
This is a great idea!

I'm going to build on this in a follow up PR to add fat records.

docs/integrations/sources/e2e-test-cloud.md
```diff
@@ -0,0 +1,34 @@
+# End-to-End Testing Source for Cloud
+
+## Overview
+
+This is a mock source for testing the Airbyte pipeline. It can generate arbitrary data streams. It is a subset of what is in [End-to-End Testing Source](e2e-test.md) in Open-Source to avoid Clouder users accidentally in curring a huge bill.
```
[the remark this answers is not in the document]
cgardens: thanks!

docs/integrations/sources/e2e-test-cloud.md
cgardens: works for me.

airbyte-integrations/connectors/source-e2e-test-cloud/Dockerfile
cgardens: cool.

cgardens:
@evantahler did i do the publish incantation wrong? i thought this was right but looked like nothing triggered. ^

evantahler:
Seems like Github Actions are back - I'll re-trigger the publishes

PR #23671, opened 2023-03-02
Source Instagram: decrease give up rate

darynaishchenko:
Give up reason: the endpoint requires the pages_read_engagement permission. For now we haven't an access to our test account to test these permissions(add/remove). See [here](https://airbytehq.slack.com/archives/C02U9R3AF37/p1677748264473119) for more info.
**Phase 1**, in this PR was added info about main permissions in spec and added doc links in error message to navigate user.


lazebnyi (approved):

airbyte-integrations/connectors/source-instagram/source_instagram/api.py
```diff
@@ -88,7 +88,13 @@ def _find_accounts(self) -> List[Mapping[str, Any]]:
                         }
                     )
         except FacebookRequestError as exc:
```
erohmensing, line 90: Do all `FacebookRequestError`s we might run into here pertain to permissions issues? Is there something more specific we could except in order to raise this specific error message in order to avoid accidentally masking other issues as permission issues?

erohmensing (approved):
One question, but 👍🏻

airbyte-integrations/connectors/source-instagram/source_instagram/api.py
darynaishchenko: thanks for commenting, I added improvement for this code, now we checking the error status code and if this code rated to permissions we also say about checking permission needed for our connector.

erohmensing: Awesome, thanks!

erohmensing (approved):

PR #23681, opened 2023-03-02
Connector Acceptance Test: Fix Test Discovery backward_compatibility

artem1205:
## What
Fix Test Discovery `backward_compatibility` fixture.

In `test_backward_compatibility` we call `skip_backward_compatibility_tests_fixture`, which, in turn, calls another fixtures:

1. `inputs`
2. `previous_connector_docker_runner`
3. `discovered_catalog`
4. `previous_discovered_catalog`

In my case, call of `discovered_catalog` and `previous_discovered_catalog` with the same config file will raise an error due to schema error. 

## How

To bypass it, I divided this fixture into 2 to check input version before catalogs comparison:

1. Check flag from provided configuration
2. Check and compare catalogs

In this case, we will not face error in second fixture if we have a flag in first one.

### DIFF
--- a/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
+++ b/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
@@ -501,20 +501,25 @@ class TestDiscovery(BaseTest):
     def skip_backward_compatibility_tests_fixture(
         self, inputs: DiscoveryTestConfig, previous_connector_docker_runner: ConnectorRunner
     ) -> bool:
+        if previous_connector_docker_runner is None:
+            pytest.skip("The previous connector image could not be retrieved.")
+
         # Get the real connector version in case 'latest' is used in the config:
         previous_connector_version = previous_connector_docker_runner._image.labels.get("io.airbyte.version")
 
         if previous_connector_version == inputs.backward_compatibility_tests_config.disable_for_version:
             pytest.skip(f"Backward compatibility tests are disabled for version {previous_connector_version}.")
 
-        def test_catalog_compatibility(
-            discovered_catalog: MutableMapping[str, AirbyteStream], previous_discovered_catalog: MutableMapping[str, AirbyteStream]
-        ):
-            if discovered_catalog == previous_discovered_catalog:
-                pytest.skip("The previous and actual discovered catalogs are identical.")
+        return False
 
-            if previous_connector_docker_runner is None:
-                pytest.skip("The previous connector image could not be retrieved.")
+    @pytest.fixture(name="skip_backward_catalog_tests")
+    def skip_backward_catalog_fixture(
+            self,
+            discovered_catalog: MutableMapping[str, AirbyteStream],
+            previous_discovered_catalog: MutableMapping[str, AirbyteStream]
+    ) -> bool:
+        if discovered_catalog == previous_discovered_catalog:
+            pytest.skip("The previous and actual discovered catalogs are identical.")
 
         return False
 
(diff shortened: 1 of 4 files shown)


lazebnyi (approved):

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
```diff
@@ -499,23 +499,23 @@ def test_check(self, connector_config, inputs: ConnectionTestConfig, docker_runn
...
-        if previous_connector_docker_runner is None:
-            pytest.skip("The previous connector image could not be retrieved.")
-
         # Get the real connector version in case 'latest' is used in the config:
         previous_connector_version = previous_connector_docker_runner._image.labels.get("io.airbyte.version")
 
         if previous_connector_version == inputs.backward_compatibility_tests_config.disable_for_version:
             pytest.skip(f"Backward compatibility tests are disabled for version {previous_connector_version}.")
+
+        def test_catalog_compatibility(
```
davydov-d, line 510: where is this function called?

artem1205: updated

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
```diff
@@ -516,6 +509,16 @@ def skip_backward_compatibility_tests_fixture(
 
         if previous_connector_version == inputs.backward_compatibility_tests_config.disable_for_version:
             pytest.skip(f"Backward compatibility tests are disabled for version {previous_connector_version}.")
+
+        return False
+
+    @pytest.fixture(name="skip_backward_catalog_tests")
+    def skip_backward_catalog_fixture(
```
davydov-d, line 516: I'd suggest giving this routine a more meaningful name, otherwise looks good to me. But let's also wait for operations team to review

davydov-d (approved):

alafanechere (changes requested):
@artem1205 I look like you spotted an edge case but I'm not sure what it is exactly and how you code fixes it.
Could you please write a test that reproduce the edge case and assess the correct behavior is now happening? 
You mention spec change in the PR description and made change to discovery/catalog tests, this is why I'm a bit puzzled.

artem1205:
@alafanechere, the problem is not in tests itself, but in **fixtures**.

In `test_backward_compatibility` we call `skip_backward_compatibility_tests_fixture`, which, in turn, calls another fixtures:

1. `inputs`
2. `previous_connector_docker_runner`
3. `discovered_catalog`
4. `previous_discovered_catalog`

In my case, call of `discovered_catalog` and `previous_discovered_catalog` with the same config file will raise an error due to schema error. To bypass it, I divided this fixture into 2 to check input version **before** catalogs comparison:

1. Check flag from provided configuration
2. Check and compare catalogs

In this case, we will not face error in second fixture if we have a flag in first one.

alafanechere:
@artem1205 could you please provide an example of a context that leads to an error?
And it would help if you would add a test (of the tests) to make sure this behavior is not happening after your fix.
I'd like to make sure I understand this change.

artem1205:
> @artem1205 could you please provide an example of a context that leads to an error? And it would help if you would add a test (of the tests) to make sure this behavior is not happening after your fix. I'd like to make sure I understand this change.

@alafanechere , as an example, where it fails: https://github.com/airbytehq/airbyte/actions/runs/4347181053/jobs/7594196845#step:11:1882

link to simplified version of test(another branch ):
https://github.com/airbytehq/airbyte/commit/bb512148d44021f4ef45499a16317070f49b3b9d

davydov-d:
hey @alafanechere did you have a chance to review this PR? does it need some additional work to do? It seems like I'm blocked by this issue as well so looking forward to merge it

alafanechere:
@artem1205 @davydov-d I merged and published a new CAT version in https://github.com/airbytehq/airbyte/pull/23992 
Please open an issue if your error is still not resolved with links to PR where you're stuck. Thanks!

PR #23690, opened 2023-03-02
Low CDK: duration macro added

grubberr:
## What
Converts ISO8601 duration to datetime.timedelta

Usage:
`"{{ now_utc() - duration('P1D') }}"`


girarda (approved):
I like it! :shipit:

PR #23705, opened 2023-03-02
Source Pipedrive: disable oAuth

grubberr:
## What

- Remove oAuth from connector spec:
> OAuth for Pipedrive is not allowed only if we want to have an app on their marketplace.
> We’ve been asking them if we can enable OAuth for our users and not be on the marketplace, but no answers so far.

Screenshot 2023-03-03 at 09 32 49


airbyte-integrations/connectors/source-pipedrive/source_pipedrive/spec.json
```diff
@@ -12,6 +12,24 @@
...
+                "title": "API Token",
+                "type": "string",
+                "description": "The Pipedrive API Token.",
+                "airbyte_secret": true
+              }
+            }
+          },
           {
             "title": "Sign in via Pipedrive (OAuth)",
             "type": "object",
```
evantahler, line 35: Shouldn't this be removed as well?

evantahler (changes requested):

erohmensing (commented):
Only one person has set up Pipedrive (other than Yowan) so far, and with an API token 👍🏻 

Will they have to re-auth if we change the spec? Probably fine if it's an alpha connector but just want to check in any case.

airbyte-integrations/connectors/source-pipedrive/source_pipedrive/spec.json
grubberr: I have removed "Sign in via Pipedrive (OAuth)"

grubberr:
> Only one person has set up Pipedrive (other than Yowan) so far, and with an API token 👍🏻
> 
> Will they have to re-auth if we change the spec? Probably fine if it's an alpha connector but just want to check in any case.

Yes it has to work

girarda (approved):
looks good assuming OAuth needs to be disabled on OSS too

evantahler (approved):

PR #23711, opened 2023-03-03
Connector Health: Fixed SAT for Instagram, Facebook pages, Google Search Console, Stripe, Typeform, Hubspot, Sentry, Faker, Zendesk support, Slack, Jira

lazebnyi:
## What
Fixed SAT for Instagram, Facebook pages, Google Search Console, Stripe, Typeform, Faker, Zendesk support, Slack, Jira

## How
Updated expected records. For Facebook pages updated ignored field in full refresh test to new format.


airbyte-integrations/connectors/source-google-search-console/acceptance-test-config.yml
```diff
@@ -38,11 +44,3 @@ acceptance_tests:
         timeout_seconds: 1800
         future_state:
           future_state_path: "integration_tests/abnormal_state.json"
-        cursor_paths:
```
davydov-d, line 41: is this change intentional? what's the reason?

lazebnyi: No. Cursor fields already present in `configured_catalog_incremental.json`.

midavadim (approved):

airbyte-integrations/connectors/source-google-search-console/acceptance-test-config.yml
davydov-d: Those are different. Here we should describe a path to a cursor value in a state, not in a record. Like [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-google-search-console/integration_tests/abnormal_state.json) and [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-google-search-console/source_google_search_console/streams.py#L240). If this change fixes the build, I'm afraid it is false positive

lazebnyi: I'm not sure because cursors for incremental tests generated here - https://github.com/airbytehq/airbyte/blob/787b2c33d7586c9a3401dd72e4fdbc2c4ccbf80c/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py#L61

davydov-d: That's right. It falls back to default value (what is in `configured_catalog_incremental.json`) if you dont set it in the acceptance-test-config.yml manually

airbyte-integrations/connectors/source-sentry/acceptance-test-config.yml
```diff
@@ -8,6 +8,19 @@ acceptance_tests:
...
+          bypass_reason: "Order features return randomly"
+        - name: organization/features
+          bypass_reason: "Order features return randomly"
+        - name: plugins/*/features
+          bypass_reason: "Order features return randomly"
+        projects:
+          - name: features
+            bypass_reason: "Order features return randomly"
+          - name: organization/features
+            bypass_reason: "Order features return randomly"
```
evantahler, line 23: For the random ordering-type problems.  If it isn't a crazy ask, I think it would be better to sort the record properties ourself rather than skip testing that stream or property.

lazebnyi: I was create follow up issue https://github.com/airbytehq/airbyte/issues/23860 to unblock this PR.

PR #23723, opened 2023-03-03
Source Files: decrease give up rate

darynaishchenko:
Updated description in spec, made more user-friendly error messages and docs.


lazebnyi (approved):

airbyte-integrations/connectors/source-file/source_file/source.py
```diff
@@ -113,13 +113,12 @@ def check(self, logger, config: Mapping) -> AirbyteConnectionStatus:
...
             reason = (
-                f"Failed to load {source_url}\n Please check File Format and Reader Options are set correctly"
-                f"\n{repr(err)}\n{traceback.format_exc()}"
+                f"Failed to load {source_url}\n Please check File Format and Reader Options are set correctly. \n{repr(err)}"
             )
             logger.error(reason)
             return AirbyteConnectionStatus(status=Status.FAILED, message=reason)
         except Exception as err:
-            reason = f"Failed to load {source_url}: {repr(err)}\n{traceback.format_exc()}"
+            reason = f"Failed to load {source_url}. Maybe you provided invalid url, please verify it: {repr(err)}."
```
alafanechere, line 121:
```suggestion
            reason = f"Failed to load {source_url}. You could have provided an invalid URL, please verify it: {repr(err)}."
```

airbyte-integrations/connectors/source-file/source_file/spec.json
```diff
@@ -33,12 +33,13 @@
...
         "title": "Reader Options",
         "description": "This should be a string in JSON format. It depends on the chosen file format to provide additional options and tune its behavior.",
-        "examples": ["{}", "{\"sep\": \" \"}"]
+        "examples": ["{}", "{\"sep\": \" \"}", { "sep" : "\t", "header" : 0, "names": "column1, column2"}]
       },
       "url": {
         "type": "string",
         "title": "URL",
-        "description": "The URL path to access the file which should be replicated."
+        "description": "The URL path to access the file which should be replicated. The format for that is https://drive.google.com/uc?export=download&id=[DRIVE_FILE_ID] where [DRIVE_FILE_ID] is the string found in the Share URL here https://drive.google.com/file/d/[DRIVE_FILE_ID]/view?usp=sharing",
```
alafanechere, line 41: I don't really get the example here. Isn't it a simple URL and not a portion of it as you mentioned in the example?

alafanechere (changes requested):
I don't really get the description of the url field 😄

airbyte-integrations/connectors/source-file/source_file/source.py
darynaishchenko: changed

airbyte-integrations/connectors/source-file/source_file/spec.json
darynaishchenko: thanks, I removed my new description, because it was for [Google Drive case](https://docs.airbyte.com/integrations/sources/file#getting-started), and added examples from [this table](https://docs.airbyte.com/integrations/sources/file#examples).

alafanechere (approved):
Please consider my latest suggestion. Approving to not block.

lazebnyi (approved):

PR #23731, opened 2023-03-03
qa-engine: exclude destination weviate and exasol

alafanechere:
## What
For safety, we want to not mark the destinations without a strict encrypt variant available on cloud.
This PR explicitly excludes `destination-weviate` and `destination-exasol`.


evantahler (commented):
Process question:

Should the connectors be excluded here, or should they be added to cloud, and commented out like this:  https://github.com/airbytehq/airbyte-platform-internal/blob/master/cloud-config/cloud-config-seed/src/main/resources/seed/source_definitions_mask.yaml#L49-L50

I think the cloud_mask file is more discoverable, but 🤷.  I'm more interested in having one place this data lives.

alafanechere:
> Process question:
> 
> 
> 
> Should the connectors be excluded here, or should they be added to cloud, and commented out like this:  https://github.com/airbytehq/airbyte-platform-internal/blob/master/cloud-config/cloud-config-seed/src/main/resources/seed/source_definitions_mask.yaml#L49-L50
> 
> 
> 
> I think the cloud_mask file is more discoverable, but 🤷.  I'm more interested in having one place this data lives.  

I don't have a good answer for this. I think it logically makes sense to define these connectors on the oss repo side instead of making Cloud declaring which connector it does not want.

evantahler:
ok, let's keep these things in OSS then.

evantahler (approved):

PR #23740, opened 2023-03-03
Add connector ops' required review when changing bypass_reasons on GA connectors

erohmensing:
## What
1. Require a review from connector ops when bypass_reasons are changed/added on GA connectors only
2. Refactor test of tests a bit

## How
1. Get a list of connectors which have bypass_reason changes, cross reference against the list of ga connectors that were changed. In order to do this, update the `find_changed_ga_connectors` method to return set(str) like the other checks. I believe this should be fine since it's not used anywhere else and we don't expose it as a console script in setup.py.
2. Make all unit testing of the connector acceptance test run the same `check_review_requirements_file` check, which will verify whether the requirements file was created as well as if it contains the right reviewers. 

## Recommended reading order
1. tools/ci_connector_ops/ci_connector_ops/acceptance_test_config_checks.py
2. tools/ci_connector_ops/tests/test_acceptance_test_config_checks.py

### DIFF
--- a/tools/ci_connector_ops/README.md
+++ b/tools/ci_connector_ops/README.md
@@ -19,3 +19,17 @@ pip install -e . # assuming you are in the ./tools/ci_connector_ops directory
 ```
 
 pip will make binaries for all the commands in setup.py, so you can run `allowed-hosts-checks` directly from the virtual-env
+
+## Testing Locally
+
+To install requirements to run unit tests, use:
+
+```
+pip install -e ".[tests]"
+```
+
+Unit tests are currently configured to be run from the base `airbyte` directory. You can run the tests from that directory with the following command:
+
+```
+pytest -s tools/ci_connector_ops/tests
+```
--- a/tools/ci_connector_ops/ci_connector_ops/acceptance_test_config_checks.py
+++ b/tools/ci_connector_ops/ci_connector_ops/acceptance_test_config_checks.py
@@ -53,10 +53,13 @@ def find_mandatory_reviewers() -> List[Union[str, Dict[str, List]]]:
     ga_connector_changes = find_changed_ga_connectors()
     backward_compatibility_changes = utils.get_changed_acceptance_test_config(diff_regex="disable_for_version")
     test_strictness_level_changes = utils.get_changed_acceptance_test_config(diff_regex="test_strictness_level")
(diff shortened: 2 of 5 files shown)


tools/ci_connector_ops/README.md
```diff
@@ -19,3 +19,17 @@ pip install -e . # assuming you are in the ./tools/ci_connector_ops directory
...
+
+```
+pip install -e ".[tests]"
+```
+
+Unit tests are currently configured to be run from the base `airbyte` directory. You can run the tests from that directory with the following command:
+
+```
+pytest -s tools/ci_connector_ops/tests
+```
```
alafanechere, line 35: 👍

tools/ci_connector_ops/ci_connector_ops/acceptance_test_config_checks.py
```diff
@@ -39,23 +40,36 @@ def find_connectors_with_bad_strictness_level() -> List[utils.Connector]:
...
+def find_changed_ga_connectors() -> Set[str]:
     """Find GA connectors modified on the current branch.
 
     Returns:
-        List[utils.Connector]: The list of GA connectors that were modified on the current branch.
+        Set[str]: Set of connector names e.g {"source-github"}: The set of GA connectors that were modified on the current branch.
     """
     changed_connectors = utils.get_changed_connectors()
-    return [connector for connector in changed_connectors if connector.release_stage == "generally_available"]
+    return {connector.technical_name for connector in changed_connectors if connector.release_stage == "generally_available"}
```
alafanechere, line 50: Could we make it continue to return and iterable of  `Connector`? This function might need to be reused elsewhere and will benefit from getting a full connector object

alafanechere (approved):
LGTM, thank you, glad it was quite an incremental change.
I just left a minor request.
Please don't forget to bump the setup.py version (even if it's not used at all atm 😄 )

tools/ci_connector_ops/ci_connector_ops/acceptance_test_config_checks.py
erohmensing: Sure, I'll either make the regex one also return `Connector` objects, or will parse the `.technical_name` on the other side before comparing 👍🏻

erohmensing: You're right, this was the better option! Done :) (made everything return Connector objects)

PR #23759, opened 2023-03-06
:bug: Source File: fix column data types for numerical values

### LINKED ISSUE #23683: Source File: integer column is given as string for large files.
## Environment
- **Source Connector and version**:airbyte/source-file:0.2.33
- **Step where error happened**: discover source

## Current Behavior
When we have very large file with over 10k records in csv, integer column for csv file is depicted as string where all columns are integer in my file. This works when we have file with less than 10k records.

## Expected Behavior
All number columns should be number.

## Steps to Reproduce
1. Run discover for csv file over 10k records with column as integer

##Solution
image
 this line here will never get true for subsequent batch run and fallback to string.
yes

Alihassanc5:
Closes #23683 

## What
Number columns are sometimes given as strings for larger files. If the number of records in a file exceeds the default limit of CSV_CHUNK_SIZE, the type of the number column is changed to a string column.

## How
The current_type check for number columns has been fixed to use 'number' instead of checking the current type of the column by data types such as (int64, float64). This change ensures that the current type of the column is checked based on its actual type, rather than relying on the underlying data types of its values.


airbyte-integrations/connectors/source-file/source_file/client.py
```diff
@@ -350,7 +350,7 @@ def dtype_to_json_type(current_type: str, dtype) -> str:
             return current_type
         if dtype == object:
             return "string"
-        if dtype in number_types and (not current_type or current_type in number_types):
+        if dtype in number_types and (not current_type or current_type == "number"):
```
marcosmarxm, line 353: @Alihassanc5 I think the changes you're submitting are not improving the connector.
The variable `number_types = ("int64", "float64")` already transform int and floats. Those are dtypes from pandas.

marcosmarxm (changes requested):
Left a comment. Are you able to run the local tests to validate your change?

airbyte-integrations/connectors/source-file/source_file/client.py
Alihassanc5: @marcosmarxm Sorry for the confusion, let's iterate this code for 2 chunks of CSV file.

During the first chunk, the current_type is None, and the dtype is "int64". As expected, the code correctly identifies this as a "number" type.

However, during the second chunk, the current_type is "number", but the dtype is still "int64". At this point, the code incorrectly compares the current_type with "number_types" instead of previous column type "number". This leads to the code setting the column type to "string" instead of "integer" for over 10,000 records.

To fix this issue, the code needs to be modified to compare the current_type with "number" instead of "number_types". This will ensure that the column type is correctly identified as an integer for all integer values in the CSV file.

marcosmarxm: @Alihassanc5 Can you add a unit test to validate the change?

marcosmarxm (changes requested):

airbyte-integrations/connectors/source-file/source_file/client.py
Alihassanc5: @marcosmarxm I have updated the unit test to validate this change.
current_type, d_type, expected

("number", "int64", "number"),
("number", "float64", "number"),

Alihassanc5: @marcosmarxm any update regarding this PR?

marcosmarxm:
@Alihassanc5 adding to my to-do list for this week

marcosmarxm (approved):

arsenlosenko (approved):

marcosmarxm:
@Alihassanc5 please grant permission to Airbyte maintainers to edit your branch. Instructions here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

Alihassanc5:
@marcosmarxm @arsenlosenko there is an issue to give access on cross organizations. 
Issue: https://github.com/orgs/community/discussions/5634
 
I added both of you in my repository, you should have an access to edit my branch now.

PR #23777, opened 2023-03-06
Create command stub

rodireich:
## What
Create `/test-connector-performance` command stub.
This is needed as peter-evans/slash-command-dispatch seems to be looking at master's slash-command.yml when dispatching commands.


davinchia (approved):

octavia-approvington:

octavia-approvington (approved):

PR #23796, opened 2023-03-06
[Low-Code CDK] Enforce manifest against the airbyte-cdk version and the Beta version 0.29.0

### LINKED ISSUE #22668: [Low-Code CDK] Connector commands should fail when the low-code manifest version does not match the current airbyte-cdk version
## Problem

Source connectors are versioned independently from one another using the version specified in setup.py A simple way of ensuring that the defined manifest is always running on a compatible version of the low-code framework version is to compare them against each other when processing a connector operation. At runtime, after parsing the manifest, we can compare the airbyte-cdk major version (which adheres to semantic versioning) against the one defined on the manifest’s version field.

## Solution

Low-code manifests specify under the `version` field what major version of the low code CDK is being run. When attempting to parse a declarative low-code manifest, we should verify that the major version matches the version of the `airbyte-cdk` package that is being run. The reason why we are gating on major version is because in a low-code connectors `setup.py` we pin to the major but get the latest minor + patch at all times which fits with the paradigm of a specified major version in the manifest.

If the major versions do not match, then we should throw an error.

We can use the `from importlib.metadata import version` which is a built in Python library to derive the running version because we're on Python 3.9.

## Acceptance Criteria

- [ ] Running a connector whose major  matches the version specified in `setup.py` should run `spec` + `read` correctly
- [ ] Runnng a connect whose major version specified in `setup.py` should fail w/ an error when `spec` is run
- [ ] The flow should work when running a sync where the connector is build against the local version of the CDK (though this should already work)

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/22668

## What
For the early access of the low-code framework, version was just a place holder. However, now that we instituted breaking changes in promoting the language to Beta in `0.29.0`, we should start explicitly verifying that the version specified in the manifest is compatible with the version of `airbyte-cdk` that a low-code connector is using at runtime. Most connectors only pin `airbyte-cdk` to a major version, so we don't usually know the minor version until runtime. At a high level, this PR adds the following validations:

1. The manifest version should never be higher than the CDK version
2. Manifest versions lower than 0.29.0 should only be running a CDK version lower than 0.29.0

## How
We can retrieve the current version of `airbyte-cdk` at runtime using a built-in Python helper function. Versions in the manifest and using this function are represented as string. We first parse these versions into their respective major, minor, and patch versions and throw any errors if they are ill-formatted. Then we perform some assertions against each other and against the Beta anchor version `0.29.0`.

We technically could have skipped the parsing to integers altogether for simplicity, but the string comparison has some little edge cases that are less obvious especially for invalid manifest version formats. Parsing first and performing the more granular checks leaves less room for error.

## Recommended reading order
2. `test_manifest_declarative_source.py`


airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -142,6 +144,33 @@ def _validate_source(self):
         except ValidationError as e:
             raise ValidationError("Validation against json schema defined in declarative_component_schema.yaml schema failed") from e
 
+        cdk_version = metadata.version("airbyte_cdk")
+        cdk_major, cdk_minor, cdk_patch = (int(part) for part in cdk_version.split("."))
```
flash1293, line 148: why not using `_get_version_parts` here?

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -142,6 +144,33 @@ def _validate_source(self):
...
+        manifest_version = self._source_config.get("version")
+        manifest_major, manifest_minor, manifest_patch = self._get_version_parts(manifest_version)
+
+        if cdk_major < manifest_major or (cdk_major == manifest_major and cdk_minor < manifest_minor):
```
flash1293, line 152: why not checking the patch version here?

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -142,6 +144,33 @@ def _validate_source(self):
...
+            raise ValidationError(
+                f"The manifest version {manifest_version} is greater than the airbyte-cdk package version ({cdk_version}). Your "
+                f"manifest may contain features that are not in the current CDK version."
+            )
+        elif manifest_major == 0 and manifest_minor < 29 and (0 < cdk_major or 29 <= cdk_minor):
```
flash1293, line 157: the `(0 < cdk_major or 29 <= cdk_minor)` part of this condition is unnecessary because it can never get false (we will never go back before 0.29 from this point on)

flash1293 (commented):
Left some nits, but looks good in general. Did not test locally

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
```diff
@@ -142,6 +144,33 @@ def _validate_source(self):
...
+            raise ValidationError(
+                f"The low-code framework was promoted to Beta in airbyte-cdk version 0.29.0 and contains many breaking changes to the "
+                f"language. The manifest version {manifest_version} is incompatible with the airbyte-cdk package version "
+                f"{cdk_version} which contains these breaking changes."
+            )
+
+    @staticmethod
+    def _get_version_parts(version: str) -> (int, int, int):
```
maxi297, line 165: Should we document somewhere (maybe in `setup.py`) that if the versioning changes, this will be impacted? I remember some people talking about release candidate versions for example which would add things at the end of the version

maxi297 (approved):
LGTM!

clnoll (approved):
LGTM pending response to @flash1293's comments.

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
flash1293: Very valid point, maybe using a lib like https://pypi.org/project/semver/ would be worth it - your call @brianjlai

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
brianjlai: The idea here is that if we're sticking with semantic versioning, the patch should always be backwards compatible and that we should be encouraging or supporting use of the latest patch since it has fixes. I'm happy to add it, but I also feel like we shouldn't be strict on patches

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
brianjlai: I am not sure that is completely true. We _could_ have an individual connector that pins it's CDK version back to `0.25.0` and the manifest is also running below `0.29.0`. In that case we would not want to throw an error because it is technically a valid pre-beta manifest. Basically it covers the case that manifest + cdk are pre-beta

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
brianjlai: I do like the idea of not building a version comparator that might already exist, although my main concern is that in GH repo in particular they have at the top **"This is a development version. Do NOT use it in production before the final 3.0.0 is released."** 😟 

As far as I can tell they're still on the dev release so I'm not sure we want to quite rely on it just yet. I don't see too many alternatives or widely supported libraries. The other caveat that I can also see us simplifying our versioning notation on the manifest to just v1 or v1.2 and if we do that we might scrap using library anyway.

Interesting, I don't think I was part of those convos or maybe I just forgot. I think a comment in the CDK setup makes sense and I'll add that. And the other nice thing that is the feedback loop on this is very quick, any command run during testing will fail loudly with the error message so theres no way to check these connectors in w/o a fix and we should notice it during testing.

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
brianjlai: we could, the main value of `_get_version_parts()` was it does some error checking since manifests could have any string, whereas for the package string is "guaranteed". That being said, its easy to reuse the two and still ensure a good error message so i'll fix this

brianjlai:
Just a note that I spaced and never pushed a commit w/ the version fixes to existing manifests, it looks like a lot of changes, but only the versions have been bumped from `0.x.0` to `0.29.0` and there was some auto-linting to some files that were not formatted properly.

Results from our validate script which will catch version errors on `spec`
```
% ./gradlew validateSourceYamlManifest                                                                           
Starting a Gradle Daemon (subsequent builds will be faster)
Building all of Airbyte.
/Users/brian.lai/dev/airbyte/airbyte-integrations/connectors

> Task :airbyte-cdk:python:validateSourceYamlManifest
Source activecampaign is fine
Source aha is fine
Source alpha_vantage is fine
Source ashby is fine
Source braze is fine
Source breezometer is fine
Source callrail is fine
Source chargebee is fine
Source chartmogul is fine
Source clickup_api is fine
Source close_com is fine
Source coin_api is fine
Source coingecko_coins is fine
Source coinmarketcap is fine
Source configcat is fine
Source convertkit is fine
Source courier is fine
Source datascope is fine
Source delighted is fine
Source dremio is fine
Source emailoctopus is fine
Source facebook_pages is fine
Source getlago is fine
Source gnews is fine
Source gocardless is fine
Source gong is fine
Source google_pagespeed_insights is fine
Source google_webfonts is fine
Source greenhouse is fine
Source gutendex is fine
Source instatus is fine
Source intruder is fine
Source ip2whois is fine
Source k6_cloud is fine
Source launchdarkly is fine
Source lokalise is fine
Source mailerlite is fine
Source mailersend is fine
Source mailjet_mail is fine
Source mailjet_sms is fine
Source metabase is fine
Source monday is fine
Source n8n is fine
Source news_api is fine
Source newsdata is fine
Source nytimes is fine
Source omnisend is fine
Source oura is fine
Source partnerstack is fine
Source pexels_api is fine
Source plausible is fine
Source pocket is fine
Source polygon_stock_api is fine
Source posthog is fine
Source postmarkapp is fine
Source prestashop is fine
Source punk_api is fine
Source pypi is fine
Source railz is fine
Source recreation is fine
Source recruitee is fine
Source reply_io is fine
Source rocket_chat is fine
Source sap_fieldglass is fine
Source secoda is fine
Source sendgrid is fine
Source sendinblue is fine
Source senseforce is fine
Source sentry is fine
Source smaily is fine
Source smartengage is fine
Source sonar_cloud is fine
Source spacex_api is fine
Source square is fine
Source statuspage is fine
Source survey_sparrow is fine
Source tempo is fine
Source the_guardian_api is fine
Source tmdb is fine
Source toggl is fine
Source tvmaze_schedule is fine
Source twilio_taskrouter is fine
Source twitter is fine
Source tyntec_sms is fine
Source vantage is fine
Source vitally is fine
Source waiteraid is fine
Source whisky_hunter is fine
Source wikipedia_pageviews is fine
Source woocommerce is fine
Source workable is fine
Source workramp is fine
Source zapier_supported_storage is fine
Source zenloop is fine
Source zoom is fine
```

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
flash1293: Fair point, if we make sure to never add functionality in a patch version it should be fine

airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py
flash1293: But this change won’t be part of cdk version 0.25.0 - it will go in 0.31.0 or so and onward.

Do you plan to release 0.25.1 separately which cherry-picks this check back to the state of 0.25.0?

brianjlai: ah duh yeah you're right the old cdk version wouldn't have this logic. i'll get rid of it

Issue #23767, opened 2023-03-06
Python CDK: Retry Token Refresh Requests


davydov-d:
Unlike regular http requests that are initiated during syncs, requests to refresh oauth tokens are not retried. As the number of connectors is growing, this becomes a bottleneck for more connectors. Especially, when the token expires right after the rate limits are exceeded. The connector tries to refresh the token but gets a 429 error in response and it fails a sync. This is a known (and fixed) issue for the [`snapchat-marketing`](https://github.com/airbytehq/airbyte/pull/17596) and [`linkedin-ads`](https://github.com/airbytehq/airbyte/pull/17724) source connectors. Now, it is also reported to be the issue for the [`salesloft`](https://github.com/airbytehq/airbyte/issues/13660) source connector. Since it is already third connector to face this problem, I think it's time to move the patch to the CDK so it wouldn't arise anymore.

davydov-d:
https://github.com/airbytehq/airbyte/pull/23815

PR #23815, opened 2023-03-07
CDK: retry token refresh requests

davydov-d:
## What
#23767 

## How
Using backoff exponential strategy


airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
```diff
@@ -64,22 +70,34 @@ def build_refresh_request_body(self) -> Mapping[str, Any]:
 
         return payload
 
+    @backoff.on_exception(
```
sherifnada, line 73: Just out of curiosity. Is it possible to disable or configure this behavior if needed? Not urgent/blocking.

sherifnada (approved):

airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py
davydov-d: Well disabling may be a bit more more challenging than configuring it but still I think both are possible. Anyway, I think it's worth doing in a dedicated PR if we do need these options

PR #23817, opened 2023-03-07
Fixes required fields on SurveyMonkey UI

### LINKED ISSUE #23769: Set up source SurveyMonkey fails when not providing optional field Client ID
The optional information for Client ID in the source set up SurveyMonkey is wrong. Should be optional, but it fails.

## Environment
- **Airbyte version**:  0.40.26
- **OS Version / Instance**: Debian GNU/Linux 11
- **Deployment**: Docker
- **Source Connector and version**: Surveymonkey 0.1.15; tested and problem exists as well with 0.1.13
- **Destination Connector and version**: -
- **Step where error happened**: Setup new connection

## Current Behavior
I get the error `'client_id in credentials is not provided'`

## Expected Behavior
The Client ID configuration field is optional. I should not get this error here.

## Logs
image

## Steps to Reproduce
1. Set up connector and leave the "Client ID" empty
2. Click on "Set up source"

Alihassanc5:
Closes #23769 

## What
The SurveyMonkey UI displays the 'Client ID' and 'Client Secret' fields as optional, but they are actually required for successful setup.
image

## How
This pull request fixes the issue by updating the UI to reflect their true required status.
image

### DIFF
--- a/airbyte-integrations/connectors/source-surveymonkey/source_surveymonkey/spec.json
+++ b/airbyte-integrations/connectors/source-surveymonkey/source_surveymonkey/spec.json
@@ -19,7 +19,7 @@
         "title": "SurveyMonkey Authorization Method",
         "description": "The authorization method to use to retrieve data from SurveyMonkey",
         "type": "object",
-        "required": ["auth_method", "access_token"],
+        "required": ["auth_method", "access_token", "client_id", "client_secret"],
         "order": 2,
         "properties": {
           "auth_method": {


marcosmarxm:
Thanks for the contribution @Alihassanc5 right now I don't have access to your branch to push changes. Also there are a lot of changes need to be made besides you proposed in this PR. I'm going to finish the work in https://github.com/airbytehq/airbyte/pull/25080

PR #23829, opened 2023-03-07
temporary fix for Connectors Base CI

maxi297:
## What
Fixes error like https://github.com/airbytehq/airbyte/actions/runs/4327429815/jobs/7556319706

Caching seems to screw the venv. I'll investigate further but this should enable the PRs to be merged at least

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -261,11 +261,6 @@ jobs:
         with:
           ref: ${{ github.head_ref }}
 
-      - name: Cache Build Artifacts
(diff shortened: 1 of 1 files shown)


clnoll:
@maxi297 I'm wondering if we could be a bit more surgical and just remove [this line](https://github.com/airbytehq/airbyte/blob/d76151976ffcfe673465af269a8e9f6b63f5c5a3/.github/actions/cache-build-artifacts/action.yml#L50)?

clnoll (approved):
Cool, approved in the meantime in order to unblock people.

PR #23870, opened 2023-03-08
Property to limit test parallelisation

sergio-ropero:
## What
After adding parallelisation in the test pipelines, some of the connectors started to fail (specially the ones that uses shared resource)
[PR](https://github.com/airbytehq/airbyte/pull/23506)

## How
Both `test` and `integrationTest` has been modified to include a property in `gradle.properties` file in the connector folder so we keep parallelisation by default with maximum threads possible but at the same time we let connector owners to configure the number of threads.
That is really useful for connectors that shares resources and are limited (like Redshift or Snowflake that are limited by number of connections).

To limit the parallelism, is it enough to create `gradle.properties` file and include this line on it:
```
numberThreads=X
```
Where X is the desired parallelisation.

Also, this PR removes the concatenation of test for performance tests, so are not run multiple times.

Example with a `println` when I had set up 3 threads, now should show 2 threads (removed as we don't want to have prints ):
Screenshot 2023-03-08 at 12 09 42

## Recommended reading order
No recommendation

## 🚨 User Impact 🚨
The users that have issues with the test should create a `gradle.properties` file with the limitation desired.

### DIFF
--- a/airbyte-integrations/connectors/destination-snowflake/gradle.properties
+++ b/airbyte-integrations/connectors/destination-snowflake/gradle.properties
@@ -1,0 +1,1 @@
+numThreads=3
--- a/airbyte-integrations/connectors/destination-snowflake/gradle.properties
+++ b/airbyte-integrations/connectors/destination-snowflake/gradle.properties
@@ -1,1 +1,1 @@
-numThreads=3
+numberThreads=3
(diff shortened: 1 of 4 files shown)


ryankfu (approved):

sergio-ropero:
Merging as the errors in the `destination-snowflake` tests are related to the test itself when it runs in parallel and not because issues in the logging.

PR #23888, opened 2023-03-08
New `connector_builder` module for handling requests from the Connector Builder

clnoll:
## What
Initial implementation of Connector Builder handlers in the CDK.
This PR just implements `resolve_manifest` handler, but provides the structure we'll use for implementing `list_streams` and `stream_read`.

## How
Modifies the `source_declarative_manifest` to handle Connector Builder requests. These are identified by the presence of a `__command` key in the config.

If a `__command` key is not present in the config, we create the `ManifestDeclarativeSource` and then call `launch` on the source as usual.

Otherwise, we instantiate a `ConnectorBuilderSource` to handle the command. The `ConnectorBuilderSource` is instantiated with a `ManifestDeclarativeSource`. To handle the `resolve_manifest` request, the `ConnectorBuilderSource` calls `resolved_manifest` on the injected `ManifestDeclarativeSource`, and returns the result as an AirbyteRecordMessage whose structure matches the Connector Builder server’s [ResolveManifest object](https://github.com/airbytehq/airbyte-platform/blob/4911d653323122e85f3ed0a2772b96dfab47e80c/airbyte-connector-builder-server/src/main/openapi/openapi.yaml#L273).

### DIFF
--- a/airbyte-cdk/python/connector_builder/README.md
+++ b/airbyte-cdk/python/connector_builder/README.md
@@ -10,7 +10,7 @@ This is the backend for requests from the [Connector Builder](https://docs.airby
 python main.py read --config secrets/config.json
 ```
 
-Note: Requires the keys `__injected_declarative_manifest` and `__command` in its config, where `__injected_declarative_manifest` is a JSON manifest and `__command` is one of the commands handled by the ConnectorBuilderSource (`stream_read`, `list_streams`, or `resolve_manifest`).
+Note: Requires the keys `__injected_declarative_manifest` and `__command` in its config, where `__injected_declarative_manifest` is a JSON manifest and `__command` is one of the commands handled by the ConnectorBuilderHandler (`stream_read`, `list_streams`, or `resolve_manifest`).
 
 ### Locally running the docker image
 
--- a/airbyte-cdk/python/connector_builder/README.md
+++ b/airbyte-cdk/python/connector_builder/README.md
@@ -1,0 +1,32 @@
+# Connector Builder Backend
+
+This is the backend for requests from the [Connector Builder](https://docs.airbyte.com/connector-development/config-based/connector-builder-ui/).
+
+## Local development
+
+### Locally running the Connector Builder backend
+
+```
+python main.py read --config secrets/config.json
+```
+
+Note: Requires the keys `__injected_declarative_manifest` and `__command` in its config, where `__injected_declarative_manifest` is a JSON manifest and `__command` is one of the commands handled by the ConnectorBuilderSource (`stream_read`, `list_streams`, or `resolve_manifest`).
+
+### Locally running the docker image
+
+#### Build
+
+First, make sure you build the latest Docker image:
+```
+./gradlew airbyte-cdk:python:airbyteDocker
+```
+
+The docker image name and tag, respectively, are the values of the `io.airbyte.name` and `io.airbyte.version` `LABEL`s in the Dockerfile.
+
+#### Run
+
+Then run any of the connector commands as follows:
+
+```
+docker run --rm -v $(pwd)/secrets:/secrets airbyte/source-declarative-manifest:dev read --config /secrets/config.json
+```
--- a/airbyte-cdk/python/connector_builder/README.md
+++ b/airbyte-cdk/python/connector_builder/README.md
@@ -7,10 +7,21 @@ This is the backend for requests from the [Connector Builder](https://docs.airby
 ### Locally running the Connector Builder backend
 
 ```
-python main.py read --config secrets/config.json
+python main.py read --config path/to/config --catalog path/to/catalog
 ```
 
-Note: Requires the keys `__injected_declarative_manifest` and `__command` in its config, where `__injected_declarative_manifest` is a JSON manifest and `__command` is one of the commands handled by the ConnectorBuilderHandler (`stream_read`, `list_streams`, or `resolve_manifest`).
+Note:
+- Requires the keys `__injected_declarative_manifest` and `__command` in its config, where `__injected_declarative_manifest` is a JSON manifest and `__command` is one of the commands handled by the ConnectorBuilderHandler (`stream_read`, `list_streams`, or `resolve_manifest`), i.e.
+```
+{
+  "config": <normal config>,
+  "__injected_declarative_manifest": {...},
+  "__command": <"resolve_manifest" | "list_streams" | "stream_read">
+}
+```
+*See [ConnectionSpecification](https://docs.airbyte.com/understanding-airbyte/airbyte-protocol/#actor-specification) for details on the `"config"` key if needed.
+
+- When the `__command` is `list_streams` or `resolve_manifest`, the argument to `catalog` should be an empty string.
 
 ### Locally running the docker image
 
--- a/airbyte-cdk/python/connector_builder/connector_builder_handler.py
+++ b/airbyte-cdk/python/connector_builder/connector_builder_handler.py
@@ -10,7 +10,7 @@ from airbyte_cdk.sources.declarative.manifest_declarative_source import Manifest
 from airbyte_cdk.utils.traced_exception import AirbyteTracedException
 
 
-class ConnectorBuilderSource:
+class ConnectorBuilderHandler:
     def __init__(self, source: ManifestDeclarativeSource):
         self.source = source
 
--- a/airbyte-cdk/python/connector_builder/connector_builder_handler.py
+++ b/airbyte-cdk/python/connector_builder/connector_builder_handler.py
@@ -3,40 +3,31 @@
 #
 
 from datetime import datetime
-from typing import Any, Mapping, Union
+from typing import Union
 
 from airbyte_cdk.models import AirbyteMessage, AirbyteRecordMessage
-from airbyte_cdk.sources.declarative.manifest_declarative_source import ManifestDeclarativeSource
 from airbyte_cdk.utils.traced_exception import AirbyteTracedException
 
 
-class ConnectorBuilderHandler:
-    def __init__(self, source: ManifestDeclarativeSource):
-        self.source = source
+def list_streams() -> AirbyteRecordMessage:
+    raise NotImplementedError
 
-    def list_streams(self) -> AirbyteRecordMessage:
-        raise NotImplementedError
 
-    def stream_read(self, command_config) -> AirbyteRecordMessage:
-        raise NotImplementedError
+def stream_read() -> AirbyteRecordMessage:
+    raise NotImplementedError
 
-    @staticmethod
-    def _emitted_at():
-        return int(datetime.now().timestamp()) * 1000
 
-    def resolve_manifest(self) -> Union[AirbyteMessage, AirbyteRecordMessage]:
-        try:
-            return AirbyteRecordMessage(
-                data={"manifest": self.source.resolved_manifest},
-                emitted_at=self._emitted_at(),
-                stream="",
-            )
-        except Exception as exc:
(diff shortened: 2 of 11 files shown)


airbyte-cdk/python/connector_builder/connector_builder_source.py
```diff
@@ -0,0 +1,42 @@
...
+
+from datetime import datetime
+from typing import Any, Mapping, Union
+
+from airbyte_cdk.models import AirbyteMessage, AirbyteRecordMessage
+from airbyte_cdk.sources.declarative.manifest_declarative_source import ManifestDeclarativeSource
+from airbyte_cdk.utils.traced_exception import AirbyteTracedException
+
+
+class ConnectorBuilderSource:
```
girarda, line 13: this isn't really a source, right?

airbyte-cdk/python/connector_builder/README.md
```diff
@@ -0,0 +1,32 @@
+# Connector Builder Backend
```
girarda, line 1: should the readme be in the `source_declarative_manifest` module since that's where the main is?

airbyte-cdk/python/connector_builder/connector_builder_source.py
```diff
@@ -0,0 +1,42 @@
...
+    @staticmethod
+    def _emitted_at():
+        return int(datetime.now().timestamp()) * 1000
+
+    def resolve_manifest(self) -> Union[AirbyteMessage, AirbyteRecordMessage]:
+        try:
+            return AirbyteRecordMessage(
+                data={"manifest": self.source.resolved_manifest},
+                emitted_at=self._emitted_at(),
+                stream="",
```
girarda, line 32: would it make sense to emit the records in a `manifest` or `resolved_manifest` stream?

airbyte-cdk/python/connector_builder/README.md
clnoll: That has a separate README, which is to do with running the Connector Builder manifests. Based on this comment I added a line in there to indicate that it handles Connector Builder requests too, but feel like it makes sense for this to have its own README.

airbyte-cdk/python/connector_builder/connector_builder_source.py
clnoll: Mind clarifying? I'm not sure I see what you're suggesting.

airbyte-cdk/python/connector_builder/connector_builder_handler.py
```diff
@@ -0,0 +1,42 @@
...
+                data={"manifest": self.source.resolved_manifest},
+                emitted_at=self._emitted_at(),
+                stream="",
+            )
+        except Exception as exc:
+            error = AirbyteTracedException.from_exception(exc, message="Error resolving manifest.")
+            return error.as_airbyte_message()
+
+    def handle_request(self, config: Mapping[str, Any]) -> Union[AirbyteMessage, AirbyteRecordMessage]:
+        command = config.get("__command")
```
maxi297, line 39: nit: I'm wondering if this doesn't leak a little bit of the API in the handler layer. For example, if we say we want to have a connector builder as an HTTP service, would this method be used or we would have another mechanism to call the proper method? My guess is that we would have different endpoints like `/manifest/resolve` and the API layer would call `handler. resolve_manifest(...)` directly. Based on that, `handle_request` is probably part of our API and not our handle and this could be moved to `airbyte-cdk/python/source_declarative_manifest/main.py`

maxi297 (approved):
LGTM! I like the clear and concise documentation that comes with the added valud

airbyte-cdk/python/connector_builder/connector_builder_handler.py
```diff
@@ -0,0 +1,42 @@
...
+from typing import Any, Mapping, Union
+
+from airbyte_cdk.models import AirbyteMessage, AirbyteRecordMessage
+from airbyte_cdk.sources.declarative.manifest_declarative_source import ManifestDeclarativeSource
+from airbyte_cdk.utils.traced_exception import AirbyteTracedException
+
+
+class ConnectorBuilderHandler:
+    def __init__(self, source: ManifestDeclarativeSource):
+        self.source = source
```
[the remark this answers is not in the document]
clnoll: The original motivation was to use `ConnectorBuilder` as a wrapper around the source, but your argument makes sense to me. Updated.

airbyte-cdk/python/connector_builder/connector_builder_handler.py
clnoll: I buy it. Updated.

airbyte-cdk/python/connector_builder/connector_builder_source.py
girarda: I'm suggesting setting the stream name for these records

airbyte-cdk/python/source_declarative_manifest/README.md
```diff
@@ -7,6 +7,8 @@ This entrypoint is used for connectors created by the connector builder. These c
 
 The spec operation is not supported because the config is not known when running a spec.
 
+This entrypoint is also the entrypoint for requests from the [Connector Builder](https://docs.airbyte.com/connector-development/config-based/connector-builder-ui/) Server. In addition to the `__injected_declarative_manifest`, the Connector Builder backend config requires the `__command` key, whose value is one of the commands handled by the ConnectorBuilderHandler (`stream_read`, `list_streams`, or `resolve_manifest`).
```
girarda, line 10: nit: should this link to the other readme?

airbyte-cdk/python/connector_builder/connector_builder_handler.py
```diff
@@ -0,0 +1,33 @@
...
+def stream_read() -> AirbyteRecordMessage:
+    raise NotImplementedError
+
+
+def resolve_manifest(source) -> Union[AirbyteMessage, AirbyteRecordMessage]:
+    try:
+        return AirbyteRecordMessage(
+            data={"manifest": source.resolved_manifest},
+            emitted_at=_emitted_at(),
+            stream="",
```
girarda, line 25: other comment disappeared - I think we should set the stream name so the record contains enough information to be interpreted without additional context

girarda (approved):
small nits but looks good otherwise

airbyte-cdk/python/connector_builder/connector_builder_handler.py
```diff
@@ -0,0 +1,33 @@
...
+
+def list_streams() -> AirbyteRecordMessage:
+    raise NotImplementedError
+
+
+def stream_read() -> AirbyteRecordMessage:
+    raise NotImplementedError
+
+
+def resolve_manifest(source) -> Union[AirbyteMessage, AirbyteRecordMessage]:
```
brianjlai, line 20: It feels a little inconsistent that we return either an AirbyteMessage or an AirbyteRecordMessage. The "parent" AirbyteMessage object has fields for each of the relevant message unless the protocol has changed. Shouldn't we be returning an AirbyteMessage w/ the `record` or `trace` field set and the relevant `type`. Or returning a `Union[AirbyteRecordMessage, AirbyteTraceMessage]`?

brianjlai (approved):
question about the return types on the handler, but lgtm!

airbyte-cdk/python/source_declarative_manifest/main.py
```diff
@@ -3,27 +3,66 @@
...
+    parser.add_argument("command", type=str, help="Airbyte Protocol command")
+    parser.add_argument("--config", type=str, required=True, help="path to the json configuration file")
+    parsed, _ = parser.parse_known_args(args)
+    return parsed.command, parsed.config
+
+
+def handle_connector_builder_request(source: DeclarativeSource, config: Mapping[str, Any]):
+    command = config.get("__command")
+    if command == "resolve_manifest":
+        return connector_builder_handler.resolve_manifest(source)
```
girarda, line 48: shouldn't this print the output?

airbyte-cdk/python/connector_builder/connector_builder_source.py
clnoll: Ah okay good call, updated.

airbyte-cdk/python/source_declarative_manifest/README.md
clnoll: 👍 done

airbyte-cdk/python/source_declarative_manifest/main.py
clnoll: Yes! Done.

airbyte-cdk/python/connector_builder/connector_builder_handler.py
clnoll: Makes sense to me, set it to `"resolve_manifest"`.

airbyte-cdk/python/connector_builder/connector_builder_handler.py
clnoll: Thanks for pointing that out, I think always returning `AirbyteMessage` makes sense.

clnoll: (updated)

girarda (changes requested):

airbyte-cdk/python/connector_builder/main.py
```diff
@@ -0,0 +1,67 @@
...
+
+    if "__injected_declarative_manifest" not in config:
+        raise ValueError(
+            f"Invalid config: `__injected_declarative_manifest` should be provided at the root of the config but config only has keys {list(config.keys())}"
+        )
+
+    return config
+
+
+def preparse(args: List[str]) -> Tuple[str, str]:
```
girarda, line 36: the interface should be the same as other connectors so this should also accept `--catalog` and and optional `--state`

It might make sense to always use `AirbyteEntrypoint.parse_args`

airbyte-cdk/python/source_declarative_manifest/README.md
```diff
@@ -7,6 +7,8 @@ This entrypoint is used for connectors created by the connector builder. These c
 
 The spec operation is not supported because the config is not known when running a spec.
 
+This entrypoint is also the entrypoint for requests from the [Connector Builder](https://docs.airbyte.com/connector-development/config-based/connector-builder-ui/) Server. In addition to the `__injected_declarative_manifest`, the [Connector Builder backend](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/connector_builder/README.md) config requires the `__command` key, whose value is one of the commands handled by the ConnectorBuilderHandler (`stream_read`, `list_streams`, or `resolve_manifest`).
```
girarda, line 10: it would make sense to provide complete config examples (maybe without the manifest) to clarify how this is meant to be used

airbyte-cdk/python/connector_builder/main.py
girarda: @clnoll is there a reason to use a preparse method instead of directly calling `AirbyteEntrypoint.parse_args`? Using the same method ensures we're using the same interface

girarda (approved):
question about using `preparse` vs `AirbyteEntrypoint.parse_args` but looks good otherwise

airbyte-cdk/python/connector_builder/main.py
clnoll: Good call, updated to use `AirbyteEntrypoint.parse_args`.

airbyte-cdk/python/source_declarative_manifest/README.md
clnoll: Added to the `connector_builder` README.

PR #23907, opened 2023-03-09
avoid cache overlap between cdk-build and connectors-build

maxi297:
## What
Create specific caches for cdk and connector builds. This will avoid cache overlap and bad cache entries being created


maxi297:
The issue we will have is that we already have polluted cache entries. I was considering manually deleting them but even when I merge this in master, there will still be branches with cdk-build creating cache entries. The other option is to add something in the cache key and regenerate cache entries

clnoll (approved):
👍

PR #23917, opened 2023-03-09
fix: Cast the number of threads to integer in the Java connector tests 😵‍💫 

sergio-ropero:
# What
When introducing the property to limit the parallelism, it executes with more threads than expected ([PR](https://github.com/airbytehq/airbyte/pull/23870)).

# Solution
The problem is that when gradle reads the property file (parameter `numberThreads=2`), it was reading the ASCII value of `2` (50) and the test were executed with 50 threads instead of 2 😭 

To sort this out, we cast the property to `int`.


ryankfu (approved):
🥳

PR #23919, opened 2023-03-09
Running the formatter on Java connectors

akashkulk:
Ran `./gradlew :spotlessJavaApply`


edgao (approved):
I only looked at the first few files; looks fine though

PR #23921, opened 2023-03-09
support reading "expires_in" when the API passes the value as string

leo-schick:
## What
I was dealing with the OAuth2 implementation of trustpilot.com and they pass the `expires_in` OAuth parameter as string and not as integer.

This results then in the following error:
```
bad operand type for abs(): 'str'
```

With this PR, we would cast the value from `expires_in` into an integer.

## How
using `int(..)` around the value. This then supports both use cases: string and integer values

## 🚨 User Impact 🚨
No breaking changes.


sherifnada (commented):
@leo-schick thanks for your contribution! We actually just received this same contribution yesterday here: https://github.com/airbytehq/airbyte/pull/20301/files and it will be released very soon. So you should be able to receive the fix today

sherifnada:
@leo-schick  actually come to think of it, I see you have one additional bugfix not included in that PR. 

I've just merged that PR to master. Could you pull master into your branch, add a test case for the extra file you modified not modified in that PR, and I'll merge your contribution as well?

sherifnada:
@leo-schick please pull master and lmk when it's ready for a review

leo-schick:
@sherifnada done

airbyte-cdk/python/airbyte_cdk/sources/streams/http/auth/oauth.py
```diff
@@ -95,7 +95,7 @@ def refresh_access_token(self) -> Tuple[str, int]:
             )
             response.raise_for_status()
             response_json = response.json()
-            return response_json["access_token"], response_json["expires_in"]
+            return response_json["access_token"], int(response_json["expires_in"])
```
sherifnada, line 98: @leo-schick could you add a unit test for this case?

sherifnada (approved):

PR #23925, opened 2023-03-10
Connector builder: support for test read with message grouping per slices

girarda:
## What
* Implement the test read method

Sample command
`python main.py read --config secrets/config.json --catalog secrets/configured_catalog.json`

## Test read

The test read method will run if the `"__command"` field is set to `"test_read"`. The source will read data from stream defined in the configured catalog. Only one stream at a time is currently supported.

The test_read command can be configured with the optional `"__test_read_config"` field in the config
The `"__test_read_config"` should have the following fields
- max_pages_per_slice
- max_slices
- max_records

sample config for a test read:
```
{
  "api_key": "",
  "__command": "test_read"
  "__test_read_config": {
    "stream_name": "Accounts",
    "max_pages_per_slice": 5,
    "max_slices": 5,
    "max_records": 10
  },
  "__injected_declarative_manifest": {
}
```

configured catalog:
```
{
  "streams": [
    {
      "stream": {
        "name": "Accounts",
        "json_schema": {
          "type": "object",
          "properties": {}
        },
        "supported_sync_modes": ["full_refresh"],
        "source_defined_cursor": false
      },
      "sync_mode": "full_refresh",
      "destination_sync_mode": "overwrite"
    }
  ]
}
```

sample configured catalog for resolving the manifest:
```
{
  "streams": [
    {
      "stream": {
        "name": "resolve_manifest",
        "json_schema": {
          "type": "object",
          "properties": {}
        },
        "supported_sync_modes": ["full_refresh"],
        "source_defined_cursor": false
      },
      "sync_mode": "full_refresh",
      "destination_sync_mode": "overwrite"
    }
  ]
}
```

## Interface change
* Implement the read_stream method, which will be called if the config's command == "test_read"

### Test read
* Copy the [models from the connector builder server](https://github.com/airbytehq/airbyte-platform-internal/tree/master/oss/airbyte-connector-builder-server/connector_builder/generated/models) and convert them to dataclasses
* Copy the tests from 
* https://github.com/airbytehq/airbyte-platform-internal/blob/master/oss/airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py
* Copy the bulk of the [default_api logic](https://github.com/airbytehq/airbyte-platform-internal/blob/master/oss/airbyte-connector-builder-server/connector_builder/impl/default_api.py) to a new MessageGrouper class
* Implement connector_builder.connector_builder_handler.read_stream
* Call read_stream from the main

## Recommended reading order
1. `airbyte-cdk/python/source_declarative_manifest/main.py`
2. `airbyte-cdk/python/connector_builder/connector_builder_handler.py`
3. `airbyte-cdk/python/connector_builder/message_grouper.py`
4. `airbyte-cdk/python/connector_builder/models.py`
5. `airbyte-cdk/python/airbyte_cdk/sources/source.py`
6. `airbyte-cdk/python/unit_tests/connector_builder/test_message_grouper.py`
7. `airbyte-cdk/python/airbyte_cdk/sources/source.py`
8. `airbyte-cdk/python/unit_tests/connector_builder/utils.py`


airbyte-cdk/python/unit_tests/connector_builder/test_read.py
```diff
@@ -0,0 +1,503 @@
+#
```
girarda, line 1: I ported these tests from https://github.com/airbytehq/airbyte-platform-internal/blob/master/oss/airbyte-connector-builder-server/unit_tests/connector_builder/impl/test_default_api.py

airbyte-cdk/python/connector_builder/connector_builder_handler.py
```diff
@@ -0,0 +1,68 @@
...
+        )
+    except Exception as exc:
+        error = AirbyteTracedException.from_exception(exc, message="Error resolving manifest.")
+        return error.as_airbyte_message()
+
+
+def _emitted_at():
+    return int(datetime.now().timestamp()) * 1000
+
+def _create_configure_catalog(stream_name: str) -> ConfiguredAirbyteCatalog:
```
girarda, line 52: creating the catalog on the Java side would allow us to keep the same read interface as connectors

airbyte-cdk/python/connector_builder/message_grouper.py
```diff
@@ -0,0 +1,192 @@
...
+        if len(slices) >= self._max_slices:
+            return True
+
+        for slice in slices:
+            if len(slice.pages) >= self._max_pages_per_slice:
+                return True
+        return False
+
+    @classmethod
+    def _create_configure_catalog(cls, stream_name: str) -> ConfiguredAirbyteCatalog:
```
girarda, line 177: creating the catalog on the Java side would allow us to keep the same read interface as connectors

airbyte-cdk/python/connector_builder/message_grouper.py
```diff
@@ -0,0 +1,192 @@
...
+
+from airbyte_cdk.models import AirbyteLogMessage, AirbyteMessage, Type
+from airbyte_cdk.sources.declarative.declarative_source import DeclarativeSource
+from airbyte_cdk.utils.schema_inferrer import SchemaInferrer
+from connector_builder.models import StreamRead, StreamReadPages, HttpResponse, HttpRequest, StreamReadSlices
+import logging
+from copy import deepcopy
+
+
+class MessageGrouper:
```
girarda, line 20: bulk of the logic is pasted from https://github.com/airbytehq/airbyte-platform-internal/blob/master/oss/airbyte-connector-builder-server/connector_builder/impl/default_api.py

airbyte-cdk/python/connector_builder/message_grouper.py
```diff
@@ -0,0 +1,192 @@
...
+        self.max_record_limit = max_record_limit
+
+    def get_message_groups(self,
+                           source: DeclarativeSource,
+                           config: Mapping[str, Any],
+                           stream: str,
+                           record_limit: Optional[int] = None,
+                           ) -> StreamRead:
+        if record_limit is not None and not (1 <= record_limit <= 1000):
+            raise ValueError(f"Record limit must be between 1 and 1000. Got {record_limit}")
```
girarda, line 35: moved from a validator because dataclasses don't have builtin validators

airbyte-cdk/python/connector_builder/connector_builder_handler.py
```diff
@@ -2,28 +2,46 @@
...
+
+from airbyte_cdk.models import AirbyteRecordMessage, AirbyteMessage
+from airbyte_cdk.models import Type as MessageType
+from airbyte_cdk.sources.declarative.declarative_source import DeclarativeSource
 from airbyte_cdk.utils.traced_exception import AirbyteTracedException
+from connector_builder.message_grouper import MessageGrouper
 
 
-def list_streams() -> AirbyteRecordMessage:
+def list_streams() -> AirbyteMessage:
```
girarda, line 19: this should return an AirbyteMessage containing an AirbyteRecord

airbyte-cdk/python/connector_builder/connector_builder_handler.py
```diff
@@ -2,28 +2,46 @@
...
+    stream_read = handler.get_message_groups(source, config, stream_name, max_records)
+    return AirbyteMessage(type=MessageType.RECORD, record=AirbyteRecordMessage(
+        data=dataclasses.asdict(stream_read),
+        stream="_test_read",
+        emitted_at=_emitted_at()
+    ))
 
 
-def resolve_manifest(source) -> Union[AirbyteMessage, AirbyteRecordMessage]:
+def resolve_manifest(source) -> AirbyteMessage:
```
girarda, line 38: this can always return an AirbyteMessage

airbyte-cdk/python/source_declarative_manifest/main.py
```diff
@@ -3,63 +3,47 @@
...
-    AirbyteEntrypoint.parse_args(args)
-    launch(source, sys.argv[1:])
-
-
-def handle_request(args: List[str]):
-    config = get_config_from_args(args)
-    source = create_source(config)
-    if "__command" in config:
-        print(handle_connector_builder_request(source, config))
+    if command == "read":
```
clnoll, line 36: nit: WDYT about putting some exception handling around `get_connector_builder_request_handler` to prevent exceptions from blowing up the production codepath, and moving the `catalog` lines from this function to `get_connector_builder_request_handler` to make a more obvious separation between what is/isn't relevant to a connector builder request.

clnoll (commented):
@girarda this lgtm, though it feels like in this design we're being quite implicit about whether a request is related to the connector builder. "whether a request is related to the connector builder" isn't important in itself - the important thing is whether a request is for a production versus a non-production read, since we don't want non-production logic to interfere with syncs. It feels like this design does put a bit more logic than I'd ideally want in front of the prod codepath. So I'm mulling over whether there's a way that we can be more explicit. Do we definitely want to have to read the catalog before we can determine that we're doing a connector builder operation? Perhaps we can have a `__command` key in the config per the original plan, but the connector builder `stream_read` requests can take in a catalog argument in the normal Airbyte protocol format?

airbyte-cdk/python/connector_builder/connector_builder_handler.py
```diff
@@ -2,19 +2,36 @@
...
     raise NotImplementedError
 
 
-def stream_read() -> AirbyteMessage:
-    raise NotImplementedError
+def read_stream(source: DeclarativeSource, config: Mapping[str, Any], configured_catalog: ConfiguredAirbyteCatalog) -> AirbyteMessage:
+    command_config = config.get("__test_read_config")
+    max_pages_per_slice = command_config["max_pages_per_slice"]
+    max_slices = command_config["max_slices"]
+    max_records = command_config["max_records"]
```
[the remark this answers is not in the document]
girarda: Is there a reason to set a default here? This will only get used by the Java-side of the connector builder so it should be sufficient to set the values there

airbyte-cdk/python/connector_builder/main.py
```diff
@@ -41,11 +42,15 @@ def preparse(args: List[str]) -> Tuple[str, str]:
     return parsed.command, parsed.config
 
 
-def handle_connector_builder_request(source: ManifestDeclarativeSource, config: Mapping[str, Any]):
-    command = config.get("__command")
+def handle_connector_builder_request(source: ManifestDeclarativeSource, config: Mapping[str, Any], catalog: ConfiguredAirbyteCatalog):
+    command = config["__command"]
```
girarda, line 23: "__command" is required

airbyte-cdk/python/connector_builder/main.py
```diff
@@ -55,11 +60,15 @@ def handle_connector_request(source: ManifestDeclarativeSource, args: List[str])
 
 
 def handle_request(args: List[str]):
-    config = get_config_from_args(args)
-    source = create_source(config)
+    parser = AirbyteEntrypoint.parse_args(args)
```
girarda, line 40: always use the same parser for simplicity

airbyte-cdk/python/unit_tests/connector_builder/test_connector_builder_handler.py
```diff
@@ -196,4 +203,5 @@ def test_invalid_command(command):
     config["__command"] = command
     source = ManifestDeclarativeSource(CONFIG["__injected_declarative_manifest"])
     with pytest.raises(ValueError):
-        handle_connector_builder_request(source, config)
+        handle_connector_builder_request(source, config, create_configured_catalog("my_stream"))
```
girarda, line 201: pass in a catalog

airbyte-cdk/python/connector_builder/main.py
```diff
@@ -4,46 +4,52 @@
...
+from airbyte_cdk.models import ConfiguredAirbyteCatalog
 from airbyte_cdk.sources.declarative.manifest_declarative_source import ManifestDeclarativeSource
-from connector_builder.connector_builder_handler import resolve_manifest
+from connector_builder.connector_builder_handler import read_stream, resolve_manifest
 
 
 def create_source(config: Mapping[str, Any]) -> ManifestDeclarativeSource:
     manifest = config.get("__injected_declarative_manifest")
-    return ManifestDeclarativeSource(manifest)
+    return ManifestDeclarativeSource(manifest, True)
```
girarda, line 18: set debug to True so the source returns raw requests and responses

airbyte-cdk/python/connector_builder/connector_builder_handler.py
```diff
@@ -2,19 +2,47 @@
...
-def stream_read() -> AirbyteMessage:
-    raise NotImplementedError
+DEFAULT_MAXIMUM_NUMBER_OF_PAGES_PER_SLICE = 5
+DEFAULT_MAXIMUM_NUMBER_OF_SLICES = 5
+DEFAULT_MAX_RECORDS = 100
+
+
+def read_stream(source: DeclarativeSource, config: Mapping[str, Any], configured_catalog: ConfiguredAirbyteCatalog) -> AirbyteMessage:
+    try:
+        if "__test_read_config" not in config:
```
clnoll, line 29: nit: Feels like this should be optional since we're not requiring any of the keys to be present, and are handling the case where it isn't set.

airbyte-cdk/python/connector_builder/message_grouper.py
```diff
@@ -0,0 +1,190 @@
...
+from airbyte_cdk.sources.declarative.declarative_source import DeclarativeSource
+from airbyte_cdk.utils.schema_inferrer import SchemaInferrer
+from airbyte_protocol.models.airbyte_protocol import ConfiguredAirbyteCatalog
+from connector_builder.models import HttpRequest, HttpResponse, StreamRead, StreamReadPages, StreamReadSlices
+
+
+class MessageGrouper:
+    logger = logging.getLogger("airbyte.connector-builder")
+
+    def __init__(self, max_pages_per_slice: int, max_slices: int, max_record_limit: int = 1000):
```
clnoll, line 22: nit (also I know this may just be copy/pasted): Would it be preferable to enforce the maximums in this class, instead of allowing them to be fully configurable?

clnoll (approved):
Looks great @girarda! A couple of questions but no blockers.

airbyte-cdk/python/connector_builder/connector_builder_handler.py
girarda: agreed. fixed

airbyte-cdk/python/connector_builder/message_grouper.py
girarda: the main reason why I didn't enforce maximums here is that we'll need to update the value in two places if we decide to increase it. I think the risk of not enforcing a limit here is negligible because we control the caller

Issue #23926, opened 2023-03-10
CDK: token expiry times are parsed incongruently


sherifnada:
# Current Behavior
[This PR](https://github.com/airbytehq/airbyte/pull/20301) tried to parse the `expires_in` response field (which tells you how long an access token is valid for, in seconds) of an OAuth token refresh request as an integer because when it's returned as a string, it causes issues down the line when trying to determine the timestamp at which the access token will expire. 

The reason for this problem is the following: 
* The `AbstractOauthAuthenticator`  class originally assumed the HTTP response of the refresh_access_token request will contain a field `expires_in` representing how many seconds are left till the access token expires
* In [this PR](https://github.com/airbytehq/airbyte/pull/19718/files) to enable migrating the Square connector to Lowcode, the PR added the ability to accept as input a datetime format representing the time at which a token expires. While this was a useful feature addition, IMO there was a few issues in how it was implemented:
  * This functionality is implicit i.e: in the case that an API returns an explicit expiry date for the access token (instead of how many seconds left till it expires), the consumer must pass in a datetime format for parsing that date in the response, and IF this datetime format is passed then the authenticator "knows" that `expiry_time = parse(response.expires_at, <format>)` rather than `expiry_time = now() + response.expires_in_seconds`. 
  * This fork in logic was implemented in `DeclarativeOAuthAuthenticator` rather than exposed as an option in `AbstractOauthAuthenticator`. This had two downsides: first, `AbstractOauthAuthenticator` always assumes that response field is an int indicating seconds till expiry, so when #20301 tried to parse it as such, `DeclarativeOAuthAuthenticator` tests started failing. Meaning we now have a subclass which overrides an assumption of the superclass it extends. Second, this is a broadly useful utility that should have been made available in `AbstractOAuthAuthenticator` so any consumer of the CDK benefits from it. 

# Expected behavior
- This behavior should be supported directly in `AbstractOauthAuthenticator`
- Once that is done, adjust `DeclarativeOAuthAuthenticator` to use that capability, and no longer override `refresh_access_token` in that class like introduced [in this PR](https://github.com/airbytehq/airbyte/pull/23927)

Issue #20761, opened 2022-12-21
connector builder does not expose requests and response sent to refresh token endpoint when using oauth authenticator


girarda:
## Current Behavior
* Requests sent to the refresh token endpoint are not directly submitted by the HttpStream so they are not intercepted and emitted as log message
* This means connector developers are blind when trying to troubleshoot why they cannot authenticate to the endpoint

## Expected Behavior
* all requests and responses should be emitted to the connector builder

## Additional notes
* Is there a way to ensure all requests and responses are emitted? Any custom component could also submit requests without going through the HttpStream. Maybe by patching the requests library?
* This breaks the 1 req per page model we currently have

PR #23927, opened 2023-03-10
Low-code CDK: Override refresh_access_token logic DeclarativeOAuthAuthenticator

sherifnada:
## What
Enable airbytehq/airbyte#20301 to be merged. Initially it failed unit tests due to the issues mentioned here https://github.com/airbytehq/airbyte/issues/23926

TL;DR `DeclarativeOAuthAuthenticator` breaks some assumptions made by `OAuthAuthenticator` in service of adding a new feature. This PR fixes the issue temporarily by pushing down the new feature logic into `DeclarativeOAuthAuthenticator` instead to enable airbytehq/airbyte#20301 to go through. 

Also contains the changes from airbytehq/airbyte#20301

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
@@ -3,7 +3,7 @@
 #
 
 from dataclasses import InitVar, dataclass, field
-from typing import Any, List, Mapping, Optional, Union
+from typing import Any, List, Mapping, Optional, Union, Tuple
 
 import pendulum
 from airbyte_cdk.sources.declarative.auth.declarative_authenticator import DeclarativeAuthenticator
@@ -91,6 +91,26 @@ class DeclarativeOauth2Authenticator(AbstractOauth2Authenticator, DeclarativeAut
 
     def get_refresh_request_body(self) -> Mapping[str, Any]:
         return self._refresh_request_body.eval(self.config)
+
+    def refresh_access_token(self) -> Tuple[str, Any]:
+        """
+        Returns the refresh token and its lifespan in seconds
+
+        This overrides the parent class method because the parent class assumes the "expires_in" field is always an int representing
+         seconds till token expiry.
+
+        However, this class provides the ability to determine the expiry date of an access token either by using (pseudocode):
+        * expiry_datetime = datetime.now() + seconds_till_access_token_expiry # in this option we have to calculate expiry timestamp, OR
+        * expiry_datetime = parse(response.headers["expires_at"]) # in this option the API tells us exactly when access token expires
+
+        :return: a tuple of (access_token, either token_lifespan_in_seconds or datetime_of_token_expiry)
+
+        # TODO this is a hack and should be better encapsulated/enabled by the AbstractOAuthAuthenticator i.e: that class should have
+                a method which takes the HTTP response and returns a timestamp for when the access token will expire which subclasses
+                such as this one can override or just configure directly.
+        """
+        response_json = self._get_refresh_access_token_response()
+        return response_json[self.get_access_token_name()], response_json[self.get_expires_in_name()]
 
     def get_token_expiry_date(self) -> pendulum.DateTime:
         return self._token_expiry_date
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
@@ -3,7 +3,7 @@
 #
 
 from dataclasses import InitVar, dataclass, field
-from typing import Any, List, Mapping, Optional, Union, Tuple
+from typing import Any, List, Mapping, Optional, Tuple, Union
 
 import pendulum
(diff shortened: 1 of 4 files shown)


airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
```diff
@@ -92,6 +92,26 @@ def get_grant_type(self) -> InterpolatedString:
...
+        This overrides the parent class method because the parent class assumes the "expires_in" field is always an int representing
+         seconds till token expiry.
+
+        However, this class provides the ability to determine the expiry date of an access token either by using (pseudocode):
+        * expiry_datetime = datetime.now() + seconds_till_access_token_expiry # in this option we have to calculate expiry timestamp, OR
+        * expiry_datetime = parse(response.headers["expires_at"]) # in this option the API tells us exactly when access token expires
+
+        :return: a tuple of (access_token, either token_lifespan_in_seconds or datetime_of_token_expiry)
+
+        # TODO this is a hack and should be better encapsulated/enabled by the AbstractOAuthAuthenticator i.e: that class should have
```
girarda, line 108: nit: I don't think, the authenticator should make requests by itself because it [prevents us from catching them in the connector builder](https://github.com/airbytehq/airbyte/issues/20761). Would be really nice to submit all requests from the same class

airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
```diff
@@ -75,11 +75,11 @@ def get_grant_type(self) -> str:
     def get_token_expiry_date(self) -> pendulum.DateTime:
         return self._token_expiry_date
 
-    def set_token_expiry_date(self, initial_time: pendulum.DateTime, value: Union[str, int]):
+    def set_token_expiry_date(self, value: Union[str, int]):
```
girarda, line 78: should we rename this method because it's not setting the expiry date to `value`?

airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
```diff
@@ -92,14 +92,34 @@ def get_grant_type(self) -> InterpolatedString:
...
+        Returns the refresh token and its lifespan in seconds
+
+        This overrides the parent class method because the parent class assumes the "expires_in" field is always an int representing
+         seconds till token expiry.
+
+        However, this class provides the ability to determine the expiry date of an access token either by using (pseudocode):
+        * expiry_datetime = datetime.now() + seconds_till_access_token_expiry # in this option we have to calculate expiry timestamp, OR
+        * expiry_datetime = parse(response.body["expires_at"]) # in this option the API tells us exactly when access token expires
+
+        :return: a tuple of (access_token, either token_lifespan_in_seconds or datetime_of_token_expiry)
```
girarda, line 106: oof this is a confusing interface

airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
```diff
@@ -92,14 +92,34 @@ def get_grant_type(self) -> InterpolatedString:
     def get_refresh_request_body(self) -> Mapping[str, Any]:
         return self._refresh_request_body.eval(self.config)
 
+    def refresh_access_token(self) -> Tuple[str, Any]:
+        """
+        Returns the refresh token and its lifespan in seconds
```
girarda, line 97: this comment doesn't match the actual interface

airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
sherifnada: I agree - should it be part of the scope of this PR?

airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
sherifnada: yup....

airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
sherifnada: thanks for the callout, copy pasta

girarda (approved):
approved pending an issue for cleaning up the declarative oauth interface

airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
sherifnada: I think the name of this method is accurate -- it's expected to do whatever processing necessary to store a timestamp of when the access token expires

airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
sherifnada: FWIW I agree, but this PR just documents the behavior which already exists

sherifnada:
@girarda issue [here](https://github.com/airbytehq/airbyte/issues/23926) documenting the larger problem

Issue #22575, opened 2023-02-08
Declarative connectors management: Add declarative definition auto-updater on startup


flash1293:
Similar to #22453, docker image tags for existing declarative source definitions need to be kept in sync with the used CDK version for both OSS and Cloud Airbyte instances.

This is done by moving the existing `CDK_VERSION` file into a central location and bundling it together with the application (see `LocalDefinitionsProvider` for regular connector definitions).

Also update the publish api implementation in the handler to use the cdk version provider instead of using a hardcoded value.

As the locally bundled seed file can only change when updating the distribution version of the Airbyte platform, it is sufficient to run the update once on startup as part of the bootloader in the `DefaultPostLoadExecutor` class.

There should be no SQL join between actor_definition table AND declarative manifest tables to allow splitting these tables into different databases in a future step. Instead, all joining should be done in the application layer.

## Acceptance criteria

* For OSS and cloud, all declarative source definitions are updated to the version specified by the local `CDK_VERSION` file on startup/redeploy of the Airbyte instance.

flash1293:
@flash1293 Create a new issue for using the seed file for connector builder server and webapp CDK version as well

flash1293:
Implementation note:
* Best way to make sure it runs on cloud as well is if the cloud bootloader wraps the OSS bootloader - make sure this is the case and if not add an extra call to the auto-updater logic defined in this issue

flash1293:
Story point estimation: 3

PR #23935, opened 2023-03-10
Adjust CDK publish command

flash1293:
## What

Part of https://github.com/airbytehq/airbyte/issues/22575

With the centralized CDK version file, we need to make sure its set correctly, also within the python lockfile.

## How

Adjust the workflow accordingly and use pip-compile when creating the PR.

Note: The legacy connector builder server change causes the python stuff - not sure whether we still need this as it's close to be removed anyway.


maxi297 (approved):
LGTM!

PR #23965, opened 2023-03-13
Destination Databricks: Support managed tables

### LINKED ISSUE #23036: Destination Databricks: support managed tables
Currently, Databcirks destination supports only external tables, which are harder to set up than managed tables. We should add support for managed tables (https://docs.databricks.com/lakehouse/data-objects.html#what-is-a-managed-table).

suhomud:
## What
Destination Databricks: Support new data source type [Managed tables](https://github.com/airbytehq/airbyte/pull/23965). 

## How
In order to support Databricks Managed tables new `DatabricksManagedTablesDestination.java` is  implemented based on `AbstractJdbcDestination.java` with a slight modification of `JdbcSqlOperations`. On the UI we can see the new data source type [Recommended] Managed tables without datasource property since it has to be modified on the Databricks side: 
image

 
### Additional changes
- In origin implementation the Databricks port was ignored for some reason. Updated `DATABRICKS` url in `DatabaseDriver.java` to custom port.
- `spec.json` was updated with new field `database` to specify a Databricks `Catalog`. So now it looks like: 
image

- Some refactoring was added to navigate through connectors's code more easily.  
- Added Utility clases to avoid code duplications.
- Nice Java 14 `Records Class` feature was added for `DatabricksDestinationConfig.java`. 

## Recommended reading order
1. `spec.json` 
2. `DatabricksManagedTablesDestination.java` - The new datasource type based on `AbstractJdbcDestination.java`.
3. `DatabricksSqlOperations.java`
4. `DatabricksStorageType.java`
5. `DatabaseDriver.java` - New url to support custom port.
6. `DatabricksDestinationConfig.java` - Modified default schema to `default` instead of `public` since Databricks creates `default` schema when Catalog is created. Also changed it to Java record type which looks much better then all of those getters-setters staff.
7. `DatabricksStorageConfigProvider.java` - Renamed and refactored. Note: `return switch` is feature of Java 17.
8. `DatabricksDestinationResolver.java`
9. `DatabricksExternalStorageBaseDestination.java` - Renamed from `DatabricksBaseDestination` since now it is used by external data source type only.
10. `DatabricksConstants.java` - Made utility class with constants
11. `DatabricksDatabaseUtil.java` - Utility class with shared code
12. `DatabricksS3DestinationAcceptanceTest.java` - As discussed was disabled for now since it looks like the external tables are not fully working. Should be revisit after external data source will be fixed.
13. `DatabricksUtilTest.java` - Integration tests util. Note: there is an issue with retrieving the data which was magically fixed. See code comments to know the details. _This fix exists only for integration tests_
14. `DatabricksManagedTablesDestinationAcceptanceTest.java` - Note: if you want to run it locally you should take a new secret from google secret manager `SECRET_DESTINATION-DATABRICKS__MANAGED_TABLES_CREDS`. Additionally all Databricks secrets were updated.
15. Others changes are mostly related to new folder structure, re-naming, small refactoring changes. 

## 🚨 User Impact 🚨
Since the `spec.json` has changed I think it should break change for existing connections if there is any.


airbyte-integrations/connectors/destination-databricks/Dockerfile
```diff
@@ -16,5 +16,5 @@ ENV APPLICATION destination-databricks
 
 COPY --from=build /airbyte /airbyte
 
-LABEL io.airbyte.version=0.3.1
+LABEL io.airbyte.version=0.3.2
```
grishick, line 19: let's bump this to 1.0.0, because this is a major breaking change

suhomud: Updated

grishick (approved):

airbyte-integrations/connectors/destination-databricks/src/test-integration/java/io/airbyte/integrations/destination/databricks/DatabricksUtilTest.java
```diff
@@ -0,0 +1,47 @@
...
+import org.jooq.SQLDialect;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DatabricksUtilTest {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DatabricksUtilTest.class);
+
+  /**
+   * Magic Databricks JDBC param. This param exists because of issue with fetching data via SELECT. It produces next error: Error occured while
```
ameyabapat-bsft, line 19: Hi @suhomud I am working on writing databricks source. I also faced this error. It got resolved by adding jvm args `'--add-opens=java.base/java.nio=ALL-UNNAMED'` in `build.gradle`.
Ex: 
```   
applicationDefaultJvmArgs = ['-XX:+ExitOnOutOfMemoryError', '-XX:MaxRAMPercentage=75.0', '--add-opens=java.base/java.nio=ALL-UNNAMED']
```

Issue #23778, opened 2023-03-06
Docs: Update data residency info


sophia-wiley:
Discover and check workflows now run in the default data residency. This means that a user can ensure more of their data is processed in the default data residency of their choosing. Users must choose the default data residency before setting up the connection because when we run the discover or check workflows, the connection has not been established yet, so we refer to the default data residency settings. 

Add something like this to the [Manage data residency](https://docs.airbyte.com/cloud/managing-airbyte-cloud/manage-data-residency) (?) documentation:

"To ensure that your data is processed in the correct region, set your default data residency before creating a new connection."

Make a ticket for OSS docs.

sophia-wiley:
Related to this [issue](https://app.zenhub.com/workspaces/platform-compose-6259ee50ab53b4001d625d79/issues/gh/airbytehq/airbyte-platform-internal/4990)

sophia-wiley:
Done in [#23990](https://github.com/airbytehq/airbyte/pull/23990) available [here](https://docs.airbyte.com/cloud/managing-airbyte-cloud/manage-data-residency/)

PR #23990, opened 2023-03-13
Docs: added info about default data residency

sophia-wiley:
Added info to docs about default data residency because the discover and check workflows now run in the default data residency. 

Related to #23778 and [#4990](https://app.zenhub.com/workspaces/platform-compose-6259ee50ab53b4001d625d79/issues/gh/airbytehq/airbyte-platform-internal/4990)


docs/cloud/managing-airbyte-cloud/manage-data-residency.md
```diff
@@ -8,7 +8,7 @@ Default data residency allows you to choose where your data is processed.
 
 :::note 
 
-Configuring default data residency only applies to new connections and does not affect existing connections.   
+To ensure your data is processed in the correct region, set your default data residency before creating a new connection as some workflows rely on the default data residency. When you set the default data residency, it applies to new connections and does not affect existing ones. If you do not set the default data residency, the [Airbyte Default](https://docs.airbyte.com/cloud/getting-started-with-airbyte-cloud/#united-states-and-airbyte-default) region is used.
```
xiaohansong, line 11: Looks good! a minor comment though:

Should we be more specific, such as mentioning some workflows during **connection creation** will rely on the data residency, maybe even giving examples such as workflow for fetching data schema or checking connection could be setup, so user could understand why it's important to setup their data residency before creating a connection? (just my 2 cents, feel free to push back if you think this is too much detailed information for users)

malikdiarra: I agree with Xiaohan here: specifically we could say "testing sources and destinations and fetching schema rely on the default data ..." instead of "some workflows"

sophia-wiley: Thank you for the feedback! I updated the text with info about testing sources and destinations and fetching the schema :)

docs/cloud/managing-airbyte-cloud/manage-data-residency.md
```diff
@@ -4,21 +4,15 @@ In Airbyte Cloud, you can set the default data residency and choose the data res
 
 ## Choose your default data residency
 
-Default data residency allows you to choose where your data is processed.
+Default data residency allows you to choose where your data is processed. Set the default data residency before creating a new connection so workflows that rely on the default data residency, such as fetching the schema or testing the source or destination, can process data in the correct region.
```
Amruta-Ranade, line 7: we should add a note to the new connection workflow as well

docs/cloud/managing-airbyte-cloud/manage-data-residency.md
```diff
@@ -4,21 +4,15 @@ In Airbyte Cloud, you can set the default data residency and choose the data res
 
 ## Choose your default data residency
 
-Default data residency allows you to choose where your data is processed.
+Default data residency allows you to choose where your data is processed. Set the default data residency before creating a new source or connection so workflows that rely on the default data residency, such as fetching the schema or testing the source or destination, can process data in the correct region.
```
sophia-wiley, line 7: I changed it to "a new source or connection" instead of just connection since these workflows run before creating a new source or connection.

docs/cloud/getting-started-with-airbyte-cloud.md
```diff
@@ -83,6 +89,12 @@ If you need to use [cron scheduling](http://www.quartz-scheduler.org/documentati
 
 To set up a connection:
 
+:::note
+
+Set your [default data residency](https://docs.airbyte.com/cloud/managing-airbyte-cloud/manage-data-residency#choose-your-default-data-residency) before creating a new connection to ensure your data is processed in the correct region.
+
```
sophia-wiley, line 95: @Amruta-Ranade Is it overkill to have this note twice in the Getting Started doc? I added it to both the "To set up a source" and "To set up a connection" sections because it applies to both, but if the users are following the instructions they would have probably set their default data residency before they set up their source.

xiaohansong (approved):
LGTM, thanks!

docs/cloud/getting-started-with-airbyte-cloud.md
Amruta-Ranade: We do need the note in both places. Users won't always read the instructions in order. What if they already had the source set up and are reading this guide to figure out how to set up a connection?

Amruta-Ranade (approved):

PR #23992, opened 2023-03-13
CAT: handle discovery error on previous catalog retrieval

alafanechere:
## What
We can't retrieve the previous connector version discovered catalog when the connector's current configuration is not compatible with the previous connector version. This is expected if backward incompatible changes where introduced on a branch and should make the backward compatible tests fail.

**But even if the backward compatibility tests are bypassed the tests are still failing because we're using a fixture to check if the backward compatibility tests can be skipped. And in this fixture, we check the content of the previous catalog**.

## How
Handle discovery error in the `previous_discovered_catalog_fixture`: a None catalog is returned if the discovery failed.
This will make the `skip_backward_compatibility_tests_fixture` execution not fail.


davydov-d (approved):

davydov-d:
Looks good to me, thanks!

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/conftest.py
```diff
@@ -308,7 +308,13 @@ def previous_discovered_catalog_fixture(
     if not cache_discovered_catalog:
         previous_cached_schemas.clear()
     if not previous_cached_schemas:
-        output = previous_connector_docker_runner.call_discover(config=connector_config)
+        try:
+            output = previous_connector_docker_runner.call_discover(config=connector_config)
+        except errors.ContainerError:
+            logging.warning(
+                "\n DISCOVER on the previous connector version failed. This could be because the current connector config is not compatible with the previous connector version."
```
evantahler, line 315: Could we log what we think the previous version is? That might help debugging.

evantahler (approved):

alafanechere:

PR #24042, opened 2023-03-14
[source-instagram] bump version to 1.0.2 (no changes)

erohmensing:
# Actual changes in this pr

Bumps version to 1.0.2 since the image was already published when we decided we didn't want that version to ever see cloud. We determined that the failures were likely transient, and therefore did not want to mask them. **[The 1.0.2 tag was manually updated to be a copy of 1.0.1 tag](https://airbytehq-team.slack.com/archives/C04TSP9JM8A/p1678780159849929?thread_ts=1678774672.384079&cid=C04TSP9JM8A).**

# The below changes were reverted and never merged / the published version was overridden

## What
* Stopgap fix for https://github.com/airbytehq/oncall/issues/1640
* Avoid `KeyError: 'date'` in `get_updated_state`

## How
If the previous record doesn't have a `date` param, use `_start_date` instead of previous record's date for getting the latest cursor value for updated state

## 🚨 User Impact 🚨
Failed syncs will succeed, but will temporarily act as full refresh if they don't have a `date` field.


airbyte-integrations/connectors/source-instagram/source_instagram/streams.py
```diff
@@ -251,7 +251,7 @@ def _state_has_legacy_format(self, state: Mapping[str, Any]) -> bool:
 
     def get_updated_state(self, current_stream_state: MutableMapping[str, Any], latest_record: Mapping[str, Any]):
         """Update stream state from latest record"""
-        record_value = latest_record[self.cursor_field]
+        record_value = latest_record.get(self.cursor_field, str(self._start_date))
         account_id = latest_record.get("business_account_id")
         state_value = current_stream_state.get(account_id, {}).get(self.cursor_field) or record_value
         max_cursor = max(pendulum.parse(state_value), pendulum.parse(record_value))
```
erohmensing, line 257: As far as I can tell, if the `date` is missing in the previous record and current stream state, this will essentially cause a full refresh, since we'll output with the cursor being equal to the start date. 

That seems acceptable-ish as a stopgap but definitely not as a long term solution for an incremental stream. 

Is there any chance of this causing some sort of endless loop where we keep re-syncing data since we keep outputting state messages that don't increase?

sherifnada: I think this wouldn't be an infinite loop per say, but it just won't advance the state. So every future sync will sync that same data.

sherifnada (approved):
This seems fine as an emergency fix. @davydov-d could you or someone from the team take the action item to figure out the correct long-term fix?

davydov-d (approved):

PR #24043, opened 2023-03-14
Source Instagram: do not emit incomplete records

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1640

## How
 - do not emit record if it's incomplete so that
   - no checkpointing would be done
   - sync would not be failed
 - log a warning with base request parameters when no data is received
 - exit stream gracefuully


davydov-d:
> /test connector=connectors/source-instagram
> 
> > clock2 connectors/source-instagram https://github.com/airbytehq/airbyte/actions/runs/4415128542
> > x connectors/source-instagram https://github.com/airbytehq/airbyte/actions/runs/4415128542
> > bug https://gradle.com/s/3tpfo34p4mvg6
> 
> ### Build Failed
> Test summary info:
> 
> ```
> =========================== short test summary info ============================
> FAILED test_core.py::TestDiscovery::test_backward_compatibility[inputs0] - As...
> ERROR test_core.py::TestSpec::test_backward_compatibility[inputs0] - docker.e...
> SKIPPED [1] ../usr/local/lib/python3.9/site-packages/connector_acceptance_test/tests/test_incremental.py:212: Skipping new incremental test based on acceptance-test-config.yml
> ========= 1 failed, 34 passed, 1 skipped, 1 error in 157.47s (0:02:37) =========
> ```

:point_up: we're having issues with published images of source-instagram. Tagging @erohmensing @sherifnada here as well

lazebnyi (approved):
Nice!

airbyte-integrations/connectors/source-instagram/source_instagram/streams.py
```diff
@@ -222,6 +230,9 @@ def stream_slices(
                 continue
             for since in pendulum.period(start_date, self._end_date).range("days", self.days_increment):
                 until = since.add(days=self.days_increment)
+                if self.time_to_exit_gracefully:
+                    self.logger.warning("Exiting gracefully")
```
erohmensing, line 234:
```suggestion
                    self.logger.info(f"Stopping syncing stream '{self.name}'")
```

airbyte-integrations/connectors/source-instagram/source_instagram/streams.py
```diff
@@ -203,7 +205,13 @@ def read_records(
             if not insight_record.get(self.cursor_field):
                 insight_record[self.cursor_field] = insight.get("values")[0]["end_time"]
 
-        yield insight_record
+        complete_records = [insight_record]
+        # if insight_list is empty, we don't want to yield an incomplete record and want to stop syncing this stream gracefully
+        if not insight_list:
+            complete_records = []
+            self.logger.warning(f"No data received for base params {json.dumps(base_params)}. It is time to exit gracefully.")
```
erohmensing, line 212:
```suggestion
            # https://developers.facebook.com/docs/instagram-api/guides/insights/
            # If insights data you are requesting does not exist or is currently unavailable
            # the API will return an empty data set instead of 0 for individual metrics.
            self.logger.warning(f"No data received for base params {json.dumps(base_params)}. Since we can't know whether there is no data or the data is temporarily unavailable, stop syncing so as not to miss temporarily unavailable data.")
```

erohmensing: @davydov-d does this describe the situation, would you say? Was working to make it clearer both for the developer and the user what was leading to the early exit. 

If this is indeed the case, couldn't a true negative (the data is not unavailable, but the data doesn't exist/there's not data to report able) for a given stream slice stop us from ever advancing the state because we check it every time?

erohmensing: Or i guess since we don't have per-stream state, we'd still be emitting a later state and not going back to re-check this... right 🤔

erohmensing (commented):
I'm a bit lost on how stopping the stream early affects subsequent runs in the case that we actually might expect the data to be empty for legit reasons

airbyte-integrations/connectors/source-instagram/source_instagram/streams.py
erohmensing: @sherifnada maybe you also have insight here with regards to the fix you were recommending: 
> It feels like the fix might be to gracefully exit syncing that stream so that this behavior (re-sync in a few hours) happens naturally
> i.e: don’t checkpoint state but also don’t fail the sync

davydov-d: @erohmensing yes, thanks! I dont think it could stop us advancing the state because all the syncs (3 on cloud + 1 integration test) used to failed on the last slice (i.e. fetching today's data). Once the latest data becomes available (I think it's a matter of hours), the state should advance and the sync should succeed

erohmensing: In that case this thread is resolved, still might be nice to fail if the cursor gets too old just in case though

erohmensing (approved):
LGTM, would still like @sherifnada to take a look if possible though

sherifnada (approved):
LGTM though I suggest we use caught exceptions rather than this Boolean. Please leave a source code comment explaining the issue for the next person looking at this too 😅

airbyte-integrations/connectors/source-instagram/source_instagram/streams.py
```diff
@@ -165,6 +166,7 @@ class UserInsights(InstagramIncrementalStream):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self._end_date = pendulum.now()
+        self.time_to_exit_gracefully = False
```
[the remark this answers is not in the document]
davydov-d: I chose boolean over throwing an exception because in the latter case I would have to catch that exception in the overridden source method `_read_incremental`/`_read_full_refresh`/etc with tons of code and new logic which would affect readability badly.
Variable renamed, code comment added.

PR #24051, opened 2023-03-14
Source Stripe: cache customers

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1632

## How
 - cache `Customers` stream
 - do not make requests to /<customer_id>/balance_transactions if customer has zero balance and no invoices
 - refactor: move slicing logic to the `stream_slices` methods

## 🚨 User Impact 🚨
 - I expect one of `bank_accounts` / `customer_balance_transactions` streams to cache the customers so the next one wont have to read them again. This should speed up one of two streams (probably the one to be synced later) or both of them if the `customers` stream is synced in a full refresh mode beforehand.
  - `Customer Balance Transactions` stream should also speed up if there are lots of users with zero balance


lazebnyi (approved):

airbyte-integrations/connectors/source-stripe/source_stripe/streams.py
```diff
@@ -196,23 +190,26 @@ def path(self, **kwargs) -> str:
...
+    def stream_slices(
+        self, *, sync_mode: SyncMode, cursor_field: List[str] = None, stream_state: Mapping[str, Any] = None
+    ) -> Iterable[Optional[Mapping[str, Any]]]:
+        parent_stream = Customers(authenticator=self.authenticator, account_id=self.account_id, start_date=self.start_date)
+        slices = parent_stream.stream_slices(sync_mode=SyncMode.full_refresh)
         for _slice in slices:
-            for customer in customers_stream.read_records(sync_mode=SyncMode.full_refresh, stream_slice=_slice):
-                yield from super().read_records(stream_slice={"customer_id": customer["id"]}, **kwargs)
+            for customer in parent_stream.read_records(sync_mode=SyncMode.full_refresh, stream_slice=_slice):
+                if customer["next_invoice_sequence"] == 1 and customer["balance"] == 0:
```
erohmensing, line 210: Please add a comment here about why we're skipping this

airbyte-integrations/connectors/source-stripe/source_stripe/streams.py
```diff
@@ -562,8 +566,8 @@ class CheckoutSessionsLineItems(SingleEmptySliceMixin, IncrementalStripeStream):
 
     cursor_field = "checkout_session_expires_at"
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
erohmensing, line 570: Just wondering, is adding these just for good measure? Don't see us passing any non-keyword args

airbyte-integrations/connectors/source-stripe/source_stripe/streams.py
```diff
@@ -108,13 +108,6 @@ def read_records(
                 self.logger.error(f"Syncing stream {self.name} is failed, due to {error_code}. Full message: {error_message}")
 
 
-class SingleEmptySliceMixin(object):
-    def stream_slices(
-        self, *, sync_mode: SyncMode, cursor_field: List[str] = None, stream_state: Mapping[str, Any] = None
-    ) -> Iterable[Optional[Mapping[str, Any]]]:
-        return [{}]
-
-
```
erohmensing, line 117: Nice refactor here! Much clearer IMO

erohmensing (approved):
> I expect one of bank_accounts / customer_balance_transactions streams to cache the customers so the next one wont have to read them again. 

I see how it would be cached for `customer_balance_transactions` but I don't see where `bank_accounts` uses the `Customers` stream. What am I missing?

erohmensing:
Nevermind, answered my own question - was looking at ExternalAccounts, not BankAccounts 😄

airbyte-integrations/connectors/source-stripe/source_stripe/streams.py
davydov-d: Done

PR #24057, opened 2023-03-14
Add a note about publishing with dockerx

evantahler:


erohmensing (approved):
Thanks for adding. Should we make a note about making sure to push the spec to the cache? I'm worried this wasn't done here

Issue #24017, opened 2023-03-13
Validate metadata file(s)


bnchrch:
1. Create a jsonschema for metadata files
2. Wrap a cli program around it that will validate any file against the schema
3. Run the cli program only when a metadata.yml file is found in a PR

PR #24080, opened 2023-03-15
Add validation tool for new metadata files

bnchrch:
## What
We need to be able to validate metadata files.

Part one of two for #24017

## How
This adds a library / cli tool to do so

## Notes for reviewer
1. All the python model classes are generated from a json spec
2. All json specs create one model currently (may expand in the future)
3. We are not using this anywhere yet
4. The way we are using these jsonschema files is temporary, until we can get an answer on how the Prod engineering wants to handle types that may cross projects

## Recommended reading order
1. Start with the readme
2. Then look at the test files to confirm you agree with the schema choices
3. Afterwards the bulk of the logic exists in [metadata_validator.py](https://github.com/airbytehq/airbyte/compare/bnchrch/poc-dagster-gcs-sensor...bnchrch/validate-metadata?expand=1#diff-2eeb2e1bbbf00a5b6ce2f9a0e38e5364de2b0910f229d18d88d9c0848efa4680)


airbyte-ci/connectors_ci/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
evantahler, line 1: Nice!

airbyte-ci/connectors_ci/metadata_service/lib/tests/fixtures/valid/metadata_catalog_complex_override.yaml
evantahler, line 1: Nice way to test this!

evantahler (approved):
Big 👍 from me, well done.

I'll bet in the future we will wan to share some of these YAML definitions with the platform to prevent drift... we can sort that out in a future PR.

PR #24117, opened 2023-03-15
Add semantic versioning guidelines for contributions

erohmensing:
* Add to documentation
* Update PR template to reference

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -23,10 +23,12 @@ Expand the relevant checklist and delete the others.
 - [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
 - [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
 - [ ] Code reviews completed
+- [ ] Connector version is set to `0.0.1`
+    - [ ] `Dockerfile` has version `0.0.1`
 - [ ] Documentation updated
     - [ ] Connector's `README.md`
     - [ ] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
-    - [ ] `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
+    - [ ] `docs/integrations/<source or destination>/<name>.md` including changelog with an entry for the initial version. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
     - [ ] `docs/integrations/README.md`
     - [ ] `airbyte-integrations/builds.md`
 - [ ] PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests)
@@ -53,10 +55,14 @@ If this is a community PR, the Airbyte engineer reviewing this PR is responsible
 - [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
 - [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
 - [ ] Code reviews completed
+- [ ] Connector version has been incremented
+    - [ ] Version has been bumped according to our [Semantic Versioning for Connectors](TODO LINK) guidelines
+    - [ ] `Dockerfile` has updated version
 - [ ] Documentation updated
     - [ ] Connector's `README.md`
     - [ ] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
-    - [ ] Changelog updated in `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
+    - [ ] Changelog updated in `docs/integrations/<source or destination>/<name>.md` with an entry for the new version. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
+
 - [ ] PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests)
 
 ### Airbyter
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -56,7 +56,7 @@ If this is a community PR, the Airbyte engineer reviewing this PR is responsible
 - [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
 - [ ] Code reviews completed
 - [ ] Connector version has been incremented
-    - [ ] Version has been bumped according to our [Semantic Versioning for Connectors](TODO LINK) guidelines
+    - [ ] Version has been bumped according to our [Semantic Versioning for Connectors](https://docs.airbyte.com/contributing-to-airbyte/semantic-versioning-for-connectors) guidelines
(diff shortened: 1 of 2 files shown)


docs/contributing-to-airbyte/README.md
```diff
@@ -72,6 +73,31 @@ The CDK currently does not support creating destinations, but it will very soon.
...
 
+### Semantic Versioning for Connectors
+
+Changes to connector behavior should always be accompanied by a version bump and a changelog entry. We use [semantic versioning](https://semver.org/) to version changes to connectors. Since connectors are a bit different from APIs, we have our own take on semantic versioning, focusing on maintaining the best user experience of using a connector.
+
+- Major: version when you make a change which requires manual intervention (update to config or configured catalog) for an existing connection to continue to succeed, or one in which data that was previously being synced will no longer be synced
+- Minor: version when you add user-facing functionality in a backwards compatible manner
+- Patch: version when you make backwards compatible bug fixes or performance improvements
+
+## Examples
```
erohmensing, line 84: Any more to add? What about e.g. renaming a stream or field? Those would be breaking as they're essentially add + deletes right? 

This is an interesting format to keep track of which changes are breaking and which ones we could potentially automate so that they're no longer breaking :)

docs/contributing-to-airbyte/README.md
```diff
@@ -35,11 +35,12 @@ Before jumping into the code please first:
 This will enable our team to make sure your contribution does not overlap with existing works and will comply with the design orientation we are currently heading the product toward.
 If you do not receive an update on the issue from our team, please ping us on [Slack](https://slack.airbyte.io)!
 
-#### 2. Let's code
+### 2. Let's code
 1. To contribute to a connector, fork the [Connector repository](https://github.com/airbytehq/airbyte). To contribute to the Airbyte platform, fork our [Platform repository](https://github.com/airbytehq/airbyte-platform).  
 2. Open a branch for your work.
 3. Code, and please write **tests**.
 4. Ensure all tests pass. For connectors, this includes acceptance tests as well. 
+5. For connectors, make sure to increment the connector's version according to our [Semantic Versioning](#semantic-versioning-for-connectors) guidelines.
```
erohmensing, line 43: Not sure if this is the correct way to link to an anchor on the same page

docs/contributing-to-airbyte/README.md
```diff
@@ -22,11 +22,11 @@ We appreciate first time contributors and we are happy to assist you in getting
 
 Here is a list of easy [good first issues](https://github.com/airbytehq/airbyte/labels/good%20first%20issue) to do.
 
-## Areas for contributing
+## Steps to contributing
```
erohmensing, line 25: Don't see any links to this anchor from anywhere else, think it's ok to rename

docs/contributing-to-airbyte/README.md
erohmensing: - Adding a non-required param to spec: does that break configs?

erohmensing: Maybe we should also consider users to think about backwards compatibility when e.g. renaming a parameter (still supporting the old name)

marcosmarxm (approved):

.github/pull_request_template.md
```diff
@@ -53,10 +55,14 @@ If this is a community PR, the Airbyte engineer reviewing this PR is responsible
 - [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
 - [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
 - [ ] Code reviews completed
+- [ ] Connector version has been incremented
+    - [ ] Version has been bumped according to our [Semantic Versioning for Connectors](https://docs.airbyte.com/contributing-to-airbyte/semantic-versioning-for-connectors) guidelines
```
evantahler, line 59: 👍 added to PR template

docs/contributing-to-airbyte/README.md
evantahler: It is!

docs/contributing-to-airbyte/README.md
evantahler: Oh! Maybe add a note at the end like:

If you are not sure what type of change your commit is, ask us! Call this out in your PR and we are happy to help out.

evantahler: I think:

> Adding a non-required param to spec: minor?
minor!

> turning a single param into a oneOf?
I think it sort of depends on where we land [here](https://docs.google.com/document/d/1tbMeG_v29zVcza6JNhw7-oY50H7aY1AbVvZr9yzFmVw/edit#).
... but I think for now it is a breaking change.

evantahler (approved):
Thank you!

docs/contributing-to-airbyte/README.md
erohmensing: > Oh! Maybe add a note at the end

True, good callout! Then we can encourage them to add it to the list too :)

erohmensing: Added `Trying to contribute, and don't see the change you want to make in this list? Call it out in your PR and your reviewer will help you pick the correct type of version change. Feel free to contribute the results back to this list!`

erohmensing: > I think it sort of depends on where we land [here](https://docs.google.com/document/d/1tbMeG_v29zVcza6JNhw7-oY50H7aY1AbVvZr9yzFmVw/edit#).
... but I think for now it is a breaking change.

yup agreed. I'm hoping that config control messages will reduce the rows in this table that constitute major changes, but for now I think that it is indeed breaking

docs/contributing-to-airbyte/README.md
erohmensing: sweet!

docs/contributing-to-airbyte/README.md
erohmensing: @evantahler what about renaming a stream/field?

evantahler: > @evantahler what about renaming a stream/field?

I think that's breaking too.  A 'rename' is an addition and a removal until schema migrations get smarter.  I think.  CC @andyjih

docs/contributing-to-airbyte/README.md
```diff
@@ -72,6 +73,35 @@ The CDK currently does not support creating destinations, but it will very soon.
...
+
+Changes to connector behavior should always be accompanied by a version bump and a changelog entry. We use [semantic versioning](https://semver.org/) to version changes to connectors. Since connectors are a bit different from APIs, we have our own take on semantic versioning, focusing on maintaining the best user experience of using a connector.
+
+- Major: a version in which a change is made which requires manual intervention (update to config or configured catalog) for an existing connection to continue to succeed, or one in which data that was previously being synced will no longer be synced
+- Minor: a version that introduces user-facing functionality in a backwards compatible manner
+- Patch: a version that introduces backwards compatible bug fixes or performance improvements
+
+## Examples
+
+Here are some examples of code changes and their respective version changes:
```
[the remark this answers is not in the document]
bleonard: I love the examples. It makes things much more concrete. Good work.

docs/contributing-to-airbyte/README.md
```diff
@@ -72,6 +73,35 @@ The CDK currently does not support creating destinations, but it will very soon.
...
+| Changing the datatype for a column of a stream in a connector's `catalog`                     | Users will have to update that data type in their `catalog`, data that was being synced will have changed format | Major          |
+| Adding a non-required parameter to a connector's `spec`                                       | Users will have the option to use the required parameter in the future                                           | Minor          |
+| Adding a stream in a connector's `catalog`                                                    | Additional data will be synced                                                                                   | Minor          |
+| Adding a column to a stream's schema in a connector's `catalog`                               | Additional data will be synced                                                                                   | Minor          |
+| Updating the format of the connector's `STATE`                                                | Incremental streams will automatically run a full refresh only for the next sync                                 | Patch          |
```
bleonard, line 100: I think this is a tricky one. Full refreshes can be quite painful. I would love to see connectors migrating the state if at all possible.

bleonard: That being said, I don't have a strong opinion it fits into the semantic versioning. So carry on.

erohmensing: I would love to see that too! I think that's a good input into our conversation about improving the way we automate things, but at the moment, this is unfortunately the current outcome of making the change. 

Depending on how painful we think the experience of running a full refresh is, we could update our notion of majors:
> - Major: a version in which a change is made which requires manual intervention (update to config or configured catalog) for an existing connection to continue to succeed, or one in which data that was previously being synced will no longer be synced

to include things that force full refreshes. I kept our reasoning behind the categorization (`we have our own take on semantic versioning, focusing on maintaining the best user experience of using a connector`) pretty vague so that we can play around with it.

Issue #23098, opened 2023-02-15
connector-acceptance-test: incremental tests failures on java sources


alafanechere:
We have failing incremental tests for `source-postgres` and `source-snowflake`. 
They are both failing at running `test_two_sequential_reads`.

I suspect the root cause of it to be that the `records_with_state` function can't handle per-stream state.
I think it went under the radar because API connectors emit legacy + per_stream states while java sources only emit per-stream state. 

Related PR: https://github.com/airbytehq/airbyte/pull/23014

These failure happened after the addition of this new (legit) assertion in https://github.com/airbytehq/airbyte/pull/22353.
````
 assert parsed_records_1, "At least one valid state should be produced, given a cursor path"
````

I suspect this assertion highlighted a silenced failures and unexpected behavior in `test_two_sequential_reads`

bnchrch:
A little more investigation on my side

I ran the tests locally against source-postgres and set a breakpoint in `test_two_sequential_reads`.

Below is the result of the failing values for records, state, stream path etc.

https://app.warp.dev/block/Q3yKgMkN7NomlmVm68NuRj
```py

(Pdb++) records_1
13[AirbyteMessage(type=<Type.RECORD: 'RECORD'>, log=None, spec=None, connectionStatus=None, catalog=None, record=AirbyteRecordMessage(namespace='public', stream='id_and_name', data={'id': 1, 'name': 'picard'}, emitted_at=1676580707185), state=None, trace=None, control=None), AirbyteMessage(type=<Type.RECORD: 'RECORD'>, log=None, spec=None, connectionStatus=None, catalog=None, record=AirbyteRecordMessage(namespace='public', stream='id_and_name', data={'id': 2, 'name': 'crusher'}, emitted_at=1676580707185), state=None, trace=None, control=None), AirbyteMessage(type=<Type.RECORD: 'RECORD'>, log=None, spec=None, connectionStatus=None, catalog=None, record=AirbyteRecordMessage(namespace='public', stream='id_and_name', data={'id': 3, 'name': 'vash'}, emitted_at=1676580707185), state=None, trace=None, control=None)]


14(Pdb++) latest_state
15{'id_and_name': {'cursor_record_count': 1, 'stream_name': 'id_and_name', 'stream_namespace': 'public', 'cursor_field': ['id'], 'cursor': '3'}}

16(Pdb++) stream_mapping
[...]

PR #24127, opened 2023-03-16
Simplify postgres and GitHub forms

### LINKED ISSUE #23696: Migrate 1-2 connectors to new simplify format
## Context

As part of the simplify connector project, we would like to migrate a couple of connectors to use the new features to serve as an example of what a simpler connector can look like.

## Acceptance Criteria
- Migrate the github and postgres connectors to use the new simplify features so that they look like they do in these designs: https://www.figma.com/proto/id4s6OcakcKAIdroQdk7Iv/SIMPLIFY-CONFIGURATION?page-id=160%3A5361&node-id=160%3A5554&viewport=676%2C390%2C0.5&scaling=min-zoom&starting-point-node-id=160%3A5554

lmossman:
## What
Resolves https://github.com/airbytehq/airbyte/issues/23696

As part of the simplify project definition of done, we want to migration a couple of connectors to use the new fields. 

## How
Migrates the postgres and github source connectors to use the new simplify fields, so that they match the designs linked in the above issue.

**Note**: This PR should not be merged until the outstanding PRs for the simplify project are merged:
https://github.com/airbytehq/airbyte-platform-internal/pull/4833
https://github.com/airbytehq/airbyte-platform-internal/pull/5091
https://github.com/airbytehq/airbyte-platform-internal/pull/5171

In fact, the full effect of these changes won't be visible until all of those are merged, because changes like pattern descriptors and group titles are split into separate PRs at the moment.


airbyte-integrations/bases/base-java/src/main/resources/ssh-tunnel-spec.json
```diff
@@ -2,6 +2,7 @@
   "type": "object",
   "title": "SSH Tunnel Method",
   "description": "Whether to initiate an SSH tunnel before connecting to the database, and if so, which kind of authentication to use.",
+  "group": "security",
```
lmossman, line 5: Unfortunately the ssh tunnel spec field is in this separate JSON file and gets copied into all of the JDBC source connectors.

Adding the `group` here is fine for postgres because as part of this PR, this group is also set on the other security fields. But this will be a problem for the non-postgres connectors that don't have any groups set, as this will cause the SSH tunnel method to go into its own group. 

Maybe the best approach here is to also migrate the other JDBC connectors to have groups like we did with postgres here? Open to other suggestions though

flash1293: This is a little annoying because we would need to make sure the groups are set completely, otherwise things will look weird (as named groups are sorted to the top).

This is how the injection is happening for postgres:
https://github.com/airbytehq/airbyte/blob/009940404cd90e72cb42f6b2ab91549ced54e7d5/airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java#L118

https://github.com/airbytehq/airbyte/blob/2e099acc52aeddc172c5ee66ed98426c69449a4a/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/ssh/SshHelpers.java#L23

https://github.com/airbytehq/airbyte/blob/2e099acc52aeddc172c5ee66ed98426c69449a4a/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/ssh/SshWrappedSource.java#L37

Maybe we can parameterize `SshWrappedSource` and `injectSshIntoSpec` with an optional resource file name and create `grouped-ssh-tunnel-spec.json` and use it for postgres specifically?

airbyte-integrations/connectors/source-postgres/src/main/resources/spec.json
```diff
@@ -10,7 +10,8 @@
         "title": "Host",
         "description": "Hostname of the database.",
         "type": "string",
-        "order": 0
+        "order": 0,
```
flash1293, line 13: If the feature flag is off, starting the order from 0 for each group will shuffle the fields.

It seems like we can keep the current orders except for the JDBC-params / Replication method case in the "Advanced" section, maybe it's fine to keep them in the current order (just making sure that we can roll back with the feature flag)

flash1293 (commented):
Left two comments

airbyte-integrations/connectors/source-postgres/src/main/resources/spec.json
lmossman: Good catch - I've updated this to undo the `order` changes I made to this spec. I also moved the `schemas` field into the `db` group since I think it belongs better there.

I think the advanced section is fine with the optional fields coming first, so we can leave that as is for now.

airbyte-integrations/bases/base-java/src/main/resources/ssh-tunnel-spec.json
lmossman: @flash1293 I went mostly with your suggestion, but I wanted to avoid duplicating the `ssh-tunnel-spec.json` just to add a single `group` field, so I instead updated these classes to parametrize those methods to take in an optional `sshGroup` string which is set on the tunnel_method spec if present before it is set on the connector spec. I used overloaded methods here so other connectors should not need any changes.

See the commit of these changes here: https://github.com/airbytehq/airbyte/pull/24127/commits/b0404a1c5ca5693e9b1943ed0aacc04a05cd3ab1

flash1293 (approved):
Built postgres and github locally and checked with and without the feature flag on the https://github.com/airbytehq/airbyte-platform-internal/pull/4833 branch - LGTM

One nit (but I think that's OK) - as the optional fields have orders we don't introduce a footer-section but an inline one:

Screenshot 2023-03-17 at 13 12 52

Design:
Screenshot 2023-03-17 at 13 14 08

lmossman:
The above comment has been addressed by making the last collapsed-inline section automatically convert to footer in https://github.com/airbytehq/airbyte-platform-internal/pull/4833/commits/8ecb437896134b35d4430104f875ad564cc1a276

lmossman:
@rodireich this build failed with the following error:
```
ERROR:root:The following GA connectors must enable high test strictness level: [source-postgres]. Please check this documentation for details: https://docs.airbyte.com/connector-development/testing-connectors/connector-acceptance-tests-reference/#strictness-level
```
I started looking into enabling the high test strictness level on this connector, but I saw that you commented out the `incremental` tests which seem to be required for high test strictness level.

Would you be able to look into enabling the high test strictness level on this connector so that this build can pass and we can release this new version with the updated connector setup form UX?

rodireich:
This is expected as we commented out the `strictness: high` .
Currently blocked by #23098

rodireich:
Note that the connector running in Cloud is /connectors/source-postgres-strict-encrypt.
It has code that adds another logical layer and [modifies spec](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-postgres-strict-encrypt/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceStrictEncrypt.java#L41)
Please ensure that none of the changes there conflict with the new spec.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/ssh/SshHelpers.java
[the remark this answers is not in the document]
lmossman: @rodireich This should not affect any other connectors, because these changes just add an overloaded method with a new `group` parameter. All other connectors that are calling this code will call the old method that does not take in the `group` parameter, which in turn calls the new method with `Optional.empty()` as the second parameter, in which case the logic is the same it was before this change

lmossman:
> Note that the connector running in Cloud is /connectors/source-postgres-strict-encrypt. It has code that adds another logical layer and [modifies spec](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-postgres-strict-encrypt/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceStrictEncrypt.java#L41) Please ensure that none of the changes there conflict with the new spec.

@rodireich I just tested this by building the `source-postgres-strict-encrypt` and it looks like the spec is still generated properly and working as expected.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/ssh/SshHelpers.java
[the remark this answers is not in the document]
lmossman: To be safe, I will run a test command on the mysql connector, as that also uses this code, to verify that the tests still run successfully for that connector

rodireich:
Should be good if tests passing. I approved

rodireich (approved):

lmossman:
@rodireich this wasn't quite ready to merge, as I hadn't published these connectors yet which includes committing the changes to the `source_specs.yaml` and `source_definitions.yaml` files. I've opened another PR here where I will be re-testing those connectors and running the publish command if successful: https://github.com/airbytehq/airbyte/pull/24212

rodireich:
Oh my. Did I merge this?! 😱

PR #24136, opened 2023-03-16
connectors-ci: fix secrets upload to GSM

alafanechere:
## What
We previously updated the `upload` signature to take a `context` object but we passed it a `client`...


octavia-approvington:

octavia-approvington (approved):

PR #24147, opened 2023-03-16
MySQL Source : support different table structure during "DESCRIBE" query

VitaliiMaltsev:
## What
**DESCRIBE <table_name>**  query in MySQL can return a different table structure depending on the version of the database

In one case, this structure has the form
_{"Field":"<column_name>","Type":"<datatype_in_db>","Null":"<nullability(YES, NO)>","Key":"<key>","Extra":"<extra_params>"}_
in other cases
_{"COLUMN_NAME":"<column_name>","COLUMN_TYPE":"<datatype_in_db>","IS_NULLABLE":"<nullability(YES, NO)>","COLUMN_KEY":"<key>","EXTRA":"<extra_params>"}_

## How
Support both of these query results

Monitor these logs for whether the sync was successful or not : https://github.com/airbytehq/oncall/issues/1654


sashaNeshcheret (approved):

akashkulk:
Looks good in general. Can you quickly mention which versions of MySQL correspond to which versions of the `DESCRIBE` query?

PR #19223, opened 2022-11-09
🎉 New Source: Demo - Youtube Analytics Business

bazarnov:
## What
Resolving: https://github.com/airbytehq/airbyte/issues/19017

PR #24148, opened 2023-03-16
🎉 Source YoutubeAnalyticsBusiness Demo: remove the `source-youtube-analytics-business` from `OSS` part 2/3

bazarnov:
## What
PART 2/2: Remove the connector itself and it's leftovers from OSS repo.

As far as Google has approved the `youtube-business` specific scopes, we don't want to store 2 equal connectors doing the same thing but using 2 different scopes. Thus, we need to revert adding the copy-connector from OSS repo.

Reverting: https://github.com/airbytehq/airbyte/pull/19223
Specifically reverting:
- adding copy-connector `source-youtube-analytics-business`  and it's leftovers from repo.

OOO: (order of operations)
1. Merge the Cloud seed changes https://github.com/airbytehq/airbyte-platform-internal/pull/5285 and remove the DB OAuth creds row from prod (part 1/3)
2. Merge the changes from this PR to OSS repo (part 2/3)
3. Merge [the changes for `platform-internal`](https://github.com/airbytehq/airbyte-platform-internal/pull/5222) (part 3/3)

## How
- removed the public docs
- removed `source-youtube-analytics-business` connector from the OSS repo
- removed seed for it.
 
## Recommended reading order
3. `y.python`


erohmensing:
@bazarnov Do we know how many people are using this? Do we have instructions on how to migrate to the regular youtube analytics connector? Same question for open source and cloud actually

bazarnov:
> @bazarnov Do we know how many people are using this? Do we have instructions on how to migrate to the regular youtube analytics connector? Same question for open source and cloud actually

Yes, we know. No one uses it, because it was not exposed to the audience, because the target was to satisfy the demo requirements before the scope is approved, and remove everything afterwards.

erohmensing:
@YowanR I think you might have metabase dashboards for OSS usage?

pedroslopez:
@erohmensing This demo connector was only showing up to be selectable for a specific workspaceId, both in cloud and OSS. I think this is what @bazarnov is referring to when saying "this was not exposed to the audience".

pedroslopez (approved):
This side of things LGTM, as long as we do things in the right order 😄 

To recap, this should only be merged after we've commented out or removed this source from the source_definition_mask in cloud.

erohmensing:
@pedroslopez 👀 where does one configure that in OSS?

pedroslopez:
> @pedroslopez 👀 where does one configure that in OSS?

Looking more closely at it, I lied 😛 - the exclusions don't apply to OSS. (missed the `isCloudApp()` check)

I was referring to this in the webapp:

https://github.com/airbytehq/airbyte-platform-internal/blob/e65dc619bce4166ec333652dd5e43650d1a49e37/oss/airbyte-webapp/src/core/domain/connector/constants.ts#L19-L37

Hopefully folks could look at the docs for this connector and realize that it was a demo thing, but it could have been more clear. I think we could look at metabase to confirm. In any case, removing the connector from the OSS catalog won't _delete_ the connector for people that already have it loaded in.

bazarnov:
> @bazarnov What do you mean by "it wasn't exposed to the audience" - wouldn't that only indicate cloud usage? If we published a public docker image for it and have public docs for it, any number of open source users could be using it. I'm not sure how we check that (I think we have segment data maybe?) but we should look into whether anyone is using it if we have the data for that

The `source-youtube-analytics-business` was available only for Cloud and only for 1 workspace, which is not accessed by anyone except the one who has access and knows the workspace exists (wks:  oauth-testing)
Here is the screenshot of the current Cloud situation:
Screenshot 2023-03-17 at 16 43 02

@YowanR 
For OSS users it makes sense to check whether or not anyone uses it, but I'm very skeptical at this point that someone configured a not-working source connector listed as `alpha`, but worth checking. 
The name of our target is `source-youtube-analytics-business` or just `Youtube Analytics Business` for your easy checking in Metabase.

bazarnov:
> > @pedroslopez 👀 where does one configure that in OSS?
> 
> @erohmensing Looking more closely at it, I lied 😛 - the exclusions don't apply to OSS. (missed the `isCloudApp()` check)
> 
> I was referring to this in the webapp:
> 
> https://github.com/airbytehq/airbyte-platform-internal/blob/e65dc619bce4166ec333652dd5e43650d1a49e37/oss/airbyte-webapp/src/core/domain/connector/constants.ts#L19-L37
> 
> Hopefully folks could look at the docs for this connector and realize that it was a demo thing, but it could have been more clear. I think we could look at metabase to confirm. In any case, removing the connector from the OSS catalog won't _delete_ the connector for people that already have it loaded in.

Thank you) To be clear - this demo source connector will not do the syncs) It was adjusted that way to pass the `scope` validation (on the consent screen for OAuth) and pass the `check connection` step. No more.

erohmensing:
>  In any case, removing the connector from the OSS catalog won't delete the connector for people that already have it loaded in.

That's true and I totally forgot about that - if they were using it previously they could load it in as a custom connector since we're not removing the docker image from dockerhub. That addresses my concerns :shipit:

pedroslopez:
> edit: or I guess you said it won't be deleted for them - is that that "we have to remove it from the db in order for it to be truly gone" thing?

@erohmensing Yeah, we never actually delete things from the db based on a catalog change. The good thing about it is we don't break things for people that may be using the connector, but the downside is that existing users are still going to have this demo connector around forever.

evantahler:
> @YowanR I think you might have metabase dashboards for OSS usage?

Here is the metabase [data](https://airbyte.metabaseapp.com/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7InR5cGUiOiJuYXRpdmUiLCJuYXRpdmUiOnsicXVlcnkiOiItLSBTRUxFQ1QgKlxuLS0gRlJPTSBgYWlyYnl0ZV93YXJlaG91c2UuY29ubmVjdGlvbmAgXG4tLSBXSEVSRSBjb25uZWN0b3Jfc291cmNlX2RlZmluaXRpb25faWQgPSBcIjc4NzUyMDczLTZkOTYtNDQ3ZC04YTkzLTJiNjk1M2YzYzc4N1wiXG4tLSBMSU1JVCAxMDAwXG5cblNFTEVDVCBcbiAgICBjb3VudCgqKSBhcyB0b3RhbFxuICAgICwgd29ya3NwYWNlX2lkIFxuICAgICwgaXNfY2xvdWRcbkZST00gYGFpcmJ5dGVfd2FyZWhvdXNlLmNvbm5lY3Rpb25gIFxuV0hFUkUgY29ubmVjdG9yX3NvdXJjZV9kZWZpbml0aW9uX2lkID0gXCI3ODc1MjA3My02ZDk2LTQ0N2QtOGE5My0yYjY5NTNmM2M3ODdcIlxuZ3JvdXAgYnkgd29ya3NwYWNlX2lkLCBpc19jbG91ZCIsInRlbXBsYXRlLXRhZ3MiOnt9fSwiZGF0YWJhc2UiOjJ9LCJkaXNwbGF5IjoidGFibGUiLCJwYXJhbWV0ZXJzIjpbXSwidmlzdWFsaXphdGlvbl9zZXR0aW5ncyI6e319)

In addition to the 1 cloud workspace using the connector (which ran 66 syncs), there are 7 OSS workspaces who ran the connector as well during the demo period, some more than once.

I think the lesson here is that in the future we need to hide the connector from both OSS and Cloud users in this scenario.  Better yet, this connector should not have been in the registry, and loaded in as a custom connector.  

@YowanR as there are OSS users who are using this connector, you might want to connect with @jerri-airbyte on a messaging plan

pedroslopez:
>  Better yet, this connector should not have been in the registry, and loaded in as a custom connector.

Strong +1 to this.

YowanR:
Thanks @evantahler and @pedroslopez ! I'm following up with @jerri-airbyte and will also make sure we don't repeat this mistake in the future 🙏

PR #24158, opened 2023-03-16
Connector Builder: set `exclude_unset=True` when returning `AirbyteMessage`s

clnoll:
## What
Fixes an issue where we were sending `AirbyteMessage` json from the CDK without setting `exclude_unset=True` in the json call. This was causing these messages to fail [validation](https://github.com/airbytehq/airbyte-platform-internal/blob/2e2c9466de22821076a853c31de0ecaa04a14e20/oss/airbyte-commons-worker/src/main/java/io/airbyte/workers/internal/DefaultAirbyteStreamFactory.java#L114) performed by the connector builder server.


girarda (approved):

clnoll:
This was already fixed on master

PR #24166, opened 2023-03-16
cdc: fix debezium shutdown interruption bug

subodh1810:
For issue https://github.com/airbytehq/oncall/issues/1596
See this [comment](https://github.com/airbytehq/oncall/issues/1596#issuecomment-1473485800) for detailed explanation


sergio-ropero:
Why don't you just trigger the shutdown in a different thread and let Airbyte main thread to finish the consumption of the queue as usual?
I just did a quick read on the changes and it's super confusing without any details on the PR or comments in the code itself. Do we really need this type of classes: `DebeziumShutdownProcedure<ChangeEvent<String, String>>`?

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIterator.java
```diff
@@ -134,15 +134,33 @@ protected ChangeEvent<String, String> computeNext() {
...
       this.tsLastHeartbeat = null;
       this.lastHeartbeatPosition = null;
       this.receivedFirstRecord = true;
       this.maxInstanceOfNoRecordsFound = 0;
       return next;
     }
+
+    if (!signalledDebeziumEngineShutdown) {
+      LOGGER.warn("Debezium engine has not been signalled to shutdown, this is unexpected");
+    }
```
sergio-ropero, line 148: Why is this unexpected?

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIterator.java
```diff
@@ -134,15 +134,33 @@ protected ChangeEvent<String, String> computeNext() {
...
+      } catch (InterruptedException e) {
+        throw new RuntimeException(e);
+      }
+      if (event == null || isHeartbeatEvent(event)) {
+        continue;
+      }
+      hasSnapshotFinished = hasSnapshotFinished(Jsons.deserialize(event.value()));
+      return event;
+    }
+    throwExceptionIfSnapshotNotFinished();
```
sergio-ropero, line 163: Why the process is not getting the events directly from `queue`?

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIterator.java
```diff
@@ -189,13 +206,13 @@ private boolean hasSnapshotFinished(final JsonNode eventAsJson) {
...
-      throw new RuntimeException(e);
+  private void requestClose(final String closeLogMessage) {
+    if (signalledDebeziumEngineShutdown) {
+      return;
     }
-    throwExceptionIfSnapshotNotFinished();
+    LOGGER.info(closeLogMessage);
+    debeziumShutdownProcedure.initiateShutdownProcedure();
+    signalledDebeziumEngineShutdown = true;
   }
```
sergio-ropero, line 216: From my point of view we should be aware that `requestClose` is called only once in the code and not just figure out if it has been called before using an if statement and the flag `signalledDebeziumEngineShutdown`.

airbyte-integrations/bases/debezium/src/test/java/io/airbyte/integrations/debezium/internals/DebeziumShutdownProcedureTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+    Thread.sleep(1000);
+    debeziumShutdownProcedure.initiateShutdownProcedure();
+
+    Assertions.assertEquals(99, recordsInserted.get());
+    Assertions.assertEquals(0, sourceQueue.size());
+    Assertions.assertEquals(100, debeziumShutdownProcedure.getRecordsRemainingAfterShutdown().size());
+
+    for (int i = 0; i < 99; i++) {
+      Assertions.assertEquals(i, debeziumShutdownProcedure.getRecordsRemainingAfterShutdown().poll());
```
sergio-ropero, line 42: How is this happening? Is `getRecordsRemainingAfterShutdown` not being actually remaining but the total?
If we have already inserted 99 records, the source queue is empty, and the total expected by the for loop is 100... Remaining should be 1?

Also what is happening with last record, are we losing it?

sergio-ropero (changes requested):

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIterator.java
subodh1810: Cause we should have shutdown the engine by this point of execution. This will never happen today but it's more of a safety log that I placed to catch bugs in case of future changes to the code.

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIterator.java
subodh1810: We need to make sure that the `queue` always has enough space but since we use a capacity queue, we need to read data from the queue in order to make space. The problem (as explained in the loom video attached in the description of the PR) is that when we trigger the shutdown of the Debezium engine, the main thread pauses until the engine shuts down. And since the reading from the `queue` is also happening in the main thread, the reading stops and if queue is already at capacity, it will not accept new records from Debezium. And Debezium wont shutdown gracefully cause it failed to hand over all the records that it fetched to the `queue`. 

As a solution now when we trigger shutdown, in order make sure the reading from the `queue` continues to happen, we start moving data from the `queue` in a separate new queue in a new thread. This allows debezium to shutdown gracefully, while the main thread continues to wait. Once the engine is shutdown, all we have to do is to read the remaining records that we moved to the new queue and we are good. 

**Please note that the main thread needs to be in pause state while debezium shuts down in order for the system to work without any edge cases.** And since the reading of the data happens inside the main thread, this the solution I chose. 

I did try another solution where we trigger the shutdown of the debezium engine in a separate thread and the main thread continues to run and read data but the implementation was even more complicated and the logic to track the state of the new thread was leaking all over the place (inside DebeziumRecordIterator and DebeziumRecordPublisher). Also it would have not worked cause of another new edge case that it would introduce. 

Right now the entire logic is contained inside the class `DebeziumShutdownProcedure` at one place which makes any changes to the other two classes in future easy to be done. 

I agree that this is not straight forward to understand but neither would any other solution. CDC itself is a complex piece of code. And anytime you are going to work with multi threading designs, things are complex.

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIterator.java
subodh1810: We call `requestClose` when certain conditions are met. For instance if the LSN of a record is higher than the target LSN that we  aimed for. Please remember that Debezium doesn't understand the concept of target LSN and it will continue to fetch data until it's shutdown. Also Debezium is running in a separate thread. So often Debezium pulls in multiple records which are after the target LSN and only when we see the first of the record which is after target LSN, we ask Debezium to shutdown but by that time Debezium has already fetched even more records. And we need to process these records. But each of the record that we process from that point is after the target LSN which makes the condition to call `requestClose` true. Here is sample screenshot of that happening. 
Screenshot 2023-03-23 at 12 49 39 PM

As you can see in the above screenshot we processed three such records which had been pulled which were after the target LSN and `requestClose` was called thrice. To tackle this, we need the boolean flag.

airbyte-integrations/bases/debezium/src/test/java/io/airbyte/integrations/debezium/internals/DebeziumShutdownProcedureTest.java
subodh1810: Ah its just a mistake on my end. The loop should be <100. My bad.

sergio-ropero (approved):
Nice work!! hard to tackle with the spaghetti code we have... 🤦

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumShutdownProcedure.java
```diff
@@ -0,0 +1,126 @@
...
+    });
+  }
+
+  private Runnable transfer() {
+    return () -> {
+      while (!sourceQueue.isEmpty() || !MoreBooleans.isTruthy(publisherStatusSupplier.get())) {
+        try {
+          T event = sourceQueue.poll(10, TimeUnit.SECONDS);
+          if (event != null) {
+            targetQueue.put(event);
```
rodireich, line 55: Suggestion: use offer to avoid blocking / error in case target queue is at capacity

subodh1810: The targetQueue is not bound by size so it should not be at capacity.

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIterator.java
```diff
@@ -134,15 +134,34 @@ protected ChangeEvent<String, String> computeNext() {
...
+      try {
+        event = debeziumShutdownProcedure.getRecordsRemainingAfterShutdown().poll(10, TimeUnit.SECONDS);
+      } catch (InterruptedException e) {
+        throw new RuntimeException(e);
+      }
+      if (event == null || isHeartbeatEvent(event)) {
+        continue;
+      }
+      hasSnapshotFinished = hasSnapshotFinished(Jsons.deserialize(event.value()));
+      return event;
```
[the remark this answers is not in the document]
subodh1810: We can not/should not discard it cause it's still relevant data. Debezium updates its offset uptill the point it has fetched records so we need to process these records. See [this](https://github.com/airbytehq/airbyte/pull/24166#discussion_r1145793818)

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumRecordIterator.java
```diff
@@ -134,15 +134,34 @@ protected ChangeEvent<String, String> computeNext() {
...
+
+    if (!signalledDebeziumEngineShutdown) {
+      LOGGER.warn("Debezium engine has not been signalled to shutdown, this is unexpected");
+    }
+
+    // Read the records that Debezium might have fetched right at the time we called shutdown
+    while (!debeziumShutdownProcedure.getRecordsRemainingAfterShutdown().isEmpty()) {
+      final ChangeEvent<String, String> event;
+      try {
+        event = debeziumShutdownProcedure.getRecordsRemainingAfterShutdown().poll(10, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
subodh1810: WE cant read form the regular queue cause the reading happens inside the main thread and the main thread pauses once we request shutdown and thus the reading stops. The records can not be discarded.
https://github.com/airbytehq/airbyte/pull/24166#discussion_r1145793818

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumPropertiesManager.java
```diff
@@ -55,7 +55,8 @@ protected Properties getDebeziumProperties() {
     props.setProperty("max.queue.size", "8192");
 
     props.setProperty("errors.max.retries", "10");
-    props.setProperty("errors.retry.delay.initial.ms", "1000");
+    props.setProperty("errors.retry.delay.initial.ms", "300");
```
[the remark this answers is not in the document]
subodh1810: Cause in our Airbyte code we only wait for 60 seconds for Debezium to return a record. So Debezium needs to hurry and restart in case of an error. If it waits for 10 seconds, there is a possibility by the time it restarts, we would ask it to shutdown anyway

subodh1810:
/publish connector=connectors/source-mssql-strict-encrypt run-tests=false

ifun248:
@subodh1810 Hi Subodh1810, I am testing source MySQL 2.0.17, and find an interesting problem maybe is related to this bug. refs https://github.com/airbytehq/airbyte/issues/25988.

PR #24184, opened 2023-03-17
Connector builder: handle empty catalog

clnoll:
## What
Fixes an issue where the CDK attempts to validate an empty catalog on requests where the catalog is not needed.

## Why
The Connector Builder server has two request types that don't require a catalog: `resolve_manifest` and `list_streams`. For these, the filepath provided in the `--catalog` argument points to an empty catalog. We don't actually want or need to try to validate the catalog, so let's skip it, to avoid unnecessary file reads if nothing else.

### DIFF
--- a/airbyte-cdk/python/connector_builder/main.py
+++ b/airbyte-cdk/python/connector_builder/main.py
@@ -4,7 +4,7 @@
 
 
 import sys
-from typing import Any, List, Mapping, Tuple
+from typing import Any, List, Mapping, Optional, Tuple
 
 from airbyte_cdk.connector import BaseConnector
 from airbyte_cdk.entrypoint import AirbyteEntrypoint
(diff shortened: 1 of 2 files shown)


airbyte-cdk/python/connector_builder/main.py
```diff
@@ -19,25 +19,36 @@ def create_source(config: Mapping[str, Any]) -> ManifestDeclarativeSource:
...
         )
 
-    return config, catalog
+    return command, config, catalog
 
 
-def handle_connector_builder_request(source: ManifestDeclarativeSource, config: Mapping[str, Any], catalog: ConfiguredAirbyteCatalog):
-    command = config.get("__command")
+def handle_connector_builder_request(
+    source: ManifestDeclarativeSource, command: str, config: Mapping[str, Any], catalog: Optional[ConfiguredAirbyteCatalog]
```
girarda, line 50: nit: `read_stream` expects `catalog` to be non-None, so we should validate its presence before passing it

girarda (approved):

airbyte-cdk/python/connector_builder/main.py
clnoll: Ah good call. I didn't see this before I hit merge so will open a new one for that.

PR #24199, opened 2023-03-17
Connector Builder: assert catalog exists on `test_read` requests

clnoll:
Adds a check that the catalog exists for `test_read` requests.


airbyte-cdk/python/connector_builder/main.py
```diff
@@ -52,6 +52,7 @@ def handle_connector_builder_request(
     if command == "resolve_manifest":
         return resolve_manifest(source)
     elif command == "test_read":
+        assert catalog is not None, "`test_read` requires a valid `ConfiguredAirbyteCatalog`, got None."
```
maxi297, line 55: This would be part of the API between the Connector Builder and the CDK. Is this what we expect?

clnoll: Yes, it ultimately gets wrapped in an AirbyteTraceMessage, which is handled by the connector builder server.

maxi297 (approved):
LGTM!

PR #24218, opened 2023-03-18
CDK: build_refresh_request_headers added

grubberr:
## What
I would like to have the possibility to customize headers `_get refresh access_token response` in inherited classes.


sherifnada (changes requested):
Looks good! could you update the tests to verify that headers are included on outgoing refresh requests

PR #24240, opened 2023-03-20
CDK: remove unexpected error swallowing on abstract source's `check` method

### LINKED ISSUE #23281: CDK: remove unexpected error swallowing on abstract source's `check` method
As part of the improve synchronous job error reporting tech spec, the following improvement to the CDK was identified: 

Connectors output trace messages instead of failed check connection statuses for unexpected errors
The current implementation of many connectors, including CDK sources and destinations, involves excepting any unhandled exception during the check method, and returning a connection_status message with status=failure. This effectively turns a failed check job into a successful check job with a failed connection test. It does so without sending a trace message. 

In order to be able to provide more information to the user than can be contained only in the ‘message’ portion of the connection status message, connectors should output trace messages when they run into unhandled errors, instead of output connection status messages.  In this project, we will add this as the default behavior for abstract sources in the CDK.  It is not in scope for this project to update existing connectors that override this default behavior. 

The CDK already has code to be able to handle this if general catch-all error handling is not in place. [This error handling](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/exception_handler.py#L11) which outputs trace messages for uncaught exceptions, is currently not being used in the check method because uncaught exceptions are being caught [here](https://github.com/airbytehq/airbyte/blob/2e099acc52aeddc172c5ee66ed98426c69449a4a/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py#L86-L87). This not only makes us lose information when going back to the API, but also stops us from doing our normal error handling for uncaught exceptions, which reports errors to sentry. For example, there was no sentry error for [this P0](https://docs.google.com/document/d/1c6H50vpIOT8eQEj5i8fYsndgMB0kM19s21n_BGr2gVc/edit) due to the fact that the error was caught and output as a connection status failure.

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/23281
We don't want to make unexpected error in `check`  swallowed into a failed Connection Status. We want these error to be handled at a higher level with the [`uncaught_exception_handler`](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/exception_handler.py#L11), **to emit trace messages**.

## How
* Remove the handling of broad exception in `check` method of the AbstractSource
* Update CAT to reflect this change
* Update Airtable's `acceptance-test-config.yml` to assess the correct behavior of the implementation.

## Recommended reading order
1. https://github.com/airbytehq/airbyte/blob/578561a42a1aba2f4986df51763a73d3f227d783/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py#L83
2. https://github.com/airbytehq/airbyte/blob/578561a42a1aba2f4986df51763a73d3f227d783/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py#L490

## 🚨 User Impact 🚨
* All the connectors not handling all Exceptions in an overridden `check` will emit trace message in case of failure. (This is what we want)


alafanechere:
@clnoll I think I'm hitting a edge case while trying to use local CDK and local CAT. When I pass the local_cdk flag the `buildDockerfileWithLocalCdk` gradle function in `airbyte-docker` gets called to build CAT. The built image has wrong entrypoint leading to the error you can check in the logs of the /test above: 
`main.py: error: argument command: invalid choice: 'integration_tests.acceptance' (choose from 'spec', 'check', 'discover', 'read')`

clnoll:
Thanks for the heads up @alafanechere, fix is incoming in a few.

clnoll:
The fix is merged @alafanechere.

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
```diff
@@ -488,11 +488,14 @@ def test_check(self, connector_config, inputs: ConnectionTestConfig, docker_runn
...
-            assert err.value.exit_status != 0, "Connector should exit with error code"
-            assert "Traceback" in err.value.stderr, "Connector should print exception"
+            output = docker_runner.call_check(config=connector_config, raise_container_error=False)
+            trace_messages = filter_output(output, Type.TRACE)
+            assert len(trace_messages) == 1, "A trace message should be emitted in case of unexpected errors"
+            trace = trace_messages[0].trace
+            assert isinstance(trace, AirbyteTraceMessage)
+            assert trace.error is not None
+            assert trace.error.message is not None
+            assert trace.error.internal_message is not None
```
erohmensing, line 498: 👍🏻 I like this test because it doesn't make any opinions about whether a `connection_status` message is output, only confirms that a trace message is output. This follows the behavior that I now use in the platform, where if there was a trace message, we declare that there was an error and don't take a connection_status message (if existing) into account. So connectors that currently output both won't fail the test. 

Are we concerned about connectors that aren't CDK ones that have the same behavior (e.g. most destinations - not sure if we regularly run CAT on destinations but we could) will suddenly start failing this?

airbyte-integrations/bases/connector-acceptance-test/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.7.2
+TestConnection: assert that a check with `exception` status emits a trace message.
+
```
erohmensing, line 5: Wondering if this should be bumped differently, as it's a new test we expect some to fail 🤔

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
erohmensing, line 498: Do we want to enforce `internal_message` even for connectors not using CDK tooling to create their traces? At the moment, the airbyte protocol only enforces `message` as a required property.

erohmensing (approved):
Thank you for taking this on! Code LGTM, I only wonder about other connectors that might start failing the test and how we want to handle them. 

Would also probably be good to get a list of CDK connectors that override `check` instead of overriding `check_connection`, since they won't inherit the change - not that we have to fix them here, but would be good to have a list for e.g. GL to look into

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
erohmensing: I guess this is only tested if someone provides a `status: "exception"` config to test though

alafanechere: > I guess this is only tested if someone provides a status: "exception" config to test though

Yes indeed.
I'd love to find a way to create a global not configurable test for this. Because it's unlikely developers will fill in a `status: "exception"` as they'll rather use `status: "fail"` to make sure the expected errors are correctly handled.
Do you have a solution in mind to always make check fail in this test context? (Maybe preventing internet access to the container under test?)

airbyte-integrations/bases/connector-acceptance-test/CHANGELOG.md
alafanechere: Only two connectors are using `status: "exception"` in their acceptance test config (`dixa` and `file`). They both are not passing CAT at the moment.
I don't expect additional connectors to fail as only these connectors + airtable will go into the `inputs.status == ConnectionTestConfig.Status.Exception:` statement.

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
alafanechere: Good call! I removed the assertion on `internal_message`

erohmensing: I believe check currently fails if we pass in a path to a config that doesn't exist. We could try to leverage that?

alafanechere: Yep this could work. The problem is that `test_check` relies on the `inputs` fixture which is reading the config. Passing an unexisting config leads to a FileNotFoundError on CAT.
I'll try to create a separate test not using this fixture.

erohmensing: Haha, our fixtures are too good to make fake errors 😄

alafanechere:
@erohmensing I can't find an obvious way to make all the connectors specifically fail at `check`. I tried to create a test context in which connector container do not have internet access, but this does not work for all connectors as they are not all making HTTP requests in `check`. 
I'll keep the current CAT change but I doubt about its usefulness. As @pedroslopez mentioned in our yesterday meeting: If we can fake an unexpected error it means it's not an unexpected error and we should improve the connector to properly manage these error. I'm pretty confident the CDK is properly handling unexpected error by logging trace messages with the [`uncaught_exception_handler`](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/exception_handler.py#L11) . This function is already unit tested and manual test validated this behavior.

airbyte-cdk/python/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.31.2
```
maxi297, line 3: This can be done as part of [the CDK release process] to avoid clash with other branches. Same with the changes in airbyte-cdk/python/setup.py

maxi297 (commented):
I think the concern of failing tests has been raised by Ella. I'm wondering if there is more problem this compatibility change might incur. In theory, this is a breaking change (a user expecting an `AirbyteConnectionStatus` will now receive an exception) even though I understand it's probably "soft"

erohmensing:
@maxi297 FWIW, the breaking part of this is handled so long as you are using airbyte's platform, as I changed the platform to be able to raise trace messages as errors to the user during the check step. When using it via docker, though, I do understand that the behavior changes - however, in docker you don't need the output of check in order to proceed to to the next step (so if you were ignoring that you got a failed check connection and moved on, you still can). The only way I could see this potentially changing someone's workflow significantly is if they had their own UI 🤔

I don't have very strong opinions on the versioning here, I'd be happy with bumping it more too, either one works for me

alafanechere:
> I think the concern of failing tests has been raised by Ella.
What can indeed happen is that some connectors expecting a "failed" `check` in CAT do not have a failed connection status anymore. I'm going to test a couple of GA connector.

erohmensing:
Ah good point, the test itself that expects a failed check might make an exception instead - makes sense (and that's indeed something we should add handling for)

alafanechere:
The following GA connectors are failing CAT "connection" checks now:
```
Source-paypal
source-google-analytics-v4
Source-zendesk-talk
Source-amazon-ads
Source-google-ads 
Source-pinterest
```

alafanechere:
For google-analytics-v4 the same tests are failing on [master](https://github.com/airbytehq/airbyte/actions/runs/4496658337/jobs/7911552150) so I can say that the failures are not related to my current changes.

alafanechere:
@erohmensing @maxi297 I'll merge it now that the impacted GA connectors are not likely to fail at CAT after their acceptance test config updates.
The following connectors required an `acceptance-test-config.yml` update:
* source-paypal 
* source-google-analytics-v4
* source-zendesk-talk 
* source-amazon-ads 
* source-google-ads 
* source-marketo
* source-pinterest

erohmensing:
Thank you @alafanechere !

PR #24245, opened 2023-03-20
Update default threshold value for Jira

benmoriceau:
## What
Move the default timeout threshold for jira. Move it to 6 hours


jdpgrailsdev (approved):
:shipit:

PR #24252, opened 2023-03-20
Source zendesk support: validate datetime on check

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1683

## How
Add validation of a `start_date` param when checking connectivity


airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/source.py
```diff
@@ -82,8 +82,9 @@ def check_connection(self, logger, config) -> Tuple[bool, any]:
         auth = self.get_authenticator(config)
         settings = None
         try:
+            datetime.strptime(config["start_date"], DATETIME_FORMAT)
             settings = UserSettingsStream(config["subdomain"], authenticator=auth, start_date=None).get_settings()
-        except requests.exceptions.RequestException as e:
+        except Exception as e:
```
bnchrch, line 87: Could we be slightly more restrictive here and catch a datetime format issue in addition to the request exception?

To catch all exceptions could lead to some silent issues for customers.

bnchrch (approved):
Nice work. 👍 

One comment I do think we should address.

I will approve so as not to be a blocker.

airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/source.py
davydov-d: @bnchrch this is done on purpose. We want every exception to be caught, I would say this is a pattern for the `check` command implementation

PR #24255, opened 2023-03-20
Simplify Github and Postgres forms #2

### LINKED ISSUE #23696: Migrate 1-2 connectors to new simplify format

lmossman:
## What
Resolves https://github.com/airbytehq/airbyte/issues/23696

As part of the simplify project definition of done, we want to migration a couple of connectors to use the new fields.

There was a [previous PR](https://github.com/airbytehq/airbyte/pull/24127) up to address this, but it was accidentally merged before the tests were properly updated, so this PR is the replacement for that one.

## How
Migrates the postgres and github source connectors to use the new simplify fields (`group` and `groups`), so that they match the designs linked in the above issue.

The Postgres connector was a bit trickier to migrate, because it dynamically creates its spec by merging in the `ssh-tunnel-spec.json` through the `SshHelpers` class. 
So, in order to set a `group` on the tunnel_method field of the postgres spec (so that the field is placed in the right place on the UI), I had to update this class to also take in an optional `group` parameter and set it on the tunnel_method if present, which is set to `"security"` in the Postgres source connector.

As a result I also had to update the acceptance tests accordingly. To make this change slightly more maintainable, I created an `AbstractPostgresSourceAcceptanceTest` class which overrides the `getSpec` method with the right group setting, and made the other acceptance tests extend that one (except the strict encrypt one since it does things slightly differently and it already extends another class).

## Recommended reading order
1. */spec.json - Update the github and postgres connectors to add `group` and `groups` to the fields to make them look like the designs in the linked tie
2. SshHelpers.java - add the overloaded methods that take in an optional group argument, and set that on the tunnel method if present.
3. SshWrappedSource.java - add an overloaded constructor that takes in a group and passes it to SshHelpers if provided.
4. PostgresSource.java - use the new constructor to pass the ssh group in
5. */Dockerfile - bump the versions in the dockerfiles
6. AbstractPostgresSourceAcceptanceTest.java - define a new abstract class that sets the image name and spec as it is set in all of the other acceptance tests, and update them to extend this
7. expected_spec.json - update the expected spec accordingly

## 🚨 User Impact 🚨
The postgres and github source connectors will have a simplified look for users, only if they have the `connector.form.simplifyConfiguration` feature flag enabled.


airbyte-integrations/connectors/source-postgres/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/AbstractPostgresSourceAcceptanceTest.java
```diff
@@ -0,0 +1,18 @@
+package io.airbyte.integrations.io.airbyte.integration_tests.sources;
+
+import io.airbyte.integrations.base.ssh.SshHelpers;
+import io.airbyte.integrations.standardtest.source.SourceAcceptanceTest;
+import io.airbyte.protocol.models.v0.ConnectorSpecification;
+import java.util.Optional;
+
+public abstract class AbstractPostgresSourceAcceptanceTest extends SourceAcceptanceTest {
```
lmossman, line 8: Created this abstract class so that I didn't need to repeat the `"security"` group string in every test

airbyte-integrations/connectors/source-postgres/src/test/resources/expected_spec.json
```diff
@@ -1,383 +0,0 @@
-{
```
lmossman, line 1: Deleted this file because it was unused, and I wasted some time figuring out which `expected_spec.json` file I was supposed to update

lmossman:
@flash1293 @rodireich I have put up this new PR to replace the old one that we reverted, and all of the acceptance tests are now passing, so this is ready for a review.

Once this gets approval, I will run the `/publish` commands for github, postgres, and postgres-strict-encrypt, then I'll merge this in

flash1293 (approved):
Build github and postgres locally and tested - LGTM. The only thing we might need to address is to allow collapsing of oneOfs:
Screenshot 2023-03-21 at 11 55 46

But that's a separate discussion

PR #24273, opened 2023-03-21
connectors-ci: new workflow for dagger nightly builds

### LINKED ISSUE #23970: connectors-ci: use the new connectors-ci command to run connector nightly builds
When the `connectors-ci` pipeline will be able to test and build all our connectors we shall use it to power our nightly builds.

We should be able to have the tests/builds for GA, Beta, and alpha connectors run in the same GHA as we'll benefit from the beefy runner pool the infra team will provide. This runner pool will have docker and pip caches that should make the nightly build faster than today.

The two following workflows should be replaced by a single one calling `connectors-ci --is-ci test-connectors` (all our connectors will be tested by default with this command).
* https://github.com/airbytehq/airbyte/blob/master/.github/workflows/connector_integration_tests_alpha.yml
* https://github.com/airbytehq/airbyte/blob/master/.github/workflows/connector_integration_tests.yml

Inspiration should be taken from this workflow to run `connectors-ci` in GHA:
https://github.com/airbytehq/airbyte/blob/master/.github/workflows/connector_integration_tests.yml#L1

alafanechere:
## What
Closes #23970
This new workflow is made to run nightly builds on the **large warm runner** the infra team provisioned.
It should be able to leverage caching to speed up nightly builds.

## How
* Declare a new scheduled workflow (everyday at 11AM UTC to not overlap with the future legacy nightly builds running earlier in the day).
* Run nightly builds for GA and Beta connectors only.

### DIFF
--- a/.github/workflows/connector_nightly_builds_dagger.yml
+++ b/.github/workflows/connector_nightly_builds_dagger.yml
@@ -1,0 +1,55 @@
+name: POC Connectors CI - nightly builds
+
+on:
+  schedule:
+    # 11AM UTC is 12AM CET, 4AM PST.
+    - cron: "0 11 * * *"
+  workflow_dispatch:
+
+jobs:
+  connectors_ci:
+    name: Connectors CI
+    timeout-minutes: 240 # 4 hours
+    runs-on: large-runner
+    steps:
+      - name: Get start timestamp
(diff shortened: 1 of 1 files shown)


c-p-b (approved):
Let’s see how it goes :)

PR #24278, opened 2023-03-21
Source file: do not read whole file on check and discover

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1681
Fix timing out `check` and `discover` commands

## How
Read only the header of a file on `check`.
Read a single chunk of data on `discover`.


airbyte-integrations/connectors/source-file/integration_tests/file_formats_test.py
```diff
@@ -79,29 +78,3 @@ def run_load_nested_json_schema(config, expected_columns=10, expected_rows=42):
     df = data_list[0]
     assert len(df) == expected_rows  # DataFrame should have 42 items
     return df
-
-
-# https://github.com/airbytehq/alpha-beta-issues/issues/174
-# this is to ensure we make all conditions under which the bug is reproduced, i.e.
-# - chunk size < file size
-# - column type in the last chunk is not `string`
-@patch("source_file.client.Client.CSV_CHUNK_SIZE", 1)
```
davydov-d, line 88: dropped this test because it doesn't represent the expected behavior from now on

airbyte-integrations/connectors/source-file/source_file/source.py
```diff
@@ -108,9 +109,8 @@ def check(self, logger, config: Mapping) -> AirbyteConnectionStatus:
         client = self._get_client(config)
         source_url = client.reader.full_url
         try:
-            with client.reader.open():
-                list(client.streams)
-                return AirbyteConnectionStatus(status=Status.SUCCEEDED)
+            list(client.streams(empty_schema=True))
```
davydov-d, line 112: Read only file header when running `check` to ensure the connection succeeds

airbyte-integrations/connectors/source-file/source_file/client.py
```diff
@@ -321,11 +323,16 @@ def load_dataframes(self, fp, skip_data=False) -> Iterable:
         reader_options = {**self._reader_options}
         try:
             if self._reader_format == "csv":
+                bytes_read = 0
                 reader_options["chunksize"] = self.CSV_CHUNK_SIZE
                 if skip_data:
                     reader_options["nrows"] = 0
                     reader_options["index_col"] = 0
-                yield from reader(fp, **reader_options)
```
davydov-d, line 328: Read only `self.CSV_CHUNK_SIZE` bytes of data to generate schema. Otherwise a time out is possible in case a large file is read

bnchrch (approved):

PR #24280, opened 2023-03-21
Connector Builder: Make connector_builder part of the CDK package

clnoll:
## What
We need to be able to install the `connector_builder` into the Connector Builder server's docker image, so it can use its entrypoint for CDK requests. The simplest way to do that is to have the `connector_builder` be part of the `airbyte-cdk` package, so that we can pip install `airbyte-cdk` when creating the Connector Builder server image.

In its current location, the `connector_builder` module was adjacent to the CDK package, so was not included in the package published to pypi. This PR moves it inside the package.

This PR doesn't change any functionality.


girarda (changes requested):
change looks directionally good, but the tests are failing because the imports in `test_message_grouper` weren't updated

clnoll:
Thanks @girarda! Fixed.

girarda (approved):
question @clnoll : would it make sense to extract the connector builder module from the CDK into an entirely new module, or do we want to keep it here long-term?

clnoll:
@girarda I've been going back and forth on that. Since we're using a new entrypoint it feels like we could definitely put it in a different package. OTOH that would mean we need to keep track of the connector builder package's version, in addition to the CDK version. i.e. in the Connector builder server's Dockerfile, we'd pip install the connector builder package at a particular connector builder version but would have to know which CDK version the connector builder version was pointing to. Since a new package would require this extra overhead, I feel that it makes sense to go with the original design for now.

PR #24324, opened 2023-03-21
Source Quickbooks: migrate to low-code

roman-yermilov-gl:
## What
*Migrate QuickBooks Singer to Low-Code*


airbyte-integrations/connectors/source-quickbooks/acceptance-test-config.yml
```diff
@@ -0,0 +1,46 @@
...
+  discovery:
+    tests:
+      - config_path: "secrets/config.json"
+  basic_read:
+    tests:
+      - config_path: "secrets/config.json"
+        configured_catalog_path: "integration_tests/configured_catalog.json"
+        empty_streams:
+          - name: vendor_credits
+            bypass_reason: "enable to populate"
```
lazebnyi, line 23: Typo in `bypass_reason`. enable -> unable

lazebnyi (changes requested):

airbyte-integrations/connectors/source-quickbooks/integration_tests/abnormal_state.json
```diff
@@ -0,0 +1,68 @@
+{
+  "accounts": {
```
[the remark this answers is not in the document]
roman-yermilov-gl: updated

airbyte-integrations/connectors/source-quickbooks/acceptance-test-config.yml
roman-yermilov-gl: fixed

lazebnyi (approved):
LGTM

davydov-d:
depends on https://github.com/airbytehq/airbyte/pull/24891

bazarnov (approved):

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
```diff
@@ -1676,14 +1676,6 @@
...
   sourceType: api
   releaseStage: alpha
-- name: QuickBooks
-  sourceDefinitionId: 29b409d9-30a5-4cc8-ad50-886eb846fea3
-  dockerRepository: airbyte/source-quickbooks-singer
-  dockerImageTag: 0.1.5
-  documentationUrl: https://docs.airbyte.com/integrations/sources/quickbooks-singer
-  icon: qb.svg
-  sourceType: api
-  releaseStage: alpha
```
evantahler, line 1686: This looks like you are just removing the connector?

docs/integrations/sources/quickbooks.md
```diff
@@ -39,45 +103,11 @@ This Source is capable of syncing the following [Streams](https://developer.intu
 - [VendorCredits](https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/vendorcredit)
 - [Vendors](https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/vendor)
 
-### Data type mapping
```
evantahler, line 42: Is this no longer useful?

davydov-d: Overlooked it. Now reverted, thanks!

airbyte-config/init/src/main/resources/seed/source_definitions.yaml
davydov-d: I'm replacing a singer based Quickbooks with a lowcode-based one. The seed for the new one should be generated after I publish the connector. So the old one is removed, am I doing it the wrong way?

airbyte-config-oss/init-oss/src/main/resources/seed/source_definitions.yaml
```diff
@@ -1676,14 +1676,6 @@
...
   sourceType: api
   releaseStage: alpha
-- name: QuickBooks
-  sourceDefinitionId: 29b409d9-30a5-4cc8-ad50-886eb846fea3
-  dockerRepository: airbyte/source-quickbooks-singer
-  dockerImageTag: 0.1.5
-  documentationUrl: https://docs.airbyte.com/integrations/sources/quickbooks-singer
-  icon: qb.svg
-  sourceType: api
-  releaseStage: alpha
```
evantahler, line 1686: There's still something wrong here - the connector should not be removed from the definitions

davydov-d: @evantahler this connector has never been enabled in cloud and the new one has a different name. Can you suggest what I should do to delete the old one and add the new connector?

evantahler: Shouldn't this low-code connector be using the same sourceDefinitionId and name?

davydov-d: It seems weird to me that the lowcode connector would be named `source-quickbooks-singer`

evantahler: Keep the ID, but change the name! That's OK!

evantahler: Tagging @pedroslopez (who is out until next week) - What's the right thing to do here?  We want to keep the same soureId so that people can upgrade, but we want to change the name of the connector

bazarnov: In this case, we should:
1) change the name first > merge
2) migrate to low-code (which is done) > merge
3) have a bunch of backward compatibility bypass for CAT between 1 and 2 to cover the name change (this will touch lots of files not just a single line, if we want to have a name changed everywhere, not just spec)

Apologies, but quickbooks-singer was not even worked, as far as i know, so it's a question what is the pool of users affected? @evantahler do you have possibility to check this? 
If there is any, we should probably change the name > migrate to low-code, otherwise - there is nothing to do but remove sourceDefinition of existing QuickBooks source, and replace with the new one for low-code. WDYT?

There is option number 3, of course: 
Release this change as separate connector (alpha) as low code connector and forget about singer-one for now. Announce the breaking change for the users > start using new implementation > come back and remove the old one. To much effort.

docs/integrations/sources/quickbooks.md
```diff
@@ -1,12 +1,76 @@
 # QuickBooks
 
-## Overview
+This page contains the setup guide and reference information for the QuickBooks Source connector.
 
-The QuickBooks source supports both Full Refresh and Incremental syncs. You can choose if this connector will copy only the new or updated data, or all rows in the tables and columns you set up for replication, every time a sync is run.
+## Prerequisites
 
-This source wraps the [Singer QuickBooks Tap](https://github.com/singer-io/tap-quickbooks).
+- Account at [Intuit Developer account](https://developer.intuit.com/app/developer/qbo/docs/get-started)
```
sherifnada, line 7: I don't think this is correct. A Cloud user only needs an Intuit Quickbooks account. This should only be relevant for OSS setup.

docs/integrations/sources/quickbooks.md
```diff
@@ -1,12 +1,76 @@
...
+
+**For Airbyte Cloud:**
+
+1. [Log into your Airbyte Cloud](https://cloud.airbyte.com/workspaces) account.
+2. In the left navigation bar, click **Sources**. In the top-right corner, click **+ new source**.
+3. On the source setup page, select **QuickBooks** from the Source type dropdown and enter a name for this connector.
+4. Click `Authenticate your QuickBooks account`
+5. Log in and Authorize to the QuickBooks account.
+6. **Start date** - The date from which you'd like to replicate data for streams.
+7. **Realm ID** - The Labeled Company ID you'd like to replicate data for streams.
```
sherifnada, line 43: where do I get this realm ID? Can we get it on behalf of the user?

docs/integrations/sources/quickbooks.md
```diff
@@ -1,12 +1,76 @@
...
+This page contains the setup guide and reference information for the QuickBooks Source connector.
 
-The QuickBooks source supports both Full Refresh and Incremental syncs. You can choose if this connector will copy only the new or updated data, or all rows in the tables and columns you set up for replication, every time a sync is run.
+## Prerequisites
 
-This source wraps the [Singer QuickBooks Tap](https://github.com/singer-io/tap-quickbooks).
+- Account at [Intuit Developer account](https://developer.intuit.com/app/developer/qbo/docs/get-started)
+- Start date
+- Realm ID
+- User Agent
```
sherifnada, line 10: Can't we retrieve this on behalf of the user? Or fill it in for them e.g: `airbyte-connector`?

docs/integrations/sources/quickbooks.md
```diff
@@ -1,12 +1,76 @@
...
 
-## Overview
+This page contains the setup guide and reference information for the QuickBooks Source connector.
 
-The QuickBooks source supports both Full Refresh and Incremental syncs. You can choose if this connector will copy only the new or updated data, or all rows in the tables and columns you set up for replication, every time a sync is run.
+## Prerequisites
 
-This source wraps the [Singer QuickBooks Tap](https://github.com/singer-io/tap-quickbooks).
+- Account at [Intuit Developer account](https://developer.intuit.com/app/developer/qbo/docs/get-started)
+- Start date
```
sherifnada, line 8: this shouldn't be in the pre-requisites section. Only things which need upfront preparation should be here e.g: if I need to go obtain an API key _before_ I start setting up the connector.

docs/integrations/sources/quickbooks.md
```diff
@@ -1,12 +1,76 @@
...
+### Step 2: Set up the QuickBooks connector in Airbyte
+
+**For Airbyte Cloud:**
+
+1. [Log into your Airbyte Cloud](https://cloud.airbyte.com/workspaces) account.
+2. In the left navigation bar, click **Sources**. In the top-right corner, click **+ new source**.
+3. On the source setup page, select **QuickBooks** from the Source type dropdown and enter a name for this connector.
+4. Click `Authenticate your QuickBooks account`
+5. Log in and Authorize to the QuickBooks account.
+6. **Start date** - The date from which you'd like to replicate data for streams.
```
sherifnada, line 42:
```suggestion
6. **Start date** - The date starting from which you'd like to replicate data.
```

airbyte-config-oss/init-oss/src/main/resources/seed/source_definitions.yaml
sherifnada: I think we shouldn't worry too much about existing users. It's dubious this connector ever really worked.

airbyte-integrations/connectors/source-quickbooks/source_quickbooks/components.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def __iter__(self) -> typing.Iterator:
+        return self._record
+
+    def __bool__(self):
+        return bool(self._record)
+
+
+@dataclass
+class CustomDatetimeBasedCursor(DatetimeBasedCursor):
```
sherifnada, line 51: please follow the process outlined [here](https://docs.google.com/document/d/1LB4gBk0slHj3zNmq1Qtj1U7VkKyTN9eFsVldgQ8T4HQ/edit) for recording issues for these components

docs/integrations/sources/quickbooks.md
```diff
@@ -1,12 +1,76 @@
...
+<!-- /env:cloud -->
+
+<!-- env:oss -->
+**For Airbyte Open Source:**
 
-### Output schema
+1. **Client ID** - The OAuth2.0 application ID
+2. **Client Secret** - The OAuth2.0 application secret
+3. **Refresh Token** - Refresh token used to get new access token every time the current one is expired
+4. **Access Token** - Access token to perform authenticated API calls with
```
sherifnada, line 56: Please prefix optional inputs with `(Optional)` otherwise this is overwhelming for the user to read

sherifnada (changes requested):

docs/integrations/sources/quickbooks.md
davydov-d: These fields are not optional, this [PR](https://github.com/airbytehq/airbyte/pull/20923) tells why

docs/integrations/sources/quickbooks.md
davydov-d: Good catch, thanks! Fixed

docs/integrations/sources/quickbooks.md
davydov-d: fixed

docs/integrations/sources/quickbooks.md
davydov-d: fixed

docs/integrations/sources/quickbooks.md
davydov-d: that's a good call -- done

docs/integrations/sources/quickbooks.md
davydov-d: The doc states we can retrieve the realm ID along with the code from the OAuth response (first one).
image

I'll try implementing this approach but it looks like we have never done this before

airbyte-integrations/connectors/source-quickbooks/source_quickbooks/components.py
davydov-d: done

docs/integrations/sources/quickbooks.md
sherifnada: Doesn't this mean we can retrieve it on behalf of the user during the oauth flow? (And therefore this field should be hidden in the UI in cloud)?

davydov-d: Done https://github.com/airbytehq/airbyte-platform-internal/pull/5743

davydov-d:
@sherifnada please take a look once more

sherifnada (approved):

PR #24336, opened 2023-03-21
Generate metadata files from catalog

### LINKED ISSUE #24010: Backfill metadata files
# Problem
We need metadata files to power our metadata service

The first step is to attempt to generate these from the catalog and note in which way we are missing any.

# Solution

What this is
1. Generating metadata.yml and spec files from the catalogs
2. Validating that they are correct
3. Noting which connectors are missing metadata or have invalid metadata
4. Backfilling the above manually
5. Creating a PR with the metadata files

What this is not
- Validating metadata files on each PR

### LINKED ISSUE #24017: Validate metadata file(s)

bnchrch:
closes #24017 
closes #24010
# Video
This is slightly out of date but it communicates the spirit of this all!
https://www.loom.com/share/9a005ea32d2442f08aa6e4f825829390

## What
We need to be able to backfill metadata files from the catalog for the migration to the new catalog definition system

## How
This adds a Dagster assets responsible for
1. Checking if we have the connector checked into source (github assets!)
2. Checking if we have a cached spec (spec cache assets!)
3. Converting into individual metadata files from the oss and cloud catalogs
4. Checking that those metadata files are valid with the spec and the catalog they came from
5. Adds a new file manager for writing files locally
6. Add the begining of spec cache specific code to the metadata service.

## Notes for reviewer
So sorry this crossed the 1k line mark. The comments snuck up on me.

Some things to keep in mind
1. dev_assets will be short lived until the orchestrator is used in production
2. a lot of this is in service of preparation for the cut over to metadata files.
3. this is a start of a new project so style is important, if you dont like the feel of how I did something let me know

## Recommended reading order
1. [airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py](https://github.com/airbytehq/airbyte/pull/24336/files#diff-ca730c97dbba6558268fb071f3d0a4295e204dfef475c65816a102750427163a)
4. [airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/github_assets.py](https://github.com/airbytehq/airbyte/pull/24336/files#diff-325007862ff6be5564363445c9a6173561d24a156045b08f0e8b7a3884b59d7d)
5. [airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/spec_cache_assets.py](https://github.com/airbytehq/airbyte/pull/24336/files#diff-4684bf8e9ee7f5c42853af944780b7f20af7e9c4721323ea13043dad6f547ccb)
6. [airbyte-ci/connectors/metadata_service/lib/metadata_service/spec_cache.py](https://github.com/airbytehq/airbyte/pull/24336/files#diff-ed9b4698908a369dbca0a8a00c429a9bb4bdc70aac8e450d671d69ee8f125f1c)
7. [airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata_assets.py](https://github.com/airbytehq/airbyte/pull/24336/files#diff-26c1f0601c26ea8c3b1322ec0d85c008b21ce4a436082dcbe5b32d01682f7e2f)
8. [...-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_report_assets.py](https://github.com/airbytehq/airbyte/pull/24336/files#diff-e67a768db69cf4c9a06fd1d7da1148e1b4ab9058232ce2aad6fdf496ce3f0239)


airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_report_assets.py
```diff
@@ -0,0 +1,164 @@
...
+from dagster import MetadataValue, Output, asset
+
+from ..templates.render import render_connector_catalog_locations_html, render_connector_catalog_locations_markdown
+
+GROUP_NAME = "catalog_reports"
+
+# HELPERS
+
+
+def augment_and_normalize_connector_dataframes(
```
bnchrch, line 11: This was largely existing code, refactored to a named function.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_report_assets.py
```diff
@@ -0,0 +1,164 @@
...
+    # Replace NaN values in the 'is_cloud' and 'is_oss' columns with False
+    total_catalog[["is_cloud", "is_oss"]] = total_catalog[["is_cloud", "is_oss"]].fillna(False)
+
+    catalog_with_metadata = pd.merge(
+        total_catalog, valid_metadata_list[["definitionId", "is_metadata_valid"]], left_on=primaryKey, right_on="definitionId", how="left"
+    )
+
+    # merge with cached_specs on dockerRepository and dockerImageTag
+    cached_specs["is_spec_cached"] = True
+    merged_catalog = pd.merge(
```
bnchrch, line 39: Net new

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_report_assets.py
```diff
@@ -0,0 +1,164 @@
...
+        catalog_with_metadata,
+        cached_specs,
+        left_on=["dockerRepository", "dockerImageTag"],
+        right_on=["docker_repository", "docker_image_tag"],
+        how="left",
+    )
+
+    # Set connectorType to 'source' or 'destination'
+    merged_catalog["connector_type"] = connector_type
+    merged_catalog["is_source_controlled"] = merged_catalog["dockerRepository"].apply(
```
bnchrch, line 49: net new

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/dev_assets.py
```diff
@@ -0,0 +1,166 @@
+import yaml
+from dagster import Output, asset
+
+"""
+NOTE: This file is temporary and will be removed once we have metadata files checked into source control.
+
+TODO (ben): Remove this file once we have metadata files checked into source control.
```
bnchrch, line 7: this file is meant to be short lived as once weve created these metadata files and checked them into source control we no long need to generate them

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/resources/file_managers/local_file_manager.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class SimpleLocalFileManager(LocalFileManager):
+    """
+    HACK WARNING: This is a hack to get around the fact that the LocalFileManager in dagster does not
+    expose the key parameter and does not handle nested directories.
+
+    Much of this code is borrowed from the LocalFileManager in dagster and modified slightly.
+
+    TODO(ben): Remove this once catalogs are live
```
bnchrch, line 23: Another file thats around only so long as metadata files are not yet checked in

airbyte-ci/connectors/metadata_service/orchestrator/tests/test_debug.py
```diff
@@ -1,20 +1,22 @@
...
+from orchestrator.assets.metadata_assets import (
+    catalog_derived_metadata_definitions,
 )
 
-from orchestrator.config import REPORT_FOLDER, CATALOG_FOLDER
+
+from orchestrator.config import REPORT_FOLDER, CATALOG_FOLDER, CONNECTORS_PATH, CONNECTOR_REPO_NAME
 
 
 def debug_catalog_projection():
```
bnchrch, line 22: expand to see the comment below on why we have this function at all

evantahler (commented):
I don't feel qualified to review the Python, but if this acomplishes the steps showcased in the video, 👍 from me!

evantahler (approved):

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/dev_assets.py
```diff
@@ -0,0 +1,166 @@
...
+"""
+
+GROUP_NAME = "dev"
+
+OVERRIDES = {
```
erohmensing, line 12: This is all the info we added which was missing in the loom, right?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata_assets.py
```diff
@@ -0,0 +1,182 @@
...
+    for override_col in allowed_overrides:
+        oss_col = override_col + OSS_SUFFIX
+        cloud_col = override_col + CLOUD_SUFFIX
+
+        cloud_value = merged_df.get(cloud_col)
+        oss_value = merged_df.get(oss_col)
+
+        # if the columns are different, add the cloud value to the overrides
+        if cloud_value and not are_values_equal(oss_value, cloud_value):
+            catalogs["cloud"][override_col] = merge_values(oss_value, cloud_value)
```
erohmensing, line 90: If I understand it correcty, this part checks whether, for an identical list of connectors on cloud and connectors on OSS, if the columns are different, and if so can perform overrides based on column differences (e.g. pinning a version). 

If we have connectors that are only on oss or cloud, it looks like we will exit early on L63 - do we then miss out on this column-based override?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata_assets.py
```diff
@@ -0,0 +1,182 @@
...
+            "connectionType": get_field_with_fallback(merged_df, "sourceType"),
+            "releaseStage": get_field_with_fallback(merged_df, "releaseStage"),
+            "license": "MIT",
+            "supportsDbt": get_field_with_fallback(merged_df, "supportsDbt"),
+            "supportsNormalization": get_field_with_fallback(merged_df, "supportsNormalization"),
+            "allowedHosts": get_field_with_fallback(merged_df, "allowedHosts"),
+            "normalizationConfig": get_field_with_fallback(merged_df, "normalizationConfig"),
+            "suggestedStreams": get_field_with_fallback(merged_df, "suggestedStreams"),
+            "resourceRequirements": get_field_with_fallback(merged_df, "resourceRequirements"),
+        }
```
erohmensing, line 120: So much metadata 🤩

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata_assets.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # remove none values
+        data = {k: v for k, v in raw_data.items() if v is not None}
+
```
erohmensing, line 124: Doing so allows it to fail the `validate_metadata` with regards to missing keys, right?

erohmensing (approved):
This is exciting, and your loom was once again very helpful! Can't wait to have those views easily accessible - the ✅ and ❌ make things so quick to parse.

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/generated/ConnectorMetadataDefinitionV1.py
```diff
@@ -137,7 +137,7 @@ class Data(BaseModel):
     license: str
     supportUrl: AnyUrl
     githubIssueLabel: str
-    sourceType: str
+    connectionType: str
```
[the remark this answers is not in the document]
bnchrch: Ya, looks like git made an incorrect looking diff 😅 

However if you have better names for `connectorType` and `connectionType` im all ears... not happy with how close they are. Feels like the Dagger vs Dagster name problem

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/dev_assets.py
bnchrch: Oooh I realized I should add a comment here.

But yes! These are all the manual catalog adjustments we must make for all generated metadata files to pass validation

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata_assets.py
bnchrch: Good eye Ella!

I think this logic is still correct. But it always felt like a scratchy sweater. So let me lay out some background context, then how this is meant to work.

**Types of catalog entries**
There are four scenarios to know about for a connector as it relates to a catalog
1. available in oss but not cloud
2. available in oss and cloud 
3. not available in oss but available in cloud
4. not available in either (we can ignore this for now)

**How they relate to overrides**
Scenario 1 its straight forward. All oss values are declared under `data` and `data.catalogs.oss.enabled` is `true`

Scenario 2 is also straight forward. All oss values are declared under `data`, `data.catalogs.oss.enabled` is `true`, `data.catalogs.cloud.enabled` is `true`, and any cloud specific values is under `data.catalogs.cloud`

Scenario 3 is straight forward to say, but the logic in the code feels funny. All cloud values are declared under `data` and `data.catalogs.cloud.enabled` is `true`. So no overrides occur, because its only in one catalog

**Onto the code**
Ok so looking at the code were commenting on.
1. If its a cloud or oss only connector, we can skip overrides. 
2. For differences in field values between the connector definition in cloud vs oss, the cloud value should be placed under `data.catalogs.cloud`

Does that make sense? Is there an edge case im missing?

oh and also for reference there is a [child PR](https://github.com/airbytehq/airbyte/pull/24545) to this one that diffs the new catalogs with the old one

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata_assets.py
bnchrch: Basically this is meant to remove optional fields that didn't have a value defined

With out this a lot of connectors metadata files would look like
```yaml
metadata_version: 1.0
data:
    connectorId: abc123
    allowedHosts: null
    supportsNormalization: null
```

bnchrch: Which isnt invalid, we would just prefer to if youre not using an optional field to not have it declared

erohmensing: That makes sense for things that don't apply to all connectors (e.g. supportsNormalization) - didn't think of that, thought all were required. Makes more sense

erohmensing: My only thought is that if there were ones that are technically optional but highly encouraged (allowedhosts, maybe docs url, etc) - having the keys in there with null values would make us more aware that they're missing and should be fixed, versus it being harder to notice when the keys are missing. but with some being completely irrelevant, we probably can't have it both ways - this way works!

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_report_assets.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    Normalize the cloud and oss connector dataframes and merge them into a single dataframe.
+    Augment the dataframe with additional columns that indicate if the connector is in the cloud catalog, oss catalog, and if the metadata is valid.
+    """
+
+    # Add a column 'is_cloud' to indicate if an image/version pair is in the cloud catalog
+    cloud_df["is_cloud"] = True
+
+    # Add a column 'is_oss' to indicate if an image/version pair is in the oss catalog
+    oss_df["is_oss"] = True
```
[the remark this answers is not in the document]
erohmensing: I didn't notice that these were 2 columns on separate data frames! I thought we were instantiating 2 cols on one data frame and then filling with either true or false, this makes sense with that context 

Thanks for the explanation!

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/generated/ConnectorMetadataDefinitionV1.py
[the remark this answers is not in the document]
bnchrch: `connectionType` is in use

Before it was called `sourceType` and used only for sources to inidicate whether it was a `api`, `file` or `database` source.

But this updates `sourceType` to now be called `connectionType` allowing destinations to do the same self reporting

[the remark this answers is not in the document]
erohmensing: Ah, I'm recommend we call it `connectorType` instead! Since that's what we use elsewhere and I think would be the correct term. Or do you think of `connectorType` as referring to source/destination vs `connectionType` being api/file/database?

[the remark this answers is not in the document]
bnchrch: Thats the tricky part! `connectorType` is being used just like you described

>  connectorType as referring to source/destination vs connectionType being api/file/database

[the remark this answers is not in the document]
erohmensing: Aha. Word! Okay, so yeah, a dagger/dagster situation. 

How about something like `typeTags`? I'm imagining your could have multiple if we ended up defining more (e.g. `[api, python]`) - this feels less like a hard classification and more like tagging for extra context. 

Or even `connectorSubtype` or something of the sort? I feel like we might end up using this to convey more than just the `://` connection type... but I could be wrong!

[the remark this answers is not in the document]
bnchrch: Im into that! but! we do program on this field in cloud if I recall. Basically the value corresponds to the same column that can have the value `custom` which then the platform has a few lines of code sprinkeled around the codebase that go `if sourceType === 'custom'`

Its not ideal, but Im currently thinking we dont open that refactor can of worms, by changing this to an array.

Yet 😅 

But connectorSubtype works! I like that!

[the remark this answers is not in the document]
erohmensing: Cool, yeah a refactor is unnecesary at the moment, but definitely something to think about! Good to know we're actually using this field at the moment

PR #24345, opened 2023-03-22
Add missing type hints for the message grouper

girarda:
## What
* Some methods of the message grouper are missing type hints

## Recommended reading order
1. `message_grouper.py`


brianjlai (approved):
lgtm!

PR #24369, opened 2023-03-22
Airbyte CDK: skip slices in a substream slicer

davydov-d:
## What
Some source APIs may return deleted entities, whereis subentities associated with them would be inaccessible and result in 404 or other errors -- this should be handled.

## How
Introduce a `skip_records` parameter for a `SubStreamPartitionRouter.ParentStreamConfig` class. Usage example:
```
  contact_stream:
    retriever:
      $ref: "#/definitions/retriever"
      partition_router:
        type: SubstreamPartitionRouter
        parent_stream_configs:
          - type: ParentStreamConfig
            stream: "#/definitions/customer_stream"
            parent_key: id
            partition_field: id
            skip_slices: "{{ parent_record['deleted'] }}"
```

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -852,6 +852,8 @@ definitions:
         "$ref": "#/definitions/DeclarativeStream"
       partition_field:
         type: string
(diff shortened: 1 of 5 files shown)


girarda (commented):
I don't have much context on the connector we're trying to fix the issue for, but would a error handler that skips requests resulting in 404 solve your problem @davydov-d ?

PR #24373, opened 2023-03-22
feat: Enable CDC checkpointing on Postgres

sergio-ropero:
## What
Enable CDC checkpointing for Postgres DB source.

## How
Add a try-catch block when we try to read the offset file when Debezium is writing, so we can skip the current loop and try to send it back in next iteration.

As in the example below we see that **it generated only 1 state message** for a 1M row sync even it's configured by default to do it every 10k or 15 minutes (see details in the test).
It's expected as the Offset should be in a position we already has been reached in our sync with the destination.

### MySQL
Screenshot 2023-03-22 at 19 46 31

### Postgres
Tested two sync of 1M rows:
Screenshot 2023-03-22 at 19 45 16

Several messages during the sync (27) show us that the file can not be accessed but it will retry:
Screenshot 2023-03-22 at 19 56 27

On the second sync (CDC) output, we see 2 states sent by the source (one checkpoint and the state at the end):
```
2023-03-22 18:37:33 INFO i.a.w.g.DefaultReplicationWorker(getReplicationOutput):538 - sync summary: {
  "status" : "completed",
  "recordsSynced" : 1000000,
  "bytesSynced" : 155000000,
  "startTime" : 1679510159334,
  "endTime" : 1679510253775,
  "totalStats" : {
    "bytesEmitted" : 155000000,
    "destinationStateMessagesEmitted" : 2,
    "destinationWriteEndTime" : 1679510253775,
    "destinationWriteStartTime" : 1679510159405,
    "meanSecondsBeforeSourceStateMessageEmitted" : 44,
    "maxSecondsBeforeSourceStateMessageEmitted" : 87,
    "maxSecondsBetweenStateMessageEmittedandCommitted" : 0,
    "meanSecondsBetweenStateMessageEmittedandCommitted" : 0,
    "recordsEmitted" : 1000000,
    "recordsCommitted" : 1000000,
    "replicationEndTime" : 1679510253775,
    "replicationStartTime" : 1679510159334,
    "sourceReadEndTime" : 1679510253503,
    "sourceReadStartTime" : 1679510159370,
    "sourceStateMessagesEmitted" : 2
  },
  "streamStats" : [ {
    "streamName" : "test_table",
    "stats" : {
      "bytesEmitted" : 155000000,
      "recordsEmitted" : 1000000,
      "recordsCommitted" : 1000000
    }
  } ]
}
```


airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
```diff
@@ -144,9 +145,13 @@ protected AirbyteMessage computeNext() {
...
           // the assignation
-          final HashMap<String, String> temporalOffset = (HashMap<String, String>) offsetManager.read();
-          if (!cdcStateHandler.isSameOffset(previousCheckpointOffset, temporalOffset)) {
-            checkpointOffsetToSend.putAll(temporalOffset);
+          try {
+            final HashMap<String, String> temporalOffset = (HashMap<String, String>) offsetManager.read();
+            if (!cdcStateHandler.isSameOffset(previousCheckpointOffset, temporalOffset)) {
+              checkpointOffsetToSend.putAll(temporalOffset);
+            }
+          } catch (ConnectException e) {
```
subodh1810, line 153: This can be final

subodh1810 (approved):
Good stuff :D

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumStateDecoratingIterator.java
sergio-ropero: Done 😊

sergio-ropero:
/publish connector=connectors/source-mssql-strict-encrypt auto-bump-version=false run-tests=false

PR #24421, opened 2023-03-23
🎉 Compare versions tool

lazebnyi:
## What
To compare in memory records outputs need some tool

## How
Created a bash script that pull the images for the two docker versions and calculated the difference of the record output

### DIFF
--- a/tools/internal/compare_versions.sh
+++ b/tools/internal/compare_versions.sh
@@ -29,27 +29,20 @@ if [ "$?" -eq 1 ]; then
     state="/config_files/state.json"
 else
     state_status="without"
-    state='" "'
+    state=""
 fi
-
-
-printf "\nPulling connectors images:\n"
-
-docker pull airbyte/"$connector":"$version_1"
-docker pull airbyte/"$connector":"$version_2"
-
 
 printf "\nStarting comparing connector %s version %s with version %s %s state:\n" "$connector" "$version_1" "$version_2" "$state_status"
 
 run_docker_image()
 {
-  docker run --rm -v $(pwd)/config_files:/config_files airbyte/"$connector":"$1" read --config /config_files/secrets/config.json --catalog /config_files/configured_catalog.json --state "$state"
+  docker run --pull=missing --rm -v $(pwd)/config_files:/config_files airbyte/"$connector":"$1" read --config /config_files/secrets/config.json --catalog /config_files/configured_catalog.json --state "$state"
 }
 
-result=$( diff <(run_docker_image "$version_1") <(run_docker_image "$version_2") | grep 'RECORD' | sed '/\"emitted_at\"/,/}/ d')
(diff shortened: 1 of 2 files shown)


evantahler (commented):
Nice!  

✔️ This keeps records in memory
✔️ It compares the output

Some docs might be nice, e.g. a readme or a comment at the top
* Looks like this needs to be run from the root dir of the connector (where /secrets is relatively nearby)
* You probably don't want to run a whole sync, so some thought should be used when choosing a starting state?  Or, maybe you do want to run a "full" sync

It also might be helpful to use a visual diff tool.  [Here are some](https://stackoverflow.com/questions/1991248/any-visual-diff-in-linux-console).

tools/internal/compare_versions.sh
```diff
@@ -0,0 +1,56 @@
...
+else
+    state_status="without"
+    state='" "'
+fi
+
+
+printf "\nPulling connectors images:\n"
+
+docker pull airbyte/"$connector":"$version_1"
+docker pull airbyte/"$connector":"$version_2"
```
evantahler, line 39: Is there ever a scenario where you would be testing this on a pre-release (#dev) version?  If so, docker-pull won't work.  I think `docker run` will pull the image if needed, so you might be able to skip these

tools/internal/compare_versions.sh
```diff
@@ -0,0 +1,56 @@
...
+
+
+printf "\nStarting comparing connector %s version %s with version %s %s state:\n" "$connector" "$version_1" "$version_2" "$state_status"
+
+run_docker_image()
+{
+  docker run --rm -v $(pwd)/config_files:/config_files airbyte/"$connector":"$1" read --config /config_files/secrets/config.json --catalog /config_files/configured_catalog.json --state "$state"
+}
+
+result=$( diff <(run_docker_image "$version_1") <(run_docker_image "$version_2") | grep 'RECORD' | sed '/\"emitted_at\"/,/}/ d')
```
evantahler, line 49: `diff` - neat!

evantahler (approved):

tools/internal/compare_versions.sh
sherifnada: should we also compare STATE messages?

sherifnada (changes requested):
should we also be comparing STATE messages? it's an important part of the behavior

tools/internal/compare_versions.sh
lazebnyi: shifted pull to run stage

tools/internal/compare_versions.sh
lazebnyi: Updated

YowanR:
@lazebnyi Please take a look at the latest comments and let's aim to wrap this up on Monday, please 🙏

lazebnyi:
@YowanR thanks. It look like I already fixed few day ago) - https://github.com/airbytehq/airbyte/pull/24421#discussion_r1147578631

sherifnada (approved):

PR #24429, opened 2023-03-23
Source S3: read a single record on check

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1697

## How
When calling `check`, we make a test read. To save time and memory, we need to read a single record instead of a whole slice. Since the block size in a config may be huge, we lower its value for the `check` command.


wmikemoon:
@davydov-d any update here?

lazebnyi (approved):
Nice!

PR #24431, opened 2023-03-23
Fix destination integration tests

grishick:
Fixing integration tests for:
 - destination-mssql
 - destination-postgres
 - destination-postgres-strict-encrypt

Fixing some of the integration tests for:
 - destination-mysql
 - destination-clickhouse


airbyte-config/init/src/main/resources/seed/destination_definitions.yaml
```diff
@@ -93,7 +93,7 @@
     normalizationRepository: airbyte/normalization-clickhouse
     normalizationTag: 0.4.0
     normalizationIntegrationType: clickhouse
-  supportsDbt: true
+  supportsDbt: false
```
ryankfu, line 96: did clickhouse not support `dbt` or are we turning off normalization for clickhouse here?

airbyte-integrations/bases/standard-destination-test/src/main/java/io/airbyte/integrations/standardtest/destination/DestinationAcceptanceTest.java
```diff
@@ -130,7 +130,8 @@ private String getImageNameWithoutTag() {
     return getImageName().contains(":") ? getImageName().split(":")[0] : getImageName();
   }
 
-  private Optional<StandardDestinationDefinition> getOptionalDestinationDefinitionFromProvider(final String imageNameWithoutTag) {
+  protected Optional<StandardDestinationDefinition> getOptionalDestinationDefinitionFromProvider(
+                                                                                                 final String imageNameWithoutTag) {
```
ryankfu, line 134: Nit: seems weird this got moved to another line

ryankfu (approved):
LGTM, minor nit but better to unblock integration tests

airbyte-config/init/src/main/resources/seed/destination_definitions.yaml
grishick: Clickhouse does not actually support custom `dbt` transformations and clickhouse `spec.json` has `supportsDbt` set to `false`. So there was a discrepancy between `destination_definitions.yaml `and `spec.json`. My guess is this happened with normalization configs were moved to `destination_definitions.yaml` via a script (@evantahler is this a fair assumption?)

davinchia (commented):
Looks good! I just came out of a meeting where we were discussed being able to note incremental normalisation support per connection in the test - looks like you already got there!

PR #24433, opened 2023-03-23
Update user impact section with more info about breaking changes

erohmensing:
## What
* Update the PR template to better explain how PR authors know if their changes are breaking and what they should do if they are.


evantahler (approved):
🖊️ Thanks!

marcosmarxm (approved):

PR #24437, opened 2023-03-23
Update strava and paystack acceptance-test-config.ymls

erohmensing:
These connectors were alpha when I originally automated the adding of bypass reasons in https://github.com/airbytehq/airbyte/pull/23985. After pulling in master, they're beta, and we don't want them to fail in connector health! (I'd already run all the tests, including alphas -- these ones failed due to extra columns).


bnchrch (approved):
go go go 🏃

PR #24446, opened 2023-03-24
Make ssl mode default to `require` for postgres on cloud

lmossman:
## What
As discussed on slack [here](https://airbytehq-team.slack.com/archives/C04QVJR9E78/p1679502400854919?thread_ts=1679433393.329999&cid=C04QVJR9E78), we would like to make the SSL Mode for postgres on airbyte cloud default to `require`, so that it works by default when configuring the connector on cloud (currently the default option of `disable` causes an error when users try to set up the source).

## How
Change the `modifySpec` method in the postgres strict encrypt source (the one used for cloud) to set the `default` key on the SSL Mode property to `require`, which will make the dropdown in the UI default to that option.


flash1293 (approved):
Tested and works as expected, LGTM

rodireich (approved):

PR #24447, opened 2023-03-24
Always show Service Account Key JSON in BigQuery destination

lmossman:
## What
As discussed on slack [here](https://airbytehq-team.slack.com/archives/C02TL38U5L7/p1679637379139249?thread_ts=1679522831.462279&cid=C02TL38U5L7), the `Service Account Key JSON` field in BigQuery is labelled as being required for cloud and optional for OSS, but since the field is technically marked as optional it is being put into a collapsed `Optional fields` group at the moment.

In order to reduce some confusion here, this field should always be shown, so that cloud users see the label describing it as required.

## How
Set the `always_show` prop to true on this field, causing it to look like this in the UI: 
Screen Shot 2023-03-23 at 11 15 23 PM

Ideally this should be solved by not having this wonky field that is marked as optional but actually required for cloud. It should be marked as required in the json spec for cloud, and optional for OSS, but we don't currently have the ability to do that (though I believe connector ops is going to be working on implementing this ability soon?)


flash1293 (approved):
Spec changes LGTM, did not test locally

lmossman:
/publish connector=connectors/destination-bigquery-denormalized

PR #24454, opened 2023-03-24
Source Slack: revert version 0.1.24

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1709

## How
Reverting to previous version because the Slack API seems to apply heavy rate limiting when we use page size much bigger then recommended (1000 vs 200).

### DIFF
--- a/airbyte-integrations/connectors/source-slack/source_slack/source.py
+++ b/airbyte-integrations/connectors/source-slack/source_slack/source.py
@@ -21,7 +21,7 @@ from pendulum import DateTime, Period
 class SlackStream(HttpStream, ABC):
     url_base = "https://slack.com/api/"
     primary_key = "id"
-    page_size = 1000
+    page_size = 100
 
     @property
     def max_retries(self) -> int:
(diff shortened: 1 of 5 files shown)


airbyte-integrations/connectors/source-slack/source_slack/source.py
```diff
@@ -21,7 +21,7 @@
 class SlackStream(HttpStream, ABC):
     url_base = "https://slack.com/api/"
     primary_key = "id"
-    page_size = 1000
+    page_size = 100
```
marcosmarxm, line 24: @davydov-d can we add a note explaining the reason of choosing 100 and why we're not going to change? I'm saying this because in Slack API and from our tests increasing to 1000 works.

davydov-d: Sure, will do

lazebnyi (approved):

pedroslopez (commented):
I'm fine with rolling back, but I don't think this was the cause of that oncall issue since the connections ran before the page size increase was introduced. See my comments on https://github.com/airbytehq/oncall/issues/1709#issuecomment-1483069970

PR #24457, opened 2023-03-24
Source Github: decrease give up rate

### LINKED ISSUE #24314: Decrease Give-up rate for GitHub

darynaishchenko:
Added validation and transformation for repositories config.
- added removing simple incorrect symbols that user can provide to repositories config
- added validation to highlight invalid config repositories config 
- fixed test `/test connector=connectors/source-github` - added correct repository name, because  user can provide only repository name and connector calls endpoint that return 404. (it can be the reason of give up)


lazebnyi (approved):
Nice!

garrrikkotua:
@darynaishchenko configuration check is not implemented correctly. For instance, it doesn't allow ligit repository CrowdDotDev/crowd.dev and it breakes our pipeline in AirByte Cloud

PR #24471, opened 2023-03-24
Source Stripe: fix stream slicing

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1713

## How
Refactoring in this [PR](https://github.com/airbytehq/airbyte/pull/24051) introduced a bug: some streams yielded `None` as a slice instead of `{}`. As a rule we yield a `None` slice if a `start_date` is > `now()` so we dont make request to read the data in the future and pass an abnormally large incremental read test. A check in `read_records` skips this slice.

This wasn't caught by the CAT because we dont have those streams well seeded (perhaps it's impossible). Now the slicing is fixed.


pedroslopez (approved):

pedroslopez:
Latest failure is due to expected records mismatch, which is happening on master and is a  known failure, so publishing anyway.

PR #24523, opened 2023-03-24
Source Github: Add undeclared columns to spec

grubberr:
## What
Issue: https://github.com/airbytehq/airbyte/issues/24566

- Add new fields to different streams
- Cosmetic fix: `"type": ["string"]` -> `"type": "string"`
- Make `"repository"` field required: `"type": ["null", "string"] -> "type": "string"`
- Extract some sub-schemas for `"user.json"`, `"reactions.json"`
- disable backward compatibility checks reasons:
  - a       Item root['collaborators']['properties']['role_name'] added to dictionary.
  - b       Type of root['assignees']['properties']['repository']['type'] changed from list to str and value changed from `['string']` to `"string"`.
  - c       Type of root['deployments']['properties']['repository']['type'] changed from list to str and value changed from `['null', 'string']` to `"string"`.
  - d       Type of root['project_cards']['properties']['project_id']['type'] changed from list to str and value changed from `['null', 'integer']` to `"integer"`.
  - e       Type of root['deployments']['properties']['creator']['type'] changed from str to list and value changed from `"object"` to `['null', 'object']`.

- [ ] Code reviews completed
- [ ] Connector version has been incremented
    - [ ] Version has been bumped according to our [Semantic Versioning for Connectors](https://docs.airbyte.com/contributing-to-airbyte/semantic-versioning-for-connectors) guidelines
    - [ ] `Dockerfile` has updated version
- [ ] Documentation updated
    - [ ] Connector's `README.md`
    - [ ] Changelog updated in `docs/integrations/<source or destination>/<name>.md` with an entry for the new version. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)

### Airbyter

- [ ] Build is successful

</details>


erohmensing:
Please make sure to remove the `fail_on_extra_columns: false` when you're done adding the undeclared columns! :)

airbyte-integrations/connectors/source-github/source_github/schemas/branches.json
```diff
@@ -3,7 +3,7 @@
   "type": "object",
   "properties": {
     "repository": {
-      "type": ["string"]
+      "type": "string"
```
erohmensing, line 6: Just wondering, this doesn't functionally change anything as to how it works, right?

erohmensing: Also, is `repository` a required field on all of these objects? If so should we declare that in the json schema?

airbyte-integrations/connectors/source-github/source_github/schemas/commit_comments.json
```diff
@@ -45,6 +45,9 @@
     },
     "author_association": {
       "type": ["null", "string"]
+    },
+    "reactions": {
+      "$ref": "reactions.json"
     }
```
erohmensing, line 51: Do we know if we want to pass along this information with the commit_comments (and other things that can also report reactions as a column), when we allow users to configure streams for these specific reactions? 

In other words, if I were a user who decided I didn't want to sync the `commit_comment_reactions` stream, I might not be stoked that the reactions are still being stored in the `commit_comments` stream. This is just a thought and not an opinion either way - thought it was an interesting thing to think about. (One could probably say the same thing about their API, but you can't as explicitly opt out of information in that case)

airbyte-integrations/connectors/source-github/source_github/schemas/issue_events.json
```diff
@@ -73,6 +90,244 @@
           "$ref": "user.json"
         }
       }
+    },
+    "performed_via_github_app": {
```
erohmensing, line 94: Nit: should we make a json for this, since we define it twice and it's long?

airbyte-integrations/connectors/source-github/source_github/schemas/organizations.json
```diff
@@ -118,9 +118,27 @@
...
+    },
     "members_can_create_pages": {
       "type": ["null", "boolean"]
     },
+    "members_can_fork_private_repositories": {
+      "type": ["null", "integer"]
+    },
+    "web_commit_signoff_required": {
+      "type": ["null", "integer"]
+    },
```
erohmensing, line 141: I think most of these should be `boolean`, according to the docs [here](https://docs.github.com/en/enterprise-server@3.6/rest/orgs/orgs#get-an-organization) (the names tipped me off)

airbyte-integrations/connectors/source-github/source_github/schemas/workflow_jobs.json
```diff
@@ -24,21 +30,25 @@
...
-      "format": "date-time",
-      "type": "string"
+      "type": "string",
+      "format": "date-time"
     },
     "completed_at": {
-      "format": "date-time",
-      "type": ["string", "null"]
+      "type": ["null", "string"],
+      "format": "date-time"
```
erohmensing, line 51: Any other (old or new) date strings that we should add a `format` for? Can you remind me how this is used when in a schema vs in a spec? (In a spec its used for the frontend to create a date-time-picker for the user to use)

airbyte-integrations/connectors/source-github/source_github/schemas/workflow_jobs.json
```diff
@@ -84,13 +94,16 @@
...
     },
     "runner_group_id": {
       "type": ["integer", "null"]
     },
     "runner_group_name": {
-      "type": ["string", "null"]
+      "type": ["null", "string"]
+    },
+    "repository": {
+      "type": ["null", "string"]
```
erohmensing, line 106: Is this nullable? Most `repository` keys don't seem like they are

airbyte-integrations/connectors/source-github/source_github/schemas/workflow_runs.json
```diff
@@ -526,6 +423,12 @@
           "type": ["null", "string"]
         }
       }
+    },
+    "actor": {
+      "$ref": "user.json"
+    },
+    "triggering_actor": {
+      "$ref": "user.json"
```
erohmensing, line 431: I see that in the repo, there is a `users.json` schema for the users stream. I don't see a `user.json` anywhere - is this ref correct? (Used many other places too)

erohmensing (changes requested):
This looks like it was a significant amount of work - thank you! 

My 2 blocking concerns are 
* `integer` vs `boolean` types for the organization params
* `users.json` vs `user.json` reference

airbyte-integrations/connectors/source-github/source_github/schemas/organizations.json
grubberr: oops my fault, I have fixed!

I tried to understand how this could happen because I am using scripts for schema generation stuff
I have realized when I tweaked the schema for this stream I used buggy  `obfuscate.py` script (later it was fixed)

anyway, thank for catching it - schema regenerated!

airbyte-integrations/connectors/source-github/source_github/schemas/branches.json
grubberr: 2. about `repository`
yes I made it required, except stream Teams stream - `repository` key removed!

airbyte-integrations/connectors/source-github/source_github/schemas/workflow_jobs.json
grubberr: I have changed it to be `non-nullable`

airbyte-integrations/connectors/source-github/source_github/schemas/workflow_jobs.json
grubberr: in stream schema `"format": "date-time"` is used for normalization to create on SQL databases DATETIME column (not STRING column)

BTW, here I just moved `format` below "type" - it's  cosmetic change, I did not add any new formats

airbyte-integrations/connectors/source-github/source_github/schemas/workflow_runs.json
grubberr: all referenced sub-schemas located here `schemas/shared`

airbyte-integrations/connectors/source-github/source_github/schemas/issue_events.json
grubberr: Yes, I thought about it initially 

I have added `"performed_via_github_app"` sub-object to 3 streams: `comments`, `issue_events`, `issues`

But in all 3 streams, this sub-object is different here and there 
I just decided to inline this sub-object to all these 3 streams until we will know some new information.

erohmensing: Ah gotcha 👍🏻

airbyte-integrations/connectors/source-github/source_github/schemas/organizations.json
erohmensing: Sweet, happy to have caught something in the script :)

airbyte-integrations/connectors/source-github/source_github/schemas/workflow_runs.json
erohmensing: Aha, that's what I was missing! Thanks

airbyte-integrations/connectors/source-github/acceptance-test-config.yml
```diff
@@ -15,7 +15,11 @@ acceptance_tests:
   discovery:
     tests:
     - config_path: "secrets/config.json"
+      backward_compatibility_tests_config:
+        disable_for_version: "0.4.6"
```
erohmensing, line 19: @grubberr This is because of removing "repository" from the `teams` stream, which was never outputting data, is that correct? Anything else that flagged as non-compatible?

grubberr: @erohmensing  I have wrote all possible compatibility problems in PR descriptions

grubberr: as for me we need to improve compatibility tests some of them are definitely false positive

airbyte-integrations/connectors/source-github/source_github/schemas/comments.json
```diff
@@ -39,6 +39,64 @@
...
+        },
+        "html_url": {
+          "type": ["null", "string"]
+        },
+        "created_at": {
+          "type": ["null", "string"]
+        },
+        "updated_at": {
+          "type": ["null", "string"]
+        },
```
evantahler, line 78: These are probably not strings, but "type": "string" + "format": "date-time"

airbyte-integrations/connectors/source-github/source_github/schemas/issue_events.json
```diff
@@ -73,6 +90,244 @@
...
+        },
+        "html_url": {
+          "type": ["null", "string"]
+        },
+        "created_at": {
+          "type": ["null", "string"]
+        },
+        "updated_at": {
+          "type": ["null", "string"]
+        },
```
evantahler, line 126: Same - type": "string" + "format": "date-time"

airbyte-integrations/connectors/source-github/source_github/schemas/issues.json
```diff
@@ -164,9 +165,11 @@
         },
         "patch_url": {
           "type": ["null", "string"]
+        },
+        "merged_at": {
+          "type": ["null", "string"]
         }
```
evantahler, line 171: Same - type": "string" + "format": "date-time"

airbyte-integrations/connectors/source-github/source_github/schemas/issues.json
```diff
@@ -182,6 +185,94 @@
...
+        },
+        "html_url": {
+          "type": ["null", "string"]
+        },
+        "created_at": {
+          "type": ["null", "string"]
+        },
+        "updated_at": {
+          "type": ["null", "string"]
+        },
```
evantahler, line 230: Same - type": "string" + "format": "date-time"

Also, they can't ever be null

evantahler (changes requested):

evantahler:
I've been trying to fix up these date-times in https://github.com/airbytehq/airbyte/pull/22832, but I keep getting stuck.  If you want to tackle the removal of nulls here for created-at/updated-at as well, that would be great

airbyte-integrations/connectors/source-github/source_github/schemas/issues.json
grubberr: it can be incompatible schema change database `string` type -> `datetime` time
we need to decide when it's good time to do this change

grubberr:
@evantahler
1. I have improve date-time fields making them non-null when this fields are really non-null
2. I have added fields from DataDog

airbyte-integrations/connectors/source-github/source_github/schemas/workflow_jobs.json
erohmensing: > in stream schema "format": "date-time" is used for normalization to create on SQL databases DATETIME column (not STRING column)

Aha, that makes sense. I'm not super familiar with normalization. Thanks!

airbyte-integrations/connectors/source-github/acceptance-test-config.yml
erohmensing: I think, looking at the error message, that it's only[ this specific one](https://github.com/airbytehq/airbyte/actions/runs/4546946491/jobs/8016229228#step:11:1595) that is incompatible, though I agree the logging of the diff is quite confusing. It would make sense as to why the backwards compatibility test didn't start failing until after you removed that repository param, despite having added a lot of params.

erohmensing:
Have we tested this with configured catalogs from users?

grubberr:
@erohmensing 

I have tested such way:
1. New sync with previous version `0.4.7` - OK
2. Subsequent sync with this `dev` version but with old catalog (from step 1) - OK
3. Refresh source schema without Reset Data - FAIL - which is resonable because this new catalog is not compatible with catalog from version `0.4.7`
4. Reset Data - OK

erohmensing:
@grubberr this is interesting to me. So it sounds like when users get the new version, their syncs will continue to succeed, unless they try to update the source schema, at which time things will break unless they perform a full refresh? Is that something we expect to happen even if we only add new columns, or because we've also changed some types, or only because we removed one field?

 I'm trying to determine whether this is a breaking change (the user needs to do something in order to avoid things breaking), for context, and to see if we can avoid it being breaking at all (e.g. if its only the one thing that got removed, it might be worth keeping it in for backwards compatibility's sake)

grubberr:
@erohmensing 
Yes, it's a good topic to discuss.
Even from the CAT tests point of view this scheme change PR is breaking, 
as for me, nothing will break and all continue to work using old schama.

I am 90% sure we can say that this change is not breaking but I can underestimate some use cases and miss something. May we need to ask the whole team?

evantahler:
Any progress on this?  Let's get this in!

grubberr:
@evantahler @erohmensing 

> Any progress on this? Let's get this in!
I have fixed tests, to finish this PR I need to get approve(s)

Thank you

grubberr:
@evantahler @erohmensing can you please look on this for review ?

airbyte-integrations/connectors/source-github/source_github/schemas/issue_events.json
maxi297: Are we sure about that? I have compared the field schema in https://docs.github.com/en/enterprise-server@3.6/rest/issues/events, https://docs.github.com/en/enterprise-server@3.6/rest/issues/comments and https://docs.github.com/en/enterprise-server@3.6/rest/issues/issues and they are actually the same. I think this would be relevant to figure this out to avoid other schema changes and prevent loss of data for the customers

airbyte-integrations/connectors/source-github/source_github/schemas/issues.json
evantahler: I think this should be a "format": "date-time" - all the properties that end in `_at` are timestamps

evantahler (changes requested):

evantahler (approved):
Oops, I was looking at an old diff - things look good now!

airbyte-integrations/connectors/source-github/source_github/schemas/teams.json
```diff
@@ -5,9 +5,6 @@
     "organization": {
       "type": ["null", "string"]
     },
-    "repository": {
```
maxi297, line 8: Is removing a field a breaking change?

maxi297 (commented):
> a Item root['collaborators']['properties']['role_name'] added to dictionary.

This does not seem like a breaking change to me. Why is it mentioned in the `disable backward compatibility checks reasons` section? 

> b Type of root['assignees']['properties']['repository']['type'] changed from list to str and value changed from ['string'] to "string".
> c Type of root['deployments']['properties']['repository']['type'] changed from list to str and value changed from ['null', 'string'] to "string".
> d Type of root['project_cards']['properties']['project_id']['type'] changed from list to str and value changed from ['null', 'integer'] to "integer".
> e Type of root['deployments']['properties']['creator']['type'] changed from str to list and value changed from "object" to ['null', 'object'].

Removing the fact that an object is nullable does not seem like a breaking change to me but who from the destination team could validate that it does not cause any issue to ensure we are 100% sure instead of 90%? 

Also in airbyte-integrations/connectors/source-github/source_github/schemas/teams.json, the `repository` field is removed. Is this a breaking change? Usually for a REST API, it is but I could see it not being an issue for something like Airbyte.

Given those questions answered, I'll give my approval!

airbyte-integrations/connectors/source-github/source_github/schemas/teams.json
grubberr: According to CAT tests removing field - is not breaking change
I just checked, I have removed one field and run CAT tests - no warnings about breaking changes

airbyte-integrations/connectors/source-github/source_github/schemas/issue_events.json
grubberr: Actually, all of these PR changes are based on real data. We gathered data from our testing accounts, and then I added updates to the schema to conform to the data.

Here are some details:

Building a good schema is not an easy task.

1. incoming JSON objects from our testing accounts may not cover all possible variations.
2. Fully relying on documentation can also be dangerous. For example, I found some date-time fields that were marked as non-null in the documentation but were actually nullable.
2. GitHub API data has greatly increased over the last year. Even if we build a good schema initially, it could quickly become outdated.

RESUME:  building a good schema is a continuous process, and we shouldn't rely solely on documentation.

P. S. I can actually obtain the "performed_via_github_app" object from the documentation, and there's a high chance that this object is the same across all streams.

airbyte-integrations/connectors/source-github/source_github/schemas/comments.json
```diff
@@ -27,18 +27,78 @@
...
     "author_association": {
       "type": ["null", "string"]
+    },
+    "reactions": {
+      "$ref": "reactions.json"
+    },
+    "performed_via_github_app": {
+      "type": ["null", "object"],
+      "properties": {
+        "id": {
```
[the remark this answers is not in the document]
grubberr: 99% of all fields in all our schemas for all source connectors have `["null", "type"]`, even when it's non-null in reality.

We try to be more precise about non-nullability only for:
1. primary key field(s)
2. cursor field
3. All other fields were you 100% sure about non-nullability

For other cases, we usually add `["null", "type"]` just because it's safer.
For us, it's much cheaper to have NULL in the database than be killed trying to insert null in a non-nullable column.

That's a way we are not very strict for such cases and usually automatically add `["null", "type"]`

airbyte-integrations/connectors/source-github/source_github/schemas/workflow_runs.json
```diff
@@ -53,18 +59,22 @@
...
     "updated_at": {
-      "type": ["null", "string"],
+      "type": "string",
       "format": "date-time"
     },
     "run_attempt": {
       "type": ["null", "integer"]
     },
+    "referenced_workflows": {
+      "type": "array"
```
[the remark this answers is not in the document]
maxi297: There is an example and the schema [here](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28) if you think we can rely on the API doc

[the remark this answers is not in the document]
grubberr: I have added from documentation field "referenced_workflows" to stream `workflow_runs`

grubberr:
@maxi297 @erohmensing @evantahler 

The motivation for this PR.

I have created this PR to add missed fields to schemas to allow CAT tests to continue to work without errors
with `fail_on_extra_columns: true`.

All of these schema changes in this PR are based on real data!
I gathered data from our testing accounts, and then I added updates to the schema to conform to the data. But the final result can still be not 100% perfect. As soon as we get more data we would be able to continue to extend schemas.

Building a good schema is not an easy task:

- Incoming JSON objects from our testing accounts may not cover all possible variations.

- Fully relying on documentation can also be dangerous. For example, I was founding some date-time fields that were marked as non-null in the documentation but were actually nullable.

- GitHub API schemas have greatly increased over the last year. Even if we build a good schema initially, it could quickly become outdated.

RESUME: building a good schema is a continuous process, and we shouldn't rely solely on documentation.

maxi297 (approved):
I've resolved my comments and therefore I'm approving this PR

grubberr:
@erohmensing can you please look? because I cannot merge this PR without your approval

erohmensing (approved):
approving to unblock "changes requested"

PR #24545, opened 2023-03-25
Add dev job to regenerate catalog from metadata and view diff

### LINKED ISSUE #24516: Add a job that parses metadata files into their catalogs
1. Take existing metadata files in dagster
2. Recompile them into the catalogs
3. diff the new catalogs with the old to ensure they match

bnchrch:
## What
image

closes #24516 

We need a way to see if the catalog that is generated by the new system matches the old system.

## How
1. Add the code to generate a catalog from metadata files
2. Adds a few dev assets that diff the old catalog and the new catalog and display the results


airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_assets.py
```diff
@@ -1,15 +1,112 @@
...
+
+
+def construct_catalog_from_metadata(catalog_derived_metadata_definitions: List[dict], catalog_name: str) -> dict:
+    # get only definitions with data.catalogs.cloud.enabled = true
+    metadata_definitions = [
+        metadata for metadata in catalog_derived_metadata_definitions if metadata["data"]["catalogs"][catalog_name]["enabled"]
+    ]
+
+    metadata_sources = [metadata for metadata in metadata_definitions if metadata["data"]["connectorType"] == "source"]
+    metadata_destinations = [metadata for metadata in metadata_definitions if metadata["data"]["connectorType"] == "destination"]
```
alafanechere, line 70: nit to avoid an extra iteration.
```suggestion
    metadata_sources = [metadata for metadata in metadata_definitions if metadata["data"]["connectorType"] == "source" and metadata["data"]["catalogs"][catalog_name]["enabled"]
    metadata_destinations = [metadata for metadata in metadata_definitions if metadata["data"]["connectorType"] == "destination" and metadata["data"]["catalogs"][catalog_name]["enabled"]]
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/dev_assets.py
```diff
@@ -126,6 +132,42 @@
     "a7bcc9d8-13b3-4e49-b80d-d020b90045e3": {"connectionType": "file"},
 }
 
+# HELPERS
+
+
+def key_catalog_entries(catalog_dict: List[dict]) -> dict:
```
alafanechere, line 138:
```suggestion
def key_catalog_entries(catalog: dict) -> dict:
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/dev_assets.py
```diff
@@ -126,6 +132,42 @@
...
+    catalog_dict_keyed = catalog_dict.copy()
+    for connector_type, id_field in [["sources", "sourceDefinitionId"], ["destinations", "destinationDefinitionId"]]:
```
alafanechere, line 140:
```suggestion
    for connector_type, id_field in [("sources", "sourceDefinitionId"), ("destinations", "destinationDefinitionId")]:
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_assets.py
```diff
@@ -1,15 +1,112 @@
...
 import json
 from dagster import asset, OpExecutionContext
+from typing import List
 
 from ..utils.dagster_helpers import OutputDataFrame
 
 
 GROUP_NAME = "catalog"
 
+# HELPERS
```
alafanechere, line 11: What do you think of keeping helpers in a separate module / package ?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/dev_assets.py
```diff
@@ -164,3 +206,25 @@ def persist_metadata_definitions(context, overrode_metadata_definitions):
...
+
+@asset(group_name=GROUP_NAME)
+def oss_catalog_diff(oss_catalog_from_metadata: dict, latest_oss_catalog_dict: dict):
+    """
+    Compares the OSS catalog from the metadata with the latest OSS catalog.
+    """
+    diff = diff_catalogs(latest_oss_catalog_dict, oss_catalog_from_metadata)
+    diff_df = pd.DataFrame.from_dict(diff)
+
+    return OutputDataFrame(diff_df)
```
alafanechere, line 230: Could you share why these assets return dataframe and not pure DeepDiff objects?

alafanechere (commented):
I don't know to what extent this code is temporary but I think the helpers you defined would greatly benefits from more typed models instead of pure dict interactions. I would find it more readable and maintainable to have a Metadata model with methods to transform it to desired intermediate models  or dict.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/dev_assets.py
bnchrch: Good question: I have it return a dataframe to take advantage of the `to_markdown` method for previews

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_assets.py
```diff
@@ -1,15 +1,112 @@
...
+
+def apply_overrides_from_catalog(metadata_data: dict, override_catalog: str) -> dict:
+    catalog_fields = metadata_data["catalogs"][override_catalog]
+    del catalog_fields["enabled"]
+    for field, value in catalog_fields.items():
+        metadata_data[field] = value
+    return metadata_data
+
+
+def metadata_to_catalog_entry(metadata_definition: dict, connector_type: str, override_catalog: str) -> dict:
```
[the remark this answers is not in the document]
bnchrch: I was definitely lazy with these as its meant to be short lived.

However lets move this to a dataclass like you suggest. The script is short lived but the translation from metadata to catalog is not

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_assets.py
```diff
@@ -1,15 +1,112 @@
...
 GROUP_NAME = "catalog"
 
+# HELPERS
+
+
+def apply_overrides_from_catalog(metadata_data: dict, override_catalog: str) -> dict:
+    catalog_fields = metadata_data["catalogs"][override_catalog]
+    del catalog_fields["enabled"]
+    for field, value in catalog_fields.items():
+        metadata_data[field] = value
```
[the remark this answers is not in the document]
bnchrch: Good call! And I appreciate all the python nits. Still getting back into the language!

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_assets.py
```diff
@@ -1,15 +1,112 @@
...
 
 
 GROUP_NAME = "catalog"
 
+# HELPERS
+
+
+def apply_overrides_from_catalog(metadata_data: dict, override_catalog: str) -> dict:
+    catalog_fields = metadata_data["catalogs"][override_catalog]
+    del catalog_fields["enabled"]
```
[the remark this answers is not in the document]
bnchrch: Great call out. To be honest I wasnt happy I was using a mutating fn here either. Will change.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_assets.py
bnchrch: Hmm good question, and generally im totally for helper/util files/modules/classes.

We already have the helper file pattern in this orchestration project as proof 😅 

However in my mind I see helper functions as filling two possible roles
1. Providing a piece of reusable code
2. Providing a way to self document code and break up the cognitive load when reading a function that uses the helper. (code should read like a book principle) 

If a helper functions only purpose is item #2 (self documentation) but we anticipate it will only ever be used in one circumstance by one other function/class/file, then I typically think it should be collocated. in the same file.

So in this instance I would opt to leave these specific helpers here. 

Does that change your mind any?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_assets.py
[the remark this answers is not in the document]
bnchrch: Fair call! Will change

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/dev_assets.py
alafanechere: Would it make sense to provide a final asset only exposing the dataframe and depending on the oss_catalog_diff which returns the diff object? Might be overkill at the moment.

bnchrch: It may slightly violate the "only write the code when you need to" rule 😛 

but its a good idea and what dagster is meant for so lets do it!

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog_assets.py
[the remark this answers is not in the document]
bnchrch: @alafanechere I added a PartialMetadataDefinition. Let me know what you think about the pattern.

Im going to hold off on adding a catalog definition model, as I wait for a decision on the airbyte-types repo at the end of this week.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+GROUP_NAME = "catalog"
+
+# HELPERS
+
+
+def apply_overrides_from_catalog(metadata_data: dict, override_catalog_key: str) -> dict:
+    # Make a deep copy of the input dictionary to avoid side effects
+    metadata_data_copy = copy.deepcopy(metadata_data)
```
alafanechere, line 19: Nit: as you reuse this *deepcopy* pattern you could implement a [decorator](https://realpython.com/primer-on-python-decorators/) to always deepcopy the input dict

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog.py
```diff
@@ -0,0 +1,161 @@
...
+
+def is_metadata_catalog_enabled(metadata_definition: dict, catalog_name: str) -> bool:
+    return metadata_definition["data"]["catalogs"][catalog_name]["enabled"]
+
+
+def is_metadata_connector_type(metadata_definition: dict, connector_type: str) -> bool:
+    return metadata_definition["data"]["connectorType"] == connector_type
+
+
+def construct_catalog_from_metadata(catalog_derived_metadata_definitions: List[dict], catalog_name: str) -> dict:
```
alafanechere, line 84:
```suggestion
def construct_catalog_from_metadata(catalog_derived_metadata_definitions: List[PartialMetadataDefinition], catalog_name: str) -> dict:
```

I'm not sure if you're passing a list of dict or PartialMetadataDefinition because `cloud_catalog_from_metadata` passes its `catalog_derived_metadata_definitions` input list composed of PartialMetadataDefinition.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog.py
```diff
@@ -0,0 +1,161 @@
...
+
+    TODO (ben): This asset should be updated to use the GCS metadata definitions once available.
+    """
+    return construct_catalog_from_metadata(catalog_derived_metadata_definitions, "oss")
+
+
+@asset(group_name=GROUP_NAME)
+def cloud_sources_dataframe(latest_cloud_catalog_dict: dict) -> OutputDataFrame:
+    sources = latest_cloud_catalog_dict["sources"]
+    return output_dataframe(pd.DataFrame(sources))
```
alafanechere, line 127: nit: I think a custom decorator can be created and use  as you consistently return the same types in the same way on each assets below.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog.py
```diff
@@ -0,0 +1,161 @@
+import pandas as pd
+import json
+import copy
+from typing import List
+
+from dagster import asset, OpExecutionContext
+
+from ..utils.dagster_helpers import OutputDataFrame, output_dataframe
```
alafanechere, line 8: nit 
```suggestion
from metadata_service.orchestrator.utils.dagster_helpers import OutputDataFrame, output_dataframe
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata.py
```diff
@@ -127,14 +130,14 @@ def build_metadata(merged_df):
         catalogs = compute_catalog_overrides(merged_df)
         metadata["data"]["catalogs"] = catalogs
 
-        return metadata
+        return PartialMetadataDefinition.construct(**metadata)
```
alafanechere, line 133:
```suggestion
        return PartialMetadataDefinition(**metadata)
```
Just wondering why we're not building the model directly

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata.py
```diff
@@ -92,13 +94,14 @@ def compute_catalog_overrides(merged_df):
     return catalogs
 
 
+# todo change this return value
 def merge_into_metadata_definitions(id_field, connector_type, oss_connector_df, cloud_connector_df) -> pd.Series:
```
alafanechere, line 98: I think a docstring would be helpful here.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata.py
```diff
@@ -92,13 +94,14 @@ def compute_catalog_overrides(merged_df):
...
     merged_connectors = pd.merge(
         oss_connector_df, cloud_connector_df, on=id_field, how="outer", suffixes=(OSS_SUFFIX, CLOUD_SUFFIX), indicator=True
     )
     sanitized_connectors = merged_connectors.where(pd.notnull(merged_connectors), None)
```
alafanechere, line 102:
```suggestion
sanitized_connectors = merged_connectors.fillna(None)
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata.py
```diff
@@ -92,13 +94,14 @@ def compute_catalog_overrides(merged_df):
...
 
-    def build_metadata(merged_df):
+    def build_metadata(merged_df: dict) -> PartialMetadataDefinition:
```
alafanechere, line 104:
```suggestion
    def build_metadata(connector_catalog_entry: dict) -> PartialMetadataDefinition:
```

I found it hard to understand what was happening in this function until I realized you're manipulating a dict and not a dataframe will all the catalog entry.
Please add a docstring to explain what the format of the expected input and where it comes from.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata.py
alafanechere: Oh I think its because you explicitly want to delay the validation in a separate step.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata.py
```diff
@@ -146,15 +149,15 @@ def validate_metadata(metadata: object) -> tuple[bool, str]:
 
 
 @asset(group_name=GROUP_NAME)
-def valid_metadata_list(overrode_metadata_definitions):
+def valid_metadata_list(overrode_metadata_definitions: List[PartialMetadataDefinition]) -> OutputDataFrame:
```
alafanechere, line 152: I would expect this asset to return a list of valid `ConnectorMetadataDefinitionV1`.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/models/metadata.py
```diff
@@ -0,0 +1,36 @@
+from metadata_service.models.generated.ConnectorMetadataDefinitionV1 import ConnectorMetadataDefinitionV1
+from pydantic import ValidationError
+
+from pydantic import ValidationError
+from typing import Tuple, Any, Optional
+
+
+class PydanticDelayValidationMixin:
```
alafanechere, line 8: I find it's a breaking a bit the pydantic purpose if you delay to model validation.
I found an interesting suggestion [here](https://github.com/pydantic/pydantic/issues/668#issuecomment-512525178). Let me know what you think.

alafanechere (commented):
Thank you for the use of a pydantic model 🙏 .
While reading the code I found it hard to figure what types you were actually handling in the functions due to some typehints / variable naming inconsistencies.
I would suggest to only use dataframe to merge oss and cloud catalog. And then serialize dataframe to tuple of `(ExpectedPydanticModel`, list_of_dict)` (a bit like what's suggested [here](https://github.com/pydantic/pydantic/issues/668#issuecomment-512525178)). And then reconvert to dataframe for the "view", the edge of your DAG. But maybe there's a dagster constraint I did not grasp.
I would be fine keep dataframe all along the way if you were always manipulating dataframe on which you apply pandas functions. You could also perform dataframe validation with pydantic with [this kind of approach.](https://www.inwt-statistics.com/read-blog/pandas-dataframe-validation-with-pydantic-part-2.html)

I'm fine approving this PR if you want to mainline it quickly but I'd love that we figure together a logic of when to use dict / list of dict / list of pydantic models / dataframe.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+# ASSETS
+
+
+@asset(group_name=GROUP_NAME)
+def cloud_catalog_from_metadata(catalog_derived_metadata_definitions: List[PartialMetadataDefinition]) -> dict:
+    """
+    This asset is used to generate the cloud catalog from the metadata definitions.
+
```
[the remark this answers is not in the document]
bnchrch: Thank you! Installing

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog.py
bnchrch: I think in this case, because we're already under the weight of dagster decorators I would like to keep whats happening for each asset much more explicit. 

My worry is two layers of decorators, is two layers of indirection, which can make debugging and understanding a little more difficult.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata.py
bnchrch: Yup exactly that, these aren't "valid" until overrides are applied

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata.py
bnchrch: I can rename this!

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/models/metadata.py
bnchrch: @alafanechere Im glad you noticed this!

Its definately an awkward duck because we
1. Want strong type guarantees of what is metadata and what is not
2. Want these types to originate (be generated) from a jsonschema file to support java and python types
3. Want partial versions of these types to allow us to delay validation while we apply overrides

I went through a whole bunch of libraries / discussions on it.

Generally
1. Pydantic doesnt support partials out of the box
2. Pydantic doesnt seem to want to support this use case
3. The `pydantic-partial` library doesn't support nested validations 😢 

Ok so looking at the solution you linked. I like it, but I think theres a few small issues
1. Out of the box it doesn't give us access to the subscriptable methods __getitem__ and __setitem__
2. I actually tried to get this working but unfortunately theres its also not as easy as adding the subscriptable methods and we need to call some more internal apis to properly update the internal raw data dict (see screenshot below)
3. When looking at implementing this there were enough bugs I had to patch over that make me nervous to take at face value

In general I think im at the point where im happy to stay with the current implementation if it works. So long as you are 😅 

image


### Appendix
https://github.com/team23/pydantic-partial
https://github.com/pydantic/pydantic/issues/1673
https://github.com/tiangolo/fastapi/issues/125

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/metadata.py
bnchrch: Had to go a slightly different way but thanks for the prompt

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/models/metadata.py
alafanechere: Fair enough, let's keep current implementation!

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/catalog.py
alafanechere: Fair enough 😄

alafanechere (approved):
Thank you for your changes! I'm ok with the tradeoffs you made. I think you still want to finish the tests as you commented out some lines in these files.

airbyte-ci/connectors/metadata_service/orchestrator/tests/test_debug.py
```diff
@@ -56,9 +56,9 @@ def debug_catalog_projection():
...
 
     catalog_derived_metadata_definitions(context, cloud_sources_df, cloud_destinations_df, oss_sources_df, oss_destinations_df).value
-    # valid_metadata_list_df = valid_metadata_list(metadata_definitions_df).value
+    # valid_metadata_report_dataframe_df = valid_metadata_report_dataframe(metadata_definitions_df).value
 
-    # all_sources_df = all_sources_dataframe(cloud_sources_df, oss_sources_df, github_connector_folders_list, valid_metadata_list_df)
+    # all_sources_df = all_sources_dataframe(cloud_sources_df, oss_sources_df, github_connector_folders_list, valid_metadata_report_dataframe_df)
     # all_destinations_df = all_destinations_dataframe(cloud_destinations_df, oss_destinations_df)
 
     # connector_catalog_location_html(context, all_sources_df, all_destinations_df)
```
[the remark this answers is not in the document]
bnchrch: Ahh yes.

Sorry this is a little bit of jank introduced by Dagster.

See this comment at the top of the file:
https://github.com/airbytehq/airbyte/pull/24545/files/823e1f59eafe45dbabbbbabbe3316b2a0ee3fffd#diff-350aa46358bd8f37293007092536a25fb77f278e282f0b18ed61ab5d49053839L23-L28

airbyte-ci/connectors/metadata_service/orchestrator/tests/test_catalog.py
```diff
@@ -44,9 +44,11 @@ def test_merged_catalog_dataframes(oss_catalog_dict, cloud_catalog_dict):
...
-    all_sources_df = all_sources_dataframe(cloud_sources_df, oss_sources_df, github_connector_folders, valid_metadata_list, cached_specs)
+    all_sources_df = all_sources_dataframe(
+        cloud_sources_df, oss_sources_df, github_connector_folders, valid_metadata_report_dataframe, cached_specs
+    )
     all_destinations_df = all_destinations_dataframe(
-        cloud_destinations_df, oss_destinations_df, github_connector_folders, valid_metadata_list, cached_specs
+        cloud_destinations_df, oss_destinations_df, github_connector_folders, valid_metadata_report_dataframe, cached_specs
     )
 
     # assert that all_sources_df has a entry for each sourceDefinitionId in the cloud catalog and oss catalog
```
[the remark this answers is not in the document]
bnchrch: @alafanechere  its actually a comment!

I can see the mix up as it starts with `assert`

PR #24569, opened 2023-03-27
connectors-ci: better handle unexpected failures

alafanechere:
## What
Fix the following pipeline behaviors:
* Acceptance tests and integration tests were skipped when no unit test is defined. We want to skip acceptance tests and integration tests only when unit tests are failing, not when they are not declared.
* When a failure outside of tests raised an exception it was caught by the test context but did not lead to explicit failure of the test run and creation of a test report. 

## How
* Only skip acceptance and integration tests when unit tests fail
* Wrap the `run` method of the abstract Step class: `run` now handles dagger Query error and calls the `_run` method of the step (which is overriden by child classes like AcceptanceTests, CodeFormatChecks etc.)
* Declare a report to not be successful when no step rans, add failed and skipped steps to the test report.


tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
```diff
@@ -150,32 +150,32 @@ async def __aexit__(self, exception_type, exception_value, traceback) -> bool:
...
+        self.test_report.print()
+        self.logger.info(self.test_report.to_json())
+        local_test_reports_path_root = "tools/ci_connector_ops/test_reports/"
+        connector_name = self.test_report.connector_test_context.connector.technical_name
+        connector_version = self.test_report.connector_test_context.connector.version
+        git_revision = self.test_report.connector_test_context.git_revision
+        git_branch = self.test_report.connector_test_context.git_branch.replace("/", "_")
+        suffix = f"{connector_name}/{git_branch}/{connector_version}/{git_revision}.json"
+        local_report_path = Path(local_test_reports_path_root + suffix)
+        await local_report_path.parents[0].mkdir(parents=True, exist_ok=True)
```
bnchrch, line 170: Nit: I think this function could benefit from a docstring and a few comments :)

bnchrch (approved):
LGTM!

PR #24600, opened 2023-03-27
Source Facebook Marketing: override SDK Cursor

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1657

## How
Monkey patch FB SDK Cursor class, please see more details in docstrings

for context: [this](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-facebook-marketing/source_facebook_marketing/streams/common.py#L36) is the place were we're decreasing the limit in that I'm referring to in the docstrings


airbyte-integrations/connectors/source-facebook-marketing/integration_tests/configured_catalog.json
```diff
@@ -1,280 +0,0 @@
-{
```
lazebnyi, line 1: Why Why was the configured catalog deleted?

lazebnyi (approved):
LGTM! Small question below.

PR #24405, opened 2023-03-23
1705 p0 revert regex

rodireich:
## What
Revert a change previously merged in #23904 
as customers [seeing cases](https://github.com/airbytehq/oncall/issues/1705) hostname validation failed for seemingly valid hosts.
## How
revert original commit in 0.4.56

PR #24615, opened 2023-03-27
Destination Snowflake: Verify hostname via regex in spec

### LINKED ISSUE #23172: Snowflake destination should verify hostname
In this OC issue a user typed in an invalid hostname (included spaces), which caused oAuth to fail: https://github.com/airbytehq/oncall/issues/1259. If we want to prevent this in the future, we should add hostname validation to the UI. This is already possible in specs by adding pattern validation. Here is an example: 
https://github.com/airbytehq/airbyte/blob/b6c6b7affb67c29d16efd203b7a18c019c075c06/airbyte-integrations/connectors/source-google-pagespeed-insights/source_google_pagespeed_insights/spec.yaml#L26

### LINKED ISSUE #24409: Fix Snowflake host pattern validation
We added host name validation with a regular expression, but found that there is more variability in hostnames than we expected: https://github.com/airbytehq/oncall/issues/1705

edgao:
## What
Revert the revert (https://github.com/airbytehq/airbyte/pull/24405) of the hostname check, and:
* add an optional `https://` prefix
* add an optional fifth subdomain

Relevant diff is in https://github.com/airbytehq/airbyte/pull/24615/commits/9b51107c7c5a7737c0d4120a3e810c83684d04ec

closes https://github.com/airbytehq/airbyte/issues/23172; closes https://github.com/airbytehq/airbyte/issues/24409

## How
Fetched all our hosts:

```sql
select distinct actor.configuration ->> 'host' from connection
inner join jobs on jobs.scope :: uuid = connection.id
inner join actor on actor.id = connection.destination_id
where jobs.status = 'succeeded' and actor.actor_definition_id = '424892c4-daac-4491-b35d-c6688ba547ba'
```

And verified that the new regex matches all of them: (this query returns no results, i.e. all hostnames matched the regex)
```sql
select distinct actor.configuration ->> 'host' as host from connection
inner join jobs on jobs.scope :: uuid = connection.id
inner join actor on actor.id = connection.destination_id
where jobs.status = 'succeeded' and actor.actor_definition_id = '424892c4-daac-4491-b35d-c6688ba547ba'
and actor.configuration ->> 'host' !~ replace('^(http(s)?:\\/\\/)?([^./?#]+\\.)?([^./?#]+\\.)?([^./?#]+\\.)?([^./?#]+\\.snowflakecomputing\\.com)$', '\\', '\')
order by host
```

(the `replace` call is because postgres treats `'\\'` as literally two backslashes)

Also added/modified some tests based on Snowflake documentation.

## Recommended reading order
1. spec.json
2. SnowflakeDestinationTest.java
3. dockerfile, snowflake.md

## 🚨 User Impact 🚨
No breaking change expected. At the very least, this won't break any existing connections in our DB.

This is patch bump because it just reduces the time it takes for us to tell users that their hostname is invalid. Previously, that would fail in `check`; now the UI can validate it while the user types.


airbyte-config/init/src/main/resources/seed/destination_specs.yaml
edgao, line 1: reviewers, please ignore this file - it's autogenerated and will be overwritten by `/publish`

airbyte-config/init/src/main/resources/seed/destination_definitions.yaml
edgao, line 1: ignore this file also.

docs/integrations/destinations/snowflake.md
```diff
@@ -287,8 +287,8 @@ Otherwise, make sure to grant the role the required permissions in the desired n
 
 | Version | Date       | Pull Request                                               | Subject                                                                                                                                             |
 |:--------|:-----------|:-----------------------------------------------------------|:----------------------------------------------------------------------------------------------------------------------------------------------------|
-| 0.4.57  | 2023-03-23 | [#23904](https://github.com/airbytehq/airbyte/pull/24405)  | Revert host validation change                                                                                                                       |
-| 0.4.56  | 2023-03-22 | [#23904](https://github.com/airbytehq/airbyte/pull/23904)  | Added host validation by pattern on UI                                                                                                              |
+| 0.4.58  | 2023-03-27 | [#24615](https://github.com/airbytehq/airbyte/pull/24615)  | Added host validation by pattern on UI                                                                                                              |
```
ryankfu, line 290: minor nit: maybe instead of `added` to have this as `fixed` since the two changelog comments are the same between `0.4.56` and `0.4.48`

ryankfu (approved):
LGTM

docs/integrations/destinations/snowflake.md
edgao: good point, done

PR #24622, opened 2023-03-27
🎉 Postgres Source : Allow streams not in CDC publication to be synced in Full-refresh mode

### LINKED ISSUE #24611: Postgres should not restrict discovered tables to those present in publication
Currently, Postgres only surfaces tables that are present in the publication while running in CDC mode. This is to prevent the user from performing an incremental sync with The side effect of this is that users have to set up a different connection to sync streams not in the current publication.

Was discussing this with @subodh1810 and we should add functionality to show tables not in the publication as well. However, incremental sync should be disabled for such tables

Previous discussion : https://github.com/airbytehq/airbyte/issues/13207

akashkulk:
Closes https://github.com/airbytehq/airbyte/issues/24611

Changes source-postgres to discover streams that are not in the CDC publication and allows to sync those streams in full-refresh mode


airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -113,6 +112,8 @@ public class PostgresSource extends AbstractJdbcSource<PostgresType> implements
   public static final String MODE = "mode";
 
   private List<String> schemas;
+
+  private Set<AirbyteStreamNameNamespacePair> publicizedTablesInCdc;
```
subodh1810, line 116: I dont like the idea of having a global variable and hoping/assuming it will get initialised cause a chain of methods will be executed before we use it. Any change in the execution steps will lead to a potential bug

subodh1810 (changes requested):

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
akashkulk: Unfortunately, we already have that with `schemas()`

I think the real issue is that we have a chain of methods where you can't predict what order they'd run in. Ideally, it should be clear what the ordering and so eventually we should gut a lot of the abstract methods deeper in the call stack. 

For now, I've moved the initialization of this global variable in `createDatabase()`. In general, this is part of the initialization phase and is the first thing that runs in every protocol method. So, any code that refers to this in check/discover/read methods should have this variable initialized

subodh1810 (approved):
Honestly I still dont like the way we are doing this but I understand that there is no better way! I think the dependency on the `database` object for all these operations is a big pain. We need to sort it out

akashkulk:

akashkulk:

akashkulk:
/publish connector=connectors/source-alloydb-strict-encrypt run-tests=false

akashkulk:

PR #24649, opened 2023-03-29
Add ci/cd support for metadata unit tests with Dagger

bnchrch:
## What
We need to run our unit tests for the metadata library

## How
1. This creates and uses a new dagger pipeline to accomplish that.
2. Updates the dagger command structure to have one top level command called `airbyte-ci-pipelines`
3. Refactors the context to be slightly more generic


.github/workflows/metadata_service_tests_dagger.yml
```diff
@@ -0,0 +1,49 @@
...
+        if: github.event_name == 'workflow_dispatch'
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: extract_branch
+      - name: Install Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: Install pipeline package
+        run: pip install ./tools/ci_connector_ops\[pipelines]\
+      - name: Run airbyte-ci-pipeline metadata-service test-metadata-service-lib [PULL REQUESTS]
```
bnchrch, line 32: @alafanechere I opted not to have a workflow dispatch, good idea or bad idea?

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -185,15 +189,48 @@ async def with_ci_credentials(context: ConnectorTestContext, gsm_secret: Secret)
...
+
+    Returns:
+        Container: A python environment with poetry installed.
+    """
+    python_base_environment: Container = context.dagger_client.container().from_("python:3-alpine")
+    python_with_git = python_base_environment.with_exec(["apk", "add", "gcc", "libffi-dev", "musl-dev", "git"])
+    python_with_poetry = python_with_git.with_exec(INSTALL_POETRY_PACKAGE_CMD)
+
+    poetry_cache: CacheVolume = context.dagger_client.cache_volume("poetry_cache")
+    poetry_with_cache = python_with_poetry.with_mounted_cache("/root/.cache/pypoetry", poetry_cache, sharing=CacheSharingMode.PRIVATE)
```
bnchrch, line 220: @alafanechere I opted not to share cache for this, good idea or bad idea?

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/metadata_service.py
```diff
@@ -0,0 +1,66 @@
...
+
+async def run_metadata_lib_test_pipeline(metadata_pipeline_context):
+    async with dagger.Connection(DAGGER_CONFIG) as dagger_client:
+        metadata_pipeline_context.dagger_client = dagger_client.pipeline(metadata_pipeline_context.pipeline_name)
+        async with metadata_pipeline_context:
+            metadata_lib_module = environments.with_poetry_module(metadata_pipeline_context, METADATA_LIB_MODULE_PATH)
+            exit_code = await with_exit_code(metadata_lib_module.with_exec(["poetry", "run", "pytest"]))
+
+            # Raise an exception if the exit code is not 0
+            if exit_code != 0:
```
bnchrch, line 30: @alafanechere I didnt like that this was where I ended up. Do you have any ideas as to how we can make the pipeline context own the handling of the status code?

For the record the reason I raise is to take advantage of the git error status reporting in `__aexit__` inside the pipeline class

.github/actions/run-dagger-pipeline/action.yml
```diff
@@ -0,0 +1,55 @@
+name: "Run Dagger pipeline"
```
bnchrch, line 1: @alafanechere What do you think about the dagger action? If your into it I can update the other dagger workflows

alafanechere: I think it's a nice initiative. 
I think @cpdeethree would like to eventually have a single workflow which will run on every PR.
The logic of what pipeline runs would be inferred by the git diff in the entrypoint.
I we go down this way declaring an action could be overkill, but I like it for our current way of doing things.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
alafanechere: I you want the cache to by reused throughout builds you should use LOCKED or SHARED.
If concurrent access is problematic you should chose LOCKED, if it's not problematic use SHARED as it will allow concurrent pipeline to access it without waiting.

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
```diff
@@ -80,7 +80,7 @@ The default behavior of the CLI is to run in a local context.
 You can tell the CLI that it is running in a CI context with the following flag:
 
 ```bash
-connectors-ci --is-ci
+airbyte-ci-pipeline --is-ci connectors-ci
 ```
```
alafanechere, line 85: Could you please document your new command to test metadata here?

tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
```diff
@@ -43,20 +130,20 @@ def __init__(
         pipeline_start_timestamp: Optional[int] = None,
         ci_context: Optional[str] = None,
     ):
+        pipeline_name = f"CI test for {connector.technical_name}"
+        super().__init__(
```
alafanechere, line 134: As the parent init is responsible for updating commit status check can we call super at the end of the child context constructor?

tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
```diff
@@ -7,26 +7,113 @@
...
+    def dagger_client(self, dagger_client: Client):
+        self._dagger_client = dagger_client
 
+    @property
+    def is_ci(self):
+        return self.is_local is False
+
+    @property
+    def is_pr(self):
+        return self.ci_context == "pull_request"
```
alafanechere, line 71: nit: not related to your PR but I think we should use an enum for this.

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/metadata_service.py
alafanechere: I would suggest you implement a `Step` class like `MetadataLibTest(Step)` and make its `_run` return  `self.get_step_result(metadata_lib_module)` .
Like what I've done for [QaChecks](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/checks.py#L12)

alafanechere: I would also suggest you make use of [ConnectorTestReport](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/bases.py#L118) (or a Metadata version of it). This will help normalize the way we report success or failure to users.

tools/ci_connector_ops/ci_connector_ops/pipelines/github.py
```diff
@@ -19,6 +19,8 @@
 def update_commit_status_check(
     sha: str, state: str, target_url: str, description: str, context: str, should_send=True, logger: Logger = None
 ):
+    logger.info(f"Attempting to create {state} status for commit {sha} on Github in {context} context.")
```
alafanechere, line 22: This call will fail if logger is None, which is the default value.

tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
```diff
@@ -7,26 +7,113 @@
...
-    SUCCESSFUL = {"github_state": "success", "description": "All tests ran successfully."}
-    FAILURE = {"github_state": "failure", "description": "Test failed."}
+    INITIALIZED = {"github_state": "pending", "description": "Pipelines are being initialized..."}
+    RUNNING = {"github_state": "pending", "description": "Pipelines are running..."}
+    ERROR = {"github_state": "error", "description": "Something went wrong while running the Pipelines."}
+    SUCCESSFUL = {"github_state": "success", "description": "All Pipelines ran successfully."}
+    FAILURE = {"github_state": "failure", "description": "Pipeline failed."}
+
+
+class PipelineContext(ABC):
```
alafanechere, line 31: This is an abstract class but you created instance of it in `metadata_service`. 

I suggest you create a `MetadataPipelineContext(PipelineContext)`. I would also declare `__aenter__` and `__aexit__` as abstract methods, and you make your   `MetadataPipelineContext` implement these methods.
I we eventually spot a pattern on a third child class we could consider unabstracting these methods and implement a sensible default one.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -185,15 +189,48 @@ async def with_ci_credentials(context: ConnectorTestContext, gsm_secret: Secret)
...
+    """Installs poetry in a python environment.
+
+    Args:
+        context (PipelineContext): The current test context, providing the repository directory from which the ci_credentials sources will be pulled.
+
+    Returns:
+        Container: A python environment with poetry installed.
+    """
+    python_base_environment: Container = context.dagger_client.container().from_("python:3-alpine")
+    python_with_git = python_base_environment.with_exec(["apk", "add", "gcc", "libffi-dev", "musl-dev", "git"])
```
alafanechere, line 216: `with_ci_connector_ops` also does the same, I suggest creating a new `with_python_and_git` function.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -26,16 +26,20 @@
 
 INSTALL_LOCAL_REQUIREMENTS_CMD = ["python", "-m", "pip", "install", "-r", "requirements.txt"]
 INSTALL_CONNECTOR_PACKAGE_CMD = ["python", "-m", "pip", "install", "."]
+INSTALL_POETRY_PACKAGE_CMD = ["python", "-m", "pip", "install", "poetry"]
+POETRY_INSTALL_DEPENDENCIES_CMD = ["poetry", "install"]
+
 DEFAULT_PYTHON_EXCLUDE = [".venv"]
+POETRY_EXCLUDE = ["__pycache__"] + DEFAULT_PYTHON_EXCLUDE
 CI_CREDENTIALS_SOURCE_PATH = "tools/ci_credentials"
 CI_CONNECTOR_OPS_SOURCE_PATH = "tools/ci_connector_ops"
```
alafanechere, line 35: nit: what do you think about moving these constant directly to the environment function they're used in. I think it my help for readability and they are not reused.

alafanechere (changes requested):
I'm glad you plugged in this codebase!!! #collaboration
I think you can benefit a bit more from the classes I've declared in `bases.py`.
I would also like to challenge the existence of subcommand group and would prefer simple CLI like: `airbyte-ci test-connectors` or `airbyte-ci test-metadata`

.github/actions/run-dagger-pipeline/action.yml
bnchrch: Im thinking Ill add a few more commands before conors repo is ready.

Should I just use this action for my new commands then and leave the existing ones alone to avoid refactoring twice?

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/main.py
```diff
@@ -0,0 +1,57 @@
...
+    "--diffed-branch",
+    help="Branch to which the git diff will happen to detect new or modified connectors",
+    default="origin/master",
+    type=str,
+)
+@click.option("--gha-workflow-run-id", help="[CI Only] The run id of the GitHub action workflow", default=None, type=str)
+@click.option("--ci-context", default="manual", envvar="CI_CONTEXT", type=click.Choice(["manual", "pull_request", "nightly_builds"]))
+@click.option("--pipeline-start-timestamp", default=get_current_epoch_time, envvar="CI_PIPELINE_START_TIMESTAMP", type=int)
+@click.pass_context
+def airbyte_ci_pipeline(
```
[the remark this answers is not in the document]
bnchrch: Good idea, my thought with this was that these commands were always going to relate to running dagger pipelines (hence the prefix) but Im happy to change it to `airbyte-ci`

On the note of groups. Im thinking that as this grows we will want different command groups? one set thats specific for connector related tasks, another specific to metadata related tasks. Another perhaps for docs specific tasks?

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/main.py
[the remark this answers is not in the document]
bnchrch: I dont have exactly a strong opinion here. Just wanted a separation from this is the pipeline code and this is the command that runs the code.

thoughts about
```
commands/airbyte_ci.py
commands/airbyte_ci/test_connectors.py
commands/airbyte_ci/metadata_service.py
```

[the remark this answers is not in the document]
bnchrch: Feel free to say no, do this instead. This is your project captain!

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
bnchrch: Generally in favor! With the exception of the paths.

Love getting the path in constants pattern set up early so we dont have to play find where we hardcoded the path later

tools/ci_connector_ops/ci_connector_ops/pipelines/github.py
bnchrch: Ahh my mistake.

Question, why do we do this above?
```py
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from logging import Logger
```

tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
bnchrch: Hmm good callout.

I want to push back slightly. But only because I may not fully understand.

But would sensible defaults be what we have
1. On error -> set commit status to error
2. On success -> set commit status to success

For the pipelines im intending to implement I think thats all I need?

These are the tests im thinking about
1. Run metadata lib unit test -> pass/fail
2. Run metadata orchestrator unit test -> pass/fail 
3. Run metadata validator -> pass/fail

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/metadata_service.py
```diff
@@ -0,0 +1,73 @@
...
+from rich.logging import RichHandler
+
+
+logging.basicConfig(level=logging.INFO, format="%(name)s: %(message)s", datefmt="[%X]", handlers=[RichHandler(rich_tracebacks=True)])
+logger = logging.getLogger(__name__)
+
+METADATA_LIB_MODULE_PATH = "airbyte-ci/connectors/metadata_service/lib"
+
+
+class MetadataLibRunTest(Step):
```
bnchrch, line 23: @alafanechere putting this up now slightly incomplete (needs some comments and classes moved to new files) just to get your thoughts on this.

Generally do you feel like im using the classes you defined in the correct spirit of things?

Also I had an idea while writing this. What if we made the PipelineContext hold onto a collection of steps and the be responsible for running them?

```python
async def run_metadata_pipeline(cli_context):
        metadata_pipleline_context = PipelineContext(
            pipeline_name="Metadata Service Lib Unit Test Pipeline",
            is_local=cli_context.obj["is_local"],
            git_branch=cli_context.obj["git_branch"],
            git_revision=cli_context.obj["git_revision"],
            gha_workflow_run_url=cli_context.obj.get("gha_workflow_run_url"),
            pipeline_start_timestamp=cli_context.obj.get("pipeline_start_timestamp"),
            ci_context=cli_context.obj.get("ci_context")
            dagger_client = dagger_client
            steps = [
                MetadataLibRunTest,
            ]
        )

        async with metadata_pipleline_context:
            test_report = await metadata_pipleline_context.run()
            return test_report
            ```
Where the pipeline context becomes concerned with everything that happens around a pipeline run
1. Logging
2. Reporting status to github
3. Ordering steps
4. Reporting outcome

File: tools/ci_connector_ops/ci_connector_ops/pipelines/commands/main.py (hunk 12 above)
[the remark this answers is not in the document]
Comment (alafanechere) [reply tools/ci_connector_ops/ci_connector_ops/pipelines/commands/main.py]: Yes sure, but I'd like to keep these commands as concise as possible... Would you mind following my suggestion for now and we'll probably re-evaluate the command structure in the future according to use cases growth?

File: tools/ci_connector_ops/ci_connector_ops/pipelines/github.py (hunk 8 above)
Comment (alafanechere) [reply tools/ci_connector_ops/ci_connector_ops/pipelines/github.py]: I'm not sure this is required in this context but sometime we only make imports to use these as type hints. And sometime we can't import a module because of circular dependencies.
`if TYPE_CHECKING` allows to only import a module when `mypy` is performing type checks.

File: tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py (hunk 11 above)
Comment (alafanechere) [reply tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py]: Sounds good to me.

File: tools/ci_connector_ops/ci_connector_ops/pipelines/commands/metadata_service.py (hunk 13 above)
Comment (alafanechere) [reply tools/ci_connector_ops/ci_connector_ops/pipelines/commands/metadata_service.py:23]: > Also I had an idea while writing this. What if we made the PipelineContext hold onto a collection of steps and the be responsible for running them?

I like it but it would mean implementing a PipelineContext for each pipeline type and PipelineContext would need to be abstract (as it was before). I feel like its too big of a lift to make this happen in the current PR as it will change quite a lot of things in your PR for adapting the connector pipeline, and in this [PR](https://github.com/airbytehq/airbyte/pull/24225). 
Would you mind creating an issue on the "Connectors CI in production" epic to track this future improvement?

File: tools/ci_connector_ops/ci_connector_ops/pipelines/commands/metadata_service.py [hunk 14]
```diff
@@ -0,0 +1,73 @@
...
+    title = "Run Metadata Service Lib Unit Tests"
```
alafanechere, line 24: nit for later: I'm wondering if we should add a `description` attribute that could explain to the world the philosophy of the implemented pipeline, with links to the thing under test code etc. It would be interesting metadata to propagate to logs or automatically generate docs.

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/airbyte_ci.py
```diff
@@ -0,0 +1,60 @@
...
+@click.option("--git-branch", default=get_current_git_branch, envvar="CI_GIT_BRANCH")
+@click.option("--git-revision", default=get_current_git_revision, envvar="CI_GIT_REVISION")
+@click.option(
+    "--diffed-branch",
+    help="Branch to which the git diff will happen to detect new or modified connectors",
+    default="origin/master",
+    type=str,
+)
+@click.option("--gha-workflow-run-id", help="[CI Only] The run id of the GitHub action workflow", default=None, type=str)
+@click.option("--ci-context", default="manual", envvar="CI_CONTEXT", type=click.Choice(CI_CONTEXT_VALUES))
```
alafanechere, line 30: I think you can directly use the enum class in `Choice`. It worked for ConnectorLanguage:
https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/connectors_ci.py#L137

alafanechere (approved):
I approve to unblock you. I shared my last suggestions.
For the command group discussion, on a second thought I'm not against subgroup but I would love to have a concise CLI. 
`airbyte-ci connectors test --name=source-pokeapi`
`airbyte-ci connectors publish --name=source-pokeapi`
`airbyte-ci metadata test lib (lib being an argument)`
`airbyte-ci cdk test`
`airbyte-ci cdk publish`
`airbyte-ci connector-acceptance-test test`
etc...

alafanechere:
@bnchrch as you modified the CLI could you please update the existing GHA workflow using it in this PR? (or refactor them to use the action you wrote)

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/metadata_service.py
bnchrch: Added the issue here: [Update Dagger pipeline class to manage the running of steps](https://app.zenhub.com/workspaces/connector-operations-6318ede5eda32e0010ef8566/issues/gh/airbytehq/airbyte/24752)

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/metadata_service.py
bnchrch: I like that idea! Feel free to add an issue for it!

PR #24657, opened 2023-03-29
24153 Destination Databricks: Fix support for external tables on S3

### LINKED ISSUE #24153: Fix support for external tables on S3 in DatabBricks destination
Definition of done:
 - demo of setting up destination-databricks with external tables starting from Databricks set-up and ending with a working Airbyte connection
 - documentation on how to configure all settings for using destination-databricks with external tables in S3

suhomud:
## What
Make Databricks external tables working. Update Databricks Setup doc.  Closes https://github.com/airbytehq/airbyte/issues/24153

## How
- To make External tables working the `Storage Locations`  should be added on Databricks Workspace console side: 
image
- The Databrciks AWS Setup process was documented in databricks.md file.

## Recommended reading order
1. all java code is just related to adding a new spec field `catalog` to external tables related code.
2. `databricks.md` file has been modified with documenting Databricks AWS setup including external tables. 

## 🚨 User Impact 🚨
No user impact.


edgao:
@airbytehq/destinations - I probably won't get to this in a timely fashion. Dumping contents of a convo with yevhen:

this PR does not modify the destination spec in any way

but it does fix behavior to respect the `catalog` config option. This may be a breaking change for existing connections, _if_ they have a non-null catalog. We should check active connections to see if they fall into that bucket - if none of them are actively using the feature, then we should be OK to ship.... but otherwise, we'll need to coordinate breaking change comms with TCS.

evantahler:
@bgroff and I just ran though this PR in detail, and we successfully ran syncs storing data to both the default location (managed tables to an s3 holding the default hive-metastore belonging to the workspace) and managed tables which live within a unity catalog managed S3 bucket outside of the workspace

evantahler (approved):

TBernstein4:
> @airbytehq/destinations - I probably won't get to this in a timely fashion. Dumping contents of a convo with yevhen:
> 
> this PR does not modify the destination spec in any way
> 
> but it does fix behavior to respect the `catalog` config option. This may be a breaking change for existing connections, _if_ they have a non-null catalog. We should check active connections to see if they fall into that bucket - if none of them are actively using the feature, then we should be OK to ship.... but otherwise, we'll need to coordinate breaking change comms with TCS.

@edgao a user chatted in letting me know that this was a breaking change for them. According to the user they previously had a null value for the Databricks Catalog which would then default to "hive_metastore", however after this update it instead supplied a `null` value. Leading to the following error:
SQL state: org.apache.hive.service.cli.HiveSQLException: Error running query: org.apache.spark.sql.AnalysisException: spark_catalog.**_null_**.airbyte._airbyte_tmp_*** is not a valid

TBernstein4:
That is correct

edgao:
should we publish a new version to treat `null` as equivalent to hive_metastore? It sounds like that would be identical to past behavior

TBernstein4:
That is what the docs indicate the expected behavior to be. 
Screenshot 2023-04-20 at 2 48 40 PM

edgao:
submitted https://github.com/airbytehq/airbyte/pull/25366. assuming things go smoothly, it'll get published by end of day.

Affected users will need to run a reset or manually copy data from the `null` catalog back into the hive_metastore catalog, after that PR gets deployed.

edgao:
@TBernstein4 1.0.2 is live now

evantahler:
Thank you for the fix, @edgao!

TBernstein4:
Yes, thanks for such a quick fix Ed!

PR #24660, opened 2023-03-29
Source Salesforce: set default start date

roman-yermilov-gl:
## What
*Set default start_date. Sync for last two years if start date is not present in config*


airbyte-integrations/connectors/source-salesforce/source_salesforce/source.py
```diff
@@ -25,6 +27,9 @@ class AirbyteStopSync(AirbyteTracedException):
 
 
 class SourceSalesforce(AbstractSource):
+    DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+    START_DATE_OFFSET_IN_YEARS = 2
```
lazebnyi, line 31: Please update doc to this changes.

lazebnyi (approved):
LGTM! Need update doc.

PR #24567, opened 2023-03-27
🎉 Source Orb: Add invoices incremental stream

kgrover:
## What
This adds the invoices resources to the Orb source connector. Like other resources, this supports an incremental sync in the sense that it should captures resources once but updates to the same resource will require a full resync (for example, when an invoice goes from `issued` -> `paid`, its `invoice_date` will still be in the past).

## How
We paginate invoices based on `invoice_date`, and otherwise keep the incremental orb stream contract the same.

## Recommended reading order
1. `source.py`
2. `schemas/invoices.json`

## 🚨 User Impact 🚨
This should enable syncing invoices for existing connectors, but it's not a breaking change for existing connectors.

## Pre-merge Checklist
<details><summary><strong>Updating a connector</strong></summary>

### Community member or Airbyter

- [ ] Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [x] Secrets in the connector's spec are annotated with `airbyte_secret`
- [x] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [x] Connector version has been incremented
    - [ ] Version has been bumped according to our [Semantic Versioning for Connectors](https://docs.airbyte.com/contributing-to-airbyte/semantic-versioning-for-connectors) guidelines
    - [ ] `Dockerfile` has updated version
- [ ] Documentation updated
    - [ ] Connector's `README.md`
    - [ ] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - [ ] Changelog updated in `docs/integrations/<source or destination>/<name>.md` with an entry for the new version. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)

- [ ] PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests)

### Airbyter

If this is a community PR, the Airbyte engineer reviewing this PR is responsible for the below items.

[...]

PR #24737, opened 2023-03-30
🎉 Source Orb: Add invoices incremental stream

girarda:
## What
* PR to test and publish @kgrover's [work](https://github.com/airbytehq/airbyte/pull/24567)

> This adds the invoices resources to the Orb source connector. Like other resources, this supports an incremental sync in the sense that it should captures resources once but updates to the same resource will require a full resync (for example, when an invoice goes from issued -> paid, its invoice_date will still be in the past).


docs/integrations/sources/orb.md
```diff
@@ -54,6 +54,7 @@ an Orb Account and API Key.
 
 | Version | Date | Pull Request | Subject |
 | --- | --- | --- | --- |
+| 1.1.0 | 2023-03-26 |  | Add Invoices incremental stream
```
pmossman, line 57: nit: could add a link to the original PR

pmossman (approved):
Looks good to me!

PR #24630, opened 2023-03-28
Source TikTok Marketing: migrate to v1.3 API (incompatible schema change!)

grubberr:
## What
Issue: https://github.com/airbytehq/airbyte/issues/23375
Issue: https://github.com/airbytehq/airbyte/issues/15736
Issue: https://github.com/airbytehq/airbyte/issues/23509

- **incompatible schema change!**
- API changed: v1.2 -> v1.3
- fail_on_extra_columns - removed
- Stream: "advertisers" primary_key changed: "id" -> "advertiser_id"
- expected_records fully regenerated

#### Stream "ad_groups"
old name  | new name
----------|-------------
placement | placements
enable_inventory_filter | inventory_filter_enabled
is_comment_disable (integer) | comment_disabled (boolean)
external_type | promotion_type
external_action | optimization_event
deep_external_action | secondary_optimization_event
modify_time (string) | modify_time (datetime)
create_time (string) | create_time (datetime)
audience | audience_ids
excluded_audience | excluded_audience_ids
location | location_ids
interest_category_v2 | interest_category_ids
interest_keywords | interest_keyword_ids
age | age_groups
operation_system | operating_systems
connection_type | network_types
device_price | device_price_ranges
android_osv | min_android_version
ios_target_device | ios14_targeting
device_models | device_model_ids
ios_osv | min_ios_version
schedule_start_time (string) | schedule_start_time (datetime)
schedule_end_time (string) | schedule_end_time (datetime)
optimize_goal | optimization_goal
bid | bid_price
conversion_bid | conversion_bid_price
deep_cpabid | deep_cpa_bid
status | secondary_status
opt_status | operation_status
carriers_v2 | carrier_ids
video_download | video_download_disabled
pangle_block_app_list_id | blocked_pangle_app_ids
action_categories | action_category_ids
rf_buy_type | rf_purchased_type
buy_impression | purchased_impression
buy_reach | purchased_reach
rf_predict_cpr | rf_estimated_cpr
rf_predict_frequency | rf_estimated_frequency
pangle_audience_package_include | included_pangle_audience_package_ids
pangle_audience_package_exclude | excluded_pangle_audience_package_ids
catalog_authorized_bc | catalog_authorized_bc_id
include_custom_actions | included_custom_actions
exclude_custom_actions | excluded_custom_actions
dpa_retargeting_type | shopping_ads_retargeting_type
brand_safety | brand_safety_type
action_v2 | actions
 . | schedule_infos
 . | share_disabled
[...]

PR #24741, opened 2023-03-31
Source TikTok: correct PK for daily reports streams

### LINKED ISSUE #23728: Source TikTok Marketing: Incorrect PK resulting in deleted data
## Environment
- **Airbyte version**:  0.39.42-alpha
- **OS Version / Instance**: 
- **Deployment**: 
- **Source Connector and version**:  TikTok Marketing 1.0.0
- **Destination Connector and version**: BigQuery 1.1.11
- **Step where error happened**: Run (Append + Dedup Histo)

## Current Behavior
When using the TikTok marketing Source, for every `<object>_daily_reports` using `Incremental | Dedup + History` mode we only get one line x `<object>_id` on the denormalized table.
_ex._
On the stream `ad_groups_reports_daily`, the PK configured is `adgroup_id`.
The denormalization process creates a table called `ad_groups_reports_daily`.
The resulting data, after a few days of run is 1 line per `adgroup_id`
The result is to have only partial data, the last available point of data for a given `adgroup` instead of all of the data points available for this `adgroup`.

This behavior is the same for every layer on the TikTok data (ie. advertisers, campaigns, adgroups, and ads) for every xxx_report_daily. We did not had the opportunity to test the `xxx_report_hourly` and `xxx_report_lifetime` and neither the audience reports.

## Expected Behavior
We would think to have 1 line x `<object>` x `stat_time_day`

_ex._
On the same scenario, we would think to have 1 line per `adgroup_id` x `stat_time_day`

## Possible Solution
One possible solution could be to add the `stat_time_day` (a composite one) as a primary key that will be used on the dedup process. (ie. `<object>_id` and  `stat_time_day`  as a PK)
Here is an exemple of the BingAds Connector that have composite PK for report streams wich should have the exact same behavior (almost all media platform are working the same way, with a X day window where the data can be updated) :
image

## Logs
-

## Steps to Reproduce
1. Create a Connection using TikTok marketing as source and BQ as a destination
2. Enable the ad_groups_reports_daily stream and put it in Incremental dedup + history mode.
3. Run the connection for at least 2 days
4. Take a look at the resulting data (denormalized one).

I don't feel I would be able to 😄

darynaishchenko:
resolved: https://github.com/airbytehq/airbyte/issues/23728
Added `stat_time_day` to base PK for daily reposrt stream to avoid incorrect data deletion.


grubberr:
@darynaishchenko 

such primary_key for stream `ads_reports_hourly` looks suspicious

```
python3 main.py discover --config secrets/config.json | jq '.catalog.streams[] | select(.name == "ads_reports_hourly")' | jq .source_defined_primary_key[][]
"ad_id"
```

I think it has to be `["ad_id", "stat_time_hour"]`
?

grubberr (changes requested):
pk for `ads_reports_hourly` has to be `["ad_id"], ["stat_time_hour"]`

bazarnov (approved):

grubberr:
This branch was merged to this PR https://github.com/airbytehq/airbyte/pull/24630
That's why we can close this PR and related issue

Issue #24416, opened 2023-03-23
Fix Destination Integration Tests


ryankfu:
### What

Fix Destination Connector Integration Tests

This is currently blocking changes for a few in-flight features: checkpointing and check with catalog

```[tasklist]
### Tasks
- [ ] Clickhouse (checkpoint)
- [ ] Cassandra
- [ ] Databricks
- [ ] Apache Iceberg
- [ ] MSSQL (checkpoint)
- [ ] MySQL (checkpoint)
- [ ] Postgres (checkpoint)
- [ ] Rockset
```

jbfbell:
Starting on Clickhouse

ryankfu:
@jbfbell as mentioned yesterday a couple of integration tests have been fixed:

* [destination-mssql](https://dnsgjos7lj2fu.cloudfront.net/tests/summary/connectors/destination-mssql/)
* [destination-postgres](https://dnsgjos7lj2fu.cloudfront.net/tests/summary/connectors/destination-postgres/)

from this [PR](https://github.com/airbytehq/airbyte/pull/24431)

After your change gets merged, we can remove [Clickhouse](https://github.com/airbytehq/airbyte/pull/24646#pullrequestreview-1367284009) from this list

Just for brevity, is this the [ticket](https://github.com/airbytehq/airbyte/issues/24170) that also addressed MySQL failures or should there be another ticket filed?

jbfbell:
@ryankfu for whatever reason it won't save my clicking of the checkbox for Clickhouse, but it is done

jbfbell:
Starting Cassandra

PR #24763, opened 2023-03-31
Fix Cassandra Integration Test

jbfbell:
Fixes the Cassandra Integration Test.

The Session was closing prior to the `checkFinalState` test, presumably as a result of closing the messageconsumer? But this test will now create a new connection


jbfbell:
Related to #24416

evantahler (approved):
`/test` passed, and this was a PR to fix the test... so... 👍 from me!

Looks like the main logical change was the `AfterAll` to close the connection.  Makes sense!

ryankfu (approved):

PR #22362, opened 2023-02-03
🚨🚨 Source Amplitude: migrate to YAML

artem1205:
## What
Resolving:

- #22142 
- #21010
- #19036
- #16943

Migrate to YAML

## How
*Describe the solution*

## Recommended reading order
1. `y.python`

## 🚨 User Impact 🚨
`stream state` and `cursor field` for  stream event has been changed.

## Pre-merge Checklist
Expand the relevant checklist and delete the others.

<details><summary><strong>Updating a connector</strong></summary>

### Community member or Airbyter

- [ ] Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
- [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`
    - [ ] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - [ ] Changelog updated in `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
- [ ] PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests)

### Airbyter

If this is a community PR, the Airbyte engineer reviewing this PR is responsible for the below items.

- [ ] Create a non-forked branch based on this PR and test the below items on it
- [ ] Build is successful
- [ ] If new credentials are required for use in CI, add them to GSM. [Instructions](https://docs.airbyte.io/connector-development#using-credentials-in-ci).
- [ ] [`/test connector=connectors/<name>` command](https://docs.airbyte.io/connector-development#updating-an-existing-connector) is passing
[...]

PR #24772, opened 2023-04-02
Low-Code CDK: make RecordFilter.filter_records as generator

grubberr:
This PR Improves this issue: https://github.com/airbytehq/airbyte/pull/22362

## Description

The `source-amplitude` connector's `events` stream returns a large amount of data from the API, which consumes a significant amount of memory. This data is then processed by the `RecordFilter.filter_records` function to produce a new list that duplicates memory usage, causing the connector to consume even more memory before issuing the first record.

## Solution

To improve memory usage and speed up the production of the first record, we can modify `RecordFilter.filter_records` to be a generator instead of returning a list. This will enable the function to yield one record at a time, without having to process the entire dataset at once. By doing so, we can reduce memory consumption and improve the efficiency of the connector.


girarda (approved):

maxi297:
@grubberr @girarda How does this work? The API clearly expect a List at multiple place, one of them being [here](https://github.com/airbytehq/airbyte/blob/4da621f96ab8ba84f3d630b40b35238fd27a5639/airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py#L56)

PR #24774, opened 2023-04-02
Group documentation navigation

timroes:
## What

Groups the navigation in the documentation into several categories, to make it easier to parse.

Also since they are grouped now I removed the duplicate "Airbyte" or "Open Source" from some of the menu entries, if they were redundant. So the new structure looks as follows:

screenshot-20230402-232008

The only links changed due to that, are the `category/` links if the page itself doesn't have a content. I fixed the only link I found we have to one, but I don't think we need a redirect for them, since we don't link from any other places to them afaik.

Also fixed some broken links as part of this PR, as well as making sure those carets in the menu are now all the same size.

### DIFF
--- a/docs/operator-guides/security.md
+++ b/docs/operator-guides/security.md
@@ -102,7 +102,7 @@ Only certain Airbyte staff can access Airbyte infrastructure and technical logs 
 
 ### Network security
 
-Depending on your [data residency](https://docs.airbyte.com/cloud/managing-airbyte-cloud#choose-your-default-data-residency) location, you may need to allowlist the following IP addresses to enable access to Airbyte:
+Depending on your [data residency](https://docs.airbyte.com/cloud/managing-airbyte-cloud/manage-data-residency) location, you may need to allowlist the following IP addresses to enable access to Airbyte:
 
 #### United States and Airbyte Default
 
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -6,7 +6,7 @@ Whether you are an Airbyte user or contributor, we have docs for you!
 
 Browse the [connector catalog](https://docs.airbyte.com/integrations/) to find the connector you want. In case the connector is not yet supported on Airbyte Cloud, consider using [Airbyte Open Source](#for-airbyte-open-source-users).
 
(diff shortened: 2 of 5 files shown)


docs/operator-guides/security.md
```diff
@@ -102,7 +102,7 @@ Only certain Airbyte staff can access Airbyte infrastructure and technical logs
 
 ### Network security
 
-Depending on your [data residency](https://docs.airbyte.com/cloud/managing-airbyte-cloud#choose-your-default-data-residency) location, you may need to allowlist the following IP addresses to enable access to Airbyte:
```
timroes, line 105: :information_source: Was a broken link.

docs/readme.md
```diff
@@ -6,7 +6,7 @@ Whether you are an Airbyte user or contributor, we have docs for you!
 
 Browse the [connector catalog](https://docs.airbyte.com/integrations/) to find the connector you want. In case the connector is not yet supported on Airbyte Cloud, consider using [Airbyte Open Source](#for-airbyte-open-source-users).
 
-Next, check out the [step-by-step tutorial](https://docs.airbyte.com/cloud/getting-started-with-airbyte-cloud) to sign up for Airbyte Cloud, understand Airbyte [concepts](https://docs.airbyte.com/cloud/core-concepts), and run your first sync. Then learn how to [manage your Airbyte Cloud account](https://docs.airbyte.com/category/managing-airbyte-cloud).
```
timroes, line 9: :information_source: Only link that needs to change due to category names.

docusaurus/src/css/custom.css
```diff
@@ -88,7 +95,7 @@ img {
 }
 
 .menu__link--sublist-caret:after {
-    background: var(--ifm-menu-link-sublist-icon) 50%/1.5rem 1.5rem;
+    background: var(--ifm-menu-link-sublist-icon) 50%/2rem 2rem;
```
timroes, line 98: :information_source: Make sure the carets are the same size as the carets defined in the theme.

nataliekwong (approved):
Looks good to me 👍

PR #24816, opened 2023-04-03
Add CICD test for orchestrator using Dagger

### LINKED ISSUE #24209: Ensure airbyte-metadata tests are run on CI/CD
Currently we dont run tests for the orchestrator or metadata service on ci/cd

we should

bnchrch:
## What
We need to run orchestrator unit tests in CICD

closes #24209

## How
This adds that using a dagger pipeline

### Not using an alpine image
Unfortunately dagster using the grpcio library which currently doesnt have "wheel" binaries for alpine. So to avoid 10 minute plus recompile times I changed to use a debian base package

Interesting reading:
https://medium.com/swlh/alpine-slim-stretch-buster-jessie-bullseye-bookworm-what-are-the-differences-in-docker-62171ed4531d
https://github.com/grpc/grpc/issues/22815
https://github.com/grpc/grpc/issues/24722

### Poetry sibling modules
I updated with_poetry to let you copy a parent working directory of the poetry module to allow for relative imports

### DIFF
--- a/.github/actions/run-dagger-pipeline/action.yml
+++ b/.github/actions/run-dagger-pipeline/action.yml
@@ -33,7 +33,7 @@ runs:
         python-version: "3.10"
     - name: Install ci-connector-ops package
       shell: bash
-      run: pip install -e ./tools/ci_connector_ops\[pipelines]\
+      run: pip install ./tools/ci_connector_ops\[pipelines]\
     - name: Run airbyte-ci
       shell: bash
       run: |
--- a/.github/actions/run-dagger-pipeline/action.yml
+++ b/.github/actions/run-dagger-pipeline/action.yml
@@ -33,7 +33,10 @@ runs:
         python-version: "3.10"
     - name: Install ci-connector-ops package
       shell: bash
-      run: pip install ./tools/ci_connector_ops\[pipelines]\
+      run: |
+        pip install --upgrade pip
+        pip install ./tools/ci_connector_ops[pipelines]
+        echo $PATH
     - name: Run airbyte-ci
       shell: bash
       run: |
--- a/.github/actions/run-dagger-pipeline/action.yml
+++ b/.github/actions/run-dagger-pipeline/action.yml
@@ -33,10 +33,7 @@ runs:
         python-version: "3.10"
     - name: Install ci-connector-ops package
       shell: bash
-      run: |
-        pip install --upgrade pip
-        pip install ./tools/ci_connector_ops[pipelines]
-        echo $PATH
+      run: pip install --quiet -e ./tools/ci_connector_ops\[pipelines]\
     - name: Run airbyte-ci
       shell: bash
       run: |
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
@@ -199,6 +199,18 @@ def with_git(base_container: Container) -> Container:
 
     return base_container.with_exec(package_install_command + packages_to_install)
 
+def with_packages(base_container: Container, packages_to_install) -> Container:
+    """Installs  in a alpine based container.
+    Args:
+        context (Container): A alpine based container.
+
+    Returns:
+        Container: A container with the packages installed.
+
+    """
+    package_install_command = ["apk", "add"]
+    return base_container.with_exec(package_install_command + packages_to_install)
+
 
 async def with_ci_connector_ops(context: PipelineContext) -> Container:
     """Installs the ci_connector_ops package in a Container running Python > 3.10 with git..
@@ -248,5 +260,6 @@ def with_poetry_module(context: PipelineContext, parent_dir_path: str, module_pa
 
     src = context.dagger_client.host().directory(parent_dir_path, exclude=poetry_exclude)
     python_with_poetry = with_poetry(context)
+    python_with_cplus = with_packages(python_with_poetry, ["build-base", "linux-headers"])
 
-    return python_with_poetry.with_mounted_directory("/src", src).with_workdir(f"/src/{module_path}").with_exec(poetry_install_dependencies_cmd)
+    return python_with_cplus.with_mounted_directory("/src", src).with_workdir(f"/src/{module_path}").with_exec(poetry_install_dependencies_cmd)
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
@@ -24,7 +24,7 @@ CONNECTOR_TESTING_REQUIREMENTS = [
     "pytest-custom_exit_code",
 ]
 
-DEFAULT_PYTHON_EXCLUDE = ["**/.venv"]
+DEFAULT_PYTHON_EXCLUDE = ["**/.venv", "**/__pycache__"]
 CI_CREDENTIALS_SOURCE_PATH = "tools/ci_credentials"
 CI_CONNECTOR_OPS_SOURCE_PATH = "tools/ci_connector_ops"
 
@@ -185,22 +185,8 @@ async def with_ci_credentials(context: PipelineContext, gsm_secret: Secret) -> C
     return ci_credentials.with_env_variable("VERSION", "dev").with_secret_variable("GCP_GSM_CREDENTIALS", gsm_secret).with_workdir("/")
 
 
-def with_git(base_container: Container) -> Container:
-    """Installs git in a alpine based container.
-    Args:
-        context (Container): A alpine based container.
-
-    Returns:
-        Container: A container with git installed.
-
-    """
-    package_install_command = ["apk", "add"]
-    packages_to_install = ["gcc", "libffi-dev", "musl-dev", "git"]
-
-    return base_container.with_exec(package_install_command + packages_to_install)
-
-def with_packages(base_container: Container, packages_to_install) -> Container:
-    """Installs  in a alpine based container.
+def with_alpine_packages(base_container: Container, packages_to_install: List[str]) -> Container:
+    """Installs packages using apk-get.
     Args:
         context (Container): A alpine based container.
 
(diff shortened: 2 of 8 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -199,6 +199,18 @@ def with_git(base_container: Container) -> Container:
 
     return base_container.with_exec(package_install_command + packages_to_install)
 
+def with_packages(base_container: Container, packages_to_install) -> Container:
+    """Installs  in a alpine based container.
+    Args:
+        context (Container): A alpine based container.
```
alafanechere, line 205:
```suggestion
        base_container (Container): A alpine based container.
```
I'm not sure its an alpine based container. Isn't the package manager `apk` on Alpine?
I like this new function but I'm a bit worried it won't work for any container you pass in depending of its base image...

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -235,18 +247,18 @@ def with_poetry(context: PipelineContext) -> Container:
...
         context (PipelineContext): The current test context, providing the repository directory from which the ci_credentials sources will be pulled.
     Returns:
         Container: A python environment with dependencies installed using poetry.
     """
-    poetry_exclude = ["__pycache__"] + DEFAULT_PYTHON_EXCLUDE
+    poetry_exclude = ["**/__pycache__"] + DEFAULT_PYTHON_EXCLUDE
     poetry_install_dependencies_cmd = ["poetry", "install"]
 
-    src = context.dagger_client.host().directory(src_path, exclude=poetry_exclude)
+    src = context.dagger_client.host().directory(parent_dir_path, exclude=poetry_exclude)
```
alafanechere, line 261: I think we should not access FS directly in environment. I would suggest you use `get_repo_dir` on the context object. It wraps a "smart" logic to return a `Directory` object from either the Host (if local) or the Git repository (if you're running in the CI)

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -41,7 +47,45 @@ async def run_metadata_lib_test_pipeline(
...
+
+    return metadata_pipeline_context.test_report.success
+
+async def run_metadata_orchestrator_test_pipeline(
+    is_local: bool,
+    git_branch: str,
+    git_revision: str,
+    gha_workflow_run_url: Optional[str],
+    pipeline_start_timestamp: Optional[int],
+    ci_context: Optional[str],
```
alafanechere, line 67: This signature will be lighter once we have a GlobalContext 🙏

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/metadata.py
```diff
@@ -46,6 +46,27 @@ def lib(ctx: click.Context):
...
+            ctx.obj.get("gha_workflow_run_url"),
+            ctx.obj.get("pipeline_start_timestamp"),
+            ctx.obj.get("ci_context"),
+        )
+        if not pipeline_success:
+            raise dagger.DaggerError("Metadata Service Lib Unit Test Pipeline failed.")
+
+    except dagger.DaggerError as e:
+        click.secho(str(e), err=True, fg="red")
+        return sys.exit(1)
```
alafanechere, line 68: nit: As we reuse this try  except block quite a lot I think we could use a decorator to decorate the commands with. I could manage the `logger.info` and the `try except`

alafanechere (commented):
Nice! My main concern is around file system access / mount, I suggest you use the `get_repo_dir` method of the context object if you need to mount directories to containers.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
bnchrch: Ah sorry I saw that "python:3-alpine" was used when calling

But I did not see the `slim` default.

We should probably pick one type of linux distro and stick to it.

I would vote for debian for the same reasons in [this post](https://medium.com/swlh/alpine-slim-stretch-buster-jessie-bullseye-bookworm-what-are-the-differences-in-docker-62171ed4531d)

> However, some teams are moving away from alpine because these images can cause compatibility issues that are hard to debug. Specifically, if using python images, some wheels are built to be compatible with Debian and will need to be recompiled to work with an Apline-based image.

Ive already ran into this in another PR

bnchrch: > What do you think about improving the with_python_base environment and pass optional packages_to_install?

I think in general im against? I really like the composability of what you have set up here. as its very chainable

I know right now its a bit annoying defining all the variables
```py
python_container = with_python_base(context)
python_with_packages = with_packages(python_container, ["git"])
python_with_packages_poetry = with_poetry(python_with_packages)
# etc...
```

But I think its the right direction because with a little bit of work we could extend this pattern to work with a chain function
```py
container = chain_functions(
    context,
    with_python_base,
    with_packages(["git"]),
    with_poetry,
)
```

or start using these like class methods

```py
context.with_python_base().with_packages(["git"]).with_poetry()
```

I think these are better than ending up with a lot of arguments to `with_python_base`

```py
with_python_base(
    context,
    includes=[],
    excludes=[],
    system_packages=[],
    pip_packages=[],
    local_poetry_packages=[],
)
```

But thats an opinion! 😅

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -235,18 +247,18 @@ def with_poetry(context: PipelineContext) -> Container:
     return poetry_with_cache
 
 
-def with_poetry_module(context: PipelineContext, src_path: str) -> Container:
+def with_poetry_module(context: PipelineContext, parent_dir_path: str, module_path: str) -> Container:
```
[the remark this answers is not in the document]
bnchrch: Typically I would agree. But something feels weird about calling context functions (`get_repo_dir`) explicitly in our CLI commands. 

Should this not be an implementation detail that our primary abstractions are only aware of internally?

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/metadata.py
bnchrch: Great idea!

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/metadata.py
```diff
@@ -27,24 +26,32 @@ def test(ctx: click.Context):
...
+        ctx.obj["is_local"],
+        ctx.obj["git_branch"],
+        ctx.obj["git_revision"],
+        ctx.obj.get("gha_workflow_run_url"),
+        ctx.obj.get("pipeline_start_timestamp"),
+        ctx.obj.get("ci_context"),
+    )
+
+
+@test.command(help="Run unit tests for the metadata service orchestrator.")
```
alafanechere, line 42: nit: I think it does not matter to detail which  nature of test you run. If you add integrations tests I bet you'll make unit tests and integration tests run in the same command.
```suggestion
@test.command(help="Run tests for the metadata service orchestrator.")
```

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -155,3 +157,18 @@ def get_modified_connectors(modified_files: Set[str]) -> Set[Connector]:
         if file_path.startswith(SOURCE_CONNECTOR_PATH_PREFIX) or file_path.startswith(DESTINATION_CONNECTOR_PATH_PREFIX):
             modified_connectors.append(Connector(get_connector_name_from_path(file_path)))
     return set(modified_connectors)
+
+
+def pipeline_command(func):
```
alafanechere, line 162: I think you could achieve the same thing by creating a class inheriting from `click.Command` and overriding its `invoke`command:
```
@test.command(cls=PipelineCommand, help="Run unit tests for the metadata service library.")
```
It's something I did for `octavia-cli` [here](https://github.com/airbytehq/airbyte/blob/master/octavia-cli/octavia_cli/base_commands.py#L10).

alafanechere (approved):
LGTM.
Please consider my suggestion about the `with_poetry_module` parameters.
I agree that we should favor the use of debian base images, but please keep also what've been on top of alpine so far. My work on this [PR](https://github.com/airbytehq/airbyte/pulls/alafanechere) has a lot of moving pieces that I think I succeeded in keeping stable and I'd prefer to not change the base images I used so far. I'll create an issue to do it later.

.github/actions/run-dagger-pipeline/action.yml
```diff
@@ -33,7 +33,7 @@ runs:
         python-version: "3.10"
     - name: Install ci-connector-ops package
       shell: bash
-      run: pip install ./tools/ci_connector_ops\[pipelines]\
+      run: pip install -e ./tools/ci_connector_ops\[pipelines]\
```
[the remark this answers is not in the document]
bnchrch: Nope, whoops, left over from debugging

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -24,7 +24,7 @@
     "pytest-custom_exit_code",
 ]
 
-DEFAULT_PYTHON_EXCLUDE = [".venv"]
+DEFAULT_PYTHON_EXCLUDE = ["**/.venv", "**/__pycache__"]
```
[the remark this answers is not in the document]
bnchrch: Really like that

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
bnchrch: Great way to do this, will update

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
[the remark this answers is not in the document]
bnchrch: Ok im onboard! will update

.github/actions/run-dagger-pipeline/action.yml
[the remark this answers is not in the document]
bnchrch: Without `-e` we were getting this failure.
https://github.com/airbytehq/airbyte/actions/runs/4632791811/jobs/8197275855?pr=24816

Seems like some weird caching behaviour with the hostedtools cache.

As the version `0.2.1` is most recent but its certainly trying to reference code from <0.2.0

When looking around the code base though it seems like in the majority of actions where we install `tools/ci_*` we are using the `-e` flag.

I imagine for this reason.

PR #24859, opened 2023-04-04
Remove SSL toggle for source-postgres in OSS

nguyenaiden:
## What
[Issue](https://app.zenhub.com/workspaces/db--dw-source-connectors-6333360e0a41155061efbcbd/issues/gh/airbytehq/airbyte/24680)
Currently for DB sources that have a strict encrypt variant, the connection will fail even if the user has selected, for example, `verify-ca` and filled out all relevant information. We want to mitigate confusion by having the `ssl_mode` be the sole source of truth for enabling/disabling SSL connection.

## Solution
Remove the ssl key from `spec.json` for `source-postgres` and rely on `ssl_mode` key as the source of truth. This also entails removing any reference to the ssl key in the protocol. Note: this needs to be backwards compatible with historical config file. We will need to ignore the ssl key and only honor `ssl_mode`.

Aside from automated tests, manual tests conducted:
- Successfully synced between an SSL only DB source with the config shown below before the change
image
- Made the code change to remove the toggle and successfully synced between the same DB source and destination
image


airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -182,8 +177,7 @@ public String toJDBCQueryParams(final Map<String, String> sslParams) {
             .map((entry) -> {
               try {
                 final String result = switch (entry.getKey()) {
-                  case JdbcSSLConnectionUtils.SSL_MODE -> PARAM_SSLMODE + EQUALS + toSslJdbcParam(SslMode.valueOf(entry.getValue()))
-                      + JdbcUtils.AMPERSAND + PARAM_SSL + EQUALS + (entry.getValue() == DISABLE ? PARAM_SSL_FALSE : PARAM_SSL_TRUE);
```
rodireich, line 186: Does it work without PARAM_SSL?
I think it should but to verify that it's not needed on either disabled or enabled modes

rodireich:
Also in source-postgres-strict-encrypt I think we're removing this ssl toggle programmatically. 
Maybe cleanup from there altogether (it's something @lmossman is also working on - maybe already done).

lmossman:
I'm not working on this anymore, but I agree that since this field is being removed from the spec, then this line that removes it for strict encrypt specifically should be removed: https://github.com/airbytehq/airbyte/blob/9fb381d1a7dd543ef3693b9e4d48f69c68421865/airbyte-integrations/connectors/source-postgres-strict-encrypt/src/main/java/io/airbyte/integrations/source/postgres_strict_encrypt/PostgresSourceStrictEncrypt.java#L42

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
nguyenaiden: Just manually verified with a successful sync without SSL.

rodireich (approved):

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
akashkulk: Why exactly are we removing this again?

akashkulk (approved):

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
rodireich: The toggle is logically redundant.
Setting ssl mode to disabled has the same effect of setting this toggle to false.

PR #24881, opened 2023-04-04
Decouple flags for debug messages from connector builder log messages

### LINKED ISSUE #24702: Suppress logging to STDOUT for Connector Builder requests

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/24702

## What
Suppresses emitting `DEBUG` Airbyte messages when servicing test read (and other commands) requests from the connector builder module. We want to continuing to emit record messages and log messages for the request/response/slice. 

## How
From the beginning we have coupled the logic to emit connector builder messages and the `--debug` flag functionality. However, these are really two separate use cases and this PR splits them apart so one can be enable with/without the other.

Changes:
- Adds a new flag `emit_connector_builder_messages` to the component factory. This separates how we toggle debug messages vs. messages for the builder server. They should really be decoupled because the `--debug` use case is really meant for connector devs who want more granular info when testing their syncs and wasn't really intended to service the builder. IT was just convenient at the time to reuse it.
- The rest of the code changes integrate this add on behavior in the same pattern as how we restrict page and slice limits for connector builder read commands. 

## Recommended reading order
1. `main.py`
3. `simple_retriever.py`

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/connector_builder/README.md
+++ b/airbyte-cdk/python/airbyte_cdk/connector_builder/README.md
@@ -16,7 +16,7 @@ Note:
 {
   "config": <normal config>,
   "__injected_declarative_manifest": {...},
-  "__command": <"resolve_manifest" | "list_streams" | "stream_read">
+  "__command": <"resolve_manifest" | "list_streams" | "test_read">
 }
 ```
 *See [ConnectionSpecification](https://docs.airbyte.com/understanding-airbyte/airbyte-protocol/#actor-specification) for details on the `"config"` key if needed.
--- a/airbyte-cdk/python/airbyte_cdk/connector_builder/main.py
+++ b/airbyte-cdk/python/airbyte_cdk/connector_builder/main.py
@@ -11,12 +11,15 @@ from airbyte_cdk.connector_builder.connector_builder_handler import list_streams
 from airbyte_cdk.entrypoint import AirbyteEntrypoint
 from airbyte_cdk.models import ConfiguredAirbyteCatalog
 from airbyte_cdk.sources.declarative.manifest_declarative_source import ManifestDeclarativeSource
+from airbyte_cdk.sources.declarative.parsers.model_to_component_factory import ModelToComponentFactory
 from airbyte_cdk.utils.traced_exception import AirbyteTracedException
 
 
 def create_source(config: Mapping[str, Any]) -> ManifestDeclarativeSource:
     manifest = config.get("__injected_declarative_manifest")
-    return ManifestDeclarativeSource(manifest, True)
+    return ManifestDeclarativeSource(
+        source_config=manifest, debug=False, component_factory=ModelToComponentFactory(emit_connector_builder_messages=True)
+    )
 
 
 def get_config_and_catalog_from_args(args: List[str]) -> Tuple[str, Mapping[str, Any], Optional[ConfiguredAirbyteCatalog]]:
--- a/airbyte-cdk/python/airbyte_cdk/connector_builder/main.py
+++ b/airbyte-cdk/python/airbyte_cdk/connector_builder/main.py
@@ -18,7 +18,7 @@ from airbyte_cdk.utils.traced_exception import AirbyteTracedException
 def create_source(config: Mapping[str, Any]) -> ManifestDeclarativeSource:
     manifest = config.get("__injected_declarative_manifest")
     return ManifestDeclarativeSource(
-        source_config=manifest, debug=False, component_factory=ModelToComponentFactory(emit_connector_builder_messages=True)
+        source_config=manifest, component_factory=ModelToComponentFactory(emit_connector_builder_messages=True)
     )
 
 
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -107,10 +107,13 @@ DEFAULT_BACKOFF_STRATEGY = ExponentialBackoffStrategy
 
 
 class ModelToComponentFactory:
-    def __init__(self, limit_pages_fetched_per_slice: int = None, limit_slices_fetched: int = None):
+    def __init__(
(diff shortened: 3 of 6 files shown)


airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -107,10 +107,13 @@
 
 
 class ModelToComponentFactory:
-    def __init__(self, limit_pages_fetched_per_slice: int = None, limit_slices_fetched: int = None):
+    def __init__(
+        self, limit_pages_fetched_per_slice: int = None, limit_slices_fetched: int = None, emit_connector_builder_messages: bool = False
```
brianjlai, line 111: We had a recent discussion about also reducing the backoff rate when making subsequent requests for retryable errors. And that is really hinting that we should just have a general flag to pass to the factory to indicate that we should build components for the connector builder and cut down on the amount of args we pass in. However, I think we can address renaming this when that second use case is worked on.

It's not the best to have a connector builder specific context flag bleed into the factory implementation, but its also the lightest weight way to easily apply the decorator components across the whole factory.

airbyte-cdk/python/airbyte_cdk/connector_builder/main.py
```diff
@@ -11,12 +11,15 @@
...
 from airbyte_cdk.sources.declarative.manifest_declarative_source import ManifestDeclarativeSource
+from airbyte_cdk.sources.declarative.parsers.model_to_component_factory import ModelToComponentFactory
 from airbyte_cdk.utils.traced_exception import AirbyteTracedException
 
 
 def create_source(config: Mapping[str, Any]) -> ManifestDeclarativeSource:
     manifest = config.get("__injected_declarative_manifest")
-    return ManifestDeclarativeSource(manifest, True)
+    return ManifestDeclarativeSource(
+        source_config=manifest, debug=False, component_factory=ModelToComponentFactory(emit_connector_builder_messages=True)
```
girarda, line 21: do we still need the `debug` field in the constructor?

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -406,13 +406,13 @@ def state(self, value: StreamState):
...
         self.stream_slicer.update_cursor(value)
 
-    def _parse_records_and_emit_request_and_responses(self, request, response, stream_state, stream_slice) -> Iterable[StreamData]:
-        # Only emit requests and responses when running in debug mode
-        if self.logger.isEnabledFor(logging.DEBUG):
-            yield _prepared_request_to_airbyte_message(request)
-            yield _response_to_airbyte_message(response)
-        # Not great to need to call _read_pages which is a private method
-        # A better approach would be to extract the HTTP client from the HttpStream and call it directly from the HttpRequester
+    def parse_records(
```
girarda, line 409: do we still need this method or can we call self.parse_response directly from https://github.com/airbytehq/airbyte/pull/24881/files#diff-574e7f982441af95b199fa0d628677b9a654543eaea1534b7516c8679375a235R371 and overwrite it in `SimpleRetrieverTestReadDecorator`?

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -424,6 +424,7 @@ class SimpleRetrieverTestReadDecorator(SimpleRetriever):
     """
 
     maximum_number_of_slices: int = 5
+    emit_connector_builder_messages: bool = False
```
girarda, line 427: I naively expect this to be true for `SimpleRetrieverTestReadDecorator`. When should we not emit the connector builder messages from the `SimpleRetrieverTestReadDecorator`?

Follow up question: If this should always be true, can we get rid of the field entirely and always emit them in https://github.com/airbytehq/airbyte/pull/24881/files#diff-574e7f982441af95b199fa0d628677b9a654543eaea1534b7516c8679375a235R441?

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
girarda: I'm not super opinionated on this. More flags makes it more difficult to use, but it allows us to better control the blast radius of features we implement for the connector builder.

Either way, I agree with waiting until we add a second flag to rename.

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
brianjlai: fair point. i think this feeds into the question of the really granular control over the decorator. My initial thought was maybe we want page limits, but not emitting request/response messages. But discussing now I don't really see a use case for not including these because the connector builder grouper needs them. I'll drop this and we'll always have the decorator emit the messages

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
brianjlai: It isn't quite plug and play because part of what this intermediary `parse_records()` method is that it takes in request + response as parameters which we need in the decorated method to emit log messages. The original `parse_response()` method implementing the `HttpStream`'s abstract method doesn't have the necessary parameters so that's presumably why we had to add `parse_records()` on top

airbyte-cdk/python/airbyte_cdk/connector_builder/README.md
```diff
@@ -16,7 +16,7 @@ Note:
 {
   "config": <normal config>,
   "__injected_declarative_manifest": {...},
-  "__command": <"resolve_manifest" | "list_streams" | "stream_read">
+  "__command": <"resolve_manifest" | "list_streams" | "test_read">
```
brianjlai, line 19: noticed this typo in the readme during my testing

girarda (approved):

PR #24888, opened 2023-04-05
🎉Source Salesforce: add checkpointing

roman-yermilov-gl:
## What
- *Add more frequent checkpointing by implementing stream slicer*
- *Fix duplicates for first two requests*


airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -633,36 +633,62 @@ def get_updated_state(self, current_stream_state: MutableMapping[str, Any], late
...
         if self.name not in UNSUPPORTED_FILTERING_STREAMS:
-            page_token: str = last_record[self.cursor_field]
-            res = {"next_token": page_token}
-            # use primary key as additional filtering param, if cursor_field is not increased from previous page
-            if self.primary_key and self.prev_start_date == page_token:
-                res["primary_key"] = last_record[self.primary_key]
-            return res
+            return {
+                "next_token": last_record[self.cursor_field],
+                "primary_key": last_record[self.primary_key]
```
grubberr, line 642: I have found such streams without primary_key:

contact_lists
contacts_list_memberships
deal_pipelines
form_submissions
property_history
subscription_changes

grubberr (changes requested):

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -633,36 +647,62 @@ def get_updated_state(self, current_stream_state: MutableMapping[str, Any], late
 
 
 class BulkIncrementalSalesforceStream(BulkSalesforceStream, IncrementalRestSalesforceStream):
+    STREAM_SLICE_STEP = 120
```
brianjlai, line 650: How did we arrive at the decision to slice over 120 day windows?

There is still a risk that we will drop records even w/ the slicing. Alex details this pretty well here: https://airbytehq-team.slack.com/archives/C04UY2A9Z53/p1680717935424479?thread_ts=1680710973.672509&cid=C04UY2A9Z53 . The larger the window, the higher chances we hit a rate limit mid way through a slice and we checkpoint with records missing.

airbyte-integrations/connectors/source-salesforce/source_salesforce/source.py
```diff
@@ -147,7 +147,8 @@ def _read_stream(
         except exceptions.HTTPError as error:
             error_data = error.response.json()[0]
             error_code = error_data.get("errorCode")
+            url = error.response.url
             if error.response.status_code == codes.FORBIDDEN and error_code == "REQUEST_LIMIT_EXCEEDED":
-                logger.warn(f"API Call limit is exceeded. Error message: '{error_data.get('message')}'")
+                logger.warning(f"API Call {url} limit is exceeded. Error message: '{error_data.get('message')}'")
                 raise AirbyteStopSync()  # if got 403 rate limit response, finish the sync with success.
```
brianjlai, line 153: I am thinking whether this is the right behavior we want to continue. Due to the 24 hour rate limit, we didn't want to block future syncs, so we just marked it successful. This has its drawbacks and we could lose records. However, now that we have checkpointing at date slice windows, maybe we should more concretely throw back an error and not swallow them. And on the next sync we pick up where the previous bookmark left off.

And now with slices, we can still make incremental progress even if we hit the rate limit issue again instead of retrying the whole sync again

brianjlai (commented):
I have a few high level questions about our existing approach and how it related to the new slice behavior. The new sliced queries make sense, but I have some general concerns about how we still swallow the rate limit errors and return a successful sync which will be compounded now that we do checkpointing

airbyte-integrations/connectors/source-salesforce/source_salesforce/source.py
roman-yermilov-gl: We have checkpointing now but not for full refresh sync. I also wanted to remove it but decided to leave as is

girarda: Failing due to daily rate limits will trigger alerts if 3 workspaces start moving more data than they can. I'm not sure what the best way to expose this kind of limitations this without introducing a new status type

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
roman-yermilov-gl: If to make it smaller then integration tests will last for about 4 hours until a process get killed. While testing we can immediately see that small step leads to big performance issue. One more disadvantage is that amount of queries also increased. So after some testing I decided to make ~3 checkpoints per year because it is better then 1 checkpoint in the end of synchronization and performance is not so bad

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -633,36 +647,62 @@ def get_updated_state(self, current_stream_state: MutableMapping[str, Any], late
...
-            if self.primary_key and self.prev_start_date == page_token:
-                res["primary_key"] = last_record[self.primary_key]
-            return res
+            return {
+                "next_token": last_record[self.cursor_field],
+                "primary_key": last_record.get(self.primary_key)
+            }
         return None
 
+    def stream_slices(
```
davydov-d, line 660: shouldn't we move this to the base class i.e. to implement slicing for the full refresh as well?

roman-yermilov-gl: Full Refresh in BulkSalesforce streams is also using this slicing. Do you think we need not only Bulk API but also REST API to use slices?

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
roman-yermilov-gl: fixed

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
roman-yermilov-gl: Decided to not to move slicing for bulk full refresh stream base class

brianjlai: To bring this up again, if we implement `stream_slices()` on this class `BulkIncrementalSalesforceStream`. But based on the changes to the `IncrementalRestSalesforceStream` above, I would assume that the class also need to rely on this slicing logic so we have slices w/ start/end date. Since `IncrementalRestSalesforceStream` doesn't inherit behavior from this class, is `IncrementalRestSalesforceStream.request_params()` going to work?

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
girarda: nit: we could make this configurable with an optional parameter in the spec

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
clnoll, line 660: @roman-yermilov-gl we want this logic to apply to both IncrementalRestSalesforceStream and BulkIncrementalSalesforceStream, so it should be moved up. That should also allow for cleanup of slice-related logic that was added to `request_params`, since much of it is duplicated between IncrementalRestSalesforceStream and BulkIncrementalSalesforceStream.

Can you also add unit tests for the IncrementalRestSalesforceStream case?

clnoll (changes requested):

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
roman-yermilov-gl: IncrementalRestSalesforceStream was also updated because I wanted to have the same SQL query for both implementations. It does work the same way as before

bnchrch: Hey @roman-yermilov-gl Just wanted to circle back here. If we address these changes we should be able to merge this in :)

roman-yermilov-gl: @girarda 
Moved slicer to `IncrementalRestSalesforceStream` so all the incremental streams use slicing now

roman-yermilov-gl: @bnchrch 
What should I do?)

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -607,14 +630,28 @@ def request_params(
...
 
-        query = f"SELECT {','.join(property_chunk.keys())} FROM {self.name} "
         if start_date:
-            query += f"WHERE {self.cursor_field} >= {start_date} "
+            where_conditions.append(f"{self.cursor_field} >= {start_date}")
+        if end_date:
+            where_conditions.append(f"{self.cursor_field} < {end_date}")
         if self.name not in UNSUPPORTED_FILTERING_STREAMS:
-            query += f"ORDER BY {self.cursor_field} ASC"
+            order_by_clause = f"ORDER BY {self.cursor_field} ASC"
```
clnoll, line 650: Why is it that we're only ordering by `self.cursor_field` in the incremental sync case, but ordering by cursor & primary key in the bulk case? Should they be consistent?

roman-yermilov-gl: We are using primary key in bulk operations query in `WHERE` clause and in `ORDER BY` clause when primary key exists. For this type of queries we need to handle pagination by ourselves so we are slicing by primary key inside base stream slicer. But rest streams pagination works in another way and we don't need it here.

**Why we need primary key for Bulk Streams:**
Given a table
|id|     date     |
|:-|:--------------|
|1|01.01.2023|
|2|01.01.2023|
|3|01.01.2023|
|4|01.01.2023|
|5|01.01.2023|
|6|01.03.2023|

Page size = 2
cursor field = date
primary key = id

Query for first slice would be:
```
SELECT fields FROM table WHERE date >= 01.01.2023 AND date < 01.02.2023 ORDER BY date LIMIT 15000;
```
Salesforce prepares data (max 15000 records but imagine it handles only 2 for example purpose):
|id|     date     |
|:-|:--------------|
|1|01.01.2023|
|2|01.01.2023|

So for now we have only 2 of 5 records satisfied first query and it means we are not ready to move to second slice. And we also see that all the 5 records have the same date `01.01.2023`. This is where `primary key` comes in handy. In order to get next two records we are making second query like that:
```
SELECT fields FROM table WHERE date >= 01.01.2023 AND date < 01.02.2023 AND id > 2 ORDER BY date, id LIMIT 15000;
```
This will return:
|id|     date     |
|:-|:--------------|
|3|01.01.2023|
|4|01.01.2023|

**Why we don't need primary key in REST Stream**:
Here (first rows of current method)
```
if next_page_token:
    """
    If `next_page_token` is set, subsequent requests use `nextRecordsUrl`, and do not include any parameters.
    """
    return {}
```
 we can see that there is a link made by Salesforce and we just use it as is for getting next page.

clnoll: Makes sense, thank you for clarifying!

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -633,36 +670,39 @@ def get_updated_state(self, current_stream_state: MutableMapping[str, Any], late
...
-        query = f"SELECT {','.join(selected_properties.keys())} FROM {self.name} "
-        if start_date:
-            if primary_key and self.name not in UNSUPPORTED_FILTERING_STREAMS:
-                query += f"WHERE ({self.cursor_field} = {start_date} AND {self.primary_key} > '{primary_key}') OR ({self.cursor_field} > {start_date}) "
-            else:
-                query += f"WHERE {self.cursor_field} >= {start_date} "
         if self.name not in UNSUPPORTED_FILTERING_STREAMS:
-            order_by_fields = [self.cursor_field, self.primary_key] if self.primary_key else [self.cursor_field]
-            query += f"ORDER BY {','.join(order_by_fields)} ASC LIMIT {self.page_size}"
+            primary_key = (next_page_token or {}).get("primary_key", "")
```
girarda, line 698: nit: can you rename this variable to `last_key` for clarity

girarda (approved):
left a few small comments, but the flow looks good to me. Thank you for the hard work!

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -592,6 +593,28 @@ def format_start_date(start_date: Optional[str]) -> Optional[str]:
...
         return None
 
+    def stream_slices(
+        self, *, sync_mode: SyncMode, cursor_field: List[str] = None, stream_state: Mapping[str, Any] = None
+    ) -> Iterable[Optional[Mapping[str, Any]]]:
+        base, start, end = (None, None, None)
+
+        now = pendulum.now(tz="UTC")
+        initial_date = pendulum.parse((stream_state or {}).get(self.cursor_field, self.start_date), tz="UTC")
+        period_end = initial_date.add(days=now.diff(initial_date).in_days())
```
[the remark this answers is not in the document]
roman-yermilov-gl: You are right, we don't need those calculations. I am going to get rid of `period_end` and `base` (going to replace `base` with `initial_date`)

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -592,6 +593,28 @@ def format_start_date(start_date: Optional[str]) -> Optional[str]:
...
+
+        slice_number = 1
+        while not end == now:
```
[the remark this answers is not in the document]
roman-yermilov-gl: This is not possible with current logic because `end` is equal to `None` when `now` is a datetime. So only `==` operation can be applied in this comparison

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
roman-yermilov-gl: done

clnoll (approved):
LGTM @roman-yermilov-gl!

roman-yermilov-gl:
@clnoll Thanks
Going to publish and merge on Monday

clnoll:
Awesome, thanks @roman-yermilov-gl!

arsenlosenko (approved):

grubberr (approved):

PR #24891, opened 2023-04-05
Lowcode CDK: implement single use refresh token oauth authenticator

davydov-d:
## What
https://github.com/airbytehq/airbyte/issues/24658

## How
Extrapolating the Single use refresh token authenticator from Python CDK to lowcode CDK.
- Subclassing AbstractOauth2Authenticator and DeclarativeAuthenticator
- Declaring component schema in yaml and python
- Implementing unit tests


airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
```diff
@@ -126,3 +128,166 @@ def access_token(self) -> str:
     @access_token.setter
     def access_token(self, value: str):
         self._access_token = value
+
+
+@dataclass
+class DeclarativeSingleUseRefreshTokenOauth2Authenticator(AbstractOauth2Authenticator, DeclarativeAuthenticator):
```
sherifnada, line 134: this feels like a lot of duplication -- shouldn't we use the existing single use authenticator and configure it using the user inputs instead?

davydov-d: @sherifnada yeah I had the same thoughts but decided to duplicate because of two things:
 - thought we should probably have independent code bases for python & low code CDKs
 - reusing the existing single use authenticator would make it impossible to use interpolated strings as input parameters. 

On the other hand, I think we hardly need those fields to be interpolated.
Now that you have dispelled my doubts - I updated the code, please take a look again

airbyte-cdk/python/airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -180,6 +180,26 @@ class OAuthAuthenticator(BaseModel):
     parameters: Optional[Dict[str, Any]] = Field(None, alias="$parameters")
 
 
+class SingleUseRefreshTokenOAuthAuthenticator(BaseModel):
+    type: Literal["SingleUseRefreshTokenOAuthAuthenticator"]
+    token_refresh_endpoint: str
+    client_id_config_path: Optional[List[str]] = ["credentials", "client_id"]
```
girarda, line 186: While this is inconsistent with the other authenticators, I think this is a good way to avoid the need for string interpolation since we know the client_id should come from the config. 

CC @brianjlai - I'm not particularly interested in introducing more widespread breaking changes, but I think this might be a good pattern to use when possible

girarda (approved):

PR #24916, opened 2023-04-05
Add connector builder compatibility guide

girarda:
## What
* Provide a how-to guide to help users know whether they can use the connector builder

## How
The problem boils down to a 6-item checklist
- The integration is a REST API returning a collection of records synchronously
- Authentication is one of (Basic HTTP, API key, OAuth 2.0, Session)
- If using OAUth 2.0, the refresh token is long-lived
- The data is returned as JSON format
- Records are paginated with one of (limit-offset, page count, link to next page)
- Endpoints do not require GraphQL or query params that are not key-value pairs

This doesn't guarantee compatibility. The more [comprehensive flow](https://whimsical.com/how-should-i-build-my-connector-YNXaNQhRuPuWy68Xg8Akhy@7YNFXnKbZ1CoYnAphCZt4) is 20-question long and requires a lot of upfront investment from the user.

That flow might be useful as a separate guide to help users determine whether they're stuck on a bug or a known limitation.


docs/connector-development/config-based/connector-builder-ui.md
```diff
@@ -8,6 +8,10 @@ The connector builder UI is in alpha, which means it’s still in active develop
 **Developer updates will be announced via our #using-the-cdk Slack channel. If you are using the CDK, please join to stay up to date on changes and issues.**
 :::
 
+
+## Can I integrate with this API using the Connector Builder UI?
+[Follow this guide to determine whether the Connector Builder UI is the right tool to build the connector you need](./connector-builder-compatibility.md).
```
girarda, line 13: it would be nice to embed the checklist in the overview page. For now, I just linked to the guide to avoid duplicating it

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Link to next page (CursorPagination)
+Endpoints paginated with a link to the next page of records typically include either a “Link” field in the response header, or in the response body.
+
+You can search the documentation and the sample response for the “next” keyword.
+
+Example: [Greenhouse](https://developers.greenhouse.io/harvest.html#pagination)
+
+### Does the integration require query params that are not key-value pairs?
+Do requests require complex query parameters? Does it support GraphQL querying?
```
flash1293, line 197: Not really sure what to do about these questions - if I answer yes, what does it mean? If I answer no, what does it mean? Should be clearly answered in the very next sentence. Especially for graphql it's not clear if you don't know what graphql is / how it works.

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+    - [OAuth2.0 with long-lived refresh token](#is-the-oauth-refresh-token-long-lived)
+    - [Session authentication](#session)
+- [ ] [Is the data returned as JSON format?](#is-the-data-returned-as-json-format)
+- [ ] [If records are paginated, are they using one of the following mechanism?](#how-are-records-paginated)
+    - [Limit-offset](#limit-offset--offsetincrement-)
+    - [Page count](#page-count)
+    - [Link to the next page](#link-to-next-page--cursorpagination-)
+- [ ] [Do endpoints only require query parameters that are key-value pairs?](#does-the-integration-require-query-params-that-are-not-key-value-pairs)
+
+You can follow [this chart](https://whimsical.com/how-should-i-build-my-connector-YNXaNQhRuPuWy68Xg8Akhy) to guide you through the process.
```
flash1293, line 16: SHould we add a hint to this sentence that there is more nuance to what's supported and what not and that the chart explains those? From the context I expected that it would only guide through the 6 questions above.

flash1293 (commented):
I think you did a great job distilling it dow to the relevant bits, left a few small comments but it reads pretty well already.

Side note about the chart: A bunch of these branches we are already working on removing in the connector builder (e.g. transformations, error handlers) and others seem relatively easy to do (compressed data, more complex data structures for parameters) - we should tackle these things in the near term

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+
+The sections below provide more details to help you answer the questions.
+
+## Is the integration a REST API returning a collection of records synchronously?
+
+Is the integration a REST API? In doubt, look for the word “REST” in the documentation.
+
+Rook at the shape of the resource path.
```
lmossman, line 24:
```suggestion
Look at the shape of the resource path.
```

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+
+Taking the [Congress API](https://api.congress.gov/#/bill) as an example,
+
+`GET /bill`
+
+Indicates the records can be retrieved by submitting a GET request to the /bill path.The sample response shows that the response returns a collection of records, so the Congress API is REST API returning a collection of records.
```
lmossman, line 30:
```suggestion
Indicates the records can be retrieved by submitting a GET request to the /bill path. The sample response shows that the response returns a collection of records, so the Congress API is a REST API returning a collection of records.
```

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+  ]
+}
+```
+
+Some endpoints are parameterized. An example of such endpoint is the [Congress API’s bill by congress endpoint](https://api.congress.gov/#/bill/bill_list_by_congress).
+`GET /bill/:congress`
+
+These endpoints are also valid synchronous REST endpoints.
+
+This differs from the [Amazon Ads reports endpoint](https://advertising.amazon.com/API/docs/en-us/info/api-overview)), which returns a report ID, which will be generated asynchronously by the source. Because the reports need to be downloaded separately, this is not a synchronous REST API.
```
lmossman, line 75:
```suggestion
This differs from the [Amazon Ads reports endpoint](https://advertising.amazon.com/API/docs/en-us/info/api-overview), which returns a report ID, which will be generated asynchronously by the source. Because the reports need to be downloaded separately, this is not a synchronous REST API.
```

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+Are requests authenticated using the Basic HTTP authentication method? You can search the documentation page for one of the following keywords
+- "Basic Auth"
+- "Basic HTTP"
+- "Authorization: Basic"
+- "Base64"
+
+Example: [Greenhouse](https://developers.greenhouse.io/harvest.html#introduction)
+
+### API Key
+Are requests authenticated using an API key injected either as a query parameter or as a request header?
```
lmossman, line 96: > injected either as a query parameter or as a request header

The current API Key auth method in the connector builder/low-code CDK only supports injecting the API key into the request header at the moment. 

Based on what you wrote here, it sounds like we may want to allow users to inject this into a query parameter instead? They can technically do this by adding a Request Parameter field to a stream, but it feels like we should make this a first-class option in the Auth component if it is common

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+- Yes: [Congress API](https://api.congress.gov/#/)
+- No: [Amazon Ads](https://advertising.amazon.com/API/docs/en-us/info/api-overview)
+
+If the integration is not a synchronous REST API, use the Python CDK.
+
+## What type of authentication is required?
+Look up the authentication mechanism in the API documentation, and identify which type it is.
+
+### Basic HTTP
+Are requests authenticated using the Basic HTTP authentication method? You can search the documentation page for one of the following keywords
```
lmossman, line 87: These section just ask yes/no questions, without indicating what the answer means in terms of compatibility with the builder.

After each of these questions are asked, the next sentence should say something like `If yes, then that is compatible with the Builder. If no, use the Python CDK.`

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+
+## How are records paginated?
+Look up the pagination mechanism in the API documentation, and identify which type it is.
+
+Here are the standard pagination mechanisms:
+
+### Page count
+Endpoints using page count pagination accept two pagination parameters
+1. The number of records to be returned (typically called “page_size”)
+2. The page to request (typically called “page”)
```
lmossman, line 174:
```suggestion
2. The page to request (typically called “page” or “page number“)
```

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+
+Example: [newsapi.ai](https://newsapi.ai/documentation)
+
+![Page-count-example](./assets/connector_builder_compatibility/page_count_example.png)
+
+### Limit-Offset (OffsetIncrement)
+Endpoints using limit-offset pagination accept two pagination parameters
+1. The number of records to be returned (typically called “limit”)
+2. The ID of the first record to return (typically called “offset”)
```
lmossman, line 183:
```suggestion
2. The index of the first record to return (typically called “offset”)
```
I think `index` is probably the more correct term to use here than `ID`?

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+
+Example: [Greenhouse](https://developers.greenhouse.io/harvest.html#pagination)
+
+### Does the integration require query params that are not key-value pairs?
+Do requests require complex query parameters? Does it support GraphQL querying?
+
+An example of an endpoint that would be tricky to implement with the Connector Builder is [SproutSocial’s Analytics endpoint](https://api.sproutsocial.com/docs/#analytics-endpoints).
+
+The endpoint requires a list of filters and metrics.
+![Filters-example](./assets/connector_builder_compatibility/filters_example.png)
```
lmossman, line 202:
```suggestion

![Filters-example](./assets/connector_builder_compatibility/filters_example.png)
```
This was rendering on the same line as the text for me, so adding another newline would probably fix that:

Screen Shot 2023-04-06 at 12 55 35 PM

lmossman (commented):
Left a bunch of small comments but overall this is looking good, I really like the breakdown by component and it is very readable and straightforward to follow

docs/connector-development/config-based/connector-builder-compatibility.md
girarda: I think the description is valid because we do allow for setting the API key as a query param with the workaround, but I agree with fixing the issue instead of documenting the workaround

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+    - [Limit-offset](#limit-offset--offsetincrement-)
+    - [Page count](#page-count)
+    - [Link to the next page](#link-to-next-page--cursorpagination-)
+- [ ] [Do endpoints only require query parameters that are key-value pairs?](#does-the-integration-require-query-params-that-are-not-key-value-pairs)
+
+You can follow [this chart](https://whimsical.com/how-should-i-build-my-connector-YNXaNQhRuPuWy68Xg8Akhy) to guide you through the process.
+
+The sections below provide more details to help you answer the questions.
+
+## Is the integration a REST API returning a collection of records synchronously?
```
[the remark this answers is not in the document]
girarda: Agreed. changed to HTTP

docs/connector-development/config-based/connector-builder-compatibility.md
```diff
@@ -0,0 +1,210 @@
...
+Example: [Congress API](https://api.congress.gov/)
+
+### Link to next page (CursorPagination)
+Endpoints paginated with a link to the next page of records typically include either a “Link” field in the response header, or in the response body.
+
+You can search the documentation and the sample response for the “next” keyword.
+
+Example: [Greenhouse](https://developers.greenhouse.io/harvest.html#pagination)
+
+### Does the integration require query params that are not key-value pairs?
```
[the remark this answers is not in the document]
girarda: fixed

docs/connector-development/config-based/connector-builder-compatibility.md
girarda: rephrased. let me know if this is clearer

docs/connector-development/config-based/connector-builder-compatibility.md
girarda: I updated the whimsical link to point to a board in our shared workspace that does not included the more complicated questions

flash1293 (approved):
LGTM, thanks for the changes!

PR #24923, opened 2023-04-05
Fix version bump mistakes

gosusnp:
## What

Some versions from our docs/changelogs have been auto-bumped by mistake. Restore them to what they were meant to be originally.

## How

Restore the versions to what they were meant to be, new runs of bumpversion should ignore them in the future.

### DIFF
--- a/docs/operator-guides/upgrading-airbyte.md
+++ b/docs/operator-guides/upgrading-airbyte.md
@@ -33,7 +33,7 @@ If you use custom connectors, this upgrade requires all of your connector specs 
 
 :::note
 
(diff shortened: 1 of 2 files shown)


colesnodgrass (approved):

davinchia (approved):
Thanks Jimmy!

Curious, any idea how this came to be? Should we be updating any bumpconf files?

gosusnp:
@davinchia, I think bumpcfg will replace any occurrences of the old version by the new version in the list of files. I suspect some of those files were added to the list so we'd have up-to-date examples using the current version. The downside is that versions that matched the current in those files (changelog for example) ended up drifting because of it.

PR #24225, opened 2023-03-20
connectors-ci: test java connectors

alafanechere:
## What
Closes #23971
This PR is an attempt at running Java connectors tests with Dagger.
I implemented:
* the logic to run different test suite according to connector language.
* the containerization of Gradle task execution in Dagger

## How to test locally
Follow the instructions to install the CLI [here](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/README.md#L13)
run from Airbyte repo root:
```bash
_EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-image://registry.dagger.io/engine:main@sha256:7983a34a8fbcf6fbed3ec96d462556f67ceff52497cd8b60808359ddb8f867a3 airbyte-ci connectors test --name=source-openweather --name=source-bigquery
```

@dagger folks: you can't run this command if you don't have a `credentials.json` file in `airbyte-integrations/connectors/source-bigquery/secrets`. 
You should create a `credentials.json` file following [this structure](https://gist.github.com/alafanechere/455c68c6253923751206246da65cb798). And run
```bash
_EXPERIMENTAL_DAGGER_RUNNER_HOST=docker-image://registry.dagger.io/engine:main@sha256:7983a34a8fbcf6fbed3ec96d462556f67ceff52497cd8b60808359ddb8f867a3 airbyte-ci connectors --use-remote-secrets=False test --name=source-openweather --name=source-bigquery
```

## 🎉 What is working
* Wrapping Gradle in Dagger and providing Gradle an access to a shared docker host works pretty well. The shared docker host is required to expose docker images built with Gradle to other steps that might require it (Connector Acceptance Test). The Gradle caching works.
* Disabling connector acceptance test run in Gradle in favor of the Dagger AcceptanceTest step works well.
* Running  the`source-postgres` tests using `testcontainers` (which is spawning containers to run test database) is working pretty well. I had to disable the cleanup container (Ryuk) because it is not working in the docker-in-docker context. I'm not sure of the impact there.

## 💔 What is not working
See updates in comments
* ~~The destination `integrationTestJava` Gradle task is failing because the Python Gradle plugin can't find Python in the container. Python is required for java destination build because it depends on normalization build.~~
[...]

PR #24929, opened 2023-04-06
Add Dagster Deployment Action skeleton

bnchrch:
## What
We want the orchestrator code to deploy to dagster automatically

## How
We have discarded the hybrid model to use dagster cloud.

As a result we have a dagger pipeline that
1. Installs dagster-cloud
2. Generates a setup.py file
3. Deploys to dagster cloud

## Notes for reviewers
After a longer conversation with Dagster. Generating a setup.py and using it during deployment with Dagster cloud was our best option for relative imports.

They have plans to get rid of the setup.py requirement, but that is not implemented yet.


evantahler:
Oh! I didn't know dagster required a docker image.  I would have thought they could run the python directly

alafanechere (changes requested):
To be consistent with the centralization efforts we started I think that this pipeline should be declared in Dagger and not in GHA. 
I suggest to create a new command like:
`airbyte-ci metadata orchestrator release` 
You can declare create a dagger container replicating the Dockerfile content and then use the `publish` function to push it to the docker registry.

Once the image is pushed to docker I suggest you create a new environment using the python base environment with the `dagster-cloud` python packaged installed. And then exec: ` dagster-cloud workspace add-location YOUR_LOCATION_NAME \
    --image LOCATION_IMAGE \
    --package-name PACKAGE_NAME `

cpdeethree:
> Oh! I didn't know dagster required a docker image. I would have thought they could run the python directly

I'm not 100% on this, but I think it can do both. It just so happens that our hybrid deployment uses the docker image approach

bnchrch:
@alafanechere This is back up for review with some key differences.

1. We are no longer targeting Dagster Hybrid as a deployment. Instead we are using Dagster Cloud (serverless)
2. To do deployments to Dagster Cloud in a simple fashion we needed a setup.py file. I didnt like the idea of having a setup.py file that was only used for Dagster Cloud deployments, so we generate it during deploy, then remove it.
3. I would have loved to continue using our Dockerfile but to do so we would have had to do a lot of complicated steps. None of them worthwhile
4. But! by ditching the dockerfile, we are using [Dagsters PEX deployments which are reportedly alot faster](https://dagster.io/blog/fast-deploys-with-pex-and-docker)
5. We now use **Dagger Pipelines** to deploy

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/metadata.py
```diff
@@ -56,6 +57,20 @@ def validate(ctx: click.Context):
     )
 
 
+@metadata.command(cls=DaggerPipelineCommand, help="Deploy the metadata service orchestrator to production")
+@click.pass_context
+def deploy_orchestrator(ctx: click.Context):
```
alafanechere, line 62: What do you think about creating an `orchestrator` command group to have a command like
`airbyte-ci metadata deploy orchestrator` 

If we decide to release the lib I can easily see something like this coming
`airbyte-ci metadata deploy lib`

.github/workflows/metadata_service_deploy_orchestrator_dagger.yml
```diff
@@ -0,0 +1,31 @@
...
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    paths:
+      - "airbyte-ci/connectors/metadata_service/orchestrator/**"
+jobs:
+  connector_metadata_service_orchestrator_deploy:
```
alafanechere, line 11: consistency nit
```suggestion
  connector_metadata_service_deploy_orchestrator:
```

.github/workflows/metadata_service_deploy_orchestrator_dagger.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - name: Checkout Airbyte
+        uses: actions/checkout@v2
+      - name: Run test for the metadata orchestrator
+        id: metadata-orchestrator-test-pipeline
+        uses: ./.github/actions/run-dagger-pipeline
+        with:
+          subcommand: "metadata test orchestrator"
+          context: "pull_request"
+      - name: Run test for the metadata orchestrator
```
alafanechere, line 25:
```suggestion
      - name: Deploy the metadata orchestrator
```

.github/workflows/metadata_service_deploy_orchestrator_dagger.yml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  connector_metadata_service_orchestrator_deploy:
+    name: Connector metadata service deploy orchestrator
+    runs-on: medium-runner
+    env:
+      CI_GITHUB_ACCESS_TOKEN: ${{ secrets.GH_PAT_MAINTENANCE_OCTAVIA }}
+    steps:
+      - name: Checkout Airbyte
+        uses: actions/checkout@v2
+      - name: Run test for the metadata orchestrator
```
alafanechere, line 19: If `test` is a `deploy` dependency I think we should declare this dependency directly in the dagger pipeline. This is to follow the philosophy of centralizing the CI logic.

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -38,6 +39,29 @@ async def _run(self) -> StepStatus:
...
+        python_with_dependencies = with_pip_packages(python_base, ["dagster-cloud==1.2.6", "poetry2setup"])
+
+        container_to_run = (
+            python_with_dependencies.with_mounted_directory("/src", self.parent_dir)
+            .with_workdir(f"/src/{self.module_path}")
+            .with_env_variable("DAGSTER_CLOUD_API_TOKEN", self.dagster_cloud_api_token)
+            .with_entrypoint(["bash"])
+            .with_exec(["deploy_to_prod.sh"])
+        )
+        # poetry_run_exec = self.poetry_run_container.with_exec(poetry_run_args)
```
alafanechere, line 61: I think you forgot to remove this comment.

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -38,6 +39,29 @@ async def _run(self) -> StepStatus:
...
+    async def _run(self) -> StepStatus:
+        python_base = with_python_base(self.context)
+        python_with_dependencies = with_pip_packages(python_base, ["dagster-cloud==1.2.6", "poetry2setup"])
+
+        container_to_run = (
+            python_with_dependencies.with_mounted_directory("/src", self.parent_dir)
+            .with_workdir(f"/src/{self.module_path}")
+            .with_env_variable("DAGSTER_CLOUD_API_TOKEN", self.dagster_cloud_api_token)
+            .with_entrypoint(["bash"])
+            .with_exec(["deploy_to_prod.sh"])
```
alafanechere, line 59: Interesting: I understand that you want to keep this script under the `metadata_service` but I think for long run maintainability we should avoid writing bash scripts. 
Do you mind splitting up the script in `with_exec` steps?

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -38,6 +39,29 @@ async def _run(self) -> StepStatus:
...
+        self.dagster_cloud_api_token = dagster_cloud_api_token
+
+    async def _run(self) -> StepStatus:
+        python_base = with_python_base(self.context)
+        python_with_dependencies = with_pip_packages(python_base, ["dagster-cloud==1.2.6", "poetry2setup"])
+
+        container_to_run = (
+            python_with_dependencies.with_mounted_directory("/src", self.parent_dir)
+            .with_workdir(f"/src/{self.module_path}")
+            .with_env_variable("DAGSTER_CLOUD_API_TOKEN", self.dagster_cloud_api_token)
```
alafanechere, line 57:
```suggestion
            .with_secret_variable("DAGSTER_CLOUD_API_TOKEN", the_token_loaded_in_a_secret_object)
```
Please try to use [`with_secret_variable`](https://dagger-io.readthedocs.io/en/sdk-python-v0.5.0/api.html#dagger.api.gen.Container.with_secret_variable). Here's an [example](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py#L193) in the connectors pipelines.

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -132,3 +156,42 @@ async def run_metadata_orchestrator_test_pipeline(
...
+    git_branch: str,
+    git_revision: str,
+    gha_workflow_run_url: Optional[str],
+    pipeline_start_timestamp: Optional[int],
+    ci_context: Optional[str],
+) -> bool:
+    # Get the DAGSTER_CLOUD_API_TOKEN from the environment
+    dagster_cloud_api_token = os.environ.get("DAGSTER_CLOUD_API_TOKEN")
+    if not dagster_cloud_api_token:
+        raise ValueError("DAGSTER_CLOUD_API_TOKEN not found in environment")
```
alafanechere, line 172: I think this validation should happen earlier, at the command level.
You can pass the token as a required command option with an `envvar` default to `DAGSTER_CLOUD_API_TOKEN`. If you make the option required and the envvar is missing I think click will raise an error.

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -38,6 +39,29 @@ async def _run(self) -> StepStatus:
         return await super()._run(["metadata_service", "validate", "metadata.yaml"])
 
 
+class DeployOrchestratorStep(Step):
```
alafanechere, line 42:
```suggestion
class DeployOrchestrator(Step):
```
nit: I'm not sure we've been consistent in the step naming, but I find if redundant to add a Step suffix when the class explicitely inherits from Step

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -132,3 +156,42 @@ async def run_metadata_orchestrator_test_pipeline(
...
+        ci_context=ci_context,
+    )
+
+    async with dagger.Connection(DAGGER_CONFIG) as dagger_client:
+        metadata_pipeline_context.dagger_client = dagger_client.pipeline(metadata_pipeline_context.pipeline_name)
+
+        async with metadata_pipeline_context:
+            deploy_orch_step = DeployOrchestratorStep(
+                context=metadata_pipeline_context,
+                title="Deploy Metadata Orchestrator to Dagster Cloud",
```
alafanechere, line 190: As this title is static I think you should declare it as a class variable constant, not with a constructor parameter.

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -132,3 +156,42 @@ async def run_metadata_orchestrator_test_pipeline(
...
+                parent_dir_path=METADATA_DIR,
+                module_path=METADATA_ORCHESTRATOR_MODULE_PATH,
```
alafanechere, line 192: Same as previous comment: these parameters are static and can be declared at with class variable constants.

airbyte-ci/connectors/metadata_service/orchestrator/deploy_to_prod.sh
```diff
@@ -0,0 +1,14 @@
...
+dagster-cloud serverless deploy-python-executable \
+--location-name metadata_service_orchestrator \
+--location-file dagster_cloud.yaml \
+--organization airbyte-connectors \
+--deployment prod \
+--api-token $DAGSTER_CLOUD_API_TOKEN \
+--python-version 3.9
+
+# Remove the temporary setup.py file
+rm setup.py
```
alafanechere, line 14: Missing new line at the end of the file.

alafanechere (changes requested):
My two main concerns are the following:
- I think we should try to avoid bash scripts as much as possible as their use so far as led to discoverability problems to understand what a pipeline is actually doing
- If you want to make `deploy` depend on successful `test` I think this should be declared at the dagger pipeline level and not at the GHA workflow level. This will help us use a single GHA workflow for everything in the future.

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/metadata.py
bnchrch: Great ideas!

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
bnchrch: 🤦 Its in a commit I did not push

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
bnchrch: Can do!

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
bnchrch: Will change :)

However im going to put my reasoning on this naming out there just to context share for fun!

With my background being on the functional side, I find classes hard to understand at a glance. You never quite know whats going on until youve read the class, its subclasses, then trace the method you call. 

This is in contrast to functions where the types tell you what happened at a glance, this data type turned into this datatype can tell a pretty good story as you trace through code.

So I always find myself adding Parent Class names to child classes so I get the same glance when reading code.

"Oh this has step in the name, it probably subclasses Step"

Anyway, this is just a side effect of me coming from one style of programming into another.

Thus, TLDR, I will change the name here

alafanechere: IMO the context in which these classes are instantiated helps you infer what's likely there parents. To be honest I don't have a strong opinion and if it makes you more comfortable with the code base feel free to keep it!

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -212,3 +262,34 @@ async def run_metadata_upload_pipeline(
...
+
+    async with dagger.Connection(DAGGER_CONFIG) as dagger_client:
+        metadata_pipeline_context.dagger_client = dagger_client.pipeline(metadata_pipeline_context.pipeline_name)
+
+        async with metadata_pipeline_context:
+            steps = [
+                TestOrchestrator(context=metadata_pipeline_context),
+                DeployOrchestrator(context=metadata_pipeline_context)
+            ]
+            steps_results = await run_steps(steps)
```
bnchrch, line 293: @alafanechere I addressed your comments and this is back up and ready for review.

Though there is one net new function added that you might recognize from [here](https://github.com/airbytehq/airbyte/pull/24225#discussion_r1162064279).

I added it as the logic was useful for "dont run next steps" and should be able to be moved into the pipeline class once the pipeline also contains the steps.

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -212,3 +262,34 @@ async def run_metadata_upload_pipeline(
...
+    )
+
+    async with dagger.Connection(DAGGER_CONFIG) as dagger_client:
+        metadata_pipeline_context.dagger_client = dagger_client.pipeline(metadata_pipeline_context.pipeline_name)
+
+        async with metadata_pipeline_context:
+            steps = [
+                TestOrchestrator(context=metadata_pipeline_context),
+                DeployOrchestrator(context=metadata_pipeline_context)
+            ]
```
alafanechere, line 292: I'm wondering if we shall make `DeployOrchestrator` depend on `TestOrchestrator` success explicitely or make it depend on the full orchestrator testing pipeline by calling `run_metadata_orchestrator_test_pipeline` and checking the success of the pipeline before running `DeployOrchestrator`.

I feel like the later is more maintainable: if  we add more tests or steps to `run_metadata_orchestrator_test_pipeline`  we won't have to update `run_metadata_orchestrator_deploy_pipeline`.

The current problem is that `run_metadata_orchestrator_deploy_pipeline` and `run_metadata_orchestrator_test_pipeline` will instantiate two differents dagger clients. Instantiating two different clients can lead to strange edge cases as two different client might have a different perceptions of the filesystem state if file changes between the two instantiation. 
I think it's not risky at the moment, but we might want to pass the dagger client directly to the `run_metadata_orchestrator_*` functions.

What do you think?

alafanechere (approved):
I'm approving because the concerns I originally have are solved! Thank you!
I added more suggestions, related to `run_steps` and how to express the dependency of Deploy on Test. I'm curious about your opinion.

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
bnchrch: I like this, but perhaps we do this during our larger refactor when we might tackle the idea of steps are pipelines.

alafanechere: Sounds good to me

PR #24840, opened 2023-04-04
Add validate metadata step to CICD with Dagger

bnchrch:
## What
Screenshot 2023-04-04 at 6 23 07 AM
We need to validate metadata.yml files for all connectors
closes #24018 
## How
This adds a dagger pipeline that runs the metadata service validate function on all changed connectors

## Notes for reviewer
1. This makes use of [pipx](https://github.com/pypa/pipx) to run the metadata-service cli command. Its pretty awesome and might be useful for other tools!

## BEFORE MERGE

- [x] Ensure that merging this wont interfere with merging in connector related PRs

PR #24964, opened 2023-04-06
create metadata files for all connectors

### LINKED ISSUE #24848: generate metadata files and commit them to the airbyte repo
## What this is
1. Updating the `persist_metadata_definitions` to write to the proper file structure `/tmp/connectors/<connector-name>/metadata.yml
2. Moving these files to their respective locations in the airbyte-repo (maybe `persist_metadata_definitions` can do it directly 😉)

## What you should note
1. Double check the cloud diff job to see if you need to modify the generation and override of the metadata files

2. There are some net-new thirdparty connectors that will only be a folder with a single metadata file in it. ([Slack conversation on this](https://airbytehq-team.slack.com/archives/C03VDJ4FMJB/p1679432433101039))

### More context on third party connectors
 From Ella in [Slack](https://airbytehq-team.slack.com/archives/C03VDJ4FMJB/p1679499521132149?thread_ts=1679432433.101039&cid=C03VDJ4FMJB):

Here is what I think I'd naturally imagine for how to keep these metadata:
```
airbyte-integrations
  destination-a
    metadata.yml
  destionation-b
    metadata.yml
  source-a
    metadata.yml
  source-b
    metadata.yml
  external
    farosai
      airbyte-pager-duty-source
        metadata.yml
      airbyte-harness-source
        metadata.yml
    some-other-vendor
      source-custom
        metadata.yml
      thiscouldbenamedinanyway
        metadata.yml
```
It is basically evan's #1 suggestion, but without having to change the path of our existing connectors. I'd also be fine with implementing evan's number 1 suggestion. I think option 2 is not great because externals can use whatever naming convention they like. For example, faros's connector's docker images (not repository) all start with airbyte-  which is confusing to say the least - I think we want to subfolder them instead of mixing them in

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/24848
I ran the `persist_metadata_definition` Dagster asset to generate these files.


evantahler (commented):
It's Happening 🥳

bnchrch:
@alafanechere Before we merge lets make sure this PR is in
https://github.com/airbytehq/airbyte/pull/24840

bnchrch (changes requested):
@alafanechere Looking good!

I think we should do the following  adjustments
1. Move any non airbyte connectors under `connectors/third-party/<AUTHOR>/<CONNECTOR>`. That should currently apply to `farosai` and `ghcr.io`
2. Remove the `devmate-cloud` subfolder of `ghcr.io`
3. Rename the `ghcr.io` folder to `ghcr`

airbyte-integrations/connectors/source-mongodb-v2/metadata.yaml
```diff
@@ -0,0 +1,20 @@
+data:
+  catalogs:
+    cloud:
+      dockerImageTag: 0.1.7
+      dockerRepository: airbyte/source-mongodb-strict-encrypt
```
bnchrch, line 5: @alafanechere Also I think for strict encrypt we need to manually add a metadata.yml file for each.

These metadata files should
1. Have both cloud and oss catalogs disabled
2. Contain enough information to build the correct docker image
3. Have a comment acknowledging that the get deployed by their non strict encrypt sibling

alafanechere:
@bnchrch I implemented your suggestions:
> Move any non airbyte connectors under connectors/third-party/<AUTHOR>/<CONNECTOR>. That should currently apply to farosai and ghcr.io

✅ 

> Rename the ghcr.io folder to ghcr

✅ 
> Remove the devmate-cloud subfolder of ghcr.io

✅ - The `ghcr` directory is now empty, is this expected @bnchrch?

> I think for strict encrypt we need to manually add a metadata.yml file for each.

I added metadata.yaml files for strict encrypt variants. They are all disabled from cloud and oss.

bnchrch (approved):
### On the note of GHCR
> ✅ - The ghcr directory is now empty, is this expected @bnchrch?

@alafanechere Oh whoops! I couldve been more clear when I said

> Remove the devmate-cloud subfolder of ghcr.io

We had the following file path in the last commit
`connectors/ghcr.io/devmate-cloud/source-something/metadata.yml`

And I was thinking it should have been
`connectors/third-party/ghcr/source-something/metadata.yml`

### On metadata validation action
I noticed our validation action isnt running. Simply because yml != yaml

https://github.com/airbytehq/airbyte/blob/8da6c4ae495b7a745b2d7b4d36f03e7a65b09fe2/.github/workflows/metadata_validate_manifest_dagger.yml

Would you be able to rename these to end in .yml just so we can check that the validation works? We can add a better check to see if its a yaml extension later.

## conclusion

These are all minor so ill approve now so when your back online if you address correct those two small things then feel free to merge!

alafanechere:
@bnchrch I
- I re-added the deleted connector in `ghrc`
- I fixed the metadata not passing validation
- I updated the model to accept `message_queue` `connectorSubType` (do you think this should be done in a separate PR?)
- The workflow to run validation was quite bulky and sometime stuck on this PR. I tweaked it a bit to make it faster and reuse existing components [here](https://github.com/airbytehq/airbyte/pull/25107).

alafanechere:
> what led you to adding the subtype?

destination-mqtt uses the `message_queue` subtype

> Were we able to do this in the overrides in the dev assets?

I did this manually . I don't think we can do it in the override for all the cases:
* Missing release stages: we can fallback to alpha
* Missing subtype: We could fallback to unknown, but I think it's better to curate these  connector subtype properly now as there was only a few.

bnchrch:
> > what led you to adding the subtype?
> 
> destination-mqtt uses the `message_queue` subtype
> 
> > Were we able to do this in the overrides in the dev assets?
> 
> I did this manually . I don't think we can do it in the override for all the cases:
> 
> * Missing release stages: we can fallback to alpha
> * Missing subtype: We could fallback to unknown, but I think it's better to curate these  connector subtype properly now as there was only a few.

Hmm no big deal but if I understand correctly you should be able to override releaseStage and subType for any connector by its connectorId [here](https://github.com/airbytehq/airbyte/blob/augustin/metadata-service/upload-metadata-files-to-gcs/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/dev.py#L136)

alafanechere:
The Dagger CI infra is currently down. I ran the metadata validation locally with success. Here are the [logs](https://gist.github.com/alafanechere/41c970ed9d5a99c15f5474e7be1ca7dd). Ready to merge.

PR #24965, opened 2023-04-06
Add `planetscale/source-airbyte`

evantahler:
Things to fix:

* The naming convention for the docker image doesn't match.  Ideally, it would be called `source-planetscale`
* This image doesn't build ARM docker images
* `additionalProperties` needs to be true in the spec - this is why CAT currently fails


nickvanw:
Hey @evantahler! We at PlanetScale are super excited about this - let me know if there's any help that y'all need to get this merged!

PR #24851, opened 2023-04-04
🎉 Source JDBC: changed the column type info log to debug log

arditarapi:
## What
On SQL integrations (MySQL, Postgres, BigQuery, ..etc), we see many log entries that follow this pattern:
Table {} column {} (type {}[{}], nullable {}) -> {}
Table _family_file column rfam_acc (type VARCHAR[7], nullable false) -> JsonSchemaType({type=string})
In big schemas these can go more than 3000 lines.  
Screenshot 2023-04-04 at 5 30 41 PM

## How
Changed the info to debug log

PR #24969, opened 2023-04-06
Add versioning, indexed fields to db source connector logging

nguyenaiden:
## What
Issue: [24794](https://app.zenhub.com/workspaces/db--dw-source-connectors-6333360e0a41155061efbcbd/issues/gh/airbytehq/airbyte/24794)
Currently, we don't have any visibility into:

- The version of a DB a source connector is running against (e.g. Postgres 9.5.x)
- Current indices on the source schema.

## How
In the `createDatabase` method `AbstractJdbcSource` class, the source product type, version, and all of its relevant index and column information are extracted via the `JdbcDatabase` object and logged accordingly. 
**_Note_**: Compound indexes will show up as its own log line but with the same index name. See screenshot below:
image


airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java
```diff
@@ -410,6 +414,29 @@ public JdbcDatabase createDatabase(final JsonNode sourceConfig) throws SQLExcept
...
+    while (tables.next()) {
+      final String tableName = tables.getString(JDBC_COLUMN_TABLE_NAME);
+      final ResultSet indexInfo = database.getMetaData().getIndexInfo(null,
+          null,
+          tableName,
+          false,
+          false);
+      LOGGER.info("Discovering indexes for table \"{}\"", tableName);
+      while (indexInfo.next()) {
+        LOGGER.info("Index name: {}, Column: {}, Unique: {}",
```
marcosmarxm, line 433: This should be a info or debug?

marcosmarxm: Saying because there is some suggestion from community to remove some excessive info logs from databases https://github.com/airbytehq/airbyte/pull/24851

nguyenaiden: No strong feelings for either for me. More than happy to upgrade it to `Debug` for this information to stand out during the logs. Thoughts? @rodireich, @akashkulk?

rodireich:
nit: adding "closes ##### (issue number)" would link this PR to the issue which will be marked as closed upon merging.
the words "resolves" and "closes" are parsed I think

airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java
rodireich: My thinking is that if this log is not overly verbose and can help tackling customer cases - especially cloud where turning debug logging on is not trivial, then it should be `INFO`

akashkulk: +1 on that. This is not too verbose (1 log per stream + 1 overall) and will greatly help in debugging

airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java
```diff
@@ -410,6 +414,29 @@ public JdbcDatabase createDatabase(final JsonNode sourceConfig) throws SQLExcept
     quoteString = (quoteString == null ? database.getMetaData().getIdentifierQuoteString() : quoteString);
     database.setSourceConfig(sourceConfig);
     database.setDatabaseConfig(jdbcConfig);
+    LOGGER.info("Data source product recognized as {}:{}",
+        database.getMetaData().getDatabaseProductName(),
+        database.getMetaData().getDatabaseProductVersion());
+    final ResultSet tables = database.getMetaData().getTables(null,
```
akashkulk, line 420: Can we abstract all of this into a helper method somewhere? the logging of indexes is super complicated

airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java
```diff
@@ -410,6 +414,29 @@ public JdbcDatabase createDatabase(final JsonNode sourceConfig) throws SQLExcept
...
+        null,
+        "%",
```
akashkulk, line 422: what does this parameter indicate?

airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java
```diff
@@ -410,6 +414,29 @@ public JdbcDatabase createDatabase(final JsonNode sourceConfig) throws SQLExcept
...
+      final String tableName = tables.getString(JDBC_COLUMN_TABLE_NAME);
+      final ResultSet indexInfo = database.getMetaData().getIndexInfo(null,
+          null,
+          tableName,
+          false,
+          false);
+      LOGGER.info("Discovering indexes for table \"{}\"", tableName);
+      while (indexInfo.next()) {
+        LOGGER.info("Index name: {}, Column: {}, Unique: {}",
+            indexInfo.getString(JDBC_INDEX_NAME),
```
akashkulk, line 434: Unfortunately, since JDBC drivers differ quite a bit in how they implement methods, you might have do null checks every time you're checking a field. see : `AbstractJdbcSource.getColumnMetadata` for an example : https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java

Alternatively, you can try running all of the integration tests for the sources that inheirit from this class to see if it leads to an NPE since this codepath is covered in every integration test

nguyenaiden: Got it. Good call. I'm still manually testing MySQL, but I'll run integration tests for less used connectors.

airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java
nguyenaiden: That argument position refers to the table name pattern, so `%` refers to everything (wildcard)

airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java
nguyenaiden: Got you. I was debating that as well but went against it since we're only using it here. But i'll put it in its own helper method so that it's easier to read and I won't have to leave code comments in the `createDatabase` method.

airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java
arditarapi: In big Schemas this becomes very big, i have cases with more than 3000 lines so it completely blocks the log window.

akashkulk: Fair point. Let's do this : 
1. Log Db version once here in `createDatabase`
2. Move logging indices to the `read()` version, since we have access to the configured catalog there. Probably close to where we are  logging schema change/validating cursor fields. 

Fine to do in a separate PR as well.

airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java
```diff
@@ -99,6 +103,47 @@ public AbstractJdbcSource(final String driverClass,
     this.sourceOperations = sourceOperations;
   }
 
+  @Override
+  public AutoCloseableIterator<AirbyteMessage> read(final JsonNode config,
```
akashkulk, line 107: I don't think we should be overriding the `read()` method here. We can add a generic logDatabaseData() method to the parent AbstractDbSource class, which has a default impl to not do anything. Then, we can override it in AbstractJdbcSource

nguyenaiden: Got it. Followed suggested pattern.

nguyenaiden:
Closing per https://github.com/airbytehq/airbyte/pull/25345

PR #24972, opened 2023-04-06
run-ab-platform.sh uses docker compose v2

gosusnp:
## What

The `run-ab-platform.sh` is checking docker compose v1 syntax, since our `docker-compose.yaml` leverages v2 feature, we should explicitly check for v2.

## How

Check and use docker compose v2

## 🚨 User Impact 🚨

We'll fail explicitly if docker compose v2 isn't found.

### DIFF
--- a/run-ab-platform.sh
+++ b/run-ab-platform.sh
@@ -134,8 +134,9 @@ if test $(tput cols) -ge 64; then
 fi
 
 ########## Dependency Check ##########
-if ! which -s docker-compose; then
-  echo "$red_text""docker compose not found! please install docker compose!""$default_text"
+if ! docker compose version >/dev/null 2>/dev/null; then
+  echo -e "$red_text""docker compose v2 not found! please install docker compose!""$default_text"
+  exit 1
 fi
 
 Download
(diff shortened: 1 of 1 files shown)


run-ab-platform.sh
```diff
@@ -134,8 +134,9 @@ if test $(tput cols) -ge 64; then
 fi
 
 ########## Dependency Check ##########
-if ! which -s docker-compose; then
-  echo "$red_text""docker compose not found! please install docker compose!""$default_text"
+if ! docker compose version >/dev/null 2>/dev/null; then
```
supertopher, line 137: I spent a pretty long time writing why I think this should be written as 
```bash
docker compose version > /dev/null 2>&1
```
but I think the way I have written this historically is more cryptic than yours.  You method seems easier to understand than me assuming my reader understand the incantation `2>&1` which throws `STDERR` into `STDOUT` which then makes `>` of `STDOUT` work for both streams.  

no changes requested, but I thought about this too much to not say something
¯\\_(ツ)_/¯

supertopher (approved):
This looks great Jimmy, thanks for updating the version and the protection block

c-p-b (approved):

run-ab-platform.sh
gosusnp: Yeah, I usually get less questions without the `2>&1`, happy to go either way, I don't have a strong preference.

PR #24974, opened 2023-04-06
low-code: Remove NoAuth from the interface

girarda:
## What
The `NoAuth` object does not need to be exposed in the DSL because the HttpRequester's authenticator field is optional.

This removes the `NoAuth` from the manifest schema so we don't need to document and maintain it.

The `NoAuth` object is only used explicitly once for source-alpha-vantage

## How
- Remove `NoAuth` from the definitions
- Update the model to component factory
- Update source-alpha-vantage to implicitly use no auth

## Recommended reading order
2. `airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py`
3. `airbyte-integrations/connectors/source-alpha-vantage/source_alpha_vantage/manifest.yaml`

## Testing
Tested locally because we can't run CATs with the local version of the CDK using CI.
I tested the following:
- `./airbyte-cdk/python/bin/run-cats-with-local-cdk.sh -c source-alpha-vantage`
- `./airbyte-cdk/python/bin/run-cats-with-local-cdk.sh -c source-gnews`
- `./airbyte-cdk/python/bin/run-cats-with-local-cdk.sh -c source-greenhouse`

Dagger CI pipeline is failing because it doesn't test with local changes to the CDK

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -678,7 +678,6 @@ definitions:
           - "$ref": "#/definitions/CustomAuthenticator"
           - "$ref": "#/definitions/OAuthAuthenticator"
           - "$ref": "#/definitions/SingleUseRefreshTokenOAuthAuthenticator"
-          - "$ref": "#/definitions/NoAuth"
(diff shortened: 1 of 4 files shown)


brianjlai (approved):
:shipit: Do we need a follow up to deprecate NoPagination? It's trivial fix but way more usage than `NoAuth` which appears to be just alpha vantage

sherifnada:
does this impact the UI?

PR #24976, opened 2023-04-07
low-code: remove mentions of NoPagination from existing manifests

girarda:
## What
The NoPagination object does not need to be exposed in the DSL because the SimpleRetriever's paginator field is optional.

This PR removes explicit mentions of `NoPagination` from the existing manifests. Removing the field from the schema will be done in a follow up PR.

This PR also removes the field from the template, which explains why it is so widely used.

Removing the explicit mention of NoPagination does not change the connector's behavior, so I do not plan on releasing a new version.

Tested with:
- `./airbyte-cdk/python/bin/run-cats-with-local-cdk.sh -c source-alpha-vantage`
-


sherifnada:
@girarda are we still trying to merge this?

PR #24994, opened 2023-04-07
Publish stream status messages in CDK

jdpgrailsdev:
## What
* Report transitions in status while reading a source

## How

Publishes the stream status message on the various transitions in status for each stream in the source.  This is a parallel PR to https://github.com/airbytehq/airbyte/pull/24971.

## Recommended reading order
1. `abstract_source.py`
2. `stream_status_utils.py`


jdpgrailsdev:
This PR has been updated to use the new protocol message and is now ready for review.

airbyte-cdk/python/setup.py
```diff
@@ -46,7 +46,7 @@
     packages=find_packages(exclude=("unit_tests",)),
     package_data={"airbyte_cdk": ["py.typed", "sources/declarative/declarative_component_schema.yaml"]},
     install_requires=[
-        "airbyte-protocol-models==1.0.0",
+        "airbyte-protocol-models==0.3.5",
```
evantahler, line 49: This _looks_ like a downgrade... but I don't think we ever were on v1.0.0?

jdpgrailsdev: Correct.  We were never on 1.0.0 even though we set that as the dependency version.  The new changes were released as a patch on the current version (0.3), which is the reason for the correction.

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -336,3 +344,9 @@ def _get_message(self, record_data_or_message: Union[StreamData, AirbyteMessage]
             return record_data_or_message
         else:
             return stream_data_to_airbyte_message(stream.name, record_data_or_message, stream.transformer, stream.get_json_schema())
+
+    def _emit_stream_status(self, stream: ConfiguredAirbyteStream, stream_status: AirbyteStreamStatus, successful: bool):
+        """
+        Emits a new AirbyteStreamStatusTraceMessage
+        """
+        print(stream_status_as_airbyte_message(stream, AirbyteStreamStatus.RUNNING, None).json(exclude_unset=True))
```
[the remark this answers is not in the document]
jdpgrailsdev: @girada Mainly ignorance/copy paste from other parts in the code that I thought was producing a message.  Will fix per your suggestion.

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -113,17 +115,20 @@ def read(
...
                         configured_stream=configured_stream,
                         state_manager=state_manager,
                         internal_config=internal_config,
                     )
+                    self._emit_stream_status(configured_stream, AirbyteStreamStatus.STOPPED, True)
                 except AirbyteTracedException as e:
                     raise e
                 except Exception as e:
                     logger.exception(f"Encountered an exception while reading stream {configured_stream.stream.name}")
+                    self._emit_stream_status(configured_stream, AirbyteStreamStatus.STOPPED, False)
```
girarda, line 131: this message won't be yielded.

Option 1:
1. `_emit_stream_status` returns `stream_status_as_airbyte_message(stream, stream_status, successful)`
2. change these lines to yield the output of `_emit_stream_status` (`yield self._emit_stream_status(configured_stream, AirbyteStreamStatus.STOPPED, True)`)

Option 2:
change this line to `stream_status_as_airbyte_message(configured_stream, AirbyteStreamStatus.STOPPED, True)` and forget about the `_emit_stream_status` method

girarda (changes requested):

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
jdpgrailsdev: @girarda Do you have a preference between the two options?  I'm inclined to implement option 1, unless you feel strongly about this.

girarda: sounds good to me @jdpgrailsdev

jdpgrailsdev: > sounds good to me @jdpgrailsdev

@girarda I didn't really understand the difference between options 1 and 2.  After looking at the code again, I went with option 2, as the method really didn't add any value anymore.

airbyte-cdk/python/airbyte_cdk/utils/stream_status_utils.py
```diff
@@ -0,0 +1,29 @@
...
+#
+
+
+from datetime import datetime
+
+from airbyte_cdk.models import AirbyteMessage, AirbyteStreamStatusTraceMessage, AirbyteTraceMessage, StreamDescriptor, TraceType
+from airbyte_cdk.models import Type as MessageType
+
+
+def as_airbyte_message(stream, current_status, successful) -> AirbyteMessage:
```
girarda, line 12: nit: it would be useful to add type annotations
`def as_airbyte_message(stream: Stream, current_status: AirbyteStreamStatus, successful: Optional[bool]) -> AirbyteMessage`

girarda (commented):
the change looks good, but some unit tests will need to be updated to reflect the new messages that are published https://github.com/airbytehq/airbyte/actions/runs/4743777368/jobs/8423891418?pr=24994

airbyte-cdk/python/airbyte_cdk/utils/stream_status_utils.py
jdpgrailsdev: Done.

jdpgrailsdev:
> the change looks good, but some unit tests will need to be updated to reflect the new messages that are published https://github.com/airbytehq/airbyte/actions/runs/4743777368/jobs/8423891418?pr=24994

@girarda I don't understand why this test is now failing.  All tests were passing before I merged in changes from master so this feels like an external change to my PR that is now failing the tests.  Any advice/pointers on what I should change to get this passing would be appreciated.

jdpgrailsdev:
> As an example, this list of expected messages will need to be updated

Interesting.  Why wasn't this failing previously?  Was it because I wasn't actually yielding/publishing the message?  Trying to understand why it is now failing and didn't fail previously on this PR.

airbyte-cdk/python/unit_tests/sources/test_abstract_source.py
```diff
@@ -78,7 +83,7 @@ class MockStreamOverridesStateMethod(Stream, IncrementalMixin):
     primary_key = None
     cursor_field = "updated_at"
     _cursor_value = ""
-    start_date = "1984-12-12"
+    start_date = "19-12-12"
```
girarda, line 86: this date looks weird. Why is the year changed to `19`?

jdpgrailsdev: Definitely not intentionally.  Will revert this.

girarda (approved):
:shipit:

jdpgrailsdev:
Per a discussion with @alafanechere I have reverted the changelog/version bump to the source-facebook-marketing connector to avoid an order of operation issue between updating and publishing the CDK and publishing the connector.  This PR will be merged first and once the CDK has been published, a second PR will be created to publish the source-facebook-marketing connector.

PR #25002, opened 2023-04-07
Override the github threshold

benmoriceau:
## What
Increase the threshold of the maximum seconds between messages for the github api. The github documentation mention that the rate limitation is based on an hourly period, switching from a 1h to a 1h30 limit should limit the amout of termination that were not needed


davinchia (approved):

tryangul (approved):

jdpgrailsdev (approved):
:shipit:

Issue #23975, opened 2023-03-13
connectors-ci: split `ci-connector-ops` in multiple packages.


alafanechere:
## Tell us about the problem you're trying to solve
We originally stuffed a lot of different tooling in the ci-connector-ops package.
* QA checks
* QA engine
* CI pipelines declared with Dagger
* Acceptance test configuration validation (high test strictness level)
* Allowed host checks

It has the following problems:
* The tools do not have the same requirements in terms of python version (dagger pipeline requires Python 3.10). It means we have to rely "[extras_requires](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/setup.py#L58)" to make sure `dagger-io` (requiring Python 3.10) is only installed when we install the "pipelines" extra. (e.g pip install .\[pipelines\])
* Some tools require libraries that are not light and always easy to containerize due to their dependency to some system C system dependencies (`pandas`) that are not always available on python alpine image. But we need to containerize this tool in the CI context. Again, playing with `extra_requires` in `setup.py` is the [workaround](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/setup.py#L58) so far. 
* We also have other ci_* packages (`ci_common_utils`, `ci_credentials`) which are not in `ci_connector_ops`, which makes the `tools` directory structure no clear at all.

## Suggestion:
* Follow's Ben suggestion to move these Python packages under an airbyte-ci top directory: 

![image](https://user-images.githubusercontent.com/5551758/224691530-4e80cc59-ab21-4304-8766-501438a5aa80.png)

@bnchrch I might challenge the legitimacy of a `common` / `connectors_ci` directory.

I think we should have something like:
- ci-credentials: to download/upload secrets to GSM
- ci-code-validator: perform static code analysis (QA checks, acceptance-test-config checks, allowed host checks)
- ci-qa-engine: generate QA report for all our connectors in our catalog + opens PR on the internal repo to promote
- ci-pipelines: holds our CI pipelines logic, and can call the containerized packages listed above.

bnchrch:
Notes from a huddle with @alafanechere 

### What we decided
1. projects/libraries/services in `airbyte-ci` should be responsible for their own dependencies
2. We should be able to import between these systems and have the dependencies resolve automatically
3. This is a good oppourtunity to start using Poetry
4. But the first phase should be get the directory structure and imports working before we do any refactoring of dependency management for existing projects.


### Outstanding questions
[...]

PR #25044, opened 2023-04-11
airbyte-ci: orchestrate the upload of modified metadata.yaml files to GCS

### LINKED ISSUE #24850: Upload metadata.yaml to GCS bucket on merge to master
## What this is
On merge to master
1. Identify any new metadata files
2. Check if the file already exist in GCS (/connector/v1.0.0/metadata.yaml and /connector/latest/metadata.yaml)
3. If not upload them to their respective GCS paths (latest/version).

The reason not to re-upload an existing file is to not mess with updated-at timestamps

## Things to note
1. We want this to be done in a similar fashion to how the validate click command works

alafanechere:
## What
Closes #24850
When a `metadata.yaml` file is modified in a commit on the `master` branch we want to upload this file to a GCS bucket.

## How
1. Expose a new CLI command: 
```bash

# The gcs-credentials option defaults to the GCS_CREDENTIALS env var
# This will upload all the modified metadata files in the last commit on master
airbyte-ci  metadata upload <target-bucket-name> --gcs-credentials <json service account>
# This will upload all the connector metadata files - useful for initial provisionning
airbyte-ci  metadata upload <target-bucket-name> --all
```

2. Add a util function to list modified file in a commit (we only had a branch diff, not a commit diff on a branch):
https://github.com/airbytehq/airbyte/blob/05fa4d87656318e186ad6ddb50f9f5ac93fa03e8/tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py#L188

4. Implement a new pipeline function that performs parallel upload of modified metadata files:
https://github.com/airbytehq/airbyte/blob/1545009b56364acc604a84bc2f736faa73d874b9/tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py#L180

5. Implement a `MetadataUpload` step: https://github.com/airbytehq/airbyte/blob/1545009b56364acc604a84bc2f736faa73d874b9/tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py#L55

I also consolidate the logic for `MetadataValidation`

## 🚨 User Impact 🚨
Once this code is on `master` the updated `metadata.yaml`


.github/workflows/metadata_upload_manifest_dagger.yml
```diff
@@ -0,0 +1,22 @@
...
+        uses: actions/checkout@v2
+      - name: Run metadata upload pipeline
+        id: metadata-upload-pipeline
+        uses: ./.github/actions/run-dagger-pipeline
+        with:
+          subcommand: "metadata upload prod-airbyte-cloud-connector-metadata-service"
+          context: "master"
+        env:
+          CI_GITHUB_ACCESS_TOKEN: ${{ secrets.GH_PAT_MAINTENANCE_OCTAVIA }}
+          GCS_CREDENTIALS: ${{ secrets.METADATA_GCS_CREDENTIALS }}
```
alafanechere, line 22: @bnchrch this secret does not exists, we need to create it. Do you already have a service account with R/W permissions to the bucket you want to write these files to?

.github/workflows/metadata_upload_manifest_dagger.yml
```diff
@@ -0,0 +1,22 @@
...
+    name: Upload modified metadata files to GCS
+    runs-on: medium-runner
+    steps:
+      - name: Checkout Airbyte
+        uses: actions/checkout@v2
+      - name: Run metadata upload pipeline
+        id: metadata-upload-pipeline
+        uses: ./.github/actions/run-dagger-pipeline
+        with:
+          subcommand: "metadata upload prod-airbyte-cloud-connector-metadata-service"
```
alafanechere, line 18: @bnchrch do you confirm this is the destination bucket we want?

airbyte-config-oss/init-oss/src/main/resources/seed/oss_catalog.json
```diff
@@ -12594,11 +12594,7 @@
             "type": "string",
             "title": "Reader Options",
             "description": "This should be a string in JSON format. It depends on the chosen file format to provide additional options and tune its behavior.",
-            "examples": [ "{}", "{\"sep\": \" \"}", {
```
alafanechere, line 12597: I don't get why this change was introduce by the "Automated change" commit.

.github/workflows/metadata_upload_manifest_dagger.yml
bnchrch: Lets use this bucket while we are still developing
https://console.cloud.google.com/storage/browser/dev-airbyte-cloud-connector-metadata-service;tab=objects?forceOnBucketsSortingFiltering=false&project=dev-ab-cloud-proj&prefix=&forceOnObjectsSortingFiltering=false

bnchrch (changes requested):
@alafanechere Putting a higher level review on this one.

Loom:
<a href="https://www.loom.com/share/8a959dd1e062451385ce4c1023969d6e" targer="_blank">image</a>
https://www.loom.com/share/8a959dd1e062451385ce4c1023969d6e


**TLDR I think we should move the upload logic into the metadata_service/lib folder**

One of the issues I noticed with how our current cicd today was that a lot of the logic was spread out with no clear owner.

Some example of this were
1. The bucket name for our spec cache being inlined in multiple areas throughout the codebase
2. Manage.sh containing logic for many loosely related systems (creating docker images, publishing connector docker images, generating spec files and caching them)
3. report.sh caching the result of the latest test run, pulling down the last 5 and aggregating them into a slack report.
4.  connector catalogs being generated and regenerated across two codebases

With this in mind a goal I think is important for us to strive for is to keep ownership of logic really clear. If we establish strong lines it should help us avoid the shell script soup we had before.

Looking at our two systems that are coming together I see them owning logic in the following way:

1. Dagger owns all pipeline logic. That includes, what steps are run, when, in what order, in what environment, and how the result is reported to the user and to the greater CICD context (has this passed or failed on github)

2. metadata-service owns all logic for how we validate, upload and create files as they relate to the metadata system.

So in the case of this PR I think we have some logic in Dagger, that should move to the metadata-service
1. What bucket and paths (latest and version) should the metadata files be uploaded to
2. If a metadata file needs to be uploaded or can be skipped because the file hash is the same in the gcs bucket

However I think theres a few areas where there could be a case that Dagger owns the logic, and that is:
1. Determining if we run metadata upload on all metadata files or just a subset or none at all

We have a few tools in this project that can help us to achieve this separation
1. Folder/file structure
2. Local libraries/modules
3. CLI tools

which I think are shown in how we implemented the validate pipeline:
https://github.com/airbytehq/airbyte/blob/172bf93ac427a06bccb0c3b72bcabee314e18370/tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py#L85-L86

Ok thats a lot of context and a big ask. Let me know what you think

alafanechere:
Thank you for raising this concern. I totally agree with your point. I fell into the trap of implementing it this way because of my familiarity with the pipeline codebase. I'll refactor this code to move as much logic as possible back to the metadata service! Let's draw clear lines between projects 👍

alafanechere:
@bnchrch I've opened a different [PR](https://github.com/airbytehq/airbyte/pull/25115) to update move the upload logic to the metadata service. I'll iterate on this current PR to implement the orchestration logic only. I'll reuse the PoetryRun step I introduced in https://github.com/airbytehq/airbyte/pull/25107

alafanechere:
@bnchrch feel free to give a second review. I consolidated the `MetadataValidation` and `MetadataUpload` step logics. The dagger pipeline is now only responsible to spot modified metadata files and call the `metadata_service upload` command with these files.

airbyte-ci/connectors/metadata_service/lib/tests/test_commands.py
```diff
@@ -45,11 +45,13 @@ def test_upload(mocker, valid_metadata_yaml_files, uploaded):
...
         commands.upload, [metadata_file_path, "my-bucket", "-sa", metadata_file_path]
     )  # Using valid_metadata_yaml_files[0] as SA because it exists...
-    assert result.exit_code == 0
     if uploaded:
         commands.click.secho.assert_called_with(f"The metadata file {metadata_file_path} was uploaded to blob_id.", color="green")
+        assert result.exit_code == 0
     else:
         commands.click.secho.assert_called_with(f"The metadata file {metadata_file_path} was not uploaded.", color="yellow")
+        # We exit with 5 status code to share with the CI pipeline that the upload was skipped.
+        assert result.exit_code == 5
```
bnchrch, line 54: 👏

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
```diff
@@ -125,14 +125,19 @@ The new metadata service also uses dagger for its reproducible CI pipeline.
...
 
 For example to run the unit tests for the metadata service library you can run
 ```bash
 airbyte-ci metadata test lib
 ```
 
+For example to upload all the connector metadata files to GCS:
+```bash
+# You should set a valid JSON service account in the GCS_CREDENTIALS env var before uploading the files.
+airbyte-ci metadata upload <target-bucket> --all
```
bnchrch, line 139: Im thinking we may want the bucket set as an env variable.

Reasoning is when we want to "go live" we could just change the github actions env var for GCS_CREDENTIALS and METADATA_SERVICE_BUCKET without any code change.

This value could default to our dev bucket.

What do you think?

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -28,14 +29,56 @@ async def _run(self, poetry_run_args: list) -> StepStatus:
...
+            METADATA_FILE_NAME, get_metadata_file_from_path(context, metadata_path)
+        )
+
+    async def _run(self) -> StepStatus:
+        return await super()._run(["metadata_service", "validate", METADATA_FILE_NAME])
+
+
+class MetadataUpload(PoetryRun):
+
+    GCS_CREDENTIALS_CONTAINER_PATH = "gcs_credentials.json"
```
bnchrch, line 57: Question: Why not keep it as an env var in the container?

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -151,19 +156,58 @@ async def get_modified_files_remote(current_git_branch: str, current_git_revisio
...
     """Retrieve the list of modified files on the branch."""
     if is_local:
-        return get_modified_files_local(current_git_revision, diffed_branch)
+        return get_modified_files_in_branch_local(current_git_revision, diffed_branch)
     else:
-        return anyio.run(get_modified_files_remote, current_git_branch, current_git_revision, diffed_branch)
+        return anyio.run(get_modified_files_in_branch_remote, current_git_branch, current_git_revision, diffed_branch)
+
+
+async def get_modified_files_in_commit_remote(current_git_branch: str, current_git_revision: str) -> Set[str]:
```
bnchrch, line 174: 👏

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -175,6 +219,14 @@ def get_modified_connectors(modified_files: Set[str]) -> Set[Connector]:
     return set(modified_connectors)
 
 
+def get_modified_metadata_files(modified_files: Set[str]) -> Set[Path]:
+    return {Path(f) for f in modified_files if f.endswith(METADATA_FILE_NAME) and f.startswith("airbyte-integrations/connectors")}
+
+
+def get_all_metadata_files() -> Set[Path]:
```
bnchrch, line 226: Hmm its stuff like this thats making me wish the pipeline was located in `airbyte-ci` already so we could have this logic in metadata-service/lib and import it

bnchrch: Can you add a TODO here for that?

Ive added an issue to discuss in grooming on the subject: [Move pipelines into the airbyte-ci project](https://app.zenhub.com/workspaces/connector-operations-6318ede5eda32e0010ef8566/issues/gh/airbytehq/airbyte/25170)

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -249,10 +301,11 @@ def invoke(self, ctx: click.Context) -> Any:
             return sys.exit(1)
 
 
-async def execute_concurrently(steps: List[Callable], concurrency: int = 5):
-    semaphore = anyio.Semaphore(concurrency)
-    async with semaphore:
+async def execute_concurrently(steps: List[Callable], concurrency=5):
+    tasks = []
+    # Asyncer does not have builtin semaphore, so control concurrency via chunks of steps
```
bnchrch, line 306: 👏

bnchrch (commented):
Just a question on env vars id like to discuss.

After that this should be good to go!

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
alafanechere: It's more or less what I planned for https://github.com/airbytehq/airbyte/issues/23975

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
alafanechere: No hard reason. I stored it in a constant because this string is reused a couple of time. I'm using it in ```
                "--service-account-file-path",
                self.GCS_CREDENTIALS_CONTAINER_PATH,
```
instead of using the `GOOGLE_APPLICATION_CREDENTIALS` env var because I find that explicit declaration of option is maybe more readable.

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
alafanechere: I don't think I get your suggestion. Even if we used an env var to set the bucket we'd need to switch the env var value in the GHA workflow. Maybe you suggest to use a GitHub secrets that we could change without code change. I still find it useful for discoverability to explicitly declare the bucket name in the GHA workflow and would prefer to not use a secret for it.

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
alafanechere: I added the todo.
But I'm not sure this is a great idea use it in the metadata service lib. It would mean that the command would have to be run from the airbyte root repo. I like the fact that this command is agnostic from its environment: pass it a file path and upload/validate it. I think it's better to consider that the pipeline is in a context where all the metadata file resides, not the CLI tool itself.

bnchrch: > It would mean that the command would have to be run from the airbyte root repo.

Ah good point! I had been thinking maybe it becomes a `get_all_metadata_files_at_path` type of command.

But as I type it out that does feel a bit silly to leave a TODO for.

Sorry for not thinking about that clearly.

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
bnchrch: Ah sorry, I had meant secret. But youve convinced me. Lets leave it searchable!

bnchrch (approved):
Looks fantastic!

Let ship this bad boy!

PR #25056, opened 2023-04-11
Source Mixpanel: added HttpAvailabilityStrategy and undeclared fields for export and annotations streams

### LINKED ISSUE #22262: Source source-mixpanel: implement AvailabilityStrategy
## What

## How
  in the connector.

and that's totally fine.

### LINKED ISSUE #24490: Source Mixpanel: Add undeclared columns to spec
## What

Mixpanel currently fails the updated test. Its `acceptance-test-config.yml` was edited with the

**How this will help**:
  previously being sent. Doing this allows us to enable column selection for Mixpanel!

## How

```
columns.txt:./source-mixpanel.txt-80-The annotations stream has the following schema errors:
columns.txt:./source-mixpanel.txt:81:Additional properties are not allowed ('user' was unexpected)
columns.txt:./source-mixpanel.txt-100-The export stream has the following schema errors:
columns.txt:./source-mixpanel.txt:101:Additional properties are not allowed ('Referred by', 'import', 'URL', 'insert_id', 'mp_api_timestamp_ms', 'mp_api_endpoint', 'mp_processing_time_ms' were unexpected)

```

5. Profit!

Definition of done: Mixpanel passes CAT without declaring `fail_on_extra_columns: false`. If the
movement.

darynaishchenko:
- resolved https://github.com/airbytehq/airbyte/issues/24490 (set HttpAvailabilityStrategy)
- resolved https://github.com/airbytehq/airbyte/issues/22262 (added fields to schema)
Also was updated backoff: set exponential backoff as [docs](https://developer.mixpanel.com/reference/import-events#rate-limits) recommends it. 60s in backoff calculation as delay were set on testing result. 1 request per minute were removed because API allows to send more request per minute, so we can decrease time of work of connector. If we receive 429 error backoff will handle it.


airbyte-integrations/connectors/source-mixpanel/source_mixpanel/streams/base.py
```diff
@@ -58,7 +57,7 @@ def __init__(
 
     @property
     def availability_strategy(self) -> Optional["AvailabilityStrategy"]:
-        return None
+        return HttpAvailabilityStrategy()
```
artem1205, line 60: you can remove this method from here

airbyte-integrations/connectors/source-mixpanel/source_mixpanel/schemas/annotations.json
```diff
@@ -14,6 +14,20 @@
...
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": ["null", "integer"]
+        },
+        "first_name": {
+          "type" : "string"
+        },
+        "last_name": {
+          "type" : "string"
```
artem1205, line 28: please add `null` in type

artem1205 (approved):

lazebnyi (approved):
Nice!

bnchrch (approved):
Looks good.

Before you merge would you
1. Please make sure you update the version in both the source_definition.yaml and metadata.yaml file
2. Run processResources

PR #25097, opened 2023-04-11
destination bigquery: run normalization inside container if env var is set

### LINKED ISSUE #24878: Run normalization inside destination-bigquery docker container

edgao:
## What
closes https://github.com/airbytehq/airbyte/issues/24878

Wraps all normalization logs inside AirbyteLogMessages, and aggregates error logs into a single AirbyteTraceMessage at the end. This mimics existing behavior. Compare:
* dev incident (from a local run) https://airbytehq.sentry.io/issues/4102096279/?project=4504087736090624&referrer=issue-stream&stream_index=0
* some random other normalization incident from production https://airbytehq.sentry.io/issues/4004565639/?query=dbtunhandlederror&referrer=issue-stream&statsPeriod=90d&stream_index=2

They both have external/internal message set correctly, along with metadata + stacktrace. (the metadata now also has `connector_command=write` because it's running inside the connector!)

## How
Add a very simple shell script to run the java destination process, and then run normalization if necessary.

Add a new java class to handle the logs appropriately. It's largely based on these classes:
* https://github.com/airbytehq/airbyte-platform/blob/main/airbyte-commons-worker/src/main/java/io/airbyte/workers/normalization/NormalizationAirbyteStreamFactory.java#L51
* this part of DefaultNormalizationRunner https://github.com/airbytehq/airbyte-platform/blob/main/airbyte-commons-worker/src/main/java/io/airbyte/workers/normalization/DefaultNormalizationRunner.java#L154-L171

Note that we're also duplicating part of SentryExceptionHelper - this is intentional. That file isn't touched particularly frequently, and we're only duplicating the pieces that relate directly to dbt. It's also reasonable to say that connectors should own the translation from dbt error to sentry error, rather than trying to hardcode that into platform.

## Recommended reading order
1. run_with_normalization.sh + dockerfile + dockerignore
2. NormalizationLogParser + test (these are in base-java, so they're available to all destinations)
3. SentryExceptionHandler (this file is 99% copy-pasted)

## 🚨 User Impact 🚨
None. As with the previous PR (to install dbt/python), this has no runtime behavior change yet.


airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/normalization/NormalizationLogParser.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  /**
+   * Wrap the line in an AirbyteLogMessage, and do very naive dbt error log detection.
+   * <p>
+   * This is needed for dbt < 1.0.0, which don't support json-format logs.
+   */
+  private Stream<AirbyteMessage> nonJsonLineToMessage(final String line) {
+    // Super hacky thing to try and detect error lines
+    if (line.contains("[error]")) {
```
jcowanpdx, line 71: will this ever be ERROR? Do we need case insensitivity?

edgao: I sure hope not... this logic is mostly copied from https://github.com/airbytehq/airbyte/blob/a6d6b063037b62574143e1d09af6b342cf8fc240/airbyte-connector-test-harnesses/acceptance-test-harness/src/main/java/io/airbyte/workers/normalization/NormalizationAirbyteStreamFactory.java#L77

de facto - our current pre-1.0 dbt versions all use the exact string `[error]`, and >=1.0 dbt supports json-formatted logs. So I think we'll get away with it.

airbyte-integrations/bases/base-java/Dockerfile
```diff
@@ -10,6 +10,7 @@ WORKDIR /airbyte
 ADD https://dtdg.co/latest-java-tracer dd-java-agent.jar
 
 COPY javabase.sh .
+COPY run_with_normalization.sh .
```
alafanechere, line 13: @edgao I think this should be dynamically added by Dagger on connectors supporting destination and not declared in the base java image. To not make your interaction on this topic blocked by Dagger implementation I suggest you keep this script in this folder but directly copy it to the bigquery destination.

airbyte-config-oss/init-oss/src/main/resources/seed/oss_catalog.json
```diff
@@ -1111,6 +1111,54 @@
     "public": true,
     "custom": false,
     "releaseStage": "alpha"
+  }, {
```
alafanechere, line 1114: Why is this file updated? Catalog updates should happen on merge.

edgao: this was from ci 🤔 https://github.com/airbytehq/airbyte/commit/13100e008528c7a76759f9538251e026d62c61dd

(same with the connectors.md change)

airbyte-integrations/bases/base-java/Dockerfile
edgao: hm... are you thinking of a gradle copy task to get the script, and then a docker COPY? that feels kind of heavyweight

at the very least, a naive `COPY ../../bases/base-java/run_with_normalization.sh .` doesn't work, iirc docker copies files to a separate build context

jbfbell (approved):

alafanechere (approved):

airbyte-integrations/bases/base-java/Dockerfile
alafanechere: Ok! Feel free to keep it like it but on the Dagger side I think this should be only added on top of the java base if we need to package normalization.

edgao:
/publish connector=connectors/destination-bigquery,connectors/destination-bigquery-denormalized

airbyte-integrations/bases/base-java/Dockerfile
edgao: sounds good. I'll tag you on the dagger pr later this week to make sure I have the right idea

PR #25101, opened 2023-04-12
Make mysql YEAR type behave like an integer rather than a string

### LINKED ISSUE #8722: Re-evaluate MySQL year type
The returned year value from MySQL can either be a `java.sql.Short` (when `yearIsDateType=false`) or a `java.sql.Date` with the date set to January 1st, at midnight (when `yearIsDateType=true`). Currently, `JsonSchemaPrimitive` does not support integer, but only supports number. Because the number type will be interpreted as a double in many destinations, and it is weird to show a year as a double, we set `yearIsDateType=true` in the JDBC connection string, and parse the returned year value as a string (#7970). The case can be re-evaluated when `JsonSchemaPrimitive` supports integer.

rodireich:
## What
The YEAR mysql datatype is intended to be used as a shorthand common year value from `1901 to 2155`.
`1 to 69` maps to `2001 to 2069`
`70 to 99` maps to `1970 to 1999`

Previously we turned YEAR columns to string mainly because the airbyte protocol had `NUMBER` as in decimal number, but lacked an `INTEGER`, which in turned cause destination to turn years value to double - `1980.0`.
This change intends making year values captured as integer.

## How
Convert incoming mysql YEAR into a `JsonSchemaType.INTEGER` - similarly to how we handle tinyint (short).

## Recommended reading order
1. `MySqlSourceOperations`

## 🚨 User Impact 🚨
This may be a breaking change as streams containing a YEAR field previously synced as string value will now be created on the destination as an integer.
This [comment ](https://github.com/airbytehq/airbyte/issues/8722#issuecomment-1231949854) suggests that the change we're making is backward compatible - but I'm not sure why.

*For connector PRs, use this section to explain which type of semantic versioning bump occurs as a result of the changes. Refer to our [Semantic Versioning for Connectors](https://docs.airbyte.com/contributing-to-airbyte/#semantic-versioning-for-connectors) guidelines for more information. **Breaking changes to connectors must be documented by an Airbyte engineer (PR author, or reviewer for community PRs) by using the [Breaking Change Release Playbook](https://docs.google.com/document/d/1VYQggHbL_PN0dDDu7rCyzBLGRtX-R3cpwXaY8QxEgzw/edit).***


airbyte-integrations/connectors/source-mysql/src/main/java/io/airbyte/integrations/source/mysql/MySqlSource.java
```diff
@@ -191,7 +191,7 @@ public JsonNode toDatabaseConfig(final JsonNode config) {
     jdbcUrl.append("&tinyInt1isBit=true");
     // ensure the return year value is a Date; see the rationale
     // in the setJsonField method in MySqlSourceOperations.java
-    jdbcUrl.append("&yearIsDateType=true");
+    jdbcUrl.append("&yearIsDateType=false");
```
rodireich, line 194: The behavior I'm seeing doesn't match documentation. But I changed to `false` as it makes a YEAR to be read into a `SMALLINT` rather than a `DATE`.
But in both cases value converted correctly to the same integer.

edgao (approved):
WOOHOOOOO INTEGERS

(tests look reasonable; I have no opinion about the actual code change)

subodh1810 (commented):
Is CDC fine or do we need to make changes there as well?

rodireich:
CDC is fine and also tested - for both snapshot and binlog.
Where I did find a problem which I think is a debezium issue is:
Documented behavior With `yearIsDateType=false` a YEAR value of
Integer `0` → should return `0`
string `'0'` → should return `2000`.
This works as expected with standard jdbc, but CDC always returns `2000`.

This is an edge case of this obscure datatype so I removed the test.

subodh1810 (changes requested):
What's the migration going to look like? We have had trouble in the past whenever we have made changes to the data type handling. Right now YEAR columns are created as STRIng right? How is the transformation going to be handled by normalization? @edgao just wanted to make sure we dont get into the same trouble that happened with JSONB data type.

subodh1810:
I think we should carry out few tests! Create a connector on latest version of Mysql and then run a sync for a table with YEAR data type and then upgrade the connector version and then again run sync to make sure that the normalization is happy with upgrade. We should also test roll back. Also use a destination like snowflake which has normalization support and not LOCAL CSV/JSON destination

edgao:
that's a fair point, I think it's worth running some tests (ideally with all of snowflake+bigquery+postgres). I expect this to be a lot safer just because strings and ints behave predictably, but testing would be a good confirmation

rodireich:
I tested this and can confirm that rolling in the change and rolling back are all seamless.
Tested with mysql <> snowflake.
Normalization can overcome the mismatch when destination table YEAR column is a string and after upgrade value coming in as an integer.
Similarly, in case of rollback destination table YEAR column is an integer and values coming as string - normalization convert to correct destination type.

The only thing I see is the expected warning on schema change "Source schema changed for table benchmark.products! Potential mismatches: [yy]. Actual schema…"
But in  both cases destination tables has the expected values and syncs complete without an error.

rodireich:
I tested on mysql <> postgres also as part of another ticket. Normalization can overcome mismatch there as well so roll in and rollback go smoothly as well.

subodh1810 (approved):

PR #25105, opened 2023-04-12
metadata: cleanup GHA workflows

alafanechere:
## What
An attempt at simplifying the metadata related workflow.

## How
* Fix typo in job or step names due to copy/pasting
* Gather metadata lib and orchestrator testing in the same workflow.
* Try to use meaningful and consistent job / step / workflow names.
* Remove declarative timeout to fallback to GHA defaults as these jobs should not be long running ones.


bnchrch (approved):
Im all for this!

PR #25107, opened 2023-04-12
airbyte-ci: improve metadata validation perfs

alafanechere:
## What
The previous approach lacked of concurrency and I think the use of Pipx is not required as simply running poetry run in containers provide enough isolation.
## How
* Create a PoetryRun step used to run `poetry run pytest` 
* Create a MetadataValidation(PoetryRun) step to run the ` poetry run metadata_service validate <yaml file>`
* Mount metadata files to the validation container with poetry to avoid mounting full connector dir.

**Validating all the connectors metadata takes 30s vs 220s before**

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
@@ -55,7 +55,7 @@ def with_python_base(context: PipelineContext, python_image_name: str = "python:
         context.dagger_client.container()
         .from_(python_image_name)
         .with_mounted_cache("/root/.cache/pip", pip_cache, sharing=CacheSharingMode.LOCKED)
-        .with_mounted_directory("/tools", context.get_repo_dir("tools", include=["ci_credentials", "ci_common_utils"], exclude=[".venv"]))
+        .with_mounted_directory("/tools", context.get_repo_dir("tools", include=["ci_credentials", "ci_common_utils"]))
         .with_exec(["pip", "install", "--upgrade", "pip"])
     )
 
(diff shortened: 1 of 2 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -452,25 +452,11 @@ def with_poetry(context: PipelineContext) -> Container:
     python_with_poetry = with_pip_packages(python_with_git, ["poetry"])
 
     poetry_cache: CacheVolume = context.dagger_client.cache_volume("poetry_cache")
-    poetry_with_cache = python_with_poetry.with_mounted_cache("/root/.cache/pypoetry", poetry_cache, sharing=CacheSharingMode.PRIVATE)
```
alafanechere, line 455: I private mode the cache is not shared between pipelines, I think we can opt for SHARED so that different poetry project we run in different pipelines can share the same cache.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -55,7 +55,7 @@ def with_python_base(context: PipelineContext, python_image_name: str = "python:
         context.dagger_client.container()
         .from_(python_image_name)
         .with_mounted_cache("/root/.cache/pip", pip_cache, sharing=CacheSharingMode.LOCKED)
-        .with_mounted_directory("/tools", context.get_repo_dir("tools", include=["ci_credentials", "ci_common_utils"], exclude=[".venv"]))
```
alafanechere, line 58: `.venv` is implicitely excluded as a default excluded dir in `get_repo_dir`

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -113,13 +89,13 @@ async def run_metadata_lib_test_pipeline(
     async with dagger.Connection(DAGGER_CONFIG) as dagger_client:
         metadata_pipeline_context.dagger_client = dagger_client.pipeline(metadata_pipeline_context.pipeline_name)
         async with metadata_pipeline_context:
-            test_lib_step = TestPoetryModule(
+            test_lib_step = PoetryRun(
```
alafanechere, line 92: @bnchrch let me know if the creation of a  `PoetryRunPytest(PoetryRun)` step calling `pytest`  would make this code more readable.

bnchrch: I think it would, or at least would be a good way to reduce line of code. But not nessesary at all!

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -72,21 +59,10 @@ async def run_metadata_validation_pipeline(
...
-            )
-
-            validation_steps = [
-                metadata_validation_step(metadata_pipeline_context, metadata_path, parent_container).run
-                for metadata_path in metadata_source_paths
-            ]
+            validation_steps = [MetadataValidation(metadata_pipeline_context, metadata_path).run for metadata_path in metadata_source_paths]
 
-            results = await execute_concurrently(validation_steps)
+            results = await execute_concurrently(validation_steps, concurrency=len(validation_steps))
```
bnchrch, line 65: Great call

bnchrch (approved):
Very well done!

Thanks for the speed improvements

PR #25115, opened 2023-04-12
metadata service: upload metadata files to gcs

alafanechere:
## What
Add a command in the metadata service library to upload metadata files to a GCS Bucket.

(1/2) for https://github.com/airbytehq/airbyte/issues/24850 : the orchestration logic will come in https://github.com/airbytehq/airbyte/pull/25044

## Logic
* If the file already exists it is not overwritten
* If the file is new with a different etag than the one stored under the `latest` prefix: update the latest file.

The blob name structure are the following :
`metadata/<connector-docker-registry>/<connector-tag>/metadata.yaml`
`metadata/<connector-docker-registry>/latest/metadata.yaml`


airbyte-ci/connectors/metadata_service/lib/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.4.2 and should not be changed by hand.
```
alafanechere, line 1: @bnchrch I'm worried that we may not be using the same poetry version. Could you please checkout this branch and check if a `poetry install` works correctly?

bnchrch: `poetry install` worked 
https://app.warp.dev/block/MXAlAwrwhk3rO83WmFZ2Zg

`poetry run metadata_service upload --help` worked
https://app.warp.dev/block/pqWT4J1W1yi55eMaAWYKqT

But! my poetry is behind `Poetry (version 1.3.2)`

So I updated to v1.4.2 and the above commands still work.

What was going wrong on your end?

airbyte-ci/connectors/metadata_service/lib/metadata_service/commands.py
```diff
@@ -1,7 +1,12 @@
-import click
+#
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+#
```
bnchrch, line 3: New guy question: What is the command that adds this?

airbyte-ci/connectors/metadata_service/lib/metadata_service/commands.py
```diff
@@ -12,7 +17,7 @@ def metadata_service():
 @metadata_service.command(help="Validate a given metadata YAML file.")
 @click.argument("file_path", type=click.Path(exists=True, path_type=pathlib.Path))
 def validate(file_path: pathlib.Path):
-    file_path = file_path if not file_path.is_dir() else file_path / "metadata.yml"
+    file_path = file_path if not file_path.is_dir() else file_path / "metadata.yaml"
```
bnchrch, line 20: nit: Maybe `metadata.yaml` moves to a reusable constant?

airbyte-ci/connectors/metadata_service/lib/poetry.lock
alafanechere: Nothing was wrong :D It's just that these changes on the lock file looked suspicious.

airbyte-ci/connectors/metadata_service/lib/tests/test_commands.py
```diff
@@ -0,0 +1,73 @@
+#
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+#
+import pytest
+from click.testing import CliRunner
+from metadata_service import commands
+from pydantic import BaseModel, ValidationError, error_wrappers
+
+
+# TEST VALIDATE COMMAND
```
bnchrch, line 10: Thank you for these 🙏

bnchrch (approved):
This all looks good, no major issues at all.

Thanks so much for taking the feedback well.

Approved!

airbyte-ci/connectors/metadata_service/lib/metadata_service/commands.py
alafanechere: It's the pre-commit hook that does it.
`pre-commit install` and you should be good to go.

PR #25125, opened 2023-04-12
Bump versions in atelier dockerfile

girarda:
## What
* Bump the CDK version in the connector-atelier-server's Dockerfile when releasing a new version of the CDK

## How
* Get the previous version from the CDK_VERSION file
* Use sed to replace instances of the previous version with the new version in the Dockerfile

## Recommended reading order
1. `.github/workflows/publish-cdk-command-manually.yml`

## Testing
To test the flow, I hardcoded the sed to replace old version with 0.0.0, removed the publish steps, and ran the flow. [This PR](https://github.com/airbytehq/airbyte-platform-internal/pull/5861/files) shows the "old version" (0.34.1) was replaced with the new version (0.0.0)


clnoll (approved):
LGTM!

PR #25142, opened 2023-04-13
Remove session token auth section from compatibility guide

flash1293:
As discussed offline, the session token authentication mechanism will be removed from the connector builder.


girarda (approved):
:shipit:

Issue #25099, opened 2023-04-12
Create a dagger pipeline that builds connectors for multiple architectures using buildx


bnchrch:
## What this is
1. Adding a new airbyte-ci command to create a docker image for a connecter `airbyte-ci connectors build --connector <name>`
2. This command should build the docker images for multiple architectures (buildx) (see [manage.sh](https://github.com/airbytehq/airbyte/blob/172bf93ac427a06bccb0c3b72bcabee314e18370/tools/integrations/manage.sh#L238-L239) for details on how we currently do this)

PR #25197, opened 2023-04-14
connectors-ci: publish dev images to DockerHub

### LINKED ISSUE #25203: connectors-ci: build and use pre-release and published image to test connectors
To build connector dev images and share them with acceptance / integration test step I originally took the following approach:

- Build connector images with dagger
- Use `.export()` to export them as tar file to the host filesystem
- Integration and acceptance test step get this tar file path as an input and load it to their docker host with docker load command.

**Exporting images as tar files leads to the following problem:**

- Disk writes to the host filesystem are long
- Nightly builds require huge disk space to store all our connector images as tar
- Dagger caches the export results in buildkit layers. Its stuffing the layers and their S3 export and download for caching is long.

**Workaround**
I suggest to publish built dev image to DockerHub.
Acceptance/integration test step then pull these images instead of loading them from tar files.
Intermediate Write/Reads on the host file system is no longer required.

alafanechere:
## What
Closes #25203 
To build connector dev images and share them with acceptance / integration test step I originally took the following approach:
* Build connector images with dagger
* Use `.export` to export them as `tar` file to the host filesystem
* Integration and acceptance test step get this tar file path as an input and load it to their docker host with `docker load` command.

Exporting images as tar files leads to the following problem:
* Disk writes to the host filesystem are long
* Nightly builds  require huge disk space to store all our connector images as tar
* Dagger caches the export results in buildkit layers. Its stuffing the layers and their S3 export and download for caching is long.

## How
In this PR is suggest to publish built dev image to DockerHub. 
Acceptance/integration test step then pull these images instead of loading them from tar files. 
Intermediate Write/Reads on the host file system is no longer required.

In the CI the images are published to the `airbyte` DockerHub registry. 
When the dagger pipeline is run locally users can target their private docker registry with:
`airbyte-ci connectors --docker-registry=augustinairbyte test --name=source-openweather`

An example of an image published by the pipeline locally is  available [here](https://hub.docker.com/repository/docker/augustinairbyte/source-openweather/tags?page=1&ordering=last_updated) on DockerHub. I followed the tag semantics @pedroslopez defined for pre release images in https://github.com/airbytehq/airbyte/commit/9bf183a439495e5ebd9f6185b295c97c675500ac

**A succesful test run on `destination-bigquery` on this branch can be found [here](https://github.com/airbytehq/airbyte/actions/runs/4698361668/jobs/8330545659), the published pre-released image is [here](https://hub.docker.com/layers/airbyte/destination-bigquery/1.2.20-dev.c90f98ef1e/images/sha256-96942a03d5b68fe2b41e9a40ea6974b1731201537f8f03a26cad9ecf64da0013?context=explore).**


tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
```diff
@@ -33,11 +33,13 @@
...
 
 
-def validate_environment(is_local: bool, use_remote_secrets: bool):
+def validate_environment(is_local: bool, use_remote_secrets: bool, docker_registry: str):
     """Check if the required environment variables exist."""
     if is_local:
         if not (os.getcwd().endswith("/airbyte") and Path(".git").is_dir()):
             raise click.UsageError("You need to run this command from the airbyte repository root.")
+        if docker_registry == "airbyte":
+            raise click.UsageError("You should not use the airbyte docker hub registry when running locally. Please use a personal one.")
```
erohmensing, line 42: Would we consider having a separate testing repository that everyone could use, instead of each person having to set up a personal one? 

If using a personal, do you have to do anything to your personal registry beforehand to make this work?

tools/ci_connector_ops/ci_connector_ops/pipelines/tests/java_connectors.py
```diff
@@ -250,28 +222,28 @@ class IntegrationTestJava(GradleTask):
...
 
-    async def _load_connector_image(self, connector_tar_file: File):
+    async def _load_connector_image(self, connector_image_ref: str):
         connector_image_tag = f"airbyte/{self.context.connector.technical_name}:dev"
-        self.context.logger.info("Load the connector image to the docker host")
-        await environments.load_image_to_docker_host(
-            self.context, connector_tar_file, connector_image_tag, docker_service_name=self.docker_service_name
+        self.context.logger.info("Pull the connector image to the docker host")
+        await environments.pull_image_to_docker_host(
+            self.context, connector_image_ref, connector_image_tag, docker_service_name=self.docker_service_name
```
erohmensing, line 237: If the disk size incurred by saving all the tar files was too large, will the same issue happen when pulling all the docker images to use? I know after running all the CAT tests locally by pulling the images (plus a few more images) I currently use 123.1GB of disk space on docker images

evantahler:
+1 on @erohmensing's concern above about if downloading all the docker images takes more or less space than the TARs.  

Also, this adds a dependecy on dockerhub, which we've seen to be flaky, onto our tests.

Are we building a new docker image and pushing it every night, with a new tag?  Is there a limit to how many tags/versions one docker image can have?

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
alafanechere: If you have a dockerhub account you have a registry by default whose name is your username, so no extra setup is required.
I agree that a separate global testing registry could be more convenient.

cpdeethree:
> +1 on @erohmensing's concern above about if downloading all the docker images takes more or less space than the TARs.
> 
> Also, this adds a dependecy on dockerhub, which we've seen to be flaky, onto our tests.
> 
> Are we building a new docker image and pushing it every night, with a new tag? Is there a limit to how many tags/versions one docker image can have?

I'm also concerned about the network overhead this introduces. If we do go down this route I think we need a registry in EKS alongside the cluster where these jobs run so we aren't paying the time tax to go back and forth to dockerhub.

IMO we should explore the tar file issue more. Disk space is cheap, I don't mind spinning up runners with 500gb-1tb of disk space per each. I'm more concerned about the extra time it takes to upload them to the dagger cache

tools/ci_connector_ops/ci_connector_ops/pipelines/tests/java_connectors.py
alafanechere: The thing is that when you pull an image its content is repesented by buildkit layers that can be fat but there storage is optimized and can be shared between images with same layers. When using tar file we dump to disk these full images. So we store the image content two times: in buildkit layers and on disk with tar files.

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
erohmensing: Yeah, I'm thinking that there are some people who probably prefer not to have 300+ images on their public repo, especially if they leave the company and don't want to delete them all, or have other ones that become hard to find in the mix. I was worried about the public/private distinction, but looks like you can have as many public images as you like (which is the default)

alafanechere:
@erohmensing @evantahler @cpdeethree 
I agree with all of your concerns. 
The thing I don't like with tar file storage is that we basically keep image contents at two locations: in buildkit layers and on disk, just for the sake of sharing these images with other steps.
If the disk space is not a concern, I still observe the export operations that write to disk to be slower than the network calls on docker push. 

I'm all in for a sidecar local registry we could use to share images with acceptances / integration tests. 
But in the meantime I'd like to unblock our capacity of running connectors testing succesfully, to proceed with the more "organizational" change that are remaining on this project.

[FYI I was able to run successfully a pipeline on all GA connectors on this branch:.
](https://github.com/airbytehq/airbyte/actions/runs/4698734917/jobs/8331380781)

tools/ci_connector_ops/ci_connector_ops/pipelines/tests/java_connectors.py
erohmensing: The same things happen (the buildkit sharing) when building with docker locally, right? I didn't pull all of the images from dockerhub, I built them all locally. Could we, instead of build + push + pull docker images, only build and use the local ones? I guess I'm not clear on why we went the .tar path in the first place

erohmensing: Sounds like that'd be the `I'm all in for a sidecar local registry we could use to share images with acceptances / integration tests.` approach you mentioned below 👍🏻

alafanechere:
> I'm also concerned about the network overhead this introduces.

I think the problem with the tar file approach is also that it introduces network overhead: exporting the cache to S3 is very slow as the number of image built grows. When we run `export`, the output of `export` (the tar binary) is stored in the cache, explaining its huge size.
I agree we shall reach out the dagger team on this topic.

alafanechere:
@cpdeethree @erohmensing @evantahler I've explain in details the motivation behind this PR [here](https://github.com/airbytehq/airbyte/issues/25229#issuecomment-1511312205). I consider it as a stop gap while Dagger helps us figure how to overcome the current bottleneck. 
I find it's a quick workaround that is easy to revert and allows us to unblock the other issues of the epic.
@pedroslopez was concerned that we're not publishing multi-architecture images. This is something that will be done in https://github.com/airbytehq/airbyte/issues/25099

PR #25204, opened 2023-04-14
Source Facebook Marketing: fix data retention period validation

davydov-d:
## What
https://github.com/airbytehq/oncall/issues/1678

## How
For some specific dates like 2023-03-31, subtracting 37 months gives us 2020-02-29. But this date is still out of range from the FB API perspective. That's why we have to adjust dates like this to the first day of the next month - 2020-03-01.

Additionally, fix date range validation in `check` for the dates in the future.


lazebnyi (approved):

arsenlosenko (approved):

clnoll (approved):
LGTM!

Issue #24951, opened 2023-04-06
Builder documentation: Partitioning page


flash1293:
Extend connector builder documentation. Details in https://docs.google.com/document/d/17LXN5RC_CQd6O_v0Z2YqZmuNKaykDxCu2IOEj5DTqNQ/edit

PR #25209, opened 2023-04-14
Connector builder: Partitioning documentation

flash1293:
## What

Related to https://github.com/airbytehq/airbyte/issues/24951

(not complete, screen recording still missing)

### DIFF
--- a/docs/connector-development/connector-builder-ui/partitioning.md
+++ b/docs/connector-development/connector-builder-ui/partitioning.md
@@ -1,0 +1,116 @@
+# Partitioning
+
+Partitioning is required if the records of a stream are nested within multiple buckets or parent resources that need to be queried separately to extract the records.
+
+Sometimes records belonging to a single stream are partitioned into subsets that need to be fetched separately. In most cases, these partitions are a parent resource type of the resource type targeted by the connector. The partitioning feature can be used to configure your connector to iterate through all partitions. In API documentations, this concept is showing up as mandatory parameters that need to be set on the path, query or request body of the request.
+
+Common API structures look like this:
+* The [SurveySparrow API](https://developers.surveysparrow.com/rest-apis/response#getV3Responses) allows to fetch a list of responses to surveys. For the `/responses` endpoint, the id of the survey to fetch responses for needs to be specified via the query parameter `survey_id`. The API does not allow to fetch responses for all available surveys in a single request, there needs to be a separate request per survey. The surveys represent the partitions of the responses stream.
+* The [Woocommerce API](https://woocommerce.github.io/woocommerce-rest-api-docs/#order-notes) includes an endpoint to fetch notes of webshop orders via the `/orders/<id>/notes` endpoint. The `<id>` placeholder needs to be set to the id of the order to fetch the notes for. The orders represent the partitions of the notes stream.
+
+There are some cases that require multiple requests to fetch all records as well, but partitioning is not the right tool to configure these in the connector builder:
+* If your records are spread out across multiple pages that need to be requested individually if there are too many records, use the Pagination feature.
+* If your records are spread out over time and multiple requests are necessary to fetch all data (for example one request per day), use the Incremental sync feature.
+
+## Dynamic and static partition routing
+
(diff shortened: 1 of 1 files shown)


docs/connector-development/connector-builder-ui/partitioning.md
```diff
@@ -0,0 +1,116 @@
...
+
+The first two options are a "static" form of partition routing (because the partitions won't change as long as the Airbyte configuration isn't changed). The API providing the partitions via one or multiple separate requests is a "dynamic" form of partition routing because the partitions can change any time.
+
+### List partition router
+
+To configure static partitioning, choose the "List" method for the partition router. The following fields have to be configured:
+* The "partition values" can either be set to a list of strings, making the partitions part of the connector itself or delegated to a user input so the end user configuring a Source based on the connector can control which partitions to fetch.
+* The "Current partition value identifier" can be freely choosen and is the identifier of the variable holding the current partition value. It can for example be used in the path of the stream using the `{{ stream_slice.<identifier> }}` syntax.
+* The "Inject partition value into outgoing HTTP request" option allows to configure how to add the current partition value to the requests
```
brianjlai, line 26:
```suggestion
* The "Inject partition value into outgoing HTTP request" option allows you to configure how to add the current partition value to the requests
```

docs/connector-development/connector-builder-ui/partitioning.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+### Substream partition router
+
+To fetch the list of partitions (in this example surveys or orders) from the API itself, the "Substream" partition router has to be used. It allows you to select another stream of the same connector to serve as the source for partitions to fetch. Each record of the parent stream is used as a partition for the current stream.
+
+The following fields have to be configured to use the substream partition router:
+* The "Parent key" is the property on the parent stream record that should become the partition value (in most cases this is some form of id)
+* The "Current partition value identifier" can be freely choosen and is the identifier of the variable holding the current partition value. It can for example be used in the path of the stream using the `{{ stream_slice.<identifier> }}` interpolation placeholder.
+* The "Parent stream" defines the records of which stream should be used as partitions
```
brianjlai, line 63: Nit: This bullet might make sense as the first bullet because the remaining bullets how we use the parents records from the stream

docs/connector-development/connector-builder-ui/partitioning.md
```diff
@@ -0,0 +1,116 @@
...
+
+#### Example
```
brianjlai, line 65: well done on the explanation for substream routers. I always feel like this one can get confusing given the added complexity of multiple streams and identifiers

docs/connector-development/connector-builder-ui/partitioning.md
```diff
@@ -0,0 +1,116 @@
...
+
+If a user configures the URLs `example.com` and `example.org` and the strategies `desktop` and `mobile`, then the following requests will be triggered
+```
+curl -X GET https://www.googleapis.com/pagespeedonline/v5/runPagespeed?url=example.com&strategy=desktop
+curl -X GET https://www.googleapis.com/pagespeedonline/v5/runPagespeed?url=example.com&strategy=mobile
+curl -X GET https://www.googleapis.com/pagespeedonline/v5/runPagespeed?url=example.org&strategy=desktop
+curl -X GET https://www.googleapis.com/pagespeedonline/v5/runPagespeed?url=example.org&strategy=mobile
+```
+
+## Adding the partition value to the record
```
brianjlai, line 100: good addition!

brianjlai (approved):

brianjlai:
i also see the authentication.md docs are included, but presumably they'll be removed/rebased when that PR is merged.

docs/connector-development/connector-builder-ui/partitioning.md
```diff
@@ -0,0 +1,116 @@
+# Partitioning
+
+Partitioning is required if the records of a stream are nested within multiple buckets or parent resources that need to be queried separately to extract the records.
```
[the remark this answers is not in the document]
flash1293: Yeah, I struggled a bit with finding the balance between being general enough to capture all cases but specific enough so you still understand what it's about

docs/connector-development/connector-builder-ui/partitioning.md
flash1293: Good point, and that applies as well to the UI. I will change it there as well

girarda (approved):

PR #25215, opened 2023-04-14
Run doc deploys on docosaurus changes

timroes:
## What

Just notices that the docs won't deploy if you change somethign in the docusaurus folder, but they should, now they do.


supertopher (approved):
ty tim!

supertopher:
auto-merged

PR #25180, opened 2023-04-13
connector builder: Emit message at start of slice

girarda:
## What
Emit slice log messages when running a test read from the connector builder.

The AbstractSource currently only emits those messages if the debug flag is turned on, which is not correct since we introduced the `emit_connector_builder_messages` dedicated to the connector builder.

## How
* Move the condition (is the log level debug) to a method that can be overwritten
* Overwrite the condition in `ManifestDeclarativeSource` so the log messages are also emitted if `emit_connector_builder_messages` is True
* Move `emit_connector_builder_messages` from the `ModelToComponentFactory` to the `ManifestDeclarativeSource`
* Pass the flag to `ModelToComponentFactory.create_component` and downstream so it can be used to configure the retriever
* Add an integration test

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py`
2. `airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py`
4. `airbyte-cdk/python/airbyte_cdk/connector_builder/connector_builder_handler.py`
5. `airbyte-cdk/python/unit_tests/connector_builder/test_connector_builder_handler.py`
6. `airbyte-cdk/python/unit_tests/sources/declarative/parsers/test_model_to_component_factory.py`

## Testing with the connector-atelier-server
Setup:
1. Build connector-atelier-server
2. from platform-internal: `./gradlew -p oss airbyte-connector-atelier-server:run`
7. from webapp-2e2-tests: `npm run createdummyapi` 

Then run the following curl command. The partition router has 6 values, but the connector builder limits the number of slices to 5, so the jq command evaluating the length of the `slices` array should return 5
```
curl -X POST 'localhost/v1/stream/read' \
  -H 'Accept: */*' \
  -H 'Accept-Language: en-US,en;q=0.9' \
  -H 'Authorization: Basic YWlyYnl0ZTpwYXNzd29yZA==' \
  -H 'Connection: keep-alive' \
  -H 'Cookie: ajs_user_id=868ac191-c872-417c-8935-e09e782c5ae5; ajs_anonymous_id=1387f288-277d-41d4-940e-7f23777ab1b7; _gcl_au=1.1.960841038.1678276956' \
  -H 'Origin: http://localhost:8000' \
[...]

PR #25217, opened 2023-04-14
Fix manifest_declarative_source + add unit tests

girarda:
## What
1. Fix a typo I introduced in my [last PR](https://github.com/airbytehq/airbyte/pull/25180)
2. Add a few "integration tests" for manifest declarative source

## How
1. `super(self)` -> `super()`
2. test with and without pagination and partition router

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/manifest_declarative_source.py`
3. `airbyte-cdk/python/unit_tests/sources/declarative/test_manifest_declarative_source.py`

### DIFF
--- a/airbyte-cdk/python/unit_tests/sources/declarative/test_manifest_declarative_source.py
+++ b/airbyte-cdk/python/unit_tests/sources/declarative/test_manifest_declarative_source.py
@@ -6,7 +6,7 @@ import json
 import logging
 import os
 import sys
-from typing import List
+from typing import List, Mapping, Any
 from unittest.mock import call, patch
 
 import pytest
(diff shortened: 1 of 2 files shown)


airbyte-cdk/python/unit_tests/sources/declarative/test_manifest_declarative_source.py
```diff
@@ -727,3 +742,419 @@ def test_given_debug_when_read_then_set_log_level(self, declarative_source_read)
...
+             _create_page({"rates": [{"USD": 3, "partition": 0}], "_metadata": {}}),
+             _create_page({"rates": [{"ABC": 2, "partition": 1}], "_metadata": {}}),
+     ),
+     [{"ABC": 0, "partition": 0}, {"AED": 1, "partition": 0}, {"USD": 3, "partition": 0}, {"ABC": 2, "partition": 1}],
+     [call({"partition": "0"}, {}, None), call({"partition": "0"}, {}, {"next_page_token": "next"}), call({"partition": "1"}, {}, None),]
+     )
+])
+def test_read_manifest_declarative_source(test_name, manifest, pages, expected_records, expected_calls):
+    _stream_name = "Rates"
+    config = {"__injected_declarative_manifest": manifest}
```
brianjlai, line 1148: why do we need to inject the manifest into the config? Can't we just pass the manifest as its own parameter to `_run_read()` and then skip calling `config.get("__injected_declarative_manifest")`. 

Reading this gave me the impression we're testing the `connector_builder` entrypoint because of the way we inject it into the config which is slightly misleading since this is very much just the common flow for all low-code processing

brianjlai (approved):
1 line fix, 432 lines of tests. Now this is podracing! Nit about testing, but lgtm!

Issue #24950, opened 2023-04-06
Builder documentation: Incremental sync page


flash1293:
Extend connector builder documentation. Details in https://docs.google.com/document/d/17LXN5RC_CQd6O_v0Z2YqZmuNKaykDxCu2IOEj5DTqNQ/edit

Issue #21927, opened 2023-01-26
Connector builder: Allow format configuration for start/end time in dateslicer


flash1293:
A lot of existing manifests use different formats for the start and end time in the date slicer component. Right now this can't be configured in the UI (it will always inherit the format from the slicing component itself which enforces consistent formats used in the API response and the configuration which can be problematic). It's possible to work around this by doing date formatting within jinja templates, but this case is common enough to provide it as a top-level UI element.

<img width="957" alt="Screenshot 2023-02-02 at 10 58 08" src="https://user-images.githubusercontent.com/1508364/216293337-dd278b45-c661-47fc-8c8b-59c8f054cd87.png">

## Acceptance criteria

* It's possible to specify an optional format only for the end date / start date of the date stream slicer component which is different than the main format
* The default behavior is still the same as before
* Remove the default value for the Datetime Format field, as that will now be only the format used for API requests

lmossman:
Technically speaking I think we can achieve this by always using the MinMaxDatetime component for start_ and end_datetime, which allows you to optionally specify a format. 

And in the UI, we can just have one optional format field under the start datetime input, and another under the end datetime input

flash1293:
Hypothesis:
* The separate format for the start and end date is not something that's absolutely required, but it allows a connector developer to decouple the config format from the API format
* It's super rare (and almost never required) to have _different_ formats for start and end time
* Start date time is almost always a user input
* End date time is almost always "now", sometimes a user input

Suggestion:
* Make start date time a dropdown with two options: "user input" or "custom"
* Make end date time a dropdown with three options: "user input", "now" or "custom"
* If custom is selected, the start datetime can be defined using jinja macros like right now 
* If user input is selected, an implicit user input (like for auth) is added which is referenced using the jinja marco logic correctly under the hood (also taking into account format conversions, see below)
* Add an additional input "start/end date time format" which is optional (applied to both start and end time)

flash1293:
Set start datetime's default to "user input" and end datetime to "now"

flash1293:
@lmossman follow-up on this:

> Does the main datetime format also dictate how the date is injected?

Yes, it does, tested this like this:
<img width="1030" alt="Screenshot 2023-04-05 at 09 54 08" src="https://user-images.githubusercontent.com/1508364/230017154-95b545ef-5371-4eb2-997f-28706d4f8c1b.png">

[...]

PR #25238, opened 2023-04-17
Connector builder: Incremental sync documentation

flash1293:
## What

Related to https://github.com/airbytehq/airbyte/issues/24950

(not complete, screen recording still missing)


docs/connector-development/connector-builder-ui/incremental-sync.md
```diff
@@ -0,0 +1,133 @@
...
+
+To use incremental syncs, the API endpoint needs to fullfil the following requirements:
+* Records contain a date/time field that defines when this record was last updated (the "cursor field")
+* It's possible to filter/request records by the cursor field
+
+To learn more about how different modes of incremental syncs, check out the [Incremental Sync - Append](/understanding-airbyte/connections/incremental-append/) and [Incremental Sync - Deduped History](/understanding-airbyte/connections/incremental-deduped-history) pages.
+
+## Configuration
+
+To configure incremental syncs for a stream in the connector builder, you have to specify how the records specify the **"last changed" / "updated at" timestamp**, the **initial time range** to fetch records for and **how to request records from a certain time range**.
```
clnoll, line 17:
```suggestion
To configure incremental syncs for a stream in the connector builder, you have to specify how the records will represent the **"last changed" / "updated at" timestamp**, the **initial time range** to fetch records for and **how to request records from a certain time range**.
```

docs/connector-development/connector-builder-ui/incremental-sync.md
```diff
@@ -0,0 +1,133 @@
...
+
+The `from-date` is set to the cutoff date of articles synced already and the `to-date` is set to the new "now".
+
+## Advanced settings
+
+The description above is sufficient for a lot of APIs. However there are some more subtle configurations which sometimes become relevant. 
+
+### Step
+
+When incremental syncs are enabled, the connector is not fetching all records since the cutoff date at once - instead it's splitting up the time range between the cutoff date and the desired end date into intervals based on the "Step" configuration (by default it's set to 1 month)
```
clnoll, line 89: > by default it's set to 1 month

How do we define "1 month"? Given that months are variable numbers of days and there are edge cases that may not have obvious defaults (e.g. what is 1 month from Jan 31?), I wonder if it would be preferable to choose another default (a week?) instead.

docs/connector-development/connector-builder-ui/incremental-sync.md
```diff
@@ -0,0 +1,133 @@
...
+
+curl 'https://content.guardianapis.com/search?order-by=oldest&from-date=2023-01-20T00:00:00Z&to-date=2023-01-30T00:00:00Z'
+
+...
+```
+
+After an interval is processed, the cursor value of the last record will be saved as part of the connection as the new cutoff date.
+
+In most cases, the default step size is fine, there are two reasons to change it:
+* **The API is unreliable** and the cutoff date should be saved more often to prevent resync of a lot of records - if the "Step" size is a day, then at most one day worth of data needs to be resync in case the sync fails halfway through. However, a smaller step size might cause more requests to the API and more load on the system.
```
clnoll, line 106:
```suggestion
* **The API is unreliable** we may want to reduce the step size. This results in more frequent saving of the "most recent" date, which prevents resync of a lot of records - if the "Step" size is a day, then at most one day worth of data needs to be resynced if the sync fails halfway through. However, it should be noted that a smaller step size might cause more requests to the API and more load on the system.
```

docs/connector-development/connector-builder-ui/incremental-sync.md
```diff
@@ -0,0 +1,133 @@
...
+* **The API requires the connector to fetch data in pre-specified chunks** - for example the [Exchange Rates API](https://exchangeratesapi.io/documentation/) makes the date to fetch data for part of the URL path and only allows to fetch data for a single day at a time
+
+### Lookback window
+
+The "Lookback window" specifies a duration that is subtracted from the last cutoff date before starting to sync.
+
+Same APIs update records over time but do not allow to filter or search by modification date, only by creation date. For example the API of The Guardian might change the title of an article after it got published, but the `webPublicationDate` still shows the original date the article got published initially.
```
clnoll, line 113:
```suggestion
Some APIs update records over time but do not allow to filter or search by modification date, only by creation date. For example the API of The Guardian might change the title of an article after it got published, but the `webPublicationDate` still shows the original date the article got published initially.
```

docs/connector-development/connector-builder-ui/incremental-sync.md
```diff
@@ -0,0 +1,133 @@
...
+To learn more about how different modes of incremental syncs, check out the [Incremental Sync - Append](/understanding-airbyte/connections/incremental-append/) and [Incremental Sync - Deduped History](/understanding-airbyte/connections/incremental-deduped-history) pages.
+
+## Configuration
+
+To configure incremental syncs for a stream in the connector builder, you have to specify how the records specify the **"last changed" / "updated at" timestamp**, the **initial time range** to fetch records for and **how to request records from a certain time range**.
+
+In the builder UI, these things are specified like this:
+* The "Cursor field" is the property in the record that defines the date and time when the record got changed. It's used to decide which records are synced already and which records are "new"
+* The "Datetime format" specifies the [format](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes) the cursor field is using to specify date and time,
+* The "Cursor granularity" is the smallest time unit supported by the API to specify the time range to request records for
```
girarda, line 22: this should either have a link to some external docs or a description / enumeration of the values that are accepted. `PT1S` doesn't mean much without context

docs/connector-development/connector-builder-ui/incremental-sync.md
```diff
@@ -0,0 +1,133 @@
...
+* The "Start datetime" is the initial start date of the time range to fetch records for. When doing incremental syncs, the second sync will overwrite this date with the last record that got synced so far. In most cases, it is defined by the end user when configuring a Source using your connector.
+* The "End datetime" is the end date of the time range to fetch records for. In most cases it's set to the current date and time when the sync is started to sync all changes that happened so far.
```
girarda, line 24: I think it's worth mentioning that `{{ now_utc() }}` is how the value is set to the current date and time

docs/connector-development/connector-builder-ui/partitioning.md
```diff
@@ -0,0 +1,116 @@
+# Partitioning
+
+Partitioning is required if the records of a stream are nested within multiple buckets or parent resources that need to be queried separately to extract the records.
+
+Sometimes records belonging to a single stream are partitioned into subsets that need to be fetched separately. In most cases, these partitions are a parent resource type of the resource type targeted by the connector. The partitioning feature can be used to configure your connector to iterate through all partitions. In API documentations, this concept is showing up as mandatory parameters that need to be set on the path, query or request body of the request.
+
+Common API structures look like this:
+* The [SurveySparrow API](https://developers.surveysparrow.com/rest-apis/response#getV3Responses) allows to fetch a list of responses to surveys. For the `/responses` endpoint, the id of the survey to fetch responses for needs to be specified via the query parameter `survey_id`. The API does not allow to fetch responses for all available surveys in a single request, there needs to be a separate request per survey. The surveys represent the partitions of the responses stream.
```
clnoll, line 8:
```suggestion
* The [SurveySparrow API](https://developers.surveysparrow.com/rest-apis/response#getV3Responses) allows users to fetch a list of responses to surveys. For the `/responses` endpoint, the id of the survey to fetch responses for needs to be specified via the query parameter `survey_id`. The API does not allow to fetch responses for all available surveys in a single request, so there needs to be a separate request per survey. The surveys represent the partitions of the responses stream.
```

docs/connector-development/connector-builder-ui/incremental-sync.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Advanced settings
+
+The description above is sufficient for a lot of APIs. However there are some more subtle configurations which sometimes become relevant. 
+
+### Step
+
+When incremental syncs are enabled, the connector is not fetching all records since the cutoff date at once - instead it's splitting up the time range between the cutoff date and the desired end date into intervals based on the "Step" configuration (by default it's set to 1 month)
+
+For example if the "Step" is set to 10 days (`P10D`) for the Guardian articles stream described above and a longer time range, then the following requests will be performed:
```
girarda, line 91: Can we point to some docs or an explanation of what the values are? `P10D` is very cryptic

docs/connector-development/connector-builder-ui/incremental-sync.md
girarda, line 106: Can we focus on the checkpointing ability instead of blaming the API for being unreliable? This is also useful when Airbyte is responsible for the sync failure

docs/connector-development/connector-builder-ui/incremental-sync.md
```diff
@@ -0,0 +1,133 @@
...
+* **The API requires the connector to fetch data in pre-specified chunks** - for example the [Exchange Rates API](https://exchangeratesapi.io/documentation/) makes the date to fetch data for part of the URL path and only allows to fetch data for a single day at a time
+
+### Lookback window
+
+The "Lookback window" specifies a duration that is subtracted from the last cutoff date before starting to sync.
+
+Same APIs update records over time but do not allow to filter or search by modification date, only by creation date. For example the API of The Guardian might change the title of an article after it got published, but the `webPublicationDate` still shows the original date the article got published initially.
+
+In these cases, there are two options:
+* **Do not use incremental sync** and always sync the full set of records to always have a consistent state - depending on the amount of data this might not be feasable
```
girarda, line 116: nit: this is also undesirable for some use cases even if the amount of data is small. This prevents the user from keeping a history of the changes (dedup + history)

docs/connector-development/connector-builder-ui/partitioning.md
```diff
@@ -0,0 +1,116 @@
...
+
+The first two options are a "static" form of partition routing (because the partitions won't change as long as the Airbyte configuration isn't changed). The API providing the partitions via one or multiple separate requests is a "dynamic" form of partition routing because the partitions can change any time.
+
+### List partition router
+
+To configure static partitioning, choose the "List" method for the partition router. The following fields have to be configured:
+* The "partition values" can either be set to a list of strings, making the partitions part of the connector itself or delegated to a user input so the end user configuring a Source based on the connector can control which partitions to fetch.
+* The "Current partition value identifier" can be freely choosen and is the identifier of the variable holding the current partition value. It can for example be used in the path of the stream using the `{{ stream_slice.<identifier> }}` syntax.
+* The "Inject partition value into outgoing HTTP request" option allows to configure how to add the current partition value to the requests
```
clnoll, line 26:
```suggestion
* The "Inject partition value into outgoing HTTP request" option allows to configure how to add the current partition value to the requests.
```

docs/connector-development/connector-builder-ui/partitioning.md
```diff
@@ -0,0 +1,116 @@
...
+
+#### Example
+
+To enable static partition routing for the [SurveySparrow API](https://developers.surveysparrow.com/rest-apis/response#getV3Responses) responses, the list partition router needs to be configured as following:
```
clnoll, line 30:
```suggestion
To enable static partition routing for the [SurveySparrow API](https://developers.surveysparrow.com/rest-apis/response#getV3Responses) responses, the list partition router needs to be configured as follows:
```

docs/connector-development/connector-builder-ui/partitioning.md
```diff
@@ -0,0 +1,116 @@
...
+* "Partition values" are set to the list of survey ids to fetch
+* "Current partition value identifier" is set to `survey` (this is not used for this example)
```
clnoll, line 32: I don't understand "(this is not used for this example)". Sounds like it's an implementation detail - can we remove it?

clnoll (approved):
Nice @flash1293! Just a few suggestions.

docs/connector-development/connector-builder-ui/incremental-sync.md
flash1293: I left that out because I want to make this the default setting in a simple dropdown as part of https://github.com/airbytehq/airbyte/issues/21927 (started with that today)

docs/connector-development/connector-builder-ui/incremental-sync.md
flash1293: We are using ISO 8601 durations (https://en.wikipedia.org/wiki/ISO_8601#Durations) so a month is a calendar month which can have varying length. I'm not sure about the perfect default, I think @girarda also put some thoughts into making this field optional and instead do a number-of-records based checkpointing which would be preferable as it makes it easier for the user.

However, AFAIK that's not planned for the immediate future  (before public beta) so I'm documenting the current state of affairs.

docs/connector-development/connector-builder-ui/incremental-sync.md
flash1293: I changed this paragraph based on @girarda s comments above

flash1293:
Thanks for the comments @girarda and @clnoll , I addressed them in the most recent version. Catherine, I've seen you made some comments on the partition page as well which was accidentally showing up in the diff, sorry for that. I will address them in a separate PR

girarda (approved):

docs/connector-development/connector-builder-ui/incremental-sync.md
```diff
@@ -0,0 +1,129 @@
+# Incremental sync
```
girarda, line 1: do we want to document the use of `stream_interval` anywhere?

flash1293: Good call, added an info box for that and updated the tutorial

PR #25239, opened 2023-04-17
performance demo

rodireich:


rodireich:
/connector-performance connector=connectors/source-postgres dataset=10m

PR #25280, opened 2023-04-18
Implement core xmin sync logic

### LINKED ISSUE #25752: Implement core sync logic for Xmin

akashkulk:
Initial MVP of xmin sync (with the option disabled in the UX so users cannot select it yet) : 
Closes https://github.com/airbytehq/airbyte/issues/25752

Recommended reading order : 
1. `XminPostgresSourceTest.java` : Unit test to understand core reading behavior
2. `XminStatus.java` : Object/data model representing state
3. `PostgresXminHandler.java` : Core logic to get incremental iterators for Xmin. Heavily adapted from the code in `AbstractDbSource`
4. All files under `../airbyte/integrations/source/relationaldb/state/` : To best expose the state in Postgres, I've added a "get raw state" functionality to the stream managers. This could be useful for future db-specific sync modes to access raw state to parse/interact with it. The other option was to add a Postgres-specific dependency in this package, which I decided against. Happy to discuss. 
5. `XminStateManager.java` and `XminRecordIterator.java` : Core logic to parse and create xmin state
6. `PostgresSource.java`, `PostgresUtils.java`, `PostgresCatalogHelper.java` : Contains logic to identify a xmin sync configuration, and constructing an xmin catalog.
7. Everything else

Work still to do : 
1. Enable the new xmin replication method in `spec.json` and writing an integration test. This has been written and validated locally, but needs the new spec change to be in for it to work. 
2. Handle xmin wraparound
3. Emit frequent state

Loom : https://www.loom.com/share/f28b031f7576432ca0a69cc2e95f3a7a


subodh1810 (commented):
Can we create a package within Postgres and move all the classes related to xmin in the class. We could do the same for CDC as well.

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/XminStateIterator.java
```diff
@@ -0,0 +1,77 @@
...
+      return endOfData();
+    }
+
+    if (messageIterator.hasNext()) {
+      // Use try-catch to catch Exception that could occur when connection to the database fails
+      try {
+        final AirbyteMessage message = messageIterator.next();
+        return message;
+      } catch (final Exception e) {
+        hasCaughtException = true;
```
subodh1810, line 66: Why are we swallowing the exception? We should fail the sync right?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/XminStateIterator.java
```diff
@@ -0,0 +1,77 @@
...
+        LOGGER.error("Message iterator failed to read next record.", e);
+        return endOfData();
+      }
+    } else if (!hasEmittedFinalState) {
+      hasEmittedFinalState = true;
+      return XminStateManager.createStateMessage(pair, xminStatus);
```
subodh1810, line 72: Why are we not emitting state every 10K records or 15 minutes?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -425,6 +439,10 @@ public List<AutoCloseableIterator<AirbyteMessage>> getIncrementalIterators(final
           AutoCloseableIterators.concatWithEagerClose(AirbyteTraceMessageUtility::emitStreamStatusTrace, snapshotIterator,
               AutoCloseableIterators.lazyIterator(incrementalIteratorSupplier, null)));
 
+    } else if (PostgresUtils.isXmin(sourceConfig) && isIncrementalSync(catalog)) {
+      final XminStateManager xminStateManager = new XminStateManager(stateManager.getRawStateMessages());
+      final PostgresXminHandler handler = new PostgresXminHandler(database, sourceOperations, getQuoteString(), xminStatus, xminStateManager);
```
subodh1810, line 444: There is a chance that xminStatus could be null at this point right?

subodh1810 (changes requested):

subodh1810 (changes requested):
Did an initial pass on the PR

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/XminStateIterator.java
akashkulk: Good catch - changed this to throw an exception now. When we do move to emitting intermediate state this can be changed to return `endofData()`

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
akashkulk: Actually, we are saving `xminStatus` in `logPresyncDebugData`, which runs prior to `getIncrementalIterators()` so this will not be null

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
```diff
@@ -237,13 +247,20 @@ public AirbyteCatalog discover(final JsonNode config) throws Exception {
...
+          .map(PostgresCatalogHelper::setIncrementalToSourceDefined)
+          .map(PostgresCatalogHelper::addCdcMetadataColumns)
           // If we're in CDC mode and a stream is not in the publication, the user should only be able to sync
           // this in FULL_REFRESH mode
-          .map(stream -> PostgresCdcCatalogHelper.setFullRefreshForNonPublicationStreams(stream, publicizedTablesInCdc))
+          .map(stream -> PostgresCatalogHelper.setFullRefreshForNonPublicationStreams(stream, publicizedTablesInCdc))
+          .collect(toList());
+
+      catalog.setStreams(streams);
+    } else if (PostgresUtils.isXmin(config)) {
```
rodireich, line 260: This means that no existing incremental connection will automatically switch to xmin, right?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresQueryUtils.java
```diff
@@ -68,14 +69,16 @@ public class PostgresQueryUtils {
    * value of the xmin snapshot (which is a combination of 1 and 2). If no wraparound has occurred,
    * this should be the same as 2.
    */
-  public static void logXminStatus(final JdbcDatabase database) throws SQLException {
+  public static XminStatus getXminStatus(final JdbcDatabase database) throws SQLException {
     LOGGER.debug("xmin status query: {}", XMIN_STATUS_QUERY);
     final List<JsonNode> jsonNodes = database.bufferedResultSetQuery(conn -> conn.prepareStatement(XMIN_STATUS_QUERY).executeQuery(),
```
rodireich, line 74: Can there be other system columns that may be useful even for just logging

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/xmin/PostgresXminHandler.java
```diff
@@ -0,0 +1,195 @@
...
+      final String sql = String.format("SELECT %s FROM %s WHERE xmin::text::bigint > ?",
+          wrappedColumnNames, fullTableName);
+
+      final PreparedStatement preparedStatement = connection.prepareStatement(sql.toString());
+
+      final XminStatus currentStreamXminStatus = xminStateManager.getXminStatus(airbyteStream);
+      if (currentStreamXminStatus != null) {
+        preparedStatement.setLong(1, currentStreamXminStatus.getXminXidValue());
+      } else {
+        preparedStatement.setLong(1, 0L);
```
rodireich, line 146: This would effectively be a full refresh sync? is this expected to ever happen?
Maybe log a warning here?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/xmin/PostgresXminHandler.java
```diff
@@ -0,0 +1,195 @@
...
+        });
+  }
+
+  private AutoCloseableIterator<AirbyteMessage> augmentWithState(final AutoCloseableIterator<AirbyteMessage> recordIterator,
+      final AirbyteStreamNameNamespacePair pair) {
+    return  AutoCloseableIterators.transform(
+        autoCloseableIterator -> new XminStateIterator(
+            autoCloseableIterator,
+            pair,
+            xminStatus),
```
rodireich, line 191: Is `xminStatus` ever changing?
I think this would just put the last saved `xminStatus` back into a state message, no?

airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/XminPostgresSourceTest.java
```diff
@@ -0,0 +1,285 @@
...
+
+  @Test
+  void testReadSuccess() throws Exception {
+    // Perform an initial sync with the configured catalog, which is set up to use xmin_replication.
+    // All of the records in the configured stream should be emitted.
+    final ConfiguredAirbyteCatalog configuredCatalog =
+        CONFIGURED_XMIN_CATALOG
+            .withStreams(CONFIGURED_XMIN_CATALOG.getStreams().stream().filter(s -> s.getStream().getName().equals(STREAM_NAME)).collect(
+                Collectors.toList()));
+    final List<AirbyteMessage> actualMessages =
```
rodireich, line 205: For my understanding, what is the initial state ?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/xmin/XminStateIterator.java
```diff
@@ -0,0 +1,78 @@
...
+      // This occurs when the previous iteration emitted state so this iteration cycle will indicate
+      // iteration is complete
+      hasCaughtException = false;
+      return endOfData();
+    }
+
+    if (messageIterator.hasNext()) {
+      // Use try-catch to catch Exception that could occur when connection to the database fails
+      try {
+        final AirbyteMessage message = messageIterator.next();
```
subodh1810, line 63: We could return directly from here

subodh1810 (commented):
@akashkulk I pushed a commit to use generated POJO from YAML.

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSource.java
akashkulk: That's correct. It's a new sync mode that users will have to choose. There isn't currently a migration path

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresQueryUtils.java
akashkulk: Which ones were you thinking of? We can always add more here later if you come up with them afterwards

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/xmin/PostgresXminHandler.java
akashkulk: If a user adds a new stream, there won't be any xmin state associated with that stream and we'll have to issue a full refresh in that case (for that particular table)

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/xmin/PostgresXminHandler.java
akashkulk: `xminStatus` is being calculated at the beginning of each sync in `logPresyncDebugData`. So, assuming the stream successfully synced this would indeed emit a new state message (different from the last saved `xminStatus`)

airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/XminPostgresSourceTest.java
akashkulk: There are 3 reads (syncs) here : 
1. Sync 1 : no state (starting from scratch)
2. Sync 2 : State from sync 1 (the xmin that was calculated then)
3. Sync 3 : State from sync 2

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/xmin/PostgresXminHandler.java
rodireich: OK I see
My concern was with mid sync checkpoint but I see from another response above that we only emit a state by the end because there's no guaranteed order.

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/XminStateIterator.java
akashkulk: Actually, thinking more on this - we should throw endOfData(). This stream may not sync successfully in the v0 case (since only one state message is emitted at the end), but we don't want one stream's failure to cause all subsequent syncs to fail

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/xmin/PostgresXminHandler.java
rodireich: So in this case it is full refresh by xmin.
Does it differ from "regular" full refresh?
Does it matter that we go into a different flow?

akashkulk: Logically, it is the same flow as the vanilla full refresh flow. It's just contained here when we do also want to query `xmin` status with every query for intermediate state checkpointing in future PRs

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/xmin/PostgresXminHandler.java
```diff
@@ -0,0 +1,196 @@
...
+    try {
+      LOGGER.info("Preparing query for table: {}", tableName);
+      final String fullTableName = getFullyQualifiedTableNameWithQuoting(schemaName, tableName,
+          quoteString);
+
+      final String wrappedColumnNames = RelationalDbQueryUtils.enquoteIdentifierList(columnNames, quoteString);
+      final String sql = String.format("SELECT %s FROM %s WHERE xmin::text::bigint > ?",
+          wrappedColumnNames, fullTableName);
+
+      final PreparedStatement preparedStatement = connection.prepareStatement(sql.toString());
```
subodh1810, line 141: `sql.toString()` not required. `sql` is already string

subodh1810 (approved):

rodireich (approved):

Issue #24892, opened 2023-04-05
Connector builder / probably all of low code CDK: AddFields transformation throws exception


flash1293:
When trying to define an `AddFields` transformation, every test read fails with an exception.

```
    transformations:
      - type: AddFields
        fields:
          - path:
              - two_times_two
            value: '{{ 2 * 2 }}'
```

```
'AirbyteMessage' object does not support item assignment - Traceback (most recent call last):
  File "/home/connector-builder-server/./connector_builder/impl/low_code_cdk_adapter.py", line 63, in read_stream
    yield from generator
  File "/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/declarative/manifest_declarative_source.py", line 119, in read
    yield from super().read(logger, config, catalog, state)
  File "/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/abstract_source.py", line 130, in read
    raise e
  File "/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/abstract_source.py", line 116, in read
    yield from self._read_stream(
  File "/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/abstract_source.py", line 185, in _read_stream
    for record in record_iterator:
  File "/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/abstract_source.py", line 303, in _read_full_refresh
    for record_data_or_message in record_data_or_messages:
  File "/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/declarative/declarative_stream.py", line 101, in read_records
    yield self._apply_transformations(record, self.config, stream_slice)
  File "/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/declarative/declarative_stream.py", line 106, in _apply_transformations
    output_record = transformation.transform(record, config=config, stream_state=self.state, stream_slice=stream_slice)
  File "/usr/local/lib/python3.9/site-packages/airbyte_cdk/sources/declarative/transformations/add_fields.py", line 115, in transform
    dpath.util.new(record, parsed_field.path, value)
  File "/usr/local/lib/python3.9/site-packages/dpath/util.py", line 65, in new
    return dpath.segments.set(obj, segments, value)
  File "/usr/local/lib/python3.9/site-packages/dpath/segments.py", line 328, in set
    current[segments[-1]] = value
TypeError: 'AirbyteMessage' object does not support item assignment
```

girarda:
[...]

PR #25290, opened 2023-04-18
low-code: Do not apply transforms on AirbyteLogMessages and AirbyteTraceMessages

girarda:
## What
* Resolves issue #24892
* The AddFields's transform method fails when emitting connector builder messages because it cannot add fields to `AirbyteLogMessage`s
* Solution is for the `DeclarativeStream` to only apply transformations when on record data and ignore `AirbyteLogMessage`s and `AirbyteTraceMessage`s

## How
* Update `_apply_transformations` to make it clear it accepts AirbyteLogMessage and AirbyteTraceMessage in addition to Mapping
* only apply transforms if the input is a dict

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_stream.py`
2. `airbyte-cdk/python/unit_tests/sources/declarative/test_declarative_stream.py`
3. `airbyte-cdk/python/unit_tests/sources/declarative/test_manifest_declarative_source.py`


clnoll (approved):
LGTM!

PR #25296, opened 2023-04-18
Ignore expired certificate

clnoll:
Ignore expired certificate for https://packages.microsoft.com/keys/microsoft.asc.


evantahler (changes requested):
I think rather than check this in to skip a temporary HTTPS cert problem MS is having, I'd rather merge in the other PRs without this test passing.  We can do that!

clnoll:
Makes sense to me @evantahler! Actually I noticed that a PR of mine just auto-merged, which means that the build ultimately passed - so I imagine it's all good now for the other build that this was blocking, too. Closing this PR.

PR #25307, opened 2023-04-19
Dagster create new connectors report

### LINKED ISSUE #25190: Export connectors.md information file to GCS for data team
I would like to map GitHub oncall issues to specific connectors. 
The best way to do this is by mapping the GitHub label to a connector id. 

@evantahler Has created this mapping [here](https://github.com/airbytehq/airbyte/blob/master/connectors.md) 🎉 
(Thanks Evan!)

The next step is to export this file, possibly as json, to a GCS bucket so that we can use it in the data warehouse.
Ideally, the GitHub label could be an attribute on the Connector Metadata json file that we already have in the warehouse.
Open to any other ideas here!

We've tried parsing the GH label and joining on connector name, but it is not reliable. 

## Additional context
The primary use case is for tracking connector issues on a dashboard that @AnilEMaharjan  is working on

### LINKED ISSUE #25246: Merge builds.md and connectors.md into one report
## What this is
Creating a new report, generated by dagster, that merges the functionality of builds.md and connectors.md

and adds the following info
1. Where is each deployed
2. What version is deployed
3. What CDK is used? (if possible)

## What this is not
1. Deleting these files

## What would be extra awesome
1. If we can share this from connectors.airbyte.com

bnchrch:
## What
1. Adds a new report that is a fusion of builds.md and connectors.md
2. Generates the report automatically on legacy catalog changes ([example](https://storage.googleapis.com/ben-ab-test-bucket/generated_reports/connector_registry_report.html))
3. Updates our asset imports to be dynamic 🎉 
4. Adds a new GCSFileManager so that we can set the appropriate content type. This lets us view the html url as opposed to just download it
5. Move legacy_registry to its own file

closes #25190 
closes #25246


airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
```diff
@@ -1,99 +1,43 @@
...
-    oss_registry_diff_report,
-    oss_registry_diff,
-    oss_registry_from_metadata,
-    oss_sources_dataframe,
-    overrode_metadata_definitions,
-    persist_metadata_definitions,
-    specs_secrets_mask_yaml,
-    valid_metadata_report_dataframe,
-]
+ASSETS = load_assets_from_modules(
```
alafanechere, line 28: 👍 . But according to Dagster simple example [here](https://docs.dagster.io/tutorial/writing-your-first-asset) they never mention the list of exposed assets should be explicitly declared. I looks mandatory in our codebase though, because I spent a fair amount of time don't understanding why my new assets was not available in the UI.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/legacy_registry.py
alafanechere, line 1: 👍

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/legacy_registry.py
```diff
@@ -0,0 +1,66 @@
...
+from dagster import asset, OpExecutionContext
+
+
+from orchestrator.utils.dagster_helpers import OutputDataFrame, output_dataframe
+
+
+from metadata_service.models.generated.ConnectorRegistryV0 import ConnectorRegistryV0
+
+
+GROUP_NAME = "legacy_registry"
```
alafanechere, line 16: nit: if we're good at naming assets modules we could use `__name__` instead of repeating the module name
```suggestion
GROUP_NAME = __name__
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
```diff
@@ -1,15 +1,98 @@
...
 
-# HELPERS
+OSS_SUFFIX = "_oss"
+CLOUD_SUFFIX = "_cloud"
+
+
+# 🔗 HTML Renderers
+
+
+def simple_link_html(url):
```
alafanechere, line 18:
```suggestion
def simple_link_html(url: str) -str:
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
```diff
@@ -1,15 +1,98 @@
...
+    if url:
+        return f'<a href="{url}" target="_blank">🔗 Link</a>'
+    else:
+        return None
+
+
+def icon_image_html(icon):
```
alafanechere, line 25:
```suggestion
def icon_image_html(icon: str) -> str:
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
```diff
@@ -1,15 +1,98 @@
...
+OSS_SUFFIX = "_oss"
+CLOUD_SUFFIX = "_cloud"
+
+
+# 🔗 HTML Renderers
+
+
+def simple_link_html(url):
+    if url:
+        return f'<a href="{url}" target="_blank">🔗 Link</a>'
```
alafanechere, line 20:
```suggestion
        return f"<a href='{url}' target='_blank'>🔗 Link</a>"
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
```diff
@@ -1,15 +1,98 @@
...
+def icon_image_html(icon):
+    github_icon_base = (
+        f"https://raw.githubusercontent.com/{CONNECTOR_REPO_NAME}/master/airbyte-config-oss/init-oss/src/main/resources/icons"
+    )
+    icon_size = "30"
+    icon_link = f'<img src="{github_icon_base}/{icon}" height="{icon_size}" height="{icon_size}"/>' if icon else "x"
+    return icon_link
+
+
+def test_badge_html(test_summary_url):
```
alafanechere, line 34:
```suggestion
def test_badge_html(test_summary_url: str) -> str:
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
```diff
@@ -1,15 +1,98 @@
...
+    if not test_summary_url:
+        return None
+
+    image_shield_base = "https://img.shields.io/endpoint"
+    test_summary_url_encoded = urllib.parse.quote(test_summary_url)
+    return f'<img src="{image_shield_base}?url={test_summary_url_encoded}">'
```
alafanechere, line 40:
```suggestion
    return f"<img src='{image_shield_base}?url={test_summary_url_encoded}'>"
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
```diff
@@ -1,15 +1,98 @@
...
+
+    return f"{CONNECTORS_TEST_RESULT_BUCKET_URL}/tests/summary/connectors/{connector}/badge.json"
+
+
+# 📊 Dataframe Augmentation
 
 
 def augment_and_normalize_connector_dataframes(
-    cloud_df, oss_df, primaryKey, connector_type, valid_metadata_report_dataframe, github_connector_folders, cached_specs
+    cloud_df: pd.DataFrame, oss_df: pd.DataFrame, primaryKey: str, connector_type: str, github_connector_folders: List[str]
```
alafanechere, line 95:
```suggestion
    cloud_df: pd.DataFrame, oss_df: pd.DataFrame, primary_key: str, connector_type: str, github_connector_folders: List[str]
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
```diff
@@ -22,148 +105,185 @@ def augment_and_normalize_connector_dataframes(
...
 
+    total_registry["issue_url"] = total_registry.apply(issue_url, axis=1)
+    total_registry["test_summary_url"] = total_registry.apply(test_summary_url, axis=1)
 
-# ASSETS
-
+    # Merge docker repo and version into separate columns
+    total_registry["docker_image_oss"] = total_registry.apply(lambda x: merge_docker_repo_and_version(x, OSS_SUFFIX), axis=1)
+    total_registry["docker_image_cloud"] = total_registry.apply(lambda x: merge_docker_repo_and_version(x, CLOUD_SUFFIX), axis=1)
+    total_registry["docker_images_match"] = total_registry.apply(lambda x: x["docker_image_oss"] == x["docker_image_cloud"], axis=1)
```
alafanechere, line 128:
```suggestion
    total_registry["docker_images_match"] = total_registry["docker_image_oss"] == total_registry["docker_image_cloud"]
```
Apply operations on DF are expensive, it's better to avoid them when possible.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
```diff
@@ -22,148 +105,185 @@ def augment_and_normalize_connector_dataframes(
...
-        valid_metadata_report_dataframe=valid_metadata_report_dataframe,
         github_connector_folders=github_connector_folders,
-        cached_specs=cached_specs,
     )
 
 
 @asset(group_name=GROUP_NAME)
 def all_destinations_dataframe(
-    cloud_destinations_dataframe, oss_destinations_dataframe, github_connector_folders, valid_metadata_report_dataframe, cached_specs
+    legacy_cloud_destinations_dataframe, legacy_oss_destinations_dataframe, github_connector_folders
```
alafanechere, line 182: As we might want to switch this back to the not legacy version soon, what do you think of:
- exposing a like `USE_LEGACY_REGISTRY` in this module,
- make `cloud_destinations_dataframe` and `oss_destinations_dataframe` assets  return the legacy DF if `USE_LEGACY_REGISTRY is True`.

#feature flags 😆

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/jobs/registry.py
```diff
@@ -3,10 +3,10 @@
 registries_inclusive = AssetSelection.keys(
     "metadata_directory_report", "cloud_registry_from_metadata", "oss_registry_from_metadata"
 ).upstream()
-registry_reports_inclusive = AssetSelection.keys("connector_registry_location_html", "connector_registry_location_markdown").upstream()
+registry_reports_inclusive = AssetSelection.keys("connector_registry_report").upstream()
```
alafanechere, line 6: Could you explain the purpose of this?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/resources/gcp.py
```diff
@@ -47,7 +92,7 @@ def gcs_file_manager(resource_context) -> GCSFileManager:
 
     storage_client = resource_context.resources.gcp_gcs_client
 
-    return GCSFileManager(
+    return ContentTypeAwareGCSFileManger(
```
alafanechere, line 95:
```suggestion
    return ContentTypeAwareGCSFileManager(
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/resources/gcp.py
```diff
@@ -4,7 +4,52 @@
...
+        check_file_like_obj(file_obj)
+        gcs_key = self.get_full_key(key + (("." + ext) if ext is not None else ""))
+        bucket_obj = self._client.bucket(self._gcs_bucket)
+        blob = bucket_obj.blob(gcs_key)
+        blob.content_type = self.get_content_type(ext)
+        blob.upload_from_file(file_obj)
+        return GCSFileHandle(self._gcs_bucket, gcs_key)
+
+
+def get_public_url_for_gcs_file_handle(file_handle: GCSFileHandle):
```
alafanechere, line 51: nit: wdyt about subclassing `GCSFileHandle` to add a `public_url` property?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/sensors/registry.py
```diff
@@ -24,8 +24,9 @@ def registry_updated_sensor_definition(context: SensorEvaluationContext):
             etag_cursor = context.cursor or None
             context.log.info(f"Old etag cursor: {etag_cursor}")
 
+            # TODO (ben) stop using legacy resources when we remove the legacy registry
```
alafanechere, line 27: Same as priror suggestion, can we use a "feature flag" like approach?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
alafanechere, line 1: Could we not use `df.to_html` and instead:
- pass a list of dict to the jinja templates
- keep the html formatting logic in jinja templates

The templating currently happens both in the python code and in the jinja templates.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/templates/connector_registry_locations.html
alafanechere, line 1: missing new line at the end of the file

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/templates/connector_registry_locations.html
```diff
@@ -1,12 +1,16 @@
...
-{% block title%}Connector Registries{% endblock %}
+{% block title%}Airbyte Connector Registry{% endblock %}
 
 {% block content %}
-<h1>Connector Registries</h1>
+<h1>Airbyte Connector Registries</h1>
 <h2>Sources</h2>
-{{ sources_table }}
+<div class="table-container test">
+  {{ sources_table }}
```
alafanechere, line 9: I think we should pass a list of dict here and generate the table from this list of dict with a for loop in the template.

alafanechere (approved):
I made a lot of minor suggestions I'll let you decide whether to implement or not 😄

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
bnchrch: Ah nice catch!

So this is a failure of how Dagster does documentation IMO

You expect the decorators to wire everything up.

And their docs are paged in a way that makes it look like that.

But if you go further back in their tutorial you will see they call a project scaffold function
https://docs.dagster.io/tutorial/setup

Which creates an `__init__.py` file with this template:
https://github.com/dagster-io/dagster/blob/master/python_modules/dagster/dagster/_generate/templates/CODE_LOCATION_NAME_PLACEHOLDER/CODE_LOCATION_NAME_PLACEHOLDER/__init__.py

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/legacy_registry.py
bnchrch: ooooh I like that!

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
bnchrch: Very good idea!

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/jobs/registry.py
bnchrch: So the `.upstream` method tells dagster to materialize
1. Any asset matching the key
2. Any assets it depends on in the DAG

without it the report generation would fail or get stale data as it wouldn't materialize the spec cache, or the registry file data

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/legacy_registry.py
bnchrch: Ah actually this will render as `orchestrator.assets.legacy_registry` which is invalid.

We could parse it but that seems like more effort and bug surface area than its worth.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
bnchrch: Looking into it, we'd still have to update the sensor. Could still be flagged but I think this juice might not be worth the squeeze as its an informational report and not critical to rollback if something goes wrong

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
bnchrch: This is a good callout.

My only hesitation was that I didnt want to reinvent the header, alignment and "apply render function" logic that `to_html` already has.

So I think I found a good inbetween by moving the small html components into the render.py function

In the future I imagine these naturally becoming jinja macros/templates as the use case expands

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
alafanechere: Thank you for clarifying!

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_report.py
alafanechere: Sounds reasonable if pandas already provide you with more  boilerplate html than the one you have to write.

PR #25309, opened 2023-04-19
Source Hubspot: update expected records

arsenlosenko:
## What
Fix failed build for Source Hubspot

## How
* values of `amount` and `amount_in_home_currency` of field `properties` in `deals_archived` stream records became null
* value of `hs_time_in_1` field in `tickets` stream records is changed


lazebnyi (approved):
Please delete not necessary info from PR body.

PR #25311, opened 2023-04-19
Use proper release tag when copying files

joeykmh:
## What
The [run-ab-platform.sh](https://github.com/airbytehq/airbyte/blob/master/run-ab-platform.sh) script copies some files from `airbytehq/airbyte-platform` before running `docker compose up`. When copying those files, we were referencing the `main` branch when really we should be referencing the tag for the current release. This PR fixes the script to copy files from the current release.

## How
- Checks the version specified in `.env` of this repo
- Copies files from `airbytehq/airbyte` using the tag `v${version}` (e.g. `v0.44.0`)

### DIFF
--- a/run-ab-platform.sh
+++ b/run-ab-platform.sh
@@ -4,6 +4,8 @@
 set -o nounset # -u exit if a variable is not set
 set -o errexit # -f exit for any command failure
 
+# Get the version specified in the .env file
+version_tag="v$(cat .env | grep ^VERSION | cut -d '=' -f 2)"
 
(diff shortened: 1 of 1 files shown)


timroes (approved):
This makes more sense to me like that. Script looks good and seems to be working. Would appreciate a review from prodeng before merging this.

c-p-b (approved):
This is a pretty good workaround for now, given our current OSS release strategy. I think as we move into rethinking how we handle releases we should probably revisit this, but it solves the immediate pain for users now.

Thanks Joey!

PR #25315, opened 2023-04-19
source-amplitude: Only fetch date_time_fields once per request

clnoll:
## What
Refactor so we're only fetch `date_time_fields` once per `EventsExtractor` instance, instead of once per record.

The `_get_date_time_items_from_schema` method that fetches `date_time_fields` is doing a lot of work - instantiating a `JsonFileSchemaLoader` which reads the schema, so hopefully this will help alleviate memory pressure issues.


artem1205 (approved):

evantahler (approved):
Nice find!

girarda (approved):

PR #25316, opened 2023-04-19
connectors-ci: publish connectors on merge

### LINKED ISSUE #22170: Check Images Exist action incorrectly reporting successful status
## Current Behavior

An error in the check_images_exist.sh script is causing all checks to be reported as successful. 

This was originally reported in[ this slack thread ](https://airbytehq-team.slack.com/archives/C03VDJ4FMJB/p1675197085141189), where a normalization image had not been published but the check was still being reported as successful.

## Expected Behavior

The check should fail if the image tag does not exist on docker hub. We should also modify the script to report a failure if there was any sort of error when executing the script (such as these expected integer errors)

## Logs

```
URL: https://hub.docker.com/v2/repositories/airbyte/connector-builder-server/tags/0.40.32
./tools/bin/check_images_exist.sh: line 67: test: -lt: unary operator expected
./tools/bin/check_images_exist.sh: line 62: test: : integer expression expected
	STATUS: found
./tools/bin/check_images_exist.sh: line 67: test: -lt: unary operator expected

	airbyte/proxy: 0.40.32
	URL: https://hub.docker.com/v2/repositories/airbyte/proxy/tags/0.40.32
./tools/bin/check_images_exist.sh: line [62](https://github.com/airbytehq/airbyte/actions/runs/4058177408/jobs/6984795147#step:3:63): test: : integer expression expected
./tools/bin/check_images_exist.sh: line 67: test: -lt: unary operator expected
	STATUS: found

Checking Normalization images exist...
Checking normalization images with version 0.3.2 exist...
	airbyte/normalization: 0.3.2
	URL: https://hub.docker.com/v2/repositories/airbyte/normalization/tags/0.3.2
./tools/bin/check_images_exist.sh: line 62: test: : integer expression expected
./tools/bin/check_images_exist.sh: line [67](https://github.com/airbytehq/airbyte/actions/runs/4058177408/jobs/6984795147#step:3:68): test: -lt: unary operator expected
	STATUS: found
```

Example run https://github.com/airbytehq/airbyte/actions/runs/4058177408/jobs/6984795147

### LINKED ISSUE #25100: Update the connector image build pipe to publish to dockerhub
## Prerequisites
1. Provision (or use an existing) test registry for docker images to test with.

## What this is
1. Adding a new airbyte-ci command to publish a docker image for a connecter `airbyte-ci connectors build-image --connector <name>`
(https://github.com/airbytehq/airbyte/blob/172bf93ac427a06bccb0c3b72bcabee314e18370/tools/integrations/manage.sh#L238-L239) for details on how we currently do this)
2. This command should upload the images to docker hub and ensure they are available

alafanechere:
## What
Closes airbytehq/airbyte#25100 
Closes airbytehq/airbyte#22170
This PR re-implements the [`manage.sh`](https://github.com/airbytehq/airbyte/blob/master/tools/integrations/manage.sh#L1) logic with a Dagger pipeline.

It exposes a new CLI command:
`airbyte-ci connectors --name=source-pokeapi publish`

The new GHA workflow in this PR will be triggered on `metadata.yaml` file change on master. 
I've set the workflow to only publish pre-release images in order to test it at scale and not deprecate /publish too early.

## How
The publish pipelines accomplishes the following steps:

1. Validate the metadata file.
2. Check if the connector image already exists.
3. Build the connector, with platform variants.
4. Upload the connector spec to the spec cache bucket.
5. Push the connector to DockerHub, with platform variants.
6. Upload its metadata file to the metadata service bucket.

## Recommended reading order
* The 🥩 of the publish logic is [here](https://github.com/airbytehq/airbyte/blob/c4ae341869710a01ff759c10e7cda01b5d808062/tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py#L160)
* The new GHA workflow which is calling the `publish`command is [here](https://github.com/airbytehq/airbyte/blob/c4ae341869710a01ff759c10e7cda01b5d808062/.github/workflows/publish_connectors.yml#L1).

## 🚨 User Impact 🚨
I've set the workflow to only publish pre-release images in order to test it at scale and not deprecate /publish too early. No impact for developers or on the connector release path should happen.


tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+class PushConnectorImageToRegistry(PublishStep):
+    title = "Push connector image to registry"
+
+    async def _run(self, built_containers_per_platform: List[Container]) -> StepResult:
+        try:
+            image_ref = await built_containers_per_platform[0].publish(
+                f"docker.io/{self.docker_image_name}", platform_variants=built_containers_per_platform[1:]
+            )
```
evantahler, line 85: How does this work? I thought you couldn't have multiple arch variants of the same connector locally.  Are you renaming them?

alafanechere: In this situation the connector does not have actual images in the user's docker host. These are dagger containers objects. Container objects are lazily evaluated. The `publish` function actually perform the `docker build`.
But it's dagger directly interacting with buildkit to compose the image, it's never materialized as a real image on the host. It's why I implemented [LoadContainerToLocalDockerHost](https://github.com/airbytehq/airbyte/blob/3dd75291fa51b4e1608e6d84d58986039c748e88/tools/ci_connector_ops/ci_connector_ops/pipelines/builds/common.py#L24) to make these connector images built with dagger available on the user's docker host. But it happens only on `build` not on `publish` command.

TL;DR: When calling `publish` on a Dagger container you can pass other platform variants containers. Dagger is internally responsible for the `buildx, manifest, push etc.` steps we originally have in `manage.sh`.

tools/ci_connector_ops/ci_connector_ops/pipelines/builds/__init__.py
```diff
@@ -46,3 +47,24 @@ async def run_connector_build(context: ConnectorContext) -> dict[str, Tuple[Step
         per_platform_containers[build_platform] = await BuildConnectorImage(context, build_platform).run()
 
     return per_platform_containers
+
+
+async def run_connector_build_pipeline(context: ConnectorContext, semaphore: anyio.Semaphore) -> ConnectorReport:
```
alafanechere, line 52: I just moved this function here but did not change its code.

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
```diff
@@ -193,5 +198,65 @@ def build(ctx: click.Context) -> bool:
...
+    help="The name of the GCS bucket where specs will be cached.",
+    type=click.STRING,
+    required=True,
+    envvar="SPEC_CACHE_BUCKET_NAME",
+)
+@click.pass_context
+def publish(ctx: click.Context, pre_release: bool, spec_cache_service_account_key: str, spec_cache_bucket_name: str):
+    if ctx.obj["is_local"]:
+        click.confirm(
+            "Publishing from a local environment is not recommend and require to be logged in to Airbyte's DockerHub registry, do you want to continue?",
```
alafanechere, line 226: Other env variable likes `SPEC_CACHE_BUCKET_NAME` and `SPEC_CACHE_SERVICE_ACCOUNT_KEY` are required, but the CLI will fail beforehand if these are not set.

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
```diff
@@ -121,6 +124,8 @@ def connectors(
         click.secho("No connector were selected according to your inputs. Please double check your filters.", fg="yellow")
         sys.exit(0)
 
+    ctx.obj["modified_connectors_in_commit"] = get_modified_connectors(ctx.obj["modified_files_in_commit"])
+    ctx.obj["modified_connectors_name_in_commit"] = [c.technical_name for c in ctx.obj["modified_connectors_in_commit"]]
```
alafanechere, line 128: On publish, as we're on `master`, we care about the connectors that got modified in the `head` commits, not on the branch.

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/connectors.py
```diff
@@ -3,97 +3,24 @@
...
-        return context.report
-
-
-async def run_connectors_build_pipelines(contexts: List[ConnectorContext], concurrency: int = 5):
-    """Run a build pipeline for all the connector contexts.
-
-    Args:
-        contexts (List[ConnectorContext]): List of connector contexts for which a build pipeline needs to be run.
-        concurrency (int): Number of test pipeline that can run in parallel. Defaults to 5
-    """
```
alafanechere, line 93: I moved these function to their specific modules: `builds`, `tests`

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -0,0 +1,206 @@
+#
```
alafanechere, line 1: This module holds the 🥩 of the publish logic. I tried to replicate in Python the [`manage.sh` steps](https://github.com/airbytehq/airbyte/blob/cda2b7dde2495a49c95ca331ed11d2348feacb2c/tools/integrations/manage.sh)

.github/workflows/metadata_upload_manifest_dagger.yml
alafanechere, line 1: I deleted this workflow because the publish connectors workflow is responsible for the upload of metadata files to GCS.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -548,3 +547,23 @@ def with_normalization(context, normalization_dockerfile_name: str) -> Container
     normalization_directory_with_build = normalization_directory.with_new_directory("build")
     normalization_directory_with_sshtunneling = normalization_directory_with_build.with_file("build/sshtunneling.sh", sshtunneling_file)
     return normalization_directory_with_sshtunneling.docker_build(normalization_dockerfile_name)
+
+
+def with_airbyte_python_connector(context: ConnectorContext, build_platform: Platform):
```
alafanechere, line 552: @evantahler for consistency with Java connectors I decided to build Python with inline dagger instructions instead of reading their dockerfile. It means that we could eventually get rid of our connector Dockerfile.
If a developers needs to build a connector image locally they could just use:
`airbyte-ci connectors --name=source-faker build`

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -0,0 +1,232 @@
...
+
+    @property
+    def docker_image_name(self):
+        if self.pre_release:
+            return f"{self.context.docker_image_from_metadata}-dev.{self.context.git_revision[:10]}"
+        else:
+            return self.context.docker_image_from_metadata
+
+
+class CheckConnectorImageDoesNotExist(PublishStep):
```
bnchrch, line 35: Looks like we have some competing PRs!
https://github.com/airbytehq/airbyte/pull/25306

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -0,0 +1,232 @@
...
+
+class PublishStep(Step, ABC):
+    def __init__(self, context: ConnectorContext, pre_release: bool = True) -> None:
+        super().__init__(context)
+        self.pre_release = pre_release
+
+    @property
+    def docker_image_name(self):
+        if self.pre_release:
+            return f"{self.context.docker_image_from_metadata}-dev.{self.context.git_revision[:10]}"
```
bnchrch, line 30: 👏

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
bnchrch: In the very least its not bad to have backup logic :)

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -0,0 +1,232 @@
...
+    async def _run(self) -> StepResult:
+        manifest_inspect = (
+            environments.with_docker_cli(self.context)
+            .with_env_variable("CACHEBUSTER", str(uuid.uuid4()))
+            .with_exec(["docker", "manifest", "inspect", self.docker_image_name])
+        )
+        manifest_inspect_stderr = await with_stderr(manifest_inspect)
+        manifest_inspect_stdout = await with_stdout(manifest_inspect)
+
+        if "no such manifest" in manifest_inspect_stderr:
```
bnchrch, line 47: Any way to do this on an error code? or a message we define.

For example
```py
our_error_message = "SUPERUNIQUEERROR"
manifest_inspect = (
            environments.with_docker_cli(self.context)
            .with_env_variable("CACHEBUSTER", str(uuid.uuid4()))
            .with_exec(["docker", "manifest", "inspect", self.docker_image_name, "&&", "echo", our_error_message])
        )
        manifest_inspect_stderr = await with_stderr(manifest_inspect)
        manifest_inspect_stdout = await with_stdout(manifest_inspect)

        if our_error_message in manifest_inspect_stderr:
```

Definitely a hack, would still prefer an error code if possible

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
bnchrch: This is all very cool and much cleaner than buildx

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -0,0 +1,232 @@
...
+
+    @property
+    def cloud_spec_key(self):
+        return f"{self.spec_key_prefix}/{self.cloud_spec_file_name}"
+
+    @property
+    def oss_spec_key(self):
+        return f"{self.spec_key_prefix}/{self.default_spec_file_name}"
+
+    def __init__(self, context: ConnectorContext, pre_release: bool, spec_bucket_name: str, gcs_credentials: Secret) -> None:
```
bnchrch, line 137: nit: `__init__` should be at the top

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -0,0 +1,232 @@
...
+
+    async def _get_connector_spec(self, connector: Container, deployment_mode: str) -> str:
+        spec_output = await connector.with_env_variable("DEPLOYMENT_MODE", deployment_mode).with_exec(["spec"]).stdout()
+        return self._parse_spec_output(spec_output)
+
+    def _get_spec_as_file(self, spec: str, name="spec_to_cache.json") -> File:
+        return self.context.get_connector_dir().with_new_file(name, spec).file(name)
+
+    @property
+    def spec_key_prefix(self):
```
bnchrch, line 126: Can we move this to library/location where more of our app has access to the naming logic?

The reason I ask is that we have a few services that have to access the spec cache. Would be great to start standardizing that.

Let me know if thats too big of a task and needs to be a separate issue

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -0,0 +1,232 @@
...
+            metadata_upload_step = metadata.MetadataUpload(
+                context, context.metadata_path, metadata_bucket_name, await metadata_service_account.plaintext()
+            )
+            step_results = []
+
+            # TODO this sequence of "return on Failure" is not beautiful.
+            # We'll improve this soon with a different and global approach to chain steps.
+            metadata_validation_result = await metadata_validation_step.run()
+            step_results.append(metadata_validation_result)
+            if metadata_validation_result.status is StepStatus.FAILURE:
```
bnchrch, line 199: Im thinking that all these return on failures could benefit from the skip on failure logic in the [run_steps](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py#L22) function in the metadata pipeline.

What do you think of extending that function to work here. Would be a good step in the direction of the global approach you mention

bnchrch (commented):
Overall looks good @alafanechere !

Let me know your thoughts on `run_steps` :)

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
alafanechere: Interesting! It does not hurt of course to perform this check two times but I think it raises a separation of concerns question.
Should the metadata service be aware of the life cyle of the artifacts it refers to? 
IMO it's the continuous deployment logic that should provide guarantees.  
In the same logic as we're not running connectors test on publish because the fact it's been merged on master with the required tests to pass gives us the guarantee we're publishing an healthy image.
I'd be more inclined to implement a frequent (nightly?) health check on metadata files to be alerted about potential inconsistencies.

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
alafanechere: I tend to follow the following pattern in term of declaration order
1. Class constant variables
2. Property functions
3. Init function
4. class methods 

I'll move the class methods below init but will keep properties above if you don't mind.

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
alafanechere: I'd indeed prefer to do it in a separate issue. Probably when we'll move the pipeline code to `airbyte-ci`. It'll be easier to share code between metadata and connectors CI/CD. I logged an issue: #25419

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
alafanechere: Yes you're right. I opened a side PR to adapt it for connector use cases and use it https://github.com/airbytehq/airbyte/pull/25422

evantahler (approved):
Nice.

There's a lot to like in this PR:
* Testing this out safely by publishing pre-releases
* Using only one file change to trigger the publish (`metadata.yaml`)

I think you may also have solved the problem of how to keep all the connector docker files in sync (e.g to update out-of-date packages) by removing dockerfiles and moving this into Dagger.  I have a bunch of followup questions on this topic, like:
* What about base/normalization "sub" images - do we still need dockerfiles for these?
* What's the plan to remove the Dockerfiles for the connectors?  We wan to be very clear that there is "one right way" to do things. 

I also don't think we should fail the action if the version already exists in the case that the connector version in metadata.yaml hasn't changed.  But that's not a blocking change, as it's logically OK, but will make our metrics messy.

.github/workflows/publish_connectors.yml
```diff
@@ -0,0 +1,48 @@
+name: (WIP) Publish connectors
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - "airbyte-integrations/connectors/**/metadata.yaml"
```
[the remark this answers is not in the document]
alafanechere: The pipeline will still run but no image will get pushed if the version referenced in the metadata.yaml file is already available in DockerHub. We can implement a finer grained `diff` to run only if the version changed in the file but I find it overkill at the moment.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -619,8 +618,30 @@ async def with_airbyte_java_connector(context: ConnectorContext, connector_java_
...
     )
+
+
+def with_airbyte_python_connector(context: ConnectorContext, build_platform: Platform):
+    pip_cache: CacheVolume = context.dagger_client.cache_volume("pip_cache")
+
+    return (
+        context.dagger_client.container(platform=build_platform)
+        .from_("python:3.9-slim")
+        .with_exec(["apt-get", "update"])
```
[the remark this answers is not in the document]
alafanechere: @evantahler I did not thoroughly checked all the images and assumed all the Dockferfile the same (except for java connectors that optionally enable sentry).
Can we go away with this assumption and reconsider it if the acceptance tests are failing?
For this `source-faker` example I think we shall expose a timezone setting in the spec instead of inferring it from the build environment.

PR #25323, opened 2023-04-19
remove files that are needed to run airbyte-platform

c-p-b:
## What
These were already gitignored, but the actual deletion of the filesmust have slipped through the cracks as part of the monorepo migration. They are currently causing confusion for end users as the versions are mismatching with what is currently in airbyte-platform.

All of these files are downloaded and updated on-the-fly when running "run-ab-platform.sh". So they should be removed from here and gitignored from this repository

## How
Delete the files


.env
timroes, line 1: This will actually colide with the PR Joey just did: https://github.com/airbytehq/airbyte/pull/25311 since we're reading the `VERSION` from this file at the moment to figure out which OSS release to load. This also seems to be correctly bumped still in the release process. Should we write that in some other specific file and adjust the run platform script again, or keep just a `VERSION` around in this file, or some fully other solution?

cpdeethree: I think we should create something like an extensionless VERSION file that has this format, and only has one line in it that has version line identically as it looks in the .env (to save work changing things downstream). Maybe call it PLATFORM_VERSION or something. I would rather leave .env gitignored so that it's super clear that it's identical. Any particular feelings about the name or file format for this hypothetical one line file?

cpdeethree: Another, kind of cleaner possibly solution is to just store that version in run-airbyte-platform.sh and have bumpversion.cfg hit that instead

timroes: I feel `PLATFORM_VERSION` would be a nice name and being clear enough, alternatively we could call it `CURRENT_PLATFORM_VERSION`, but at some point this name is just getting too long :D

cpdeethree: The canonical version now lives in run-ab-platform.sh after this change (in order to make downloading and running airbyte just wget a single file instead of having to clone entire airbyte repo)

cpdeethree:
internal pr to bump run-ab-platform.sh as part of the OSS release process: https://github.com/airbytehq/airbyte-platform-internal/pull/6056

.bumpversion.cfg
```diff
@@ -16,6 +16,8 @@ serialize =
 
 [bumpversion:file:octavia-cli/README.md]
 
+[bumpversion:file:run-ab-platform.sh]
```
cpdeethree, line 19: I have no idea why, but putting this at the bottom caused bumpversion to fail. It only worked when I put it here before the setup.py line

run-ab-platform.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+VERSION=0.44.1
```
josephkmh, line 3: I think we want this injected down on line 44 too:

```
base_github_url="https://raw.githubusercontent.com/airbytehq/airbyte-platform/v$VERSION/"
```

Otherwise this script will download `docker-compose.yaml`, `.env` and co from `main` rather than from the latest release tag. `main` can contain breaking changes in `docker-compose.yaml` (as a recent PR of mine did) and then people were running the new `docker-compose.yaml` with "old" images from the most recent release, which caused problems.

cpdeethree: Done and done!

joeykmh (approved):

PR #25333, opened 2023-04-19
Dagger: Return true on valid exit in Dagger Commands

### LINKED ISSUE #25326: Bug: Metadata Upload is failing on skip
https://github.com/airbytehq/airbyte/actions/runs/4745534412/jobs/8427989274#step:4:911

bnchrch:
## What
Airbyte-ci commands were failing on skip because we were not returning a boolean value denoting pass or fail

## How
This updates our commands to return True when skipped. It also improved the start/failed error messaging

closes #25326


evantahler (approved):
Maybe the method should be -> Boolean typed to catch this kind of thing?

PR #25340, opened 2023-04-19
New dagger cli version

c-p-b:
## What
This new version contains some performance optimizations for the new caching strategy implemented on EKS. It also, very helpfully, now prints the dagger engine that a client connects to

## How
update dagger cli version


cpdeethree:
@alafanechere FYI

PR #25348, opened 2023-04-20
Perf test of preparethreshold=0

akashkulk:


akashkulk:
/connector-performance connector=connectors/source-postgres dataset=10m

PR #25351, opened 2023-04-20
Connector Builder UI Error Handler Documentation

### LINKED ISSUE #25154: Builder documentation: Error handling page

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/25154

## What

Adds documentation for the Error handler component in the Connector Builder UI


docs/connector-development/connector-builder-ui/error-handling.md
```diff
@@ -0,0 +1,84 @@
...
+
+When the API documentation recommends that requests be retried after waiting a constant amount of time, the "Constant" backoff strategy should be set on the error handler.
+
+#### Example
+
+The [Intercom API](https://developers.intercom.com/intercom-api-reference/reference/http-responses) is an API that recommends a constant backoff strategy when retrying requests.
+
+### Exponential
+
+The exponential backoff strategy is similar to constant where the connection waits to retry a request based on a numeric value defined on the connector. When the API documentation recommends that requests be retried after waiting an exponentially increasing amount of time, the "Exponential" backoff strategy should be set on the error handler.
```
flash1293, line 29: Would it make sense to give the formula here and the backoff times it will produce? It seems a little abstract in the current form.

The example below doesn't really illustrate it because it's not clear what will come after 4 (6? 8?)

docs/connector-development/connector-builder-ui/error-handling.md
```diff
@@ -0,0 +1,84 @@
...
+
+Although a lot of API documentation does not call out using an exponential backoff, some APIs like the [Posthog API](https://posthog.com/docs/api) mention rate limits that are advantageous to use an exponential backoff. In this case, the rate limit of 240 requests/min should work for most syncs. However, if there is a spike in traffic, then the exponential backoff allows the connection to avoid sending more requests than the endpoint can support.
+
+### Wait time from header
+
+The "Wait time from header" backoff strategy allows the connection to wait before retrying a request based on the value specified in the API response.
+
+#### Example
+
+The [Chargebee API](https://apidocs.chargebee.com/docs/api/error-handling) documentation recommends using the `Retry-After` in the response headers to determine when to retry the request.
```
flash1293, line 45: Could you add an example here how that looks in practice? Like "this request" returns "this header", which means we need to refetch at "time"

docs/connector-development/connector-builder-ui/error-handling.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Wait until time from header
+
+The "Wait until time from header" backoff strategy allows the connection to wait until a specific time before retrying a request according to the API response.
+
+#### Example
+
+The [Recurly API](https://recurly.com/developers/api/v2021-02-25/index.html#section/Getting-Started/Limits) is an API that defines a header `X-RateLimit-Reset` which specifies when the request rate limit will be reset. 
+
+During a sync, if the connection receives a 429 error response, it will read the `X-RateLimit-Reset` header and pause making further requests to the Recurly API until after the time specified in the header.
```
flash1293, line 57: Same here - I think this is especially important to emphasize the difference between wait time and wait until time

docs/connector-development/connector-builder-ui/error-handling.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Response filter
+
+A response filter should be used when a connection needs to interpret an API response to decide how the sync should proceed. Common use cases for this feature include ignoring error codes to continue fetching data, retrying requests for specific error codes, and stopping a sync based on the response received from the API.
+
+### Response conditions
+
+The following conditions can be specified on the "Response filter" and are used to determine if attributes of the response match the filter:
+- "If error message matches": The filter will match the provided text against the text body of the API response.
```
flash1293, line 66: Could we expand these three cases and have examples for them? Reading this as a novice user the following things are not clear to me:
* How is the error message determined?
* What does "matches" mean? Full match or include?
* How do I fullfill a predicate? How would that look for some example? I think especially for this one we need multiple examples because it's pretty complex
* If I specify multiple of these, is it and AND or an OR match?

docs/connector-development/connector-builder-ui/error-handling.md
```diff
@@ -0,0 +1,84 @@
...
+- "and predicate is fulfilled": This allows for more granular control over how the filter matches by evaluating an interpolation expression against the API response.
+- "and HTTP codes match": The filter will check to see if the HTTP status code of the response is in the provided set of HTTP status codes.
+
+### Then execute action
+
+If a response from the API matches the predicates of the response filter the connection will continue the sync according to the "Then execute action" definition. This is a list of the actions that a connection can take:
+- SUCCESS: The response was successful and the connection will continue fetching the next set of records from the API.
+- RETRY: The response was unsuccessful, but the error is transient and may be successful on subsequent attempts. The request will be retried according to the backoff policy defined on the error handler.
+- IGNORE: The response was unsuccessful, but the error should be ignored and the connection can continue fetching the next set of records from the API.
```
flash1293, line 75: What's the difference between success and ignore?

docs/connector-development/connector-builder-ui/error-handling.md
```diff
@@ -0,0 +1,84 @@
+# Error Handling
+
+Error handlers allow for the connection to decide how to continue fetching data according to the contents of the response from the partner API. Depending on attributes of the response such as status code, text body, or headers, the connection can continue making requests, retry unsuccessful attempts, or fail the sync.
```
girarda, line 3: When I think of connection, I think of a [source/destination pair + their config](https://docs.airbyte.com/understanding-airbyte/connections/). How about source or connector?

docs/connector-development/connector-builder-ui/error-handling.md
```diff
@@ -0,0 +1,84 @@
+# Error Handling
```
girarda, line 1: probably worth calling out that retries will be disabled on the connector builder. This is making me think maybe there should be a toggle to enable them if the developer wants to test the config.

docs/connector-development/connector-builder-ui/error-handling.md
girarda: good question :)

We should specify we'll extract records from a successful response, but not from an error we ignore

docs/connector-development/connector-builder-ui/error-handling.md
```diff
@@ -0,0 +1,84 @@
...
+
+Error handlers allow for the connection to decide how to continue fetching data according to the contents of the response from the partner API. Depending on attributes of the response such as status code, text body, or headers, the connection can continue making requests, retry unsuccessful attempts, or fail the sync.
+
+An error handler is made of two parts, "Backoff strategy" and "Response filter". When the conditions of the response filter are met, the connection will proceed with the sync according to behavior specified. See the [Response filter](#response-filter) section for a detailed breakdown of possible response filter actions. In the event of a failed request that needs to be retried, the backoff strategy determines how long the connection should wait before attempting the request again. 
+
+When an error handler is not implemented for a stream, the connection will default to retrying requests that received a 429 and 5XX status code in the response 5 times using a 5-second exponential backoff. This default retry behavior is recommended if an API's documentation does not specify error handling or retry behavior.
```
clnoll, line 7: If other errors are encountered, the sync will fail?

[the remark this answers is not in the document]
brianjlai: fair point, and configured seems to be how we phrase this in other parts of the docs. Will update this and any other places I use implemented

docs/connector-development/connector-builder-ui/error-handling.md
brianjlai: yah connector is also more consistent w/ other documentation will fix this throughout

docs/connector-development/connector-builder-ui/error-handling.md
brianjlai: added

flash1293 (approved):
That's great, thanks!

PR #25366, opened 2023-04-20
Destination Databricks: use hive_metastore if catalog not set

edgao:
This was a regression from https://github.com/airbytehq/airbyte/pull/24657.

I'm kind of hoping that one of the DATs will fail.... but if that's not the case, then I'll figure out the right place to add a test for this :/


ryankfu (approved):
LGTM pending passing integration tests

ifun248:
When I sync data from MySQL to Databricks, there is a data type mapping error which bigint data in MySQL maps into int data in Databricks. 

The detail error see https://github.com/airbytehq/airbyte/issues/25944. Hope this bug can be fixed soon!

PR #25370, opened 2023-04-20
connectors-ci: use shared pip cache instead of locked

alafanechere:
## What
Following Dagger team's advice the `pip_cache` sharing mode should be "SHARED" (the default sharing mode) f


octavia-approvington:

octavia-approvington (approved):

PR #25117, opened 2023-04-12
Add descriptions and examples to component schema

girarda:
## What
* Add descriptions, examples, and interpolation_context to component schema

## How
Add the following to all components:
* title: str
* description: str
* examples: array

Add interpolation_context array to fields that can interpolate values. The array list the values that can be accessed.

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml`
2. `airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/request_option.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/declarative/models/declarative_component_schema.py`

## 🚨 User Impact 🚨
*Are there any breaking changes? What is the end result perceived by the user?*

*For connector PRs, use this section to explain which type of semantic versioning bump occurs as a result of the changes. Refer to our [Semantic Versioning for Connectors](https://docs.airbyte.com/contributing-to-airbyte/#semantic-versioning-for-connectors) guidelines for more information. **Breaking changes to connectors must be documented by an Airbyte engineer (PR author, or reviewer for community PRs) by using the [Breaking Change Release Playbook](https://docs.google.com/document/d/1VYQggHbL_PN0dDDu7rCyzBLGRtX-R3cpwXaY8QxEgzw/edit).***

*If there are breaking changes, please merge this PR with the 🚨🚨 emoji so changelog authors can further highlight this if needed.*

## Pre-merge Checklist
*Expand the relevant checklist and delete the others.*

<details><summary><strong>New Connector</strong></summary>

### Community member or Airbyter

- [ ] **Community member?** Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
- [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [ ] Connector version is set to `0.0.1`
    - [ ] `Dockerfile` has version `0.0.1`
- [ ] Documentation updated
    - [ ] Connector's `README.md`
[...]

Issue #24723, opened 2023-03-30
low-code: Access partition router's values using stream_partition variable 


girarda:
We're trying to remove the concept of stream_slice from the connector builder.

The last reference to stream_slice is the variable used in jinja macros to access the partition router and incremental_sync values.

flash1293:
This is already renamed in the UI, the only part that's left is renaming the `stream_slice` object in the YAML to something else (maybe `stream_partition`?)

Problem: `stream_slice` is also used for the incremental sync slices (maybe that should have a separate object? or is it fine to keep that on `stream_partition` as well?)

flash1293:
Grooming notes:
* Splitting it up is a lot of effort
* Having `stream_partition` as an alias to `stream_slice` can easily be added

girarda:
@flash1293 just as a heads up - I renamed the issue and changed the description to better match what we groomed on Tuesday - rename the `stream_slice` jinja variable.

The original issue was about renaming a field in the schema. I created a new issue since that's different: https://github.com/airbytehq/airbyte/issues/25375

PR #25373, opened 2023-04-20
low-code: Alias stream_interval and stream_partition to stream_slice in interpolation context

girarda:
## What
* This PR handles a subset of https://github.com/airbytehq/airbyte/issues/24723.

Allow connector developer to refer to the stream slice using two new keywords:
- stream_interval
- stream_partition

The end-goal is for `stream_interval` to be used to refer to incremental_sync's values and `stream_partition` to refer to partition router's values. This will enable us to stop referring to `stream_slice` from the connector builder.

This is a fairly heavy lift today because
1. These concepts do not exist in all layers of the CDK. It's all just stream slices from the SimpleRetriever's point of view.
2. Removing the stream_slice keyword would be an expensive breaking change.
Since we don't have the appetite to completely replace the `stream_slice` keyword, we instead copy the value to `stream_partition` and `stream_interval`. 

Not included in this PR:
- Updating the reference docs. Will be done as part of https://github.com/airbytehq/airbyte/pull/25117 (https://github.com/airbytehq/airbyte/pull/25117/commits/0e5f5b51aabd5ba2c1f8b73a2112dcf3800f73b3)

## How
* If stream_slice is the Jinja context, copy it's value to the `stream_interval` and `stream_partition` keys
* If `stream_interval` or `stream_partition` are already in the context, raise an exception because that is unexpected

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/interpolation/jinja.py`
4. `airbyte-cdk/python/unit_tests/sources/declarative/interpolation/test_jinja.py`

### DIFF
--- a/airbyte-cdk/python/unit_tests/sources/declarative/interpolation/test_jinja.py
+++ b/airbyte-cdk/python/unit_tests/sources/declarative/interpolation/test_jinja.py
@@ -46,6 +46,7 @@ def test_literals(test_name, s, value):
     val = interpolation.eval(s, None)
     assert val == value
 
+
 @pytest.mark.parametrize(
     "test_name, context, input_string, expected_value",
     [
@@ -73,6 +74,19 @@ def test_stream_slice_alias(test_name, context, input_string, expected_value):
     config = {}
     val = interpolation.eval(input_string, config, **context)
     assert val == expected_value
+
+
+@pytest.mark.parametrize(
+    "test_name, alias", [
+        ("test_error_is_raised_if_stream_interval_in_context", "stream_interval"),
+        ("test_error_is_raised_if_stream_partition_in_context", "stream_partition"),
+    ]
+)
+def test_error_is_raised_if_alias_is_already_in_context(test_name, alias):
+    config = {}
+    context = {alias: "a_value"}
+    with pytest.raises(ValueError):
+        interpolation.eval("a_key", config, **context)
 
 
 def test_positive_day_delta():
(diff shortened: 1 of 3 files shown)


airbyte-cdk/python/unit_tests/sources/declarative/interpolation/test_jinja.py
```diff
@@ -47,6 +47,48 @@ def test_literals(test_name, s, value):
     assert val == value
 
 
+@pytest.mark.parametrize(
+    "test_name, context, input_string, expected_value",
+    [
+        ("test_get_value_from_stream_slice",
```
brianjlai, line 53: Using the `pytest.param` class is more idiomatic for how we should be structuring our tests.
```
"context, input_string, expected_value"
[
  pytest.param(
     "input_1",
     { ... },
      id="test_get_value_from_stream_slice",
  )
]
```

airbyte-cdk/python/unit_tests/sources/declarative/interpolation/test_jinja.py
```diff
@@ -47,6 +47,48 @@ def test_literals(test_name, s, value):
...
+        ("test_get_value_from_stream_slice_no_stream_slice",
+         {},
+         "{{ stream_slice['stream_slice_key'] }}", None),
+        ("test_get_value_from_stream_partition",
+         {"stream_slice": {"stream_slice_key": "hello"}},
+         "{{ stream_partition['stream_slice_key'] }}", "hello"),
+        ("test_get_value_from_stream_partition_no_stream_slice",
+         {},
+         "{{ stream_partition['stream_slice_key'] }}", None),
+        ("test_get_value_from_stream_partition",
```
brianjlai, line 65: This should be `"test_get_value_from_stream_interval"`

brianjlai (approved):
nits on a few parts of the tests, but nothing to block on

PR #25376, opened 2023-04-21
Fix: Add repo checkout step to dagster deploy pipeline

bnchrch:
## What
Action to deploy dagster is broken because it cant find any actions
This is because we dont checkout the repo at the beginning of the workflow

## How
This adds that step


alafanechere (approved):
FYI for this kind of changes I heavily used the approve-and-merge slash command :)

Issue #25325, opened 2023-04-19
`destination-redshift` is failing nightly connector health checks


evantahler:
Check the #connector-health channel to see that redshift is still failing
I think that what's wrong is operational (passwords, auth, the db we are testing with)
This is a debugging story and a chance to explore our infra

* How do you get to the redsheft server? - Google Secret Manager - config + https://internal-docs.airbyte.io/
* How to you learn the test credentials? - Google Secret Manager
* What changes need to be made?

A way to start this investigation is to make a no-op PR and run `/test` and see what happens

cynthiaxyin:
Documenting current observations.

`/test` for Redshift via Github actions only fails 1 test:

```
RedshiftInsertDestinationAcceptanceTest > testIfSuperTmpTableWasCreatedAfterVarcharTmpTable() FAILED
    java.sql.SQLException: [Amazon](500310) Invalid operation: column "_airbyte_data" is of type character varying but expression is of type super 
    Hint: You will need to rewrite or cast the expression.;
        at com.amazon.redshift.client.messages.inbound.ErrorResponse.toErrorException(Unknown Source)
        at com.amazon.redshift.client.PGMessagingContext.handleErrorResponse(Unknown Source)
        at com.amazon.redshift.client.PGMessagingContext.handleMessage(Unknown Source)
        at com.amazon.jdbc.communications.InboundMessagesPipeline.getNextMessageOfClass(Unknown Source)
        at com.amazon.redshift.client.PGMessagingContext.doMoveToNextClass(Unknown Source)
        at com.amazon.redshift.client.PGMessagingContext.getParameterDescription(Unknown Source)
        at com.amazon.redshift.client.PGClient.prepareStatement(Unknown Source)
        at com.amazon.redshift.dataengine.PGQueryExecutor.<init>(Unknown Source)
        at com.amazon.redshift.dataengine.PGDataEngine.prepare(Unknown Source)
        at com.amazon.jdbc.common.SPreparedStatement.<init>(Unknown Source)
        at com.amazon.jdbc.jdbc41.S41PreparedStatement.<init>(Unknown Source)
        at com.amazon.jdbc.jdbc42.S42PreparedStatement.<init>(Unknown Source)
        at com.amazon.redshift.core.jdbc42.PGJDBC42PreparedStatement.<init>(Unknown Source)
        at com.amazon.redshift.core.jdbc42.PGJDBC42ObjectFactory.createPreparedStatement(Unknown Source)
[...]

PR #25379, opened 2023-04-21
no-op

cynthiaxyin:


cynthiaxyin:
This was just used to repro #25325.

PR #25390, opened 2023-04-21
Fix Java destination template

penberg:
## What

The code generated by the connector generator doesn't actually compile.

## How

Fix up the the function signatures to include checked exceptions and fix the import paths.

## 🚨 User Impact 🚨

No.

Developers will appreciate that generated code compiles, though.


marcosmarxm (approved):
Thanks @penberg

PR #25404, opened 2023-04-21
low-code: Delete now_local macro

girarda:
## What
- The `now_local()` macro is unpredictable because it uses the time local to the machine running the sync. 
- We do not recommend connector developers to use this method
- No existing connectors use the macro
- We should just delete it to prevent future headaches

## How
* Delete the macro
* Delete the unit test
* Remove the macro from the reference docs

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/interpolation/macros.py`
2. `airbyte-cdk/python/unit_tests/sources/declarative/interpolation/test_macros.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml`


clnoll (commented):
Nice! Found a couple more places where references to it can be removed.
- https://github.com/airbytehq/airbyte/blob/8d39efb8718599606cd0962535e38f2573dad8e6/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml#L592
- https://github.com/airbytehq/airbyte/blob/8d39efb8718599606cd0962535e38f2573dad8e6/airbyte-cdk/python/airbyte_cdk/sources/declarative/models/declarative_component_schema.py#L863

girarda:
good catch @clnoll. Removed those references too

clnoll (approved):

PR #25422, opened 2023-04-24
connectors-ci: refactor to use `run_steps`

alafanechere:
## What
@bnchrch implemented a nice function to run multiple pipeline steps and handle their failures:
`run_steps`.
In this PR I'm adapting it so that it can call the `run` method of `Step` with custom arguments.
I'm also changing the `StepResult` model to receive a `output_artifact` attribute.
This attribute will help enforce the fact that `_run` method should return `StepResult` and not tuple of `StepResult` and output artifact. Some `Step._run` indeed returned additional objects that are used downstream for other steps. This `output_artifact` attributes should receive these objects.


bnchrch (approved):
Not one, nit, issue, or comment. This PR is perfect. Nice Job @alafanechere !

PR #25428, opened 2023-04-24
Connector builder documentation: Auto-generate yaml reference and link up in fitting places

### LINKED ISSUE #24893: Auto-generate low code YAML reference page
The current YAML reference file is only linking to the schema file: https://docs.airbyte.com/connector-development/config-based/understanding-the-yaml-file/reference

To make it easier to consume for users, for all definitions from the schema a markdown section should be generated with title, description, examples, required and optional fields and so on

POC: https://github.com/airbytehq/airbyte/pull/21254

This should also include a section for jinja macros: https://github.com/airbytehq/airbyte/issues/24576

flash1293:
## What

Closes https://github.com/airbytehq/airbyte/issues/24893
Screenshot 2023-04-24 at 16 38 57

Built on https://github.com/airbytehq/airbyte/pull/25440

## To test

* Go to `docusaurus` directory
* Install deps: `yarn` (requires yarn: https://classic.yarnpkg.com/lang/en/docs/install/#mac-stable )
* Start dev server: `yarn start`

## How

We are using mdx for docusaurus which means you can use react components to render documentation pages. By adding a yaml loader to load the manifest schema and implementing a component that iterates over the individual parts and translates the information into HTML elements, the reference page can be filled dynamically based on the schema file.

**Note** Right now this code is reaching into the `airbyte-cdk` directory to get the manifest, which means the documentation will always show the most recent state of the schema (which is not necessarily the released version). We could represent this by copying the schema into the docs directory on release to decouple it from the master state - if you feel like we should do that I can add it on this PR.

Some details about this generation:
* The descriptions can also contain markdown - however, docusaurus is not evaluating the output from custom components as markdown again, so `react-markdown` is pulled in to render these "inner" markdown texts
* Heading are not regular headings - they need to have the right classes set to make the table of contents to the right work and to take the sticky header into account when scrolling to a heading - for this reason, the `Heading` component has to be imported
* The automatic table-of-contents generation is working on the ast of the parsed mdx, so it doesn't include dynamic headings generated by components. By exporting the `toc` variable from `reference.md`, it's possible to set it manually. For this reason it's also pulling in the schema to generate the right entries
* The declarative source definition itself is the root level object of the schema, so it needs to be added as a special case to get a complete list
* Variables that can be used in an interpolation context, refs and arrays of refs are linked up so it's possible to jump through the reference, following some trace. For simplification it's only going one way (like things are linked up in the schema itself - we could add the other direction as well (e.g. "variable used in following components:"), but I'm not sure whether it's relevant.
* I'm ignoring titles because in almost all cases they wouldn't add any additional information and would look weird (e.g. "BearerAuthenticator" vs "Bearer Authenticator")

I tried to link up the variables/macros section in all places on the connector builder UI documentation pages where this makes sense.


docs/connector-development/config-based/understanding-the-yaml-file/reference.md
```diff
@@ -1,3 +1,50 @@
...
+
+import schema from "../../../../airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml";
+
+
 # YAML Reference
 
-To see every component and each possible parameter, see the full YAML schema reference [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml).
+This page lists all components, interpolation variables and interpolation macros that can be used when defining a low code YAML file.
+
+For the technical JSON schema definition that low code manifests are validated against, see [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml).
```
girarda, line 10: Should we link to the $parameters section since it's a field in all the components?

Maybe something like

> All components have an optional `$parameters` field used to inherit propertyies. See the section in the advanced topics for more details (../advanced-topics.md#parameters).

girarda (approved):
this is great!

One thing this is making me think is that it could make sense to have two different description fields
- a short one for the tooltips
- a potentially longer one for the docs page
(not something to do today)

docs/connector-development/config-based/understanding-the-yaml-file/reference.md
flash1293: Good idea, added that:
Screenshot 2023-04-25 at 10 54 08

flash1293:
> One thing this is making me think is that it could make sense to have two different description fields

Yeah, I had the same idea here:
https://docs.google.com/document/d/1iWjJu-ZwK5wHcOxJZUO2kMUhh-5gbNKN3ju7PexpEqY/edit#bookmark=id.64a9zpnf6dkw

But I didn't want to make it too complex from the start. If you can see places where you would like to differentiate it's super cheap to add that.

Issue #23382, opened 2023-02-23
Connector builder: Ignore backoff configuration


flash1293:
When running test reads in the connector builder UI, quick feedback is very important as the user is actively iterating on the configuration and long wait times (>10s) are very disruptive for interactive processes like this (see https://www.nngroup.com/articles/powers-of-10-time-scales-in-ux/).

In some cases, the connector builder server will attempt to execute the configured backoff strategy which is a very helpful feature while running a connector in production but is problematic for the user experience during connector development.

For this reason all backoff strategies should be ignored in the connector builder, propagating errors back to the user immediately.

flash1293:
cc @sherifnada as discussed yesterday, not sure whether you created something to track this already.

girarda:
This is somewhat blocked on this bug which prevents us from customizing backoffs when there are no responses

maxi297:
This is the bug @girarda is referring to I think https://github.com/airbytehq/airbyte/issues/23942

flash1293:
Grooming notes:
* Similar to slice/page limit functionality
* Consolidate into a single flag passed to the model constructor
* Look up flag in the requester to not apply backoff configuration

lmossman:
We should add a message/tooltip to the UI explaining that backoff configuration won't be used for Test reads, but should be configured for production syncs.

We don't want to ignore the error handling/filter configuration - just the backoff.

So when running a read for the connector builder, the backend should set the number of retries to 0.

lmossman:
We should add a message/tooltip to the UI explaining that backoff configuration won't be used for Test reads, but should be configured for production syncs.

We don't want to ignore the error handling/filter configuration - just the backoff.

So when running a read for the connector builder, the backend should set the number of retries to 0.

PR #25429, opened 2023-04-24
[ISSUE #23382] ignore backoff configuration on test reads

maxi297:
## What
Addresses https://github.com/airbytehq/airbyte/issues/23382

Previous behaviour:
```
2023-04-24 10:41:12 [    INFO] Starting syncing ManifestDeclarativeSource (abstract_source.py:95)
2023-04-24 10:41:12 [    INFO] Syncing stream: test  (abstract_source.py:184)
2023-04-24 10:41:12 [    INFO] Backing off _send(...) for 5.0s (requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))) (_common.py:105)
2023-04-24 10:41:12 [    INFO] Caught retryable error '('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))' after 1 tries. Waiting 5 seconds then retrying... (rate_limiting.py:31)
2023-04-24 10:41:17 [    INFO] Backing off _send(...) for 10.0s (requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))) (_common.py:105)
2023-04-24 10:41:17 [    INFO] Caught retryable error '('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))' after 2 tries. Waiting 10 seconds then retrying... (rate_limiting.py:31)
```

New behaviour:
```
2023-04-24 10:29:18 [    INFO] Starting syncing ManifestDeclarativeSource (abstract_source.py:95)
2023-04-24 10:29:18 [    INFO] Syncing stream: test  (abstract_source.py:184)
2023-04-24 10:29:18 [   ERROR] Giving up _send(...) after 1 tries (requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))) (_common.py:120)
2023-04-24 10:29:18 [   ERROR] Encountered an exception while reading stream test (abstract_source.py:126)
Traceback (most recent call last):
```

## How
Override SimpleRetriever. max_retries to take into account:
* disabling of the retries
* max_retries from the error_handler

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py` where max_retries is overridden
2. `airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py` where the flag is defined to by disable retries
3. `airbyte-cdk/python/airbyte_cdk/connector_builder/connector_builder_handler.py` where it is configured for the connector builder

## 🚨 User Impact 🚨
No breaking changes
* Connector builder will not retry anymore. This also includes the case defined by https://github.com/airbytehq/airbyte-internal-issues/issues/2696 (even though it does not solve the issue because we don't rely on the ErrorHandler when `disable_retries` is not configured) which probably reduce the priority of this ticket
* The max retries will be based on the ErrorHandler and not on HttpStream.max_retries

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
@@ -112,12 +112,17 @@ DEFAULT_BACKOFF_STRATEGY = ExponentialBackoffStrategy
 
 class ModelToComponentFactory:
     def __init__(
-        self, limit_pages_fetched_per_slice: int = None, limit_slices_fetched: int = None, emit_connector_builder_messages: bool = False
+        self,
+        limit_pages_fetched_per_slice: int = None,
+        limit_slices_fetched: int = None,
+        emit_connector_builder_messages: bool = False,
+        disable_retries=False,
     ):
         self._init_mappings()
         self._limit_pages_fetched_per_slice = limit_pages_fetched_per_slice
         self._limit_slices_fetched = limit_slices_fetched
         self._emit_connector_builder_messages = emit_connector_builder_messages
+        self._disable_retries = disable_retries
 
     def _init_mappings(self):
         self.PYDANTIC_MODEL_TO_CONSTRUCTOR: [Type[BaseModel], Callable] = {
@@ -779,6 +784,7 @@ class ModelToComponentFactory:
                 config=config,
                 maximum_number_of_slices=self._limit_slices_fetched,
                 parameters=model.parameters,
+                disable_retries=self._disable_retries,
             )
         return SimpleRetriever(
             name=name,
@@ -789,6 +795,7 @@ class ModelToComponentFactory:
             stream_slicer=stream_slicer or SinglePartitionRouter(parameters={}),
             config=config,
             parameters=model.parameters,
+            disable_retries=self._disable_retries,
         )
 
     @staticmethod
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
@@ -50,6 +50,8 @@ class SimpleRetriever(Retriever, HttpStream):
         parameters (Mapping[str, Any]): Additional runtime parameters to be used for string interpolation
     """
 
+    _DEFAULT_MAX_RETRY = 5
+
     requester: Requester
     record_selector: HttpSelector
     config: Config
@@ -61,6 +63,7 @@ class SimpleRetriever(Retriever, HttpStream):
     paginator: Optional[Paginator] = None
     stream_slicer: Optional[StreamSlicer] = SinglePartitionRouter(parameters={})
     emit_connector_builder_messages: bool = False
+    disable_retries: bool = False
 
     def __post_init__(self, parameters: Mapping[str, Any]):
         self.paginator = self.paginator or NoPagination(parameters=parameters)
@@ -94,6 +97,14 @@ class SimpleRetriever(Retriever, HttpStream):
     def raise_on_http_errors(self) -> bool:
         # never raise on http_errors because this overrides the error handler logic...
         return False
+
+    @property
+    def max_retries(self) -> Union[int, None]:
+        if self.disable_retries:
+            return 0
+        if hasattr(self.requester.error_handler, "max_retries"):
+            return self.requester.error_handler.max_retries
+        return self._DEFAULT_MAX_RETRY
 
     def should_retry(self, response: requests.Response) -> bool:
         """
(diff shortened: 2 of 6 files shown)


maxi297:
Note that the new behavior could be very confusing for CompositeErrorHandler since it uses the max_retries of the first ErrorHandler (see https://github.com/airbytehq/airbyte/blob/08f83b89bd0eb3c17b0cbe031a2ba4e860abbbc4/airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/error_handlers/composite_error_handler.py#L49)

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -95,6 +98,14 @@ def raise_on_http_errors(self) -> bool:
         # never raise on http_errors because this overrides the error handler logic...
         return False
 
+    @property
+    def max_retries(self) -> Union[int, None]:
+        if self.disable_retries:
+            return 0
+        if hasattr(self.requester.error_handler, "max_retries"):
```
maxi297, line 105: CustomErrorHandler might not expose the value so we need to validate if the attribute exists

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -789,6 +795,7 @@ def create_simple_retriever(
             stream_slicer=stream_slicer or SinglePartitionRouter(parameters={}),
             config=config,
             parameters=model.parameters,
+            disable_retries=self._disable_retries,
```
brianjlai, line 798: I'm curious what the use case is for disabling retries when we are not using the `SimpleRetrieverTestReadDecorator`? Mainly because we used to be using that decorator for the test builder specific implementation details like max_slices, etc.

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
maxi297: How would you suggest to do that? I don't see any way of enforcing the usage of an interface when implementing a CustomErrorHandler

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
maxi297: This is a good question and I don't have a clear answer for this. I added this in the `SimpleRetriever` instead of the `SimpleRetrieverTestReadDecorator` because I could not find a clean way to do this using the current interface i.e. if we init `ModelToComponentFactory` with `disable_retries` but without `_limit_slices_fetched` or `_emit_connector_builder_messages`, how should it translate in terms of implementation?

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
girarda: Fine to keep out of scope of this PR, but we should aim to be in a place where we can assume custom components implement the interface they're supposed to - custom error handlers should implement the error handler interface, and custom paginators should implement the paginator interface.

One approach would be for the [`create_custom_component`](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py#L269) to verify that the class inherits the expected interface.

girarda: [Created an issue](https://github.com/airbytehq/airbyte/issues/25463). No need to worry about it here

girarda (approved):

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: we talked about this a bit on Slack, given that we have to expose `disable_retries` config on the `ModelToComponentFactory`, it is likely more confusing to only honor this on the `SimpleRetrieverTestReadDecorator` So in that regard we can leave this as is on both the decoarator and `SimpleRetriever`.

However, this does feel like we're starting to be more apparent that all these settings start to make it less clear how they affect the components that are constructed as the list of settings, `limit_slices_fetched`, `page_limts`, `disable_retries` grows more. We're basically using the config of the factory as a surrogate for telling the factory this is a `TestRead` connector. But that goes a little outside the scope of this change and we don't need to block on this

brianjlai (approved):
just added a follow up on some of our discussions on Slack. changes 👍

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
maxi297: To be honest, I wouldn't mind having only one flag `test_read` and to configure everything based on that. It's not like we have different values for all of these and we only call it in one context. If we do that, we can simplify today and extract as needed later on

PR #25442, opened 2023-04-24
Remove `ENABLE_SENTRY` from connector Dockerfiles

evantahler:


edgao (approved):
:shrug:

jbfbell (approved):
Did a quick scan of Sentry's docs and it doesn't look like this is something they use either

alafanechere (approved):
🧹 Thank you!!!

PR #25460, opened 2023-04-25
Convert source mysql decimal number with 0 decimal digits to an int

### LINKED ISSUE #25027: source-mysql decimal(x, 0) translates to an Airbyte `Number` instead of an `Integer`
This is part of #24470 .
## Environment
- **Airbyte version**:  example is 0.22.0-alpha
- **OS Version / Instance**: example macOS, Windows 7/10, Ubuntu 18.04, GCP n2. , AWS EC2
- **Deployment**: example are Docker or Kubernetes deploy env
- **Source Connector and version**: mysql 2.0.13 

## Current Behavior
Create on mysql a column of type `DECIMAL(M, 0)`
Where M can be any number (1-65) and 0 signifies decimal digits = none.
While this is effectively an integer, on airbyte this is translated to `JsonSchemaType.NUMBER` rather than an `INTEGER`.
A similar issue exists also in [source-postgress](https://github.com/airbytehq/airbyte/issues/20466) and [snowflake](https://github.com/airbytehq/airbyte/issues/19776)

## Expected Behavior
Make `DECIMAL(M, 0)` turn into an `INTEGER` to make it show up correctly on destination side.

rodireich:
## What
Improves the accuracy of decimal type conversion when it is defined with 0 decimal points (essentially an integer).

## How
Read the standard decimal digits value when parsing column.
In mysql only, A decimal number with 0 decimal point is converted to an airbyte `JsonSchemaType.INTEGER` rather than a `NUMBER`.

## Recommended reading order
1. `MySqlSourceOperations.java`
2. `AbstractJdbcSource`

## 🚨 User Impact 🚨
I validated that in case a table already synced with an older version (which would have created a NUMBER (double) on the destination side, user can still keep syncing as 1234 would be converted by normalization to 1234.0 and so on.

Similarly, in case of a rollback after a destination table was already created with an Int, normalization would convert 1234.0 to 1234.

Resolves #25027


airbyte-integrations/connectors/source-mysql/src/main/java/io/airbyte/integrations/source/mysql/MySqlSourceOperations.java
```diff
@@ -191,8 +192,17 @@ public MysqlType getDatabaseFieldType(final JsonNode field) {
...
+        case DECIMAL -> {
+          if (field.get(INTERNAL_DECIMAL_DIGITS) != null && field.get(INTERNAL_DECIMAL_DIGITS).asInt() == 0) {
+            return BIGINT;
+          }
+        }
+        case DECIMAL_UNSIGNED -> {
+          if (field.get(INTERNAL_DECIMAL_DIGITS) != null && field.get(INTERNAL_DECIMAL_DIGITS).asInt() == 0) {
+            return BIGINT_UNSIGNED;
+          }
+        }
```
rodireich, line 204: @nguyenaiden 
This is the only thing I think you'll need to put in postgres or any other connector in order to resolves the similar issues. (plus testing)

nguyenaiden (approved):
+1. Thanks for this!

prateekmukhedkar (approved):

akashkulk:
Approving, but quick questions : 
1. Which destinations is this tested against for normalization? 
2. Are we handling arrays here (I forget if we have array support for MySQL)

akashkulk (approved):

rodireich:
@akashkulk 
1. Validated with destination postgres and snowflake if I recall correctly.
2. No array types in mysql!

rodireich:
/publish connector=connectors/source-mysql-strict-encrypt auto-bump-version=false

PR #25471, opened 2023-04-25
connectors-ci: fix modified files detection on master

alafanechere:
## What
Under the connector command group the logic to detect modified connectors was relying on the list of modified files on the branch.
When running on master this logic does not make sense as diffing master with master does not lead to any modified files.

## How
- Expose a `modified_files` list in the click context object
- If we're on the master branch the list of modified files is built from the `git diff` of the latest commit. If not on master the current branch is diffed with master to retrieve the list of modified files.


octavia-approvington (approved):

octavia-approvington:
This code is at peace with itself <br/> imagine karate kid approving

PR #25473, opened 2023-04-25
Connector builder documentation: Extend record selection examples

flash1293:
## What

I noticed that the connector builder record selection page was missing one example that was explicitly called out here: https://docs.airbyte.com/connector-development/config-based/understanding-the-yaml-file/record-selector/#selecting-fields-nested-in-arrays

Added that as well.

### DIFF
--- a/docs/connector-development/connector-builder-ui/record-processing.mdx
+++ b/docs/connector-development/connector-builder-ui/record-processing.mdx
@@ -101,6 +101,42 @@ Sometimes, there is only one record returned per request from the API. In this c
 
 In this case, a record selector of `rates` will yield a single record which contains all the exchange rates in a single object.
 
+### Fields nested in arrays
+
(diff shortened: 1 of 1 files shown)


docs/connector-development/connector-builder-ui/record-processing.mdx
```diff
@@ -101,6 +101,42 @@ Sometimes, there is only one record returned per request from the API. In this c
...
+      "record": {
+        "id": "2"
+      }
+    }
+  ]
+}
+
+```
+
+In this case a record selector with a placeholder `*` selects all children at the current position in the path, in this case **`data`, `*`, `record`** will return the following records:
```
girarda, line 127: should we can `*` a glob instead of a placeholder?

girarda (approved):

PR #25474, opened 2023-04-25
Docusaurus: Change port for local development

flash1293:
When running `yarn start` to develop docs locally, it's using port 3000, just like the webapp dev server. This PR changes the port to 3005 so it's possible to run webapp dev server and docs dev server side by side.


timroes (approved):
Code LGTM

PR #25483, opened 2023-04-25
connectors-ci: use dockerfile for python connectors

alafanechere:
## What
I originally tried to replicate with Dagger the Dockerfile in the generator for python connectors.
I realized we can't adopt this Dockerfile for a couple of connectors because some of connectors python dependencies are not compatible with the alpine base image we are using.

This is pretty well explained on a recent destination:
```
# FROM python:3.9.11-alpine3.15 as base
# switched from alpine as there were tons of errors (in case you want to switch back to alpine)
# - https://stackoverflow.com/a/57485724/5246670
# - numpy error: https://stackoverflow.com/a/22411624/5246670
# - libstdc++ https://github.com/amancevice/docker-pandas/issues/12#issuecomment-717215043
# - musl-dev linux-headers g++ because of: https://stackoverflow.com/a/40407099/5246670
```

**I think it's safer to rely on connectors dockerfile to build them at the moment.**

**The connectors that can't use the alpine base images are the one relying on libraries like `pandas` and `pyarrows`**:
* source-salesforce
* source-file
* source-sendgrid
... 

## How
Change `with_airbyte_python_connector` function to use connector specific dockerfile.

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
@@ -624,7 +624,18 @@ async def with_airbyte_java_connector(context: ConnectorContext, connector_java_
     )
 
 
-def with_airbyte_python_connector(context: ConnectorContext, build_platform: Platform):
+def with_airbyte_python_connector(context: ConnectorContext, build_platform):
+    pip_cache: CacheVolume = context.dagger_client.cache_volume("pip_cache")
+    return (
+        context.dagger_client.container(platform=build_platform)
+        .with_mounted_cache("/root/.cache/pip", pip_cache)
+        .build(context.get_connector_dir())
+        .with_label("io.airbyte.version", context.metadata["dockerImageTag"])
+        .with_label("io.airbyte.name", context.metadata["dockerRepository"])
+    )
+
+
+def with_airbyte_python_connector_full_dagger(context: ConnectorContext, build_platform: Platform):
     pip_cache: CacheVolume = context.dagger_client.cache_volume("pip_cache")
     base = context.dagger_client.container(platform=build_platform).from_("python:3.9.11-alpine3.15")
     snake_case_name = context.connector.technical_name.replace("-", "_")


bnchrch (commented):
@alafanechere Why not ditch the alpine base image instead?

I know in my own dagger pipelines I wasnt able to use it as it didnt seem to have good .whl file support and if I recall the python community recommends not to use alpine at all

evantahler:
I'd really prefer to force all the python connectors to use the same base image...

alafanechere:
@evantahler and @bnchrch 
I think we've three options:
1. Build everything with the same base image, which would be python:3.9-slim . We sacrifice image size but gain consistency.
2. Build everything by default with python:3.9.11-alpine3.15  and fallback to python:3.9-slim if  we spot that the connectors is using incompatible dependencies (according to the setup.py ). No sacrifice but complexify things...
3. (The current PR) Use the connector dockerfile: the best guarantee for consistency between the original pipelines and the new one.

I suggest that I try out option 1 but if I spot other compatibility problems I'll switch to 3.

I could be in favor of 3 too because I find it a bit early in the adoption of this tool to ditch the Dockerfile now. I clearly understand the benefits of getting rid of these but I find that we are already changing a lot of process in this project, so removing one of the multiple moving parts would simplify things for me in the short term.

bnchrch (approved):
@alafanechere Thanks for laying out the options

Thinking on this more, I agree with @evantahler we should ensure we use the same base image everywhere when it comes to python. @alafanechere you pointed out that would be `python:3.9-slim`. Im a thumbs up for that.

But! your point here:
> I could be in favor of 3 too because I find it a bit early in the adoption of this tool to ditch the Dockerfile now. I clearly understand the benefits of getting rid of these but I find that we are already changing a lot of process in this project, so removing one of the multiple moving parts would simplify things for me in the short term.

Made me think these are two separate issues.

The first is standardizing what python docker base image is used. The solution to that seems straight forward.

The second is "What defines how a connector image is build" and I think here Augustin is right that the dockerfile in the connectors folder should be the source of truth and our pipeline should be unaware and unopinionated about the implementation details.

**TLDR: I approve this PR and option 3 so long as we in a separate issue we address that airbyte as a company should be using python-slim not python-alpine**

alafanechere:
Thank you @bnchrch for sharing this.
I logged this issue https://github.com/airbytehq/airbyte/issues/25523

PR #21971, opened 2023-01-27
Source Netsuite: retry with next dateformat in list

ChloeBellm:
## What
[Issue: Netsuite source connector dateformat error](https://github.com/airbytehq/airbyte/issues/21965)

## How
Loops through the possible date formats.

### Community member or Airbyter

- [X] Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [X] Secrets in the connector's spec are annotated with `airbyte_secret`
- [X] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [X] Documentation updated
    - [X] Connector's `README.md`
    - [X] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - [X] Changelog updated in `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
- [X] PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests)

### Airbyter

If this is a community PR, the Airbyte engineer reviewing this PR is responsible for the below items.

- [ ] Create a non-forked branch based on this PR and test the below items on it
- [ ] Build is successful
- [ ] If new credentials are required for use in CI, add them to GSM. [Instructions](https://docs.airbyte.io/connector-development#using-credentials-in-ci).
- [ ] [`/test connector=connectors/<name>` command](https://docs.airbyte.io/connector-development#updating-an-existing-connector) is passing
- [ ] New Connector version released on Dockerhub and connector version bumped by running the `/publish` command described [here](https://docs.airbyte.io/connector-development#updating-an-existing-connector)

</details>

Issue #21965, opened 2023-01-27
Netsuite source connector dateformat error


ChloeBellm:
## Environment
- **Airbyte version**:  0.40.27
- **OS Version / Instance**: AWS EC2
- **Deployment**: docker
- **Source Connector and version**: Netsuite 0.1.3

## Current Behavior
Getting an error with a %Y-%m-%d dateformat.
```
DATE FORMAT exception. Cannot read using known formats ['%m/%d/%Y', '%Y-%m-%d']
Search error occurred: Parse of date/time "2022-12-31" failed with date format "dd/MM/yy" in time zone America/Los_Angeles
```

## Expected Behavior
Given it's in the known formats list this should retry with the date format and succeed.

It seems as if it log that it should retry with %Y-%m-%d dateformat (after trying the first date format) but it doesn't go through the list, only uses the default value (first one in the list).

## Logs
[airbyte_netsuite_salesorder_error_logs.zip](https://github.com/airbytehq/airbyte/files/10518500/airbyte_netsuite_salesorder_error_logs.zip)

## Are you willing to submit a PR?
[Yes](https://github.com/airbytehq/airbyte/pull/21971)

octavia-squidington-iii:
At Airbyte, we seek to be clear about the project priorities and roadmap. This issue has not had any activity for 180 days, suggesting that it's not as critical as others. It's possible it has already been fixed. It is being marked as stale and will be closed in 20 days if there is no activity. To keep it open, please comment to let us know why it is important to you and if it is still reproducible on recent versions of Airbyte.

octavia-squidington-iii:
This issue was closed because it has been inactive for 20 days since being marked as stale.

PR #25487, opened 2023-04-25
Source Netsuite: Make input date format configurable (#21965)

rilla:
## What

As explained in #21965, the connector is failing to guess the correct input date format when querying the Netsuite API and this is making the requests to fail under some circumstances. 

Apparently [there's no way to retrieve the correct format from Netsuite via REST API](https://github.com/airbytehq/airbyte/pull/21971#issuecomment-1415695014), so currently the connector tries to guess it using a trial-and-error approach by testing against `DateFormatException`

The current implementation seemed to present some bugs, and @ChloeConnor tried to fix those in #21971. 

But even after fixing them, the current brute-force approach may not be a reliable solution. This is because a date like **Dec 1st 2023** could be represented in multiple ways, such as `dd/mm/YY` (01/12/2023) or `mm/dd/YY` (12/01/2023). Both representations are ambiguous and can be parsed incorrectly using either `%d/%m/%Y` or `%m/%d/%Y` without causing an exception.

We encountered that problem in our local setup,  after implementing the fix from #21971 and it led to further errors, problems and data inconsistencies.

## How

* Remove brute-force approach implementation.
* Add **Netsuite Input Date Format** as a new configurable property in the connector, so the user can populate it with the format matching their Netsuite setup. The default value remains the same (`%m/%d/%Y`). 

## 🚨 User Impact 🚨

Users using formats other than the default one – i.e. `%m/%d/%Y`, will need to populate **Netsuite Input Date Format**.  I don't think this can be considered a breaking change since the current solution wasn't really working in the first place. 

- [ ] Code reviews completed
- [x] Connector version has been incremented
    - [x] Version has been bumped according to our [Semantic Versioning for Connectors](https://docs.airbyte.com/contributing-to-airbyte/#semantic-versioning-for-connectors) guidelines
    - [x] `Dockerfile` has updated version
- [ ] Documentation updated
    - [ ] Connector's `README.md`
    - [ ] Changelog updated in `docs/integrations/<source or destination>/<name>.md` with an entry for the new version. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)

### Airbyter

- [ ] Build is successful

### DIFF
--- a/airbyte-integrations/connectors/source-netsuite/Dockerfile
+++ b/airbyte-integrations/connectors/source-netsuite/Dockerfile
@@ -35,5 +35,5 @@ COPY source_netsuite ./source_netsuite
 ENV AIRBYTE_ENTRYPOINT "python /airbyte/integration_code/main.py"
 ENTRYPOINT ["python", "/airbyte/integration_code/main.py"]
 
-LABEL io.airbyte.version=0.1.3
+LABEL io.airbyte.version=0.1.4
 LABEL io.airbyte.name=airbyte/source-netsuite
--- a/airbyte-integrations/connectors/source-netsuite/sample_files/config.json
+++ b/airbyte-integrations/connectors/source-netsuite/sample_files/config.json
@@ -6,5 +6,6 @@
   "token_secret": "a3c....",
   "start_datetime": "2022-06-01T00:00:01Z",
   "object_types": ["customer"],
-  "window_in_days": 30
+  "window_in_days": 30,
+  "netsuite_input_date_format": "%d/%m/%Y"
 }
--- a/airbyte-integrations/connectors/source-netsuite/sample_files/invalid_config.json
+++ b/airbyte-integrations/connectors/source-netsuite/sample_files/invalid_config.json
@@ -6,5 +6,6 @@
   "token_secret": "a3c....",
   "start_datetime": "2022-06-01T00:00:01Z",
   "object_types": ["customer"],
-  "window_in_days": 30
+  "window_in_days": 30,
+  "netsuite_input_date_format": "%m...",
 }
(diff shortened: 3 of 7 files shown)


bazarnov (commented):
@rilla 
According to: 
> Users using formats other than the default one – i.e. %m/%d/%Y, will need to populate Netsuite Input Date Format. I don't think this can be considered a breaking change since the current solution wasn't really working in the first place.

We should give the clear instructions `how` to get the correct values for each Customer to satisfy the Date Format input requirements, not just `test what works for you`. Can we?

evantahler:
> This approach will work for those who strongly decided to pull out their data from Netsuite, however, there are lots of customers that just want to sync by pressing the single button so I'm very skeptical about the connector setup process, which could cause more incoming issues at that stage, just because they don't actually know how to check the date format they use.

I don't think I follow the converstation.  Are you proposing that:
* Rather than try/catch to guess the date format, we ask users explicitly for the right date format, given some instructions?
* This would be a breaking change as it requires a new type of input

If we can write clear instructions about how to get the right date format, couldn't we pragmatically hit the same API in the connector and load that data up at start for converting the dates we did get?

pquadri:
hi, im stuck on the date step as we are using dd/mm/yyyy (not a choice i can argue against :D), the instructions to get the current specifications are quite easy, it's just under Setup -> Company -> General Preferences and it's literally the first option.
I assume that if anyone is using the current setup you can prefill the new field with a sensible default, and if none of that works you are not breaking anything as it's not working in the first place

image

bmcgilli:
Hi @rilla,
I've closed this pull request, as we won't be able to review it in a timely manner.  

For the time being, Airbyte is focused on improving the quality and performance of other connectors and not able to prioritize this PR for Netsuite.  We owe you a response, and unfortunately that is a 'no' to this change, at this time.

In the future, we will definitely focus on improving more connectors such as Netsuite, and welcome PRs from OSS contributors like yourself.  Please be sure to join our slack community so that you'll be able to learn when we are ready for more contributions like this again. We'll keep you in mind for Netsuite improvements in the future. 

Learn more here: https://airbyte.com/blog/upgrading-our-community-pull-requests-experience

jsvgjsvg:
@rilla Hello! Any further information on your merge request? Running to the same problem.

PR #25489, opened 2023-04-25
connectors-ci: use pip install instead of python -m pip install 

alafanechere:
## What
In the[ `publish` pipeline ](https://github.com/airbytehq/airbyte/actions/runs/4798181638/jobs/8536120015 )the metadata validation step fails because the metadata lib can't be installed:
```
│ │ ModuleNotFoundError: No module named 'click'                             │ │

```

## How
I don't know why exactly but using `pip install` instead of `python -m pip install` fixes the problem.


bnchrch (approved):
Oh man, im not sure why this isnt working now...

The python dependency story is definitely a bit rough around the edges.

Do you have any theories as to why this doesnt work?

But anyway, this does seem like a fine fix, so im happy to approve.

PR #25546, opened 2023-04-26
Connector builder documentation: Add sections about custom params

flash1293:
## What

As discussed above, add sections on how to leverage free-form parameters at the end of the stream config for cases that can't be handled by the more specific injection sections.


lmossman (approved):
A couple of small wording nits that can be applied to the other documents in this PR as well, otherwise this LGTM

PR #25549, opened 2023-04-26
metadata: generate `metadata.yaml` for new connectors.

### LINKED ISSUE #24019: Add metadata.yml to connector generator
Ensure we add a metadata.yml file to all newly generated connectors

alafanechere:
Closes airbytehq/airbyte#24019 

We want any new connector to have a pre-filled `metadata.yaml` file.
This PR : 
- created templates for `metadata.yaml` file for all our generators
- auto generate the definition id of a connector in `plopfile.js`
- update the scaffolding logic to generate scaffolds with definition id.


bnchrch (approved):
Good cloning

airbyte-integrations/connector-templates/generator/plopfile.js
```diff
@@ -105,11 +116,21 @@ module.exports = function (plop) {
       {type: 'emitSuccess', outputPath: pythonDestinationOutputRoot}
     ]
 
-  })
+  })
```
maxi297, line 119: nit: are those trailing spaces needed?

airbyte-integrations/connector-templates/destination-java/metadata.yaml.hbs
```diff
@@ -0,0 +1,19 @@
+data:
+  allowedHosts:
+    hosts:
+      - "*" # Please change to the hostname of the source.
```
maxi297, line 4: Who would change this value and when? It seems like it might be another thing the user needs to fill out. If it's the case, how can we make sure the user fills it? Will it fail in some way if it's not? Should we add this to [the tutorial](https://docs.airbyte.com/connector-development/config-based/tutorial/getting-started)?

airbyte-integrations/connector-templates/generator/build.gradle
```diff
@@ -28,7 +28,7 @@ def addScaffoldTemplateTask(name, packageName, outputDirName, scaffoldParams=[])
             }
             exec {
                 workingDir rootDir
-                def cmd = ['./tools/integrations/manage.sh', 'scaffold', name, packageName]
+                def cmd = ['./tools/integrations/manage.sh', 'scaffold', name, packageName, 'uuid']
```
maxi297, line 31: This field is expected to be a UUID and not the string `uuid`. Does it matter here? If anything can be put here, should we be more descriptive in the string that this can be any string like `any string with uuid format` or whatever?

airbyte-integrations/connector-templates/generator/plopfile.js
```diff
@@ -80,13 +80,24 @@ module.exports = function (plop) {
...
   plop.setActionType('emitSuccess', function(answers, config, plopApi){
       console.log(getSuccessMessage(answers.name, plopApi.renderString(config.outputPath, answers), config.message));
   });
 
   plop.setGenerator('Python Destination', {
     description: 'Generate a destination connector written in Python',
-    prompts: [{type:'input', name:'name', 'message': 'Connector name e.g: redis'}],
+    prompts: [
+      {type:'input', name:'name', 'message': 'Connector name e.g: redis'},
+      {
```
maxi297, line 92: This allows the user to enter whatever he would like. Do we need validation like:
* This is actually a UUID?
* The UUID does not already exist for another connector?

maxi297 (commented):
Can you provide more contexte as to why we need a metadata file for all the connectors now? What are we trying to achieve? What has changed so that we need this now? I feel like https://github.com/airbytehq/airbyte/issues/24019 describes the solution very well but not really the problem. 

Given that new information, I'll check the PR again and approve

alafanechere:
@maxi297 we introduced the `metadata.yaml` for a lot of different usecases. The first one is to not rely on central `*_definitions.yaml` files anymore and generate the connector registry (oss and cloud) from these file. We're making it a requirement for all connectors to have a `metadata.yaml` as it's about to play a central role in a lot of different project we're tackling. E.G we'll detect connector version changes from this file to trigger connector publish and connector registry update. Feel free to read this [tech spec](https://docs.google.com/document/d/1CXG0hTftomCTO3vB0ocbs00CkYBD6YURfrxm8KwozB4/edit#heading=h.fa72b39y2m99) for more details.

airbyte-integrations/connector-templates/generator/build.gradle
alafanechere: What matters is to have an artificial static value here to have idempotent results when a scaffolds happen in the CI, as the CI compare the scaffolds output to the previously versionned scaffolds. I'll change it to a more intuitive value as you suggested.

airbyte-integrations/connector-templates/destination-java/metadata.yaml.hbs
alafanechere: I feel like its a bit early in the dev process to ask developers to fill this in with an extra prompt. I think the `*` value is valid but we likely don't want wildcards there for security reasons.
I suggest we :
- Document in the tutorial the purpose of the metadata.yaml file and how to fill the fields (https://github.com/airbytehq/airbyte/issues/25612)
- Add a logic in the  CI metadata validation to fail if `*` is used without a specific comment explaining why the wildcard is used. (https://github.com/airbytehq/airbyte/issues/25613)

airbyte-integrations/connector-templates/generator/plopfile.js
alafanechere: > This is actually a UUID?
@maxi297 I added a validation function to check if the UUID is a valid one. 

> The UUID does not already exist for another connector?
I feel like it'd be a bit overkill to validate this in plop. But we can add a QA check in the CI to test that a definition id belongs to a single connector. (#25618)

maxi297 (approved):
Thanks for the context @alafanechere ! Things make sense to me and I approve the change

Issue #24577, opened 2023-03-27
Prepare real connectors as reference manifests


flash1293:
To be able to see the theoretical documentation about low code manifests in practice, some existing connectors (3-5) should be prepared and referenced in the documentation. Each of these connectors manifest should be importable in connector builder with perfect fidelity (inline schemas, no custom components, inline spec).

To select the connectors, the following properties would make for a good fit:
* Uses many features
* Is a well-known API
* Does not require authentication / authentication can easily be set up by anyone

As part of this issue, fitting connectors should be selected, prepared (e.g. inlining schemas and specs) and referenced in the docs

PR #25554, opened 2023-04-26
source-news-api: Make manifest work in builder

flash1293:
## What

Part of https://github.com/airbytehq/airbyte/issues/24577

This PR inlines schema and spec in the manifest so it can be fully loaded in the connector builder by uploading the file.
The functionality of the connector is not changed.


girarda (approved):
I know this should not introduce a change in behavior, but I think we should still release a new version of the connector to keep what's running in sync with the code.

can you update the [changelog](https://github.com/airbytehq/airbyte/blob/master/docs/integrations/sources/news-api.md?plain=1#L61) and the [version in the dockerfile](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-news-api/Dockerfile#L37) before publishing (`/publish connector=connectors/source-news-api run-tests=false`) (fine to skip the tests since they fail on master too 😢 )

airbyte-integrations/connectors/source-news-api/source_news_api/manifest.yaml
```diff
@@ -71,3 +139,191 @@ check:
...
+        description: |
+          A date and optional time for the oldest article allowed. This should
+          be in ISO 8601 format (e.g. 2021-01-01 or 2021-01-01T12:00:00).
+        pattern: "^[0-9]{4}-[0-9]{2}-[0-9]{2}(T[0-9]{2}:[0-9]{2}:[0-9]{2})?$"
+        order: 6
+      end_date:
+        type: string
+        description: |
+          A date and optional time for the newest article allowed. This should
+          be in ISO 8601 format (e.g. 2021-01-01 or 2021-01-01T12:00:00).
```
girarda, line 220: @flash1293 can we change this line since it's causing permissions error on Cloud and we plan on using this connector as an example? I think we'd just need to remove the "or"

airbyte-integrations/connectors/source-news-api/source_news_api/manifest.yaml
```diff
@@ -71,3 +139,191 @@ check:
...
+          A comma-seperated string of domains (eg bbc.co.uk, techcrunch.com,
+          engadget.com) to remove from the results.
+        items:
+          type: string
+        order: 5
+      start_date:
+        type: string
+        description: |
+          A date and optional time for the oldest article allowed. This should
+          be in ISO 8601 format (e.g. 2021-01-01 or 2021-01-01T12:00:00).
```
girarda, line 213: @flash1293 can we change this line since it's causing permissions error on Cloud and we plan on using this connector as an example? I think we'd just need to remove the "or"

flash1293:
/publish connector=connectors/source-news-api run-tests=false

flash1293:
/publish connector=connectors/source-news-api run-tests=false

Issue #25053, opened 2023-04-11
connectors-ci: compare legacy and dagger results for nightly builds


alafanechere:
After the merge of https://github.com/airbytehq/airbyte/pull/24225 all the GA and Beta connectors (Python and Java) will be tested during nightly builds. We should compare the nightly builds results to spot discrepancies.
If discrepancies are found we shall fix these.

We can iterate on the SQL queries built for this [dashboard](https://airbyte.metabaseapp.com/dashboard/279-dagger-poc) to spot discrepancies.

bnchrch:
Grooming
- For the monitoring dashboard the columns would be (connector name, legacy test result, dagger test result)

alafanechere:
I spotted a lot of CAT tests error on nightly builds that I can't reproduce locally:
The `test_config_match_spec` fails on a lot of GA connectors.
I suspect it's because the connector under test image is not the right one passed to CAT.
This could have something to do with filename collision.
I tried a different tar file naming approach in https://github.com/airbytehq/airbyte/pull/25477

alafanechere:
Some python connectors could not be built with the "one docker file for all in dagger" approach. To get stable connector build I decided to keep using the connector specific docker files:
https://github.com/airbytehq/airbyte/pull/25483

alafanechere:
On nightly builds, a lot of python connectors are failing the `expected_records` or `backward_compatibility` tests, while they are not on `/test`.
E.G on [this nightly build ](https://github.com/airbytehq/airbyte/actions/runs/4860447035/jobs/8664278862)the `source-twilio` connectors failed `test_backward_compatibility`, but it is not failing it on `/test`.

alafanechere:
`source-file` integration tests are failing: it relies on `docker-compose` which is not available in the integration test container in dagger.

alafanechere:
On today's [nightly build ](https://github.com/airbytehq/airbyte/actions/runs/5187246645)the following connector failed CAT or integration test in our new pipeline but not on /test:
* source-jira
* source-gitlab
* source-prestashop
* source-facebook-marketing

alafanechere:
I confirmed that I wasn't able to reproduce the remaining inconsistencies when running tests on impacted connector with airbyte-ci or direct CAT. When I run airbyte-ci and CAT at the same moment I get consistent results...
The remaining inconsistencies are mainly related to expected records different results.

So I consider they are due to the following reasons:
[...]

PR #25557, opened 2023-04-26
connectors-ci: load secrets on unit tests

alafanechere:
## What
Relates to #25053 

I figured that some connector rely on the existence of secrets configs to run unit tests.
I think it's a bad pattern but I'd like our CI pipeline to adapt to this reality instead of fixing these connectors.
(This connector is source-tiktok-marketing)

## How
Download secrets before running unit tests.
Mount the secret directory to the container under test in unit tests.


airbyte-integrations/connectors/source-tiktok-marketing/metadata.yaml
```diff
@@ -11,7 +11,7 @@ data:
   connectorSubtype: api
   connectorType: source
   definitionId: 4bfac00d-ce15-44ff-95b9-9e3c3e8fbd35
-  dockerImageTag: 2.0.6
+  dockerImageTag: 3.0.1
```
alafanechere, line 14: I'm bumping the version here to trigger the tests on this source and eventually publish on merge.

bnchrch (approved):
Not a great thing we have to do, but I understand the why.

PR #25558, opened 2023-04-26
source-coingecko-coins: Make manifest work in builder

flash1293:
## What

Part of https://github.com/airbytehq/airbyte/issues/24577

This PR inlines schema and spec in the manifest so it can be fully loaded in the connector builder by uploading the file.
The functionality of the connector is not changed.


girarda (approved):
lgtm! can you update the changelog and the dockerfile before merging?

PR #25562, opened 2023-04-26
connectors-ci: unpin pytest on specific connectors

alafanechere:
## What
Relates to #25053

When running unit/integration tests on python connectors we build a connector testing environment in a container by:
- Installing the testing library listed [here](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py#L21) . To replicate what's [happening in `airbyte-python.gradle` ](https://github.com/airbytehq/airbyte/blob/155a4c97654c7e90be43f7017230dd922c28fe80/buildSrc/src/main/groovy/airbyte-python.gradle#L80)
- Load the root pyproject.toml which specify minversion for pytest in this container
- Install the connector under test as a python package, with the specific test requirements it might have

The probem is that some connector explicitely pin a pytest version which is lower than the global one we pin in the main `pyproject.toml`: `pytest:6.2.5`.

## How
I suggest to unpin the version of pytest in the `setup.py` of the couple of connectors explicitely pinning a pytest version.


evantahler (approved):
Screenshot 2023-04-26 at 9 25 23 AM
Tests passed... so... :shipit:

bnchrch (approved):

alafanechere:
/publish connector=connectors/source-smartsheets run-tests=false

PR #23466, opened 2023-02-26
Destination-S3: Correctly generate int64 values

etsybaev:
## What
As noted in [#14362 (comment)](https://github.com/airbytehq/airbyte/pull/14362#discussion_r928033781), dest-s3's avro and parquet output formats currently use int32. This causes problems for users with int64 values. We should have dest-s3 produce int64 values for integer inputs.

The reason we use int32 right now is because we want to support union types, which behave kind of weirdly in edge cases.

## How
Updated S3 to use Avro.LONG type instead of Avro.INT

## 🚨 User Impact 🚨
At the first glance, no breaking changes are expected. In the edge case user might need to reset connection and re-sync data.
**Important note: If we would get oneOf ("long" and "long, specific type: timestamp" - the value would be treated as a long**

## Pre-merge Checklist
Expand the relevant checklist and delete the others.

<details><summary><strong>New Connector</strong></summary>

### Community member or Airbyter

- [ ] **Community member?** Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- [ ] Secrets in the connector's spec are annotated with `airbyte_secret`
- [ ] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- [ ] Code reviews completed
- [ ] Documentation updated
    - [ ] Connector's `README.md`
    - [ ] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - [ ] `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
    - [ ] `docs/integrations/README.md`
    - [ ] `airbyte-integrations/builds.md`
- [ ] PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests)

### Airbyter

If this is a community PR, the Airbyte engineer reviewing this PR is responsible for the below items.

[...]

PR #25570, opened 2023-04-26
Avro destinations: map airbyte_type: integer to long

### LINKED ISSUE #23760: Source MySql Destination S3 (parquet) Support: Negative IDs / integers for MySql Source to AWS S3 Destination in parquet file format
## Environment
- **Airbyte version**:  0.40.30
- **OS Version / Instance**: Ubuntu, AWS EC2
- **Deployment**: Docker
- **Source Connector and version**: airbyte/source-mysql:1.0.21
- **Destination Connector and version**: airbyte/destination-s3:0.3.19
- **Step where error happened**: 
- 1. Create a connection between Zendesk Support and S3, in a parquet format.
- 2. Load the data from parquet to DataFrame (pyspark/pandas) or any other.

## Current Behavior
* It will be possible to see IDs as negative floats when it should be integers. Also, the numbers don’t match the original ID on MySql*

## Expected Behavior
* IDs should come as in source, exact values*

## Logs
image

edgao:
## What
Closes https://github.com/airbytehq/airbyte/issues/23760

https://github.com/airbytehq/airbyte/pull/23466 only fixed this behavior for `{type: integer}`. This PR adds the fix for `{type: number, airbyte_type: integer}`.

Also updated test cases:
* add a test case for `{type: integer}`
* update the avroSchema for `{type: number, airbyte_type: integer}` to be `long`

(also, submitted an issue to clean up the data types stuff - it's a pretty messy change so will do it separately https://github.com/airbytehq/airbyte-internal-issues/issues/1726)


evantahler (approved):
Tests pass, so :shipit:!

edgao:
/publish connector=connectors/destination-bigquery,connectors/destination-bigquery-denormalized,connectors/destination-gcs,connectors/destination-s3,connectors/destination-snowflake

TBernstein4:
@edgao is there a chance that this would change the datatype format in the following manner:

`1676500610` now formatted as `1.678177401E9`?

edgao:
I would not expect that - that actually sounds like it's being treated as a number/float. Probably easier to discuss over slack / separate issue though

PR #25586, opened 2023-04-26
Don't convert a string to binary base64 unnecessarily

### LINKED ISSUE #9938: Source MySQL: support `utf8_bin` charset
## Environment
- **Airbyte version**:  0.35.12
- **OS Version / Instance**: example macOS, Windows 7/10, Ubuntu 18.04, GCP n2. , AWS EC2
- **Source Connector and version**: Mysql 0.5.1
- **Severity**: Medium

## Current Behavior
Replication of columns that are using `utf8_bin` charset results in base64 encoded data.

[Related slack conversation](https://airbytehq-team.slack.com/archives/C019WJFR1JT/p1642569581190900)

Raw data example:
Screenshot 2022-01-19 183339

## Expected Behavior
Syncs should replicate decoded utf8 strings.

## Logs
Source table schema:
```sql
CREATE TABLE `INV_LOT_LOC_ID` (
  `organizationId` varchar(20) COLLATE utf8_bin NOT NULL,
  `warehouseId` varchar(20) COLLATE utf8_bin NOT NULL,
  `locationId` varchar(60) COLLATE utf8_bin NOT NULL DEFAULT '',
  `lotNum` varchar(10) COLLATE utf8_bin NOT NULL,
  `traceId` varchar(30) COLLATE utf8_bin NOT NULL,
  `customerId` varchar(30) COLLATE utf8_bin NOT NULL,
  `sku` varchar(50) COLLATE utf8_bin NOT NULL,
  `qty` decimal(18,8) NOT NULL,
  `qtyAllocated` decimal(18,8) NOT NULL,
  `qtyRpIn` decimal(23,8) NOT NULL,
  `qtyRpOut` decimal(23,8) NOT NULL,
  `qtyMvIn` decimal(23,8) NOT NULL,
  `qtyMvOut` decimal(23,8) NOT NULL,
  `qtyOnHold` decimal(23,8) NOT NULL,
  `onHoldLocker` int(11) NOT NULL DEFAULT '0',
  `grossWeight` decimal(18,8) NOT NULL,
  `netWeight` decimal(18,8) NOT NULL,
  `cubic` decimal(18,8) NOT NULL,
  `price` decimal(24,8) DEFAULT NULL,
  `lpn` varchar(30) COLLATE utf8_bin DEFAULT NULL,
  `qtyPa` decimal(18,8) NOT NULL DEFAULT '0.00000000',
  `qcStatus` varchar(2) COLLATE utf8_bin DEFAULT NULL,
  `lastMaintenanceDate` timestamp NULL DEFAULT NULL,
  `noteText` mediumtext COLLATE utf8_bin,
  `udf01` varchar(500) COLLATE utf8_bin DEFAULT NULL,
  `udf02` varchar(500) COLLATE utf8_bin DEFAULT NULL,
  `udf03` varchar(500) COLLATE utf8_bin DEFAULT NULL,
  `udf04` varchar(500) COLLATE utf8_bin DEFAULT NULL,
  `udf05` varchar(500) COLLATE utf8_bin DEFAULT NULL,
  `currentVersion` int(11) NOT NULL DEFAULT '100',
  `oprSeqFlag` varchar(65) COLLATE utf8_bin NOT NULL DEFAULT '2016',
  `addWho` varchar(40) COLLATE utf8_bin DEFAULT NULL,
  `addTime` timestamp NULL DEFAULT NULL,
  `editWho` varchar(40) COLLATE utf8_bin DEFAULT NULL,
  `editTime` timestamp NULL DEFAULT NULL,
  `inLocTime` timestamp NULL DEFAULT NULL,
  PRIMARY KEY (`organizationId`,`warehouseId`,`locationId`,`lotNum`,`traceId`),
  KEY `auto_shard_key_organizationId` (`organizationId`),
  KEY `IDX_INV_LOT_LOC_ID_C` (`organizationId`,`warehouseId`,`locationId`,`sku`),
  KEY `IDX_INV_LOT_LOC_ID_QQQQQ` (`organizationId`,`warehouseId`,`qty`,`qtyPa`,`qtyMvIn`,`qtyRpIn`,`qtyAllocated`),
  KEY `IDX_LOTXLOCXID_SKU` (`organizationId`,`warehouseId`,`sku`),
  KEY `I_INV_LOT_LOC_ID_A` (`organizationId`,`warehouseId`,`customerId`,`sku`),
  KEY `I_INV_LOT_LOC_ID_B` (`organizationId`,`warehouseId`,`traceId`),
  KEY `I_INV_LOT_LOC_ID_C` (`organizationId`,`warehouseId`,`qty`,`qtyMvIn`,`qtyRpIn`),
  KEY `I_INV_LOT_LOC_ID_LOT` (`organizationId`,`warehouseId`,`lotNum`),
  KEY `I_INV_LOT_LOC_ID_LOC` (`locationId`),
  KEY `I_INV_LOT_LOC_ID_SKU` (`organizationId`,`customerId`,`sku`),
  KEY `I_INV_LOT_LOC_ID_LOT2` (`organizationId`,`lotNum`),
  KEY `I_INV_LOT_LOC_ID_LO` (`lotNum`,`organizationId`),
  KEY `I_INV_LOT_LOC_ID_SCW` (`sku`,`customerId`,`warehouseId`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin
```

## Steps to Reproduce
1. Create a MySQL table with `varchar(500) COLLATE utf8_bin DEFAULT NULL`
2. Replicate the table
3. Check the output data

rodireich:
## What
Mysql column that is encoded with any character set and collated (sorted) in a binary collation logic (e.g utf8mb4_bin) is today treated as a binary value and converted to base64 on detonation side.
This is not necessary as the value itself is still a simple string.
We need to treat it as a string and create one on destination.

## How
Our code is checking incorrectly the `.isBinary()` property of a column when deciding how to convert it.
Remove this logic when copying to json.

## 🚨 User Impact 🚨
Any string value previously synced as base64 will remain so until a reset or a full refresh.
Newly synced values will appear as the original on source side ("abcd").

Resolves #9938 .


airbyte-integrations/connectors/source-mysql/src/main/java/io/airbyte/integrations/source/mysql/MySqlSourceOperations.java
```diff
@@ -111,14 +111,7 @@ public void copyToJsonField(final ResultSet resultSet, final int colIndex, final
...
       case TIME -> putTime(json, columnName, resultSet, colIndex);
-      case CHAR, VARCHAR -> {
-        if (field.isBinary()) {
-          // when character set is binary, the returned value is binary
-          putBinary(json, columnName, resultSet, colIndex);
-        } else {
-          putString(json, columnName, resultSet, colIndex);
-        }
-      }
+      case CHAR, VARCHAR -> putString(json, columnName, resultSet, colIndex);
```
prateekmukhedkar, line 114: Do we need a test case for binary fields to ensure that conversion to string logic is taking place as expected?

edgao:
checking my understanding:
* current behavior is that we discover these columns as `string` and read records as `base64encode(select the_column)`
* new behavior is still discovering string, but  no longer base64 encoding the data?

this sounds like a breaking change. Do we need any coordination to release this? Should we recommend that users run a reset after upgrading? (or they can manually rewrite their raw tables + rebuild normalized tables I guess, but that sounds complicated)

airbyte-integrations/connectors/source-mysql/src/main/java/io/airbyte/integrations/source/mysql/MySqlSourceOperations.java
rodireich: This code change is putting us back to handling VARCHAR the same way under any charset or collation.

We don't have tests for variation of charset as far as I can tell, but the change itself is returning to the way this used to behave before checking for `.isBinary()` was added.

rodireich:
What you describe @edgao is correct - that's the change.
But I thought of it as non breaking as old values that already synced to destination as base64 would not affect new values synced as a simple string.
Do you think we should flag this beforehand to customers so they can be aware and reset or re-run normalization ?

edgao:
presumably there are some folks who need to manually base64-decode their data right now; we should at least tell them to stop doing that decode.

does our schema include anything like `format: base64`? I.e. can we continue reading base64 for existing users, but new users (or after schema refresh) start reading normal strings?

rodireich:
> does our schema include anything like format: base64
I don't think that's the case. 
The column in the case above was marked correctly as a string but its value was treated like a binary.
Note that the only thing binary here was the column collation order. not any encoding or data.

I'm not sure what can be done other than notifying customers that may have reverse conversion on they destination.
I'm not sure we can even check for existence of such cloud connections as the type is string.

edgao:
sorry, what does this mean?

> the only thing binary here was the column collation order. not any encoding or data

rodireich:
The case here is db with charset encoding of utf8,
and a collation (or sorting) order of utf8_bin.
```
CREATE TABLE `INV_LOT_LOC_ID` ( … ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_bin
```

The data it encoded in plain utf8 but when sorted for example in `ORDER BY`, binary logic is used which is comparing byte value rather than go into all the intricacies of variable length multibyte utf8 characters.
For example uff aware sorting may treat `ss` and `ß` as equals while binary would place `ß` after `z`

Not to be confused with Binary encoding that has its own "binary" collation order which is unrelated to this case and will still get destination strings converted to base64 as it should.
See https://dev.mysql.com/doc/refman/5.7/en/charset-binary-collations.html

edgao (approved):
ah got it :/ that seems challenging to coordinate then, if we can't even recognize which workspaces are affected...

is our plan to just tell all source-mysql users something like "check your source DB for columns with `COLLATE=utf8_bin` and run a reset if you have them?" (or maybe "update your query to attempt base64_decode, and fallback to the raw value if that decode fails"?)

approving since I have no opinion on the code change, but I think the release plan needs some clarification. (destinations has no problems with this, since the schema was and will remain `type: string`)

rodireich:
Thanks Ed
And this is largely the plan - alert users to check if their connection is affected. Thanks for suggesting fallback in case a decode fails. that should probably solve it for most.
I will coordinate with cloud and community support to send out a message.

rodireich:
Airbyte technical support:
Once this change goes in, previously synced mysql-source connections that are syncing:
A database with any charset encoding which is set with an *_bin collation order (sorting) 
For example `hebrew` charset with `hebrew_bin` collation.

Will see the following potentially breaking change:
Strings that were previously encoded on destination in base64 will now show up in their regular form 'abcd' rather than 'YWJjZA=='.
This means that a table that was previously synced with VARCHAR columns converted to base64 will now have new values synced as simple string.

What can be done:
a. Reset of table data on airbyte will get a fresh copy of all rows with now plain 'abcd' string.
b. If there's any script or procedure on destination that is consuming these base64 encoded columns, it should be updated to also be able to process plain strings (e.g if base64 decode failed use the string as is).

Note: the issue won't affect anybody who is running with default settings but only customers that intentionally created a binary collated DB or tables or columns.

cc: @erica-airbyte

erica-airbyte:
Thanks @rodireich can you fill out the[ breaking changes doc](https://docs.google.com/document/d/1VYQggHbL_PN0dDDu7rCyzBLGRtX-R3cpwXaY8QxEgzw/edit)?

rodireich:
See https://docs.google.com/document/d/1BmiM93agJ6iVGdJek0a8lAWdH-02JLFzStAEjGmdSCc/edit?usp=sharing

prateekmukhedkar (approved):

amaliaroye:
Hi @rodireich, is this still scheduled to be released tomorrow, May 24th?

PR #25589, opened 2023-04-26
low-code: Use Jinja sandbox environment and prevent use of range method

girarda:
## What
Closes https://github.com/airbytehq/airbyte/pull/25589/

## How
* Use Jinja sandbox environment as per [Jinja's recommendation](https://jinja.palletsprojects.com/en/3.0.x/sandbox/#jinja2.sandbox.SandboxedEnvironment) (I did not find any input that caused an issue without the sanbox, but was fixed with it)
* Restrict use of range method within Jinja because it can be used to clog the CPU
* Add unit tests to confirm it's not possible to import libraries and access global values that are not explicitly added to the jinja context

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/interpolation/jinja.py`
2. `airbyte-cdk/python/unit_tests/sources/declarative/interpolation/test_jinja.py`


airbyte-cdk/python/unit_tests/sources/declarative/interpolation/test_jinja.py
```diff
@@ -138,7 +137,39 @@ def test_negative_day_delta():
...
+       {% endfor %}
+        {{ a }}""", id="test_jinja_with_very_long_running_compute"),
+    pytest.param("{{ eval ('2+2') }}", id="test_jinja_with_eval"),
+    pytest.param("{{ getattr(config, 'key') }}", id="test_getattr"),
+    pytest.param("{{ setattr(config, 'password', 'hunter2') }}", id="test_setattr"),
+    pytest.param("{{ globals()  }}", id="test_jinja_with_globals"),
+    pytest.param("{{ locals()  }}", id="test_jinja_with_globals"),
+    pytest.param("{{ eval ('2+2') }}", id="test_jinja_with_eval"),
+])
+def test_restricted_builtin_functions_are_not_executed(template_string):
```
sherifnada, line 172: is there a way to make this raise exceptions?

girarda: No because they're not special keywords to Jinja. I just added the tests to confirm that assumption

brianjlai: as per in person chat, we might be able to add a condition during evaluation for disallowed phrases like `in range(` and throw a ValueError if the interpolation string contains this since its invalid. We'll still remove it from the context for extra safety, but this could fail more loudly

girarda: Updated [the `JinjaInterpolation` to raise an error if a variable is undefined](https://github.com/airbytehq/airbyte/pull/25589/files#diff-df8a64d2d501197620f175b1b69fcbb3ad0117f621cfe3fbc387b4e43b39f34aR94-R98) instead of blindly returning `None`

airbyte-cdk/python/unit_tests/sources/declarative/interpolation/test_jinja.py
```diff
@@ -57,24 +58,12 @@ def test_literals(s, value):
             "{{ stream_slice['stream_slice_key'] }}",
             "hello",
             id="test_get_value_from_stream_slice"),
-        pytest.param(
```
girarda, line 60: this is not true anymore. we'll raise an exception if the variable is not present in the context

airbyte-cdk/python/unit_tests/sources/declarative/interpolation/test_jinja.py
```diff
@@ -57,24 +58,12 @@ def test_literals(s, value):
...
-            None,
-            id="test_get_value_from_stream_slice_no_slice"
-        ),
         pytest.param(
             {"stream_slice": {"stream_slice_key": "hello"}},
             "{{ stream_partition['stream_slice_key'] }}",
             "hello",
             id="test_get_value_from_stream_slicer"
         ),
-        pytest.param(
```
girarda, line 72: this is not true anymore. we'll raise an exception if the variable is not present in the context

airbyte-cdk/python/unit_tests/sources/declarative/requesters/request_options/test_interpolated_request_options_provider.py
```diff
@@ -21,7 +21,6 @@
         ("test_value_depends_on_stream_slice", {"read_from_slice": "{{ stream_slice['start_date'] }}"}, {"read_from_slice": "2020-01-01"}),
         ("test_value_depends_on_next_page_token", {"read_from_token": "{{ next_page_token['offset'] }}"}, {"read_from_token": 12345}),
         ("test_value_depends_on_config", {"read_from_config": "{{ config['option'] }}"}, {"read_from_config": "OPTION"}),
-        ("test_missing_value", {"missing_param": "{{ fake_path['date'] }}"}, {}),
```
girarda, line 24: this is not true anymore. we'll raise an exception if the variable is not present in the context

sherifnada (approved):

airbyte-cdk/python/airbyte_cdk/sources/declarative/interpolation/jinja.py
```diff
@@ -75,6 +91,11 @@ def _literal_eval(self, result):
 
     def _eval(self, s: str, context):
         try:
+            ast = self._environment.parse(s)
+            undeclared = meta.find_undeclared_variables(ast)
+            undeclared_not_in_context = {var for var in undeclared if var not in context}
+            if undeclared_not_in_context:
+                raise ValueError(f"Jinja macro has undeclared variables: {undeclared}. Context: {context}")
```
brianjlai, line 98: this should probably be`undeclared_not_in_context` otherwise we'll list valid template variables that are in the context

brianjlai (approved):
approving, but i'm going to push a commit to fix the error message and add some unit tests to eval to verify it gets thrown not thrown. And since alex is out i'll merge this one in next week

PR #25593, opened 2023-04-26
Remove check_images_exist.sh

### LINKED ISSUE #25537: Remove check_images_exist.sh

bnchrch:
## What
Remove this as
1. It references the _definitions.yaml file
2. Our new CI should be doing this check

closes #25537


alafanechere (approved):

alafanechere (changes requested):
When you say "our new CI should be doing this check"., do you mean it's now `metadata validation` role to perform this check?
I'm OK with it but  think it might be too soon to remove this script:
- We have no guarantee ATM that the version declared in the metadata is the latest one.
- The [check](https://github.com/airbytehq/airbyte/pull/25609) I added to validate the version in the metadata file is updated is in the connector test pipeline. And this pipeline is not yet considered production ready.
- We could also make this check in the[ `validate` pipeline](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/metadata.py#L39), but It'll require a couple of adaptation on https://github.com/airbytehq/airbyte/pull/25609
- We should update docs / socialize this change for awareness.
It also mean that we should not move too fast the metadata validation inside the CI connector tests and [keep this workflow](https://github.com/airbytehq/airbyte/blob/master/.github/workflows/metadata_validate_manifest_dagger.yml#L1) until the CI connector tests are not

PR #25597, opened 2023-04-27
Remove connectors.md and builds.md

### LINKED ISSUE #25456: Remove connectors.md and builds.md (update docs as well)
## Overview

This report:
https://connectors.airbyte.com/files/generated_reports/connector_registry_report.html

is replacing builds.md and connectors.md

so we must

- [ ] delete builds.md
- [ ] delete connectors.md
- [ ] delete generation logic for builds.md
- [ ] delete generation logic for connectors.md
- [ ] update docs for builds.md
- [ ] update docs for connectors.md

bnchrch:
## What
This report:
https://connectors.airbyte.com/files/generated_reports/connector_registry_report.html

is replacing builds.md and connectors.md

closes #25456


alafanechere (changes requested):
I think we miss links to connector build page on badges to remove `builds.md`.

alafanechere:
Do you have an issue logged to improve the design of this page? 
I think it currently has a couple of columns that are not useful for users (definition id, cloud/oss image match etc.)

bnchrch:
Good catch on the badge

@alafanechere I just updated that here in this [PR](https://github.com/airbytehq/airbyte/pull/25630)

In terms of updates to the report, feel free to log issues!

My interpretation was that this report is meant for internal which is why I added the do match columns here.

Do you think we should have both a public and an internal report?

connectors.md
alafanechere, line 1: I think that some community users, or the support team, might rely on this file as their source of truth. To advertise migration I would suggest adding a simple link to your new report instead of deleting the file. 
It' be even better if we could make the orchestrator commit a md version on the repo, but overkill maybe.

alafanechere (approved):

PR #25601, opened 2023-04-27
🐛 Source Hubspot: fix `CAT` tests

bazarnov:
## What
Based on this CAT test: https://github.com/airbytehq/airbyte/actions/runs/4815554992
need to  ignore certain fields for `companies` stream to pass the CAT

## How
- ignore `properties/hs_all_*` while `basic_read` to satisfy `expected records` check


artem1205 (approved):

airbyte-integrations/connectors/source-hubspot/acceptance-test-config.yml
```diff
@@ -43,6 +43,8 @@ acceptance_tests:
         companies:
         - name: properties/hs_time_*
           bypass_reason: Hubspot time depend on current time
+        - name: properties/hs_all_*
+          bypass_reason: This property is floating from sync to sync
```
erohmensing, line 47: This is ignoring all `hs_all_whatever` fields? I find it hard to believe that all of them float if the data's not changing. E.g. these ones are probably stable: `"hs_all_accessible_team_ids": null, "hs_all_owner_ids": "52550153", "hs_all_team_ids": null,` Are there specific ones that are causing issues?

erohmensing: I don't understand why they could change though 😅 there won't be new owners or teams, etc. But now I'm also confused, as there are still `hs_all` fields in the expeceted records, so maybe this isn't ignoring what I think it is?

bazarnov: Ok, I'll remove this from `acceptance-test-config.yaml`

erohmensing: Just for posterity's sake, looking at the diff, I see some completely new `hs_all` fields, which is annoying when hubspot changes stuff, but I think its more of "hubspot is giving us new fields" and not that the data within that field is changing:

image

I think it makes sense to update the expected records to contain the new fields in that case.

PR #25609, opened 2023-04-27
connectors-ci: new checks to validate connector version

### LINKED ISSUE #25329: CI: Check that the docker version in the metadata file has changed if any other files have changed in the connector folder
## Overview
If any file has changed in a connector folder, ensure that the version has changed.

## Solution
1. parse the local metadata.yml file
1. parse the remote metadata.yml file (github raw)
1. check if the dockerImageTag version is semver and has increased, (ignore in the prerelase version case `1.2.3-dev.asd213`
1. if it has increased, pass, if not, fail

## Extra considerations
Lets use this as a way to gate expensive tests.

So this check should be run early in the CI pipeline, if it fails then skip remaining tests.

alafanechere:
## What
Closes #25329

When a connector is modified we want to enforce a connector version increment in the `metadata.yaml` file in the `dockerImageTag` field.
The version that will eventually be published on merge is pulled from the metadata file, it's why we want to check the version is bumped in this file.

## How
If the CI context is not master and connectors files were modified:
* Check that the `dockerImageTag` in the `metadata.yaml` file follows semantic versionning
* Pull the `master` version of `metadata.yaml` from `raw.github.com`
* Check that the current connector version is an increment of the master version.

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
@@ -16,6 +16,7 @@ This documentation should be helpful for both local and CI use of the CLI. We in
 ### Install
 ```bash
 # Make sure that the current Python version is >= 3.10
+pyenv shell 3.10
 pip install "ci-connector-ops[pipelines] @ git+https://github.com/airbytehq/airbyte.git@master#subdirectory=tools/ci_connector_ops"
 cd airbyte
 airbyte-ci
@@ -108,6 +109,11 @@ Test connectors changed on the current branch:
 ```mermaid
 flowchart TD
     entrypoint[[For each selected connector]]
+    subgraph version ["Connector version checks"]
+        sem["Check version follows semantic versionning"]
+        incr["Check version is incremented"]
+        sem --> incr
+    end
     subgraph static ["Static code analysis"]
       qa[Run QA checks]
       fmt[Run code format checks]
@@ -126,8 +132,9 @@ flowchart TD
         build-->integration
         build-->cat
     end
-    entrypoint-->static
-    entrypoint-->tests
+    entrypoint-->version
+    version-->static
+    version-->tests
     report["Build test report"]
     tests-->report
     static-->report
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/tests/__init__.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/tests/__init__.py
@@ -8,7 +8,7 @@ from typing import List
 
 import anyio
 import asyncer
-from ci_connector_ops.pipelines.bases import ConnectorReport, StepResult, StepStatus
+from ci_connector_ops.pipelines.bases import ConnectorReport, StepResult
 from ci_connector_ops.pipelines.contexts import ConnectorContext
 from ci_connector_ops.pipelines.tests import java_connectors, python_connectors
 from ci_connector_ops.pipelines.tests.common import QaChecks, VersionFollowsSemverCheck, VersionIncrementCheck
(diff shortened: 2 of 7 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
```diff
@@ -16,6 +16,7 @@ This documentation should be helpful for both local and CI use of the CLI. We in
 ### Install
 ```bash
 # Make sure that the current Python version is >= 3.10
+pyenv shell 3.10
```
alafanechere, line 19: Unrelated doc addition 😄

tools/ci_connector_ops/ci_connector_ops/pipelines/README.md
```diff
@@ -108,6 +109,11 @@ Test connectors changed on the current branch:
 ```mermaid
 flowchart TD
     entrypoint[[For each selected connector]]
+    subgraph version ["Connector version checks"]
```
bnchrch, line 112: <3

tools/ci_connector_ops/ci_connector_ops/pipelines/tests/__init__.py
```diff
@@ -28,6 +28,19 @@
...
+    """Run the version checks on a connector.
+
+    Args:
+        context (ConnectorContext): The current connector context.
+
+    Returns:
+        List[StepResult]: The results of the version checks steps.
+    """
+    context.logger.info("Run version checks.")
+    return [await VersionFollowsSemverCheck(context).run(), await VersionIncrementCheck(context).run()]
```
bnchrch, line 41: Like this pattern!

tools/ci_connector_ops/ci_connector_ops/pipelines/tests/__init__.py
```diff
@@ -87,13 +100,18 @@ async def run_connector_test_pipeline(context: ConnectorContext, semaphore: anyi
     """
     async with semaphore:
         async with context:
+            version_checks_results = await run_version_checks(context)
```
bnchrch, line 103: Just thinking about the skip on failure login in `run_tests`, could we use that here?

tools/ci_connector_ops/ci_connector_ops/pipelines/tests/common.py
```diff
@@ -4,13 +4,90 @@
...
+    GITHUB_URL_PREFIX_FOR_CONNECTORS = "https://raw.githubusercontent.com/airbytehq/airbyte/master/airbyte-integrations/connectors"
+    failure_message: ClassVar
+
+    @property
+    def github_master_metadata_url(self):
+        return f"{self.GITHUB_URL_PREFIX_FOR_CONNECTORS}/{self.context.connector.technical_name}/{METADATA_FILE_NAME}"
+
+    @cached_property
+    def master_metadata(self) -> dict:
+        response = requests.get(self.github_master_metadata_url)
```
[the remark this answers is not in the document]
bnchrch: I think this was the right call :)

tools/ci_connector_ops/ci_connector_ops/pipelines/tests/common.py
```diff
@@ -4,13 +4,90 @@
...
+        return self.success_result
+
+
+class VersionFollowsSemverCheck(VersionCheck):
+
+    title = "Connector version semver check."
+    failure_message = f"The dockerImageTag in {METADATA_FILE_NAME} is not following semantic versioning."
+
+    def validate(self) -> StepResult:
+        if not len(str(self.current_connector_version).split(".")) == 3:
```
bnchrch, line 88: nit: What do you think about using the [semver](https://pypi.org/project/semver/) library instead? so we can have the same logic for parsing and validating?

bnchrch (commented):
Only real question I have is about using run_steps for our fail early logic :)

If thats not a good idea though lmk and we can merge

tools/ci_connector_ops/ci_connector_ops/pipelines/tests/__init__.py
alafanechere: I agree but in in this context I think `run_steps` is not immediately compatible as I wrapped steps executions inside function like `run_versions_checks`, `run_qa_checks` etc. 
I think we might eventually want to make these function Steps themselves (or `Pipeline`). And a `Pipeline` could nest multiple other pipelines execution.

alafanechere:
@bnchrch I did the following changes:
* I discarded the fail early logic because this check is likely to fail on a lot of connectors and I'd like other tests to run to monitor the health of running pipelines . We might reconsider it once the test pipelines are ready for production
* I use semver lib as suggested
* I added the logic to not require a version bump if only a specific set of files were modified.

I also realize that we'll probably soon want to migrate the[ QA checks](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/qa_checks.py) into specific pipeline steps because the pipelines have awareness on the modified files and it'll help add smarter static code analysis checks.

Finally, I opened https://github.com/airbytehq/airbyte/pull/25661 to encourage connector developers to bump the version in the metadata before we discard /test.

bnchrch (approved):
Thanks for the explanation! 

Lets merge!

PR #25588, opened 2023-04-26
Destination Bigquery: add AIRBYTE_ENTRYPOINT env var for kube process

edgao:
## What
When running in kube, airbyte overrides the entrypoint using AIRBYTE_ENTRYPOINT. Set that variable correctly.

Issue #25506, opened 2023-04-25
connectors-ci: validate the correctness of the images built with the new publish pipeline


alafanechere:
## Tell us about the problem you're trying to solve
If we want to rely on the new publish pipeline (publish on merge to master) we need to triple check the correctness of the images it build and publishes.

I would especially love to validate that:
- The packaging of normalization inside Java destination is correct
- Java sources behave correctly 
- Python and low code connector behave correctly 
- The image that are push are not significantly bigger than the one pushed with /publish 

Can we assume that passing nightly builds on GA connectors on the new test pipelines is enough to guarantee image correctness?

evantahler:
## Grooming:

Extra tests we do:
* Run syncs in multiple architectures
* Add a multi-arch CAT test
* E2E Tests

Worries:
* normalization + destinations combined 


Path Forward:
* Build pre-release versions of GA Sources (1 java, 1 python, 1 low-code) and GA destinations (1 java, 1 python) 
* Run syncs locally with existing GA publish versions and time them
* Run syncs locally with pre-release + dagger versions and time them
* Be sure to test both a strict encrypt and normal version of a destination
* Be sure to test the destinations with and without normalization

If times match and no bugs, we are good!

alafanechere:
I sucessfully tested source-facebook-marketing (GA Python connector) with pre-release **0.3.6-dev.e20fb9d994**
:
![Screen Shot 2023-05-06 at 11.47.53.png](https://images.zenhubusercontent.com/62a72f6a30e85f5981338b7c/1decbb72-8463-480b-b100-26a2df719a40)

alafanechere:
Successfully tested source postgres strict encrypt with [2.0.28-dev.26a60c5a08](https://hub.docker.com/layers/airbyte/source-postgres-strict-encrypt/2.0.28-dev.26a60c5a08/images/sha256-e2275caa305b58008f12011aaa1a05c8a4af0bf0b3bf393c846fda90b2ae7187?context=explore)
<img width="1357" alt="Screen Shot 2023-05-15 at 15 16 18" src="https://github.com/airbytehq/airbyte/assets/5551758/203b9494-5a68-4e1b-8477-38608955af63">

alafanechere:
Successfully test destination-postgres-strict-encrypt with `airbyte/destination-postgres-strict-encrypt:0.3.27-dev.be6dbae6df`
[...]

PR #25653, opened 2023-04-27
Replicate recent dockerfile changes into dagger

### LINKED ISSUE #25492: Port dockerfile changes into dagger
* replicate changes from https://github.com/airbytehq/airbyte/pull/25097
* into the relevant files (see https://github.com/airbytehq/airbyte/pull/25359 )

edgao:
closes https://github.com/airbytehq/airbyte/issues/25492; relates to https://github.com/airbytehq/airbyte/issues/25506

changes sourced from: https://github.com/airbytehq/airbyte/pull/25097/files#diff-ef927a1bde217f45489a1408ed303e9783225cc41482a42db0c13403b7f2b4be + https://github.com/airbytehq/airbyte/pull/25588/files

I ran local syncs with destinations bigquery and postgres, which both behaved as expected. (i.e. normalization ran successfully in bigquery, and postgres didn't try to do anything with normalizatoin)


jbfbell (approved):

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -565,6 +589,7 @@ def with_integration_base_java_and_normalization(context: PipelineContext, build
     ]
 
     dbt_adapter_package = DESTINATION_SPECIFIC_NORMALIZATION_ADAPTER_MAPPING.get(context.connector.technical_name, "dbt-bigquery==1.0.0")
```
edgao, line 591: @alafanechere should these two lines actually raise an exception? I.e. if the build isn't configured to support this connector, we would fail loudly

alafanechere: Yes, should all destination have a custom adapter, can we fallback to a default python package, or should adapter be optionally installed?

alafanechere (approved):
👍 The new logic looks good to me. I suggest to simplify the multiple mappings we have into a single one for readability.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
edgao: normalization is somehow installing a couple packages by default:
```
$ docker run -it --entrypoint=bash airbyte/normalization:dev
proot@ef8b6bf1d4ac:/airbyte# pip list | grep dbt
dbt-bigquery               1.0.0
dbt-core                   1.0.0
dbt-extractor              0.4.0
dbt-postgres               1.0.0
dbt-redshift               1.0.0
dbt-snowflake              1.0.0
```

so I'll just copy those values :shrug:

edgao: also switched everything to just use `[...]` instead of `.get`.

PR #25655, opened 2023-04-28
add max heartbeat

### LINKED ISSUE #25437: Update Metadata.yml to have the new max source heartbeat timeout field
## Overview
Update Metadata.yml to have the new max source heartbeat timeout field that Benoit/Move team recently added to the definitions

Note we will want this overrideable

bnchrch:
## What
1. Adds the new heart beat field
2. Adds a few other new fields
3. Adds missing icons
4. Updates and removes some metadata files
5. Updates out upload to work off of md5 instead of etags
6. updates the reports for slightly better diffs

closes #25437 

related to https://github.com/airbytehq/airbyte/pull/25661


airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
```diff
@@ -53,6 +53,16 @@ properties:
         format: uri
       githubIssueLabel:
         type: string
+      maxSecondsBetweenMessages:
+        description: Number of seconds allowed between 2 airbyte protocol messages. The source will timeout if this delay is reach
```
alafanechere, line 57:
```suggestion
        description: Maximum delay between 2 airbyte protocol messages, in second. The source will timeout if this delay is reached
```

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
```diff
@@ -53,6 +53,16 @@ properties:
         format: uri
       githubIssueLabel:
         type: string
+      maxSecondsBetweenMessages:
```
alafanechere, line 56: Should we have a default value for this field?

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
alafanechere, line 1: Do you think we should update the auto generated `metadata.yaml` files to set values for this field on created connectors?

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
```diff
@@ -53,6 +53,16 @@ properties:
...
+        description: Number of seconds allowed between 2 airbyte protocol messages. The source will timeout if this delay is reach
+        type: integer
+      releaseDate:
+        description: The date when this connector was first released, in yyyy-mm-dd format.
+        type: string
+        format: date
+      protocolVersion:
```
alafanechere, line 63: I think this field deserve an extra validation on the metadata validation side

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/dev.py
alafanechere, line 1: This changes are unrelated to the new metadata fields right?

alafanechere (approved):

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
bnchrch: Since its an optional override Im thinking no

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/dev.py
bnchrch: Yup! all part of improving the diff report

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
bnchrch: Thats a good call out! Im thinking a new issue for that. WDYT?

airbyte-integrations/connectors/source-github/metadata.yaml
```diff
@@ -6,6 +6,7 @@ data:
   connectorType: source
   definitionId: ef69ef6e-aa7f-4af1-a01d-ef775033524e
   dockerImageTag: 0.4.8
+  maxSecondsBetweenMessages: 5400
```
girarda, line 9: How did we decide on these values?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry.py
```diff
@@ -173,7 +175,8 @@ def persist_registry_to_json(
         OutputDataFrame: The registry directory manager.
     """
     registry_file_name = f"{registry_name}_registry"
-    registry_json = registry.json()
+    registry_json = registry.json(exclude_none=True)
```
girarda, line 178: I don't know what to do about this, but we keep tripping over this gotcha. Maybe we need a linting rule?

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
girarda: nit: can we document the default value? Either in the description or as an actual default is fine

airbyte-integrations/connectors/source-github/metadata.yaml
bnchrch: Good question @girarda !

So the we're currently migrating to use these metadata files. 

But they are not in use yet. 

The changes you see are all related to ensuring our new oss_registry.json files match the current oss_catalog.json files that are currently in use by the airbyte system. 

So these values are all just taken from the already deployed catalog file. 

You can see these in the code base in the _definitions.yaml files as well. 

I can share some links Monday when I'm near my laptop 🙂

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry.py
bnchrch: Ya its defiantely an annoying default by pandas. Though I dont think a linting rule is a good idea here as there will certainly be times we want null fields to be shown.

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
bnchrch: So this is where we are a bit caught between our two repos.

Our `airbyte` repo holds our connectors

Then our `platform` repo hold the system that runs these connectors

This PR is in relation to the metadata service which lives in `airbyte` and creates the registry that the `platform` consumes.

What that means in practise is 
1. we have duplicated types/fields between the metadata service in `airbyte` and the platform code in `platform-internal`. (At least in the short term)
2. We have the config of a connector separate from the implementation logic that consumes that config 

Some additional context here: https://docs.google.com/document/d/1jl122NkgH9rs6mr58UgyOVqD9J4d_656BcCEMlVVfbQ/edit#

For this field the default is implemented [in the platform repo here](https://github.com/airbytehq/airbyte-platform-internal/blob/master/oss/airbyte-featureflag/src/main/kotlin/FlagDefinitions.kt#L56) which means the default used in platform can change, and this system in `airbyte` would be none the wiser. 

So we can list a default value in the description, but I think given enough time it will become incorrect

PR #25661, opened 2023-04-28
qa-checks: check metadata version matches dockerfile version

alafanechere:
## What
This make QA checks rely on the version defined in the metadata file.
It means the changelog entry update check will use the version declared in the metadata, not in the dockerfile.
I also added a new check to validate the dockerfile version label matched the metadata version.

We must re-generate the metadata files for our connectors as this new QA check will probably fail for a lot of connectors.


bnchrch (approved):
Well done! and thank you!

PR #25663, opened 2023-04-28
Source S3: fix CAT tests

grubberr:
## What
After this:
https://aws.amazon.com/it/about-aws/whats-new/2022/12/amazon-s3-automatically-enable-block-public-access-disable-access-control-lists-buckets-april-2023/

Source S3 CAT tests started to fail


bazarnov (approved):
Great!

Issue #25646, opened 2023-04-27
DatetimeBasedCursor to support timezones other than UTC


maxi297:
## Tell us about the problem you're trying to solve
Given `2023-04-27T06:29:50+0700` as a DatetimeBasedCursor.start_datetime, the `start_datetime` sent to the API is `2023-04-27T06:29:50+0000`

the problem comes from [here](https://github.com/airbytehq/airbyte/blob/733b3e75420fa511edd7fdaff0919502e0ce68d4/airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py#L29). Basically, we blindly override the timezone while we should probably only assume UTC if the datetime is naive. 

## Describe the solution you’d like
If datetime is naive, assume it is UTC and set the timezone. Else, use the timezone defined in the datetime

## Additional context
Slack: https://airbytehq-team.slack.com/archives/C027KKE4BCZ/p1682599579162349

PR #25665, opened 2023-04-28
[ISSUE #25646] support parsing of non utc dates

maxi297:
## What
Addresses https://github.com/airbytehq/airbyte/issues/25646

## How
Only assume UTC datetime when the parsed datetime is naive 

## 🚨 User Impact 🚨
It safe to deploy without migration because this is actually fixing an issue that occurs today i.e. today if an API returns a datetime with a timezone, we strip this timezone and:
* have duplicate if the timezone is -X (for example `2023-04-27T06:29:50-01:00` is converted to `2023-04-27T06:29:50-00:00` which is actually one hour less and therefore the records between `2023-04-27T06:29:50-00:00` and `2023-04-27T07:29:50-00:00` will be fetched twice)
* have missing records if the timezone is +X

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py
@@ -16,7 +16,7 @@ class DatetimeParser:
     Instead of using the directive directly, we can use datetime.fromtimestamp and dt.timestamp()
     """
 
-    def parse(self, date: Union[str, int], format: str, timezone):
+    def parse(self, date: Union[str, int], format: str):
         # "%s" is a valid (but unreliable) directive for formatting, but not for parsing
         # It is defined as
         # The number of seconds since the Epoch, 1970-01-01 00:00:00+0000 (UTC). https://man7.org/linux/man-pages/man3/strptime.3.html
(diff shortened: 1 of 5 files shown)


airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py
```diff
@@ -25,8 +25,11 @@ def parse(self, date: Union[str, int], format: str, timezone):
         # See https://stackoverflow.com/a/4974930
         if format == "%s":
             return datetime.datetime.fromtimestamp(int(date), tz=timezone)
-        else:
-            return datetime.datetime.strptime(str(date), format).replace(tzinfo=timezone)
+
+        parsed_datetime = datetime.datetime.strptime(str(date), format)
+        if self._is_naive(parsed_datetime):
+            return parsed_datetime.replace(tzinfo=timezone)
+        return parsed_datetime
```
maxi297, line 32: I don't really like the interface of this method since timezone is not necessarily used. For example, if I call:
`DatetimeParser. parse("2023-04-27T06:29:50-01:00", "%Y-%m-%dT%H:%M:%S%z", datetime.timezone.utc)`, it feels like the returned datetime should have the timezone UTC but it won't. Therefore, I'll remove the parameter completely

maxi297: With the most recent change where we only set the timezone if the datetime is naive, I could call `DatetimeParser.parse("2023-04-27T06:29:50-01:00", "%Y-%m-%dT%H:%M:%S%z", datetime.timezone.utc)` and since the datetime provide is not naive, it would keep the timezone from the datetime. Hence, the parameter `datetime.timezone.utc` is confusing. Therefore, I removed the timezone from the parameters of the method. Anyway, I don't think it added any value since the value of this parameter was always UTC

brianjlai: ah okay so you were talking about the method signature change for the whole method. I think I was just confused because I thought this was in reference to the exact line of this comment. I agree getting rid of the parameter makes sense. thanks for clarifying!

brianjlai (approved):

airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py
maxi297: Yes, sorry for the confusion. I made changes after writing this comment which might have generated even more confusion

girarda (approved):

PR #25675, opened 2023-04-28
Pin bg color of row to white for dark mode

evantahler:
Fix this:
Screenshot 2023-04-28 at 10 09 56 AM

### DIFF
--- a/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/templates/base.html
+++ b/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/templates/base.html
@@ -56,6 +56,7 @@
 
       .styled-table tbody tr {
         border-bottom: 1px solid #dddddd;
+        background-color: #ffffff;
       }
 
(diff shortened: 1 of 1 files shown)


evantahler:
@bnchrch what do I need to do to rebuild the file... will it update on merge?

bnchrch (approved):
So @evantahler this will get run on the next change to the registry file after deploy.

Or after this deploys you can go here
https://airbyte-connectors.dagster.cloud/prod/locations/metadata_service_orchestrator/jobs/generate_registry_reports

and click "materialize all"

Now there is one catch!

There is an issue with Dagger (only when called from GHA) running the deploy orchestrator command right now which we have in the backlog.

So to deploy we have to call `airbyte-ci` locally.

Ill slack you the secret, and you can let me know if you want me to do it for you :0
```sh
DAGSTER_CLOUD_METADATA_API_TOKEN="REDACTED" airbyte-ci metadata deploy orchestrator
```

Issue #24720, opened 2023-03-30
Connector builder server: set slice descriptor field so slices are named properly in the UI


girarda:
## What
We should set the slice descriptor field on the API responses from the backend so that slices in the dropdown are named something more helpful than “Slice 0”, “Slice 1”, etc.

The slice descriptor current accepts two fields
- Start datetime
- list_item

When grouping the pages, we can get these fields by [parsing the slice log message](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py#L117)

Example slice message when using incremental syncs and a list partitioner:
`slice:{"item_id": "1", "start_time": "2020-01-01", "end_time": "2020-01-31"}`

## Implementation details
- This will require updating the `StreamReadSliceDescriptor` struct to also display the end datetime

## Open questions
- How to get the start/end datetime when the partition field is customized (defaults to `start_time` and `end_time`)

## Acceptance criteria
- [ ] The UI show the slice description with the start datetime if the stream has a DatetimebasedCursor
- [ ] The UI show the slice description with the end datetime if the stream has a DatetimebasedCursor
- [ ] The UI show the slice description with the item values if the stream has at least one partition router. The values can be concatenated with `|` 
- [ ] OpenAPI spec is updated to include a end datetime and to accept arbitrary key-value pairs

girarda:
grooming notes:
- both start and end time can be optional. they will always be either both set of both unset
- @girarda  follow up issue: need a design to display values when there are multiple partition routers
- FE behavior when using multiple partition routers: concatenate the key-value pairs. the string will be too long, but that's ok until we have a better design
- we will prevent users from customizing the field keys, so we can assume the default keys are used

maxi297:
We will change the proposed solution for this for a couple of reasons:
* We can't know which fields will express the start/end datetime. This was previously addressed in [a previous comment](https://github.com/airbytehq/airbyte/issues/24720#issuecomment-1513487114) by `we will prevent users from customizing the field keys, so we can assume the default keys are used`. However, this was not a perfect solution
* At the moment of parsing the slices to create the `StreamReadSliceDescriptor`, we don't know the datetime formats. To know about them, we would have to check the inner workings of the source which is flaky and increase maintenance costs

In the end, the information is just to be displayed to the user and the frontend does not use those fields. Therefore, it has been decided to return the slice object as an object changing the API to:
```
slice_descriptor:
  type: object
[...]

PR #25677, opened 2023-04-28
[ISSUE #24720] connector builder set slice descriptor

maxi297:
## What
Addresses the CDK part of https://github.com/airbytehq/airbyte/issues/24720

## How)
Sending a `Dict[str, Any]` (see the issue for an explanation as to why)

## 🚨 User Impact 🚨
This is technically a breaking change but in practice it was always null so it shouldn't break to connector-builder-server. Note that when updating the connector-builder-server to use the new CDK version, the frontend will need to be updated at the same time to avoid it to parse the slice descriptor as the previous `StreamReadSliceDescriptor` object


lmossman (commented):
I'm not the most familiar with this code but no issues stood out to me!

girarda (approved):

PR #25694, opened 2023-04-28
Build list of connectors in docusourus react component

### LINKED ISSUE #25678: Feedback on the Docs list of Connectors
In https://github.com/airbytehq/airbyte/pull/25650, we replaced the list of connectors on https://docs.airbyte.com/integrations/ with the registry list generated by the metadata pipeline.  That was a first pass.

Now, we should make a view of this data specifically for the docs site. This issue is to collect feedback on what that page should have:
* Ideally, we don't use an iFrame, and we load the remote data at build time from the metadata manifest
* We don't need actor ids
* The links should be easier to find
* It should work better in dark mode
* ...

Add additional ideas below as comments on this issue

evantahler:
Closes https://github.com/airbytehq/airbyte/issues/25678.

We load the list of connectors from the registry. We sort them by release stage and then alphabetically.  We only show relevant information. 

Building the list in-page with react vs an iFrame allows the styling and scrolling to function as expected

Screenshot 2023-04-28 at 2 13 54 PM
Screenshot 2023-04-28 at 2 13 52 PM

### DIFF
--- a/docusaurus/src/components/ConnectorRegistry.jsx
+++ b/docusaurus/src/components/ConnectorRegistry.jsx
@@ -73,7 +73,7 @@ export default function ConnectorRegistry({ type }) {
               .replace("source-", "")
               .replace("destination-", "");
             const iconLink = `${iconBase}/${connector.icon}`;
-            const docsLink = `/integrations/${type}s/${codeName}`;
+            const docsLink = `/integrations/${type}s/${codeName}`; // not using documentationUrl so we can have relative links
             const sourceLink = `${sourceBase}/${baseName}`;
             const bugsLink = `${bugsBase}:connectors/${type}/${codeName}`;
             const isCloud = cloudRegistry[type + "s"].find(
--- a/docusaurus/src/components/ConnectorRegistry.jsx
+++ b/docusaurus/src/components/ConnectorRegistry.jsx
@@ -1,19 +1,12 @@
 import React from "react";
 import { useEffect, useState } from "react";
 
-const oss_registry_url =
-  "https://connectors.airbyte.com/api/v0/catalog/oss_catalog.json";
-const cloud_registry_url =
-  "https://connectors.airbyte.com/api/v0/catalog/cloud_catalog.json";
-const iconBase =
-  "https://raw.githubusercontent.com/airbytehq/airbyte/master/airbyte-config-oss/init-oss/src/main/resources/icons";
-const iconStyle = { maxWidth: 25 };
-const sourceBase =
-  "https://github.com/airbytehq/airbyte/tree/master/airbyte-integrations/connectors";
-const bugsBase =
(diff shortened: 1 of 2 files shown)


docusaurus/src/components/ConnectorRegistry.jsx
```diff
@@ -0,0 +1,120 @@
+import React from "react";
+import { useEffect, useState } from "react";
+
+const oss_registry_url =
```
bnchrch, line 4: nit: we did have a json file to consume: https://connectors.airbyte.com/files/generated_reports/connector_registry_report.json

docusaurus/src/components/ConnectorRegistry.jsx
```diff
@@ -0,0 +1,120 @@
...
+const bugsBase =
+  "https://github.com/airbytehq/airbyte/issues?q=is:open+is:issue+label"; // :connectors/source/activecampaign
+
+async function fetchCatalog(url, setter) {
+  const response = await fetch(url);
+  const registry = await response.json();
+  setter(registry);
+  return registry;
+}
+
```
bnchrch, line 22: We can drop the return here
```suggestion
function fetchCatalog(url, setter) {
  const response = await fetch(url);
  const registry = await response.json();
  setter(registry);
}

```

docusaurus/src/components/ConnectorRegistry.jsx
```diff
@@ -0,0 +1,120 @@
...
+export default function ConnectorRegistry({ type }) {
+  const [ossRegistry, setOssRegistry] = useState([]);
+  const [cloudRegistry, setCloudRegistry] = useState([]);
+
+  useEffect(async () => {
+    await fetchCatalog(oss_registry_url, setOssRegistry);
+    await fetchCatalog(cloud_registry_url, setCloudRegistry);
+  }, []);
```
bnchrch, line 47: nit: we can drop the async
```suggestion
  useEffect(() => {
    fetchCatalog(oss_registry_url, setOssRegistry);
    fetchCatalog(cloud_registry_url, setCloudRegistry);
  }, []);
```

docusaurus/src/components/ConnectorRegistry.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+  if (ossRegistry.length === 0 || cloudRegistry.length === 0)
+    return <div>{`Loading ${type}s...`}</div>;
+
+  // makes the assumption that the OSS registry is a superset of the cloud registry
```
bnchrch, line 52: thanks for the comment!

bnchrch (approved):
Nice work, LGTM



And the nits can be ignored if you want. Done better than perfect!

evantahler:
automerge is too aggresive!

PR #25695, opened 2023-04-28
Logic update for document deployment and slash commands

wennergr:
## What
 - Change logic when docs are deployed
 - Change logic around slash commands

### DIFF
--- a/.github/workflows/deploy-docs-site.yml
+++ b/.github/workflows/deploy-docs-site.yml
@@ -8,16 +8,6 @@ on:
       - "docs/**"
       - "docusaurus/**"
 
-  pull_request:
-    types:
-      - closed
-      - opened
-      - reopened
-      - synchronize
-    paths:
-      - "docs/**"
-      - "docusaurus/**"
-
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-docs-site.yml
git-phu, line 1: Now that I'm taking a closer look at this file, I believe at current it is not actually possible for forks to run this workflow because of line 35, which requires a secret that forked repos will not have access to.

There is also some logic in lines 37-40 that I think will prevent an actual docs deploy from happening for PR builds.

So I think this workflow is actually fine in its current state?

I think there is something to be said about this not being straightforward to determine though, so perhaps we can do some refactoring to make this more obvious, like creating a separate workflow for PR checks instead of reusing this workflow that should be dedicated to publishing.

So I think we probably can merge this change as is and have a follow-up to run the docs PR checks as a new separate workflow.

git-phu (approved):
slash command change looks good

docs workflow change i don't think is strictly necessary but a good first step to refactoring this workflow so it seems worth doing anyway

git-phu (approved):

octavia-approvington:

octavia-approvington (approved):

PR #25696, opened 2023-04-29
low-code: Update docs on incremental syncs

girarda:
## What
* Clarify that records must be sorted in ascending order according to their cursor field
* Clarify that the cursor field cannot be nested
* Document which interpolation values are supported
* Remove an erroneous example usage of day_delta()
* Add a non-obvious example usage of day_delta()

## Recommended reading order
2. `docs/connector-development/connector-builder-ui/incremental-sync.md`

## 🚨 User Impact 🚨
Better documentation

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -561,8 +561,11 @@ definitions:
         enum: [DatetimeBasedCursor]
       cursor_field:
         title: Cursor Field
-        description: The location of the value on a record that will be used as a bookmark during sync.
+        description: The location of the value on a record that will be used as a bookmark during sync. To ensure no data loss, the API must return records in ascending order based on the cursor field. Nested fields are not supported, so the field must be at the top level of the record. You can use a combination of Add Field and Remove Field transformations to move the nested field to the top.
         type: string
+        interpolation_context:
+          - config
+          - stream_state
         examples:
           - "created_at"
           - "{{ config['record_cursor'] }}"
@@ -589,6 +592,7 @@ definitions:
         examples:
           - "2021-01-1T00:00:00Z"
           - "{{ now_utc() }}"
+          - "{{ day_delta(-1) }}"
       start_datetime:
         title: Start Datetime
         description: The datetime that determines the earliest record that should be synced.
@@ -2076,7 +2080,6 @@ interpolation:
       examples:
         - "{{ day_delta(25) }}"
         - "{{ day_delta(25, format='%Y-%m-%d') }}"
-        - "{{ config['start_time'] + day_delta(2) }}"
     - title: Duration
       description: Converts an ISO8601 duratioin to datetime.timedelta.
       arguments:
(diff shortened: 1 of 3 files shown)


docs/connector-development/connector-builder-ui/incremental-sync.md
```diff
@@ -7,8 +7,10 @@ This is especially important if there are a large number of records to sync and/
 Incremental syncs are usually implemented using a cursor value (like a timestamp) that delineates which data was pulled and which data is new. A very common cursor value is an `updated_at` timestamp. This cursor means that records whose `updated_at` value is less than or equal than that cursor value have been synced already, and that the next sync should only export records whose `updated_at` value is greater than the cursor value.
 
 To use incremental syncs, the API endpoint needs to fullfil the following requirements:
-* Records contain a date/time field that defines when this record was last updated (the "cursor field")
+* Records contain a top-level date/time field that defines when this record was last updated (the "cursor field")
+  * If the record's cursor field is nested, you can use an "Add Field" transformation to copy it to the top-level, and a Remove Field to remove it from the object. This will effectively move the field to the top-level of the record
```
lmossman, line 11: Do we have plans or an issue to allow defining the cursor field as a path so that nested fields can be accessed directly? The Add/Remove Field approach feels pretty hacky to me, but an okay workaround until we are able to support nested cursor fields natively

lmossman (approved):
Had a question about a future improvement, but otherwise this LGTM

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -561,8 +561,11 @@ definitions:
         enum: [DatetimeBasedCursor]
       cursor_field:
         title: Cursor Field
-        description: The location of the value on a record that will be used as a bookmark during sync.
+        description: The location of the value on a record that will be used as a bookmark during sync. To ensure no data loss, the API must return records in ascending order based on the cursor field. Nested fields are not supported, so the field must be at the top level of the record. You can use a combination of Add Field and Remove Field transformations to move the nested field to the top.
         type: string
+        interpolation_context:
+          - config
+          - stream_state
```
lmossman, line 568: One more question I just thought of: why doesn't this interpolation context have access to the records/response? It seems like if it did, the nested cursor field issue could be solved by accessing that variable in the interpolation

girarda: the cursor field is the path on the record. This won't work if the path is in the response, but not in the record.

We could update it to be a list to support nested fields, but normalization won't support nested cursor fields so there's not much value anyway.

(I'm not sure in what scenario the stream_state would be needed here - it would probably make more sense for the cursor_field to be static or at most, defined in the config)

docs/connector-development/connector-builder-ui/incremental-sync.md
girarda: Super hacky. We should design a better way to remap fields.

That being said, there's no plan to support nested cursor fields because [normalization will stop supporting them](https://docs.google.com/document/d/1T77e4V54g7-RiG02fm406iDa6PO-kkLJU0c6wJRyZhg/).

PR #25700, opened 2023-04-30
Source SalesForce: Remove pagination and query limits

artem1205:
## What
Resolving https://github.com/airbytehq/oncall/issues/1787

## How

Remove pagination and query limits; use only stream slices instead.
After these changes, connector successfully downloaded about 440k records.

## Recommended reading order
1. `x.java`
2. `y.python`

## 🚨 User Impact 🚨

no breaking changes

## Pre-merge Actions

### Airbyter

- Build is successful

</details>


artem1205:
I'd like to make a publish with `dev` tag, and pin this version for `Doxo.com` customer only.

artem1205:
/publish connector=connectors/source-salesforce pre-release=true

artem1205:
@maxi297 , yes, i've Reduced slice interval to 30 days, we can try this once more in test Doxo connection

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -683,11 +681,8 @@ def request_params(
         order_by_clause = ""
 
         if self.name not in UNSUPPORTED_FILTERING_STREAMS:
-            last_primary_key = (next_page_token or {}).get("primary_key", "")
-            if last_primary_key:
-                where_conditions.append(f"{self.primary_key} > '{last_primary_key}'")
             order_by_fields = ", ".join([self.cursor_field, self.primary_key] if self.primary_key else [self.cursor_field])
-            order_by_clause = f"ORDER BY {order_by_fields} ASC LIMIT {self.page_size}"
+            order_by_clause = f"ORDER BY {order_by_fields} ASC"
```
maxi297, line 685: I don't remember who but I've been told the SalesForce API had a limit and this was the reason we had this `limit` statement. Is that true? If yes, isn't it dangerous to remove this limit here?

artem1205: it should not be dangerous to remove limits. The idea of changes in this PR is to replace `LIMIT` and `pagination` with stream slicing, e.g.:
before:
``` SQL
SELECT a,b,c FROM table WHERE id>2 LIMIT 12000;
```
after
``` SQL
SELECT a,b,c FROM table WHERE cursor<2021-01-30 AND cursor>2021-01-01;
```

maxi297: How are we sure that `cursor<2021-01-30 AND cursor>2021-01-01` will return less than 12000 records?

artem1205:
/publish connector=connectors/source-salesforce pre-release=true

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
maxi297: Ok! I think my confusion came from the fact that in the declarative model, we stop looping when the next_token_page is null. In the case of Salesforce, this is not case since we always return a next_token_page if the stream as a primary_key and a name. In that case, we stop the iteration using [this condition](https://github.com/airbytehq/airbyte/blob/a3079cfdc598abfd42e3dbcae617053d77034f3f/airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py#L539). Size is defined at 15000, not 12000 though but we use it explicitly in the LIMIT statement so it should be fine.

maxi297 (approved):
Tested in production as AB testing and it worked so I'll approve

maxi297 (approved):
Tested in production as AB testing and it worked so I'll approve

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -432,7 +432,7 @@ def download_data(self, url: str, chunk_size: int = 1024) -> tuple[str, str]:
         # set filepath for binary data from response
         tmp_file = os.path.realpath(os.path.basename(url))
         with closing(self._send_http_request("GET", f"{url}/results", stream=True)) as response, open(tmp_file, "wb") as data_file:
-            response_encoding = response.apparent_encoding or response.encoding or self.encoding
+            response_encoding = response.encoding or self.encoding
```
sherifnada, line 435: can you clarify why you made this change?

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
```diff
@@ -579,7 +579,7 @@ def transform_empty_string_to_none(instance: Any, schema: Any):
 
 class IncrementalRestSalesforceStream(RestSalesforceStream, ABC):
     state_checkpoint_interval = 500
-    STREAM_SLICE_STEP = 120
+    STREAM_SLICE_STEP = 30
```
sherifnada, line 582: what is the impact of changing this? Do other salesforce syncs now take longer?

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
artem1205: response.`apparent_encoding` method tries to guess encoding by reading the whole response content. this caused OOM for large responses.

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
artem1205: By decreasing the `window` we make  responses smaller and thus iterate over them quicker.
According to screen it takes roughly the same time to sync (or even slightly less):
First sync was performed using `2.0.13-dev.b206a930`, next 2 using `2.0.9`
image

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
sherifnada: makes sense!

airbyte-integrations/connectors/source-salesforce/source_salesforce/streams.py
sherifnada: 👍🏼

sherifnada:
I was just asking questions, don't block on my review!

PR #23945, opened 2023-03-10
🐛 Source Freshservice: correct data types in stream schemas to match Freshservice API

aroder:
## What
Corrects some data types in the stream schemas, so that they match the source API's data types. Resolves #12243

## How
Modifications to the data types in the schema

## Recommended reading order
1. `x.java`
2. `y.python`
^--- I'm not sure what these mean

## 🚨 User Impact 🚨
Are there any breaking changes? What is the end result perceived by the user? If yes, please merge this PR with the 🚨🚨 emoji so changelog authors can further highlight this if needed.

No breaking changes

## Pre-merge Checklist
<details><summary><strong>Updating a connector</strong></summary>

### Community member or Airbyter

- [ ] Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests)) <-- following the instructions, I don't see an option "Allow edits and access to secrets by maintainers"
- [x] Secrets in the connector's spec are annotated with `airbyte_secret`
- [x] Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
image

- [ ] Code reviews completed <-- who should do these?
- [x] Documentation updated
    - [x] Connector's `README.md`
    - [x] Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - [x] Changelog updated in `docs/integrations/<source or destination>/<name>.md` including changelog. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog) <-- can't do this until merged, right?
- [x] PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests)

PR #25743, opened 2023-05-01
[non-forked] 🚨🚨Source Freshservice: correct data types in stream schemas to match Freshservice API

sajarin:
This PR includes the changes from #23945 and also includes changes to the schema files and acceptance_test_config.yml file to make tests pass.


sajarin:
This PR correct the following data types in the following schema files, which falls under a breaking change: 

> Ticket endpoint:
> Category field should be string instead of integer
> 
> Requester endpoint:
> work_phone_number should be string instead of integer
> mobile_phone_number should be string instead of integer
> 
> Agent endpoint
> work_phone_number should be string instead of integer
> mobile_phone_number should be string instead of integer

PR #25761, opened 2023-05-02
📝 Use correct CDC field names led by underscores

emmaling27:
## What
The docs have incorrect names for the CDC metadata columns. They are all prefixed by an underscore. This PR fixes that.

## How
Fixes docs.

## 🚨 User Impact 🚨
No breaking changes. This just fixes the docs to be accurate, which will help users who are developing new connectors and want to be compatible with CDC.

No checklist seemed applicable for docs? Let me know if there is anything else I need to do, though.


marcosmarxm (approved):
Thanks @emmaling27

PR #25792, opened 2023-05-04
✨ Source Amazon Ads: add availability strategy for basic streams

### LINKED ISSUE #22247: Source source-amazon-ads: implement AvailabilityStrategy
## What

## How
  in the connector.

and that's totally fine.

roman-yermilov-gl:
## What
*Add availablity strategy*


erohmensing (approved):
LGTM! Please follow the checklist for publishing/version bumping 👍🏻

lazebnyi (approved):
LGTM with Ella's comment.

PR #25820, opened 2023-05-04
pin the version of requests lib

girarda:
## What
* urllib3 has had a major version bump which is breaking the connector builder.
 * from 1.26.15 to 2.0.2


maxi297 (approved):
Shipit!

PR #25838, opened 2023-05-05
Source Github: deprecate "page size" input parameter

### LINKED ISSUE #10847: Source Github: deprecate "page size" input parameter
Like mentioned in the [UX handbook](https://docs.airbyte.io/connector-development/ux-handbook) this parameter is asking the user for input about an implementation detail. The user has no way of knowing what the right value for this parameter should be. 
Screen Shot 2022-03-04 at 12 15 06 AM

Remove that option and use a sane value by default.

### LINKED ISSUE #25025: Repository field on GitHub connector doesn't accept full URL
Need to improve setup success rate on GitHub connector
# problem
We watched fullstory sessions after noticing a high give up rate. In the repository field, most people input the full URL e.g “https://github.com/openai/gpt-3.git” and even though the error message said specifically what the correct format is, they didn’t know what to fix. We require just the repository name, not the full URL, so it fails to setup. 

 @lazebnyi how do you recommend we fix?

grubberr:
## What

Issue: https://github.com/airbytehq/airbyte/issues/10847

- Remove "page_size_for_large_streams" from `spec.json`, but still use it for current setups

## Pre-merge Actions

### Airbyter

- Build is successful

</details>


airbyte-integrations/connectors/source-github/source_github/source.py
```diff
@@ -133,6 +133,7 @@ def _get_org_repositories(config: Mapping[str, Any], authenticator: MultipleToke
             stream = RepositoryStats(
                 authenticator=authenticator,
                 repositories=unchecked_repos,
+                # This parameter is deprecated and in future will be used sane default, page_size: 10
```
lazebnyi, line 136: Can we add sone unit test for cover backward compatibility.

lazebnyi (changes requested):
LGTM! But need add unit tests.

airbyte-integrations/connectors/source-github/source_github/source.py
grubberr: In really I have only removed OPTIONAL param "page_size_for_large_streams" from `spec.json` and nothing more for now.

Hm I am stuck finding how good way to test this case.

lazebnyi: Attempt to read data using a custom page size or default and then compare the resulting quantity of records with the expected amount.

lazebnyi (approved):
Approved to unblock. But still need add unit tests.

PR #25842, opened 2023-05-05
:tada: Source Amplitude:  added missing attrs in events schema, enabled default availability strategy

### LINKED ISSUE #22248: Source source-amplitude: implement AvailabilityStrategy
## What

## How
  in the connector.

and that's totally fine.

### LINKED ISSUE #24480: Source Amplitude: Add undeclared columns to spec
## What

Amplitude currently fails the updated test. Its `acceptance-test-config.yml` was edited with the

**How this will help**:
  previously being sent. Doing this allows us to enable column selection for Amplitude!

## How

```
columns.txt:./source-amplitude.txt-85-The events stream has the following schema errors:
columns.txt:./source-amplitude.txt:86:Additional properties are not allowed ('$insert_key', 'data_type', 'plan', 'source_id', 'partner_id', 'global_user_properties' were unexpected)

```

5. Profit!

Definition of done: Amplitude passes CAT without declaring `fail_on_extra_columns: false`. If the
movement.

### LINKED ISSUE #24941: Amplitude Improvement Epic

midavadim:
added missing attrs in events schema, 
enabled default avilability strategy

## 🚨 User Impact 🚨

*For connector PRs, use this section to explain which type of semantic versioning bump occurs as a result of the changes. Refer to our [Semantic Versioning for Connectors](https://docs.airbyte.com/contributing-to-airbyte/#semantic-versioning-for-connectors) guidelines for more information. **Breaking changes to connectors must be documented by an Airbyte engineer (PR author, or reviewer for community PRs) by using the [Breaking Change Release Playbook](https://docs.google.com/document/d/1VYQggHbL_PN0dDDu7rCyzBLGRtX-R3cpwXaY8QxEgzw/edit).***

## Pre-merge Actions

- **Community member?** Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- Connector version is set to `0.0.1`
    - `Dockerfile` has version `0.0.1`
- Documentation updated
    - Connector's `README.md`
    - Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - `docs/integrations/<source or destination>/<name>.md` including changelog with an entry for the initial version. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
    - `docs/integrations/README.md`

### Airbyter

- Build is successful

</details>

### Airbyter

- Build is successful

</details>

- Issue acceptance criteria met
- PR name follows [PR naming conventions](https://docs.airbyte.com/contributing-to-airbyte/issues-and-pull-requests#pull-request-title-convention)
- If adding a new generator, add it to the [list of scaffold modules being tested](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connector-templates/generator/build.gradle#L41)
- The generator test modules (all connectors with `-scaffold` in their name) have been updated with the latest scaffold by running `./gradlew :airbyte-integrations:connector-templates:generator:testScaffoldTemplates` then checking in your changes
- Documentation which references the generator is updated as needed

</details>


artem1205 (changes requested):
availability strategy was removed, because for this stream API may return 404 (see [docs](https://www.docs.developers.amplitude.com/analytics/apis/export-api/#response) ) during read of first slice.

midavadim:
> availability strategy was removed, because for this stream API may return 404 (see [docs](https://www.docs.developers.amplitude.com/analytics/apis/export-api/#response) ) during read of first slice.

@artem1205 
404 error is caught at line 163 and stream just returns empty list of records.
So this error does not get to 'availability strategy' handler

artem1205:
Yep, but this means that availability strategy is useless here, isn't it?

midavadim:
> Yep, but this means that availability strategy is useless here, isn't it?

I suppose it can handle another types of errors, what do you think?

artem1205 (approved):

airbyte-integrations/connectors/source-amplitude/acceptance-test-config.yml
```diff
@@ -30,7 +30,6 @@ acceptance_tests:
           extra_fields: no
           exact_order: no
           extra_records: yes
-        fail_on_extra_columns: false
```
erohmensing, line 33: Woot 🥳

airbyte-integrations/connectors/source-amplitude/source_amplitude/streams.py
```diff
@@ -68,10 +67,6 @@ def url_base(self) -> str:
         subdomain = "analytics.eu." if self.data_region == "EU Residency Server" else ""
         return f"https://{subdomain}amplitude.com/api/"
 
-    @property
-    def availability_strategy(self) -> Optional["AvailabilityStrategy"]:
-        return None
```
erohmensing, line 73: Just double checked - I think @midavadim is right. The availability strategy's `get_first_record_for_slice` calls `stream.read_records` so the 404 error handling that returns empty records should work, and the AS should say "connected to stream, but got 0 records". So this should work in that case 👍🏻

erohmensing (commented):
Nice work! Could you please also add the missing column declarations to the cohorts stream? Here are all of the missing columns from the past week (they look pretty much teh same last month) 

image

This comes from the "please check the datadog dashboard" note [here](https://github.com/airbytehq/airbyte/issues/24480#issuecomment-1488885878).  I guess our test doesn't catch it because `cohorts` is declared as an empty stream.

maxi297 (approved):
LGTM given Ella's concern is addressed!

erohmensing:
@midavadim that's fine with me! If any of the types are wrong, that's on the documentation, not us. And we'll catch them in the schema validation errors. Thank you! ⭐

erohmensing (approved):

erohmensing:
Then I think we can safely close #24480 as well :)

Issue #25853, opened 2023-05-05
Connector Builder - Freeze Python version


maxi297:
There was an issue in production on May 4th because the CDK versions were not froze and the same code was redeployed. As part of the deployment, Docker images are rebuild and the transitives dependencies where updated. The specific case was `requests` release [version 2.3.0 on May 3rd](https://github.com/psf/requests/blob/main/HISTORY.md#2300-2023-05-03). For some reasons, this would the following error:
```
sh-4.2# /root/.pyenv/versions/3.9.11/bin/python /root/.pyenv/versions/3.9.11/lib/python3.9/site-packages/airbyte_cdk/connector_builder/main.py read --config /tmp/config415767259166977059.tmp --catalog /tmp/catalog7463836077949301517.tmp
Traceback (most recent call last):
  File "/root/.pyenv/versions/3.9.11/lib/python3.9/site-packages/airbyte_cdk/connector_builder/main.py", line 9, in <module>
    from airbyte_cdk.connector import BaseConnector
  File "/root/.pyenv/versions/3.9.11/lib/python3.9/site-packages/airbyte_cdk/__init__.py", line 6, in <module>
    from .entrypoint import AirbyteEntrypoint
  File "/root/.pyenv/versions/3.9.11/lib/python3.9/site-packages/airbyte_cdk/entrypoint.py", line 19, in <module>
    from airbyte_cdk.sources import Source
  File "/root/.pyenv/versions/3.9.11/lib/python3.9/site-packages/airbyte_cdk/sources/__init__.py", line 5, in <module>
    from .abstract_source import AbstractSource
  File "/root/.pyenv/versions/3.9.11/lib/python3.9/site-packages/airbyte_cdk/sources/abstract_source.py", line 28, in <module>
    from airbyte_cdk.sources.streams.http.http import HttpStream
  File "/root/.pyenv/versions/3.9.11/lib/python3.9/site-packages/airbyte_cdk/sources/streams/http/__init__.py", line 6, in <module>
    from .exceptions import UserDefinedBackoffException
  File "/root/.pyenv/versions/3.9.11/lib/python3.9/site-packages/airbyte_cdk/sources/streams/http/exceptions.py", line 8, in <module>
    import requests
  File "/root/.pyenv/versions/3.9.11/lib/python3.9/site-packages/requests/__init__.py", line 43, in <module>
    import urllib3
[...]

PR #25855, opened 2023-05-05
[ISSUE #25953] update frozen airbyte-connector-atelier-server python …

maxi297:
…dependencies

## What
Addresses https://github.com/airbytehq/airbyte/issues/25853

Related to https://github.com/airbytehq/airbyte-platform-internal/pull/6405

## How
Remove upper bound on requests. This was to solve the issue quickly in airbyte-connector-atelier-server but this is not an issue for the airbyte-cdk per se

When creating the PR in airbyte-platform, update and freeze dependencies for airbyte-connector-atelier-server

## 🚨 User Impact 🚨
Given the new process to freeze versions in airbyte-platform, this PR will automate the management of freezing versions which will save a lot of time for the devs since they won't have to pull the automated PR branch, run pip-compile and push the result


airbyte-cdk/python/setup.py
```diff
@@ -57,7 +57,7 @@
         "pydantic~=1.9.2",
         "python-dateutil",
         "PyYAML~=5.4",
-        "requests==2.29.0",
+        "requests",
```
maxi297, line 60: Note that once this will be released, we will be able to update the airbyte-connector-atelier-server by removing https://github.com/airbytehq/airbyte-platform-internal/pull/6405/files#diff-3520a0d4b40fe9a8f724644b23d8f2c34b412f9cdfbd1f3b0e5297dc40be41f1R1 and changing `openssl-devel` to `openssl11-devel`

.github/workflows/publish-cdk-command-manually.yml
```diff
@@ -294,11 +294,11 @@ jobs:
         run: |
           PREVIOUS_VERSION=$(cat oss/airbyte-connector-builder-resources/CDK_VERSION)
           sed -i "s/${PREVIOUS_VERSION}/${{needs.bump-version.outputs.new_cdk_version}}/g" oss/airbyte-connector-atelier-server/Dockerfile
+          sed -i "s/airbyte-cdk==${PREVIOUS_VERSION}/airbyte-cdk==${{needs.bump-version.outputs.new_cdk_version}}/g" oss/airbyte-connector-atelier-server/requirements.in
           echo ${{needs.bump-version.outputs.new_cdk_version}} > oss/airbyte-connector-builder-resources/CDK_VERSION
-          cd oss/airbyte-connector-builder-server
+          cd oss/airbyte-connector-atelier-server
```
brianjlai, line 299: Just to confirm, this was the issue where we were testing the old python image and so the issue with the java server slipped unnoticed?

For the moment I agree we should fix this, but before we merge this, let's be extra careful to check that it hasn't been renamed because we've got a handful of workstreams swapping the builder for atelier

brianjlai (approved):
nothing to block, just want to make sure I'm on the same page on one thing.

Side note I do think its a little bit weird that we have a GH workflow that depends on files in internal `/oss`, it feels like the structure of the private repo is bleeding into public workflows. Out of scope, but feels like a bit of an antipattern

maxi297:
> nothing to block, just want to make sure I'm on the same page on one thing.
> 
> Side note I do think its a little bit weird that we have a GH workflow that depends on files in internal `/oss`, it feels like the structure of the private repo is bleeding into public workflows. Out of scope, but feels like a bit of an antipattern

The pattern hasn't changed the fact that we rely on a file in the OSS folder. It has only change the file since instead of freezing versions in airbyte-connector-builder-server, we freeze them in airbyte-connector-atelier-server. The workflow is a patch because our value stream is splitted between two repositories

.github/workflows/publish-cdk-command-manually.yml
maxi297: This is a very good point. I'll be extra careful when I'll merge master into this branch. Thanks for reminding me!

brianjlai:
> The pattern hasn't changed the fact that we rely on a file in the OSS folder

Right but from our public `airbyte` repo workflow we rely on checking out a private repo internal which feels weird cuz it's contingent on having proper access. It's more just an oddity, I don't have a solution in mind 😅

maxi297:
> Right but from our public airbyte repo workflow we rely on checking out a private repo internal which feels weird cuz it's contingent on having proper access. It's more just an oddity, I don't have a solution in mind 😅

Yes... the fun of having our value stream in different repositories!

PR #25856, opened 2023-05-05
Exclude scaffold during validate all

bnchrch:
## What
Exclude scaffold files during validate all

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
@@ -242,7 +242,11 @@ def get_modified_metadata_files(modified_files: Set[Union[str, Path]]) -> Set[Pa
 
 
 def get_all_metadata_files() -> Set[Path]:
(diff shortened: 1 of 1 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -242,7 +242,11 @@ def get_modified_metadata_files(modified_files: Set[Union[str, Path]]) -> Set[Pa
 
 
 def get_all_metadata_files() -> Set[Path]:
-    return {Path(metadata_file) for metadata_file in glob("airbyte-integrations/connectors/**/metadata.yaml", recursive=True)}
+    return {
+        Path(metadata_file)
+        for metadata_file in glob("airbyte-integrations/connectors/**/metadata.yaml", recursive=True)
+        if "-scaffold-" not in metadata_file
+    }
```
alafanechere, line 249: nit: 
```suggestion
    return {
        metadata_path
        for metadata_path in Path("airbyte-integrations/connectors/").rglob("**/metadata.yaml")
        if "-scaffold-" not in metadata_path.parent.name
    }
```

alafanechere (approved):

PR #25858, opened 2023-05-05
connectors-ci: send slack message and set execute timeout

### LINKED ISSUE #25844: connectors-ci: send slack messages for publish pipelines
When a publish pipeline is triggered on merge to master we want to receive slack messages in a channel:
* When a connector specific publish pipeline started
* When it fails (tag @dev-connector-ops)
* When it succeeds

### LINKED ISSUE #25845: connectors-ci: publish pipeline should timeout after 1H
We want dagger operations to fail after 1H timeout on publish pipelines

alafanechere:
## What
Closes #25844
Closes #25845

We want to send update to slack channels when a pipeline context starts and stops (success or failure).
We also want publish pipeline to fail if a dagger operation takes more than 1H.

## How
* Send message to #publish-on-merge-updates when entering or exiting 
* Set an execute timeout to 1H
Screen Shot 2023-05-05 at 16 33 58

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
@@ -77,6 +77,12 @@ def validate_environment(is_local: bool, use_remote_secrets: bool):
 )
 @click.option("--modified/--not-modified", help="Only test modified connectors in the current branch.", default=False, type=bool)
 @click.option("--concurrency", help="Number of connector tests pipeline to run in parallel.", default=5, type=int)
+@click.option(
+    "--execute-timeout",
+    help="The maximum time in seconds for the execution of a Dagger request before an ExecuteTimeoutError is raised. Passing None results in waiting forever.",
+    default=None,
+    type=int,
+)
 @click.pass_context
 def connectors(
     ctx: click.Context,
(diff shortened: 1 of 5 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
```diff
@@ -77,6 +77,12 @@ def validate_environment(is_local: bool, use_remote_secrets: bool):
 )
 @click.option("--modified/--not-modified", help="Only test modified connectors in the current branch.", default=False, type=bool)
 @click.option("--concurrency", help="Number of connector tests pipeline to run in parallel.", default=5, type=int)
+@click.option(
+    "--execute-timeout",
+    help="The maximum time in seconds for the execution of a Dagger request before an ExecuteTimeoutError is raised. Passing None results in waiting forever.",
+    default=None,
```
evantahler, line 83: Is a 1h default not good for all types of tasks (except maybe the nightly tests?)

evantahler (approved):
Nice!

Nit: In the screenshot the title of the message is `<none|publish>`... what's `none`?

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
alafanechere: I wanted to make this specific to publish as until now I haven't been setting timeouts. It should indeed be good for all types of task but to reduce moving parts I prefer to not set the timeout globally.

alafanechere:
> Nit: In the screenshot the title of the message is <none|publish>... what's none?

none should be the link to the GHA workflow. As I ran the command locally for testing the GHA workflow URL is none.

PR #25859, opened 2023-05-05
Allow sessionvariables jdbc url param in source-mysql

rodireich:
## What
source mysql can accept a jdbc url param looking like `sessionVariables=MAX_EXECUTION_TIME=123456666`

## How
Allow adding a sessionVariables param in jdbc url parsing.
In order to reduce blast radius this only applies to mysql and mysql strict encrypt.
This is an adapted code change of #25386


marcosmarxm (approved):
Thanks @rodireich

airbyte-integrations/connectors/source-mysql/src/main/java/io/airbyte/integrations/source/mysql/MySqlSource.java
```diff
@@ -382,6 +387,40 @@ protected static String toSslJdbcParamInternal(final SslMode sslMode) {
...
+    if (!jdbcPropertiesString.isBlank()) {
+      final String[] keyValuePairs = jdbcPropertiesString.split(delimiter);
+      for (final String kv : keyValuePairs) {
+        final String[] split = kv.split("=");
+        if (split.length == 2) {
+          parameters.put(split[0], split[1]);
+        } else if (split.length == 3 && kv.contains("sessionVariables")) {
+          parameters.put(split[0], split[1] + "=" + split[2]);
+        } else {
+          throw new IllegalArgumentException(
```
akashkulk, line 415: Maybe mark this is a `ConfigErrorException`?

airbyte-integrations/connectors/source-mysql/src/test/java/io/airbyte/integrations/source/mysql/MySqlSourceTests.java
```diff
@@ -197,4 +198,13 @@ CREATE VIEW test_view_null_cursor(id) as
 
   }
 
+
+  @Test
+  void testParseJdbcParameters() {
```
akashkulk, line 203: Maybe in addition to this, we can add a new test that given an existing source config containing a jdbc url param looking like `sessionVariables=MAX_EXECUTION_TIME=123456666`, that we are successfully able to connect to the database? So, it will be a little more e2e

As someone reading the unit tests, I'm not sure how `parseJdbcParameters` fits into the larger picture

akashkulk (commented):
Quick question : I understand this is mainly to reduce the blast radius of the change. But, is this something that is ONLY specific to mysql and not desired in other connectors? 

I guess my thinking is that this seems like generally useful code that  will enable all JDBC users to have params in the form of `sessionVariables=MAX_EXECUTION_TIME=123456666`

What we actually publish can be something else altogether.

rodireich:
I didn't see anything equivalent in postgres jdbc url parameters.
This isn't even a legal url as far as I can tell as this `=` was supposed to be url encoded `%3D`.
mysql however doesn't accept an encoded equals to sign.
That's why I was struggling to only put this in mysql. Not a really good change but necessary for somebody who needs to set a session variable.

airbyte-integrations/connectors/source-mysql/src/main/java/io/airbyte/integrations/source/mysql/MySqlSource.java
rodireich: I can change that. 
Should I also update the same in the JdbcUtils class from which it was copied?

akashkulk: Yes - from my understanding this indicates an error in how the user entered their JDBC params right? If so, then it should be a config error both here and the original code

akashkulk:
> I didn't see anything equivalent in postgres jdbc url parameters. This isn't even a legal url as far as I can tell as this `=` was supposed to be url encoded `%3D`. mysql however doesn't accept an encoded equals to sign. That's why I was struggling to only put this in mysql. Not a really good change but necessary for somebody who needs to set a session variable.

Out of curiosity, how many people are facing this issue? Longer term, I think we'd not want to expose JDBC param setting. Another option here is to just expose the `MAX_EXECUTION_TIME` as a setting in the UX

I think given that this is currently scoped to Mysql, this should be fine - but I really think we should file a ticket to revisit how we're setting JDBC args in the future. For example : 
- How many users are setting them? 
- Are there only a couple of distinct parameters that only set? If so, perhaps only those can be exposed in the UX?

I think this will make some of the existing code confusing. it also seems like the user found a workaround by setting a top-level JDBC param instead of a `sessionVariable`

I'd prefer to shelf this issue for now before we have a good understanding of how often this is used. WDYT?

rodireich:
This came form one user in #25386.
This user did find a workaround and I'm not very happy with this code change either but if somebody is stuck and need to set a session variable they may have no other recourse.

Any of the following variables are valid so there may be other cases - current or future - that can benefit from us supporting it:
https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html

airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java
```diff
@@ -410,14 +411,18 @@ protected long getActualCursorRecordCount(final Connection connection,
 
   @Override
   public JdbcDatabase createDatabase(final JsonNode sourceConfig) throws SQLException {
+    return createDatabase(sourceConfig, JdbcDataSourceUtils::getConnectionProperties);
```
akashkulk, line 414: Let's just override the entire createDatabase method in MySqlSource to support this use case, with a comment on why it needs overriding

akashkulk (commented):
Small nit, looks good otherwise

airbyte-integrations/connectors/source-mysql/src/test/java/io/airbyte/integrations/source/mysql/MySqlSourceTests.java
rodireich: Added

airbyte-integrations/connectors/source-mysql/src/main/java/io/airbyte/integrations/source/mysql/MySqlSource.java
rodireich: Done

akashkulk (approved):

PR #25864, opened 2023-05-06
connectors-ci: handle strict encrypt connectors

### LINKED ISSUE #25846: connectors-ci: strict encrypt connector should be testable and publishable with the CLI
Strict encrypt connectors are not yet supported by the CLI (because the list of existing connectors is taken from the OSS registry). 
We should change this.

alafanechere:
## What
Closes #25846 

The strict encrypt connectors were not testable with the CLI because we used the OSS registry to get the list of available connectors. 

## How
* Create the list of available connectors by searching all the folder having a `metadata.yaml` file
* Tweak the gradle command run to include non strict encrypt connector project when building / testing strict encrypt connectors.


tools/ci_connector_ops/ci_connector_ops/pipelines/bases.py
```diff
@@ -430,10 +430,16 @@ def build_include(self) -> List[str]:
         Returns:
             List[str]: List of directories or files to be mounted to the container to run a Java connector Gradle task.
         """
+        additional_inclusion = []
```
bnchrch, line 433: Nit: Why an array and not a single value?

tools/ci_connector_ops/ci_connector_ops/pipelines/bases.py
```diff
@@ -430,10 +430,16 @@ def build_include(self) -> List[str]:
...
+        if self.context.connector.technical_name.endswith("-strict-encrypt"):
```
bnchrch, line 434: Idea: it would be great if we had a more explicit mapping
```py
related_connectors = {
  "source-file-secure": ["source-file"],
  "source-postgres-strict-encrypt": ["source-postgres"]
  # ...
,}

bnchrch (approved):
Couple of nits, otherwise lgtm!

PR #25879, opened 2023-05-08
Connector builder documentation: Small adjustments

flash1293:
I asked for feedback about our docs, doing some small fixes here:
* Remove "base64" as a keyword to look for for basic auth - while it's often related to basic auth, it can also mean something else and the "basic" is a very official term to look for
* Removed a link to an issue for single use refresh tokens - this link won't help users if they have to implement this themselves and might only be frustrating
* Clarify some wordings
* Remove a paragraph with duplicate information (the same stuff is already stated on the same page)


brianjlai (approved):

lmossman (approved):
LGTM

Issue #20466, opened 2022-12-14
Source postgres Number(38,0) is translated to a destination Double


rodireich:
## Environment
- **Airbyte version**:  example is 0.22.0-alpha
- **OS Version / Instance**: example macOS, Windows 7/10, Ubuntu 18.04, GCP n2. , AWS EC2
- **Deployment**: example are Docker or Kubernetes deploy env
- **Source Connector and version**: (if applicable example Salesforce 0.2.3) 
- **Destination Connector and version**: (if applicable example Postgres 0.3.3) 
- **Step where error happened**: Deploy / Sync job / Setup new connection / Update connector / Upgrade Airbyte

## Current Behavior
On a Postgres <> Postgres replication
Reading from source a column of `NUMBER(38, 0)` is translated on destination to a `DOUBLE PRECISION` type. 
This is an extra information in type as the `0` in the number is the scale (=places after decimal point),

## Expected Behavior
The most correct way to create the type is an Integer with the same range.
a Snowflake <> Postgres replication will create a destination BigInt, which is correct.

akashkulk:
Do you happen to know a good way to get precision out of a Column ResultSet? It's not exposed in the `DatabaseMetaData` API : https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-

Strangely enough - that capability exists in the `getProcedureColumns` method. https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getProcedureColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-

bleonard:
Similar to snowflake #19776

bleonard:
Probably currently just seeing. DECIMAL -> FLoat
but we could check the precision == 0 to know it's an INT.

akashkulk:
First steps : Check what airbyte_type a NUMERIC(38,0) is currently converted to. 

1. Create a Postgres table w/ NUMERIC(38,0) type
2. Trace the process of how this is converted to an airbyte type (which is the type we send over the wire) : https://docs.airbyte.com/understanding-airbyte/supported-data-types
3. Check what type this ends up being discovered us. Check the catalog that is returned by discover()
4. Explore this via unit tests. The class that contains the logic for mapping DB types to airbyte types is `PostgresSourceOperations.getAirbyteType`

PR #25898, opened 2023-05-08
Source Postgres: Number(38,0) is incorrectly translated to a destination Double instead of BigInt

nguyenaiden:
## What
#20466 
### Current Behavior
On a Postgres <> Postgres replication
Reading from source a column of NUMBER(38, 0) is translated on destination to a DOUBLE PRECISION type.
This is an extra information in type as the 0 in the number is the scale (places after decimal point)
### Expected Behavior
The most correct way to create the type is an Integer with the same range.
e.g: a Snowflake <> Postgres replication will create a destination BigInt, which is correct.
## How
- Updated `copyToJsonField` and `getDatabaseFieldType` in `PostgresSourceOperations.java`  as well as `PostgresConverter` to handle the variations of the `NUMERIC` datatype:
1. `NUMERIC` - Has unspecified Precision but preserve all decimal scale. e.g: type: NUMERIC(), data: 10.23 -> 10.23
2. `NUMERIC(somePrecision)` - Has Precision but 0 decimal scale. e.g: type: NUMERIC(20), data: 10.25 -> 10
3. `NUMERIC(somePrecision, someScale)` - Both Precision and Scale are defined.  e.g: type: NUMERIC(10,4), data: 10.2523 -> 10.2523;  15.23451 -> 15.2345

- Updated `PostgresConverter.registerArray`  to map `_NUMERIC` data type to use the int64 schema when there is no decimal scale instead of `float64`
- Expanded unit tests in `AbstractPostgresSourceDataTypeTest` and `PostgresSourceOperationsTest` to account for specified and unspecified decimal scale.

## Reading Order
`PostgresSourceOperations`
`PostgresSourceOperationsTests`
`PostgresConverter`
`AbstractPostgresSourceDatatypeTest`


airbyte-integrations/connectors/source-mysql/src/test/java/io/airbyte/integrations/source/mysql/MySqlSourceOperationsTest.java
```diff
@@ -302,5 +302,4 @@ protected void executeQuery(final String query) {
       throw new RuntimeException(e);
     }
   }
-
```
prateekmukhedkar, line 305: nit: is this auto-formatting causing mischief?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
```diff
@@ -398,9 +407,16 @@ public PostgresType getDatabaseFieldType(final JsonNode field) {
         // It should not be converted to base64 binary string. So it is represented as JDBC VARCHAR.
         // https://www.postgresql.org/docs/14/datatype-binary.html
         case "bytea" -> PostgresType.VARCHAR;
+        case "numeric" -> {
+          if (field.get(INTERNAL_DECIMAL_DIGITS) != null && field.get(INTERNAL_DECIMAL_DIGITS).asInt() == 0) {
+            yield BIGINT;
+          } else {
+            yield defaultType;
```
prateekmukhedkar, line 414: shouldn't this be `PostgresType.NUMERIC`?

prateekmukhedkar, line 414: shouldn't this be PostgresType.NUMERIC?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
```diff
@@ -398,9 +407,16 @@ public PostgresType getDatabaseFieldType(final JsonNode field) {
...
+          }
+        }
         case TIMESTAMPTZ -> PostgresType.TIMESTAMP_WITH_TIMEZONE;
         case TIMETZ -> PostgresType.TIME_WITH_TIMEZONE;
-        default -> PostgresType.valueOf(field.get(INTERNAL_COLUMN_TYPE).asInt(), POSTGRES_TYPE_DICT);
+        default -> defaultType;
```
prateekmukhedkar, line 419: one unintended side effect of moving `PostgresType.valueOf(field.get(INTERNAL_COLUMN_TYPE).asInt(), POSTGRES_TYPE_DICT);` further up this method is that it will be executed for all cases and not just the default. Do we know if all `INTERNAL_COLUMN_TYPE` are mapped to `POSTGRES_TYPE_DICT`? As in will is there a chance we run into a `KeyNotFoundException`?

prateekmukhedkar: remembered KeyNotFoundException is .NET C# land :) but the question still remains for 1:1 mapping of INTERNAL_COLUMN_TYPE to POSTGRES_TYPE_DICT and what happens if `defaultType` comes to be null?

airbyte-integrations/connectors/source-mysql/src/test/java/io/airbyte/integrations/source/mysql/MySqlSourceOperationsTest.java
nguyenaiden: This one was me. I'll update it.

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
nguyenaiden: `numeric` is manually handled in the defaultType. This edge case was added specifically for the no decimals issuel; otherwise, the defaultType will handle the mapping.

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
nguyenaiden: Ah, good call. Will move it back to the original level and investigate if we could potentially running into a missing key problem.

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
```diff
@@ -370,8 +379,8 @@ public PostgresType getDatabaseFieldType(final JsonNode field) {
     try {
       final String typeName = field.get(INTERNAL_COLUMN_TYPE_NAME).asText().toLowerCase();
       // Postgres boolean is mapped to JDBCType.BIT, but should be BOOLEAN
+      final PostgresType defaultType = PostgresType.valueOf(field.get(INTERNAL_COLUMN_TYPE).asInt(), POSTGRES_TYPE_DICT);
       return switch (typeName) {
```
rodireich, line 383: If doable, An array of numerics should also be checked and return an array of integers in case scale is 0.
that's the case of `_numeric`

prateekmukhedkar (commented):
@nguyenaiden  can you run `PostgresSourceDatatypeTest` tests with this PR? I am getting a test failure with the message 
`Returned value '1234567890' from stream test_1_decimal is not in the expected list: [1.2345678901234567E9] ==> expected: <true> but was: <false>`

I think it this for the test defined here -something in the new code is breaking them? https://github.com/airbytehq/airbyte/pull/25898/files#diff-b19dae966dbd5c3edecfbcbc290661374cdc7bf85144a6d46d46f349c6fb1205L210

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
nguyenaiden: @rodireich got it. Will make the change

akashkulk:
Just wanted to check - does this work on an e2e sync w/normalization?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
```diff
@@ -398,9 +407,16 @@ public PostgresType getDatabaseFieldType(final JsonNode field) {
         // It should not be converted to base64 binary string. So it is represented as JDBC VARCHAR.
         // https://www.postgresql.org/docs/14/datatype-binary.html
         case "bytea" -> PostgresType.VARCHAR;
+        case "numeric" -> {
+          if (field.get(INTERNAL_DECIMAL_DIGITS) != null && field.get(INTERNAL_DECIMAL_DIGITS).asInt() == 0) {
+            yield BIGINT;
```
akashkulk, line 412: nit : return

airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/PostgresSourceOperationsTest.java
```diff
@@ -4,12 +4,174 @@
 
 package io.airbyte.integrations.source.postgres;
 
-import static org.junit.jupiter.api.Assertions.*;
+import static org.hamcrest.MatcherAssert.assertThat;
```
akashkulk, line 7: Thanks for adding these tests!

edgao:
it's probably worth doing tests on snowflake + bigquery (i.e. the two GA destinations with normalization). We're sticking to numeric types so I'm not super worried though.

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresSourceOperations.java
nguyenaiden: It has to be yield here since it's technically not returning within the switch statement.

airbyte-integrations/connectors/source-postgres/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/AbstractPostgresSourceDatatypeTest.java
```diff
@@ -293,11 +293,36 @@ protected void initTests() {
...
+
     // Blocked by https://github.com/airbytehq/airbyte/issues/8902
     for (final String type : Set.of("numeric", "decimal")) {
       addDataTypeTestData(
           TestDataHolder.builder()
               .sourceType(type)
+              .fullSourceDataType("NUMERIC(20,7)")
               .airbyteType(JsonSchemaType.NUMBER)
               .addInsertValues(
                   "'123'", "null", "'1234567890.1234567'")
```
nguyenaiden, line 328: @prateekmukhedkar, for this test case in its current form, the `numeric` source type will have a 0 decimal scale since the fullSourceDataType wasn't specified. With my change, `1234567890.1234567` will be truncated to `1234567890` I've updated this test case to `NUMERIC(20,7)` to accommodate the chosen value.

prateekmukhedkar (approved):

airbyte-integrations/connectors/source-postgres/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/AbstractPostgresSourceDatatypeTest.java
```diff
@@ -726,29 +751,33 @@ private void addArraysTestData() {
...
-            .airbyteType(JsonSchemaType.builder(JsonSchemaPrimitive.ARRAY)
-                .withItems(JsonSchemaType.builder(JsonSchemaPrimitive.NUMBER)
-                    .build())
-                .build())
-            .addInsertValues("'{131070.23,231072.476596593}'")
-            .addExpectedValues("[131070.23,231072.476596593]")
-            .build());
+    for (final String type : Set.of("numeric", "decimal")) {
+      // This verifies NUMERIC/DECIMAL[] types to have
+      // default precision of (38, 0) if not specified
```
subodh1810, line 756: whats the value thats being written in the posgtres source?

subodh1810 (changes requested):
So there is an issue/bug with the PR. For numeric array we are dropping the decimal points while postgres is retaining them. We need to add those tests back and fix it.

nguyenaiden:
@subodh1810, Numeric arrays in CDC mode are still running into the same [deserialization problem](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/postgres/PostgresConverter.java#L90). E.g: If a numeric array is `[34.224, 23.9901, 55]`, it will come back as `[34.224, 23.9901, 55.0]`, but I have manually verified that the value is stored as `55` in Postgres. Is this something we want to handle on our side or the destinations side?

Please tag anyone from Destinations that you see fit.

airbyte-integrations/connectors/source-postgres/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/AbstractPostgresSourceDatatypeTest.java
nguyenaiden: CDC now preserves the decimal places if no Precision **_and_** Decimal are specified.

airbyte-integrations/connectors/source-postgres/src/test-integration/java/io/airbyte/integrations/io/airbyte/integration_tests/sources/AbstractPostgresSourceDatatypeTest.java
```diff
@@ -726,29 +774,51 @@ private void addArraysTestData() {
...
-            .build());
+      addDataTypeTestData(
+          TestDataHolder.builder()
+              .sourceType(String.format("%s_array", type))
+              .fullSourceDataType(String.format("%s(30,2)[]", type.toUpperCase()))
+              .airbyteType(JsonSchemaType.builder(JsonSchemaPrimitive.ARRAY)
+                  .withItems(JsonSchemaType.builder(JsonSchemaPrimitive.NUMBER)
+                      .build())
+                  .build())
+              .addInsertValues("'{131070.23,231072.476596593}'")
```
subodh1810, line 818: Can you add a comment here that Postgres stores the rounded off value out here

subodh1810:
> @subodh1810, Numeric arrays in CDC mode are still running into the same [deserialization problem](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/postgres/PostgresConverter.java#L90). E.g: If a numeric array is `[34.224, 23.9901, 55]`, it will come back as `[34.224, 23.9901, 55.0]`, but I have manually verified that the value is stored as `55` in Postgres. Is this something we want to handle on our side or the destinations side?
> 
> Please tag anyone from Destinations that you see fit.

I think its fine. Its not really a huge bug per say

subodh1810 (approved):

subodh1810 (approved):

airbyte-config-oss/init-oss/src/main/resources/seed/oss_registry.json
akashkulk, line 1: Weird, there are no diffs here. Can you just revert this?

akashkulk (approved):

airbyte-config-oss/init-oss/src/main/resources/seed/oss_registry.json
nguyenaiden: Yep I'll revert this. I have no idea how this got added.

PR #25909, opened 2023-05-09
connectors-ci: always run metadata validation and report success if no validation happened

alafanechere:
## What
We want to make the metadata validation a required Github status check when a PR targets `master`.
Required checks can't be dynamically declared, so we have to run the metadata validation pipelines on all the PRs, even those not editing the `metadata.yaml` files.

## How
* Change the GHA workflow to perform metadata validation on all PRs.
* Emit a success status check when no metadata files were modified.

## 🚨 User Impact 🚨
@bnchrch the `Validate metadata.yaml files` check should be declared as mandatory. Not the `Connectors' metadata.yaml files validation` why:

* `Connectors' metadata.yaml files validation`: it's the GHA workflow run. IMO it should always succeeds unless an unhandled exception happenned.
* `Validate metadata.yaml files`: It's the actual status update sent by Dagger pipelines.

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/metadata.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/metadata.py
@@ -19,6 +19,7 @@ from rich.logging import RichHandler
 logging.basicConfig(level=logging.INFO, format="%(name)s: %(message)s", datefmt="[%X]", handlers=[RichHandler(rich_tracebacks=True)])
 logger = logging.getLogger(__name__)
 
+
 # MAIN GROUP
 
 
@@ -37,9 +38,6 @@ def metadata(ctx: click.Context):
 def validate(ctx: click.Context, modified_only: bool) -> bool:
     if modified_only:
         metadata_to_validate = get_modified_metadata_files(ctx.obj["modified_files"])
-        if not metadata_to_validate:
-            click.secho("No modified metadata found. Skipping metadata validation.")
-            return True
     else:
         click.secho("Will run metadata validation on all the metadata files found in the repo.")
         metadata_to_validate = get_all_metadata_files()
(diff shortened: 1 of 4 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/metadata.py
```diff
@@ -37,9 +38,6 @@ def metadata(ctx: click.Context):
 def validate(ctx: click.Context, modified_only: bool) -> bool:
     if modified_only:
         metadata_to_validate = get_modified_metadata_files(ctx.obj["modified_files"])
```
bnchrch, line 40: @alafanechere I think we have an error in this logic.

If a PR adds a new connector but no metadata file is found this will pass when it should fail.

We should add a new helper function
```py
def get_expected_metadata_file_paths(modified_files: Set[Union[str, Path]]) -> Set[Path]:
    # Any folder that matches airbyte-integrations/connectors/source-**, or airbyte-integrations/connectors/destination-**,
    # or airbyte-integrations/connectors/third-party/**/*/ should have a metadata file
    modified_connectors = get_modified_connectors(modified_files)
    modified_connector_file_paths = [connector.code_directory for connector in modified_connectors.keys()]
    expected_metadata_file_paths = [f"{connector_file_path}/{METADATA_FILE_NAME}" for connector_file_path in modified_connector_file_paths]
    return expected_metadata_file_paths;
```

bnchrch (commented):
One change we should make which I can add for you after our call today :)

bnchrch:
@alafanechere this became a little bigger than expected.

I opted to add all missing metadata.yaml files

theres a counter solution where we could add an ignore list.

currently seeing if publish on merge is useful for certain connectors

https://airbytehq-team.slack.com/archives/C043JHEEYKG/p1683669890611309

bnchrch (approved):

PR #25912, opened 2023-05-09
connectors-ci: clean gradle lock file

alafanechere:
## What
On a workflow sequentially running connector pipeline it might happen than a connector gradle build fails without releasing its lockfile. As the gradle cache is shared accross pipelines it might prevent other connector build to run and we find `Timeout waiting to lock journal cache error`. 

## How
Remove the lock file after mounting the cache volume.


tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -409,6 +409,7 @@ def with_gradle(
         .with_exec(["sh", "-c", "curl -fsSL https://get.docker.com | sh"])
         .with_exec(["mkdir", "/root/.gradle"])
         .with_mounted_cache("/root/.gradle", root_gradle_cache, sharing=CacheSharingMode.LOCKED)
+        .with_exec(["rm", "-f", "/root/.gradle/caches/journal-1/journal-1.lock"])
```
alafanechere, line 412: This is the actual change I made, other changes are code formatting.

bnchrch (approved):

PR #25915, opened 2023-05-09
🐛 Source Mixpanel, Source Pinterest, Source Freshdesk: fix builds

davydov-d:
## What
Fix builds for:
 - Source Mixpanel. Expected Records Mismatch: data went missing in one of the streams - `Export`.
 - Source Pipedrive. Expected Records Mismatch: stream emits records that contain extra fields (neither exposed in schema nor documented in the api reference).
 - Source Freshdesk. Expected Records Mismatch: stream `TicketFields` - new item added to the `choices` field in one of the records: `"Chat - Internal Task": 14`
 - Source Pinterest. Expected Records Mismatch: the schema used to be missing a field. Stream `BoardPins` now has one more field: `creative_type`

## How
- Mixpanel: refill the stream and update expected records
- Pipedrive: Do not emit undocumented fields as part of a record
- Freshdesk: update expected records
- Pinterest: update expected records + expose new field in the stream schema, bump version


airbyte-integrations/connectors/source-pipedrive/source_pipedrive/streams.py
```diff
@@ -77,9 +77,14 @@ def parse_response(self, response: requests.Response, **kwargs) -> Iterable[Mapp
...
         :return an iterable containing each record in the response
         """
+        documented_properties = self.get_json_schema()["properties"]
         records = response.json().get(self.data_field) or []
         for record in records:
             record = record.get(self.data_field) or record
+            # do not emit non-documented top level fields as part of a record
+            if isinstance(record, dict):
+                # it can also be a list, so need to handle it
+                record = {key: value for (key, value) in record.items() if key in documented_properties}
```
erohmensing, line 87: Why do we need to restrict this in the source? This is something the platform will be taking care of. If the fields are not documented, we should document them instead of not emitting them. See [this issue](https://github.com/airbytehq/airbyte/issues/24493)

airbyte-integrations/connectors/source-pinterest/source_pinterest/schemas/board_pins.json
```diff
@@ -9,6 +9,10 @@
     "created_at": {
       "type": ["null", "string"]
     },
+    "creative_type": {
+      "type": ["null", "string"],
+      "enum": ["REGULAR", "VIDEO", "CAROUSEL", "MAX_VIDEO", "SHOP_THE_PIN", "IDEA"]
+    },
```
erohmensing, line 15: Just wondering, what was the symptom of tests failing as a result here? [There are many undocumented fields in the pinterest connector](https://github.com/airbytehq/airbyte/issues/24492), so I wouldn't expect tests to fail due to `fail_on_extra_columns: false`

erohmensing (changes requested):

airbyte-integrations/connectors/source-pinterest/source_pinterest/schemas/board_pins.json
davydov-d: Tests failed because of expected and actual records mismatch

airbyte-integrations/connectors/source-pipedrive/source_pipedrive/streams.py
davydov-d: That would be too easy :smile: 
This connector's build failed due to the fields that are not even documented by the source API (Pipedrive)

davydov-d: I don't think we should document and expose fields that are not documented by Pipedrive

airbyte-integrations/connectors/source-pinterest/source_pinterest/schemas/board_pins.json
erohmensing: Got it - i see quite a few more things in the diff
image

But am fine with whatever additions make the test pass for now - I just feel like this wasn't the actual source of the mismatch 😄

airbyte-integrations/connectors/source-pipedrive/source_pipedrive/streams.py
erohmensing: Oh... that is indeed sort of yikes 🤔 What types of info are they? Is there anything in the github issue that I made that _is_ documented by pipedrive? At this point I would rather write a distinct list of columns to ignore (that we know aren't documented by pipedrive) instead of always non-emitting ones that we haven't declared. That makes us unresilient to new (documented) things in the pipedrive API - we won't notice them nor emit them.

airbyte-integrations/connectors/source-pinterest/source_pinterest/schemas/board_pins.json
davydov-d: yep, that's just reordering - the only diff here is `"creative_type": "REGULAR"`

airbyte-integrations/connectors/source-pipedrive/source_pipedrive/streams.py
davydov-d: and the data type of those fields varies a lot - strings, integers, objects etc

erohmensing: > new undocumented fields - we won't notice them in the output as well

we will, actually! They will appear on the connector problems dashboard under [Record Unexpected Fields](https://app.datadoghq.com/dashboard/hq4-ush-fht/connector-problems?from_ts=1683044883832&to_ts=1683649683832&live=true)

> And hardcoding things like 3ce5b1409718d65a8adc965be1f0da8821d8b9ab (from your issue) makes me a bit nervous 😸

Yeah I totally get this, I wonder if we can do something specific for that stream. 

> and the data type of those fields varies a lot - strings, integers, objects etc

Not sure exactly what you're getting at here - for a given field does it differ? Or all the fields have different types? 


Let me see if other people have opinions on how we want to handle fields undocumented by the API in general

davydov-d: > Not sure exactly what you're getting at here - for a given field does it differ? Or all the fields have different types?

that was just an answer to your question - "What types of info are they?"

erohmensing: Ah, I didn't mean `type`, just more meant "what information are they sending through that they're not documenting"

davydov-d: @erohmensing did you have a chance to know what people think of this?

davydov-d:
@erohmensing I've excluded Pipedrive changes from this PR as we talked before

erohmensing (approved):

erohmensing:
Thanks! Other changes look good

lazebnyi (approved):

PR #25921, opened 2023-05-09
connectors-ci: remove slack channel mention on publish failures

alafanechere:
## What
The channel mention are annoying.
Removing them until our publish pipeline is more stable and in production.


supertopher (approved):
LGTM

PR #25947, opened 2023-05-10
Source Google Ads: improve GAQL parser error messages

grubberr:
## What
Try to improve this oncall https://github.com/airbytehq/oncall/issues/1974

The customer entered GAQL query:
```
"SELECT customer.id, metrics.cost_micros, campaign_name FROM customer_campaign, labels"
```

On the first sign, this query looks valid, but in really google does not allow multiple tables in FROM clause.
I have improved error handling to provide good error message for such cases:

```
Incorrect GAQL query: multiple resources 'customer_campaign, labels' is not allowed
```


lazebnyi (approved):
LGTM! Please update PR description with how and what will be changed with this PR.

PR #25966, opened 2023-05-10
Upload metadata on merge to master with GHA

bnchrch:
## What
Use GHA instead of dagger temporarily

### DIFF
--- a/.github/workflows/upload-metadata-files.yml
+++ b/.github/workflows/upload-metadata-files.yml
@@ -11,9 +11,6 @@ jobs:
   deploy-catalog-to-stage:
     name: "Upload Metadata Files to Metadata Service"
     runs-on: ubuntu-latest
-    env:
-      DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD }}
-      DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME }}
     steps:
       - name: Checkout Airbyte Cloud
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/upload-metadata-files.yml
```diff
@@ -0,0 +1,52 @@
...
+    paths:
+      - "airbyte-integrations/connectors/**/metadata.yaml"
+
+jobs:
+  upload-metadata-files-to-gcs:
+    name: "Upload changed Metadata Files to Prod Bucket"
+    runs-on: ubuntu-latest
+    env:
+      DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_PASSWORD }}
+      DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME }}
```
pedroslopez, line 16: Can we move this to the step that requires it?

pedroslopez (approved):

.github/workflows/upload-metadata-files.yml
bnchrch: Good call out!

alafanechere (changes requested):
I think we should also remove this logic from the "publish on merge" if you want to rely on GHA instead of dagger for this job:
https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py#L222

.github/workflows/upload-metadata-files.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: tj-actions/changed-files@v35
+        with:
+          files: "airbyte-integrations/connectors/**/metadata.yaml"
+      - name: Setup Python 3.10
+        if: steps.changed-files-specific.outputs.any_changed == 'true'
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: Install metadata_service
+        if: steps.changed-files-specific.outputs.any_changed == 'true'
```
alafanechere, line 28: nit: you could add a previous step to cancel the workflow run `steps.changed-files-specific.outputs.any_changed == 'false'`

bnchrch:
@alafanechere Im thinking we should leave the upload metadata step in publish for two reasons
1. Im hoping the GHA is short lived (under 2 weeks at most)
2. I want to be sure that metadata publish works with dagger publish before we release.

Thoughts?

alafanechere (approved):
> Thoughts?

I was just concern about race conditions / extra triggers that could be caused by running both pipelines in parallel. But the MD5 sensor should prevent extra triggers.

bnchrch:
@alafanechere yup im hoping the MD5 saves us on the double uploads.

Leaving in the metadata upload in the publish pipeline will also save us from forgetting to add the upload step back when we remove `/publish` (if we dont then this GHA action will fail on a race condition if it doesnt find a docker hub image)

PR #25969, opened 2023-05-10
Update dagster sensors to use a md5 based cursor

bnchrch:
## What
We have far to many metadata files to use a concatenated string array as a cursor.

This PR updates our cursors to be MD5 versions of the underlying string arrays

### DIFF
--- a/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/sensors/metadata.py
+++ b/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/sensors/metadata.py
@@ -1,5 +1,5 @@
 from dagster import sensor, RunRequest, SkipReason, SensorDefinition, SensorEvaluationContext, build_resources, DefaultSensorStatus
-from orchestrator.utils.dagster_helpers import string_array_to_cursor
+from orchestrator.utils.dagster_helpers import string_array_to_hash
 
 
 def metadata_updated_sensor(job, resources_def) -> SensorDefinition:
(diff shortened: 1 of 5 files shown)


airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/sensors/metadata.py
```diff
@@ -20,26 +20,24 @@ def metadata_updated_sensor_definition(context: SensorEvaluationContext):
         with build_resources(resources_def) as resources:
             context.log.info("Got resources for gcs_metadata_updated_sensor")
 
-            etags_cursor_raw = context.cursor or None
-            etags_cursor = deserialize_composite_etags_cursor(etags_cursor_raw)
-            etags_cursor_set = set(etags_cursor)
+            etags_cursor_raw = context.cursor or []
+            etags_cursor = string_array_to_cursor(etags_cursor_raw)
```
alafanechere, line 24: I'm not sure I get how this work. I thought cursor are MD5 strings, `etag_cursor_raw` is a string and not a list of string when `context.cursor` is not `False`, so `string_array_to_cursor` will be called with a string, not a string array.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/utils/dagster_helpers.py
```diff
@@ -13,28 +14,15 @@ def output_dataframe(result_df: pd.DataFrame) -> Output[pd.DataFrame]:
     return Output(result_df, metadata={"count": len(result_df), "preview": MetadataValue.md(result_df.to_markdown())})
 
 
-def deserialize_composite_etags_cursor(etag_cursors: Optional[str]) -> List[str]:
-    """Deserialize a cursor string into a list of etags.
+def string_array_to_cursor(strings: List[str]) -> str:
```
alafanechere, line 17:
```suggestion
def hash_list_of_string(strings: List[str]) -> str:
```

I would find it more readable to rename it this way. I orinally thought cursor was a specific type.

alafanechere (changes requested):

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/sensors/metadata.py
bnchrch: A damn, a hasty miss by me. Youre right this doesn't make sense.

alafanechere (approved):

Issue #25687, opened 2023-04-28
Categorize Pinterest config errors


nataliekwong:
This error is shown as a system error, and should be recategorized as a config error.

This should be tackled in the Connector Spec, so that anything not entered as a date can be rejected before testing the connector.
- String does not match format YYYY-MM-DD

artem1205:
Fixed in https://github.com/airbytehq/airbyte/pull/25984

PR #25984, opened 2023-05-11
Source Pinterest: Add pattern for start_date

artem1205:
## What
Resolve :

- https://github.com/airbytehq/airbyte/issues/25687
- https://github.com/airbytehq/oncall/issues/2027

## How
Add re for start_date

## Recommended reading order
1. `x.java`
2. `y.python`

## 🚨 User Impact 🚨

no breaking changes

## Pre-merge Actions

### Airbyter

- Build is successful

</details>


airbyte-integrations/connectors/source-pinterest/source_pinterest/spec.json
```diff
@@ -11,6 +11,9 @@
         "type": "string",
         "title": "Start Date",
         "description": "A date in the format YYYY-MM-DD. If you have not set a date, it would be defaulted to latest allowed date by api (89 days from today).",
+        "format": "date",
```
lazebnyi, line 14: It look like `String does not match format YYYY-MM-DD` we get during check. 
Let's catch error and rise `AirbyteTracedException` with `config_error` failure type.

lazebnyi (changes requested):

airbyte-integrations/connectors/source-pinterest/source_pinterest/spec.json
erohmensing: I think it might be bets to do both! 
* Adding the pattern to the spec makes it so that users can only enter the right pattern in the UI, so they'll never run into it in check 
* Excepting the issue in check and emitting a config_error or failed connection_status message will make it so that anyone using the docker interface will know its not a system_error

artem1205: updated

airbyte-integrations/connectors/source-pinterest/source_pinterest/source.py
```diff
@@ -346,7 +347,15 @@ def _validate_and_transform(self, config: Mapping[str, Any]):
         if not start_date:
             config["start_date"] = latest_date_allowed_by_api
         else:
-            config["start_date"] = pendulum.from_format(config["start_date"], "YYYY-MM-DD")
+            try:
+                config["start_date"] = pendulum.from_format(config["start_date"], "YYYY-MM-DD")
```
lazebnyi, line 351: Can you add unit test for this functionality.

lazebnyi (approved):
LGTM! Minor changes requested.

erohmensing:
Thanks! Changes look good. 

Re: no breaking changes - there are users who have specs saved that don't match the new pattern. If I understand correctly, all of those syncs are failing. But now they will fail at a different step (is that correct?)

artem1205:
> Thanks! Changes look good.
> 
> Re: no breaking changes - there are users who have specs saved that don't match the new pattern. If I understand correctly, all of those syncs are failing. But now they will fail at a different step (is that correct?)

Yep, they'll still face an error, but now it will be handled as a `config_error`.

artem1205:
@erohmensing, please approve if everything looks good

sherifnada (approved):
approving on behalf of @erohmensing since she's OOO

PR #25993, opened 2023-05-11
Normalization: Better handling for interesting CDC situations

edgao:
## What
Similar to how source-mysql CDC will sort on both `_ab_cdc_updated_at` AND `_ab_cdc_log_pos`, source-postgres CDC should sort using both updated_at AND `_ab_cdc_lsn`.

Note that this adds a new `_ab_cdc_lsn` column to the final table. I think that's fine.

Long-term:
* DB sources should tell us which field(s) to use >.>
    * the protocol should support composite cursors?
* or maybe DB sources give us better guarantees about `_airbyte_emitted_at`, and we can remove the hardcoded CDC column references?

## How
Copy-paste the `_ab_cdc_log_pos` stuff.

We'll need to publish:
* `bases/base-normalization`
* `connectors/destination-bigquery` (because normalization code is copied into the bigquery connector)

## Recommended reading order
1. stream_processor.py - this is the actual code change
2. messages.txt, messages_incremental.txt, messages_schema_change.txt - these are the raw records that get written to the destination as part of the integration tests
    *also see `dbt_integration_test.py` - I swapped the comment character from `#` to `//` to align better with jsonl conventions
3. catalog.json, catalog_schema_change.json - in practice, source-postgres leaves the cursor field empty. Replicate that behavior in our tests.
4. Everything under dbt_test_config
    1. the `test_dedup_cdc_excluded....sql` files are inputs to `dbt test`. If they return nonzero records, `dbt test` will throw an error.
    2. the `dedup_cdc_excluded_....incorrect_names.sql` files are the inputs to those `test_....sql` files. We're asserting the exact set of records that we expect to see after each sync.

## 🚨 User Impact 🚨
source-postgres CDC users will see a new column `_ab_cdc_lsn` in their SCD and final tables.


airbyte-integrations/bases/base-normalization/normalization/transform_catalog/stream_processor.py
```diff
@@ -804,6 +804,13 @@ def generate_scd_type_2_model(self, from_table: str, column_names: Dict[str, Tup
             cdc_cols += f", {cast_begin}{col_cdc_log_pos}{cast_as}" + "{{ dbt_utils.type_string() }}" + f"{cast_end}"
             quoted_cdc_cols += f", {quoted_col_cdc_log_pos}"
 
+        if "_ab_cdc_lsn" in column_names.keys():
```
edgao, line 807: this is copied from the if-statement right above, which enables sorting on `_ab_cdc_log_pos`

airbyte-integrations/bases/base-normalization/integration_tests/resources/test_simple_streams/data_input/catalog.json
```diff
@@ -159,7 +159,7 @@
         "default_cursor_field": []
       },
       "sync_mode": "incremental",
-      "cursor_field": ["_ab_cdc_lsn"],
+      "cursor_field": [],
```
edgao, line 162: this replicates cdc behavior in practice, where source-postgres doesn't declare an explicit cursor field

(we should maybe reconsider that behavior, but I think that's a separate conversation)

airbyte-integrations/bases/base-normalization/integration_tests/dbt_integration_test.py
```diff
@@ -415,7 +415,7 @@ def writer():
                             line = input_data.readline()
                             if not line:
                                 break
-                            if not line.startswith(b"#"):
+                            if not line.startswith(b"//"):
```
jbfbell, line 418: should both comment types be considered?

airbyte-integrations/bases/base-normalization/integration_tests/dbt_integration_test.py
```diff
@@ -476,7 +476,9 @@ def run_check_dbt_command(self, normalization_image: str, command: str, cwd: str
         """
         Run dbt subprocess while checking and counting for "ERROR", "FAIL" or "WARNING" printed in its outputs
         """
-        if normalization_image.startswith("airbyte/normalization-oracle") or normalization_image.startswith("airbyte/normalization-clickhouse"):
+        if normalization_image.startswith("airbyte/normalization-oracle") or normalization_image.startswith(
+            "airbyte/normalization-clickhouse"
+        ):
```
jbfbell, line 481: small not totally necessary optimization for readability
```suggestion
             if any([normalization_image.startswith(x) for x in ["airbyte/normalization-oracle", "airbyte/normalization-clickhouse"]]):
```

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/stream_processor.py
```diff
@@ -804,6 +804,13 @@ def generate_scd_type_2_model(self, from_table: str, column_names: Dict[str, Tup
...
+            col_cdc_lsn = self.name_transformer.normalize_column_name("_ab_cdc_lsn")
+            quoted_col_cdc_lsn = self.name_transformer.normalize_column_name("_ab_cdc_lsn", in_jinja=True)
+            cdc_updated_order_pattern += f"\n            {col_cdc_lsn} desc,"
```
jbfbell, line 810: is the odd spacing here intentional?

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/stream_processor.py
```diff
@@ -804,6 +804,13 @@ def generate_scd_type_2_model(self, from_table: str, column_names: Dict[str, Tup
...
+            cdc_cols += f", {cast_begin}{col_cdc_lsn}{cast_as}" + "{{ dbt_utils.type_string() }}" + f"{cast_end}"
```
jbfbell, line 811: this is fine, the mixture of f strings and non f strings which represent jinja replacement tokens threw me off. Random non required suggestions
```suggestion
            cdc_cols += f", {cast_begin}{col_cdc_lsn}{cast_as}" + "{{ dbt_utils.type_string() }}" + f"{cast_end}"
            cdc_cols += "".join([", ", cast_begin, col_cdc_lsn, cast_as, "{{ dbt_utils.type_string() }}", cast_end])
```

jbfbell, line 811: this is fine, the mixture of f strings and non f strings which represent jinja replacement tokens threw me off. Random non required suggestions
```suggestion
            cdc_cols += "".join([", ", cast_begin, col_cdc_lsn, cast_as, "{{ dbt_utils.type_string() }}", cast_end])
```

jbfbell (approved):

airbyte-integrations/bases/base-normalization/integration_tests/dbt_integration_test.py
edgao: this is only applied to the `messages*.txt` files. I'd prefer for us to just pick a convention and use it everywhere (and my vs code extension is only highlighting `//` :P )

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/stream_processor.py
edgao: this eventually gets rendered to a sql statement, so the spacing is to make that sql more readable

(the real answer is that I copypasta-ed it, but I think that's the reason)

airbyte-integrations/bases/base-normalization/integration_tests/dbt_integration_test.py
edgao: done

airbyte-integrations/bases/base-normalization/normalization/transform_catalog/stream_processor.py
edgao: done

PR #25996, opened 2023-05-11
source-bing-ads: implement retry logic if SSL verification failed

alafanechere:
## What
Trying to mitigate OC issue https://github.com/airbytehq/oncall/issues/1997
Bing Ads API transiently returns invalid certificates.

## How
Retry requests when SSL Error is encountered.

## 🚨 User Impact 🚨
This change should mitigate problems related to SSL certificate validation for Bing Ads connector users. Due to the introduction of a retry logic, It might lead to slowest (but working) syncs.


alafanechere:
I think tests are failing for a reason unrelated to my changes. Going to give a CDK upgrade a try.

davydov-d (approved):

PR #26020, opened 2023-05-12
[lowcode] OAuth Authenticator: cast token expiry time to number

clnoll:
## What
Update the OAuth authenticator to cast `value` to an integer, when a time in seconds is passed in instead of a date & format.

Closes https://github.com/airbytehq/airbyte/issues/25992.


airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
```diff
@@ -117,7 +117,11 @@ def set_token_expiry_date(self, value: Union[str, int]):
         if self.token_expiry_date_format:
             self._token_expiry_date = pendulum.from_format(value, self.token_expiry_date_format)
         else:
-            self._token_expiry_date = pendulum.now().add(seconds=value)
+            try:
+                self._token_expiry_date = pendulum.now().add(seconds=int(value))
+            except ValueError:
+                raise ValueError(f"Invalid token expiry value {value}; an integer is required.")
```
sherifnada, line 123: why is an int required? what's wrong with a float?

clnoll: Based on type annotations, pendulum is expecting an int. That said, it doesn't error if given a float so we could accept one.

sherifnada: alright, I guess an int is probably fine then (`int(<float>)` should truncate down which is fine in this case I think)

sherifnada (approved):

PR #26023, opened 2023-05-12
✨ Source Jira: worklog replications speed up, add undeclared fields

### LINKED ISSUE #14670: Source Jira: worklog replication speed up
Replicating worklogs from jira source with large amount of issues.

It would be nice to have a separate stream for all worklogs for this API endpoint: https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-worklogs/#api-rest-api-3-worklog-list-post
It should use this endpoint to request the worklog ids which is in sync with the previous request (1000 record limitation)
https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-worklogs/#api-rest-api-3-worklog-updated-get

If this is an acceptable solution and the dev environment is compatible with windows OS without using WSL, yes.

darynaishchenko:
resolved: https://github.com/airbytehq/airbyte/issues/14670
**Problem**: 
Replicating worklogs from jira source with large amount of issues.
**Solution**:
Unfortunately, the API does not provide an opportunity to extract all worklogs based on the projects to which they belong. Therefore, we cannot refuse to synchronize worklogs according to the issue to which it refers. There is only an option to extract all worklogs present in the account. So a division was made by configuration, if there are projects in the configuration, we extract them by issue, but if projects are not present, we skip the issue sync and extract all updated worklogs at once.

`Issue Worklogs` and `All issue worklogs` streams work depending on the config. 
If you do not have specified projects in the configuration, `All issue worklogs` stream will work, otherwise `Issue Worklogs`.
This was done in order to speed up work in the case of a large number of issues and extract updated worklogs 
for all projects at once, bypassing the reading of all issues.

Also, Streams were divided into separate files for ease of use. `stream.py` was deleted and added module `streams`.


lazebnyi (approved):
LGTM!

darynaishchenko:
@sherifnada  @erohmensing I also have a question about the new stream. Now, if there are no projects in the config, the Issue Worklogs stream is deleted and All issue worklogs is added, which causes Breaking Change. 
From the user's point of view, he does not need to think about which stream to choose. 
However, we can just add All issue worklogs, which will not cause  Breaking Change.
Which option is better?

sherifnada:
@darynaishchenko i'm getting notifications about this from Github reminders. Is this supposed to be open?

darynaishchenko:
> @darynaishchenko i'm getting notifications about this from Github reminders. Is this supposed to be open?

HI, This PR for resolves https://github.com/airbytehq/airbyte/issues/14670, but we are waiting for more user data to understand that we really need this feature. I guess it's better to close this PR and reopen later.

PR #26025, opened 2023-05-12
Source Github: better error explanation for 401

### LINKED ISSUE #25664: Source Github: update personal access token expired error
## What
Add more transparent depiction of the personal access token expired. Now API return only `Bad credentials, re-authentication or access token renewal is required` or `HTTPError('401 Client Error: Unauthorized for url: https://api.github.com/repos/duneanalytics/spellbook?per_page=100')
`
### Workspaces: 

[link](https://cloud.airbyte.io/workspaces/315b6d66-2bcd-4eb3-839c-d0fc567d78aa/connections/5ce74abb-a663-45db-b9e7-0cebca8b28e7/status#1929826::0)

[link](https://cloud.airbyte.io/workspaces/ee09cef9-e4ac-4763-ba15-c7044f85f0b3/connections/93dccf29-49ac-41d3-82f3-7e0bb3deed38/status#1997946::0)

[link](https://cloud.airbyte.io/workspaces/315b6d66-2bcd-4eb3-839c-d0fc567d78aa/connections/5ce74abb-a663-45db-b9e7-0cebca8b28e7/status#1963648::0)

[link](https://cloud.airbyte.io/workspaces/315b6d66-2bcd-4eb3-839c-d0fc567d78aa/connections/5ce74abb-a663-45db-b9e7-0cebca8b28e7/status#1941140::0)

[link](https://cloud.airbyte.io/workspaces/315b6d66-2bcd-4eb3-839c-d0fc567d78aa/connections/5ce74abb-a663-45db-b9e7-0cebca8b28e7/status#1952283::0)

roman-yermilov-gl:
## What
*Not every case handled when 401 exception happen. Unhandled 401 error gives traceback instead of human readable error description.*

## How
*Added log with more transparent depiction of the personal access token expired instead of just getting error traceback. For this I made the following changes:
- Pass the authentication type variable to constructor method of each stream because we need to know token of what type we currently deal with (Accordion to task description we must handle personal access token only but there one more type)
- Add condition to check whether error code 401 or not and also check the token type. If all is true then log a message: `<token type> renewal is required: <original error message>`. Of course I raise original error in the end to do not break the flow
- Fix acceptance test config: there was expected records changed so I ignored fast changing field
- Fix unit test: because of changes in step 1 the `_get_authenticator` method become not static, so `SourceGithub` need to be instantiated now to call it
- Add unit test to ensure 401 error is properly handled
- Make refactoring to avoid code duplicates: move constants from source and stream modules to separate module as we use it in two different places now*


airbyte-integrations/connectors/source-github/acceptance-test-config.yml
```diff
@@ -32,6 +32,11 @@ acceptance_tests:
       - name: "events"
         bypass_reason: "Only events created within the past 90 days can be showed"
       ignored_fields:
+        workflows:
+        - name: created_at
+          bypass_reason: "Got timezone +03:00 when expected +02:00"
```
lazebnyi, line 37: Why we get +03:00?

lazebnyi (changes requested):
Please update description to what and how master will be changed with your PR.

airbyte-integrations/connectors/source-github/acceptance-test-config.yml
lazebnyi: It is important not to overlook any changes made to the field as it can result in a decrease in test coverage.
I request that you investigate this matter. It may be necessary to verify the time zone settings in the configuration and compare them with the actual summer time. Additionally, please consider updating the expected records with an appropriate justification.

lazebnyi: It is important not to overlook any changes made to the field as it can result in a decrease in test coverage.
I request that you investigate this matter. It may be necessary to verify the time zone settings in the configuration and compare them with the actual summer time. Additionally, please consider updating the expected records with an appropriate justification.

lazebnyi (changes requested):

lazebnyi (approved):

PR #26031, opened 2023-05-12
Low code CDK: Make refresh token in oauth authenticator optional

### LINKED ISSUE #23450: Low-code CDK: Support OAuth client flow as an auth option
### Component Description

Some APIs support a flavor of OAuth known as OAuth Client credentials. In this flavor of OAuth, a client ID / secret are used to obtain an access token, skipping the refresh token entirely. 

Example diagram from Auth0: 
image

Some example APIs which do this are: 
* [auth0](https://auth0.com/docs/get-started/authentication-and-authorization-flow/client-credentials-flow)
* [sellsy](https://api.sellsy.com/doc/v2/#section/Authentication)

### Proposed YAML schema

I would like to support this flow in the current OAuth authenticator or add a new OAuth authenticator which supports this flow. 

**Option 1: Add this to the current OAuth authenticator**
This should be doable if we make the `refresh_token` field optional in the current OAuth definition
```yaml
OAuthAuthenticator:
  description: Authenticator for requests using OAuth 2.0 authentication
  type: object
  required:
    - type
    - client_id
    - client_secret
    - token_refresh_endpoint
  properties:
    type:
      type: string
      enum: [OAuthAuthenticator]
    client_id:
      type: string
    client_secret:
      type: string
    refresh_token:
      type: string
    token_refresh_endpoint:
      type: string
    access_token_name:
      type: string
      default: "access_token"
    expires_in_name:
      type: string
      default: "expires_in"
    grant_type:
      type: string
      default: "refresh_token"
    refresh_request_body:
      type: object
      additionalProperties: true
    scopes:
      type: array
      items:
        type: string
    token_expiry_date:
      type: string
    token_expiry_date_format:
      description: The format of the datetime; provide it if expires_in is returned in datetime instead of seconds
      type: string
    $parameters:
      type: object
      additionalProperties: true
```
Then the user can configure the request body and grant type according to what they need. 

The downsides of this approach is that depending on the `grant_type`, `refresh_token` might actually be required. So it's not the best option. But it is workable. 

**Option 2: Optimal support in current OAuth authenticator**
Likely the optimal way to support this would be to display different required parameters depending on the `grant_type`. Something like: 

```yaml
OAuthAuthenticator:
  description: Authenticator for requests using OAuth 2.0 authentication
  type: object
  oneOf:
    - title: Refresh Token flow
      required:
      - type
      - grant_type
      - client_id
      - client_secret
      - refresh_token
      - token_refresh_endpoint
      properties:
        type:
          type: string
          enum: [OAuthAuthenticator]
        grant_type:
          type: string
          default: "refresh_token"
        client_id:
          type: string
        client_secret:
          type: string
        refresh_token:
          type: string
        token_refresh_endpoint:
          type: string
        access_token_name:
          type: string
          default: "access_token"
        expires_in_name:
          type: string
          default: "expires_in"
        refresh_request_body:
          type: object
          additionalProperties: true
        scopes:
          type: array
          items:
            type: string
        token_expiry_date:
          type: string
        token_expiry_date_format:
          description: The format of the datetime; provide it if expires_in is returned in datetime instead of seconds
          type: string
        $parameters:
          type: object
          additionalProperties: true
      - title: Client Credentials flow
        required:
        - type
        - grant_type
        - client_id
        - client_secret
        - authorization_endpoint
        properties:
          type:
            type: string
            enum: [OAuthAuthenticator]
          grant_type:
            type: string
            default: "client_credentials"
          client_id:
            type: string
          client_secret:
            type: string
          refresh_token:
            type: string
          authorization_endpoint:
            type: string
          access_token_name:
            type: string
            default: "access_token"
          expires_in_name:
            type: string
            default: "expires_in"
          refresh_request_body:
            type: object
            additionalProperties: true
          scopes:
            type: array
            items:
              type: string
          token_expiry_date:
            type: string
          token_expiry_date_format:
            description: The format of the datetime; provide it if expires_in is returned in datetime instead of seconds
            type: string
          $parameters:
            type: object
            additionalProperties: true
```

This is probably the best way to implement this change, but is also likely the most complex

**Option 3: Standalone support (additive change, not a breaking change)**

```yaml
ClientFlowOAuthAuthenticator:
    description: Authenticator for requests using OAuth 2.0 authentication
    type: object
    required:
      - type
      - client_id
      - client_secret
      - authorization_endpoint
    properties:
      type:
        type: string
        enum: [ClientFlowOAuthAuthenticator]
      client_id:
        type: string
      client_secret:
        type: string
      authorization_endpoint:
        type: string
      access_token_name:
        type: string
        default: "access_token"
      expires_in_name:
        type: string
        default: "expires_in"
      grant_type:
        type: string
        default: "client_credentials"
      authorization_request_body:
        type: object
        additionalProperties: true
      scopes:
        type: array
        items:
          type: string
      token_expiry_date:
        type: string
      token_expiry_date_format:
        description: The format of the datetime; provide it if expires_in is returned in datetime instead of seconds
        type: string
      $parameters:
        type: object
        additionalProperties: true
```

### Connector API Docs URL

_No response_

### Pull Request URL

_No response_

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/23450

## How

Makes the refresh_token parameter optional in schema and python classes, add test to fail if it's not set for grant_type "refresh_token".

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
@@ -22,7 +22,7 @@ class Oauth2Authenticator(AbstractOauth2Authenticator):
         token_refresh_endpoint: str,
         client_id: str,
         client_secret: str,
-        refresh_token,
+        refresh_token: str,
         scopes: List[str] = None,
         token_expiry_date: pendulum.DateTime = None,
         token_expiry_date_format: str = None,
--- a/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
 
-from typing import Any, List, Mapping, Optional, Sequence, Tuple, Union
+from typing import Any, List, Mapping, Sequence, Tuple, Union
 
 import dpath
 import pendulum
--- a/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
 
-from typing import Any, List, Mapping, Sequence, Tuple, Union
+from typing import Any, List, Mapping, Optional, Sequence, Tuple, Union
 
 import dpath
 import pendulum
@@ -22,7 +22,7 @@ class Oauth2Authenticator(AbstractOauth2Authenticator):
         token_refresh_endpoint: str,
         client_id: str,
         client_secret: str,
-        refresh_token: str,
+        refresh_token,
         scopes: List[str] = None,
         token_expiry_date: pendulum.DateTime = None,
         token_expiry_date_format: str = None,
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/oauth.py
@@ -31,15 +31,15 @@ class DeclarativeOauth2Authenticator(AbstractOauth2Authenticator, DeclarativeAut
         token_expiry_date (Optional[Union[InterpolatedString, str]]): The access token expiration date
         token_expiry_date_format str: format of the datetime; provide it if expires_in is returned in datetime instead of seconds
         refresh_request_body (Optional[Mapping[str, Any]]): The request body to send in the refresh request
-        grant_type: The grant_type to request for access_token
+        grant_type: The grant_type to request for access_token. If set to refresh_token, the refresh_token parameter has to be provided
     """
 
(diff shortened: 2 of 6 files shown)


sherifnada (commented):
does it make sense to encode this via JSONSchema dependency keywords? https://json-schema.org/understanding-json-schema/reference/conditionals.html

airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
```diff
@@ -22,7 +22,7 @@ def __init__(
         token_refresh_endpoint: str,
         client_id: str,
         client_secret: str,
-        refresh_token: str,
```
girarda, line 25: why is the type annotation removed?

flash1293: That was an oversight, I will add it back

flash1293:
@sherifnada JSON schema isn't great on types that depend on values of other fields - I tested around with if-then-else blocks (https://json-schema.org/understanding-json-schema/reference/conditionals.html#id6) but couldn't make that work - the approach https://stackoverflow.com/questions/9029524/json-schema-specify-field-is-required-based-on-value-of-another-field should work but it's not compatible with how we generate docs and types out of our schema. I think ultimately it would be nice to express this invariant in the schema but it's not worth the downsides based on current tooling

flash1293:
@girarda added back that annotation, could you take another look?

girarda (approved):
:shipit:

PR #26037, opened 2023-05-12
connectors-ci: less scary but harmless error log on publish

### LINKED ISSUE #25622: connectors-ci: on publish we should not display scary but harmless errors in the logs
On a normal publish flow the following log is displayed:
```
#4 ERROR: process "docker-entrypoint.sh docker manifest inspect airbyte/source-pinterest:0.2.6-dev.3e6c8a37b0" did not complete successfully: exit code: 1
```

But it's a red herring:
* We use the `docker manifest inspect` command to check if an image exists
* If the image does not exists the process returns a 1 exit code
* This is handled at a higher level then, when we have this error we know that we can publish

We should redirect the output to negate the exit code.

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/25622

To check if an image can be publish we first check with a `docker manifest` command that the image does not exists yet. When an image does not exists the `docker manifest` command returns a 1 exit code that leads to scary dagger logs and can be confusing for log readers.

## How
Make the command always return `0` status code by piping it to `true`
The rest of the logic does not change: the `docker manifest` output is available on stderr and we read it to confirm the absence of the image.

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
@@ -23,12 +23,11 @@ class CheckConnectorImageDoesNotExist(Step):
         manifest_inspect = (
             environments.with_docker_cli(self.context)
             .with_env_variable("CACHEBUSTER", str(uuid.uuid4()))
-            .with_exec(["docker", "manifest", "inspect", self.context.docker_image_name])
+            .with_exec(["sh", "-c", f"docker manifest inspect {self.context.docker_image_name} || true"])
         )
         manifest_inspect_exit_code = await with_exit_code(manifest_inspect)
         manifest_inspect_stderr = await with_stderr(manifest_inspect)
         manifest_inspect_stdout = await with_stdout(manifest_inspect)
-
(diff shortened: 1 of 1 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -23,12 +23,11 @@ async def _run(self) -> StepResult:
         manifest_inspect = (
             environments.with_docker_cli(self.context)
             .with_env_variable("CACHEBUSTER", str(uuid.uuid4()))
```
evantahler, line 25: not part of this PR, but... hilarious.

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -23,12 +23,11 @@ async def _run(self) -> StepResult:
...
-            .with_exec(["docker", "manifest", "inspect", self.context.docker_image_name])
+            .with_exec(["sh", "-c", f"docker manifest inspect {self.context.docker_image_name} || true"])
```
evantahler, line 26: Ha! wrap it in a bash 'try/catch' basically.  nice.

evantahler (approved):

PR #26018, opened 2023-05-12
Add path filter to metadata validation

bnchrch:
## What
https://airbytehq-team.slack.com/archives/C03VDJ4FMJB/p1683849907595279

PR #26040, opened 2023-05-12
Revert "Add path filter (#26018)"

bnchrch:
This reverts commit 4d59d0345def77050dc69c6c0dd4a13488a8c1af to bring back the validate check for all prs

@marcosmarxm We've had to bring this back due to an awkward state between this being both a required check and a check that exists outside of github actions.

Were exploring the best way to make this check a non blocker to merging community PRs. But for now you will have to fork their branch yourself.

We're sorry for adding a bit of work here.


octavia-approvington:

octavia-approvington (approved):

PR #26044, opened 2023-05-12
connectors-ci: fix condition on image check

alafanechere:
## What
Fix forgotten condition change in check if image exists.


octavia-approvington:

octavia-approvington (approved):

PR #26056, opened 2023-05-14
low-code cdk: make page_size optional for offset and page increment strategies

### LINKED ISSUE #25371: Low-code CDK: The page_size field in PageIncrement and OffsetIncrement should be optional
### Component Description

The page_size field is used to tell the source API how many records to include in each page.
However, some APIs do not accept a parameter specifying the number of records to fetch. [The Discourse API is an example of such API](https://docs.discourse.org/#tag/Notifications/operation/markNotificationsAsRead)

We should therefore make the field optional 

## Acceptance criteria
- [ ] page_size is an optional field for the PageIncrement component
- [ ] page_size is an optional field for the OffsetIncrement component
- [ ] An error is thrown if the DefaultPaginator's page_size_option and the pagination strategy does not set a page size
- [ ] The UI is updated to only render the page_size_option fields when page_size is set (page_size_option becomes required if page_size is set) (similar to what we do for the cursor pagination)

### Proposed YAML schema

```yaml
  PageIncrement:
    description: Pagination strategy that returns the number of pages reads so far and returns it as the next page token
    type: object
    required:
      - type
    properties:
      type:
        type: string
        enum: [PageIncrement]
      page_size:
        description: The number of records to request
        type: integer
      start_from_page:
        type: integer
        default: 0
      $parameters:
        type: object
        additionalProperties: true
  OffsetIncrement:
    description: Pagination strategy that returns the number of records reads so far and returns it as the next page token
    type: object
    required:
      - type
    properties:
      type:
        type: string
        enum: [OffsetIncrement]
      page_size:
        description: The number of records to request
        anyOf:
          - type: integer
          - type: string
      $parameters:
        type: object
        additionalProperties: true
```

### Connector API Docs URL

https://docs.discourse.org/

## Alternative considered
- We could re-structure the schema to group the page_size and the page_size_option into a single optional field. 
  - This would better encapsulate the constraint, but would require a breaking change to the schema at the DefaultPaginator level

```
    description: Default pagination implementation to request pages of results with a fixed size until the pagination strategy no longer returns a next_page_token
    type: object
    required:
      - type
      - pagination_strategy
    properties:
      type:
        type: string
        enum: [DefaultPaginator]
      pagination_strategy:
        anyOf:
          - "$ref": "#/definitions/CursorPagination"
          - "$ref": "#/definitions/CustomPaginationStrategy"
          - "$ref": "#/definitions/OffsetIncrement"
          - "$ref": "#/definitions/PageIncrement"
      decoder:
        "$ref": "#/definitions/JsonDecoder"
      page_size:
        type: object
        required:
          - request_option
          - value
        properties:
          request_option:
            "$ref": "#/definitions/RequestOption"
          value:
            type: integer
      page_token_option:
        anyOf:
          - "$ref": "#/definitions/RequestOption"
          - "$ref": "#/definitions/RequestPath"
      $parameters:
        type: object
        additionalProperties: true
```

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/25371

## What
Makes the `page_size` field optional for offset and page increment pagination strategies because some APIs do not support sending page size in the requests being made.

## How
There are two aspects to this fix. Making the field option for the pagination strategies involves changing our schema and because we are going from required -> optional, this is not a breaking change. The second aspect is that we now need to also handle when to stop pagination if there is no `page_size`. I will detail the thought process behind that in inline comments.

## Recommended reading order
2. `declarative_component_schema.py`
3. `offset_increment.py`
4. `page_increment.py`


airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
```diff
@@ -36,16 +37,23 @@ class OffsetIncrement(PaginationStrategy):
...
+            self._page_size = InterpolatedString(page_size, parameters=parameters)
+        else:
+            self._page_size = None
 
     def next_page_token(self, response: requests.Response, last_records: List[Mapping[str, Any]]) -> Optional[Any]:
-        if len(last_records) < self._page_size.eval(self.config):
+        decoded_response = self.decoder.decode(response)
+
+        # Stop paginating when there are fewer records than the page size or the current page has no records
+        if (self._page_size and len(last_records) < self._page_size.eval(self.config, response=decoded_response)) or len(last_records) == 0:
```
brianjlai, line 56: Thinking about the use case of APIs that do not allow specifying a `page_size`, there is actually one particular use case that we need to also consider. The case is when the API response contains paging metadata that we need to use to determine if there are any more records. Although the number of records to fetch is fixed and cannot be included in the request, we still need to reference a page_size limit in the response to stop paging. Which is why for the Offset strategy, we’re also decoding the response and allowing it to be part of the interpolation context.

See this example in Discourse, where `meta.limit` is needed to stop pagination:
https://docs.discourse.org/#tag/Groups/operation/listGroupMembers

airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/page_increment.py
```diff
@@ -19,15 +19,16 @@ class PageIncrement(PaginationStrategy):
...
     parameters: InitVar[Mapping[str, Any]]
     start_from_page: int = 0
 
     def __post_init__(self, parameters: Mapping[str, Any]):
         self._page = self.start_from_page
 
     def next_page_token(self, response: requests.Response, last_records: List[Mapping[str, Any]]) -> Optional[Any]:
-        if len(last_records) < self.page_size:
+        # Stop paginating when there are fewer records than the page size or the current page has no records
+        if (self.page_size and len(last_records) < self.page_size) or len(last_records) == 0:
```
brianjlai, line 31: This is just our best effort attempt at stopping when the next page attempt has no more records included

I could also include this in the PageIncrement strategy, but right now `page_size` is not interpolated and we’ve haven’t seen a use case yet so I’d rather leave this alone and convert this to an interpolated value when there is a clear need.

airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/paginators/strategies/offset_increment.py
```diff
@@ -55,7 +63,10 @@ def reset(self):
         self._offset = 0
 
     def get_page_size(self) -> Optional[int]:
-        page_size = self._page_size.eval(self.config)
-        if not isinstance(page_size, int):
-            raise Exception(f"{page_size} is of type {type(page_size)}. Expected {int}")
-        return page_size
+        if self._page_size:
+            page_size = self._page_size.eval(self.config)
```
girarda, line 67: this won't work if the page size interpolated string references `response`, right?

brianjlai: That is correct. Although I'm not sure how we can do this because this part of the code doesn't have access to the response object the same way the `next_page_token()` method does. It also is not super critical either because if an API doesn't support sending page_size (like Discourse), it wouldn't matter if we could get a page_size from the response because the API wouldn't utilize it anyway.

The unfortunate part is the inconsistency as to what interpolation contexts the strategy has access and it shouldn't be something the developer should need to worry about. But that being said, its a pretty niche case for APIs that don't need this parameter since we went almost a year before finding one. I'm not sure that a wider refactor of our request options methods is a critical need to address this.

girarda (approved):

PR #26058, opened 2023-05-15
run formatter

cgardens:
## What
* I ran `./gradlew format` and it produced a lot of diffs.
* @sherifnada are we doing something other than running the formatter for the connectors? are the diffs in this PR valid?


sherifnada:
Gradle formatting for connectors doesn't have a clear owner at the moment, though my understanding is @airbytehq/connector-operations is [going to look into this](https://airbytehq-team.slack.com/archives/C03VDJ4FMJB/p1684196001381979). Removing extensibility from the review

PR #26069, opened 2023-05-15
Remove outdated warning from GSC Source Connector docs

tybernstein:
Warning about OAuth not working is no longer accurate, removes it from docs.


nataliekwong (approved):
Looks good to me!

PR #26076, opened 2023-05-15
Bump connector builder's record limit to 1000

girarda:
## What
* Limiting the connector builder to reading 100 records is too restrictive.
* This PR bumps the limit to 1000 records.

## 🚨 User Impact 🚨
* No breaking changes
* The connector builder will return more records


girarda:
@flash1293, I approved your PR. Happy to only bump it in the platform

Issue #26085, opened 2023-05-15
dagger: corrupted image being published?


alafanechere:
I face an occasional problem while publishing an image with the code [here](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py#L62)

1. I build a connector container successfully
2. I can run commands again this container successfully (the `spec` command)
3. The image is pushed successfully to DockerHub

When I try to pull the image to test it out I face the following error:
`failed to register layer: Error processing tar file(exit status 1): archive/tar: invalid tar header`

It happens for the following image:
`docker pull  --platform linux/arm64 airbyte/destination-postgres-strict-encrypt:0.3.27-dev.bd80936152`
`docker pull  --platform linux/amd64  airbyte/destination-postgres-strict-encrypt:0.3.27-dev.bd80936152`

I tried to pull both image architecture and faced the same problem.
[Pipeline](https://github.com/airbytehq/airbyte/actions/runs/4983199283) that led to the publication of this image.

I suspect something wrong happens on call to the Dagger publish function.
**I could not reproduce this problem locally nor on subsequent run of the same pipeline in the CI.**

alafanechere:
@sipsma thank you for trying out.

The `failed to register layer: Error processing tar file(exit status 1): archive/tar: invalid tar header` is the output of my local `docker pull` command. Nothing looks suspicious during the dagger execution.

I also tried to use this image in Dagger code:
```python
import anyio
import dagger

async def main():
    async with dagger.Connection() as client:
        ctr = (
            client.pipeline("test")
            .container()
            .from_("airbyte/destination-postgres-strict-encrypt:0.3.27-dev.bd80936152")
            .with_exec(["spec"])
        )
        output = await ctr.stdout()
        print(output[:300])

anyio.run(main)
```

and face the following error:
```
dagger.exceptions.QueryError: runtime error: index out of range [0] with length 0
```

alafanechere:
Thank you @sipsma for the deep investigation and the clear explanations!
You're right that we can't easily force our users to upgrade their Docker engines and more flexibility on the Dagger side would be appreciated. Nevertheless, we currently do not require a specific Docker version [here](https://docs.airbyte.com/quickstart/deploy-airbyte) and I think it'd be great to require a recent version that matches these new performance features. @cpdeethree what's the docker version used in our production Kubernetes clusters?


[...]

PR #26088, opened 2023-05-15
connectors-ci: check if the pushed image can be pulled

alafanechere:
## What
https://github.com/airbytehq/airbyte/issues/26085 suggests that published images might very rarely not be pullable.
I'd like to add a safety net on the publish flow and to always pull images that were just pushed and try to run SPEC on them.
The networking overhead is worth the peace of mind IMO.

## How
* Pull the published images just after push
* If pull fails no spec cache or metadata upload

**Bonus: do not try to upload metadata or spec cache during a pre-release flow.**

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
@@ -89,6 +89,9 @@ class PullConnectorImageFromRegistry(Step):
                     return StepResult(self, status=StepStatus.FAILURE, stderr=f"Failed to pull {self.context.docker_image_name}")
             return StepResult(self, status=StepStatus.SUCCESS, stdout=f"Pulled {self.context.docker_image_name} and ran spec command")
         except QueryError as e:
+            if attempt > 0:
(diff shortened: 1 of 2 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -73,6 +73,21 @@ async def _run(self, built_containers_per_platform: List[Container]) -> StepResu
             return StepResult(self, status=StepStatus.FAILURE, stderr=str(e))
 
 
+class PullConnectorImageFromRegistry(Step):
+    title = "Pull connector image from registry"
+
+    async def _run(self) -> StepResult:
+        try:
+            exit_code = await with_exit_code(
```
bnchrch, line 81: Do we need any retry logic?

bnchrch (approved):
Approved as long as you dont think there will be a race condition :)

alafanechere:
> Approved as long as you dont think there will be a race condition :)

The steps are sequential so I have the guarantee docker push happenned before docker pull. 
But you're right :) I'll add a retry logic.

PR #26094, opened 2023-05-15
Destination: dev-null - Add icon

terencecho:


terencecho:
/publish connector=connectors/destination-dev-null auto-bump-version=false run-tests=false

bnchrch:
/publish connector=connectors/destination-dev-null auto-bump-version=false run-tests=false

bnchrch (approved):
@terencecho Approving! Looks like my publish failed as the image was already published.

You should be good to merge

PR #26095, opened 2023-05-15
connectors-ci: slim down Java images

alafanechere:
## What
Java connectors images built with dagger are ~2x bigger that the one built with Dockerfiles.

## How
* Use `with_mounted_directory` instead of `with_directory` to mount the built artifact: The mounted directory is not stored as a buildkit layer. This was the reason why the Dagger built images were fatter. `with_directory` instruction correspond to a docker COPY, stored as a buildkit layer. `with_mounted_directory` acts as a runtime volume mount and is what we shall use for multi stage like builds: 

## Fat (note the `copy` layer) 
Screen Shot 2023-05-15 at 23 20 13

## Slim (no more `copy` layer)

Screen Shot 2023-05-15 at 23 44 44

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
@@ -670,9 +670,8 @@ async def with_airbyte_java_connector(context: ConnectorContext, connector_java_
     return (
         base.with_workdir("/airbyte")
         .with_env_variable("APPLICATION", application)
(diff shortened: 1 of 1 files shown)


alafanechere:
It's still not matching 💯 the `latest` size (634MB) but it drastically slimmed down from 1.14GB to 764MB
Screen Shot 2023-05-16 at 00 07 40

octavia-approvington:

octavia-approvington (approved):

PR #26096, opened 2023-05-15
Update low-code reference docs to mention that %s should be used for unix timestamp

girarda:
## What
* Update low-code reference docs to mention that `%s should be used to format timestamps


sherifnada (approved):

flash1293 (approved):
LGTM, should we also add it to the list of suggested options in the frontend?

Issue #25877, opened 2023-05-08
connectors-ci: intermitent Dagger engine errors


alafanechere:
[A recent nightly build ](https://github.com/airbytehq/airbyte/actions/runs/4906113070/jobs/8760332561) and a [publish pipeline](https://github.com/airbytehq/airbyte/actions/runs/4919988172) failed with multiples Dagger engine errors:

## [Nightly build pipeline:](https://github.com/airbytehq/airbyte/actions/runs/4906113070/jobs/8760332561)

```
Unavailable: error reading from server: read unix     
@->/run/buildkit/buildkitd.sock: read: connection reset by peer 
```

```
ResourceExhausted: grpc: received message larger than max
```

## [publish pipeline](https://github.com/airbytehq/airbyte/actions/runs/4919988172) : 
```
Unavailable: error reading from server: EOF  
```

```
Unavailable: connection error: desc = "transport: error while dialing: dial unix /var/run/buildkit/buildkitd.sock: connect: connection refused"
```

```
runc run failed: unable to start container process: error during container init: error mounting "/var/lib/dagger/runc-overlayfs/snapshots/snapshots/347127/fs" to rootfs at "/var/lib/docker": stat /var/lib/dagger/runc-overlayfs/snapshots/snapshots/347127/fs: no such file or directory
```

## [another publish pipeline](https://github.com/airbytehq/airbyte/actions/runs/4924896156/jobs/8798483842):
```
│ failed to get layer download url for digest                              │ │
│ │ sha256:5c87408e8ad0e6c07c6c71856aa29c5b4ebd6795febd1ee8eb21aeee5f85ad38: │ │
│ │ Get "http://172.20.82.21:8080/layerDownloadURL": dial tcp                │ │
│ │ 172.20.82.21:8080: connect: connection refused  
```

cpdeethree:
From discord conversations: We upgraded the Dagger engine version that contained some improvements and also bumped the memory of the machines running the ci-system (both magicache and Karpenter run on here, and magicache was running out of memory). Previously we were using two t3.medium machine in this node pool (4gb memory per machine) and now we are using t3.large (8gb of memory). Erik also purged some of our cache which has grown to be quite large. Magicache has been stable since

sipsma:
[...]

PR #26118, opened 2023-05-16
connectors-ci: use crane to check if image exists

alafanechere:
## What
[`crane`](https://github.com/google/go-containerregistry/blob/main/cmd/crane/README.md) is a cool tool to inspect docker registry images. It's actively use by the dagger team for investigation.
It has the benefit of not requiring a docker daemon to run.
I want to replace the use of docker cli by crane in the `CheckConnectorImageDoesNotExist` step to not rely on docker in docker pattern that can be unstable ( see https://github.com/airbytehq/airbyte/issues/25877)

## How
**No logical change == no test change == the test are still passing**:
* Use `crane ls` to check the tag available in an image repo
* If the tag is returned we mark the step as SKIPPED and the push to the docker hub won't happen
* If the tag is not returned or the image repo does not exists: SUCCESS, push will happen
* If the `crane ls` command failed we mark the step as FAILURE and push won't hapen

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
@@ -716,3 +716,7 @@ def with_airbyte_python_connector_full_dagger(context: ConnectorContext, build_p
         .with_label("io.airbyte.version", context.metadata["dockerImageTag"])
         .with_label("io.airbyte.name", context.metadata["dockerRepository"])
     )
+
+
+def with_crane(context: PipelineContext) -> Container:
+    return context.dagger_client.container().from_("gcr.io/go-containerregistry/crane:v0.15.1")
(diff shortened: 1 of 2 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -716,3 +716,7 @@ def with_airbyte_python_connector_full_dagger(context: ConnectorContext, build_p
         .with_label("io.airbyte.version", context.metadata["dockerImageTag"])
         .with_label("io.airbyte.name", context.metadata["dockerRepository"])
     )
+
+
+def with_crane(context: PipelineContext) -> Container:
```
bnchrch, line 721: nit: lets link to the crane docs here and talk about why crane

bnchrch (approved):
One nit but other wise LGTM!

Heres a couple of cranes

PR #26119, opened 2023-05-16
connectors-ci: credentials to spec cache and metadata bucket not required on pre-release

alafanechere:
## What
When publish pre-release images no interaction with the spec cache or metadata bucket happens.
The credentials to connect to these bucket should not be required by the CLI.

## How
* Make the bucket name / credentials CLI options not required
* Add a validate function to fail the command if the credentials do not exists when running for a main release publish


bnchrch (approved):
Yup! Was feeling this pain yesterday. Thanks for the improvement

Issue #26066, opened 2023-05-15
Source Google Analytics (GA4): Fix Pagination


mpetrykin:
### Connector Name

source-google-analytics-data-api

### Connector Version

0.2.2

### What step the error happened?

Other

### Revelant information

Right now the connector does not handle pagination. The **next_page_token** method checks whether **offset** and **limit** parameters are present in the response, but it does not return them according to the [documentation](https://developers.google.com/analytics/devguides/reporting/data/v1/rest/v1beta/RunReportResponse).
You can see only the **row_count** field come from response.
![image](https://github.com/airbytehq/airbyte/assets/40140203/527ff7df-585a-41f8-95a2-e480244f297c)

### Contribute

- [X] Yes, I want to contribute

marcosmarxm:
@mpetrykin please update the title to follow the convention `Source Google Analytics: issue`

marcosmarxm:
The PR was escalated to GL Team.

mpetrykin:
The issue was resolved by https://github.com/airbytehq/airbyte/pull/26126

PR #26126, opened 2023-05-16
Source Google Analytics (GA4): Fix Pagination

mpetrykin:
## What
Fix the pagination mechanism. Problem is described in [26066 issue](https://github.com/airbytehq/airbyte/issues/26066).

## How
1. Add **next_page_offset** property to report config, that increase in condition current next_page_offset or first offset less than **rowCount** from response.
2. Add **offset** and **limit** fields to request body parameters that allowed users define them (it is possible from documentation).


airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/defaults/custom_reports_schema.json
```diff
@@ -21,6 +21,17 @@
...
+      "offset": {
+        "type": ["integer", "string"],
+        "minimum": 0,
+        "pattern": "^[0-9]+$"
+      },
+      "limit": {
+        "type": ["integer", "string"],
+        "minimum": 0,
+        "maximum": 250000,
+        "pattern": "^([0-9]|[1-9][0-9]{1,4}|1[0-9]{5}|2[0-4][0-9]{4}|250000)$"
```
bazarnov, line 33: please remove "pattern", "minimum", "maximum" from added schemas

docs/integrations/sources/google-analytics-data-api.md
```diff
@@ -110,15 +110,16 @@ This connector outputs the following incremental streams:
...
-| 0.1.0   | 2023-01-08 | [20889](https://github.com/airbytehq/airbyte/pull/20889) | Improved config validation, SAT                                                        |
-| 0.0.3   | 2022-08-15 | [15229](https://github.com/airbytehq/airbyte/pull/15229) | Source Google Analytics Data Api: code refactoring                                     |
-| 0.0.2   | 2022-07-27 | [15087](https://github.com/airbytehq/airbyte/pull/15087) | fix documentationUrl                                                                   |
-| 0.0.1   | 2022-05-09 | [12701](https://github.com/airbytehq/airbyte/pull/12701) | Introduce Google Analytics Data API source                                             |
+| Version | Date       | Pull Request                                             | Subject                                                                       |
+|:--------|:-----------|:---------------------------------------------------------|:------------------------------------------------------------------------------|
+| 0.2.3   | 2023-05-16 | [00000](https://github.com/airbytehq/airbyte/pull/00000) | Fix pagination                                                                |
```
bazarnov, line 115: please add real numbers and url for your PR

bazarnov (changes requested):

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/defaults/custom_reports_schema.json
mpetrykin: :heavy_check_mark: done

docs/integrations/sources/google-analytics-data-api.md
mpetrykin: :heavy_check_mark: done

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/defaults/custom_reports_schema.json
```diff
@@ -21,6 +21,12 @@
           "type": "string"
         }
       },
+      "offset": {
+        "type": ["integer", "string"]
```
bazarnov, line 25: What is the data type inside of that field? We should not have both types and expect both, please use the only type that fits the data as follows:
```json
"type": ["null", "integer"]
```

airbyte-integrations/connectors/source-google-analytics-data-api/unit_tests/test_streams.py
```diff
@@ -88,30 +88,36 @@ def test_request_body_json(patch_base_class):
         ],
         "dateRanges": [request_body_params["stream_slice"]],
         "returnPropertyQuota": True,
+        "offset": "0",
```
bazarnov, line 91: use `integer` instead of `string`

airbyte-integrations/connectors/source-google-analytics-data-api/unit_tests/test_streams.py
```diff
@@ -88,30 +88,36 @@ def test_request_body_json(patch_base_class):
...
+        "limit": "10000"
```
bazarnov, line 92: use `integer` instead of `string`

airbyte-integrations/connectors/source-google-analytics-data-api/unit_tests/test_streams.py
```diff
@@ -88,30 +88,36 @@ def test_request_body_json(patch_base_class):
...
     }
 
     request_body_json = GoogleAnalyticsDataApiBaseStream(authenticator=MagicMock(), config=patch_base_class["config"]).request_body_json(**request_body_params)
     assert request_body_json == expected_body_json
 
 
 def test_next_page_token_equal_chunk(patch_base_class):
-    stream = GoogleAnalyticsDataApiBaseStream(authenticator=MagicMock(), config=patch_base_class["config"])
+    additional_config = {
+        "offset": "0",
```
bazarnov, line 101: use `integer` instead of `string`

airbyte-integrations/connectors/source-google-analytics-data-api/unit_tests/test_streams.py
```diff
@@ -88,30 +88,36 @@ def test_request_body_json(patch_base_class):
...
+        "limit": "10000"
```
bazarnov, line 102: use `integer` instead of `string`

bazarnov (changes requested):
Another change is needed, we should not have the multi-types in JSON Schema, as this will cause issues during data normalization, please refer to the comments bellow.

airbyte-integrations/connectors/source-google-analytics-data-api/unit_tests/test_streams.py
mpetrykin: Change all occurrences limit and offset to integer type :heavy_check_mark:

bazarnov:
@marcosmarxm I think we're good to go with this contribution, once CAT is passed.

bazarnov (approved):

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/source.py
```diff
@@ -201,11 +209,17 @@ def request_body_json(
         stream_slice: Mapping[str, Any] = None,
         next_page_token: Mapping[str, Any] = None,
     ) -> Optional[Mapping]:
+
+        self.config["offset"] = self.config.get("offset", 0)
+        self.config["limit"] = self.config.get("limit", 10000)
```
grubberr, line 214: Do we really need to mutate `self.config` dict with default values for `offset`, `limit` in this place?
As I see you do it `instantiate_report_class` function.
Why we need to do it here too ?

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/source.py
```diff
@@ -154,13 +154,21 @@ def get_json_schema(self) -> Mapping[str, Any]:
...
+            else:
+                offset = self.config["offset"]
 
-            if total_rows <= offset:
+            if total_rows <= offset + limit:
+                self.config["next_page_offset"] = None
                 return None
 
-            return {"limit": limit, "offset": offset + limit}
+            self.config["next_page_offset"] = offset + limit
```
grubberr, line 169: Why do we need to keep `next_page_offset` inside `self.config` dict?

The main purpose of this method `def next_page_token` is to pass to next functions like `request_body_json` the value of the token

grubberr (dismissed):

grubberr:
@mpetrykin @sherifnada 

For me also not very clear why we need to define these variables `limit`, `offset` as configuration parameters.

A lot of Airbyte connectors don't bother users with these internal details and keep all this stuff internally.
The connectors just do pagination using some sensible defaults.

All our stream classes use the standard variable: `page_size` which the developer defines to some reasonable value: 

https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/streams/http/http.py#L36

Usually, we define `page_size` as the maximum possible value, but not often on some connectors it can be less value to avoid overloading.

sherifnada:
Agree with @grubberr -- these params should not be exposed

mpetrykin:
@grubberr thanks for pointing out the **page_size** property. Removed the config mutability and set the default value for the limit to 100000. Please review the changes

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/source.py
```diff
@@ -154,13 +154,19 @@ def get_json_schema(self) -> Mapping[str, Any]:
...
 
-        if all(key in r for key in ["limit", "offset", "rowCount"]):
-            limit, offset, total_rows = r["limit"], r["offset"], r["rowCount"]
+        if "rowCount" in r:
+            total_rows = r["rowCount"]
 
-            if total_rows <= offset:
-                return None
+            if self.page_size is None:
+                self.page_size = 100000
```
marcosmarxm, line 161: This is confusing to me. Why we're increasing the page size value? This should be a const value right?
Can you create a constant for the `10000` called PAGE_SIZE = 10000 and add the documentation link to this value? https://developers.google.com/analytics/devguides/reporting/core/v3/reference#maxResults this is the correct one? Should it 10k

IMHO you must rename the variable to `offset` instead of `self.page_size`.

marcosmarxm (changes requested):
Added some comments @mpetrykin

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/source.py
mpetrykin: I agree with self.page_size replacing to offset. Resolved it next commit

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/source.py
mpetrykin: Removed it

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/source.py
```diff
@@ -29,6 +29,8 @@
 # the initial values should be saved once and tracked for each stream, inclusivelly.
 GoogleAnalyticsQuotaHandler: GoogleAnalyticsApiQuota = GoogleAnalyticsApiQuota()
 
+PAGE_SIZE = 100000
```
marcosmarxm, line 32: @mpetrykin can you add a comment directing to where you can find this info?

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/source.py
```diff
@@ -201,12 +210,17 @@ def request_body_json(
...
         next_page_token: Mapping[str, Any] = None,
     ) -> Optional[Mapping]:
+
         payload = {
             "metrics": [{"name": m} for m in self.config["metrics"]],
             "dimensions": [{"name": d} for d in self.config["dimensions"]],
             "dateRanges": [stream_slice],
             "returnPropertyQuota": True,
+            "offset": str(0),
+            "limit": str(100000)
```
marcosmarxm, line 220:
```suggestion
            "limit": str(self.PAGE_SIZE)
```

airbyte-integrations/connectors/source-google-analytics-data-api/unit_tests/test_streams.py
```diff
@@ -88,6 +88,8 @@ def test_request_body_json(patch_base_class):
         ],
         "dateRanges": [request_body_params["stream_slice"]],
         "returnPropertyQuota": True,
+        "offset": str(0),
+        "limit": str(100000)
```
marcosmarxm, line 92:
```suggestion
        "limit": str(self.PAGE_SIZE)
```

marcosmarxm (changes requested):

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/source.py
mpetrykin: done :heavy_check_mark:

airbyte-integrations/connectors/source-google-analytics-data-api/unit_tests/test_streams.py
mpetrykin: done :heavy_check_mark:

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/source.py
```diff
@@ -243,6 +259,11 @@ def request_body_json(
         next_page_token: Mapping[str, Any] = None,
     ) -> Optional[Mapping]:
         payload = super().request_body_json(stream_state, stream_slice, next_page_token)
+
+        # remove offset and limit fields according to their absence in
+        # https://developers.google.com/analytics/devguides/reporting/data/v1/rest/v1beta/properties/runPivotReport
+        payload.pop("offset")
+        payload.pop("limit")
```
marcosmarxm, line 266: To make this operation safe to not break in case the key doesn't exist.
```suggestion
        payload.pop("offset", None)
        payload.pop("limit", None)
```

airbyte-integrations/connectors/source-google-analytics-data-api/unit_tests/test_streams.py
```diff
@@ -14,6 +14,8 @@
 
 from .utils import read_incremental
 
+PAGE_SIZE = 100000
```
marcosmarxm, line 17: Please import this from `source.py`

marcosmarxm (dismissed):
Some nit comments. I'll run tests locally and check what is not working

airbyte-integrations/connectors/source-google-analytics-data-api/source_google_analytics_data_api/source.py
mpetrykin: done :heavy_check_mark:

airbyte-integrations/connectors/source-google-analytics-data-api/unit_tests/test_streams.py
mpetrykin: done :heavy_check_mark:

mpetrykin:
@marcosmarxm hi, sorry to ask, did you manage to run the tests locally and find the problem or haven't had time yet?

marcosmarxm:
Not yet @mpetrykin I'm going to test and run today.

sh4sh (approved):

PR #26133, opened 2023-05-16
Metadata: Check that normalized files exist

### LINKED ISSUE #25538: Add a metadata check that normalizationRepository and normalizationTag exists

bnchrch:
## What
Checks that the normalization images exist on metadata file upload

and adds tests to ensure were doing this check

closes #25538


airbyte-ci/connectors/metadata_service/lib/tests/fixtures/__init__.py
```diff
@@ -3,15 +3,26 @@
...
 
 @pytest.fixture
 def invalid_metadata_yaml_files() -> List[str]:
-    file_path = os.path.join(os.path.dirname(__file__), "invalid")
-    # list the absolute paths of all files in the directory
-    return [os.path.join(file_path, file_name) for file_name in os.listdir(file_path)]
+    return list_all_paths_in_fixture_directory("metadata_validate/invalid")
+
+
+@pytest.fixture
```
alafanechere, line 21: nit: as these files are static I think its worth using the session scope: the fixtures will only be called once per test session and their results will be re-used throughout tests.
```suggestion
@pytest.fixture(scope="session")
```

airbyte-ci/connectors/metadata_service/lib/tests/fixtures/__init__.py
```diff
@@ -3,15 +3,26 @@
 from typing import List
 
 
+def list_all_paths_in_fixture_directory(folder_name: str) -> List[str]:
+    file_path = os.path.join(os.path.dirname(__file__), folder_name)
+    return [os.path.join(file_path, file_name) for file_name in os.listdir(file_path)]
```
alafanechere, line 8: nit: I tend to prefer using Pathlib instead of os.path
```suggestion
    return Path( os.path.dirname(__file__) / folder_name).iterdir()
```

airbyte-ci/connectors/metadata_service/lib/tests/test_gcs_upload.py
```diff
@@ -12,6 +12,44 @@
 from metadata_service.constants import METADATA_FILE_NAME
 
 
+def stub_is_image_on_docker_hub(image_name: str, version: str) -> bool:
+    return "exists" in image_name and "exists" in version
+
+
+def setup_upload_mocks(mocker, version_blob_md5_hash, latest_blob_md5_hash, local_file_md5_hash):
```
alafanechere, line 19: nit: in pytest the `setup` `teardown` pattern is usually declared through the use of fixtures.

alafanechere (approved):
Nice additional check! Just minor pytest related suggestions.

PR #26138, opened 2023-05-16
connectors-ci: check compression type on image layers

alafanechere:
## What
Relates to https://github.com/airbytehq/airbyte/issues/26085
The latest docker version occassionally uses a compression algorithm (`zstd`)  for the image layers that is not backward compatible with Docker version < 21. It means that anyone building a connector image running Docker >= 21 can publish connector images that won't work on some Airbyte deployment (Airbyte Cloud does not support these images today). 

## How
* Inspect the layers of the published images
* If a `zstd` layer is found we fail the pipeline.
* No metadata file will get uploaded to GCS, so the registry won't be regenerated so no incompatbile connector will get released.


bnchrch (approved):
LGTM! Nice find with Crane

PR #26158, opened 2023-05-16
HACKDAY: Icon CDN

bnchrch:
## What
Moves icons to their connector folders and uploads them to a gcs bucket with cdn caching enabled

demo here: https://storage.googleapis.com/dev-airbyte-cloud-connector-metadata-service/registries/v0/cloud_registry.json

## TODO before merge

- [x] Revert the hacks applied to the MetadataUploadPipeline
- [x] Remove the icon url dev asset
- [x] Run format
closes #26158


alafanechere (changes requested):
The code looks good to me but I'd like to have a simple logic discussion.
IMO icon and metadata upload should be separated.
I like the concept of: 
* The publish pipeline makes sure all the artifacts required for a connector release are distributed: docker image, icon, metadata file.
* The metadata service validate that the artifact described in a metadata file are indeed available, before uploading the metadata file.

My suggestion would be to create a different pipeline step, upstream of the metadata upload, that takes care of uploading the icon to GCS.
I think the upload logic could be exposed as a different entrypoint of the metadata service, or an interaction with a gcloud container on the dagger side.

bnchrch:
Hey @alafanechere !

Very fair call out. I just want to ask why do you feel like the icon upload and the metadata upload should be separate?

My reasoning so far for keeping them together is
1. They both are used to generate the registry
2. They are both types of connector metadata
3. They both live in the same folder (so will docs)
4. We may want to move our Icon QA check into metadata validate

bnchrch:
Thanks for replying @alafanechere 

I think were close here!

My main counter here is I would like to keep the logic for the metadata service in one location. So that someone can look in one place to understand the system. This is similar to our goal with the connector folders.

Currently the metadata service has four parts
1. Validation
2. Upload
3. Orchestration (generation)
4. The CI pipeline that encapsulates these

The logic for 1-3 lives in `airbyte-ci/metadata-service` and 4 lives in `pipelines`.

Ideally if I had to make one move to consolidate the logic I'd rather move the metadata pipeline into `airbyte-ci/metadata-service` but that feels like too big of a lift 😅

alafanechere (approved):
👍 Thank you for the discussion! I'm in as long as we have we have a clear logic.

bnchrch:
@bnchrch test

pedroslopez:
@bnchrch yo

PR #25782, opened 2023-05-03
Destination Snowflake: Enable in-connector normalization

edgao:
closes #25502

## What
Enable in-connector normalization for snowflake. See https://github.com/airbytehq/airbyte-platform-internal/pull/6349 for sync screenshot.

## How
Mostly just replicate the bigquery changes.

## 🚨 User Impact 🚨
None.

PR #26165, opened 2023-05-16
Destination Redshift: Install normalization

edgao:
basically just copy https://github.com/airbytehq/airbyte/pull/25782 but for redshift

I ran a sync locally (with platform-side hack to turn on in-connector normalization) and it worked:
image


airbyte-integrations/connectors/destination-redshift/Dockerfile
```diff
@@ -18,3 +37,6 @@ COPY --from=build /airbyte /airbyte
 
 LABEL io.airbyte.version=0.4.7
```
evantahler, line 38: Should there be a version bump here?

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -566,7 +566,7 @@ def with_integration_base_java(context: PipelineContext, build_platform: Platfor
         "dockerfile": "redshift.Dockerfile",
         "dbt_adapter": "dbt-redshift==1.0.0",
         "integration_name": "redshift",
-        "supports_in_connector_normalization": False,
+        "supports_in_connector_normalization": True,
```
evantahler, line 569: 💙

airbyte-config-oss/init-oss/src/main/resources/seed/oss_registry.json
evantahler, line 1: Screenshot 2023-05-16 at 4 36 27 PM
uhh... assuming this is fine

evantahler (approved):

airbyte-config-oss/init-oss/src/main/resources/seed/oss_registry.json
edgao: :shrug: `automated change` is a pathway to many abilities some consider to be unnatural

@bnchrch iiuc this file is going away soon?

airbyte-integrations/connectors/destination-redshift/Dockerfile
edgao: yeah, I haven't done the metadata.yaml / changelog either. just wanted to get the pr up and deal with the tedious bits later... i.e. today 😓

airbyte-config-oss/init-oss/src/main/resources/seed/oss_registry.json
bnchrch: @alafanechere yes sorry this is an artifact of us wanting the registry in the source code to be recent.

So what currently happens is on each PR we
1. Run processResources
2. Which downloads the registry from the gcs bucket
3. We commit the changes

Were going to move this to an mechanism that commits the new file directly to master, just need to discuss if that truly is a good idea.

edgao: cool. So I should resolve the merge conflict by just rerunning processResources + rejecting master changes?

PR #26171, opened 2023-05-17
hook up memory manager to free

cgardens:


airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/BufferManager.java
```diff
@@ -17,25 +17,24 @@
...
-  BufferManagerDequeue bufferManagerDequeue;
+  private final Map<StreamDescriptor, MemoryBoundedLinkedBlockingQueue<AirbyteMessage>> buffers = new HashMap<>();
+  private final BufferManagerEnqueue bufferManagerEnqueue;
+  private final BufferManagerDequeue bufferManagerDequeue;
   private final ScheduledExecutorService debugLoop = Executors.newSingleThreadScheduledExecutor();
 
   public BufferManager() {
-    buffers = new HashMap<>();
-    final var memoryManager = new GlobalMemoryManager(TOTAL_QUEUES_MAX_SIZE_LIMIT_BYTES);
+    final long availableMemory = (long) (Runtime.getRuntime().maxMemory() * 0.8);
```
davinchia, line 35: can use Line 23.

cgardens: thanks.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/BufferManager.java
```diff
@@ -202,33 +196,37 @@ public void close() throws Exception {
...
+      final var freeMem = maxMemoryBytes - currentMemoryBytes.get();
       // Never allocate more than free memory size.
       final var toAllocateBytes = Math.min(freeMem, BLOCK_SIZE_BYTES);
-      currentMemoryBytes += toAllocateBytes;
+      currentMemoryBytes.addAndGet(toAllocateBytes);
 
       return toAllocateBytes;
     }
 
     public void free(final long bytes) {
```
davinchia, line 225:
```suggestion
    public void synchronized free(final long bytes) {
```

davinchia: what calls 'free'?

cgardens: why does this need to be synchronized? i get it for request memory. but not for free.

davinchia: wait figured out what calls free

davinchia: I believe free is called in each worker pool thread

cgardens: yeah, but it is thread safe. it's just updating an atomiclong. there's no reason why 2 threads can't do that at the same time.

davinchia: you are right

davinchia (approved):

PR #26174, opened 2023-05-17
async destination fix reclaim

cgardens:
## What
* fix reclaiming memory from queues
* suggest which knobs to turn next


ryankfu (approved):
LGTM, it appears we're tuning the knobs to eagerly flush 👍

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/UploadWorkers.java
```diff
@@ -47,15 +47,19 @@ public void start() {
   }
 
   private void retrieveWork() {
+    // todo (cgardens) - i'm not convinced this makes sense. as we get close to the limit, we should
+    // flush more eagerly, but "flush all" is never a particularly useful thing in this world.
```
davinchia, line 51: agree. i think the hard req for flushAll is when we are closing down the destination.

davinchia (approved):
let's try!

PR #26176, opened 2023-05-17
async destination -- lock drain to

cgardens:
## What
* allow draining from the queue to happen in a lock to prevent interleaved reads from different workers.
* took the implementation approach from how `LinkedBlockingQueue` does it by using a `ReentrantLock`.

### DIFF
--- a/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/BufferManager.java
+++ b/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/BufferManager.java
@@ -14,6 +14,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Stream;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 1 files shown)


airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/BufferManager.java
```diff
@@ -154,34 +157,47 @@ public Optional<Instant> getTimeOfLastRecord(final StreamDescriptor streamDescri
...
+          final long newSize = memoryItem.size() + bytesRead.get();
+          if (newSize <= bytesToRead) {
+            bytesRead.addAndGet(memoryItem.size());
+            return true;
+          } else {
+            return false;
+          }
+        }).map(MemoryItem::item)
+            .toList()
+            .stream();
```
cgardens, line 191: hack for now. we can switch this to a for loop, but didn't want to slow us down on this right now.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/BufferManager.java
```diff
@@ -120,11 +121,13 @@ static class BufferManagerDequeue {
 
     private final GlobalMemoryManager memoryManager;
     private final Map<StreamDescriptor, MemoryBoundedLinkedBlockingQueue<AirbyteMessage>> buffers;
+    private final Map<StreamDescriptor, ReentrantLock> bufferLocks;
 
     public BufferManagerDequeue(final GlobalMemoryManager memoryManager,
                                 final Map<StreamDescriptor, MemoryBoundedLinkedBlockingQueue<AirbyteMessage>> buffers) {
       this.memoryManager = memoryManager;
       this.buffers = buffers;
+      bufferLocks = new HashMap<>();
```
ryankfu, line 130: Does this need to be a `ConcurrentHashMap`?

cgardens: i think actually it probably should. 2 threads could try to put the lock at once.

cgardens: buffers should also be a concurrenthasmap

ryankfu (approved):
Makes sense pending updates to `ConcurrentHashMap`

PR #26183, opened 2023-05-17
Source Jira: updated release stage to GA

darynaishchenko:
updated release stage to GA


lazebnyi (approved):
Approved to unblock. But need check why you change `oss_registry.json` file.

PR #26189, opened 2023-05-17
🐛 Source: Stripe Update invoice_items.json so period.end type matches period.start (integer)

tybernstein:
## What
In the Stripe Connector, up until about a week ago, on Airbyte cloud, these both were coming through as integer. Then the end started coming across as numeric and getting turned into scientific notation (which broke systems that were expecting integers). This aligns the two values.

## How
changes period.end from `number` to `integer`

## Recommended reading order
1. `x.java`
2. `y.python`

## 🚨 User Impact 🚨
Possibly? This is a datatype change but I am not sure if it is actually a breaking change.

### DIFF
--- a/airbyte-integrations/connectors/source-stripe/source_stripe/schemas/invoice_items.json
+++ b/airbyte-integrations/connectors/source-stripe/source_stripe/schemas/invoice_items.json
@@ -105,7 +105,7 @@
       "type": ["null", "object"],
       "properties": {
         "end": {
-          "type": ["null", "number"]
+          "type": ["null", "integer"]
         },
         "start": {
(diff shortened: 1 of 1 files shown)


sh4sh:
This field is documented as an `integer` here: https://stripe.com/docs/api/invoiceitems/object#invoiceitem_object-amount

sajarin:
This has been implemented on the master version of the connector: https://github.com/airbytehq/airbyte/blob/12f1304a67c3de024f08d7e496272f9bc89bbef9/airbyte-integrations/connectors/source-stripe/source_stripe/schemas/invoice_items.json#L107C1-L111C38

PR #26210, opened 2023-05-17
Source File: add retry on SSHException('Error reading SSH protocol banner') - 2

grubberr:
## What
Try to fix this oncall https://github.com/airbytehq/oncall/issues/1954
I found one bug in my prev PR  https://github.com/airbytehq/airbyte/pull/26115


artem1205 (approved):

airbyte-integrations/connectors/source-file/source_file/client.py
```diff
@@ -96,7 +96,7 @@ def close(self):
 
     def backoff_giveup(self, error):
         # https://github.com/airbytehq/oncall/issues/1954
-        if isinstance(error, SSHException) and str(error) == "Error reading SSH protocol banner":
+        if isinstance(error, SSHException) and str(error).startswith("Error reading SSH protocol banner"):
```
maxi297, line 99: Should we have a test changing in that case?

maxi297 (approved):
Conditional approval: Given tests are updated, this change seems to make sense!

airbyte-integrations/connectors/source-file/source_file/client.py
grubberr: improved

PR #26213, opened 2023-05-17
🐛 Destination BigQuery (+denormalized): correctly parse buffer count

edgao:
we're not actually parsing the config correctly (it's nested under the `loading_method` object). Extract it correctly.

also swap the order of parameters in the `assertEquals` calls. The convention is `assertEquals(expected, actual)`.


ryankfu (approved):
LGTM, thanks for getting this up!

edgao:
/publish connector=connectors/destination-bigquery,connectors/destination-bigquery-denormalized

PR #26216, opened 2023-05-18
use set instead of list for message type look ups

cgardens:
## What
* same as equivalent pr in the platform repo https://github.com/airbytehq/airbyte-platform-internal/pull/6655


pedroslopez (approved):
Looks fine, though I'm not sure I understand why this is "in favor of" the platform PR - presumably if this is a performance issue the change would need to be done platform-side, not just in the test harness?

Note that there's an equivalent for Sources as well.

davinchia (approved):

Issue #26205, opened 2023-05-17
Remove automatic registry commits to repos


bnchrch:
## Problem
Remove the oss_registry.json from source control

Do not run DownloadCatalog on PRs

PR #26217, opened 2023-05-18
Registry: Refactor and Remove registry download from git.

bnchrch:
## What
This is about preventing those annoying merge conflicts with updated oss_registry.json changes

This PR
1. Removes all the java based download files to a gradle download plugin (fixes a circular dependency preventing the removal of the spec-oss folder
2. Removes the spec-oss folder
3. ignores oss_regsitry.json

Related to #26205


airbyte-config-oss/init-oss/src/main/resources/seed/oss_registry.json
alafanechere, line 1: I think it should be added to a .gitignore , otherwise pending PR might recreate it

alafanechere (approved):
Cool, more 🧹 : 👏

airbyte-config-oss/init-oss/src/main/resources/seed/oss_registry.json
bnchrch: We already do! Its just that the file wasnt deleted from git after being added.

https://github.com/airbytehq/airbyte/blob/bnchrch/metadata/remove-download-registry-pr/.gitignore#L74

One of those weird things where the ignore doesn't work retro actively

PR #26218, opened 2023-05-18
Fix version in run-ab-platform.sh script

flash1293:
## What

This PR manually fixes the version that gets pulled when running Airbyte with the run-ab-platform.sh script.
0.44.4 is not the latest release, https://github.com/airbytehq/airbyte-platform-internal/actions/runs/4978002525 actually released 0.44.5

This is just a stopgap measure so users will get the most recent version, we still need to fix the release process so it bumps this to the right number (the one it builds images for)

### DIFF
--- a/run-ab-platform.sh
+++ b/run-ab-platform.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-VERSION=0.44.4
+VERSION=0.44.5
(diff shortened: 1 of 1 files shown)


davinchia (approved):

flash1293:
Thanks for the quick review @davinchia !

PR #26291, opened 2023-05-19
connectors-ci: force image's layers compression to GZIP

alafanechere:
## What
- Install latest Dagger version to force compression algorithm on the image with publish to GZIP.
- Improve the check of image layers to fail if anything else than GZIP compressed layers is detected.
- Force pipeline concurrency to 1 to avoid concurrent access failures.
- Implement a retry logic in case of transient failure of the `PushConnectorImageToRegistry` step.


alafanechere:
I successfully ran multiple publish with these changes:
https://github.com/airbytehq/airbyte/actions/runs/5022624449

PR #26321, opened 2023-05-19
Remove all select count(*) queries

akashkulk:
Context : https://github.com/airbytehq/oncall/issues/2067#issuecomment-1555340786


rodireich:
@akashkulk There is one `COUNT` we do when the cursor column is nullable (or when syncing a view of a table where nullable is implied).
That `COUNT` is necessary in order to prevent us attempting to sync null cursor values and was written as carefully as we can in oder to only do the potentially costly count when it's absolutely necessary.
I don't think that's the COUNT that's causing this issue, or is it?

akashkulk:
> @akashkulk There is one `COUNT` we do when the cursor column is nullable (or when syncing a view of a table where nullable is implied). That `COUNT` is necessary in order to prevent us attempting to sync null cursor values and was written as carefully as we can in oder to only do the potentially costly count when it's absolutely necessary. I don't think that's the COUNT that's causing this issue, or is it?

The issue here is that the `select count(*)` is quite slow and the theory is that this is causing this sync to hang. What are the side effects of not checking of nullable in general?

cgardens:
We are closing this PR due to inactivity. We regularly close PRs that were created more than 6 months ago. If you'd like to keep working on it, please feel free reopen it. We'd be happy to keep working with you.

PR #26322, opened 2023-05-19
Dagster: Remove all code used for development of new metadata files

bnchrch:
## What
1. Removes dev specific jobs
2. Removes legacy registry watchers
3. Updates generation to be based on the written file, not the dataframe used to write the file

image


evantahler (approved):
🪓 

Of course I want to remove files called "dev" and "legacy".  How could I not want to approve that?

PR #26332, opened 2023-05-22
Have integration runner pass string and its size

cgardens:
## What
* We were struggling with bad size estimates on the async snowflake destination. There's no reason we need to guess; a few a method calls before the message was serialized as a string. This PR adds a new code path to the `IntegrationRunner` that allows it to get the exact size of the message as it deserializes it to a string. It then just passes the string to the destination which let's us play with other optimizations down the road.

## How
* Add code path
* Add test to verify behavior didn't change

## Recommended reading order
1. `airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java`
2. `airbyte-integrations/bases/base-java/src/test/java/io/airbyte/integrations/base/IntegrationRunner2Test.java`


ryankfu (approved):
Overall, lgtm. Could benefit from some javadoc comments but non-blocking

PR #26375, opened 2023-05-22
hide queue inside MemoryBoundedLinkedBlockingQueue

cgardens:
## What
Hiding the actual java `Queue` has an inner class to avoid the chance that someone tries to use native queue methods that we haven't overridden. Good thing that I did this too, because one of the changes we made during hack dayz wasn't reflected in our current feature branch. We need to override `poll(time, unit)` not just `poll`. This PR makes sure we won't make that mistake again!


davinchia (approved):

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/buffers/MemoryBoundedLinkedBlockingQueue.java
```diff
@@ -7,77 +7,135 @@
...
+        return success;
+      } else {
+        currentMemoryUsage.addAndGet(-itemSizeInBytes);
+        log.debug("offer failed");
+        return false;
+      }
+    }
+
+    @Override
+    public MemoryBoundedLinkedBlockingQueue.MemoryItem<E> take() throws InterruptedException {
```
ryankfu, line 113: Do we want to make these methods package-private too since `MemoryBoundedLinkedBlockingQueue.MemoryItem<E> poll` is the public-facing side of this queue?

davinchia: looks like we cannot make this package private as the underlying class has this as public and we cannot override the existing signature with a more private visibility.

PR #26386, opened 2023-05-23
async destination -- introduce state

cgardens:
## What
* Adds a global counter to the destination.
* Uses that to keep track of high watermarks of records that are flushed so that we can decide which state messages are safe to emit.

## Approach
only worried about per-stream right now...
* every record and state message is assigned a message #
* each state message is added into the memory manager _and_ put in the queue. 
* as we pull out of the queue we look at message numbers to determine if what state messages we can emit.

## Recommended reading order
1. `StreamAwareQueue.java` -- start here. I separate anything that isn't strictly memory management out of the memory queue and pulled it up to this level. This level of the onion is aware of streams and airbyte messages.
2. `StateLifecycle.java` -- this is where we track what state messages we have and can emit
3. `MemoryAwareMessageBatch.java` -- i'm not happy with this yet. right now there's a method here that the flusher calls to update the `StateLifecycle` and find out what state message to emit.

### DIFF
--- a/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/RecordSizeEstimator.java
+++ b/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/RecordSizeEstimator.java
@@ -34,8 +34,8 @@ public class RecordSizeEstimator {
    * determined by {@code sampleBatchSize}.
    */
   public RecordSizeEstimator(final int sampleBatchSize) {
-    this.streamRecordSizeEstimation = new HashMap<>();
-    this.streamSampleCountdown = new HashMap<>();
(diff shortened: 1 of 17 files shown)


airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/RecordSizeEstimator.java
```diff
@@ -71,7 +71,7 @@ public long getEstimatedByteSize(final AirbyteRecordMessage record) {
   }
 
   @VisibleForTesting
-  static long getStringByteSize(final JsonNode data) {
+  public static long getStringByteSize(final JsonNode data) {
```
cgardens, line 74: needed access to this but technically in a different package. 🙄  i plan to remove our use of this entirely but not in this PR.

cgardens:
i can do better.

PR #26396, opened 2023-05-23
async destination -- introduce state 2

cgardens:
## What
* I closed the previous PR https://github.com/airbytehq/airbyte/pull/26386 in favor of this one.
* This allows for a unified approach across per-stream and global states for state tracking (at least from outside the state manager).

## How
* Every state and record message is issues a monotonically increasing message number.
* `AsyncStateManager` tracks `StateLifecycle` and `StreamContract`.
* `StateLifecycle` is keeping track of state messages in the order they come in. For global, it's just keeps track of all state messages. For per stream, we have a `StateLifecycle` per stream.
* `StreamContract` tracks claims made by workers on batches. By tracking the max message number for each batch, it can tell us for a give message number if there are any outstanding batches below that number that haven't completed.
* This all comes together in the state manager. Every time a worker completes, we run a check based on the outstanding contracts to understand what's the max message number we could release a state for (based on info in `StreamContract` and the buffers. Then we look up what states we actually have (`StateLifecycle`) and emit any if we have them. Some tests now exist in `AsyncStateManagerTest` and `StateLifecycleTest`.
* The logic for the guarantees we can make on what message numbers are safe to emit needs a lot more testing. This PR in it's current form is on its way to encapsulating that logic so it can be rigorously tested. Close, but needs more work.
* I also stole some tests from the old `BufferedStreamConsumer` to begin testing that this all comes together as expected in `AsyncStreamConsumer`.

## Recommended reading order
1. `AsyncStateManager.java`
2. `StateLifecycle.java`,
3. `StreamContract.java`
4. `MemoryAwareMessageBatch.java`
5. `AsyncStateManagerTest.java`
6. `StateLifecycleTest.java`


cgardens:
closed in favor of davinchia/introduce-state4

PR #26438, opened 2023-05-23
connectors-ci: improve Java connector build logic and Gradle caching strategy

### LINKED ISSUE #26429: connectors-ci: improve the java connectors build and Gradle caching logic
**Current flaws of the Gradle caching logic in our pipeline:**

We create per connector cache volume, caching the `GRADLE_HOME` folder with `LOCKED` cache sharing mode. It means that concurrent access to the cache is not possible, slowing down pipelines and potentially creating Gradle timeouts and transient failure.

Gradle has two level of caching:
* (read only) dependency cache: that can be shared and read concurrently without risk to share java dependency and not downloading them.
* build cache: it's used to store the output of tasks and do not re-run the tasks that previously run. This cache is supposed to be used by a single Gradle process. 

I suggest to:
* Persist the dependency cache in a global volume shared by all connectors with SHARED cache sharing mode
* Persist the build cache in connector specific volume, with LOCKED cache sharing mode to avoid concurrent write.

[Gradle documentation about the dependency cache.
](https://docs.gradle.org/current/userguide/dependency_resolution.html#sub:shared-readonly-cache)

alafanechere:
## What
Close #26429 
With this PR I'd like to:
- improve and rationalize the caching mechanism for Java connectors
- Create the distribution tar for java connectors in step separate from the one building the docker image
- Use Docker 20 instead of Docker 23 for `dind` usecases: Docker 20 is the version we use on the runner, so for consistency it's good that they match. 

## How
*Improve and rationalize the caching mechanism for Java connectors*: 
- Gradle has a "read only dependency cache" which is safe to read concurrently and share through pipelines. I implemented the use of a dedicated global cache for Java dependency. For `source-postgres` the build time is 2mn vs 3mn when using this cache.

- Specifically cache the `build-cache` but keep it per connector as concurrent access is not recommended. The build time went from 2mn to 1mn when using this cache.

*Create the distribution tar for java connectors in step separate from the one building the docker image*:
The step building the distribution tar with gradle is separated from the connector image build.
As we build two images (one per platform) this separation allows a single execution the distribution tar step, the tar file is reused for the 2 images builds downstream.

*Use Docker 20 instead of Docker 23:*
Use `docker:20-dind` and `docker:20-cli` tags instead of 23 for dind usecases.

## 🚨 User Impact 🚨
* This PR helps us be more granular and wary about the Gradle cache mechanism. I hope it will lead to less transient failures of our CI pipelines.
* It will likely reduce the build time a bit, and keeping all the dependency in a single volume will leads to less disk space use
* The image content of the connector we build is not changing so no impact to expect on the publish pipleine. I'll launch a couple of pre-releases as a sanity check.


tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -255,7 +255,7 @@ def with_dockerd_service(
         docker_lib_volume_name = f"{docker_lib_volume_name}-{slugify(docker_service_name)}"
     dind = (
         context.dagger_client.container()
-        .from_("docker:23.0.1-dind")
+        .from_(consts.DOCKER_DIND_IMAGE)
```
alafanechere, line 258: Using a constant to make sure we use the same docker version everywhere

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -386,34 +384,27 @@ def with_gradle(
     if sources_to_include:
         include += sources_to_include
 
+    gradle_dependency_cache: CacheVolume = context.dagger_client.cache_volume("gradle-dependencies-caching")
```
alafanechere, line 387: This cache is shared with all the java connectors. It's storing the java dependencies.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -386,34 +384,27 @@ def with_gradle(
...
+        .with_env_variable("GRADLE_HOME", "/root/.gradle")
         .with_exec(["mkdir", "/airbyte"])
-        .with_mounted_directory("/airbyte", context.get_repo_dir(".", include=include))
-        .with_mounted_cache("/airbyte/.gradle", airbyte_gradle_cache, sharing=CacheSharingMode.LOCKED)
         .with_workdir("/airbyte")
+        .with_mounted_directory("/airbyte", context.get_repo_dir(".", include=include))
+        .with_exec(["mkdir", "-p", consts.GRADLE_READ_ONLY_DEPENDENCY_CACHE_PATH])
+        .with_mounted_cache(consts.GRADLE_BUILD_CACHE_PATH, gradle_build_cache, sharing=CacheSharingMode.LOCKED)
+        .with_mounted_cache(consts.GRADLE_READ_ONLY_DEPENDENCY_CACHE_PATH, gradle_dependency_cache)
+        .with_env_variable("GRADLE_RO_DEP_CACHE", consts.GRADLE_READ_ONLY_DEPENDENCY_CACHE_PATH)
```
alafanechere, line 406: This tells Gradle that we enable their "read only dependency cache". You can read about it [here](https://docs.gradle.org/current/userguide/dependency_resolution.html#sub:shared-readonly-cache)

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -42,19 +42,6 @@ async def _run(self) -> StepResult:
             return StepResult(self, status=StepStatus.SUCCESS, stdout=f"No manifest found for {self.context.docker_image_name}.")
 
 
-class BuildConnectorForPublish(Step):
```
alafanechere, line 45: Use builds.run_connector_build() instead.

bnchrch (approved):

alafanechere:
Confirmed the publish flow still work: https://github.com/airbytehq/airbyte/actions/runs/5065538005/jobs/9094223334
Initial source-postgres build time: 3:48 mn

PR #26459, opened 2023-05-24
Source: Iterable - read timeout added

grubberr:
## What
Try to improve for this oncall problem https://github.com/airbytehq/oncall/issues/1985#issuecomment-1559276465

The connector just hangs forever after reading 3 days about 150G of data. After 1 hour of silence, Airbyte killed the connector.
I have added a read timeout for 300 seconds to retry in such cases if the connector waits reading too much.

About timeouts: https://requests.readthedocs.io/en/latest/user/advanced/#timeouts


lazebnyi (approved):

bazarnov (approved):

maxi297 (commented):
I'm a little uncertain of why we are adding a timeout. Let me try to summarize my understanding.

The current issue in production right now is a couple of retries that mentions:
Here are some stats:
* 429: 178 occurrences
* 500: 12 occurrences
* 502 Bad Gateway: 516 occurrences
* 504 Gateway Time-out: 68 occurrences

Given that we retry enough time on the same request, it can take up to 1260 seconds of wait time before the request works (20 + 40 + 80 + 160 + 320 + 640). Given all those occurrences we retry only for 20 seconds, it's already > 4 hours in wait time which might explain why the sync is so long (this is without the wait time of having a response from this server which does not seem optimal). I'm still not so sure what is the expectation with adding a timeout to our requests: Do we know how much time a requests takes in average? P(95)? Do we often hit 60 seconds connect and 300 seconds read? How will timeouting improve the sync success rate?

All that to say, I don't think this will address the root cause. There seems to be two issues:
* We send too many requests. The API doc mentions `10 requests/second, per project` on /api/campaigns/metrics and `5 requests/minute` on /api/lists/getUsers. Is it worth having a mechanism other than retrying? Personally, I don't think this is our biggest problem right now so maybe not
* The API is clearly having some kind of issue which we don't understand. Have we contacted Iterable to understand their view on this? The 500 HTTP status returns a body saying that we should contact them (or whatever the CSM is): `{"msg":"An error occurred. Please try again later. If problem persists, please contact your CSM","code":"GenericError","params":null}`

grubberr:
@maxi297 

I suspect that connector just hangs forever [here](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/streams/http/http.py#L308) looking at this https://github.com/airbytehq/oncall/issues/1985#issuecomment-1559276465
I only suspect this I don't have 100% proofs.

With the help of timeout, I just try to raise the `ReadTimeout` exception and repeat it again if we hang more than 5 minutes.

grubberr:
@maxi297 

> How will timeout improve the sync success rate?

it will not, I just tried to fix the hanging problem.

> Do we know how much time a request takes in average?

no, I just assume that we get a response less than 5 minutes

> Have we contacted Iterable to understand their view on this?

as I know - no, we just assume that that problem may be some database inconsistency on Iterable database

darynaishchenko:
> 

HI, @maxi297, I'm working on it now. I guess getting more information is a good idea, I will add this handling in this PR, because time of sync is to long and we need to decrease it. Maybe adding read timeout will not help but we need to try.

darynaishchenko:
> > 
> 
> HI, @maxi297, I'm working on it now. I guess getting more information is a good idea, I will add this handling in this PR, because time of sync is to long and we need to decrease it. Maybe adding read timeout will not help but we need to try.

@maxi297 a little update about signal handling. We can't handle signals like Kill and Stop [docs](https://docs.python.org/3/library/signal.html#signal.SIGKILL). I have tested it locally  with cancel button and catching all available signals except kill and stop, and I didn't get any signal in my signal listener. Looks like that mean that platform killing  processes with sending SIGKILL or SIGSTOP which we can not handle.

girarda (approved):
code looks fine to me 🚢 . Do we know that we're stuck reading from a socket or is this a guess?

darynaishchenko:
> code looks fine to me ship . Do we know that we're stuck reading from a socket or is this a guess?

@girarda Reading from socket is one of the reasons https://github.com/airbytehq/oncall/issues/1985#issuecomment-1570117261

PR #26460, opened 2023-05-24
add memory manager to state manager

cgardens:
## What
* Makes the State Manager memory aware.

### DIFF
--- a/airbyte-integrations/bases/base-java/src/test/java/io/airbyte/integrations/destination_async/state/AsyncStateManagerTest.java
+++ b/airbyte-integrations/bases/base-java/src/test/java/io/airbyte/integrations/destination_async/state/AsyncStateManagerTest.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+ */
+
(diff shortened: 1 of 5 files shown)


airbyte-integrations/bases/base-java/src/test/java/io/airbyte/integrations/destination_async/state/AsyncStateManagerTest.java
```diff
@@ -25,10 +27,12 @@
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import org.apache.commons.lang.RandomStringUtils;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 class AsyncStateManagerTest {
```
cgardens, line 33: made it so this class compiles, but didn't actually try to get the tests to run since @davinchia is working on them currently.

cgardens: we will want to add some tests specific to the memory mgmt.

PR #26473, opened 2023-05-24
fix(cdc): limit queue size to lower memory consumption

subodh1810:
Issue : https://github.com/airbytehq/oncall/issues/2080

Without a heapdump debugging OOM is very difficult but my guess is that the queue (both Airbyte's and Debezium's internal) is growing too big. We can limit debezium's internal queue by using the property (https://debezium.io/documentation/reference/2.2/connectors/postgresql.html#postgresql-property-max-queue-size-in-bytes). I am setting it to 256MB. Next to limit our own queue am introducing a parameter in the spec.I did think about making our own queue memory bound as well but it's more work and makes things complicated. I want to try this PR out as an immediate fix and see how it goes and then if needed we can make our internal queue memory bound as well for further improvements.


airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumPropertiesManager.java
```diff
@@ -22,6 +22,7 @@
 public class DebeziumPropertiesManager {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(DebeziumPropertiesManager.class);
+  public static final String BYTE_VALUE_256_MB = "268435456";
```
edgao, line 25: readability nit

```suggestion
  public static final String BYTE_VALUE_256_MB = Integer.toString(256 * 1024 * 1024);
```

edgao (approved):
seems reasonable, just a couple comments to help future people

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumPropertiesManager.java
```diff
@@ -89,10 +93,13 @@ public Properties getDebeziumProperties() {
     // By default "decimal.handing.mode=precise" which's caused returning this value as a binary.
     // The "double" type may cause a loss of precision, so set Debezium's config to store it as a String
     // explicitly in its Kafka messages for more details see:
-    // https://debezium.io/documentation/reference/2.1/connectors/postgresql.html#postgresql-decimal-types
+    // https://debezium.io/documentation/reference/2.2/connectors/postgresql.html#postgresql-decimal-types
     // https://debezium.io/documentation/faq/#how_to_retrieve_decimal_field_from_binary_representation
     props.setProperty("decimal.handling.mode", "string");
 
+    // https://debezium.io/documentation/reference/2.2/connectors/postgresql.html#postgresql-property-max-queue-size-in-bytes
+    props.setProperty("max.queue.size.in.bytes", BYTE_VALUE_256_MB);
```
prateekmukhedkar, line 101: `max.queue.size.in.bytes` is only applicable to the Postgres Debezium connector. Is it ok to set this property in this class because it appears to be set for all CDC source connectors?

subodh1810: Oh no its applicable for all the 3 connectors. I confirmed in the docs that's why decided to add it in the common properties section

prateekmukhedkar (approved):

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/DebeziumPropertiesManager.java
tom-haines: @prateekmukhedkar @subodh1810 fyi this indirectly exposes a downstream debezium bug re approximate queue size calculator.  see [issues-28069](https://github.com/airbytehq/airbyte/issues/28069#issuecomment-1750030783) for more info.  The downstream bug is fixed from debezium 2.4.0+.

sh4sh:
re https://github.com/airbytehq/airbyte/issues/28069#issuecomment-1750030783 it appears that source-mssql may require a Debezium upgrade (2.2.0 => 2.4.0) for compatibility with this change

PR #26474, opened 2023-05-24
Low code CDK: Allow nested objects for request_body_json

### LINKED ISSUE #20734: [Low-code CDK] Support interpolated values in request_body_json
## Context
When a JSON blob with deeply-nested fields is passed into the `request_body_json` field, this blob is directly decoded into an object, which means interpolated values are not resolved.

Users may want to use interpolated strings anywhere in the JSON blob, so this should be supported by the CDK.

One method of achieving this is to convert the JSON blob to a string, resolve the interpolated values, then convert back to JSON, before decoding the JSON object.

## Acceptance Criteria
- The `request_body_json` field supports interpolated values at any level of nesting in JSON blobs

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/20734

by allowing nested objects when specifying `request_body_json`:
```
request_body_json:
          sort:
              field: "updated_at"
              order: "{{ config.order }}"
```

will result in the following body (assuming a config object `{"order":"ASCENDING"}`):
```
{
  "sort": {
              "field": "updated_at"
              "order": "ASCENDING"
  }
}
```

allowing interpolation at every nesting level and arrays at any level except for the object root level.

Previously, the schema only allowed strings and a map of strings.

Now strings as well as arbitrary deep nested objects are allowed. Request options coming from different components (pagination, incremental sync, partition routing) can still be merged with the defined `request_body_json`). If the jinja expression returns a numeric or boolean data type, it is turned into the respective type in the final JSON as well.

The previously existing behavior of interpreting the provided string as JSON is kept for backwards compatibility (I don't see it used anywhere it existing connectors though).

## How

Some smaller decisions I made on this PR that are not specified in the original issue:
* Do not add this functionality to `request_body_data`, but to `request_body_json`. As `request_body_data` is already well-specified as being used for sending URL-encoded data if it specifies a dict, we can't use it for JSON payloads (also this would be very confusing)
* Do not change the `RequestInput` type to allow all kinds of input for all request options as it doesn't make sense semantically to have nested headers or body data - implement this as a new type and provider for nested input specifically to avoid unintentional behavior changes
* Do not allow arrays - as it's not possible to inject keys into an array structure in JSON, the top level of the request body needs to be an object because otherwise request options from different components can't be merged together. This rules out certain cases like https://github.com/airbytehq/airbyte/blob/903d34e5f16430a86a1cebafd51b2dc821595148/airbyte-integrations/connectors/source-senseforce/source_senseforce/manifest.yaml#L13 , but those can still resort to using a `request_body_data` string with a custom header to signify JSON payload. The vast majority of APIs works with objects at the root of the json body, for example the Airbyte config API. IMHO keeping request_body_json always "injectable" outweighs the benefit of catering to this special case as part of request_body_json if there exists a workaround.

## 🚨 User Impact 🚨

No breaking changes, as the behavior for `request_body_json` as a string doesn't change and is only extended for the case if the values of the root level dictionary are not strings themselves, but arrays, objects or primitive values.


airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -1127,12 +1127,11 @@ definitions:
               }, "orderBy": 1, "columnName": "Timestamp"}]/
       request_body_json:
         title: Request Body JSON Payload
-        description: Specifies how to populate the body of the request with a JSON payload.
+        description: Specifies how to populate the body of the request with a JSON payload. Can contain nested objects
```
girarda, line 1130:
```suggestion
        description: Specifies how to populate the body of the request with a JSON payload. Can contain nested objects.
```

airbyte-cdk/python/airbyte_cdk/sources/declarative/interpolation/interpolated_nested_mapping.py
```diff
@@ -0,0 +1,45 @@
...
+    def __post_init__(self, parameters: Optional[Mapping[str, Any]]):
+        self._interpolation = JinjaInterpolation()
+        self._parameters = parameters
+
+    def eval(self, config: Config, **additional_parameters):
+        return self._eval(self.mapping, config, **additional_parameters)
+
+    def _eval(self, value, config, **kwargs):
+        # Recursively interpolate dictionaries and lists
+        if type(value) == str:
```
girarda, line 37: nit: `isinstance` is generally a better pattern to use that comparing the type with equality to support inheritance https://stackoverflow.com/a/1549854

airbyte-cdk/python/airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -1094,12 +1094,13 @@ class HttpRequester(BaseModel):
         ],
         title="Request Body Payload (Non-JSON)",
     )
-    request_body_json: Optional[Union[str, Dict[str, str]]] = Field(
+    request_body_json: Optional[Union[str, Dict[str, Any]]] = Field(
         None,
-        description="Specifies how to populate the body of the request with a JSON payload.",
+        description="Specifies how to populate the body of the request with a JSON payload. Can contain nested objects",
```
girarda, line 1099:
```suggestion
        description="Specifies how to populate the body of the request with a JSON payload. Can contain nested objects.",
```

airbyte-cdk/python/unit_tests/sources/declarative/requesters/request_options/test_interpolated_request_options_provider.py
```diff
@@ -59,6 +59,9 @@ def test_interpolated_request_params(test_name, input_request_params, expected_r
         ("test_number_falsy_value", {"number_falsy": "{{ 0.0 }}"}, {"number_falsy": 0.0}),
         ("test_string_falsy_value", {"string_falsy": "{{ '' }}"}, {}),
         ("test_none_value", {"none_value": "{{ None }}"}, {}),
+        ("test_string", """{"nested": { "key": "{{ config['option'] }}" }}""", {"nested": {"key": "OPTION"}}),
```
girarda, line 62: nit: can you add a test case with a list and another one with a number to cover the branches in https://github.com/airbytehq/airbyte/pull/26474/files#diff-e94b83408916b273194cf715b73cb227bee59ecc69bb2f8148a3cc62a7162952R42-R43 and https://github.com/airbytehq/airbyte/pull/26474/files#diff-e94b83408916b273194cf715b73cb227bee59ecc69bb2f8148a3cc62a7162952R44-R45

girarda (approved):
lgtm! comments are non-blocking. Is there an existing connector using request_body_json we can run the CATs on to smoke test the change before merging?

girarda (approved):
lgtm! comments are non-blocking. Is there an existing connector using request_body_json we can run the CATs on to smoke test the change before merging?

flash1293:
The run failed, but most tests passed and the ones that failed failed in the exact same way as in the nightly run: https://github.com/airbytehq/airbyte/actions/runs/5075651770/jobs/9117112283

flash1293:
Seems like intercom worked in the nightly run, running that test to be sure

flash1293:
Intercom is still on low code pre 0.29.0, so it doesn't work due to our BWC check... Updating the version for testing

flash1293:
Looks like there is actually a bug in there - trying the fix

flash1293:
Seems like the same error happens as well when running the intercom connector whit the current master state of the CDK: https://github.com/airbytehq/airbyte/pull/26577 - looks like it's an unrelated error

@girarda could you take another look just to be sure this is safe to merge?

airbyte-cdk/python/unit_tests/sources/declarative/requesters/request_options/test_interpolated_request_options_provider.py
girarda: doesn't the union `NestedMappingEntry` include numbers and lists? https://github.com/airbytehq/airbyte/pull/26474/files#diff-e94b83408916b273194cf715b73cb227bee59ecc69bb2f8148a3cc62a7162952R12

flash1293: The entries yes, but the top level “mapping” parameter that needs to be passed in doesn’t

flash1293:
Tested in atelier server via https://github.com/airbytehq/airbyte-platform-internal/blob/master/oss/airbyte-connector-atelier-server/README.md?plain=1#L18-L27

and regular patterns still work while allowing nested objects as well.

PR #26485, opened 2023-05-24
:bug: Source Square migrate auth spec to advanced auth

davydov-d:
## What
Remove deprecated authSpecification from connector spec in favour of advancedAuth
https://github.com/airbytehq/airbyte/issues/26247
+fix build

## How
Replace `authSpecification` with `advancedAuth`

Pay attention: this PR could not be manually tested on localhost because the Square OAuth API does not accept a redirect URL, it is hardcoded in our Square account as `cloud.airbyte.io`


bazarnov (approved):
LGTM, once the CAT passes.

lazebnyi (approved):

airbyte-integrations/connectors/source-square/acceptance-test-config.yml
```diff
@@ -42,6 +42,11 @@ acceptance_tests:
             bypass_reason: "Floating data"
           - name: updated_at
             bypass_reason: "Floating data"
+        categories:
```
maxi297, line 45: How is this related to the oauth change?

airbyte-integrations/connectors/source-square/source_square/spec.yaml
```diff
@@ -75,17 +75,6 @@ connectionSpecification:
       title: Include Deleted Objects
       default: false
       order: 3
-authSpecification:
```
maxi297, line 78: I have the same question as [this other PR](https://github.com/airbytehq/airbyte/pull/26358/files#r1204215234)

maxi297 (approved):
Conditional approval on the confirmation that this isn't breaking change

airbyte-integrations/connectors/source-square/acceptance-test-config.yml
davydov-d: This was needed to fix buid, not actually related to oauth changes

PR #26491, opened 2023-05-24
Source Qualaroo: drop auth spec

davydov-d:
## What
https://github.com/airbytehq/airbyte/issues/26244
Remove authSpec from spec.json because OAuth is not supported at all by the Qualaroo API + update stream 
schema to make build green again


bazarnov (approved):
LGTM, non quite sure why it was added before, but I've tested OAuth for this connector on Prod - and it's not working.

lazebnyi (approved):

PR #26507, opened 2023-05-24
Docs: mention custom objects in hubspot docs

sherifnada:


marcosmarxm (approved):
Thanks @sherifnada I wasn't aware of the need of this message and the additional steps.
To clarify any new Hubspot stream which needs additional scope we must add this info?

sherifnada:
@marcosmarxm yes if we add a new stream which requires new oauth scopes the user needs to take those steps

nataliekwong:
Note if the user is not using OAuth and instead a private key, they'll need to instead go into Hubspot and update the scopes the key has access to instead (rather than reauthenticating OAuth app permissions).

sherifnada:
@nataliekwong good call, fixed that plus other things in https://github.com/airbytehq/airbyte/pull/26533

PR #26526, opened 2023-05-25
fix flush worker close

cgardens:
## What
* We were dropping records on close, because flush all triggered one flush per stream. The issue was that if there was more then a single batch of records in any one stream's queue, records above the batch size didn't get sent. 😵 
* Reworks the flush worker to wait until all queues are empty.

## How
* It removes `flushAll`, this concept felt like a handbrake, but it didn't actually make any sense.
* On `close` the flusher worker just waits for the normal process of flushing to empty the buffers. The only thing that changes it the threshold for uploading records is reduced from a non-zero threshold to 0. In other words, any queue with more than 0 records, is going to get picked up to be flushed.

## Recommended reading order
1. `FlushWorkers.java`
2. `FlushWorkersTest.java`


airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/AsyncStreamConsumer.java
```diff
@@ -140,7 +137,10 @@ private void propagateFlushWorkerExceptionIfPresent() throws Exception {
     }
   }
 
-  private void validateRecord(final AirbyteMessage message, final StreamDescriptor streamDescriptor) {
+  private void validateRecord(final AirbyteMessage message) {
```
cgardens, line 140: fixing previous merge conflict resolution issue.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/FlushWorkers.java
```diff
@@ -53,93 +54,145 @@ public class FlushWorkers implements AutoCloseable {
...
+          } else {
+            streamToInProgressWorkers.put(desc, new AtomicInteger(1));
+          }
+          allocatableThreads--;
+          flush(desc);
+        } else {
+          break;
+        }
+      }
+    } catch (final Exception e) {
```
cgardens, line 164: realized that we need to catch and propagate exceptions here too, because it also runs in a thread. i was confused for the longest time because there was an NPE, but it failed silently until I propagated the exception.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/buffers/MemoryAwareMessageBatch.java
```diff
@@ -48,8 +48,6 @@ public List<MessageWithMeta> getData() {
 
   @Override
   public void close() throws Exception {
-    batch = null;
```
davinchia, line 51: was this just left over lines?

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/FlushWorkers.java
davinchia: you are right. I was focused on the thread pool and forgot about the scheduler thread - good catch

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/FlushWorkers.java
```diff
@@ -53,93 +54,132 @@ public class FlushWorkers implements AutoCloseable {
...
 
   private void retrieveWork() {
-    // todo (cgardens) - i'm not convinced this makes sense. as we get close to the limit, we should
-    // flush more eagerly, but "flush all" is never a particularly useful thing in this world.
-    // if the total size is > n, flush all buffers
-    if (bufferDequeue.getTotalGlobalQueueSizeBytes() > BufferManager.TOTAL_QUEUES_MAX_SIZE_LIMIT_BYTES) {
-      flushAll();
-      return;
+    try {
+      log.info("retrieving queues to process");
```
davinchia, line 128:
```suggestion
      log.debug("retrieving queues to process");
```

davinchia: might get noisy

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/FlushWorkers.java
```diff
@@ -53,93 +54,132 @@ public class FlushWorkers implements AutoCloseable {
...
     }
+  }
 
-    final ThreadPoolExecutor threadPoolExecutor = (ThreadPoolExecutor) workerPool;
-    var allocatableThreads = threadPoolExecutor.getMaximumPoolSize() - threadPoolExecutor.getActiveCount();
-
-    // todo (cgardens) - build a score to prioritize which queue to flush next. e.g. if a queue is very
-    // large, flush it first. if a queue has not been flushed in a while, flush it next.
-    // otherwise, if each individual stream has crossed a specific threshold, flush
+  private Optional<StreamDescriptor> getNext(final long queueSizeThresholdBytes) {
```
davinchia, line 158:
```suggestion
  private Optional<StreamDescriptor> getQueueToFlush(final long queueSizeThresholdBytes) {
```

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/FlushWorkers.java
```diff
@@ -53,93 +54,132 @@ public class FlushWorkers implements AutoCloseable {
...
+          if (streamToInProgressWorkers.containsKey(desc)) {
+            streamToInProgressWorkers.get(desc).getAndAdd(1);
+          } else {
+            streamToInProgressWorkers.put(desc, new AtomicInteger(1));
+          }
+          allocatableThreads--;
+          flush(desc);
+        } else {
+          break;
+        }
```
davinchia, line 149:
```suggestion
        if (next.isEmpty()) {
          break;
        }
        final StreamDescriptor desc = next.get();
        if (streamToInProgressWorkers.containsKey(desc)) {
          streamToInProgressWorkers.get(desc).getAndAdd(1);
        } else {
          streamToInProgressWorkers.put(desc, new AtomicInteger(1));
        }
        allocatableThreads--;
        flush(desc);
```

davinchia: suggestion to avoid left hanging braces

airbyte-integrations/bases/base-java/src/test/java/io/airbyte/integrations/destination_async/FlushWorkersTest.java
```diff
@@ -25,15 +25,22 @@ public class FlushWorkersTest {
...
-    final var collector = mock(Consumer.class);
+    when(dequeue.take(desc, 1000)).thenReturn(new MemoryAwareMessageBatch(List.of(), 10, null, null));
+    when(dequeue.getQueueSizeBytes(desc)).thenReturn(Optional.of(10L));
+    when(dequeue.getQueueSizeInRecords(desc)).thenAnswer(ignored -> {
+      if (hasThrownError.get()) {
+        return Optional.of(0L);
+      } else {
+        return Optional.of(1L);
+      }
+    });
```
davinchia, line 40: for understanding: is this trying to force an exception in the supervisor thread?

davinchia (approved):
Thanks for catching the supervisor error bits!

One question for understanding.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/buffers/MemoryAwareMessageBatch.java
cgardens: yeah.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/FlushWorkers.java
cgardens: yeah. we should make a pass at the logs to fix levels, but for right now i'm going to leave it. as we debug what's broken it's helpful.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/FlushWorkers.java
cgardens: cool!

airbyte-integrations/bases/base-java/src/test/java/io/airbyte/integrations/destination_async/FlushWorkersTest.java
cgardens: nope. i changed the behavior in the worker so that it's not allowed to close until all the queues are empty. so now in this test, i can't let the queue go to zero until after we throw, otherwise the workers can close before the exception gets thrown.

Issue #26239, opened 2023-05-18
Source Lever Hiring: Migrate to advancedAuth


lazebnyi:

davydov-d:
https://github.com/airbytehq/airbyte/pull/26564 - PR is ready, but I'm not able to manually test OAuth flow as we have an inconsistency in our creds

PR #26564, opened 2023-05-25
:bug: Source Lever-hiring: migrate to advancedAuth

davydov-d:
## What
https://github.com/airbytehq/airbyte/issues/26239

## How
Delete `authSpecification`

_currently blocked because of creds inconsistency_
UPD: creds issue fixed


lazebnyi (approved):

lazebnyi (approved):

maxi297 (approved):
Conditional approval: https://github.com/airbytehq/airbyte/pull/26358/files#r1204215234

Note that we don't have any cloud user for this which makes it hard for me to validate how the config is in DB. I'm somewhat concerned about the `auth_type` [here](https://github.com/airbytehq/airbyte/pull/26564/files#diff-ef255f6a1a22963f999b497b816f2b15dae3b25db14e0024a90462cd2ff2c2e1R87) since it does not seem to be part of the code that we removed and therefore authentication might break for those who don't have this

PR #26583, opened 2023-05-25
allow running run-ab-platform.sh in detached mode

c-p-b:
## What
Sometimes end users might want to simply start the docker compose process and not retain an open terminal process. This allows them to do that in a single command in the script

## How
Add a flag, -b, that allows end users to run in "background" mode. Unfortunately -d was already taken so we use -b instead.

## Recommended reading order
1. run-ab-platform.sh

## 🚨 User Impact 🚨
No breaking changes

### DIFF
--- a/run-ab-platform.sh
+++ b/run-ab-platform.sh
@@ -169,6 +169,8 @@ if test $? -ne 0; then
 fi
 
 ########## Ending Docker ##########
-docker compose down
-
-echo -e "$blue_text""Stopping Docker Compose""$default_text"
+if [ -z "$dockerDetachedMode" ]; then
+  docker compose down
+else
+  echo -e "$blue_text""Airbyte containers are running!""$default_text"
+fi
--- a/run-ab-platform.sh
+++ b/run-ab-platform.sh
@@ -30,6 +30,7 @@ Help()
    echo -e "   -r --refresh     ${red_text}DELETE${default_text} existing assets and re-download new ones"
    echo -e "   -h --help        Print this Help."
    echo -e "   -x --debug       Verbose mode."
+   echo -e "   -b --background  Run docker compose up in detached mode."
    echo -e ""
 }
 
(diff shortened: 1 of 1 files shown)


run-ab-platform.sh
```diff
@@ -106,6 +108,9 @@ for argument in $@; do
     -x | --debug)
       set -o xtrace  # -x display every line before execution; enables PS4
       ;;
+    -b | --background)
+      dockerDetachedMode="-d"
+      ;;
```
marcosmarxm, line 113: @cpdeethree can we have a flag or command to stop the instance? Because today people only will the `docker-compose down` anyway. Having this we could update the upgrading docs and everything use the script.

run-ab-platform.sh
```diff
@@ -155,7 +160,7 @@ done
 echo
 echo -e "$blue_text""Starting Docker Compose""$default_text"
 
-docker compose up
+docker compose up $dockerDetachedMode
```
josephkmh, line 163: I think we need to conditionally call `docker compose down` on line 172, right? Currently when I run `./run-ab-platform.sh -b` the containers spin up and then are immediately stopped.

josephkmh: @cpdeethree I'm no bash wizard, but I think this should do it: https://github.com/airbytehq/airbyte/pull/26583/commits/19c58f53dbd6ebb9097c5cac73b3e6ec768c2380

joeykmh (approved):
@cpdeethree I pushed a commit that appears to be working, would appreciate a quick check, but otherwise LGTM! 🚢

run-ab-platform.sh
josephkmh: @marcosmarxm I think any command to stop the instance would basically just wrap `docker compose down`. I feel like that is an acceptable command to use, since users should know that they are running docker containers via compose. `run-ab-platform.sh` is useful for starting the instance because it pulls the correct files/images, but it does not really add anything for stopping the instance.

cpdeethree:
> @cpdeethree I pushed a commit that appears to be working, would appreciate a quick check, but otherwise LGTM! 🚢

Looks good!

PR #26633, opened 2023-05-25
Log full vacuum info

### LINKED ISSUE #26509: Query and log results for last full vacuum
Add a query to log vacuum status (both auto vacuum and manual vacuum) and log it.

rodireich:
## What
This change logs before syncing the result of 2 queries that will allow us to get information about whether a full vacuum of the table is currently running or ran.

## How
Performs two queries showing information about currently running full vacuum and also system information that allows gauging whether a full sync ran in between source-postgres attempts (relation filename).
See [tech spec](https://docs.google.com/document/d/1cwVNjX3wl5RY6Kg3wQOx1uvBcdTRKqqLJ_OE6E_Axqg/edit#bookmark=id.ll5tb3yaaabj) for more information.

## 🚨 User Impact 🚨
This is only adding logging.


airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresQueryUtils.java
```diff
@@ -51,6 +54,14 @@ public class PostgresQueryUtils {
             txid_snapshot_xmin(txid_current_snapshot()) AS xmin_raw_value;
       """;
 
+  public static final String CTID_FULL_VACUUM_IN_PROGRESS =
```
akashkulk, line 57: nit : here and below annotate with query

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresQueryUtils.java
```diff
@@ -51,6 +54,14 @@ public class PostgresQueryUtils {
...
+      """
+      SELECT phase FROM pg_stat_progress_cluster WHERE relid=to_regclass('%s')::oid
```
akashkulk, line 59: Quick question : This seems different from the query in the tech spec. What is the difference between the two queries?

rodireich: This filters only for rows related to this stream. I'll add also `AND command = 'VACUUM FULL'`

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/PostgresQueryUtils.java
```diff
@@ -51,6 +54,14 @@ public class PostgresQueryUtils {
             txid_snapshot_xmin(txid_current_snapshot()) AS xmin_raw_value;
       """;
 
+  public static final String CTID_FULL_VACUUM_IN_PROGRESS_QUERY =
+      """
+      SELECT phase FROM pg_stat_progress_cluster WHERE command = 'VACUUM FULL' AND relid=to_regclass('%s')::oid
```
akashkulk, line 59: Seems like pg_stat_progress_cluster was added in Postgres 12? https://pgpedia.info/p/pg_stat_progress_cluster. What's the behavior of this query for versions prior to that but are supported by Airbyte (PG 9.4 - 12)?

rodireich: Great catch! Thank you
What I think will work best in the real sync is if we can gracefully fail if `pg_stat_progress_cluster` function wasn't supported and fallback to standard non-ctid sync.
For the purpose of this logging only call I'll make sure we're showing a warning which doesn't affect the overall sync.
I'm testing with an older pg to understand what JDBC is giving us in that case.

akashkulk: Thanks! Just wanted to make sure it didn't cause a NPE or something

rodireich: Very good point. and another thing to look for when reading using ctid

akashkulk: Another quick thing you can do is to change the docker image we're using here to test the `logPreSyncDebugData` functionality pre-12 : Not blocking, but thought I'd mention :  https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/PostgresSourceTest.java#LL143C46-L143C46

akashkulk (approved):

PR #26644, opened 2023-05-26
connectors-ci: publish strict-encrypt first

### LINKED ISSUE #26635: Build strict-encrypt/secure version first
## Overview
When running the build pipeline for multiple connectors we should make a simple change to ensure our strict-encrypt and secure versions are built and published first.  

This is needed for metadata validation because of overrides

alafanechere:
## What
Closes #26635
 The metadata upload step on publish checks if the the connectors referenced in the metadata file are already published to DockerHub. Non strict-encrypt variant reference the strict-encrypt variant in their metadata file for cloud.
So if we publish the non strict-encrypt variant first, the metadata upload will fail if the strict-encrypt variant is not published yet.
As strict-encrypt variant are often modified in the same PR as the non strict-encrypt variant, we want to publish them first.

## How
Hacky approach: as connector names with -strict-encrypt/secure prefix are longer, they will be sorted first with a reverse sort below.


tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -289,3 +289,16 @@ def create_connector_report(results: List[StepResult]) -> ConnectorReport:
...
+    """Reorder contexts so that the ones that are for strict-encrypt/secure connectors come first.
+    The metadata upload on publish checks if the the connectors referenced in the metadata file are already published to DockerHub.
+    Non strict-encrypt variant reference the strict-encrypt variant in their metadata file for cloud.
+    So if we publish the non strict-encrypt variant first, the metadata upload will fail if the strict-encrypt variant is not published yet.
+    As strict-encrypt variant are often modified in the same PR as the non strict-encrypt variant, we want to publish them first.
+    This is an hacky approach: as connector names with -strict-encrypt/secure prefix are longer,
+    they will be sorted first with our reverse sort below.
+    """
+
+    return sorted(contexts, key=lambda context: context.connector.technical_name, reverse=True)
```
bnchrch, line 304: So I like this!

But, I think concurrency might still bite us

i.e. if we have concurrency set to 5 we might build `source-file` and `source-file-secure` at the same time

Perhaps we want the sort to pull the strict-encrypt variants to the start and their counter parts to the end of the list?

Or at the conncurreny level we explicitly try and build all strict encrypts in one block before anything else?

bnchrch (commented):
One comment

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
alafanechere: On publish concurrency is forced to 1 for safety/stability.

bnchrch (approved):

PR #26651, opened 2023-05-26
Source Amazon Ads: fix version in metadata.yaml

roman-yermilov-gl:
## What
*Connector's version in docker image tag and metadata.yaml do not match.*

## How
*Upgrade version in metadata.yaml*


lazebnyi (commented):
It's look like version tag already changed - https://github.com/airbytehq/airbyte/pull/26657

PR #26656, opened 2023-05-26
:bug: Source Instagram: migrate to advanced auth

davydov-d:
## What
#26237 

## How
Replace `authSpecification` in spec with `advancedAuth`

_this change can not be manually tested on localhost because the Oauth app does not contain `localhost` in allowed redirect URLs_


bazarnov (approved):

lazebnyi (approved):
LGTM! But need tested with pre-release tag.

PR #26675, opened 2023-05-26
Source Twilio: fix expected records and docker image tag

roman-yermilov-gl:
## What
*`executions` steram became empty for some reason. After some manipulations a reocrd become present in response. There is also new field `subresource_uris` in incoming_phone_numbers stream. Docker image tag was not updated.*

## How
*Made new execution record, updated expected records and schema for incoming_phone_numbers stream. Updated docker image tag*


lazebnyi (approved):
Please add link to failed action to PR description.

PR #26677, opened 2023-05-26
Multi-architecture normalization build (local)

evantahler:
When building and testing normalization locally, we need to force the base images to match the local host OS.

This is not a problem when publishing the connectors as `airbyte-ci`/dagger handles this for us

### DIFF
--- a/airbyte-integrations/bases/base-normalization/build.gradle
+++ b/airbyte-integrations/bases/base-normalization/build.gradle
@@ -54,11 +54,12 @@ def buildAirbyteDocker(String customConnector) {
     // We are lucky that all the python code dbt uses is mutli-arch compatible
 
     def arch = 'linux/amd64'
-    if (Os.isArch("aarch_64")) {
+    if (Os.isArch("aarch_64") || Os.isArch("aarch64")) {
         arch = 'linux/arm64'
     }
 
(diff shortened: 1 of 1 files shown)


edgao (approved):
nice. this has no impact on prod, right? because `/publish` doesn't go through this path anymore

evantahler:
> nice. this has no impact on prod, right? because `/publish` doesn't go through this path anymore

I think this should be OK for 2 reasons:
* we build and test on AMD machines, so there shouldn't be a change to the images we produce
* yep, `manage.sh` is what we use for publishing (not the new pipelines yet), and that rebuilds the normalization containers in different architectures anyway https://github.com/airbytehq/airbyte/blob/master/tools/integrations/manage.sh#L237-L243

davinchia:
@evantahler was this not detecting the local arch before this change?

evantahler:
no... because the [base image we use is so old](https://hub.docker.com/r/fishtownanalytics/dbt/tags), they only ever published AMD images... so any image built on it remains AMD unless you do something drastic... like this!  Anyway, all of this is going away soon so 🤷

evantahler:
You can update the changes if you want to look at the ENV `DOCKER_BUILD_PLATFORM` first, and if that's missing, use the local arch, that would be fine

PR #26535, opened 2023-05-25
:bug: Source Notion: migrated to advancedAuth

darynaishchenko:
## What
https://github.com/airbytehq/airbyte/issues/26242

## How
- Replace deprecated `authSpecification` with `advancedAuth` in the connector spec

PR #26683, opened 2023-05-26
CAT: Ensure we only generate iso8601 with hypothesis in backwards compatibility test

bnchrch:
## What
Ok so we have one GLOBAL problem of our formats do not match ALL valid ISO dates.

But lets ignore that.

The subproblem here is backwards compatibility tests are failing for the following reasons
1. Hypothesis generates date-times that represent all valid RFC3339 values (an ISO subset) (screenshot 1)
2. Our `format` for `date-time` in our specs are very strict about having 3 miliseconds (screenshot 2)

This causes failing backwards compatibility tests ([slack thread](https://airbytehq-team.slack.com/archives/C02UF50V9HA/p1685112086208119))

image
image

## Solution
Only generate date-times that represent our valid version

blocks https://github.com/airbytehq/airbyte/pull/26535

### DIFF
--- a/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/utils/backward_compatibility.py
+++ b/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/utils/backward_compatibility.py
@@ -5,15 +5,14 @@
 from abc import ABC, abstractmethod
 from deepdiff import DeepDiff
 from enum import Enum
-import datetime
 import jsonschema
 
-from hypothesis import HealthCheck, Verbosity, given, settings, strategies as st
+from hypothesis import HealthCheck, Verbosity, given, settings
 from hypothesis_jsonschema import from_schema
 
 from airbyte_cdk.models import ConnectorSpecification
 from connector_acceptance_test.utils import SecretDict
-from typing import Dict, Any, Tuple
+from typing import Dict, Any
 
 
 class BackwardIncompatibilityContext(Enum):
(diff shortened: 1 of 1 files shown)


bnchrch:
@evantahler moving our slack convo here
image


TLDR: That was a good callout that had me reevaluate my approach to this test

The issue here is essentially `hypothesis-jsonschema` can generate properties that define `format: date-time` and `pattern: <REGEX>` but not both.

In the case of both it blows up because many of the date-time strings do not match our pattern regex's as they are often very strict 

e.g.  "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}.[0-9]{3}Z$",

I think our best solution to this is in the case of `format: date-time` + `pattern` is to
1. Only generate the value based on the regex pattern
2. Rely on the jsonschema validation that uses both the new schema and a different jsonschema library to validate that it is indeed a date-time field..

How we do that is by removing `format: date-time` anywhere its defined alongside a `pattern` from the previous schema before using it to generate prop tests. This forces `hypothesis` to use the regex pattern as its basis for generating values.

sherifnada (approved):
makes sense to me. Approving since this is blocking a roadmap item for GL

davydov-d:
hey @bnchrch it looks like the problem still persists, could you please take a look? https://github.com/airbytehq/airbyte/actions/runs/5266559038/jobs/9520679064

bnchrch:
Just closing the loop here as my reply yesterday was on the latest PR
https://github.com/airbytehq/airbyte/pull/26535#issuecomment-1592179829

Essentially it looks like the problem is resolved!

PR #26700, opened 2023-05-27
Serialized Message Consumer

cgardens:
## What
* As part of the async version of the Snowflake Destination, instead of passing the already deserialized messages to the message consumer, we need to pass the raw string (and its size in bytes).
* This PR changes adds a new consumer to the `Destination` interface that does just (`SerializedAirbyteMessageConsumer`). It is default to just delegate to the old `AirbyteMessageConsumer`. This makes it backwards compatible and a no op for existing Destinations. The `IntegrationRunner` to calls the interface method on the Destination. 
* I have left the code (but commented out) in `SnowflakeDestination.java` for how we would turn on the async version of the Snowflake Destination. As we roll it out with the feature flag, those 3 lines of code will all that needs to be maintained on the feature branch.

## Recommended reading order
1. `airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java`
2. `airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/SerializedAirbyteMessageConsumer.java`
3. `airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java`
4. the rest..

## 🚨 User Impact 🚨
No! That's the whole point of this PR it's a no op from the user's point of view, but sets up the async snowflake destination.

### DIFF
--- a/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
+++ b/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
@@ -4,11 +4,17 @@
 
 package io.airbyte.integrations.base;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.annotations.VisibleForTesting;
 import io.airbyte.commons.json.Jsons;
 import io.airbyte.protocol.models.v0.AirbyteMessage;
+import io.airbyte.protocol.models.v0.AirbyteMessage.Type;
 import io.airbyte.protocol.models.v0.ConfiguredAirbyteCatalog;
+import java.util.Optional;
 import java.util.function.Consumer;
+import lombok.extern.slf4j.Slf4j;
 
 public interface Destination extends Integration {
 
@@ -29,8 +35,126 @@ public interface Destination extends Integration {
                                      Consumer<AirbyteMessage> outputRecordCollector)
       throws Exception;
 
+  /**
+   * Default implementation allows us to not have to touch existing destinations while avoiding a lot
+   * of conditional statements in {@link IntegrationRunner}.
+   *
+   * @param config config
+   * @param catalog catalog
+   * @param outputRecordCollector outputRecordCollector
+   * @return AirbyteMessageConsumer wrapped in SerializedAirbyteMessageConsumer to maintain legacy
+   *         behavior.
+   * @throws Exception exception
+   */
+  default SerializedAirbyteMessageConsumer getSerializedMessageConsumer(final JsonNode config,
+                                                                        final ConfiguredAirbyteCatalog catalog,
+                                                                        final Consumer<AirbyteMessage> outputRecordCollector)
+      throws Exception {
+    return new ShimToSerializedAirbyteMessageConsumer(getConsumer(config, catalog, outputRecordCollector));
+  }
+
   static void defaultOutputRecordCollector(final AirbyteMessage message) {
     System.out.println(Jsons.serialize(message));
+  }
+
+  @Slf4j
+  class ShimToSerializedAirbyteMessageConsumer implements SerializedAirbyteMessageConsumer {
+
+    private final AirbyteMessageConsumer consumer;
+
+    public ShimToSerializedAirbyteMessageConsumer(final AirbyteMessageConsumer consumer) {
+      this.consumer = consumer;
+    }
+
+    @Override
+    public void start() throws Exception {
+      consumer.start();
+    }
+
+    /**
+     * Consumes an {@link AirbyteMessage} for processing.
+     * <p>
+     * If the provided JSON string is invalid AND represents a {@link AirbyteMessage.Type#STATE}
+     * message, processing is halted. Otherwise, the invalid message is logged and execution continues.
+     *
+     * @param inputString JSON representation of an {@link AirbyteMessage}.
+     * @throws Exception if an invalid state message is provided or the consumer is unable to accept the
+     *         provided message.
+     */
+    @Override
+    public void accept(final String inputString, final Integer sizeInBytes) throws Exception {
+      consumeMessage(consumer, inputString);
+    }
+
+    @Override
+    public void close() throws Exception {
+      consumer.close();
+    }
+
+    /**
+     * Consumes an {@link AirbyteMessage} for processing.
+     * <p>
+     * If the provided JSON string is invalid AND represents a {@link AirbyteMessage.Type#STATE}
+     * message, processing is halted. Otherwise, the invalid message is logged and execution continues.
+     *
+     * @param consumer An {@link AirbyteMessageConsumer} that can handle the provided message.
+     * @param inputString JSON representation of an {@link AirbyteMessage}.
+     * @throws Exception if an invalid state message is provided or the consumer is unable to accept the
+     *         provided message.
+     */
+    @VisibleForTesting
+    static void consumeMessage(final AirbyteMessageConsumer consumer, final String inputString) throws Exception {
+
+      final Optional<AirbyteMessage> messageOptional = Jsons.tryDeserialize(inputString, AirbyteMessage.class);
+      if (messageOptional.isPresent()) {
+        consumer.accept(messageOptional.get());
+      } else {
+        if (isStateMessage(inputString)) {
+          throw new IllegalStateException("Invalid state message: " + inputString);
+        } else {
+          log.error("Received invalid message: " + inputString);
+        }
+      }
+    }
+
+    /**
+     * Tests whether the provided JSON string represents a state message.
+     *
+     * @param input a JSON string that represents an {@link AirbyteMessage}.
+     * @return {@code true} if the message is a state message, {@code false} otherwise.
+     */
+    @SuppressWarnings("OptionalIsPresent")
+    private static boolean isStateMessage(final String input) {
+      final Optional<AirbyteTypeMessage> deserialized = Jsons.tryDeserialize(input, AirbyteTypeMessage.class);
+      if (deserialized.isPresent()) {
+        return deserialized.get().getType() == Type.STATE;
+      } else {
+        return false;
+      }
+    }
+
+    /**
+     * Custom class that can be used to parse a JSON message to determine the type of the represented
+     * {@link AirbyteMessage}.
+     */
+    private static class AirbyteTypeMessage {
+
+      @JsonProperty("type")
+      @JsonPropertyDescription("Message type")
+      private AirbyteMessage.Type type;
+
+      @JsonProperty("type")
+      public AirbyteMessage.Type getType() {
+        return type;
+      }
+
+      @JsonProperty("type")
+      public void setType(final AirbyteMessage.Type type) {
+        this.type = type;
+      }
+
+    }
+
   }
 
 }
--- a/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
+++ b/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
@@ -57,6 +57,9 @@ public interface Destination extends Integration {
     System.out.println(Jsons.serialize(message));
   }
 
+  /**
+   * Backwards-compatibility wrapper for an AirbyteMessageConsumer. Strips the sizeInBytes argument away from the .accept call.
+   */
   @Slf4j
   class ShimToSerializedAirbyteMessageConsumer implements SerializedAirbyteMessageConsumer {
 
--- a/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
+++ b/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
@@ -137,8 +137,8 @@ public interface Destination extends Integration {
     }
 
     /**
-     * Custom class that can be used to parse a JSON message to determine the type of the represented
-     * {@link AirbyteMessage}.
+     * Custom class for parsing a JSON message to determine the type of the represented
+     * {@link AirbyteMessage}. Do the bare minimum deserialisation by reading only the type field.
      */
     private static class AirbyteTypeMessage {
 
--- a/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
+++ b/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
@@ -58,7 +58,8 @@ public interface Destination extends Integration {
   }
 
   /**
-   * Backwards-compatibility wrapper for an AirbyteMessageConsumer. Strips the sizeInBytes argument away from the .accept call.
+   * Backwards-compatibility wrapper for an AirbyteMessageConsumer. Strips the sizeInBytes argument
+   * away from the .accept call.
    */
   @Slf4j
   class ShimToSerializedAirbyteMessageConsumer implements SerializedAirbyteMessageConsumer {
--- a/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
+++ b/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
@@ -20,8 +20,8 @@ import io.airbyte.protocol.models.v0.AirbyteMessage;
 import io.airbyte.protocol.models.v0.AirbyteMessage.Type;
 import io.airbyte.protocol.models.v0.ConfiguredAirbyteCatalog;
 import io.airbyte.validation.json.JsonSchemaValidator;
-import java.io.BufferedInputStream;
-import java.io.ByteArrayOutputStream;
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.List;
@@ -31,6 +31,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+import org.apache.commons.io.input.CountingInputStream;
 import org.apache.commons.lang3.ThreadUtils;
 import org.apache.commons.lang3.concurrent.BasicThreadFactory;
 import org.slf4j.Logger;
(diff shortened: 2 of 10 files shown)


airbyte-integrations/connectors/destination-snowflake/src/main/java/io/airbyte/integrations/destination/snowflake/SnowflakeDestination.java
```diff
@@ -21,6 +29,27 @@ enum DestinationType {
...
+  public SerializedAirbyteMessageConsumer getSerializedMessageConsumer(final JsonNode config,
+                                                                       final ConfiguredAirbyteCatalog catalog,
+                                                                       final Consumer<AirbyteMessage> outputRecordCollector)
+      throws Exception {
+    log.info("destination class: {}", getClass());
+    // detect if running on internal staging for snowflake, if so run consumer2.
+    final boolean useAsyncSnowflake = false;
+    // final boolean useAsyncSnowflake = config.has("loading_method")
+    // && config.get("loading_method").has("method")
+    // && config.get("loading_method").get("method").asText().equals("Internal Staging");
```
cgardens, line 45: uncommenting these lines of code will all we will need to do to "turn on" the async version of the destination. as we do a feature flagged roll out, this will be the only diff in that feature branch.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/SerializedAirbyteMessageConsumer.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Consumes all {@link AirbyteMessage}s
+   *
+   * @param message {@link AirbyteMessage} as a string
+   * @param sizeInBytes size of that string in bytes
+   * @throws Exception exception
+   */
+  @Override
+  void accept(String message, Integer sizeInBytes) throws Exception;
```
davinchia, line 46: thought: should `sizeInBytes` be a long instead of an int? although seems very unlikely we'd have a record > 2.3GB, not sure we want to bake this assumption into the interface.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
```diff
@@ -29,8 +35,126 @@ AirbyteMessageConsumer getConsumer(JsonNode config,
...
+    @Override
+    public void start() throws Exception {
+      consumer.start();
+    }
+
+    /**
+     * Consumes an {@link AirbyteMessage} for processing.
+     * <p>
+     * If the provided JSON string is invalid AND represents a {@link AirbyteMessage.Type#STATE}
+     * message, processing is halted. Otherwise, the invalid message is logged and execution continues.
```
davinchia, line 78: nit: probably useful to add more details why i.e. STATE messages represent application state and we want to fail fast to ensure data correctness.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
```diff
@@ -29,8 +35,126 @@ AirbyteMessageConsumer getConsumer(JsonNode config,
...
+        return deserialized.get().getType() == Type.STATE;
+      } else {
+        return false;
+      }
+    }
+
+    /**
+     * Custom class that can be used to parse a JSON message to determine the type of the represented
+     * {@link AirbyteMessage}.
+     */
```
davinchia, line 139:
```suggestion
    /**
     * Custom class for parsing a JSON message to determine the type of the represented
     * {@link AirbyteMessage}. Do the bare minimum deserialisation by reading only the type field.
     */
```

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
```diff
@@ -147,9 +145,20 @@ private void runInternal(final IntegrationConfig parsed) throws Exception {
           final JsonNode config = parseConfig(parsed.getConfigPath());
           validateConfig(integration.spec().getConnectionSpecification(), config, "WRITE");
           final ConfiguredAirbyteCatalog catalog = parseConfig(parsed.getCatalogPath(), ConfiguredAirbyteCatalog.class);
-          try (final AirbyteMessageConsumer consumer = destination.getConsumer(config, catalog, outputRecordCollector)) {
-            runConsumer(consumer);
-          }
+
+          final Procedure consumeWriteStreamCallable = () -> {
+            try (final SerializedAirbyteMessageConsumer consumer = destination.getSerializedMessageConsumer(config, catalog, outputRecordCollector)) {
```
davinchia, line 150: for understanding: after this merge, we should be able to get rid of `getConsumer`?

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
```diff
@@ -197,31 +206,55 @@ private void produceMessages(final AutoCloseableIterator<AirbyteMessage> message
...
-        EXIT_THREAD_DELAY_MINUTES,
-        TimeUnit.MINUTES);
+  @VisibleForTesting
+  static void consumeWriteStream(final SerializedAirbyteMessageConsumer consumer,
+                                 final BufferedInputStream bis,
+                                 final ByteArrayOutputStream baos)
+      throws Exception {
+    consumer.start();
+
+    final byte[] buffer = new byte[8192]; // 8K buffer
```
davinchia, line 223: any specific reason to use 8k?

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
```diff
@@ -197,31 +206,55 @@ private void produceMessages(final AutoCloseableIterator<AirbyteMessage> message
...
+          lastWasNewLine = false;
+        }
+      }
+    }
+
+    // Handle last line if there's one
+    if (baos.size() > 0) {
+      consumer.accept(baos.toString(StandardCharsets.UTF_8), byteCount);
+    }
+    // }
```
davinchia, line 250:
```suggestion
```

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
davinchia: is this from somewhere else? looks familiar.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
davinchia: ah - didn't realise this was moved from an existing class.

davinchia (approved):
:shipit: 

I'm not 100% the shim is in the right place. It looks like `SwitchingDestination` is where this should go. I don't think this is blocking.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
ryankfu: 8k bytes is the default buffer size for `BufferedInputStream`

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
```diff
@@ -197,31 +206,55 @@ private void produceMessages(final AutoCloseableIterator<AirbyteMessage> message
...
+                                 final ByteArrayOutputStream baos)
+      throws Exception {
+    consumer.start();
+
+    final byte[] buffer = new byte[8192]; // 8K buffer
+    int bytesRead;
+    int byteCount = 0;
+    boolean lastWasNewLine = false;
+
+    while ((bytesRead = bis.read(buffer)) != -1) {
```
edgao, line 228: this looks functionally equivalent to scanner.nextLine. Is there a performance advantage to doing it manually?

I'd prefer to either explore a prebuilt class or really beef up the tests in IntegrationRunnerBackwardsCompatbilityTest. E.g. verify behavior with very large lines (>8192 bytes), multibyte characters, etc. Possibly even generating some randomized inputs / fuzzing. This sort of low-level byte handling makes me very nervous.

(SO also suggests using `BufferedReader#readLine` https://stackoverflow.com/a/5172393)

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
```diff
@@ -197,31 +206,55 @@ private void produceMessages(final AutoCloseableIterator<AirbyteMessage> message
...
+      for (int i = 0; i < bytesRead; i++) {
+        final byte b = buffer[i];
+        if (b == '\n' || b == '\r') {
+          if (!lastWasNewLine && baos.size() > 0) {
```
edgao, line 232: what's the difference between byteCount and baos.size()?

edgao (commented):
the manual system.in parsing scares me. Open to being convinced that I'm being overcautious though.

other than that, agree that this is a noop. had a few comments for my own learning + one nitpick.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
```diff
@@ -29,8 +35,126 @@ AirbyteMessageConsumer getConsumer(JsonNode config,
...
+                                                                        final Consumer<AirbyteMessage> outputRecordCollector)
+      throws Exception {
+    return new ShimToSerializedAirbyteMessageConsumer(getConsumer(config, catalog, outputRecordCollector));
+  }
+
   static void defaultOutputRecordCollector(final AirbyteMessage message) {
     System.out.println(Jsons.serialize(message));
   }
 
+  @Slf4j
```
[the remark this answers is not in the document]
cgardens: yup

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
```diff
@@ -29,8 +35,126 @@ AirbyteMessageConsumer getConsumer(JsonNode config,
...
+     *
+     * @param consumer An {@link AirbyteMessageConsumer} that can handle the provided message.
+     * @param inputString JSON representation of an {@link AirbyteMessage}.
+     * @throws Exception if an invalid state message is provided or the consumer is unable to accept the
+     *         provided message.
+     */
+    @VisibleForTesting
+    static void consumeMessage(final AirbyteMessageConsumer consumer, final String inputString) throws Exception {
+
+      final Optional<AirbyteMessage> messageOptional = Jsons.tryDeserialize(inputString, AirbyteMessage.class);
```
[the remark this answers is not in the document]
cgardens: nope. i'd be happy to switch this up. for context, this is just a lift and shift from `IntegrationRunner`, so i leaned towards not changing things that i didn't have to, but my assessment is the same as yours that we probably don't need to do a separate deserialization.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
cgardens: we can only get rid of `getConsumer` once all destinations are migrated over to use the new thing. but for snowflake, yes, in subsequent PRs it overrides this method instead.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
```diff
@@ -147,9 +145,20 @@ private void runInternal(final IntegrationConfig parsed) throws Exception {
...
-            runConsumer(consumer);
-          }
+
+          final Procedure consumeWriteStreamCallable = () -> {
+            try (final SerializedAirbyteMessageConsumer consumer = destination.getSerializedMessageConsumer(config, catalog, outputRecordCollector)) {
+              consumeWriteStream(consumer);
+            }
+          };
+
+          watchForOrphanThreads(
```
cgardens, line 155: for context, we moved this because on master it is getting called before the `close` is called on the destination. this doesn't make sense, because it is checking if the destination cleaned itself up properly. that's a ridiculous thing to change before you let the destination call `close`. now this only gets called _after_ the destination calls `close`

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
cgardens: +1 to what ryan said.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
```diff
@@ -29,8 +35,129 @@ AirbyteMessageConsumer getConsumer(JsonNode config,
...
+      if (messageOptional.isPresent()) {
+        consumer.accept(messageOptional.get());
+      } else {
+        if (isStateMessage(inputString)) {
+          throw new IllegalStateException("Invalid state message: " + inputString);
+        } else {
+          log.error("Received invalid message: " + inputString);
+        }
```
evantahler, line 119: 👍 I believe that all other types of messages that a source might emit (log, control, etc) should not be passed to the destination, so STATE messages are the only other types to check for which are meaningful

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
cgardens: we want to count the number of bytes we have per message. i went with the lower-level thing so that we could count the bytes as we pull them off the wire. otherwise we have to call `getBytes("UTF")).size()` on the string which isn't free. admittedly, i haven't tested the performance difference, but looking at the impl it looks like it has to allocate a new byte array to do this.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
```diff
@@ -29,8 +35,129 @@ AirbyteMessageConsumer getConsumer(JsonNode config,
...
+
+    /**
+     * Tests whether the provided JSON string represents a state message.
+     *
+     * @param input a JSON string that represents an {@link AirbyteMessage}.
+     * @return {@code true} if the message is a state message, {@code false} otherwise.
+     */
+    @SuppressWarnings("OptionalIsPresent")
+    private static boolean isStateMessage(final String input) {
+      final Optional<AirbyteTypeMessage> deserialized = Jsons.tryDeserialize(input, AirbyteTypeMessage.class);
```
evantahler, line 131: I'm sure you tested this, but just for my own understanding, how does this work?  Since there's only one top-level property, with the JSON parser only deser the top level, leaving `RECORD: {}` as a string?

evantahler: cc @davinchia remember when we tried this in mexico with just `string.contains()`? this is much better

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
```diff
@@ -29,8 +35,126 @@ AirbyteMessageConsumer getConsumer(JsonNode config,
...
+      throws Exception {
+    return new ShimToSerializedAirbyteMessageConsumer(getConsumer(config, catalog, outputRecordCollector));
+  }
+
   static void defaultOutputRecordCollector(final AirbyteMessage message) {
     System.out.println(Jsons.serialize(message));
   }
 
+  @Slf4j
+  class ShimToSerializedAirbyteMessageConsumer implements SerializedAirbyteMessageConsumer {
```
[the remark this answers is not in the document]
cgardens: because i don't want people to use it outside of this file. not a strongly held opinion though.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
cgardens: * `bytesRead` tracks the bytes read out of the buffer. i.e. (1 of 8192)
* `byteCount` is tracking number of bytes read for a single message.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/SerializedAirbyteMessageConsumer.java
cgardens: i don't have a strong opinion. i think we said fivetran limits record size to 200MB, so that seems like it gives us a fair amount of buffer. happy to swtich it.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
```diff
@@ -197,31 +206,55 @@ private void produceMessages(final AutoCloseableIterator<AirbyteMessage> message
...
+
+    final byte[] buffer = new byte[8192]; // 8K buffer
+    int bytesRead;
+    int byteCount = 0;
+    boolean lastWasNewLine = false;
+
+    while ((bytesRead = bis.read(buffer)) != -1) {
+      for (int i = 0; i < bytesRead; i++) {
+        final byte b = buffer[i];
+        if (b == '\n' || b == '\r') {
```
evantahler, line 231: [nit] Any worries about "encoded" newline characters within the records themselves (`{body: "hello\nworld"}`)?  Is there a builtin for `sytem.newline` or something like that?

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
edgao: sorry, I was more thinking about bytesCount vs the `baos.size()` call. I think we don't need bytesCount at all, and can rely on the bytearrayoutputstream to track it for us?

(though obv in the other comment chain, I'm trying to find a way to avoid this entirely)

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
edgao: how about something like this? it doesn't exclude newline/carriage return characters, but presumably that's a negligible difference
```java
try (CountingInputStream counter = new CountingInputStream(System.in);
    BufferedReader reader = new BufferedReader(new InputStreamReader(counter))) {
  while (true) {
    String line = reader.readLine();
    if (line == null) {
      break;
    }
    long messageBytes = counter.getCount();
    counter.reset();
  }
}
```

evantahler (commented):
None of my comments are blocking, but the `/test` seems to be failing:

```
BigQueryStandardDestinationAcceptanceTest > testCustomDbtTransformations() FAILED
    io.airbyte.workers.exception.TestHarnessException: dbt debug Failed.
        at app//io.airbyte.integrations.standardtest.destination.DestinationAcceptanceTest.testCustomDbtTransformations(DestinationAcceptanceTest.java:967)
```
So something isn't working quite right

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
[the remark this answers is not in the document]
edgao: I also don't have a strong opinion here, let's just leave it as is

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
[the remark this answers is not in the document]
edgao: ah gtocha. didn't realize this is existing code. let's just put a comment here saying we suspect it's unnecessary

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/Destination.java
cgardens: fwiw, jonathan pearlin is the one who set this up. this has been in prod for a while.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
cgardens: I have a test in `IntegrationRunnerBackwardsCompatbilityTest.java` to make sure the behavior is the same between what we have on master and this branch to make sure we don't mess anything up.

cgardens:
running tests on "master" [here](https://github.com/airbytehq/airbyte/pull/26806) to compare test results

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
cgardens: 🔥 🔥 🔥  i didn't know this was a thing. let me try it! if it works, i agree this is way better.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
cgardens: agreed.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/base/IntegrationRunner.java
edgao: womp. I might mess around with this later today. I feel like there's some ordering of reader -> counter -> buffer that could work. agree with merging this pr though.

PR #26702, opened 2023-05-27
grab bag of non-controversial clean up tasks

cgardens:
## What
* Various clean up that we did on the way for async snowflake. Pulling into its own PR to keep the harder PRs more focused.


airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/AsyncStreamConsumer.java
```diff
@@ -148,21 +139,6 @@ private void validateRecord(final AirbyteMessage message, final StreamDescriptor
...
-
-    trackerIsValidRecord(message, streamDescriptor);
-  }
-
-  private void trackerIsValidRecord(final AirbyteMessage message, final StreamDescriptor streamDescriptor) {
-    // todo (cgardens) - is valid should also move inside the tracker, but don't want to blow up more
-    // constructors right now.
-    try {
-
-      if (!isValidRecord.apply(message.getRecord().getData())) {
```
cgardens, line 160: we will have to figure out what to do when we move Redshift over to use this. But for that's the only one that uses it and trying to save this interface that relies on the data being deserialized isn't worth the effort now.

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/csv/CsvSerializedBuffer.java
```diff
@@ -62,12 +66,20 @@ protected void writeRecord(final AirbyteRecordMessage record) throws IOException
 
   @Override
   protected void flushWriter() throws IOException {
-    csvPrinter.flush();
```
cgardens, line 65: in a single thread it wasn't really possible (or likely) for this to get called without accept getting called first. in a multi-threaded world that can actually happen pretty frequently, so when it log and move on.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination/buffered_stream_consumer/RecordSizeEstimator.java
```diff
@@ -71,7 +71,7 @@ public long getEstimatedByteSize(final AirbyteRecordMessage record) {
   }
 
   @VisibleForTesting
-  static long getStringByteSize(final JsonNode data) {
+  public static long getStringByteSize(final JsonNode data) {
```
cgardens, line 74: we end up using this in some tests later on

airbyte-integrations/bases/base-java-s3/src/main/java/io/airbyte/integrations/destination/s3/csv/CsvSerializedBuffer.java
davinchia: +1 

maybe leave as a comment?

davinchia (approved):
🏃🏻‍♂️

ryankfu (approved):
LGTM, is there a reason you want to merge into serialize messge consumer instead of master?

PR #26704, opened 2023-05-27
Source E2E Test: Add multi-threading support

cgardens:
## What
* We want to see if we can get source E2E even faster. We suspect we can do this multi-threading it because STDOUT should be able to hit higher throughput than what we're seeing. 

@davinchia what do you think of this approach. I couldn't think of a way of doing this that doesn't require changing `IntegrationRunner`.

## 🚨 User Impact 🚨
no


cgardens:
/connector-performance connector=connectors/source-e2e-test ref=cgardens/threaded-source-e2e

cgardens:
/connector-performance connector=connectors/source-e2e-test ref=master

cgardens:
/connector-performance connector=connectors/source-e2e-test ref=68cf197b5a4f13ecf092e39cc366932600ac327e

cgardens:
/connector-performance connector=connectors/source-e2e-test ref=cgardens/threaded-source-e2e

davinchia:
Old.

PR #26706, opened 2023-05-28
Encapsulate and test triggers for assigning flush workers

cgardens:
## What
* The logic for which stream we assign a flush worker is one of the places where we will do the most tuning. That means as we change it, we need to be well encapsulate and well tested.
* This PR is designed to be readable to get someone up to speed on when we trigger. It's good context for the destinations team as the logic for when we decide to flush is a little different than the existing one.

## How
* Moves logic for triggers into its own class.
* Adds rigorous javadoc comments.
* Adds rigorous testing.

## Recommended reading order
1. `DetectStreamToFlush.java`
2. `RunningFlushWorkers.java`

### DIFF
--- a/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/DetectStreamToFlush.java
+++ b/airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/DetectStreamToFlush.java
@@ -1,0 +1,222 @@
+/*
+ * Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+ */
+
+package io.airbyte.integrations.destination_async;
+
+import com.google.common.annotations.VisibleForTesting;
+import io.airbyte.integrations.destination_async.buffers.BufferDequeue;
+import io.airbyte.protocol.models.v0.StreamDescriptor;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+
+/**
+ * This class finds the best, next stream to flush.
+ */
+@Slf4j
+public class DetectStreamToFlush {
+
+  private static final double EAGER_FLUSH_THRESHOLD = 0.90;
+  private static final long QUEUE_FLUSH_THRESHOLD_BYTES = 10 * 1024 * 1024; // 10MB
+  private static final long MAX_TIME_BETWEEN_REC_MIN = 5L;
+  private final BufferDequeue bufferDequeue;
+  private final RunningFlushWorkers runningFlushWorkers;
+  private final AtomicBoolean isClosing;
+  private final DestinationFlushFunction flusher;
+
+  public DetectStreamToFlush(final BufferDequeue bufferDequeue,
+                             final RunningFlushWorkers runningFlushWorkers,
+                             final AtomicBoolean isClosing,
+                             final DestinationFlushFunction flusher) {
+    this.bufferDequeue = bufferDequeue;
+    this.runningFlushWorkers = runningFlushWorkers;
+    this.isClosing = isClosing;
+    this.flusher = flusher;
+  }
+
+  /**
+   * Get the best, next stream that is ready to be flushed.
+   *
+   * @return best, next stream to flush. If no stream is ready to be flushed, return empty.
+   */
+  public Optional<StreamDescriptor> getNextStreamToFlush() {
+    return getNextStreamToFlush(computeQueueThreshold());
+  }
+
+  /**
+   * We have a minimum threshold for the size of a queue before we will flush it. The threshold helps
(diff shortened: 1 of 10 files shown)


airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/DetectStreamToFlush.java
```diff
@@ -0,0 +1,222 @@
...
+    this.isClosing = isClosing;
+    this.flusher = flusher;
+  }
+
+  /**
+   * Get the best, next stream that is ready to be flushed.
+   *
+   * @return best, next stream to flush. If no stream is ready to be flushed, return empty.
+   */
+  public Optional<StreamDescriptor> getNextStreamToFlush() {
```
davinchia, line 50: looks like this isn't wired up in `FlushWorkers` - is the plan to do that in a follow up PR?

davinchia (approved):
Great stuff! Super love the comments + tests.

I only skimmed the tests as they felt very similar to our previous branch.

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/DetectStreamToFlush.java
```diff
@@ -0,0 +1,222 @@
...
+   * @param stream stream
+   * @return is time triggered and a debug string
+   */
+  @VisibleForTesting
+  ImmutablePair<Boolean, String> isTimeTriggered(final StreamDescriptor stream) {
+    final Boolean isTimeTriggered = bufferDequeue.getTimeOfLastRecord(stream)
+        .map(time -> time.isBefore(Instant.now().minus(MAX_TIME_BETWEEN_REC_MIN, ChronoUnit.MINUTES)))
+        .orElse(false);
+
+    final String debugString = String.format("time trigger: %s", isTimeTriggered);
```
ryankfu, line 133: How useful is this for debugging? I can only see this being useful since `getNextStreamToFlush` doesn't differentiate between `isSizedTriggered` or `isTimeTriggered` to log the universal `debugString` and you want to know if it was time or size triggered

There would be greater value with the `Instant.now()` and `bufferDequeue.getTimeOfLastRecord(stream).get()` values imo

ryankfu (approved):

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/RunningFlushWorkers.java
```diff
@@ -0,0 +1,80 @@
...
+    streamToFlushWorkerToBatchSize = new ConcurrentHashMap<>();
+  }
+
+  /**
+   * Call this when a worker starts flushing a stream.
+   *
+   * @param stream the stream that is being flushed
+   * @param flushWorkerId flush worker id
+   */
+  public void trackFlushWorker(final StreamDescriptor stream, final UUID flushWorkerId) {
```
ryankfu, line 33: Is this class intended to be wired in later?

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/DetectStreamToFlush.java
evantahler, line 1: The comments in this file are 
swedish_chef

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/DetectStreamToFlush.java
```diff
@@ -0,0 +1,222 @@
...
+        AirbyteFileUtils.byteCountToDisplaySize(queueSizeThresholdBytes),
+        AirbyteFileUtils.byteCountToDisplaySize(currentQueueSize),
+        AirbyteFileUtils.byteCountToDisplaySize(sizeOfRunningWorkersEstimate),
+        AirbyteFileUtils.byteCountToDisplaySize(queueSizeAfterRunningWorkers));
+
+    return ImmutablePair.of(isSizeTriggered, debugString);
+  }
+
+  /**
+   * For a stream, determines how many bytes will be processed by CURRENTLY running workers. For the
```
evantahler, line 174: Is it possible for a stream to have more than one active worker? (I think so)

evantahler (approved):
Great comments and tests! 
There are a few things not-wired up (as others have pointed out), but I'm going to assume a later PR uses them

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/DetectStreamToFlush.java
cgardens: yes!

airbyte-integrations/bases/base-java/src/main/java/io/airbyte/integrations/destination_async/DetectStreamToFlush.java
cgardens: fair point. in practice, this is a simple trigger and we haven't seen a bug here yet. i'm sure we will one day and then want to log more, but i'm inclined to wait until then. the tradeoff is how huge the log lines gets so if we go with what you're suggesting you really need `now: 2023/05/30 12:00.000+06 record added at: 2023/05/30 11:00.000+06`.

PR #26715, opened 2023-05-29
Source Notion: fix expected records

roman-yermilov-gl:
## What
*User expected record is no longer present in actual records.*

## How
*Delete nonexistent user record*


lazebnyi (approved):
Please add link to failed action to PR description.

PR #26719, opened 2023-05-29
🐛 Source Hubspot: fix then state has explicit `empty str` state value

bazarnov:
## What
Resolving: https://github.com/airbytehq/oncall/issues/2148

## How
- added additional checks for `stream_state` and `""` state value, assign `start_date` instead
- edited `expected records` with updated `deals` and `deals_pipelines` stream records. The new property was added silently: `"hs_num_of_associated_line_items": 1`
- covered the case  with unit test `test_streams.test_empty_string_in_state()`


lazebnyi (commented):
Please follow the instruction in Pre-merge Actions item or delete it at all.

bazarnov:
> Please follow the instruction in Pre-merge Actions item or delete it at all.

Done. It's removed.

airbyte-integrations/connectors/source-hubspot/source_hubspot/streams.py
```diff
@@ -753,18 +753,22 @@ def filter_by_state(self, stream_state: Mapping[str, Any] = None, record: Mappin
...
+        state_value = self._start_date
+        # we should check the presence of `stream_state` to overcome `availability strategy` check issues
+        if stream_state:
+            state_value = stream_state.get(self.cursor_field)
+            # sometimes the state is saved as `EMPTY STRING` explicitly
+            state_value = (
+                self._start_date if str(state_value) == "" else pendulum.from_format(str(state_value), self.cursor_field_datetime_format)
+            )
+        # compare the state with record values and get the max value between of two
+        cursor_value = max(state_value, record_value)
```
lazebnyi, line 766: let's add unit test for new functionality.

lazebnyi (approved):
LGTM! Unit tests for new functionality required.

airbyte-integrations/connectors/source-hubspot/source_hubspot/streams.py
bazarnov: There is no new functionality, but ok, I'll cover this.

lazebnyi: This is new code which handle empty state, before this PR we haven't this feature. So, this is new functionality and should be covered by unit tests. Thanks! 🙂

PR #26721, opened 2023-05-29
Source Mixpanel: fix expected records

roman-yermilov-gl:
## What
*Expected records do not match*

## How
*Updated records for `revenue`, `export`, `funnels` streams because no more records present in 10 days time window*


lazebnyi (approved):

PR #26740, opened 2023-05-30
connectors-ci: dynamic dependency resolution and test trigger

### LINKED ISSUE #25051: connectors-ci: get a stable list of required gradle subprojects to build a connector
The list of gradle projects required for a connector build is currently hardcoded.
We keep this list to only mount a subset of the airbyte repo to the container the will perform the gradle build.
The hardcoded list is not a good solution because we have to update it according to repo / projects changes.
We should find a way to dynamically spot the dependencies to load the gradle containers.
Maybe mounting all the directory that have a `build.gradle` file and that are not python connectors?

alafanechere:
## What
Closes #25051

Java connectors are built with Gradle. The dependencies to build these connectors are declared in their `build.gradle` file.
So far, we've been hardcoding the usual dependencies our java connectors have, but it's weak as the list of dependencies might evolve following refactor of connectors bases.

This PR suggest an approach to: 
* Compute the "local" dependencies of any connectors, by parsing their `build.gradle` files for Java connectors
* Include the "local" dependencies in the gradle build context to only load what's actually required by the connector in the build context.
* Trigger tests on connectors using a dependency if one of the dependency was updated. In other words the `--modified` flag now also consider a connector to be modified if one of its dependency (base) was updated .

## How
1. [Parse `build.gradle` file to get the list of "local" dependencies directories](https://github.com/airbytehq/airbyte/pull/26740/files#diff-a9928a5048f2e8b3c2770108831ff0acca8e12ed4b03b563060a74292ece8f36R78)
2. [Expose a `get_local_dependencies_paths` on the `Connector` class to get the list of the directories a connector depends on](https://github.com/airbytehq/airbyte/pull/26740/files#diff-a9928a5048f2e8b3c2770108831ff0acca8e12ed4b03b563060a74292ece8f36R225)
3. [Find intersection between modified files on branch on connector dependencies to get the list of modified connectors.
](https://github.com/airbytehq/airbyte/pull/26740/files#diff-c10daddb93e9e7509d41684ef94e2f81316789f83b339d8c67acb0d7b2f17a41R224)
4. [Mount connector's gradle dependencies when a running gradle task](https://github.com/airbytehq/airbyte/pull/26740/files#diff-8483fdbe9f47e848ceffb731bc1c8acd14199bec09b52360b6b9bea0be7d8327R53)

## 🚨 User Impact 🚨
This change should improve Java connectors developers DX:
* Their changes on directories like `source-jdbc` will trigger tests on all the connectors using it.

This change could also help introduce a nice autobump logic or automated testing of a pool of connectors upon CDK changes.


tools/ci_connector_ops/ci_connector_ops/pipelines/gradle.py
```diff
@@ -28,33 +27,6 @@ class GradleTask(Step, ABC):
     BIND_TO_DOCKER_HOST = True
     gradle_task_name: ClassVar
 
-    # TODO more robust way to find all projects on which the task depends?
-    JAVA_BUILD_INCLUDE = [
```
evantahler, line 32: Nice! Deleting these custom declarations which could drift

tools/ci_connector_ops/ci_connector_ops/pipelines/gradle.py
```diff
@@ -69,23 +41,6 @@ def docker_service_name(self) -> str:
...
-        """
-        if self.context.connector.technical_name.endswith("-strict-encrypt"):
-            return [Connector(self.context.connector.technical_name.replace("-strict-encrypt", ""))]
-        if self.context.connector.technical_name == "source-file-secure":
-            return [Connector("source-file")]
-        if self.context.connector.technical_name == "destination-bigquery-denormalized":
-            return [Connector("destination-bigquery")]
-        if self.context.connector.technical_name == "destination-dev-null":
-            return [Connector("destination-e2e-test")]
-        return []
```
evantahler, line 87: Can you confirm (do a test) that changing a 'regular' connector also knows to publish the strict encrypt version?

evantahler (commented):
This is great.  

I'd like to see a demo of this working both ways: 
- A parent changes and a child is triggered (`source-postgres` -> `source-postgres-strict-encrypt`)
- A child changes and a parent isn't triggered(`source-postgres-strict-encrypt` does not trigger `source-postgres`)


We know of at least python connector (`source-file` -> `source-file-secure`) in which we do need track... some sort of partent-child linkage.  Where should we store that?  Is that out of scope here? 

Also, adding @rodireich as a reviewer to comment from the Java team's POV.

tools/ci_connector_ops/ci_connector_ops/pipelines/gradle.py
rodireich: To verify that I'm getting what's going on here. What you're saying @evantahler should happen since a build.gradle in a strict-encrypt flavor has a dependency for its non strict-encrypt.
e.g. in source-postgres-strict-encrypt you got 
```implementation project(':airbyte-integrations:connectors:source-postgres')```

File: tools/ci_connector_ops/ci_connector_ops/utils.py [hunk 3]
```diff
@@ -26,6 +26,7 @@
 DESTINATION_CONNECTOR_PATH_PREFIX = CONNECTOR_PATH_PREFIX + "/destination-"
 ACCEPTANCE_TEST_CONFIG_FILE_NAME = "acceptance-test-config.yml"
 AIRBYTE_DOCKER_REPO = "airbyte"
+GRADLE_PROJECT_RE_PATTERN = r"project\((['\"])(.+?)\1\)"
```
rodireich, line 29: Do you care that both `implementation` and `testImplementation` will match?

alafanechere: Good question!
I believe that test dependencies should be included at build time because they're problalby use in running java connectors tests. But do change to these tests dependencies should trigger a test pipeline for all the connectors using it? I'm not sure of that, as if I make update to connector acceptance tests I'm not sure I want to run a test on ALL connectors using it. 
I'll differentiate dependencies from test dependencies so that we can control this.

alafanechere:
Demo time for @evantahler : 
1. Let's make a change to `source-postgres` and check that both `source-postgres` and `source-postgres-strict-encrypt` are tested.

alafanechere:
> We know of at least python connector (source-file -> source-file-secure) in which we do need track... some sort of partent-child linkage. Where should we store that? Is that out of scope here?

I focused on java connectors here as their dependencies is explicitly declared. For source-file-secure IMO we should not use `FROM airbyte/source-file:0.3.9` in the dockerfile to express the dependency. We shall revamp this connector to install `source-file` as a python package and make it do the necessary override. Local python package dependencies is already handled in the pipeline code so this should be the way to go if we want to discard Docker files for python connectors too.

alafanechere:
@rodireich @evantahler my main question is:
* Do we want to trigger tests on connectors for all local dependencies that are updated.
Example:
* When a change is made to `base-java` do we want to run tests for all Java connectors.
I would say yes ideally.
But in the python connectors world I'm not sure we want to run tests an all connectors on CDK update...

rodireich:
I think testing all java connectors when base-java changes for example would be a very helpful thing as we keep breaking tests when making updates to shared code.
I get what you say with python/CDK as the number of connectors makes testing everything too slow.
but for java the long tail is not that long.
Is there a way to separate the two maybe and only test the entire local dependency tree for java?

airbyte-integrations/connectors/source-postgres/Dockerfile
```diff
@@ -26,3 +26,4 @@ COPY --from=build /airbyte /airbyte
 
 LABEL io.airbyte.version=2.0.30
 LABEL io.airbyte.name=airbyte/source-postgres
+LABEL dummy="dummy"
```
rodireich, line 29: Don't forget to revert

evantahler:
I really like the suggestion that `source-file-secure` use it's package-manager dependencies (pip) to call out that it depends on `source-file`.  Elegant!

tools/ci_connector_ops/ci_connector_ops/utils.py
```diff
@@ -74,6 +75,23 @@ def get_changed_acceptance_test_config(diff_regex: Optional[str] = None) -> Set[
...
 
 
+def get_gradle_project_dependencies_paths(gradle_file_path: Path):
+    dependencies_paths = []
+    with open(gradle_file_path) as gradle_file:
+        for line in gradle_file:
+            if "implementation project" in line or "implementation files" in line:
+                match = re.search(GRADLE_PROJECT_RE_PATTERN, line)
+                if match:
+                    project_name = match.group(2)
```
bnchrch, line 85: Given that were doing something a little custom and regexs are involved I would advocate that we add comments with very explicit example of 

We are looking for a line like
```java
    implementation project(':airbyte-integrations:bases:base-java')
```

and want to extract `airbyte-integrations:bases:base-java`

tools/ci_connector_ops/ci_connector_ops/utils.py
```diff
@@ -74,6 +75,23 @@ def get_changed_acceptance_test_config(diff_regex: Optional[str] = None) -> Set[
...
+                    if project_name.startswith(":"):
+                        project_path = project_name[1:].replace(":", "/")
+                        dependencies_paths.append(Path(project_path))
+                        dependency_gradle_file = Path(f"{project_path}/build.gradle")
```
bnchrch, line 89: Oh man, such a shame that gradle doesn't have a good way to list all downstream dependencies 😢

bnchrch (approved):
I believe we should add the comments, but given that gradle doesnt provide a way to list downstream dependencies I think what you have here is perfectly fine.

👍 from me

tools/ci_connector_ops/ci_connector_ops/utils.py
alafanechere: I could run `gradle dependencies` command inside a gradle container and parse its output, but it'll add a lot of overhead to this "pre pipeline run".

alafanechere:
> I really like the suggestion that source-file-secure use it's package-manager dependencies (pip) to call out that it depends on source-file. Elegant!

@evantahler issue created: https://github.com/airbytehq/airbyte/issues/26823

alafanechere:
@evantahler I confirm that the changes in [0d81b65](https://github.com/airbytehq/airbyte/pull/26740/commits/0d81b654bc59ba31a42d2d8fcef2e260855c96fb) that modified `source-postgres-strict-encrypt` and `source-pokeapi` correctly triggered the following tests:
* source-pokeapi
* source-postgres-strict-encrypt
* source-alloydb-strict-encrypt (which explicitly depends on `source-postgres-strict-encrypt` according to its [build.gradle](https://github.com/airbytehq/airbyte/blob/c0c36ca90822409e48a8161b5708d2ddc409a1ef/airbyte-integrations/connectors/source-alloydb-strict-encrypt/build.gradle#L17) file.

The tests ran as expected: failure on pokeapi acceptance because its broken, other failure because the connector version was not bumped.

PR #26741, opened 2023-05-30
CDK: Improve schema detection

### LINKED ISSUE #25669: Connector builder: Improve schema detection
Currently, the schema detection of the connector builder is differentiating between `integer` and `number`. As we are working with small sample sets, there is a high chance to recognize a number as an integer on accident because the test read did only yield integers by chance.

To increase the chance for a successful first sync, all numeric values should be detected as "number" in the schema.

It should also not produce `anyOf` or `oneOf` but instead a type union (but if it can't be avoided it should be `oneOf`):
```
 oneOf: [{type: string}, {type: integer}] vs ["string", "integer"]
```

flash1293:
## What

Closes https://github.com/airbytehq/airbyte/issues/25669

Improves the built-in schema inferrer to produce schemas that work better with the Airbyte platform by default:
* Always assume a number is "number", not "integer", because the sample size is usually small
* Ignore properties of type "null", as this is most likely not correct anyway and can cause errors later on when the schema gets refined - instead this will just be unexpected fields that are filtered out by the platform
* Do no emit types like this:
```
{
  "a": {
    "anyOf": [
      { "type: "null" },
      { "type: "object", "properties": { ... } },
    ]
  }
}
```
instead, do this:
```
{
  "a": {
    "type": ["null", "object"],
    "properties": { ... }
  }
}
```
as the platform is ignoring anyOf but can extract the nested object in the second case.

## How

The number/integer case can be handled by adding another extra strategy.

However, the null handling is baked in too deeply into genson to be able to get the desired behavior by extra strategies (e.g. the anyOf can't be changed by provided strategies and the library _always_ expects an output for a schema node, so an "ignore null" strategy can't be implemented). To solve this, a post-processing function is introduced that traverses the built schema and changes the output as desired.

## 🚨 User Impact 🚨

Schema detection in the connector builder will change for the better (existing connectors won't be affected)


airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
```diff
@@ -22,13 +23,25 @@ def to_schema(self):
         return schema
 
 
+class IntegerToNumber(Number):
+    """
+    This class has the regular Number behaviour, but it will never emit an integer type.
+    """
+
+    def __init__(self, node_class):
+        super(IntegerToNumber, self).__init__(node_class)
```
girarda, line 32: nit: this should be equivalent to `super().__init__(node_class)`

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
```diff
@@ -22,13 +23,25 @@ def to_schema(self):
...
+
 class NoRequiredSchemaBuilder(SchemaBuilder):
-    EXTRA_STRATEGIES = (NoRequiredObj,)
+    EXTRA_STRATEGIES = (NoRequiredObj, IntegerToNumber)
 
 
 # This type is inferred from the genson lib, but there is no alias provided for it - creating it here for type safety
 InferredSchema = Dict[str, Union[str, Any, List, List[Dict[str, Union[Any, List]]]]]
 
+SchemaNode = Union[dict, dict[str, str, Any, list, list[dict[str, Any, list]]]]
```
girarda, line 43: I'm a little confused by this type definition. what is a `dict[str, str, Any, list, list[dict[str, Any, list]]]` ?

girarda (approved):

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
flash1293: That's a leftover from some testing, the `InferredSchema` above is correct already - I removed it

PR #26744, opened 2023-05-30
fix: modified files detection

alafanechere:
## What
We want to trigger tests on connectors that got modified because one of their file was deleted.

## How
Fix the `--diff-filter` option value.


octavia-approvington:

octavia-approvington (approved):

PR #26747, opened 2023-05-30
Source Confluence: fix expected records

roman-yermilov-gl:
## What
*Expected records do not match for stream `pages`. A view field is an html field and some tag attributes in actual records are in different order.*

## How
*Updated view field in expected records with actual ones*


lazebnyi (approved):

PR #26775, opened 2023-05-30
:bug: Source Google Directory: migrate to advanced auth

davydov-d:
## What
#26232

## How
Remove `authSpecification` from the connector spec because the OAuth flow is not yet supported for this connector in cloud env


airbyte-integrations/connectors/source-google-directory/source_google_directory/spec.json
```diff
@@ -78,13 +78,5 @@
         ]
       }
     }
-  },
-  "authSpecification": {
```
girarda, line 82: do we automatically convert authSpecification to advancedAuth or is this a breaking change?

davydov-d: @girarda this is not a breaking change. Both advancedAuth and authSpecification are nullable, besides this connector is outdated on cloud and has no oauth support so we will not break things

girarda (approved):

bazarnov (approved):

lazebnyi (approved):

PR #26776, opened 2023-05-30
turn on async snowflake

cgardens:
## What
* This is the feature branch for Async Snowflake. It "turns on" the Async Snowflake Destination. 
* We want to incrementally roll this destination to users. In order to do that we want all the code for Async Snowflake can live in master and that all we need is this single commit to "turn it on". In other words, this PR should only ever differ from master by one commit. Then we will use this to publish the Async Snowflake Destination during roll out. Once it is 100% rolled out, this PR can be merged into master. 
* Basically this command is how we update it: `git rebase --onto origin/master HEAD~1`

## How
*Describe the solution*

## Recommended reading order
1. `SnowflakeDestination.java` <== this is the only file that should ever have a diff. if it doesn't, there's a problem!

## 🚨 User Impact 🚨
Yes, this officially switched destination-snowflake over to the async version.

### DIFF
--- a/airbyte-integrations/connectors/destination-snowflake/metadata.yaml
+++ b/airbyte-integrations/connectors/destination-snowflake/metadata.yaml
@@ -22,8 +22,8 @@ data:
     jobSpecific:
       - jobType: sync
         resourceRequirements:
-          memory_limit: 1Gi
-          memory_request: 1Gi
+          memory_limit: 2Gi
+          memory_request: 2Gi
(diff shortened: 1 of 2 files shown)


ryankfu:
/publish connector=connectors/destination-snowflake pre-release=true

ryankfu:
/legacy-publish connector=connectors/destination-snowflake pre-release=true run-tests=false

davinchia:
No longer needed.

PR #26785, opened 2023-05-30
Source Freshsales: fix expected records

roman-yermilov-gl:
## What
*Contacts stream expected records do not match. Field `open_deals_count` now has value 2 instead of 0*

## How
*FIx `open_deals_count` value to 2*


lazebnyi (approved):

PR #26792, opened 2023-05-30
Jacoco report for source-postgres

nguyenaiden:
## What
[25681](https://app.zenhub.com/workspaces/db--dw-source-connectors-6333360e0a41155061efbcbd/issues/gh/airbytehq/airbyte/25681)
Added an extra step in CI workflow to run and generate a JaCoCo test coverage report for source-postgres. Will be available to other databases when the time comes.
## 🚨 User Impact 🚨
Contributors to the `source-postgres` will be notified if total module coverage drops below 64%, the current percentage. While not blocking, this needs to be addressed by the PR author. Example: 

_Passing:_
image
_Failing:_
image

### DIFF
--- a/.github/workflows/jacoco_report.yml
+++ b/.github/workflows/jacoco_report.yml
@@ -27,4 +27,5 @@ jobs:
           paths: ${{ github.workspace }}/airbyte-integrations/connectors/source-postgres/build/reports/jacoco/test/jacocoTestReport.xml
           token: ${{ secrets.GITHUB_TOKEN }}
           min-coverage-overall: 64
+          title: Coverage report for source-postgres
           update-comment: true
--- a/.github/workflows/jacoco_report.yml
+++ b/.github/workflows/jacoco_report.yml
@@ -9,6 +9,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v3
       - name: Set up Java
         uses: actions/setup-java@v3
         with:
--- a/.github/workflows/jacoco_report.yml
+++ b/.github/workflows/jacoco_report.yml
@@ -3,7 +3,7 @@ name: Measure coverage
 on:
   pull_request:
     branches:
-      - duy/main
+      - master
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/jacoco_report.yml
```diff
@@ -25,5 +26,6 @@ jobs:
         with:
           paths: ${{ github.workspace }}/airbyte-integrations/connectors/source-postgres/build/reports/jacoco/test/jacocoTestReport.xml
           token: ${{ secrets.GITHUB_TOKEN }}
-          min-coverage-overall: 40
-          min-coverage-changed-files: 60
+          min-coverage-overall: 64
+          title: Coverage report for source-postgres
```
prateekmukhedkar, line 30: When we extend the coverage to other connectors, the title will need to be templatized so that it does not keep showing source-postgres. Don't think it is blocking the merge on this PR.

prateekmukhedkar (approved):

.github/workflows/jacoco_report.yml
nguyenaiden: It's not blocking. We can extend it by running the gradle test command for multiple db sources. We can use this [tool](https://docs.gradle.org/current/userguide/jacoco_report_aggregation_plugin.html) to display an aggregate report

akashkulk (approved):

.github/workflows/jacoco_report.yml
```diff
@@ -3,12 +3,13 @@ name: Measure coverage
 on:
   pull_request:
     branches:
-      - duy/main
+      - master
```
rodireich, line 6: Would this always checkout master branch?
Is this what we want to run a report on?

.github/workflows/jacoco_report.yml
rodireich, line 1: Who is calling into this yaml?

.github/workflows/jacoco_report.yml
nguyenaiden: This is triggered when someone is creating a PR into master.

rodireich: I see, this is for PR targeting master.
I'm not the biggest expert on GitHub Actions, but I'd expect to see a `ref` to the PR branch in the checkout operation.

Can you confirm that what we check out is the PR branch?

rodireich (approved):

PR #26801, opened 2023-05-30
update bump_version to use shell script

c-p-b:
## What

Resolves https://github.com/airbytehq/airbyte-internal-issues/issues/4408

This change was made in airbyte-platform, but we maintain two copies of this script for now and this also needed to be changed in airbytehq/airbyte

## How
Make the script match whats in airbytehq/airbyte-platform

### DIFF
--- a/tools/bin/bump_version.sh
+++ b/tools/bin/bump_version.sh
@@ -9,13 +9,29 @@ if ! test "$(tty)" == "not a tty"; then
 fi
 
 set -o xtrace
-PREV_VERSION=$(grep -w VERSION gradle.properties | cut -d"=" -f2)
+REPO=$(git ls-remote --get-url | xargs basename -s .git)
+echo $REPO
+if [ "$REPO" == "airbyte" ]; then
+  PREV_VERSION=$(grep -w VERSION run-ab-platform.sh | cut -d"=" -f2)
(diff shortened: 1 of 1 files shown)


git-phu (approved):

tools/bin/bump_version.sh
```diff
@@ -9,13 +9,29 @@ if ! test "$(tty)" == "not a tty"; then
 fi
 
 set -o xtrace
-PREV_VERSION=$(grep -w VERSION gradle.properties | cut -d"=" -f2)
+REPO=$(git ls-remote --get-url | xargs basename -s .git)
+echo $REPO
+if [ "$REPO" == "airbyte" ]; then
+  PREV_VERSION=$(grep -w VERSION run-ab-platform.sh | cut -d"=" -f2)
```
git-phu, line 15: In the workflow it looks like we run this script (bump_version) before updating`run-ab-platform.sh`.  Will this work as intended? 

https://github.com/airbytehq/airbyte/blob/e88eb48e07eb5466ff7125dda93e83c1bae24ba5/.github/workflows/release-airbyte-os.yml#L129-L141

cpdeethree: Yes, that is expected behavior, this means that airbyte will now fetch the correct platform version when running --refresh

Issue #26791, opened 2023-05-30
Investigate slow select count(*) queries


akashkulk:
### Revelant information

More context here : https://github.com/orgs/airbytehq/projects/23/views/1?pane=issue&itemId=28581474

Facing some issues where `select count(*) where cursor > cursor_value` are taking a large amount of time on some databases. This is causing a heartbeat failure and subsequently causing the sync to fail. Need to do a deeper dive to understand : 
- If we should remove all `select count(*)` queries
- How different types of indexes can affect performance of `select count(*)` queries.

PR #26810, opened 2023-05-31
Remove incremental estimate query from Postgres

akashkulk:
As part of investigating https://github.com/airbytehq/oncall/issues/2067, the `select count(*)` query can be super slow. In addition, the estimate is currently not being used anywhere. Let's remove this one in particular as we determine what to do with the other one. 

Part of https://github.com/airbytehq/airbyte/issues/26791


airbyte-integrations/connectors/source-relational-db/src/main/java/io/airbyte/integrations/source/relationaldb/AbstractDbSource.java
```diff
@@ -466,7 +453,6 @@ private AutoCloseableIterator<AirbyteMessage> getIncrementalStream(final Databas
         table.getFields().stream().anyMatch(f -> f.getName().equals(cursorField)),
         String.format("Could not find cursor field %s in table %s", cursorField, table.getName()));
 
-    estimateIncrementalSyncSize(database, airbyteStream, cursorInfo, cursorType);
```
rodireich, line 469: Postgres is the only implementation I see 👍

rodireich (approved):

PR #26812, opened 2023-05-31
Connector CI: Ensure Crane has appropriate auth

bnchrch:
## What
Login to dockerhub in the crane container

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
@@ -772,22 +772,25 @@ def with_airbyte_python_connector_full_dagger(context: ConnectorContext, build_p
 
 def with_crane(
         context: PipelineContext,
-        docker_hub_username_secret: Optional[str] = None,
-        docker_hub_password_secret: Optional[str] = None,
     ) -> Container:
     """Crane is a tool to analyze and manipulate container images.
     We can use it to extract the image manifest and the list of layers or list the existing tags on an image repository.
     https://github.com/google/go-containerregistry/tree/main/cmd/crane
     """
 
+    # We use the debug image as it contains a shell which we need to properly use environment variables
+    # https://github.com/google/go-containerregistry/tree/main/cmd/crane#images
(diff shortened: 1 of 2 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -770,9 +769,29 @@ def with_airbyte_python_connector_full_dagger(context: ConnectorContext, build_p
...
+    if context.docker_hub_username_secret and context.docker_hub_password_secret:
+        base_container = (
+            base_container.with_secret_variable("DOCKER_HUB_USERNAME", context.docker_hub_username_secret).with_secret_variable(
+                "DOCKER_HUB_PASSWORD", context.docker_hub_password_secret
+            )
+            # We need to use skip_entrypoint=True to avoid the entrypoint to be overridden by the crane command
+            # We use sh -c to be able to use environment variables in the command
+            # This is a workaround as the default crane entrypoint doesn't support environment variables
+            .with_exec(
+                ["sh", "-c", "crane auth login index.docker.io -u $DOCKER_HUB_USERNAME -p $DOCKER_HUB_PASSWORD"], skip_entrypoint=True
```
bnchrch, line 793: @alafanechere 

I went on way to long of a deep dive here but essentially I think this is our best way to do this.

Here are a few roadblocks I found
1. cranes default entrypoint didnt support env vars in any of the following formats `$ENV` `${ENV}` `"$ENV"` `\"$ENV\"`
2. crane did not have a shell in the image by default, you need to use a `debug` image to get access
3. I couldnt use a dagger secret inline `with_exec("auth", "login", "index.docker.io", "-u", context.docker_hub_username_secret)`

The only alternative I found was explictly reading the secrets and passing them in as plain text

```python
    async def _run(self) -> StepResult:
        docker_repository, docker_tag = self.context.docker_image_name.split(":")
        dhu = await self.context.docker_hub_username_secret.plaintext()
        dhs = await self.context.docker_hub_password_secret.plaintext()
        crane_ls = (
            environments.with_crane(
                self.context,
                dhu,
                dhs,
            )
```

Which I think is worse.

Let me know your thoughts

alafanechere (approved):
Thank you for the fix! I don't have a cleaner solution to suggest.

PR #26815, opened 2023-05-31
lowcode: Better describe what the cursor field is for

sherifnada:
I am always confused by what the "datetime format" field is for so I tried to make it better


flash1293 (approved):
LGTM

PR #26831, opened 2023-05-31
🐛 Source Hubspot: migrate to advancedAuth

### LINKED ISSUE #26236: Source Hubspot: Migrate to advancedAuth

tolik0:
## What
Migrate oAuth to advancedAuth.
#26236

## How
Add advanced_auth to spec.yaml.


airbyte-integrations/connectors/source-hubspot/source_hubspot/spec.yaml
```diff
@@ -66,6 +66,14 @@ connectionSpecification:
               examples:
                 - refresh_token
               airbyte_secret: true
+            access_token:
```
davydov-d, line 69: this is not implemented in OAuth flow, I think we need to stay consistent

davydov-d (changes requested):

airbyte-integrations/connectors/source-hubspot/source_hubspot/spec.yaml
tolik0: Fixed

davydov-d (approved):

lazebnyi (approved):

PR #26860, opened 2023-05-31
Improving error messages to have better messaging in datadog and the …

maxi297:
…frontend

## What
Today, we display to the user the message. However, it often do not contain information regarding the actual error since we blindly catch everything and do not provide information specific to the exception that was raised in the message.

## How
Adding the `str(exception)` in the error message for generic errors.

## 🚨 User Impact 🚨
Nothing breaking. The user will now see `Error handling request: Validation against json schema defined in declarative_component_schema.yaml schema failed` instead of `Error handling request.` for example


clnoll (approved):
👍

PR #26866, opened 2023-05-31
Add paths to only trigger on source-postgres changes

nguyenaiden:
## What
Only trigger JaCoCo workflow run if `source-postgres` files are touched


bnchrch (approved):

.github/workflows/jacoco_report.yml
```diff
@@ -4,6 +4,8 @@ on:
   pull_request:
     branches:
       - master
+    paths:
+      - "airbyte-integrations/connectors/source-postgres/**"
```
rodireich, line 8: You got all the dependencies in base-java and the abstract classes

nguyenaiden: I was thinking about that as well, but it won't affect source-postgres unless someone publishes it. And also Jacoco is scanning only for changes in that module.

PR #26888, opened 2023-06-01
connectors-ci: add retry logic at step level

alafanechere:
## What
Some Step might require a retry logic due to transient failures.
E.G: the connector package install transiently fail due to some pip errors related to not found packages.

## How
* The `Step` abstract class has a `retry` attribute set to false by default and a `max_retries` attribute set to 3 by default
* If a retry is enable on a child class, when the `_run` call returns a failed step results `_run` is retried from within `run`


tools/ci_connector_ops/ci_connector_ops/pipelines/bases.py
```diff
@@ -92,9 +93,12 @@ class Step(ABC):
 
     title: ClassVar[str]
     started_at: ClassVar[datetime]
+    retry: ClassVar[bool] = False
```
bnchrch, line 96: Lets remove the retry boolean and intead rely on `max_retries` solely

```py
max_retries: ClassVar[int] = 0
```

```
class ConnectorPackageInstall(Step):
    max_retries = 3
```

bnchrch (approved):
One nit, will approve to unblock :)

PR #26889, opened 2023-06-01
connectors-ci: get modified files from PR API

alafanechere:
## What
In our connectors CI pipeline we originally get the list of modified files with `git diff` operation.
These operations are quite expensive, especially in the CI because we run the `git diff` operation in a dagger container that is checkouting airbyte's repo.

In a pull request context the GitHub API provides the list of modified files in the pull request.
We can use this list instead of running `git diff` to speed up pipelines running on PRs.

The current pull request is an attempt at doing so.
I also did an additional operation:
We previously alway ran two different git diff commands: one to get the list of modified files in branch, the other one to get the list of modified files in the last commit.
I slightly changed this logic: we get the list of modified files in the last commits only when we're on master. When we're on a branch we git diff the full branch.  

[Before on source-pokeapi](https://github.com/airbytehq/airbyte/actions/runs/5142482162/jobs/9256383996): total run duration 3min
[After change on source-pokeapi:](https://github.com/airbytehq/airbyte/actions/runs/5142582706/jobs/9256495554) total run duration 2mn

## How
* Centralize the logic of getting the list of modified files in [`get_modified_files`](https://github.com/airbytehq/airbyte/blob/1923cdd766b727c6e8d11bffe294051c77bf41d7/tools/ci_connector_ops/ci_connector_ops/pipelines/commands/airbyte_ci.py#L26)

## User impact

Faster test run on PR: expect 1 minute speed gain.


tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -24,6 +24,7 @@
 
 if TYPE_CHECKING:
     from ci_connector_ops.pipelines.contexts import ConnectorContext
+    from github import PullRequest
```
evantahler, line 27: I was expecting some curl/API work in this PR.  I guess this just handles it!

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/airbyte_ci.py
```diff
@@ -13,12 +15,34 @@
...
 
 from .groups.connectors import connectors
 from .groups.metadata import metadata
 
 
+def get_modified_files(
+    git_branch: str, git_revision: str, diffed_branch: str, is_local: bool, ci_context: CIContext, pull_request: PullRequest
+) -> List[str]:
+    """Get the list of modified files in the current git branch.
+    If the current branch is master, it will return the list of modified files in the current commit.
```
erohmensing, line 30: Suggestion: Maybe clarify that this "current commit" is the squash commit of the branch whose merge kicked off the run

The reason I mention this is that if we decided to switch from squash commits to a full history (e.g. rebase instead of merge), I'm not sure if this "last commit" logic would work anymore. Its an edge case not worth writing more code for, but a reminder of the context that allows us to do it this way may be helpful.

evantahler (approved):
Pros - faster 
Cons - our pipelines are even more resilient on Github... but they already are, so 🤷

erohmensing (approved):
This is dope, I didn't know that the pull request context has a list of changed files. Learn something new every day :D

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/airbyte_ci.py
alafanechere: I improved the docstring, thank you for the suggestion.

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
alafanechere: This [lib](https://github.com/PyGithub/PyGithub) is pretty good at wrapping the GitHub API for typed interaction.

alafanechere:
> Cons - our pipelines are even more resilient on Github... but they already are, so 🤷

True, but we can easily fallback on the `git diff`

bnchrch (approved):
Love some more speed!

PR #26892, opened 2023-06-01
Source Pipedrive: fix expected records

roman-yermilov-gl:
## What
*Expected records do not match. Missed pipeline_ids*

## How
*Updated existing recrods with pipeline_ids field*


davydov-d (changes requested):
New fields need to be added to the stream schema; also please bump the version and update the changelog

davydov-d (approved):

Issue #26907, opened 2023-06-01
source-s3: delete TestIntegrationCsvFiles?


alafanechere:
I'm currently migrating our test pipelines to Dagger (cf [airbyte-ci](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/README.md#L1).)
In dagger test are run inside containers.
The source-s3 `TestIntegrationCsvFiles` is using minio as a docker container.
I'm having a hard time making the `test_big_file` pass because:
* I've to customize the inter container networking setup in the Dagger context (I validate I could connect to minio from the test client)
* The volume sharing between the test container and minio is clumsy, looks like minio can't access the "big file" we mount and I've error like  `An error occurred (NoSuchBucket) when calling the ListObjectsV2 operation: The specified bucket does not  exist`.

Before spending lot of developer time trying to make it work I would challenge the usefulness of this test.
This tests:
* Generates a 1GB file
* Run a read on it as assert that all the records are read.

Is it something actually beneficial to test?

I suggest to either:
* Remove this test
* Interact with a real remote bucket and discard minio use (like the `TestIncrementalFileStreamS3` is doing)
* Invest time in adapting the test to work in the Dagger context.

PR #26908, opened 2023-06-01
source-s3: delete integration tests using minio

alafanechere:
## What
More details in https://github.com/airbytehq/airbyte/issues/26907
TLDR: minio tests are hard to run in the context of Dagger and I'm challenging the usefulness of this tests. I'm removing this integration test so that this connector can be properly tested in the new Connectors CI with Dagger.

## How
* Remove the `test_big_file` integration test.


alafanechere:
You can see in the CI checks lists that the CI tests with dagger passed.

evantahler (approved):
Fine with me.

Tagging @clnoll as an FYI, as she's thinking about a CDK for file-based sources... maybe some shared tooling for testing both big and small files is a good idea.

davydov-d (approved):

clnoll:
Thanks @evantahler. Removing the test makes sense to me. The file-based source CDK will have tests for large files.

alafanechere:
Relates to #25053

Issue #26909, opened 2023-06-01
Pass control message to connector builder frontend


maxi297:
### What area the feature impact?

Connectors

### Revelant Information

As a connector developer developing for single use refresh token API, I don't want to enter a new refresh token every time I do a test read

## Context
Some APIs invalidate the refresh token once it's used (for example, quickbooks and GitLab). The platform and the CDK already have a mechanism to support this where a control message is emitted with the new configuration. We need to replicate this mechanism.

## Gap
Today, the connector builder is silencing those messages. We need to return the latest one to the frontend so that it updates its config. In order to do that, we will:
* Return the most recent AirbyteControlConnectorConfigMessage as part of the API
* Frontend will use this information to update it's config

## Acceptance criteria
- [ ] A connector developer developing for single use refresh token API only needs to enter the refresh token once when opening the connector builder project. Do an example with an API
- [ ] Disable testing values menu while request is in flight to avoid race condition

flash1293:
I'll move this back to the sprint backlog for now as it depends on the refresh token authenticator being available in the UI

flash1293:
Un-assigning myself for the next week

PR #26922, opened 2023-06-01
[ISSUE #26909] add latest connector config control message to connect…

maxi297:
…or builder API

## What
Partially addresses https://github.com/airbytehq/airbyte/issues/26909. Once the information is available to the frontend, the frontend will need to update the config

## How
Add a new field in StreamRead and populate it with the latest CONNECTOR_CONFIG message based on `emitted_at`

## 🚨 User Impact 🚨
Once released and merged in the connector builder server, we will start returning the new field

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
+++ b/airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
@@ -79,7 +79,7 @@ class MessageGrouper:
             inferred_schema=schema_inferrer.get_stream_schema(
                 configured_catalog.streams[0].stream.name
             ),  # The connector builder currently only supports reading from a single stream at a time
-            latest_config_update=latest_config_update.connectorConfig.config if latest_config_update else None,
+            latest_config_update=latest_config_update.connectorConfig.config if latest_config_update else self._clean_config(config),
         )
 
     def _get_message_groups(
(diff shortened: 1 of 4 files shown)


flash1293 (approved):
Changes LGTM

airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
```diff
@@ -63,7 +66,9 @@ def get_message_groups(
                 if message_group.type == TraceType.ERROR:
                     error_message = f"{message_group.error.message} - {message_group.error.stack_trace}"
                     log_messages.append(LogMessage(**{"message": error_message, "level": "ERROR"}))
-
+            elif isinstance(message_group, AirbyteControlMessage):
+                if not latest_config_update or latest_config_update.emitted_at <= message_group.emitted_at:
```
clnoll, line 70: Nit: the fact that the control message we see here is related to config (and therefore that it's appropriate to assign it to the `latest_config_update` variable) is down in `_get_message_groups`; I wonder if it would be slightly clearer to move `and message.control.type == OrchestratorType.CONNECTOR_CONFIG` to this method and have `_get_message_groups` yield every control message.

clnoll (approved):
LGTM, just one nitpicky comment for you @maxi297, not a blocker.

brianjlai (approved):

airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
maxi297: I see the point that it's more readable from the public `get_message_groups`. I honestly don't have a strong preference but if you think this is important enough, I can spend cycles and update it

clnoll: It's just a nit so no worries if you'd rather leave it as-is.

PR #26924, opened 2023-06-01
Update gradle command to remove gitignored files

bnchrch:
## What
1. Updates our gitignore to be a little more specific
1. Updates gradle to remove files that should be removed by gitignore


bnchrch:
@cpdeethree You can see in the automated change commit that we removed the oss_registry and only the oss_registry

[d0c4657](https://github.com/airbytehq/airbyte/pull/26924/commits/d0c465760b97c41eb2141b726f55b2263283cb64)

.github/workflows/gradle.yml
```diff
@@ -321,10 +321,14 @@ jobs:
           attempt_limit: 3
           attempt_delay: 5000 # in ms
 
+      # This is helpful in the case that we change a previously commited generated file to be ignored by git.
+      - name: Remove any files that have been gitignored
+        run: git ls-files -i -c --exclude-from=.gitignore | xargs git rm --cached
+
```
alafanechere, line 327: 👍 I understand the purpose of it but do you want it to be permanent?

alafanechere (approved):

.github/workflows/gradle.yml
bnchrch: Oh so badly I want it to be permanent!

Here's some more context on this
https://airbytehq-team.slack.com/archives/C03VDJ4FMJB/p1685647715073319

Basically multiple times now I've
1. git rm --cached airbyte-config-oss/init-oss/src/main/resources/seed/oss_registry.json
2. Double checked its in the gitignore
3. Removed the file from my local filesystem
4. Run process resources locally and confirmed the file exists but git is not tracking it
5. Merged these changes to master.

However  if there exists a PR that
1. Is out of date from master after the above changes (all of them)
2. Has a commit that changes the oss_registry.json (some of them)
3. Is merged before updating to latest. (often)

That work gets undone as the file is readded, and currently tracked files do not adhere to the gitignore 😢

PR #26926, opened 2023-06-01
cat: skip test_catalog_has_supported_data_types

alafanechere:
## What
This `test_catalog_has_supported_data_types` test currently leads to too much failures on GA connectors. We should first address the invalid format at scale before enabling it.


octavia-approvington:

octavia-approvington (approved):

PR #26930, opened 2023-06-02
Fix normalization builds

evantahler:
Fixes dbt issues like `ImportError: cannot import name 'NotRequired' from 'typing_extensions' (/usr/local/lib/python3.9/site-packages/typing_extensions.py)`

There was a new version of a transitive dependency that was released that broke stuff, so we pinned to the previous version.   If we weren't about to delete all of normalization, we'd start using pip freeze  to formally lock down our versions.

### DIFF
--- a/airbyte-integrations/bases/base-normalization/setup.py
+++ b/airbyte-integrations/bases/base-normalization/setup.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
 
+# Note: cattrs is pinned to the last known working version which does not have conflicts with typing_extensions.  Learn more https://airbytehq-team.slack.com/archives/C03C4AVJWG4/p1685546430990049
 
 import setuptools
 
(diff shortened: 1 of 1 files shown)


ryankfu (approved):

jbfbell (approved):

cynthiaxyin (approved):

evantahler:
`io.airbyte.workers.exception.TestHarnessException: dbt debug Failed.` is happening on the bigquery tests.

Any ideas @edgao? I'll run them again just in case...

edgao:
ugh. this is probably because of the cred rotation. There's also some connector failures
```
 BigQueryGcsDestinationAcceptanceTest > testIncrementalDedupeSync() FAILED
    io.airbyte.workers.exception.TestHarnessException: Destination process exit with code 1. This warning is normal if the job was cancelled.
```

snowflake also has some interesting errors, for which I have no explanation
```
SnowflakeDestinationIntegrationTest > testCheckFailsWithInvalidPermissions() FAILED
    java.nio.file.NoSuchFileException: secrets/config.json

SnowflakeInternalStagingDestinationAcceptanceTest > testSyncWriteSameTableNameDifferentNamespace() FAILED
    org.opentest4j.AssertionFailedError: expected: <10> but was: <0>
```

I'll work on the bigquery failures locally tomorrow, and then hopefully get to snowflake. I think we should merge this pr as-is since it fixes the normalization build problems though.

evantahler:
👍 from me if you want to merge it!

edgao:
dest-bq tests are now green https://github.com/airbytehq/airbyte/actions/runs/5157751904 (including denormalized https://github.com/airbytehq/airbyte/actions/runs/5157753428)

PR #26931, opened 2023-06-02
Connector CI: Use one service account

bnchrch:
## What
A demo of what using one service account would look like

## TODO before merge
1. Give the github ci SA admin access to spec cache
2. Give thegithub ci SA admin access to the metadata service 

closes: airbytehq/airbyte-internal-issues#1850


alafanechere (commented):
To move forward on this PR we'd need the global SA to be created first right?

PR #26945, opened 2023-06-02
🐛 Source Zendesk Support: forced `ticket  metrics` stream  to use `cursor pagination`

bazarnov:
## What
Resolving: https://github.com/airbytehq/oncall/issues/2164

## How
- changed the way `request_params()` pass the `page`  param
- changed the way to get `next_page_token` due to a different response structure
- covered the case with unit tests: 
     - `TestSourceZendeskSupportCursorPaginationStream.test_next_page_token[TicketMetricEvents]`
     - `TestSourceZendeskSupportCursorPaginationStream.test_request_params[TicketMetricEvents]`

## 🚨 User Impact 🚨
Non-breaking change.


airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/streams.py
```diff
@@ -654,21 +654,28 @@ class TicketMetrics(SourceZendeskSupportCursorPaginationStream):
     """TicketMetric stream: https://developer.zendesk.com/api-reference/ticketing/tickets/ticket_metrics/"""
 
     def next_page_token(self, response: requests.Response) -> Optional[Mapping[str, Any]]:
-        if self._ignore_pagination:
-            return None
-        next_page = self._parse_next_page_number(response)
```
roman-yermilov-gl, line 659: It looks like there is no more usages of the `_parse_next_page_number` method. Shouldn't we delete it as well?

davydov-d, line 659: `_parse_next_page_number` should probably be removed as well

davydov-d (approved):

airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/streams.py
```diff
@@ -654,21 +654,28 @@ class TicketMetrics(SourceZendeskSupportCursorPaginationStream):
...
-        return next_page if next_page else None
+        """
+        https://developer.zendesk.com/documentation/api-basics/pagination/paginating-through-lists-using-cursor-pagination/#when-to-stop-paginating
+        """
+        meta = response.json().get("meta")
```
roman-yermilov-gl, line 660: You are getting "meta" using safe `get` method, so you expect it may not be a part of response. In case when `meta` not present we will get `AttributeError: 'NoneType' object has no attribute 'get'` error on `meta.get("after_cursor")`

airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/streams.py
bazarnov: Not like really, other streams are  using it:
- All `SourceZendeskSupportFullRefreshStream` based streams are using it, due to different response structures for different streams.

airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/streams.py
bazarnov: Done.

roman-yermilov-gl (approved):

airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/streams.py
```diff
@@ -654,21 +654,28 @@ class TicketMetrics(SourceZendeskSupportCursorPaginationStream):
...
     def next_page_token(self, response: requests.Response) -> Optional[Mapping[str, Any]]:
-        if self._ignore_pagination:
-            return None
-        next_page = self._parse_next_page_number(response)
-        return next_page if next_page else None
+        """
+        https://developer.zendesk.com/documentation/api-basics/pagination/paginating-through-lists-using-cursor-pagination/#when-to-stop-paginating
+        """
+        meta = response.json().get("meta", {})
+        return meta.get("after_cursor") if meta.get("has_more", False) else None
```
sh4sh, line 661: Hey @bazarnov, there is some discussion about this line in https://github.com/airbytehq/airbyte/pull/27070#discussion_r1228497844 and I am hoping you could help us understand. The contributor thinks it should be:

```suggestion
        return meta.get("after_cursor") if meta.get("has_more", True) else None
``` 

From what I can tell their logic seems correct, but I couldn't reproduce any bugs with your version and I don't want to change something that works.

What do you think about this? is there some context we're missing? any thoughts would be appreciated!

sh4sh: Nevermind, figured it out that `False` is just default here 😅  Thank you!

potatozerg:
Hi @bazarnov, I noticed an error with the ticket metrics table
/api/v2/ticket_metrics?start_time=1609459200&page%5Bsize%5D=100
Response Code: 503, Response Text: {
  "error": {
    "title": "DatabaseError",
    "message": "Database timeout error"
  }
}

the page[size]=100 is causing this, maybe we should remove it from the API call?

PR #26947, opened 2023-06-02
Source Gitlab: update expected records for pipelines streams

roman-yermilov-gl:
## What
*Expected records do not match. New field `name` added to `pipelines` `pipelines_extended` streams.*

## How
*Updated expected records and schema for missing field*


davydov-d (changes requested):
Please bump the version and update the changelog

davydov-d (approved):

PR #26949, opened 2023-06-02
Source Pinterest: fix expected records

roman-yermilov-gl:
## What
*Expected records do not match. New field `note` added to `board_pins` stream.*

## How
*Updated expected records and schema for stream `board_pins`*


davydov-d (changes requested):
please bump the version and update the changelog

davydov-d (approved):

Issue #20790, opened 2022-12-22
Support for authentication with keys as query parameter in configuration based connectors


imrehg:
## Tell us about the problem you're trying to solve
I'm working on adding a new source (#20789) and their authentication requires passing in the API key as a query parameter (called `key`). I'm planning to use the configuration based / low-code connector toolkit to build this, however such an authentication pattern is not available for that connector type (looking through [the docs](https://docs.airbyte.com/connector-development/config-based/understanding-the-yaml-file/authentication/)).

## Describe the solution you’d like
Option of query-parameter type authentication, similar to [the `APIKeyAuthenticator`](https://docs.airbyte.com/connector-development/config-based/understanding-the-yaml-file/authentication/#apikeyauthenticator) except as a query parameter with a specific key, rather than a HTTP request header.

## Describe the alternative you’ve considered or used
This could be pretty easily implementable in the [Python HTTP API based connector](https://docs.airbyte.com/connector-development/cdk-python/http-streams), but that would lose the benefits of the low-code setup, naturally.

## Additional context
N/A

## Are you willing to submit a PR?
Yes.

girarda:
This is a good idea! As a work around until it's implemented, you can set the api key directly in the requester's

example:
```
  requester:
    url_base: "..."
    request_options_provider:
      request_parameters:
        api_key: "{{ config['api_key'] }}"
```

imrehg:
Thanks @girarda, that's a super good point! Yeah, it would totally work, didn't think of that one. Interesting to find the right level of coding and config even in this low-code connector environment. Will give this one a try!

girarda:
Proposed schema change:
```
  ApiKeyAuthenticator:
    title: API Key Authenticator
    description: Authenticator for requests authenticated with an API token injected as an HTTP request header.
    type: object
    required:
      - type
      - api_token
    properties:
      type:
        type: string
        enum: [ApiKeyAuthenticator]
      api_token:
        title: API Key
        description: The API key to inject in the request. Fill it in the user inputs.
        type: string
        interpolation_context:
          - config
        examples:
          - "{{ config['api_key'] }}"
          - "Token token={{ config['api_key'] }}"
      inject_into:
        title: Inject API Key Into Outgoing HTTP Request
        description: Configure how the API Key will be sent in requests to the source API.
        "ref": "#/definitions/RequestOption"
      $parameters:
        type: object
        additionalProperties: true
```
- Replace `header` field with `inject_into`

## Implementation 
[...]

PR #26953, opened 2023-06-02
Low code CDK: Allow query param / body injection for api key authenticator

flash1293:
## What

Part of https://github.com/airbytehq/airbyte/issues/20790

This PR deprecates the header param of the api key authenticator and introduces a request_option param that allows to also inject the auth key into a query param or the body

## How

To allow this I had to change the class hierarchy a bit - the important part is that the simple retriever can spread in authenticator-defined query params and body params into outgoing requests. To do so the interface of the authenticator known to the retriever has to include methods for getting query params and body params (not for headers  as this is already handled by the current setup using the Session class from the requests package). Currently this is set to `AuthBase`. As every authenticator can theoretically be able to do this, `DeclarativeAuthenticator` is extended with the required methods and the authenticator type in the requester is changed to `DeclarativeAuthenticator`.

This removes a bunch of double inheritance for other declarative connectors which is a nice side effect.

As an image:
Screenshot 2023-06-02 at 15 36 05

Note that the python class `ApiKeyAuthenticator` is changed to only support the request option, the deprecated header property is handled in the component factory to isolate this behavior to the yaml layer.

Theoretically, custom authenticators used to work fine as long as they implemented AuthBase, now they have to implement `DeclarativeAuthenticator`. I checked existing custom authenticators and they all do this already, I guess it was implied that a custom authenticator needs to do this in the first place (it just didn't bear any weight so far). I clarified this in the description.

## 🚨 User Impact 🚨

Low code CDK users can now inject the token from the api key authenticator in request params or the body:
```
type: ApiKeyAuthenticator
token: "{{ config.token }}"
inject_into:
   type: "RequestOption"
   field_name: "authKey"
   inject_into: "request_parameter"
```

The `header` property is still supported, but should not be used anymore.

The `ApiKeyAuthenticator` class from `airbyte_cdk.sources.declarative.auth.token` does not support the header parameter anymore.


airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/declarative_authenticator.py
```diff
@@ -3,20 +3,32 @@
...
 @dataclass
-class DeclarativeAuthenticator:
+class DeclarativeAuthenticator(AbstractHeaderAuthenticator):
     """
     Interface used to associate which authenticators can be used as part of the declarative framework
     """
 
+    def get_request_params(self) -> Union[MutableMapping[str, Any], None]:
+        """HTTP request parameter to add to the requests"""
+        return None
```
girarda, line 19: is there value in returning `None` instead of `{}`?

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -191,6 +195,8 @@ def request_headers(
             self.requester.get_request_headers,
             self.paginator.get_request_headers,
             self.stream_slicer.get_request_headers,
+            # auth headers are handled separately
```
girarda, line 198: nit: let's point to where in the code this is handled to help future us

airbyte-cdk/python/airbyte_cdk/sources/declarative/auth/declarative_authenticator.py
flash1293: Not really, I'll simplify

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -191,6 +197,8 @@ def request_headers(
             self.requester.get_request_headers,
             self.paginator.get_request_headers,
             self.stream_slicer.get_request_headers,
+            # auth headers are handled separately by passing the authenticator to the HttpStream constructor
```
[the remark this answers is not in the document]
flash1293: I would like to split this out from the main change as it would require changes to the other authenticators as well.

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -224,7 +224,23 @@ def create_add_fields(self, model: AddFieldsModel, config: Config, **kwargs) ->
...
+        request_option = (
+            RequestOption(
+                inject_into=RequestOptionType(model.inject_into.inject_into.value),
+                field_name=model.inject_into.field_name,
+                parameters=model.parameters,
+            )
+            if model.inject_into
+            else RequestOption(
+                inject_into=RequestOptionType.header,
+                field_name=InterpolatedString.create(model.header, parameters=model.parameters).eval(config),
```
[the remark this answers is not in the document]
flash1293: Good point, that's not necessary

flash1293:
@girarda Adjusted the PR, could you have another look?
* Fails if both inject_into and header are defined
* Cleaned up the types

girarda (approved):

PR #26964, opened 2023-06-02
Source Metabase: change activity stream

roman-yermilov-gl:
## What
*Activity stream is not available on [latest version v0.46](https://www.metabase.com/docs/v0.46/api/activity), so the stream crashes with 404 error. Becasue of Metabase is not versioned there is no way to make the stream work. ([previous version v0.45 for comparison](https://www.metabase.com/docs/v0.45/api/activity))*

## How
*Proposal: change `Activity` stream to use `/activity/recent_views` endpoint instead of deprecated `/activity`*

### DIFF
--- a/airbyte-integrations/connectors/source-metabase/source_metabase/manifest.yaml
+++ b/airbyte-integrations/connectors/source-metabase/source_metabase/manifest.yaml
@@ -43,7 +43,7 @@ definitions:
     $ref: "#/definitions/base_stream"
     $parameters:
       name: "activity"
-      path: "activity"
+      path: "activity/recent_views"
   cards_stream:
     $ref: "#/definitions/base_stream"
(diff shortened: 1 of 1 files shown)


airbyte-integrations/connectors/source-metabase/source_metabase/manifest.yaml
```diff
@@ -43,7 +43,7 @@ definitions:
     $ref: "#/definitions/base_stream"
     $parameters:
       name: "activity"
-      path: "activity"
+      path: "activity/recent_views"
```
davydov-d, line 46: Will the stream schema stay the same?

davydov-d (changes requested):
you need to update the version in metadata.yaml and Dockerfile + update the changelog

Issue #26342, opened 2023-05-22
SingleUseRefreshToken in Connector Builder


flash1293:
### Component Description

Generalizing means that we want to simplify the configuration for the user and have only one interface/YAML component to support that. However, the Python CDK implementation can stay the same.

### Proposed YAML schema

```yaml
  OAuthAuthenticator:
    required:
      <...>
    properties:
      refresh_token_updater:
        properties:
          refresh_token_config_path:
            title: Config Path To refresh_token
            description: Describe where in the config the refresh token is. To be defined when the refresh token is single use. This is necessary to update the refresh token after it's been used
            type: array
            items:
              type: string
            examples:
              - ["credentials", "refresh_token"]
          access_token_config_path:
            type: array
            items:
              type: string
          token_expiry_date_config_path:
            type: array
            items:
              type: string
```
Given that `refresh_token_updater` is defined as part of OAuthAuthenticator, we will instantiate `SingleUseRefreshTokenOauth2Authenticator`.

To make it work consistently with the connector builder, two other things need to be changed:
* Expiry date format has to be supported like on regular OAuth authenticator
* Authenticator shouldn't fail if config values are missing - if no access token / expiry date is set, it's just getting a new access token via the refresh token

## Acceptance criteria
- [ ] Create single YAML interface and wire up the Python component properly
- [ ] In frontend, add switch to oauth authenticator UI and convert accordingly
- [ ] Update documentation
- [ ] Update source-quickbooks and use its example in the documentation
- [ ] Ensure that there there are no SingleUseRefreshTokenOAuthAuthenticator in the DB `select count(*) from actor_definition_config_injection where json_to_inject::text like '%SingleUseRefreshTokenOAuthAuthenticator%';`

Out-of-scope: 
- [ ] refresh token update mechanism in the frontend (see https://github.com/airbytehq/airbyte/issues/26909)

flash1293:
To discuss: Config path arguments should be replaced by interpolated strings to make it consistent with the other authenticators:
```
        client_id_config_path: Sequence[str] = ("credentials", "client_id"),
        client_secret_config_path: Sequence[str] = ("credentials", "client_secret"),
        access_token_config_path: Sequence[str] = ("credentials", "access_token"),
        refresh_token_config_path: Sequence[str] = ("credentials", "refresh_token"),
        token_expiry_date_config_path: Sequence[str] = ("credentials", "token_expiry_date"),
```

Plan:
* update it to use interpolated string
[...]

PR #26966, opened 2023-06-02
Extend low code OAuthAuthenticator with token refresh capabilities

flash1293:
## What

Related to https://github.com/airbytehq/airbyte/issues/26342

Allow a user to specify a refresh token updater object as part of the `OAuthAuthenticator` low code component:
```
    authenticator:
      type: OAuthAuthenticator
      client_id: "some_client_id"
      client_secret: "some_client_secret"
      token_refresh_endpoint: "https://api.sendgrid.com/v3/auth"
      refresh_token: "{{ config['apikey'] }}"
      refresh_request_body:
        body_field: "yoyoyo"
        interpolated_body_field: "{{ config['apikey'] }}"
      refresh_token_updater:
        refresh_token_name: "the_refresh_token"
        refresh_token_config_path:
          - apikey
```

If specified, the refresh token, access token and access token expiry time returned from the authentication response is written back to the config object. This is important if a refresh token can only be used once.

## How

* Add refresh_token_updater object to the existing oauth configuration
* If set, instantiate a SingleUseRefreshTokenOAuth2Authenticator instead

Following changes were necessary in the CDK class:
* Replace config id and client config paths params with regular string params (resolve in factory) - if not set they still grab it from the config automatically (required to adapt the xero connector because it uses custom config paths - see https://github.com/airbytehq/airbyte/pull/27007)
* Add expiry format if it's not a number of seconds for feature compatibility as the regular oauth authenticator already allowed this
* Do not fail if config values are not available

On the low code CDK side, the jinja interpolation happens in the factory at creation time.

Changes to quickbooks and removing the existing low code component will happen on a separate PR.

Long term, it's planned to merge the functionality in a single class, this has been split out of the initial phase to reduce risk and scope.

## 🚨 User Impact 🚨

SingleUseRefreshTokenOAuth2Authenticator interface does not support custom config paths anymore for client id and client secret, needs to be adapted by resolving them before creating an instance.


girarda (approved):
lgtm! 🚢

flash1293:
@girarda Could you take another quick look? Adjusted as discussed in grooming: Make work with empty config values (remove validation and do not fail on empty config values)

The discussed removal of the direct SingleUseRefreshTokenOAuthAuthenticator in low code will be done in a follow-up PR: https://github.com/airbytehq/airbyte/pull/27053

girarda (approved):

PR #26979, opened 2023-06-02
Fix grep versionbump

c-p-b:
## What
There was a bug in our script that greps run-ab-platform.sh to get the canonical verison that was causing issue with release where, since we changed how we were searching for version it was actually matching two things in the file, making the version bump script unable to actually run

## How
Fixes the bug as well as manually bumps versions to bring them back in line with what airbyte-platform is at

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.44.8
+current_version = 0.44.9
 commit = False
 tag = False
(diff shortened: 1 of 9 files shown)


sajarin (approved):

cpdeethree:
This is the second line that was matching:

https://github.com/airbytehq/airbyte/blob/0a995bf9d622fc333ec3518334cef1f40bdc4748/run-ab-platform.sh#L46

With the new regex, this will no longer match

PR #27003, opened 2023-06-05
connectors-ci: no concurrency on java connectors for nightly builds

alafanechere:
## What
Concurrent test for Java connectors are failing (see https://github.com/airbytehq/airbyte/issues/26962)

## How
* On nightly builds, separate java connectors testing in a different worfklow step with concurrency forced to 1
* Create a separate workflow for manual run of tests

Issue #26862, opened 2023-05-31
connectors-ci: `runc run failed` errors


kpenfound:
The following category of errors have been observed in Publish pipelines:

```
runc run failed: unable to start container process: error during container init: error mounting "/var/lib/dagger/runc-overlayfs/snapshots/snapshots/347127/fs" to rootfs at "/var/lib/docker": stat /var/lib/dagger/runc-overlayfs/snapshots/snapshots/347127/fs: no such file or directory
```

Originally reported in https://github.com/airbytehq/airbyte/issues/25877

Relevant comment from Erik: https://github.com/airbytehq/airbyte/issues/25877#issuecomment-1548300551
> Yeah this one I was not able to repro on my own in the past, but I will look through your logs to find any clues about the one you most recently reported in discord.

>In general, the error means buildkit is trying to mount a cache dir into a container that no longer exists on disk, which should never happen. So the most likely explanation is a "use-after-free" type of bug somewhere, but just looking through the relevant code in buildkit I haven't spotted anything yet. I may end up just upstreaming more logs to buildkit that will help debug this further, but we'll see what I find.

>It's very hard to say what you could do to mitigate it since root cause is so unclear right now. The only thing that is at least worth a shot would be to use a single `dockerd` service for all connectors in a test suite, rather than giving each connector its own. That should greatly reduce the number of times that these containers need to be spun up and down, which theoretically could reduce how often this bug has a chance of occurring. But again, that's just a best guess for now. Re-using dockerd more is probably a performance improvement anyways, so doesn't hurt to try it anyways.

Example failure: https://github.com/airbytehq/airbyte/actions/runs/4919988172/jobs/8788287616#step:5:6107

kpenfound:
Currently in a 'monitoring' state on this one to see if it comes up again. We'll also need an upstream change to debug this further which includes the stack in the trace logs which @sipsma has more context on

sipsma:
Yeah for this one we'll need to catch it happening in one of the engines in the dev cluster w/ trace logs enabled.

I just last week added even more trace logs to buildkit that should help debug this (and related bugs) further, on top of the previous ones. We are planning on a release tomorrow, so once that's out, Airbyte should be able to upgrade and pick up those extra logs too.

---

For some more context on the logs that will be of interest:
1. Either of these logs will tell us the mapping of dagger cache volume key -> buildkit cache record id
[...]

PR #27012, opened 2023-06-05
connectors-ci: disable concurrency for Java connectors testing

alafanechere:
## What
I observed that concurrent testing of multiple connectors can lead to errors like https://github.com/airbytehq/airbyte/issues/26862

As a stopgap I suggest to disable concurrency on Java connectors tests.
[It indeed make java connector testing work without discrepancies compared to /test.
](https://github.com/airbytehq/airbyte/actions/runs/5174185398).

[I originally only set this up for nightly builds, but the same kind of problem can happen on PR.
](https://github.com/airbytehq/airbyte/pull/27003)I find that it's better to force concurrency from the pipeline logic instead of GHA workflow logic.

## How
Call `run_connector_test_pipelines` two times: 
* one for non java connectors with concurrency set to the value passed to as a CLI option
* one for java connectors with concurrency forced to 1.


tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
```diff
@@ -314,29 +332,31 @@ def publish(
...
+                ctx.obj["git_branch"],
+                ctx.obj["git_revision"],
+                gha_workflow_run_url=ctx.obj.get("gha_workflow_run_url"),
+                pipeline_start_timestamp=ctx.obj.get("pipeline_start_timestamp"),
+                ci_context=ctx.obj.get("ci_context"),
+                pull_request=ctx.obj.get("pull_request"),
+            )
+            for connector, modified_files in selected_connectors_and_files.items()
+        ]
+    )
```
alafanechere, line 359: Auto format

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
```diff
@@ -177,12 +177,30 @@ def test(
...
+        anyio.run(
+            run_connectors_pipelines,
+            [
+                connector_context
+                for connector_context in connectors_tests_contexts
+                if connector_context.connector.language == ConnectorLanguage.JAVA
+            ],
+            run_connector_test_pipeline,
+            "Test Pipeline",
+            1,
```
evantahler, line 201: That's the concurrency argument

evantahler (approved):

alafanechere:
Closing as https://github.com/airbytehq/airbyte/pull/27021 allows us to not us this stop gap.

PR #27047, opened 2023-06-06
Metadata Upload Validation: Update to use registry url

### LINKED ISSUE #26975: Update Metadata Upload to validate using crane/docker not the dockerhub api
## Problem
The dockerhub api can be out of date with the registry leading to false negatives in our publish pipeline

## Solution
metadata Update the validation to use crane or docker instead of the api

## Appendix
relevant slack thread: https://airbytehq-team.slack.com/archives/C056HGD1QSW/p1685726669445679

bnchrch:
## Overview

Docker hubs api can sometimes become out of date leading to our upload incorrectly not finding a docker image in docker hub

this updates the function to use the same url used by the docker and crane cli's

closes #26975

### DIFF
--- a/airbyte-ci/connectors/metadata_service/lib/metadata_service/docker_hub.py
+++ b/airbyte-ci/connectors/metadata_service/lib/metadata_service/docker_hub.py
@@ -37,7 +37,4 @@ def is_image_on_docker_hub(image_name: str, version: str) -> bool:
     tag_url = f"https://registry.hub.docker.com/v2/repositories/{image_name}/tags/{version}"
     response = requests.get(tag_url, headers=headers)
 
-    if response.ok:
-        return True
-
-    return False
+    return response.ok
--- a/airbyte-ci/connectors/metadata_service/lib/metadata_service/docker_hub.py
+++ b/airbyte-ci/connectors/metadata_service/lib/metadata_service/docker_hub.py
@@ -21,30 +21,6 @@ def get_docker_hub_auth_token() -> str:
     return token
 
 
-def get_image_tags(repo: str, image: str) -> List[str]:
-    token = get_docker_hub_auth_token()
-    headers = {"Authorization": f"JWT {token}"}
-
-    tags_url = f"https://hub.docker.com/v2/repositories/{repo}/{image}/tags/"
-    params = {"page_size": 100}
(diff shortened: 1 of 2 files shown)


airbyte-ci/connectors/metadata_service/lib/metadata_service/docker_hub.py
```diff
@@ -55,7 +31,13 @@ def is_image_on_docker_hub(image_name: str, version: str) -> bool:
     Returns:
         bool: True if the image and version exists on Docker Hub, False otherwise.
     """
-    repo, image = image_name.split("/", 1)
-    tags = get_image_tags(repo, image)
 
-    return version in tags
+    token = get_docker_hub_auth_token()
+    headers = {"Authorization": f"JWT {token}"}
```
alafanechere, line 36: The fact that we use the same token to connect to https://hub.docker.com makes me think https://registry.hub.docker.com is just an alias. Do you get different responses when using `registry`?

alafanechere (approved):
I believe it's a reasonable attempt at fixing the problem we had, but I'm not 💯  sure hub.docker.com and registry.docker.com are not using same backends...

airbyte-ci/connectors/metadata_service/lib/metadata_service/docker_hub.py
```diff
@@ -55,7 +31,13 @@ def is_image_on_docker_hub(image_name: str, version: str) -> bool:
...
+    tag_url = f"https://registry.hub.docker.com/v2/repositories/{image_name}/tags/{version}"
+    response = requests.get(tag_url, headers=headers)
+
+    if response.ok:
+        return True
+
+    return False
```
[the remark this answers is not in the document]
bnchrch: Oh god damn it 🤦

airbyte-ci/connectors/metadata_service/lib/metadata_service/docker_hub.py
bnchrch: They are different endpoints in the sense that before we were only able to get a list of tags, not individual tags.

It seems to work. but theres no way to know for sure.

We can see if we encounter the out of sync error down the road

PR #27051, opened 2023-06-06
Connector CI: Upload complete.json file on pipeline complete

bnchrch:
## What
Add a complete.json file for when a connector pipeline completes


tools/ci_connector_ops/ci_connector_ops/pipelines/actions/__init__.py
```diff
@@ -2,60 +2,3 @@
...
-from typing import TYPE_CHECKING, List, Tuple, Union
-
-import asyncer
-from ci_connector_ops.pipelines.bases import Step, StepStatus
-
-if TYPE_CHECKING:
-    from ci_connector_ops.pipelines.bases import StepResult
-
-
-async def run_steps(
```
bnchrch, line 17: I had to move this for circular import reasons 😢

tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
```diff
@@ -67,6 +67,8 @@ def __init__(
         slack_webhook: Optional[str] = None,
         reporting_slack_channel: Optional[str] = None,
         pull_request: PullRequest = None,
+        ci_report_bucket: Optional[str] = None,
+        ci_gcs_credentials: Optional[str] = None,
```
bnchrch, line 71: I moved pipeline report logic and creds to the base pipeline

tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
```diff
@@ -416,24 +420,7 @@ async def __aexit__(
         suffix = f"{connector_name}/{connector_version}/output.json"
         file_path_key = f"{self.report_output_prefix}/{suffix}"
 
-        local_report_path = Path(local_reports_path_root + file_path_key)
-
-        await local_report_path.parents[0].mkdir(parents=True, exist_ok=True)
```
bnchrch, line 421: I move report saving to the report model

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/connectors.py
```diff
@@ -4,16 +4,57 @@
...
+        return StepResult(step=None, status=StepStatus.FAILURE)
+
+    if state == ContextState.ERROR:
+        return StepResult(step=None, status=StepStatus.FAILURE)
+
+
+    raise ValueError(f"Could not convert context state: {state} to step status")
+
+
+# HACK: This is to avoid wrapping the whole pipeline in a dagger pipeline to avoid instability just prior to launch
```
bnchrch, line 33: @alafanechere This was the big set back.

To use the GCS uploads at the end of the pipeline we needed dagger.

This left us with a few weird options
1. Refactor to our utopia where all steps are pipelines (large effort)
2. Refactor our connector pipelines to be steps (large effort and in the wrong direction)
3. Add extra argument(s) to all our `run_connector_*_pipeline` functions (felt verbose)
4. bolt on a generic report complete pipeline at the end of the run that uses a hack to repurpose an existing connector pipeline

I went with 4 because it was fast, but more importantly it didnt change the logic of our pipelines in a large way. Which is something I wanted to avoid as we are hoping to launch these very soon.

Let me know your thoughts!

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
```diff
@@ -55,11 +52,55 @@ def validate_environment(is_local: bool, use_remote_secrets: bool):
...
 # COMMANDS
 
 
 @click.group(help="Commands related to connectors and connector acceptance tests.")
 @click.option("--use-remote-secrets", default=True)  # specific to connectors
+@click.option(
+    "--ci-gcs-credentials",
+    help="The service account to use during CI.",
+    type=click.STRING,
+    required=False,  # Not required for pre-release or local pipelines
```
alafanechere, line 101:
```suggestion
    required=False,  # Not required for pre-release or local pipelines
```
Do we have a logic to validate this is not null when we're on a pre-release?

tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
alafanechere: nit: Do we want to persist reports for metadata pipelines?
I believe dagger metadata pipelines  running in the CI will soon be removed as metadata validation will happen in test.

tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
```diff
@@ -402,24 +416,20 @@ async def __aexit__(
...
-            report_upload_exit_code = await remote_storage.upload_to_s3(
-                self.dagger_client, str(local_report_path), s3_key, os.environ["TEST_REPORTS_BUCKET_NAME"]
-            )
-            if report_upload_exit_code != 0:
-                self.logger.error("Uploading the report to S3 failed.")
+
+        suffix = f"{connector_name}/{connector_version}/output.json"
+        file_path_key = f"{self.report_output_prefix}/{suffix}"
+
+        await self.report.save(file_path_key)
```
alafanechere, line 423:
```suggestion
        await self.report.share(file_path_key)
```
What do you think about exposing a "mega" function on the report side that:
* Write report to remote storage if ci
* Write report to local file system if local
* Post a comment on the PR if ci
* Print out the report to the console
* Send a slack message if ci - on publish only but could be used for connector health reports.

It makes me think that we could revamp the connector health report logic on slack and just report to slack at runtime, one message per connector. 

We'd totally defer what needs to be reported and when  to the Report class.

Let me know what you think about orchestrating slack messages from Dagger instead of Dagster. The dagger pipeline could also retrieve the previous build statuses from GCS.
I believe this approach would totally remove the need for a "complete.json" file and downstream aggretion and processing on Dagster. The connector badge could also be uploaded from Dagger... Looks like a nice shortcut ATM. Wdyt?

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/connectors.py
alafanechere: It looks like an elegant approach and not a hack to me 😄 But let me know what you think of my suggestion about discarding `complete.json` and bet on live reporting from dagger

tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
bnchrch: @alafanechere I like this direction!

The only thing we loose is history
1. We wont have ✅❌✅✅✅❌
2. Our badges and connector specific status report history looking back

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
bnchrch: @alafanechere Yup right here: https://github.com/airbytehq/airbyte/blob/bnchrch/dagger/nightly-report-finished-step/tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py#L42

tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
alafanechere: We have timestamp on the key prefix, so we could `ls` the bucket with the prefix, order results by timestamp and pick the last 6 results.
But I believed this is a logic that's better suited for aggregation usecases in Dagster. And I like to have one big artifact produced daily that persists all our connectors tests results.
I'll let you decide if this is a beneficial shortcut compared to a Dagster job implementation that we could do later.

alafanechere (approved):
Approving to not block you, would love to see the `report.share` method come to life (even if it's not posting slack messages), but this can come in a subsequent PR.

PR #27060, opened 2023-06-06
🐛 Source Notion: added skipping 404 error for steam bloks

darynaishchenko:
## What
resolved: https://github.com/airbytehq/oncall/issues/2195
For some cases source-notion retrieves blocks ids which don't belong to the integration. 

## How
Was added handling for 404 error to skip this case and show log message.


lazebnyi (approved):
Nice!

pedroslopez (approved):
:shipit:

Issue #26748, opened 2023-05-30
Source Zendesk Support: New limits for offset-based pagination


gustavohwulee:
### Connector Name

source-zendesk-support

### Connector Version

0.3.0

### What step the error happened?

None

### Revelant information

The new limit that will be implemented on August 15, 2023 will break exiting OBP streams with over 10,000 records.

```
What is changing?

Beginning on August 15, 2023, OBP requests over the first 10,000 records (100 pages) 
will result in an error message: “400 Bad Request.” In order to request data sets larger than 10,000 records,
customers will need to transition to CBP.

OBP requests for fewer than 10,000 records will not be affected, nor will requests made using the CBP model.
```
https://support.zendesk.com/hc/en-us/articles/5591904358938-New-limits-for-offset-based-pagination

I'm not 100% sure which streams on zendesk connectors will be impacted because it looks like time-based extraction won't be limited by this change:

```
Time based Incremental Export is a special case and not considered either OBP or CBP. 
The new page 100 limit is not applicable to it.

You can export tickets using cursor-based exports or time-based exports. 
However, Zendesk strongly recommends using cursor-based exports for better performance.
```

### Contribute

- [X] Yes, I want to contribute

btkcodedev:
I would like to work on this

jrolom:
@igrankova: @btkcodedev  will pick this one and related issues #26055  and #16163

artem1205:
Fixed in https://github.com/airbytehq/airbyte/pull/26945

PR #27070, opened 2023-06-06
🐛Source Zendesk Support: Fix pagination limit and cursor pagination behaviour for Ticket Metrics stream

### LINKED ISSUE #16163: Source Zendesk Support: `ticket_metrics` pagination error
## Environment
- **Airbyte version**: 0.40.2
- **OS Version / Instance**: debian 11 bullseye / GCP e2
- **Deployment**: Docker
- **Source Connector and version**: Zendesk Support 0.2.15
- **Destination Connector and version**: BigQuery 1.1.15

## Current Behavior
I want to sync all the zendesk support tables with incremental + dedup when it's available and full refresh overwrite when it's not.
The first sync is working well.
But during the second the ticket metrics start to sync but the following error occurs : 
Response Code: 429, Response Text: Pagination limit reached for offset model
## Expected Behavior
Second sync should succeed

## Logs

[logs-135.txt](https://github.com/airbytehq/airbyte/files/9459720/logs-135.txt)

## Steps to Reproduce
1. Select all tables with incremental + dedup when it's available
2. Do a first sync
3. Launch a second sync

### LINKED ISSUE #26055: Source Zendesk Support: ticket_metrics stream pagination limit
### Connector Name

source-zendesk-support

### Connector Version

0.2.29

### What step the error happened?

During the sync

### Revelant information

`TicketMetrics` stream is defined as cursor pagination (https://github.com/airbytehq/airbyte/pull/13757) in order to avoid rate limits of the Zendesk API. This behaviour is currently not working as intended, the api calls are still paginated via offset mode.

Syncs that fetch more than 100k records will still reach pagination limits and return `Response Text: Pagination limit reached for offset model`. According to the [Zendesk documentation on cursor pagination](https://developer.zendesk.com/api-reference/introduction/pagination/#using-cursor-pagination) the correct parameter for page size is `page[size]` (instead of the current `per_page`), omitting this parameter will default to offset pagination.

### Relevant log output

_No response_

### Contribute

- [ ] Yes, I want to contribute

btkcodedev:
## What
Closes #26055
Closes #16163
Partial Close for #26748

## Bug Fix
* Added page_size as per documentation which addresses -> #26055
* Already stream is paginated via cursor but had a small issue with has_more boolean param, addresses -> #16163
* Hopefully the impact mentioned in #26748 is removed

## 🚨 User Impact 🚨
No breaking changes

## Pre-merge Checklist
<details open><summary><strong>Updating a connector</strong></summary>

### Community member or Airbyter [Yes]

- [x] Issue acceptance criteria met
- [x] Documentation which references the generator is updated as needed
+


airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/streams.py
```diff
@@ -652,13 +652,14 @@ class TicketForms(SourceZendeskSupportCursorPaginationStream):
...
-
+    page_size = 100
+    
     def next_page_token(self, response: requests.Response) -> Optional[Mapping[str, Any]]:
         """
         https://developer.zendesk.com/documentation/api-basics/pagination/paginating-through-lists-using-cursor-pagination/#when-to-stop-paginating
         """
         meta = response.json().get("meta", {})
-        return meta.get("after_cursor") if meta.get("has_more", False) else None
+        return meta.get("after_cursor") if meta.get("has_more", True) else None
```
marcosmarxm, line 662: Can you explain how the change you're doing fix the issue?

marcosmarxm (changes requested):

airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/streams.py
williamangelo: I believe `meta.get("has_more", False)` will return `False` if the `meta` dict does not have a `has_more` key in it. This will also yield the same result if `meta` has a key `has_more` with the value `False`, which is the expected value when pagination limit is reached.

What that means is that the return value of `next_page_token` will be `meta.get("after_cursor")` only when `has_more` is `True`

sh4sh: This logic makes sense to me @btkcodedev, we only need `after_cursor` to be returned when `has_more` is `True` (and there are more records to fetch)

Zendesk has a [pagination example](https://developer.zendesk.com/documentation/api-basics/pagination/paginating-through-lists-using-cursor-pagination/#python-example-with-the-after-cursor) that affirms us:

```
    if data['meta']['has_more']:
        params['page[after]'] = data['meta']['after_cursor']
    else:
        url = None
```

sh4sh:
Hey @btkcodedev I ran `read` with `--debug` on the latest release of this connector and requests seemed to correctly populate parameters for cursor pagination. 🤔 I'm not sure how it still works since your logic seems correct. It's possible I don't have enough data since my `ticket_metrics` stream only has 2 pages.

Did you come across bugs since https://github.com/airbytehq/airbyte/pull/26945 that inspired these fixes?

btkcodedev:
@sh4sh During my last check for reproducing issues,

Response Code: 429, Response Text: Pagination limit reached for offset model

Haven't tried after the migration to cursor Pagination, 
But next_page would return only None if has_more variable is false, 
It will paginate with default page[size] and make it unrecognizable, isn't it?

btkcodedev:
> Hi, @sh4sh During my last check for reproducing issues,
> `Response Code: 429, Response Text: Pagination limit reached for offset model`
> Is the response which I got. 
> 
> Haven't tried after the migration to cursor Pagination, 
> But `next_page_token` would return only None if `has_more` variable is false, 
> It will paginate with default `page[size]` and make it unrecognizable, isn't it? 
> 

Still unclear that how it works though 😄

airbyte-integrations/connectors/source-zendesk-support/source_zendesk_support/streams.py
sh4sh: I checked the [python docs](https://docs.python.org/3/library/stdtypes.html#dict.get) and it turns out `if meta.get("has_more", False)` actually just sets False as default, if `has_more` has a value then it will be fetched as is. 👍🏽

sh4sh (commented):
After review, these improvements were covered by https://github.com/airbytehq/airbyte/pull/26945/ and that implementation works well for me locally. For now, I don't see any need for any further modifications. I'll be closing this PR, but if we reproduce bugs or see behaviour that warrant these changes then we can look at re-opening.

PR #27071, opened 2023-06-06
CDK: Datetime format inferrer

### LINKED ISSUE #26903: CDK: Add datetime format inferrer
### What area the feature impact?

None

### Revelant Information

```
class DatetimeFormatInferrer:
    def accumulate(self, record: AirbyteRecordMessage):
        """
        Updates the internal state with another record
        """

     
    def get_inferred_datetime_formats(self) -> dict[str, str]:
        """
        Returns a list of detected datetime fields in the records (mapping of field name to format)
        """
```

The behavior is the following:
* When accumulate is called with the first record, all string fields in the record (no nested fields as this is not supported by the incremental sync module anyway) are checked against a hardcoded list of known formats whether they are parseable via the `DatetimeParser`.
* If parsing succeeds for a format, the format is stored in an internal state of the inferrer - the format search is stopped
* If parsing fails for all formats, the field is marked as non-parseable and skipped for all further calls of accumulate
* When accumulate is called with another record, all fields with their respective detected formats are checked whether they also exist and whether they still parse with the detected format. If this is not the case, the field is marked as non-parseable
* get_inferred_datetime_formats returns the internal state of detected formats so far

This will result in a list of “candidate cursor fields” that have been detected automatically.

The list of formats to check can be extended over time, but we can start with the ones that are currently used in connectors:
* `%Y-%m-%d`
* `%Y-%m-%d %H:%M:%S`
* `%Y-%m-%d %H:%M:%S.%f+00:00`
* `%Y-%m-%dT%H:%M:%S.%f%z`
* `%s`
* `%d/%m/%Y %H:%M`
* `%Y-%m`
* `%d-%m-%Y`
* `%Y-%m-%dT%H:%M:%SZ`
* `%Y-%m-%dT%H:%M:%S.000Z`

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/26903

This PR introduces a new class `DatetimeFormatInferrer` that can be used to detect record fields that might be dates (and qualify as cursor fields for incremental syncs).

The connector builder module is passing all records through the inferrer as part of the testread and returns the results.


girarda (approved):

airbyte-cdk/python/airbyte_cdk/utils/datetime_format_inferrer.py
```diff
@@ -0,0 +1,80 @@
...
+        if isinstance(value, int) and value in self._timestamp_heuristic_range:
+            return True
+        return False
+
+    def _matches_format(self, value: Any, format: str) -> bool:
+        """Checks if the value matches the format"""
+        try:
+            self._parser.parse(value, format)
+            return True
+        except Exception:
```
brianjlai, line 45: Should we use a more specific error to catch which is usually a better practice? A cursory test if we use bad inputs like a string ends up with a `ValueError`.

brianjlai (approved):

PR #27087, opened 2023-06-06
connectors-ci: mask secrets in GHA logs with ::add-mask::

### LINKED ISSUE #27088: connectors-ci: mask connector secrets in logs
On our connectors testing pipeline our internal ci_credentials library is used to download GSM secrets to file system.
In the context of dagger pipeline the interaction with this library is containerized.
This library uses a VERSION env var to modulate it's behavior:
If VERSION=dev:

Write secrets to current working directory: it's the behavior we want for dagger pipelines
Do not print out ::add-mask:: logs that are used to mask secrets in GHA logs
If VERSION != dev:

Write secrets to an absolute path relative to the GHA action runners file system
Print ::add-mask:: to stdout to make GHA redact the secrets from the logs.
We've originally set the env var to dev in our dagger pipeline to get secret written at the right path but we miss the secret masking capabilities while doing it.

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/27088

On our connectors testing pipeline our internal `ci_credentials` library is used to download GSM secrets to file system.
In the context of dagger pipeline the interaction with this library is containerized.
This library uses a `VERSION` env var to modulate it's behavior:
If `VERSION=dev`:
* Write secrets to current working directory: it's the behavior we want for dagger pipelines
* Do not print out `::add-mask::` logs that are used to mask secrets in GHA logs

If `VERSION != dev`:
* Write secrets to an absolute path relative to the GHA action runners file system
* Print `::add-mask::` to stdout to make GHA redact the secrets from the logs.

We've originally set the env var to `dev` in our dagger pipeline to get secret written at the right path but we miss the secret masking capabilities while doing it.

## How
This PR changes `ci_credentials` and its usage in our dagger pipeline to:

* Write secrets to a correct path while not using `VERSION=dev`
* Write secrets to mask to a /tmp file that dagger pipeline can read and print out to stdout without any logger prefix.
* Increase the list of keys considered as secrets by consuming our spec mask.

### DIFF
--- a/airbyte-integrations/connectors/source-openweather/metadata.yaml
+++ b/airbyte-integrations/connectors/source-openweather/metadata.yaml
@@ -13,7 +13,7 @@ data:
       enabled: true
     oss:
       enabled: true
-  releaseStage: generally_available
+  releaseStage: alpha
   documentationUrl: https://docs.airbyte.com/integrations/sources/openweather
   tags:
     - language:python
--- a/airbyte-integrations/connectors/source-openweather/metadata.yaml
+++ b/airbyte-integrations/connectors/source-openweather/metadata.yaml
@@ -13,7 +13,7 @@ data:
       enabled: true
     oss:
       enabled: true
-  releaseStage: alpha
+  releaseStage: generally_available
   documentationUrl: https://docs.airbyte.com/integrations/sources/openweather
   tags:
     - language:python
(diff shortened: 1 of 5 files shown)


alafanechere:
🎉 I confirm it work. In a[ test commit](https://github.com/airbytehq/airbyte/pull/27087/commits/547e7ba23c2df98f075801f60500e0646e9175fa) I printed secrets on purpose to the logs and artificially modified a connector in .
The secrets are redacted from [the GHA logs. ](https://github.com/airbytehq/airbyte/actions/runs/5193213799)
Screen Shot 2023-06-06 at 22 52 04

c-p-b (approved):

Issue #25321, opened 2023-04-19
Better UX when importing the schema


lmossman:
## Context
As identified in the most recent Build-a-connector session, the UX around importing a schema has some areas that can be improved:
- Its not clear what is happening when you import from the testing panel, since the only thing that changes is that the Import button goes away, unless the user happens to already have the Declared Schema tab open
- Users always have to manually import schema changes, which can be cumbersome as the schema continues to change while they are developing the connector
- Detected schema sets fields to `null` if that is the only value seen for these fields in the response, which is likely incorrect and is easy for the user to miss

## Acceptance Criteria
- Users are only asked to review the schema when publishing. If the schema of a stream has not changed since the last publish, exclude that stream from this "review" step
- Detect when a stream is "dirty" (i.e. has not been tested since changes were made to it), and prevent publishing if there are any dirty streams
  - Alternative: run the tests on the dirty streams with the current testing values 
- When reviewing a stream with a changed schema, require the user to change the type of any fields which were detected to be `null` type
- Add link to schema docs page somewhere in the new "review" component

lmossman:
Grooming notes:
- Allow users to toggle between "Automatic detection" and "Manual" mode, default to Automatic
- If there is a field that can't be inferred (e.g. `null`), force the mode to be manual
- When publishing, if any stream has not done a test read, we run the test read for them and update the schema if in automatic mode

- Alternative: Only ask the user to review the schema when they are publishing. If the schema hasn't changed since the last publish, ignore that stream during this step.
  - Use the last detected schema from the last test read
  - Detect if a stream has not been tested since the last time a change was made to it - if so test the stream on publish or have an error on the publish button requiring the user to test before publishing

lmossman:
Grooming notes:
- The above approach is fairly high effort and introduces several new concepts
- A more incremental approach to get most of the value here would be the following:
  - Default to automatically import the detected schema as the declared schema (store a flag in the schema, e.g. `automatic_detection: true`), but allow the user to manually edit the schema as well by switching this automatic mode off
  - On publish, detect if the new schema of a stream is incompatible with the schema from the last published manifest. If so, warn the user that this will cause issues in the destination if they just publish.
  -  Ignore the `null` type fields, i.e. don't include them in the schema

lmossman:
I have tested out the case where some records returned by the API have `null` value for some field, while other records have a string value for the same field. I tested setting the type of this field in the schema to each of the following:
- `"type": "null"`
[...]

PR #27094, opened 2023-06-07
Add metadata field to low-code schema

lmossman:
## What
Relates to https://github.com/airbytehq/airbyte/issues/25321

We now have a use case where the Connector Builder needs to store some UI-specific data in the connector manifest, as the manifest is the single resource that the builder uses to persist its state.

Specifically, the Connector Builder needs the ability to store an "auto-import schema" boolean flag per stream in the connector manifest. After discussion with Joe and Maxime, we agreed that persisting this in a generic top-level `metadata` field of the connector manifest would be the most straightforward approach, for the following reasons:
- Doesn't require any changes to the persistence layer for connector builder connectors (still just persists the manifest YAML)
- Doesn't affect any existing components in the declarative schema
- Is not specific to the builder - since this is just a generic `metadata` field, technically any consumer of low-code manifests could use this field for whatever they want
- Manifests without any `metadata` set will still be compatible with the Builder - it will just fall back to the default values that it uses when first initializing the metadata values

## 🚨 User Impact 🚨
No breaking changes, since this just adds an optional field to the schema.

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -27,6 +27,10 @@ properties:
     type: object
   spec:
     "$ref": "#/definitions/Spec"
+  metadata:
+    type: object
+    description: For internal use only - used by consumers of declarative manifests for storing related metadata
+    additionalProperties: true
 additionalProperties: false
 definitions:
   AddedFieldDefinition:
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -29,7 +29,7 @@ properties:
     "$ref": "#/definitions/Spec"
   metadata:
     type: object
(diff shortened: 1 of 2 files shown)


girarda (approved):
sounds good to me

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -27,6 +27,10 @@ properties:
     type: object
   spec:
     "$ref": "#/definitions/Spec"
+  metadata:
+    type: object
+    description: For internal use only - used by consumers of declarative manifests for storing related metadata
```
sherifnada, line 32: what does `internal` mean in this context?

lmossman: I could reword it to say "For internal Airbyte use only" ?

sherifnada: That would be useful! but yeah just asking a question, not really requesting changes

lmossman: Added "Airbyte" to the description and added a note telling people not to modify this field manually

flash1293 (approved):
LGTM

Issue #26425, opened 2023-05-23
Construct Initial Debezium state for ctid based snapshot


subodh1810:
Implement debezium cdc state containing the right LSN and other parameters that would be passed to debezium after ctid based snapshot is complete so that it can start processing WAL logs from that LSN.

PR #27106, opened 2023-06-07
postgres-cdc: implementation to construct initial debezium state manually to skip snapshot

subodh1810:
Issue: https://github.com/airbytehq/airbyte/issues/26425

This is right now unused piece of code but once we have the ctid implementation, we can call the method to get the initial state to pass onto Debezium. I have also tested this with Debezium and on passing the state generated by the method to Debezium, it accepts it and skips the snapshot phase and starts processing WAL from the given LSN. 

Am not going to publish a new version of Postgres cause this code is not being executed right now


airbyte-db/db-lib/src/main/java/io/airbyte/db/PostgresUtils.java
```diff
@@ -17,9 +17,9 @@
 public class PostgresUtils {
 
   public static PgLsn getLsn(final JdbcDatabase database) throws SQLException {
-    // pg version 10+.
+    // pg version >= 10. For versions < 10 use query select * from pg_current_xlog_location()
```
akashkulk, line 20: I see. Do we have any plans of supporting version < PG 10 for CDC workflows?

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/postgres/PostgresDebeziumStateUtil.java
```diff
@@ -224,4 +233,39 @@ private OptionalLong extractLsn(final Set<Partition> partitions,
 
   }
 
+  public JsonNode constructInitialDebeziumState(final JdbcDatabase database, final String dbName) {
```
akashkulk, line 236: nit : Could you add a quick comment here?

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/postgres/PostgresDebeziumStateUtil.java
```diff
@@ -224,4 +233,39 @@ private OptionalLong extractLsn(final Set<Partition> partitions,
...
+    try {
+      return format(currentXLogLocation(database), currentTransactionId(database), dbName, Instant.now());
+    } catch (SQLException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @VisibleForTesting
+  public JsonNode format(final Long currentXLogLocation, final Long currentTransactionId, final String dbName, final Instant time) {
```
akashkulk, line 245: Remind me again : why do we need current txn id in the DBZ/CDC state?

akashkulk (approved):
Approved - just some minor question/comments

airbyte-db/db-lib/src/main/java/io/airbyte/db/PostgresUtils.java
subodh1810: I have not seen any issues till yet + Support for Postgres versions < 10 has already ended (see [this](https://endoflife.date/postgresql)) so I doubt anyone would be using it right now. I just put the comment that in case someone raises an issue with 9.4 we know what needs to be done.

airbyte-integrations/bases/debezium/src/main/java/io/airbyte/integrations/debezium/internals/postgres/PostgresDebeziumStateUtil.java
subodh1810: Debezium expects it! Didnt deep dive into how it uses it

PR #27107, opened 2023-06-07
connectors-ci: make source-file testable in airbyte-ci

alafanechere:
## What
Relates to #25053 
### Fixing `source-file`
`source-file` integration tests uses docker-compose to spin up a sftp container used for testing.
To make the container orchestration  and connection work in `airbyte-ci` we should bind python integration test to the global docker host.
The docker host should also have access to the file we want to mount of sftp, to do so we move them to the /tmp folder so that both the container under test, docker host have access to it and the docker host can mount them to the `sftp` containers.

### Fixing `source-file-secure`
source-file-secure has an implicit dependency to source-file: its docker image is based on source-file docker image in which the source-file package was installed.
But, outside of the dockerized execution, executing  `import source_file` in python in `source-file-secure` fails because our pipeline did not handle installing other connector package.
To fix this problem:
* Dependency to source-file in source-file-secure is declared explicitely inside its `setup.py`  (@evantahler we previously chatted about the benefits of explicitely variants  dependencies in `setup.py`)
* We detect dependencies to mount in the dagger pipeline by running `python setup.py egg_ingo`
* We mount the required dependencies when installing the `source-file-secure` in the test environment.
* When building the `source-file-secure` docker image we mount and install the same dependencies.


airbyte-integrations/connectors/source-file-secure/Dockerfile
```diff
@@ -1,3 +1,11 @@
+### WARNING ###
+# This Dockerfile will soon be deprecated.
+# It is not used to build the connector image we publish to DockerHub.
+# The new logic to build the connector image is declared with Dagger here:
+# https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py#L771
+
+# If you need to add a custom logic to build your connector image, you can do it by adding a finalize_build.sh or finalize_build.py script in the connector folder.
+# Please reach out to the Connectors Operations team if you have any question.
```
alafanechere, line 8: `source-file-secure` is now fully built from Dagger.  As `source-file` needs to be mounted at build time it was easier to accomplish it with Dagger.

airbyte-integrations/connectors/source-file-secure/setup.py
```diff
@@ -2,9 +2,20 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
 
+import os
+from pathlib import Path
 
 from setuptools import find_packages, setup
 
+
+def local_dependency(name: str) -> str:
```
alafanechere, line 11: Compute the absolute path to the local dependencies. In dagger pipelines we mount local dependencies to `/local_dependencies`.

airbyte-integrations/connectors/source-file-secure/source_file_secure/source.py
```diff
@@ -4,34 +4,19 @@
...
-try:
-    import source_file.source
-except ModuleNotFoundError:
-    current_dir = os.path.dirname(os.path.abspath(__file__))
-    parent_source_local = os.path.join(current_dir, "../../source-file")
-    if os.path.isdir(parent_source_local):
-        sys.path.append(parent_source_local)
-    else:
-        raise RuntimeError("not found parent source folder")
-    import source_file.source
```
alafanechere, line 24: This is not required anymore as the source file connector is installed as a normal python deps with pip install.

evantahler:
While out of scope for this PR, I have the opinion that using local docker containers for testing connectors is bad for a number of reasons:
* anti-patterns like this
* slow to spin up and down containers for each test/suite
* false sense of isolation. we are in general poor at clearning up files/dbs after tests.  This is a problem masked by destroying the container which appears when moving to a hosted DB, like testing snowflake
* unobservable.  If a test fails, it is hard to connect to the DB or S3 bucked to see what went wrong

To that end, I would propose that we work to remove all test containers like this, and set up persistent test hosts for all connectors

airbyte-integrations/connectors/source-file-secure/Dockerfile
```diff
@@ -1,4 +1,12 @@
-FROM airbyte/source-file:0.3.9
+### WARNING ###
+# This Dockerfile will soon be deprecated.
+# It is not used to build the connector image we publish to DockerHub.
+# The new logic to build the connector image is declared with Dagger here:
+# https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py#L771
+
+# If you need to add a custom logic to build your connector image, you can do it by adding a finalize_build.sh or finalize_build.py script in the connector folder.
+# Please reach out to the Connectors Operations team if you have any question.
+FROM airbyte/source-file:0.3.10
```
evantahler, line 9: Nice.

airbyte-integrations/connectors/source-file-secure/setup.py
```diff
@@ -21,8 +32,10 @@
     "xlrd==2.0.1",
     "openpyxl==3.0.10",
     "pyxlsb==1.0.9",
+    local_dependency("source-file"),
```
evantahler, line 35: 👍

airbyte-integrations/connectors/source-file/setup.py
```diff
@@ -24,7 +24,7 @@
     "pyxlsb==1.0.9",
 ]
 
-TEST_REQUIREMENTS = ["pytest~=6.2", "pytest-docker~=1.0.0", "pytest-mock~=3.6.1"]
+TEST_REQUIREMENTS = ["pytest~=6.2", "pytest-docker~=1.0.0", "pytest-mock~=3.6.1", "docker-compose"]
```
evantahler, line 27: Is this a python package to work with docker?

evantahler (approved):
👍 from me, mainly about explicit dependencies! 
I'd wait for the other reviews on the 🐍 code before merging

airbyte-integrations/connectors/source-file-secure/setup.py
bnchrch: Oh dear python!

I like this solution.

My only reservations are having
1. Dagger relying on `file://` to mark relative dependencies, or pip -e as this isnt always the case ([example](https://github.com/airbytehq/airbyte/blob/master/airbyte-ci/connectors/metadata_service/orchestrator/pyproject.toml#L18
))
2. our connector setup.py's relying on knowing how dagger works under the hood.

So I want to ask a question for #1 and propose a possible idea for 2

1. I dont think we can do away with this, but how come we dont mount the relative import to the same relative location in the dagger file tree?

2. What if, in the dagger machine, we symlinked the local host path to the dagger `/local_dependencies` version? Would remove the need for having `DAGGER_BUILD` in our connectors folder.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -99,6 +101,47 @@ def with_python_package(
...
+    """
+    python_environment = with_python_base(context)
+    container = with_python_package(context, python_environment, package_source_code_path)
+
+    # Find local dependencies in setup.py
+    setup_dependency_paths = []
+    if await get_file_contents(container, "setup.py"):
+        container_with_egg_info = container.with_exec(["python", "setup.py", "egg_info"])
+        egg_info_output = await container_with_egg_info.stdout()
+        for line in egg_info_output.split("\n"):
```
bnchrch, line 123: This must ahve been painful, well done

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -118,20 +161,19 @@ async def with_installed_python_package(
...
-        container = container.with_exec(install_local_requirements_cmd)
 
-    container = container.with_exec(install_connector_package_cmd)
+    setup_dependencies, requirements_dependencies = await find_local_python_dependencies(context, package_source_code_path)
+    for dependency_directory in setup_dependencies + requirements_dependencies:
+        container = container.with_mounted_directory("/" + dependency_directory, context.get_repo_dir(dependency_directory))
+
+    if await get_file_contents(container, "setup.py"):
+        container = container.with_exec(install_connector_package_cmd)
+    if await get_file_contents(container, "requirements.txt"):
```
bnchrch, line 175: Nit: We likely will need on for pyproject.toml at some point.

bnchrch (approved):
Great work here @alafanechere 

This must have been annoying to work through

I had one alternative idea, but Im not positive it would work and in the spirit of "done not perfect"

airbyte-integrations/connectors/source-file/setup.py
alafanechere: The source-file integration test use docker-compose without specifying it as a test dependency. I added this so that our pipeline installs it within the test environment, that does not have `docker-compose` by default.

alafanechere:
> My only reservations are having
* Dagger relying on file:// to mark relative dependencies, or pip -e as this isnt always the case ([example](https://github.com/airbytehq/airbyte/blob/master/airbyte-ci/connectors/metadata_service/orchestrator/pyproject.toml#L18))
* our connector setup.py's relying on knowing how dagger works under the hood.

I believe there's indeed a cleaner solution I'd like to tackle later:
Have a a single function that buids a python connector container by:
* Detecting its local dependencies from pyproject.toml, requirements.txt or setup.py
* Mounting these dependencies + the connector code with path relative to the airbyte repo
* Install the dependencies according to the context (eg - install tests deps if it's the container to run tests, or install a specific cdk version if the code change has cdk change etc.) 

Then we'll have a same base container on top of which we can:
* run unit/integration tests
* build for CAT and publish

I did not take this approach right now because it was too much refacto, and as we still rely on dockerfiles for python connectors  (except source-file-secure) we still have a different container for unit/integration tests (dynamically built with dagger) and one for build and publish based on the Dockerfile.

This is definitely something we'll tackle when we'll want to remove dockerfiles for Python connectors.

PR #27109, opened 2023-06-07
📝 Fix broken connector documentation links

joeykmh:
## What
Fixes broken doc links in `metadata.yml` files, which resulted in the docs not displaying correctly in-app or in the [connector catalog](https://docs.airbyte.com/integrations/) for the following connectors:

- `source/clickup` (cloud)
- `source/exchange-rates` (cloud)
- `source/hello-baton`
- `source/kustomer` (cloud)
- `source/qonto`
- `source/zapier-supported-storage`
- `destination/local-csv`
- `destination/local-sqlite`
- `destination/azure-blob-storage` (cloud)

I'm assuming we don't need to bump connector versions for this, but please lmk if that's not the case!


evantahler (approved):
Thanks!

Issue #26753, opened 2023-05-30
File CDK: Implement get_json_schema and parse_records for CSV files


clnoll:
#### `get_json_schema`

Return a JSON Schema for a group of records as follows:
1. Create an empty master schema.
For each file (up to 100 files):
2. Open the file and read the first row; we assume this is the header.
3. Update the master schema with all columns from the row (with type as `string`).

This should be written from scratch rather than using the `source_files_abstract` [csv_parser](https://github.com/airbytehq/airbyte/blob/3aa0c284e2631a94e5b2e345fefd70220e15638d/airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/formats/csv_parser.py). `source_files_abstract` 1) only infers the schema from one file, 2) has options for making all values strings or inferring other types, and 3) uses multiprocessing and other hacks to handle issues introduced by using a user-input `block_size`.

This will include a utility to be used by all parsers discovers the correct block size to use for the parser and schema inference to account for all of the columns, without attempting to read the entire file.

#### `parse_records`
Create a parser for CSV file types. The `source_files_abstract` `csv_parser` may be useful but will have to be modified to remove options that we aren't supporting.

This ticket is for the basic implementation of these methods. We will support additional configuration options for CSVs, but those should be done in a later ticket, since this ticket is a blocker for all the other parsers.

#### Acceptance Criteria
- [ ] `discover` returns a configured catalog with the master schema for streams consisting of CSV files.
- [ ] `read` returns `AirbyteRecordMessage`s for streams consisting of CSV files.
- [ ] Errors are handled appropriately and have test coverage; error messages are clear.

clnoll:
Grooming notes:
- Break out separate ticket for block size investigation. 
- Consider not using pyarrow for csv.

PR #27122, opened 2023-06-07
File-based CDK module scaffolding

clnoll:
## What
This is the first phase of the file-based CDK, and includes base implementations of the major classes and methods required to create a file-based connector.

This also includes a test framework that will be used for creating end-to-end tests for various scenarios, file types, etc. as the work progresses. The framework is based on an in-memory files source (providing the first concrete example of a file-based source), and includes a test demonstrating discover and read for a stream of CSV files. 
 
Note: there is a basic implementation of the CSV parser & schema inference, for the purpose of demonstrating how the test suite is used. This will be updated as part of https://github.com/airbytehq/airbyte/issues/26753.

## Recommended reading order
Production code:
1. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py`
2. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream.py`
4. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_config.py`
5. `airbyte-cdk/python/airbyte_cdk/sources/file_based/availability_strategy.py`
6. `airbyte-cdk/python/airbyte_cdk/sources/file_based/discovery_concurrency_policy.py`
7. `airbyte-cdk/python/airbyte_cdk/sources/file_based/remote_file.py`
8. `airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py`
9. `airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_validation_policies.py`
10. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/base_file_type_parser.py` & File type-specific parsers
11. `airbyte-cdk/python/airbyte_cdk/sources/file_based/exceptions.py`

Test framework:
1. `airbyte-cdk/python/unit_tests/sources/file_based/test_source.py`
2. `airbyte-cdk/python/unit_tests/sources/file_based/in_memory_files_source.py`
3. `airbyte-cdk/python/unit_tests/sources/file_based/scenarios/_base_scenario.py`
4. `airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py`
5. `airbyte-cdk/python/unit_tests/sources/file_based/helpers.py`

See the [Tech Spec](https://docs.google.com/document/d/12nDEOAGCoXG0LAyiBe2fCx9CZfSfGNlHGr5Xrphk4qE/edit#) for details and background.


girarda (commented):
This is great! Left a few comments and questions, but I like where this is heading

girarda (commented):
This is great! Left a few comments and questions, but I like where this is heading

airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+def conforms_to_schema(record: Mapping[str, Any], schema: Mapping[str, Any]) -> bool:
+    """
+    Return true iff the record conforms to the supplied schema.
+
+    The record conforms to the supplied schema iff:
+    - All columns in the record are in the schema.
+    - For every column in the record, that column's type is equal to or narrower than the same column's
+      type in the schema.
+    """
+    ...
```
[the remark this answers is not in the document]
clnoll: Yes, this will be part of the [schema validation policies](https://app.zenhub.com/workspaces/connector-extensibility-6257455decf3920012f4e872/issues/gh/airbytehq/airbyte/26762) ticket.

airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def type_mapping_to_jsonschema(type_mapping: Mapping[str, Any]) -> Mapping[str, Any]:
+    """
+    Return the user input schema (type mapping), transformed to JSON Schema format.
+    """
+    ...
```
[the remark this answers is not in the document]
clnoll: Yes, this will be part of the [user input schema](https://app.zenhub.com/workspaces/connector-extensibility-6257455decf3920012f4e872/issues/gh/airbytehq/airbyte/26759) ticket.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/base_file_type_parser.py
```diff
@@ -0,0 +1,32 @@
...
+class FileTypeParser(ABC):
+    """
+    An abstract class containing methods that must be implemented for each
+    supported file type.
+    """
+
+    @abstractmethod
+    async def infer_schema(
+        self, file: RemoteFile, stream_reader: AbstractFileBasedStreamReader
+    ) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
clnoll: Makes sense to me. Done.

airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py
```diff
@@ -0,0 +1,76 @@
...
+            merged_schema[k2] = _choose_wider_type(k1, t1, t2)
+
+    return merged_schema
+
+
+def _choose_wider_type(key, t1: Any, t2: Any) -> Any:
+    # TODO: more testing
+    if t1 is None or t2 is None:
+        return t1 or t2
+    elif t1 in ("list", "object"):
```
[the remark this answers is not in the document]
clnoll: I also modified this function to only include strings. It'll be updated as we support parsers with other types.

clnoll (commented):
Thanks @girarda! Updated based on your comments.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream.py
```diff
@@ -0,0 +1,163 @@
...
+        return base_schema
+
+    async def _infer_file_schema(self, file: RemoteFile) -> Mapping[str, Any]:
+        try:
+            return await self._get_parser(self.config.file_type).infer_schema(file, self.stream_reader)
+        except Exception as exc:
+            raise SchemaInferenceError(f"Error inferring schema for file: {file.uri}") from exc
+
+    @staticmethod
+    def _get_parser(file_type: FileType) -> FileTypeParser:
```
sherifnada, line 157: is it crazy to say that this shouldn't be static/private, and that a user implementing a file-based connector should be able to supply a parser to support a new filetype in their connector without having to upstream it to the CDK? 

if this is a lot of work, we should do it as an additive change. But if it's easy, might as well do it out the gate. It seems to me like this particular function should be extensible by the user

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_config.py
```diff
@@ -0,0 +1,47 @@
...
+from typing import Any, List, Mapping, Optional
+
+from airbyte_cdk.models import ConfiguredAirbyteCatalog
+from airbyte_cdk.sources.file_based.exceptions import ConfigValidationError
+from airbyte_cdk.sources.file_based.remote_file import FileType
+from airbyte_cdk.sources.file_based.schema_validation_policies import UserValidationPolicies
+
+
+@dataclass
+class FileBasedStreamConfig:
```
sherifnada, line 15: I think we can get rid of the parsing code if we define this as a `BaseModel`, and rely on pydantic to parse . Is that desirable, or do we want to avoid pydantic in this context?

airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py
```diff
@@ -0,0 +1,78 @@
...
+
+def merge_schemas(schema1: Dict[str, JsonSchemaType], schema2: Dict[str, JsonSchemaType]) -> Dict[str, JsonSchemaType]:
+    """
+    Returns a new dictionary that contains schema1 and schema2.
+
+    Schemas are merged as follows
+    - If a key is in one schema but not the other, add it to the base schema with its existing type.
+    - If a key is in both schemas but with different types, use the wider type.
+    - If the type is a list in one schema but a different type of element in the other schema, raise an exception.
+    - If the type is an object in both schemas but the objects are different raise an exception.
```
sherifnada, line 24: Not saying we should implement it here (i'd rather we wait till we get requests for it) but I'm curious if this is a practical or philosophical concern e.g: why not recursively merge_schemas on the objects?

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/_base_scenario.py
```diff
@@ -0,0 +1,37 @@
...
+    expected_records: Dict[str, Any]
+
+    def __init__(self):
+        self.source = InMemoryFilesSource(self.files)
+        self.validate()
+
+    def validate(self):
+        streams = {s["name"] for s in self.config["streams"]}
+        expected_streams = {s["name"] for s in self.expected_catalog["streams"]}
+        assert expected_streams <= streams
```
sherifnada, line 24: n00b question since I haven't run this myself: if a class implementing this base scenario fails, is it easy to know exactly which test case failed? e.g: when we do `pytest.mark.parametrize` you can give individual names to test cases and have individualized error messages and an ❌ shows on the exact test case which failed etc.. which I think is a useful thing. Is that also the case here or do all these inputs show up as one big test?

airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
```diff
@@ -0,0 +1,77 @@
...
+    scenario = MultiCsvNFilesExceedsLimitForSchemaInferenceTestScenario()
+    with mock.patch(
+        "airbyte_cdk.sources.file_based.file_based_stream.MAX_N_FILES_FOR_STREAM_SCHEMA_INFERENCE",
+        scenario.schema_inference_limit,
+    ):
+        assert discover(capsys, tmp_path, scenario) == scenario.expected_catalog
+
+
+def discover(capsys, tmp_path, scenario) -> Dict[str, Any]:
+    launch(
```
sherifnada, line 48: why not call `discover` directly instead of launching a process?

sherifnada (commented):
I like how things are getting laid out here. Great progress! 

A more macro concern I have with this architecture is that it should be more oriented towards stream-level dependency injection. Essentially, I shouldn't need to change anything in the framework to support something specific to my connector. This includes: 
* Using a different logic for `merge_schemas`
* Supporting a new file type specific to my connector and not have to upstream it to the CDK
* Customizing or extending CDK-provided file parsers
* Writing custom error messages when a connection check fails (I should be able to have access to the exception object which generated the error if a read op fails, for example, in order to map that exception's type or stack trace to a specific end-user-friendly message)

do you have thoughts on how we could achieve that herE?

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/_base_scenario.py
maxi297: As long as the scenario is instantiated as part of a test function (either a `pytest.mark.parametrize` of a stand-alone), I don't see this problem occurring

maxi297 (commented):
Here are a couple comments. I also tried running `test_scenarios.py::test_discover_with_n_files_exceeding_limit` to answer Sherif's comment and I got the following error:

```
ImportError while importing test module '/Users/maxime/tmp/tmp2/airbyte/airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/Users/maxime/.pyenv/versions/3.9.11/lib/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
unit_tests/sources/file_based/test_scenarios.py:12: in <module>
    from unit_tests.sources.file_based.scenarios.csv_scenarios import (
unit_tests/sources/file_based/scenarios/csv_scenarios.py:5: in <module>
    from unit_tests.sources.file_based.scenarios._base_scenario import BaseTestScenario
unit_tests/sources/file_based/scenarios/_base_scenario.py:7: in <module>
    from unit_tests.sources.file_based.in_memory_files_source import InMemoryFilesSource
unit_tests/sources/file_based/in_memory_files_source.py:10: in <module>
    from airbyte_cdk.sources.file_based.file_based_source import AbstractFileBasedSource
airbyte_cdk/sources/file_based/file_based_source.py:16: in <module>
    from airbyte_cdk.sources.file_based.file_based_stream import FileBasedStream
airbyte_cdk/sources/file_based/file_based_stream.py:17: in <module>
    from airbyte_cdk.sources.file_based.file_types import AvroParser, CsvParser, FileTypeParser, JsonlParser, ParquetParser
airbyte_cdk/sources/file_based/file_types/__init__.py:1: in <module>
    from .avro_parser import AvroParser
airbyte_cdk/sources/file_based/file_types/avro_parser.py:8: in <module>
    from airbyte_cdk.sources.file_based.file_types import FileTypeParser
E   ImportError: cannot import name 'FileTypeParser' from partially initialized module 'airbyte_cdk.sources.file_based.file_types' (most likely due to a circular import) (/Users/maxime/tmp/tmp2/airbyte/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/__init__.py)
```

Repro steps:
```
git clone git@github.com:airbytehq/airbyte.git
cd airbyte/airbyte-cdk/python
git checkout file-cdk-scaffolding
python -m venv .venv
source .venv/bin/activate
pip install -e ".[dev]"
pip install -e ".[tests]"
python -m pytest unit_tests/sources/file_based/test_scenarios.py::test_discover_with_n_files_exceeding_limit
```

I didn't investigate much. Not sure if I'm doing anything wrong...

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_config.py
clnoll: I like it, updated!

airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py
clnoll: That was the original plan but during grooming we'd agreed not to support nested schemas. I took that to mean that we would just give these a generic `object` type. I made a P2 line item in the proposal for revisiting nested schemas in the future, but can add revisiting it to one of the non-CSV parser tickets if it feels higher-priority than that.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -0,0 +1,61 @@
...
+from fnmatch import fnmatch
+from io import IOBase
+from typing import Any, List, Mapping
+
+from airbyte_cdk.sources.file_based.remote_file import RemoteFile
+
+
+class AbstractFileBasedStreamReader(ABC):
+    def __int__(self, config: Mapping[str, Any]):
+        self.config = config
```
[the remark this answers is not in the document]
clnoll: Agreed - I've updated `AbstractFileBasedStreamReader` to inherit from `pydantic.BaseModel`. When a new connector is being created, its implementation of `AbstractFileBasedStreamReader` will define the required configuration and pydantic will validate it.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream.py
clnoll: This shouldn't be a lot of work. I'll look into this as part of a reorganization to give us more stream-level dependency injection.

clnoll (commented):
Thanks for the comments @sherifnada! You bring up a really good point about dependency injection so I'm going to look into reorganizing this to enable more connector-specific dependencies. I think this can be done fairly easily by changing `FileBasedStream` to be a default, which is defined on `FileBasedSource`, so the user can swap it out if they want. (There will also be a bit more reorg to bubble up the other components you mentioned so they're accessible to the user.) In the meantime I've addressed all of them comments except the ability to supply a parser, since that's relevant to the dependency injection reorg.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@dataclass
+class AbstractFileBasedSource(AbstractSource, ABC):
+    """
+    All file-based sources must extend this class, implementing `stream_reader()`.
+    """
+
+    @abstractmethod
+    def stream_reader(self, config: Mapping[str, Any]) -> AbstractFileBasedStreamReader:
```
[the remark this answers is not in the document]
clnoll: SGTM, updated.

clnoll (commented):
Thanks for the comments @maxi297, I've taken your suggestions. Regarding the circular import error, that popped up temporarily for me but should be fixed.

clnoll:
@sherifnada I've reorganized this to facilitate dependency injection. Specifically, the class of the FileBasedStream is now configurable (so users can implement their own stream-related logic such as merging schemas), and is initialized with an AvailabilityStrategy (giving connector developers the opportunity to configure different error messages), a DiscoveryPolicy (for configurability of the number of concurrent requests and number of files used for discovery), and a dictionary mapping file types to the parsers that are available for the connector.

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/abstract_file_based_stream.py
```diff
@@ -0,0 +1,113 @@
...
+    @abstractmethod
+    def primary_key(self) -> Optional[Union[str, List[str], List[List[str]]]]:
+        ...
+
+    @abstractmethod
+    def read_records(
+        self,
+        sync_mode: SyncMode,
+        cursor_field: List[str] = None,
+        stream_slice: Mapping[str, Any] = None,
```
girarda, line 65: It would be nice to replace stream_slice and stream_state with actual objects.

I think the benefits for file-based connectors would be worth it since they'll all share the same state structure. We might as well make it explicit and deal with structures instead of arbitrary Mappings

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @property
+    def primary_key(self) -> Optional[Union[str, List[str], List[List[str]]]]:
+        return self.config.primary_key
+
+    def read_records(
+        self,
+        sync_mode: SyncMode,
+        cursor_field: List[str] = None,
+        stream_slice: Mapping[str, Any] = None,
```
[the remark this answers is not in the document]
girarda: reading through this with fresh eyes...

I think we should use the stream_slice to decide which files to read.

The flow I'd expect is
1. Given a stream_state, `stream_slices` generates partitions/slices to process (probably 1 file per slice to leverage the default checkpointing behavior
2. `read_records` uses the stream_slice instead of the stream_state to know which file to read. 

Using the stream_slice instead of the state will allow us to parallelize the files and streams by pushing all the parallelization logic to the source

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -0,0 +1,91 @@
...
+    All file-based sources must provide a `stream_reader`.
+    """
+
+    def __init__(
+        self,
+        stream_reader: AbstractFileBasedStreamReader,
+        availability_strategy: AbstractFileBasedAvailabilityStrategy,
+        discovery_policy: AbstractDiscoveryPolicy = DefaultDiscoveryPolicy(),
+        parsers: Dict[FileType, FileTypeParser] = None,
+        stream_cls: Type[AbstractFileBasedStream] = DefaultFileBasedStream,
```
brianjlai, line 38: I keep getting hung up on this parameter. It feels a weird to me that we need to pass the type of FileBasedStream here because it makes it add another layer of obfuscation about what type it will actually be when we build the streams in the below `streams()` method. 

As in, let's say we have an S3 source that inherits FileBasedSource(). When we first `__init__()` the S3 file source, we also need to pass in what type of stream to be creating when first building the source during runtime. Whereas, if the S3 file source were to just reimplement the `streams()` method, it is statically specified in code exactly what type of stream it will be creating. It's readable in code either as the parameter or in the streams() method, but just feels a bit more complicated unless we needed the type to be dynamic at runtime or something.


Maybe if you cold provide a small code snippet of what it looks like to initialize a new file source that needs construct a new `AbstractFileBasedStream` that might help me understand the reason why we need to pass it in this way.

brianjlai (commented):
sorry to get in on this one a little late, added general points to clarify on and a few questions I had. I think some of my questions are a result of how the code has been refactored since the tech spec, but in agreement with the direction its going in

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
sherifnada: I think I agree with Brian. It feels circuitous to pass the class when the dev can init and return it directly from `streams` (that direction also makes it easier for the dev to keep). It potentially complicates the story for how we init those streams, but I think that's solvable

Separately, I really second @brianjlai 's suggestion about keeping a "pet implementation" of a file-based connector going just as a demo alongside this project. It shouldn't be its own module, it can just be a package inside the CDK that you remove once this project is done. But I think it helps a lot if as you are designing these abstractions, you're seeing exactly what the user experience is like. I tried doing something like this in #27297 with the stripe connector (I just added a `demo` package in the CDK to test the UX as I go along)

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/abstract_file_based_stream.py
```diff
@@ -0,0 +1,113 @@
...
+        self,
+        raw_config: Dict[str, Any],
+        stream_reader: AbstractFileBasedStreamReader,
+        availability_strategy: AbstractFileBasedAvailabilityStrategy,
+        discovery_policy: AbstractDiscoveryPolicy,
+        parsers: Dict[FileType, FileTypeParser],
+    ):
+        super().__init__()
+        try:
+            self.config = FileBasedStreamConfig(**raw_config)
```
sherifnada, line 46: why not ask the consumer to pass in a FileBasedStreamConfig?

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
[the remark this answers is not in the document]
sherifnada: agree w alex, stream slice should contain all the info needed to read the records in a file

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_spec.py
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+#
+
+from pydantic import BaseModel
+
+
+class AbstractFileBasedSpec(BaseModel):
```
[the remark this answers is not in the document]
clnoll: I'm not entirely sure whether we'll want it. I'll remove it for now and when we get around to implementing `spec` we can see if we want it back.

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/abstract_file_based_stream.py
clnoll: Sure, done.

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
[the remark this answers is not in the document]
sherifnada: I think the reasoning is: slices are a function of state, so we should be able to encode anything we need from the state into the slice

For context I think we're trying to push in the direction of having slices completely describe requests, more [here](https://www.loom.com/share/100bd529cc8f49c3877cfaa235f91ae6)

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
sherifnada: oh derp good point on inmemory -- i missed that

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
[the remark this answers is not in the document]
girarda: ^+1
Would be interesting to get some numbers to confirm, but this is also how most low-code connectors work. We usually only need to reference the stream_slice and not the stream_state in the outgoing request

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -0,0 +1,90 @@
...
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+#
+
+import logging
+import traceback
+from abc import ABC
+from typing import Any, Dict, List, Mapping, Optional, Tuple
+
+from airbyte_cdk.sources import AbstractSource
+from airbyte_cdk.sources.file_based.default_file_based_availability_strategy import DefaultFileBasedAvailabilityStrategy
```
girarda, line 11: I don't see this file anymore. Was it deleted?

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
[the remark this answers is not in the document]
clnoll: 👍 got it, makes sense to me. I've updated the signature here and in the downstream functions.

`stream_state` is required by the abstractmethod `read_records` from the `Stream` interface, which we're inheriting from so I've left that in.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
clnoll: Strange - no, it's still there. Where were you looking?

girarda: not sure what happened.  🤷

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -0,0 +1,130 @@
...
+    ) -> Iterable[Mapping[str, Any]]:
+        """
+        Yield all records from all remote files in `list_files_for_this_sync`.
+        """
+        schema = self._catalog_schema
+        if schema is None:
+            # On read requests we should always have the catalog available
+            raise MissingSchemaError("Expected `json_schema` in the configured catalog but it is missing.")
+        parser = self.get_parser(self.config.file_type)
+        for file in self.list_files_for_this_sync(stream_slice):
```
girarda, line 45: nit: I think we should compute the files to read from the `stream_slices` method and store them in the slice. I can update the `list_files_for_this_sync` to reflect this as I implement incremental syncs

airbyte-cdk/python/unit_tests/sources/file_based/in_memory_files_source.py
```diff
@@ -0,0 +1,68 @@
...
+    files: Dict[str, dict]
+    file_type: str
+
+    def list_matching_files(
+        self,
+        globs: List[str],
+        from_date: Optional[datetime] = None,
+    ) -> List[RemoteFile]:
+        return [
+            RemoteFile(f, data["last_modified"], self.file_type)
```
girarda, line 51: nit: [data["last_modified"] is a string ](https://github.com/airbytehq/airbyte/pull/27122/files#diff-261d057402b704966dbd561ac054ecb582646d2e5cdef5ee57ef81d6bfc32e62R32) so we should parse it (datetime.strptime(`data["last_modified"], "%Y-%m-%dT%H:%M:%S.%fZ"`)

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -0,0 +1,225 @@
...
+                    "json_schema": {"col1": "string", "col2": "string"},
+                    "name": "stream1",
+                    "supported_sync_modes": ["full_refresh"],
+                }
+            ]
+        }
+    )
+    .set_expected_records(
+        [
+            {"col1": "val11", "col2": "val12"},
```
girarda, line 50: nit: I think we should add the filename and its timestamp to the records [like we do for the S3 source](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/stream.py#L178-L195). 

Non-blocking because I can add the filename and the cursor field as I implement incrementals

girarda (commented):
left a few non-blocking comments based on some things I noticed while looking into how to support incrementals. 

lgtm 🚀

girarda (approved):

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -0,0 +1,225 @@
...
+            },
+        }
+    )
+    .set_file_type("csv")
+    .set_expected_catalog(
+        {
+            "streams": [
+                {
+                    "json_schema": {
+                        "col1": "string",
```
girarda, line 98: @clnoll 
shouldn't this be a json_schema instead of a col -> type mapping?

sherifnada (approved):

Issue #26904, opened 2023-06-01
Document builder incremental sync date format capabilities


flash1293:
### What area the feature impact?

None

### Revelant Information

Currently it's not clear how the date format needs to be specified in the connector builder and there is no authoritative documentation explaining all the details (except for a link within the text on the incremental sync documentation page).

This should be improved in the following way:
* In the manifest schema documentation of the `datetime_format` field, list out all available placeholders (copy of the relevant table from https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes plus `%s` - omit the system-dependent variables as we can't guarantee they will be available in every case) using a markdown table and include a link to the python documentation
* On the incremental sync concept page, link to the relevant section in the yaml reference
* Interpret markdown for the description in the tooltip generated from manifest documentation

flash1293:
Grooming notes:
* Do not use a table but a list, easier to render in a tooltip
* Give the tooltip a max height and start scrolling to prevent a super large tooltip

PR #27149, opened 2023-06-08
Low code CDK: Datetime format documentation

flash1293:
## What

Related to https://github.com/airbytehq/airbyte/issues/26904

Documents all supported placeholders in format strings.

As Python binds to the C standard lib provided by the OS, the exact supported options can differ. I chose to document the ones that are guaranteed to work so we don't over-promise.

They also depend on the current locale - as in most cases (using our base image or publishing from the connector builder) the locale will be set to `en_US` I stuck to those as examples and made a note in the bottom that it's possible to use it with other locales as well.


girarda (approved):
>  I chose to document the ones that are guaranteed to work so we don't over-promise
+1 

:shipit:

flash1293:
This can be merged once https://github.com/airbytehq/airbyte-platform-internal/pull/7115 is merged

PR #27158, opened 2023-06-08
[ISSUE #26909] adding message repository

maxi297:
## What
Addresses https://github.com/airbytehq/airbyte/issues/26909 as the in-memory solution does not work

Note that this has only been implemented for the case of passing control message to the Connector Builder but there are some currently implemented cases that would benefit from this like `AirbyteTracedException.emit_message`

## How
Adding a repository where messages can get published to be consumed upstream whenever the consumer feels like it. This solution is built on https://github.com/airbytehq/airbyte/tree/maxi297/allow-for-control-message-anywhere-in-the-code

## 🚨 User Impact 🚨
This change is meant to be backward compatible


maxi297:
greenhouse and airtable are broken on master so those test failures are not representative of the code change success

airbyte-cdk/python/airbyte_cdk/entrypoint.py
```diff
@@ -82,7 +82,8 @@ def run(self, parsed_args: argparse.Namespace) -> Iterable[str]:
         with tempfile.TemporaryDirectory() as temp_dir:
             if cmd == "spec":
                 message = AirbyteMessage(type=Type.SPEC, spec=source_spec)
-                yield message.json(exclude_unset=True)
+                yield from [self.airbyte_message_to_string(queued_message) for queued_message in self._emit_queued_messages(self.source)]
+                yield self.airbyte_message_to_string(message)
             else:
                 raw_config = self.source.read_config(parsed_args.config)
                 config = self.source.configure(raw_config, temp_dir)
```
maxi297: If we really want to preserve order, we could call `_emit_queued_messages`. For example, do we assume that `source.read_config` and `source.configure` can emit messages?

airbyte-cdk/python/airbyte_cdk/sources/message/repository.py
```diff
@@ -0,0 +1,36 @@
...
+    @abstractmethod
+    def emit_message(self, message: AirbyteMessage) -> None:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def consume_queue(self) -> Iterable[AirbyteMessage]:
+        raise NotImplementedError()
+
+
+class InMemoryMessageRepository(MessageRepository):
```
maxi297, line 21: Random thought: this should probably be a singleton; it would be pretty cool to not have to instantiate and pass this through the whole CDK. However, there are threading implications to do this and therefore I think we should delay this

airbyte-cdk/python/airbyte_cdk/entrypoint.py
maxi297: Just to clarify, this seems only an issue if other part of the code emits message as print

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -357,3 +366,7 @@ def _get_message(self, record_data_or_message: Union[StreamData, AirbyteMessage]
             return record_data_or_message
         else:
             return stream_data_to_airbyte_message(stream.name, record_data_or_message, stream.transformer, stream.get_json_schema())
+
+    @property
+    def message_repository(self) -> Union[None, MessageRepository]:
```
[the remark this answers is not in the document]
girarda: the only class extending `Source` is `AbstractSource` so I think it would make sense to either make `Source` an interface or combine them. Not something that has to be done as part of this PR though.

airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
```diff
@@ -211,7 +214,11 @@ def get_access_token(self) -> str:
             self.access_token = new_access_token
             self.set_refresh_token(new_refresh_token)
             self.set_token_expiry_date(new_token_expiry_date)
-            emit_configuration_as_airbyte_control_message(self._connector_config)
+            if self._message_repository:
+                self._message_repository.emit_message(create_connector_config_control_message(self._connector_config))
+            else:
+                # FIXME emit_configuration_as_airbyte_control_message as been deprecated in favor of package airbyte_cdk.sources.message
```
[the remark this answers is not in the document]
girarda: I think that's fine. We can figure out how to always pass the repository and delete the old behavior when we work on exposing the raw requests and responses from the authenticator.

girarda (approved):
:shipit:

flash1293 (approved):
Approach makes sense to me. Tested with the connector builder and it works as expected end-to-end, LGTM.

Question for a follow-up task - should we add an e2e test for this interaction? It already proved to be brittle when implementing it initially, I worry it will silently start failing at some point because there are so many parts that need to stay lined up.

PR #27162, opened 2023-06-08
🐛Source Klaviyo: anonymize connection error messages

lazebnyi:
## What

This PR anonymizes connection error messages to protect sensitive information.

## How

- Implemented functionality to anonymize connection error messages.
- Added regular expressions to replace URLs and API keys with generic placeholders.
- Tested the script with various error messages to ensure proper anonymization.


airbyte-integrations/connectors/source-klaviyo/source_klaviyo/source.py
```diff
@@ -21,7 +22,15 @@ def check_connection(self, logger, config: Mapping[str, Any]) -> Tuple[bool, Any
             # we use metrics endpoint because it never returns an error
             _ = list(Metrics(api_key=config["api_key"]).read_records(sync_mode=SyncMode.full_refresh))
         except Exception as e:
-            return False, repr(e)
+            original_error_message = repr(e)
+
+            # Regular expression pattern to match the API key
+            pattern = r'api_key=\b\w+\b'
```
josephkmh, line 28: I assume the API key can only be alphanumeric, because it's embedded in the URL. So this regex should be OK, although it doesn't allow for special characters.

joeykmh (approved):
Thanks for the quick turnaround, code LGTM!

PR #27176, opened 2023-06-08
File CDK: CSV parser

clnoll:
## What
Adds error handling and tests for the CSV parser.

Note: I did not find a strong reason to use pyarrow for the CSV parser. It reads entire files into memory at once and stores data in a columnar format, but since we're not doing any computation-heavy operations this is not necessary.

This PR also includes some formatting changes, in a second commit.

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/file_based/exceptions.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/file_based/exceptions.py
@@ -11,6 +11,10 @@ class MissingSchemaError(Exception):
     pass
 
 
+class RecordParseError(Exception):
+    pass
+
+
 class SchemaInferenceError(Exception):
     pass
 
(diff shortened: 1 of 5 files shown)


airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream.py
```diff
@@ -194,7 +202,7 @@ async def _infer_file_schema(self, file: RemoteFile) -> Mapping[str, Any]:
             )
         except Exception as exc:
             raise SchemaInferenceError(
-                f"Error inferring schema for file: {file.uri}"
+                f"Error inferring schema for file: {file.uri}. Is the file valid {self.config.file_type}?"
```
maxi297: Can schema inference exception be caused by the file type? In which case? If it's not always the case, should we refine this error message?

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream.py
```diff
@@ -194,7 +202,7 @@ async def _infer_file_schema(self, file: RemoteFile) -> Mapping[str, Any]:
             )
```
maxi297: For my curiosity (and therefore outside of the PR scope): why is infer_schema async?

maxi297 (approved):
Approved: All comments are unrelated to the scope of the PR

PR #27178, opened 2023-06-08
Metadata: Add Breaking Changes entry

### LINKED ISSUE #27064: Add breakingChanges and migrationGuideURl to metadata
Update [ConnectorMetadataDefinitionV0.yaml](https://github.com/airbytehq/airbyte/blob/master/airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml) to include
* breakingChanges (optional) - the type should be added as a separate json schema BreakingChanges.yaml with
    * version (string, pattern: semver, required)
    * upgradeDeadline (string, format: date-time, required)
    * message (string, required

It is fine if people start adding these entries or don't before the validation is in place, since we won't be consuming the info (and even if we did consume it, we more importantly aren't using it yet) anyway. 

Add logic to geneate `migrationDocumentationUrl` on build - A projection of documentation url + expected migration_guide.md path `f"{documentationUrl}/migration_guide"`

bnchrch:
## What
1. Adds an optional release field to the metadata.yaml files
2. Adds an optional release field to the registry
3. Adds a optional migrationDocumentationUrl field that will be set to documentationUrl/migration_guide
4. Increases test coverage

```yaml
metadataSpecVersion: 1.0
data:
  name: AlloyDB for PostgreSQL
  definitionId: 1fa90628-2b9e-11ed-a261-0242ac120002
  connectorType: source
  dockerRepository: airbyte/image-exists-1
  githubIssueLabel: source-alloydb-strict-encrypt
  dockerImageTag: 0.0.1-exists
  documentationUrl: https://docs.airbyte.com/integrations/sources/alloydb
  connectorSubtype: database
  license: MIT
  releases:
    breakingChanges:
      3.0.0:
        upgradeDeadline: 2023-08-22
        message: "This version changes the connector’s authentication method from `ApiKey` to `oAuth`, per the [API guide](https://amazon-sqs.com/api/someguide)."
  tags:
    - language:java
 ```
 
 ## Review order

 1. [...etadata_service/lib/tests/fixtures/metadata_validate/valid/metadata_breaking_changes.yaml](https://github.com/airbytehq/airbyte/pull/27178/files#diff-3300a498df0676681f551f77fb8e825330ea83324515d9d516511578246a34c1)
 2. [...yte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorReleases.yaml](https://github.com/airbytehq/airbyte/pull/27178/files#diff-9fb9be5f14e2a7479a66b00d841139fee1fb99a1ebb069b02d2ef12c32543f26)
 3. [airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry.py](https://github.com/airbytehq/airbyte/pull/27178/files#diff-30a0695c81393223cb4332ae1c164220f67eee8a8a5900def040e8a78443c4b1)
 
closes #27064

### CODE REVIEW

File: airbyte-ci/connectors/metadata_service/lib/metadata_service/models/generated/ConnectorMetadataDefinitionV0.py [hunk 1]
```diff
@@ -83,6 +100,16 @@ class Config:
...
 
 
+class ConnectorBreakingChanges(BaseModel):
+    class Config:
+        extra = Extra.allow
+
+    __root__: Dict[constr(regex=r"^\d+\.\d+\.\d+$"), VersionBreakingChange] = Field(
+        ...,
+        description="Each entry denotes a breaking change in a specific version of a connector that requires user action to upgrade.",
+    )
```
erohmensing, line 110: I am not sure it matters, but I know that this was originally designed as a list instead of a dict. I think a dict makes sense as we shouldn't expect multiple entries for one key, and the ordering is going to be done based on semver and not necessarily the order that the entries came in - but wanted to call it out just in case

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorRegistrySourceDefinition.yaml
```diff
@@ -68,3 +68,5 @@ properties:
   maxSecondsBetweenMessages:
     description: Number of seconds allowed between 2 airbyte protocol messages. The source will timeout if this delay is reach
     type: integer
+  releases:
+      "$ref": ConnectorReleases.yaml
```
erohmensing, line 72:
```suggestion
  releases:
      "$ref": ConnectorReleases.yaml

```

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorReleases.yaml
```diff
@@ -0,0 +1,44 @@
...
+    type: object
+    additionalProperties: false
+    required:
+      - upgradeDeadline
+      - message
+    properties:
+      upgradeDeadline:
+        description: The deadline by which to upgrade before the breaking change takes effect.
+        type: string
+        format: date
```
erohmensing, line 37: Do we need to do any validation on date format? Or does `date` specifically refer to `YYYY-MM-DD` format?

airbyte-ci/connectors/metadata_service/lib/tests/fixtures/metadata_validate/invalid/metadata_invalid_breaking_change_version.yaml
```diff
@@ -0,0 +1,19 @@
...
+  dockerRepository: airbyte/image-exists-1
+  githubIssueLabel: source-alloydb-strict-encrypt
+  dockerImageTag: 0.0.1-exists
+  documentationUrl: https://docs.airbyte.com/integrations/sources/alloydb
+  connectorSubtype: database
+  releaseStage: generally_available
+  license: MIT
+  releases:
+    breakingChanges:
+      notsemver:
```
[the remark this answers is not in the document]
erohmensing: Especially given this note on `patternProperties`: 

image

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry.py
```diff
@@ -92,12 +132,12 @@ def metadata_to_registry_entry(metadata_entry: LatestMetadataEntry, connector_ty
...
 
     # if there is no releaseStage, set it to "alpha"
-    # Note: this is something our current cloud registry generator does
-    # Note: We will not once this is live
     if not overrode_metadata_data.get("releaseStage"):
         overrode_metadata_data["releaseStage"] = "alpha"
 
+    # apply generated fields
     overrode_metadata_data["iconUrl"] = metadata_entry.icon_url
+    overrode_metadata_data["releases"] = apply_connector_release_defaults(overrode_metadata_data)
```
erohmensing, line 140: In the case of `if metadata_releases is None: return None`, is this setting to `releases: null` instead of leaving the key out entirely? If so is that the behavior we want? (I don't think it matters on the consumption side, but might look a little confusing)

airbyte-ci/connectors/metadata_service/orchestrator/tests/test_registry.py
erohmensing, line 1: Nice coverage improvements!

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/generated/ConnectorMetadataDefinitionV0.py
erohmensing:
```
   additionalProperties: true
    minProperties: 1
    patternProperties:
      "^\\d+\\.\\d+\\.\\d+$":
        $ref: "#/definitions/VersionBreakingChange"
```
the `minProperties: 1` and `additionalProperties: true` sort of indicates to me that a list might be better here? idk

erohmensing (approved):
LGTM! Main concern is just making sure we are validating against semver correctly, otherwise :shipit:

airbyte-ci/connectors/metadata_service/lib/tests/fixtures/metadata_validate/valid/metadata_breaking_changes_with_migration_doc_url.yaml
```diff
@@ -0,0 +1,21 @@
...
+  connectorSubtype: database
+  releaseStage: generally_available
+  license: MIT
+  releases:
+    migrationDocumentationUrl: https://docs.airbyte.com/integrations/sources/alloydb/migration_guide
+    breakingChanges:
+      2.0.0:
+        migrationDocumentationUrl: https://docs.airbyte.com/integrations/sources/alloydb/migration_guide#2.0.0
+        upgradeDeadline: 2023-08-22
+        message: "This version changes the connector’s authentication method from `ApiKey` to `oAuth`, per the [API guide](https://amazon-sqs.com/api/someguide)."
```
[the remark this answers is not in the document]
erohmensing: The unit test clarified this for me - yep!

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/generated/ConnectorMetadataDefinitionV0.py
bnchrch: ahh sorry, this was an atrifact from originally going the list route. Ill set this to false and add a test case to prove no additional props are allowed

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorReleases.yaml
bnchrch: Jsonschema takes care of this one for us
https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry.py
bnchrch: Good catch. It will set it to null. Which I think is a fine trade off to being able to keep the "if" statement isolated to `apply_connector_release_defaults`

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorReleases.yaml
erohmensing: dope!

PR #27190, opened 2023-06-09
source-file: bump version in metadata.yaml

alafanechere:
## What
The `dockerImageTag` of `source-file` in `metadata.yaml` does not match with the dockerfile.
`source-file-secure` was correctly published and bumped to 0.3.11 in previous PR but source-file was not bumped in metadata, so `source-file` 0.3.11 was not published.


octavia-approvington:

octavia-approvington (approved):

PR #27204, opened 2023-06-09
connector_ci: more granular mgmt of concurrency

alafanechere:
## What
With current implementation `run_connectors_pipelines` is called twice: once for python connector, another time for java connectors with forced concurrency.
It's problematic because the `run_connectors_pipelines` uploads the `complete.json` file which should be a singleton aggregating all the connector test reports.
This PR slightly changes the logic. 
We call `run_connectors_pipelines` but use different semaphore object according to connector language. Concurrency is forced to 1 for Java connectors.

It has the following impact:
Only a single java connectors is tested at a time, but multiple python connectors can be tested in parallel.
Python and Java connectors are now tested concurrently, not Python connectors first, then Java connectors.

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
@@ -223,30 +223,13 @@ def test(
     try:
         anyio.run(
             run_connectors_pipelines,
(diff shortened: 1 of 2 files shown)


octavia-approvington:
Lets do it! <br/> giddddy-up

octavia-approvington (approved):

PR #27229, opened 2023-06-10
initial ctid for testing

rodireich:


airbyte-db/db-lib/src/main/java/io/airbyte/db/jdbc/StreamingJdbcDatabase.java
```diff
@@ -95,20 +95,33 @@ protected <T> Stream<T> toUnsafeStream(final ResultSet resultSet,
       @Override
       public boolean tryAdvance(final Consumer<? super T> action) {
```
akashkulk, line 96: Curious, were you still facing the same issue where this is failing silently? 

I assume here the issue becomes that if there is an error - the ctid sync _appears_ to be successful and could move on to the CDC/xmin/cursor based sync. As a result, it can lead to a situation where there is data loss due to missing records from the original sync?

subodh1810:
Closing in favour of https://github.com/airbytehq/airbyte/pull/27302

PR #27248, opened 2023-06-12
Remove old growth related labeling code

timroes:
## What

Remove some legacy growth labeler, that isn't in use anymore.


evantahler (approved):

timroes:
Merging despite CI failure, since it should be unrelated and this PR doesn't touch any real code.

PR #27258, opened 2023-06-12
Registry: Disable all hidden connectors in cloud

### LINKED ISSUE #27254: disable destination-amazon-sqs in cloud

bnchrch:
## What
There are a number of connectors that are in the cloud connector registry that have been hidden from the users
https://github.com/airbytehq/airbyte-platform-internal/blob/master/oss/airbyte-webapp/src/core/domain/connector/constants.ts#L2

This PR disables them in the cloud registry so that we dont tell prospective customers they are available when they are not

closes #27254

### DIFF
--- a/airbyte-integrations/connectors/destination-amazon-sqs/metadata.yaml
+++ b/airbyte-integrations/connectors/destination-amazon-sqs/metadata.yaml
@@ -10,7 +10,7 @@ data:
   name: Amazon SQS
   registries:
     cloud:
-      enabled: true
+      enabled: false
     oss:
       enabled: true
   releaseStage: alpha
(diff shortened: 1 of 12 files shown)


evantahler (approved):

alafanechere (approved):
Now that we discarded the definitions mask on cloud did these connector went live on Cloud or not?

pedroslopez (approved):
Thanks for including docs links where relevant! Didn't go through each one individually but will trust you went through and validated each one should actually be hidden.

erohmensing (approved):
Thanks! I wonder if we should set these to private in the prod DB so that we can remove the frontend logic around them, as a followup 🤔

bnchrch:
Confirmed these should all still be disbaled

Also Moved a related issue to this to grooming for tomorrow :)

PR #27262, opened 2023-06-12
Experiments to fix CI

evantahler:
These changes make CI green again.
I don't really know why Github's build Artifacts breaks things, but here we are.  

Compare the results of this PR ([green](https://github.com/airbytehq/airbyte/actions/runs/5247676966/jobs/9478178445?pr=27262)) with master ([red](https://github.com/airbytehq/airbyte/actions/runs/5247111805/jobs/9476890267
))

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -251,10 +251,10 @@ jobs:
         with:
           ref: ${{ github.head_ref }}
 
-      - name: Cache Build Artifacts
-        uses: ./.github/actions/cache-build-artifacts
-        with:
-          cache-key: ${{ secrets.CACHE_VERSION }}-connectors
+      # - name: Cache Build Artifacts
+      #   uses: ./.github/actions/cache-build-artifacts
+      #   with:
+      #     cache-key: ${{ secrets.CACHE_VERSION }}-connectors
 
       - uses: actions/setup-java@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle.yml
```diff
@@ -251,10 +251,11 @@ jobs:
...
 
-      - name: Cache Build Artifacts
-        uses: ./.github/actions/cache-build-artifacts
-        with:
-          cache-key: ${{ secrets.CACHE_VERSION }}-connectors
+      # Commenting this out fixes the build, and I don't know wny - https://github.com/airbytehq/airbyte/pull/27262
+      # - name: Cache Build Artifacts
+      #   uses: ./.github/actions/cache-build-artifacts
+      #   with:
+      #     cache-key: ${{ secrets.CACHE_VERSION }}-connectors
```
bnchrch, line 258: So the question is
1. Is caching the problem?
2. Or is the latest cache the problem?

I imagine its the later.

wdyt about busting the cache?

```suggestion
       - name: Cache Build Artifacts
          uses: ./.github/actions/cache-build-artifacts
          with:
            cache-key: ${{ secrets.CACHE_VERSION }}-connectors-gradle-build
```

bnchrch (approved):
LGTM. Figured it might be worthwhile to bust this up to see if we can keep the caching alive

.github/workflows/gradle.yml
evantahler: nice... let's try it

Issue #27233, opened 2023-06-12
dagger: dind "failed to get digest" errors


alafanechere:
On nightly build on which we disabled conccurency for java connectors (see https://github.com/airbytehq/airbyte/issues/27168#issuecomment-1584608324) 
we face errors from the global `dockerd` service:
```
2023-06-09T13:30:59.5034536Z #16 2312.7 time="2023-06-09T13:29:15.008451175Z" level=error msg="invalid image access: \"sha256:b5fadaa0074623dc614a16fb29df59277f57e96b0bdc5a23d6d5058f14e96172\", error: \"failed to get digest sha256:b5fadaa0074623dc614a16fb29df59277f57e96b0bdc5a23d6d5058f14e96172: open /var/lib/docker/image/overlay2/imagedb/content/sha256/b5fadaa0074623dc614a16fb29df59277f57e96b0bdc5a23d6d5058f14e96172: no such file or directory\""
```

Context:
* We initialize a [global `dockerd` service ](https://github.com/airbytehq/airbyte/blob/f8a9c3ed2b63acf2f5c663f08cfacc82c7202b07/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py#L284) that is bound to containers that require docker interaction
* We mount a cache volume to `/var/lib/docker` to this dockerd service to get a persistent volume in case the dagger kills the service after the 10 seconds inactivity grace period.

alafanechere:
Hypothesis:
**A. The cache volume has grown too big and some of its data was pruned.**
I don't think this hypothesis is valid because I tried mapping new volumes to `/var/lib/docker` with the same error occuring.

**B. We're hitting a docker bug that has been solved in more recent versions.**
I'm[ trying out docker 24 images](https://github.com/airbytehq/airbyte/actions/runs/5242639625/jobs/9466292855) for dind and cli and check if it has an effect.

**C. The dockerd service is not gracefully killed by Dagger and leads to corrupted volumes** 
I [enabled debug log level ](https://github.com/airbytehq/airbyte/actions/runs/5242639625/jobs/9466292855)for dockerd to get more insights on the dockerd service lifecyle.

vito:
[...]

PR #27267, opened 2023-06-12
connectors-ci: run `dockerd` in a background task

alafanechere:
## What
This is an hacky attempt at solving #27233 

We believe the #27233 errors are due to intermittent restarts of the dockerd service container by Dagger. 
Dagger service containers *are started just-in-time, de-duplicated, and stopped when no longer needed* (cf [doc](https://docs.dagger.io/757394/use-service-containers/)). This explains the intermittent stop and restart but it looks like it leads to a dockerd corrupted state, probably because dockerd is not gracefully stopped.

## How
Start `dockerd` in a background task before connector pipelines.
Dagger will not manage the lifecycle of the dockerd containers because `dockerd.exit_code()` forces long running execution.
We kill the dockerd container once all the connector pipelines are done.
We don't care that a graceful shutdown happens because no docker data needs to be kept throughout pipelines executions (hence the removal of the /var/lib/docker volume).

## 🚨 User Impact 🚨
* No impact on publish: [publishing a java and a python connectors worked correctly.](https://github.com/airbytehq/airbyte/actions/runs/5252717687/jobs/9489117953)
* A [nightly build](https://github.com/airbytehq/airbyte/actions/runs/5247814879) ran on Java connectors worked correctly 
* Currently running a [nightly build](https://github.com/airbytehq/airbyte/actions/runs/5252723316) on all GA connectors to confirm #27233 is solved. -> great results!


octavia-approvington:
It's time <br/> fine lets go

octavia-approvington (approved):

PR #27291, opened 2023-06-13
add link to record selector docs

lmossman:
## What
In a build-a-connector session, we realized there isn't a way for users to quickly get to the docs on Record Selector today.

This PR solves the problem by adding a link in the description

## How
Add a link to the Record Selector docs in the description of the field in the declarative schema.

Since the tooltip in the Builder UI is sourced from this description, this will add the link to the tooltip as well:
Screen Shot 2023-06-12 at 5 45 24 PM


brianjlai (approved):
🪶

PR #27297, opened 2023-06-13
PoC concurrent CDK

sherifnada:


airbyte-cdk/python/airbyte_cdk/v2/abstract_source_v2.py
```diff
@@ -0,0 +1,65 @@
+from abc import ABC, abstractmethod
```
sherifnada, line 1: ignore this file, not relevant for the demo I showed

airbyte-cdk/python/airbyte_cdk/v2/concurrency/async_source.py
```diff
@@ -0,0 +1,70 @@
+import json
```
sherifnada, line 1: ignore this file, not used / relevant for the demo

airbyte-cdk/python/airbyte_cdk/v2/abstract_source_v2.py
```diff
@@ -0,0 +1,65 @@
...
+
+from airbyte_cdk.sources import Source
+
+from airbyte_cdk.sources.streams.core import StreamData
+from airbyte_protocol.models import ConfiguredAirbyteCatalog, AirbyteCatalog, AirbyteConnectionStatus, Status
+
+from airbyte_cdk.v2 import Stream
+
+
+class AbstractSource(Source, ABC):
```
girarda, line 13: I don't see where this class is used. Is it cruft / replaced by AsyncSource?

girarda: nvm. I see that the demo does not run a full source. The main follow up item I would be interested in is whether there's plausible adapter that would allow us to upgrade AbstractSource and HttpStream to use this v2 without changing their interface

airbyte-cdk/python/airbyte_cdk/v2/concurrency/stream_group.py
```diff
@@ -0,0 +1,122 @@
...
+
+@dataclass
+class StreamAndPartition(Generic[PartitionType]):
+    stream: PartitionedStream
+    partition_descriptor: PartitionType
+
+
+class ConcurrentStreamGroup(ABC, Generic[PartitionType]):
+    requester: AsyncRequester
+    concurrency_policy: ConcurrencyPolicy
```
sherifnada, line 29: in hindsight I think the requester should probably hold the concurrency policy because it should be the one coordinating retrieval of the partitions

PR #27301, opened 2023-06-13
connectors-ci: disable format check on python connectors

alafanechere:
## What
We don't want to run Python code format until we have an auto-formatting pipeline to run before tests.


octavia-approvington:
Send it! <br/> rocket socks

octavia-approvington (approved):

PR #27223, opened 2023-06-09
Issue 26581/per partition cursor

maxi297:
## What
Addresses https://github.com/airbytehq/airbyte/issues/26581. Note that this is in draft since we need to complete https://github.com/airbytehq/airbyte/issues/26607 before merging

## How
Adding a new cursor class to support per partition states.

TODO:
* Validate if requests performed are identical from master for zenloop (records and slice logs have been validated). If possible/worth, move this to an integration test
* Validate how this interacts with custom components: they have their own state and we should probably instantiate PerPartitionCursor only if the incremental is not a custom component to preserve backward compatibility. This would be only for backward compatibility reasons though and therefore, maybe we need to rely on the format of the stream state to validate this...
* [tentative] Clean up cursor/stream slicer interface if possible: it might be tricky since `SimpleRetriever` thinks stream slicers have `update_cursor` has a method but it does not make sense for ListPartitionRouter and SubstreamPartitionRouter

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/per_partition_cursor.py`
2. `airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py`

## 🚨 User Impact 🚨
Until https://github.com/airbytehq/airbyte/issues/26607 is completed, this change is not backward compatible and

Issue #26607, opened 2023-05-25
Unblock source-zenloop airbyte-cdk version following state per partition


maxi297:
### What area the feature impact?

Connectors

### Revelant Information

Follow-up to https://github.com/airbytehq/airbyte/issues/26581

## Tell us about the problem you're trying to solve
As of 2023-05-25, we have 2 OSS workspaces that uses the zenloop source. We had to fix the airbyte-cdk version to allow the release of the state per partition feature. We can also have connectors created with the connector builder using an incremental sync and partition routers. 

*source-zenloop*
```
{
    "inserted_at": "2022-09-19T08:46:22.373163Z",
    "id": "WlRBek9ESTFNREl0TmpJMk9DMDBOR0V4TFRoaE16UXRZV1UyWW1SbU56WTNPVGRs"
}
```

## Describe the solution you’d like
The pattern we need to migrate is:

### source-zenloop
```
{ 
  <cursor_field>: <cursor_value>,
  <partition_field>: <partition_value>
}
```

It should map to this format before being passed to the StatePerPartition component:
```
{
  "states": [
    {
      <partition_field>: <partition_value>,
      <cursor_field>: <cursor_value>
    }
...
]}
```

### How to identify that the migration should occur
If the stream is incremental, has a partition router and state does not have the root field `states`, then a migration is needed

### Acceptance Criteria
- [ ] ~source-zenloop can be executed with it's previous state without error~ given the low number of user, we will ask them to do a reset
- [x] Update CATs
- [x] Document breaking change

PR #27306, opened 2023-06-13
[ISSUE #26607] CATs: support list in state

maxi297:
## What
Following https://github.com/airbytehq/airbyte/pull/27223, we will have states with array in them. We would like to be able to reference them with indices like:

```
  incremental:
    - <...>
      cursor_paths:
        "answers": [ "states", 0, "cursor", "inserted_at" ]
        "answers_survey_group": [ "states", 0, "cursor", "inserted_at" ]
```

## How
Allowing to parse int and str for `cursor_paths` field allow to fetch in a dict or in a list.

## 🚨 User Impact 🚨
The change is not fully backward compatible. If a user would have defined a cursor_path as `[ "cursor", 0 ]`, today it would cast the int as a string and would generate the cursor_path [ "cursor", "0" ]. However, I have not seen such case like this in the repo as all values are strings.

 It will also unblock https://github.com/airbytehq/airbyte/issues/26607.

### DIFF
--- a/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/config.py
+++ b/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/config.py
@@ -7,7 +7,7 @@ import logging
 from copy import deepcopy
 from enum import Enum
 from pathlib import Path
-from typing import Generic, List, Mapping, Optional, Set, TypeVar
+from typing import Generic, List, Mapping, Optional, Set, TypeVar, Union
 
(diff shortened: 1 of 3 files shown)


erohmensing (approved):
This seems fine to me. Re: backwards compatibility, I agree that while some of the cursor_paths use numbers, they seem to be stringified already. If they didn't and we ran into the issue you mentioned, it would just be the acceptance tests that started failing - there wouldn't be end user impact

I'm intrigued by the introduction of the explicit use of `state` and `cursor` in the state message... but that's not what I'm reviewing here :D

PR #27325, opened 2023-06-13
Update postgres.md

### LINKED ISSUE #19082: Source Postgres: Clarify the behaviour of CDC replication with two or more destination
### Tell us about the documentation you'd like us to add or update.

Currently, it’s possible to connect more than one destination to a single CDC source (PostgreSQL), which might lead to an inconsistent replication behavior. There's no explicit documentation of what would happen in case two or more connections would run a replication from the same PostgreSQL replication slot, and the UI encourages reusing the same source multiple times. 

My assumption is that both destinations would end up with an incomplete set of changes, and perhaps the replication process itself will fail. From [PostgreSQL docs](https://www.postgresql.org/docs/current/logicaldecoding-explanation.html#LOGICALDECODING-REPLICATION-SLOTS):
> A logical replication slot knows nothing about the state of the receiver(s). It's even possible to have multiple different receivers using the same slot at different times; they'll just get the changes following on from when the last receiver stopped consuming them. Only one receiver may consume changes from a slot at any given time.

I’d say the desired behaviour would be for Airbyte to prevent setting up a second connection on a CDC source without removing the first one, but [as discussed on Slack](https://airbytehq.slack.com/archives/C021JANJ6TY/p1666815866314179), I'd start with making the documentation clearer.

### If applicable, add links to the relevant docs that should be updated
- [Postgres connector docs](https://airbytehq.github.io/integrations/sources/postgres/)

evantahler:
Closes https://github.com/airbytehq/airbyte/issues/19082

Add a note that says CDC can only work with one destination


docs/integrations/sources/postgres.md
```diff
@@ -168,6 +168,7 @@ Airbyte uses [logical replication](https://www.postgresql.org/docs/10/logical-re
 - Schema changes are not supported automatically for CDC sources. Reset and resync data if you make a schema change.
 - The records produced by `DELETE` statements only contain primary keys. All other data fields are unset.
 - Log-based replication only works for master instances of Postgres.  CDC cannot be run from a read-replica of your primary database.
+- Am Airbyte database source using CDC replication can only be used with a single Airbyte destination.
```
subodh1810, line 171: Could you add more context on why this is the case?

evantahler: 👍 a little more info added.

docs/integrations/sources/postgres.md
```diff
@@ -168,6 +168,7 @@ Airbyte uses [logical replication](https://www.postgresql.org/docs/10/logical-re
 - Schema changes are not supported automatically for CDC sources. Reset and resync data if you make a schema change.
 - The records produced by `DELETE` statements only contain primary keys. All other data fields are unset.
 - Log-based replication only works for master instances of Postgres.  CDC cannot be run from a read-replica of your primary database.
+- Am Airbyte database source using CDC replication can only be used with a single Airbyte destination.  This is due to how Postgres CDC is implemented - each destination would recieve only part of the data available in the replication slot.
```
subodh1810, line 171: *An

subodh1810 (approved):

PR #27331, opened 2023-06-13
Ignore -migrations.md files in sidebar

### LINKED ISSUE #27086: Update docs to enable unlisted (sidebar) migration guide docs

erohmensing:
## What
* Enable hosting migration guides which aren't listed in the sidebar
* Close https://github.com/airbytehq/airbyte/issues/27086

Consider a migrations file added like this: 

image

It will be hosted at `https://docs.airbyte.com/integrations/sources/aircall-migrations`
image

## How
Update `getFileNamesInDir` to ignore `-migrations.md` as well as the `.inapp.md` pattern

## Why 
To enable us to create migration docs which are navigable via link/direct URL, but unlisted in the sidebar. See https://github.com/airbytehq/airbyte/pull/27294#discussion_r1227419521 re: `migrationDocumentationUrl`


erohmensing:
@bnchrch there is another quite easy option we could do that wouldn't mess with the location of the existing connector docs. Consider a structure like this where we have a `migrations` folder under each of `sources` and `destinations`: 
image

Then the content is hosted at `/integrations/sources/migrations/aircall` instead of `/integrations/sources/aircall-migrations`
image

One thing I notice is that in files unlisted in the sidebar, I think we lose the sidebar 🤔

evantahler (approved):

evantahler:
I think this work.  We then need to be sure to add a link to "my migrations" from each connector's docs page

bnchrch (approved):

erohmensing:
> We then need to be sure to add a link to "my migrations" from each connector's docs page

Only the ones with migrations - but yes!

PR #27358, opened 2023-06-14
Issue 26607/fix cats support array in state

maxi297:
## What
Follow up on https://github.com/airbytehq/airbyte/pull/27306 as Pydantic always try to cast as the first time. source-posthog had a cursor that was "2331" as a string and it was being casted as an int which ended up failing the CATs for this source. So basically, we want "2331" to stay a string and 2331 to stay an int

## How
Pydantic aggressively casts based on the order the Union is defined. In this case, I set it up as Union[int, str] so it would cast as int first and if it didn’t work cast a str. To avoid doing that and keeping the type received, we need to define smart_union = True

### DIFF
--- a/airbyte-integrations/bases/connector-acceptance-test/CHANGELOG.md
+++ b/airbyte-integrations/bases/connector-acceptance-test/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.10.7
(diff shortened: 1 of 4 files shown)


erohmensing (approved):
Thanks for testing both cases!

octavia-approvington:
Lets merge it <br/> sheep thumbs up

octavia-approvington (approved):

PR #27377, opened 2023-06-14
Handle the case of no metadata on master for new connectors

bnchrch:
## What
We had two issues
1. We didnt gracefully handle the null case for tags
2. We didnt gracefully handle when a metadata file was not on master (because the connector was new)

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/tests/common.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/tests/common.py
@@ -31,14 +31,20 @@ class VersionCheck(Step, ABC):
         return f"{self.GITHUB_URL_PREFIX_FOR_CONNECTORS}/{self.context.connector.technical_name}/{METADATA_FILE_NAME}"
 
     @cached_property
-    def master_metadata(self) -> dict:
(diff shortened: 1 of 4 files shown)


octavia-approvington:

octavia-approvington (approved):

tools/ci_connector_ops/ci_connector_ops/pipelines/tests/common.py
```diff
@@ -31,14 +31,20 @@ def github_master_metadata_url(self):
...
 
     @cached_property
-    def master_metadata(self) -> dict:
+    def master_metadata(self) -> Optional[dict]:
         response = requests.get(self.github_master_metadata_url)
-        response.raise_for_status()
+
+        # New connectors will not have a metadata file in master
+        if not response.ok:
+            return None
```
alafanechere, line 39:
```suggestion
        if response.status_code not in [200, 404]:
            response.raise_for_status()
       if response.status_code == 404:
            return None
```

alafanechere (approved):
Please checkout my suggestion.  LGTM Otherwise. Thanks for the fix.

PR #27382, opened 2023-06-14
File-based CDK: Support for incremental syncs

girarda:
## What
Support incremental syncs for file-based connectors.

By default, all file-based connectors' cursor field is `_ab_source_file_last_modified`, which is the timestamp when a record's source was last modified. This is defined by the `DefaultFileBasedStream`.

To ensure we don't lose records due to race conditions, the connection's state keeps track of the last 10k files that were synced.

If the history has more than 10k files, we mark the state as partial, and the connector will set its start_time to the minimum between (earliest timestamp in history, now minus 3 days) to minimize the chances of losing data. The connector will then only sync the files that have a timestamp >= start_time, with a secondary comparison on the URI if the file's timestamp is equal to the timestamp of the earliest file in the history. 

Both the maximum number of files, and the time buffer can be configured by connector developers.

### DefaultFileBasedStream
- Compute the files to sync from the `compute_slices` method. The method groups the files by their last_modified timestamp. Each slice contains all the files for a given timestamp.
- Delegates the state management to a FileBasedCursor

### FileBasedCursor
- State is of the form: {history: {file_uri: ts}, history_is_partial: yes/no}
- The state's history can have at most 10k files
- history_is_partial is set to true if a single sync requires syncing more than 10k files
- if the initial state's history_is_partial is true, we need to set the start_time to min(earliest datetime in history, 3 days ago)

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/file_based/README.md`
2. `airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/abstract_file_based_stream.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py`
4. `airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/file_based_cursor.py`
5. `airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/file_based_stream_config.py`
6. `airbyte-cdk/python/unit_tests/sources/file_based/scenarios/incremental_scenarios.py`
7. The other files


airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/file_based_cursor.py
```diff
@@ -0,0 +1,75 @@
...
+    def is_history_partial(self):
+        return self._history_is_partial
+
+    def get_files_to_sync(self, all_files: List[RemoteFile]):
+        start_time = self._compute_start_time()
+        files_to_sync = [
+            f
+            for f in all_files
+            if (
+                f.last_modified >= start_time
```
girarda, line 48: opening the door to challenging this implementation:

According to the AbstractFileBasedStreamReader.list_matching_files's interface, this all the files passed to this method should have a timestamp >= start_time. 

However, it seems reasonable for the cursor to do this filtering since it owns the start_time.

I also thought about removing the start_time parameter from `AbstractFileBasedStreamReader.list_matching_files`, but I can imagine it being useful for sources that allow filtering based on a timestamp (I don't know if there are any)

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -3,37 +3,63 @@
...
-    def read_records(
-        self,
-        sync_mode: SyncMode,
-        cursor_field: List[str] = None,
-        stream_slice: Optional[StreamSlice] = None,
-        stream_state: Optional[StreamState] = None,
-    ) -> Iterable[Mapping[str, Any]]:
+    def compute_slices(self) -> Iterable[Optional[Mapping[str, Any]]]:
+        # Group all files by timestamps and return them as slices
+        # TODO: Partition the files by glob patterns to enable better checkpointing
```
clnoll, line 54: How would partitioning by glob pattern improve checkpointing?

clnoll (commented):
Nice @girarda!

Couple of comments in addition to the line-specific ones -

If the history is not full, is there any reason not to sync files older than the earliest in the history? Particularly given that we aren't aware of optimizations for querying file-based sources on last-modified date. 

I'm wondering if we need a secondary sort, on the URI, as a tiebreaker. I'm thinking about a (admittedly probably very rare) situation where there are multiple 10ks of files and they're all new. In that situation there's a risk we could
- Sync 10k files uploaded today but set the cursor to 3 days ago.
- Next sync we skip the 10k files that are already in the history, putting in a new 10k.
- Next sync we re-sync the first 10k, because they're no longer in the history.
What I'd propose is a slight modification where we sync the files on both last-modified and URI. If the history is full, we try to sync files after the earliest timestamp and with the URI alphabetically greater than the alphabetically earliest URI (similarly when deleting files from the history). Does that seem reasonable?

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
girarda: It would increase the checkpoint granularity.

Assuming we have two globs 
1. One glob with a single file for datetime = 2021-01-01T00:00:00 "*_a.csv" -> ["my_file_a.csv"]
2. One glob with many files for datetime = 2021-01-01T00:00:00 "*_b.csv" -> ["my_file_b.csv", "another_file_b.csv"]

We can only checkpoint the instant if all globs are synced. Checkpointing by glob would allow us to commit my_file_a.csv even we the _b files are large or if they fail.

The benefits are debatable since I wouldn't expect that many files having the exact same timestamp

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/incremental_scenarios.py
```diff
@@ -0,0 +1,1328 @@
...
+                            },
+                        }
+                    },
+                    "name": "stream1",
+                }
+            ]
+        }
+    )
+    .set_expected_records(
+        [
```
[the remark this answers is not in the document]
girarda: Makes sense. Updated the logic to sync all files that are not in the history if the history is not partial

girarda:
@clnoll 
>   What I'd propose is a slight modification where we sync the files on both last-modified and URI. If the history is full, we try to sync files after the earliest timestamp and with the URI alphabetically greater than the alphabetically earliest URI (similarly when deleting files from the history). Does that seem reasonable?

Super reasonable. I agree with the concern and updated with a secondary sort on the uri

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
clnoll: Thanks, I see what you mean. I suspect that the gains from partitioning by glob will be relatively low - just scrolling through [metabase](https://airbyte.metabaseapp.com/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7InR5cGUiOiJuYXRpdmUiLCJuYXRpdmUiOnsicXVlcnkiOiJzZWxlY3RcbiAgICBDb3VudCgqKSxcbiAgICBjb25maWd1cmF0aW9uIC0-ICdwYXRoX3BhdHRlcm4nXG5mcm9tIGFjdG9yIHdoZXJlIGFjdG9yX2RlZmluaXRpb25faWQgPSAnNjk1ODk3ODEtNzgyOC00M2M1LTlmNjMtODkyNWIxYzFjY2MyJyBHUk9VUCBCWSBjb25maWd1cmF0aW9uIC0-ICdwYXRoX3BhdHRlcm4nOyIsInRlbXBsYXRlLXRhZ3MiOnt9fSwiZGF0YWJhc2UiOjcyfSwiZGlzcGxheSI6InRhYmxlIiwicGFyYW1ldGVycyI6W10sInZpc3VhbGl6YXRpb25fc2V0dGluZ3MiOnt9fQ==), most connections just have a single pattern (fyi globs will be `"|"`-separated when you look through the results).

girarda: fair enough. doesn't seem too useful then

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -3,37 +3,62 @@
...
-        sync_mode: SyncMode,
-        cursor_field: List[str] = None,
-        stream_slice: Optional[StreamSlice] = None,
-        stream_state: Optional[StreamState] = None,
-    ) -> Iterable[Mapping[str, Any]]:
+    def compute_slices(self) -> Iterable[Optional[Mapping[str, Any]]]:
+        # Group all files by timestamps and return them as slices
+        files = [{"uri": f.uri, "last_modified": f.last_modified, "file_type": f.file_type} for f in self.list_files_for_this_sync()]
+
+        slices = [{"files": list(group[1])} for group in itertools.groupby(files, lambda f: f["last_modified"])]
```
maxi297, line 56: Why are we grouping files on a specific time stamp? Is it because it's harder to track state? I just want to acknowledge the downside which is when we will want to parallelize, we might have less benefits (given it's frequent that files are modified at the exact same time)

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -42,35 +67,56 @@ def read_records(
             # On read requests we should always have the catalog available
             raise MissingSchemaError("Expected `json_schema` in the configured catalog but it is missing.")
         parser = self.get_parser(self.config.file_type)
-        for file in self.list_files_for_this_sync(stream_slice):
+        for file_description in stream_slice["files"]:
+            file = RemoteFile.from_file_partition(file_description)
```
maxi297, line 71: It feels very odd that we need to rebuild this thing. Why don't we keep the RemoteFile in the stream_slices? Is it because of the debug logs for the slice?

maxi297 (commented):
I think I'll need to give it another go so I won't approve but I don't have any blocker. I just need to play a little more with the solution

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/file_based_cursor.py
```diff
@@ -0,0 +1,89 @@
...
+            "history": self._file_to_datetime_history,
+            "history_is_partial": self._history_is_partial,
+        }
+        return state
+
+    def is_history_partial(self):
+        return self._history_is_partial
+
+    def _should_sync_file(self, file: RemoteFile) -> bool:
+        if file.uri in self._file_to_datetime_history:
```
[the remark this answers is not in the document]
girarda: Catherine brought up the same point. Updated so we sync the files if their modified_at is more recent than the one in the state's history

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
girarda: I'm fine keeping them as RemoteFile in the slices. updated!

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/abstract_file_based_stream.py
```diff
@@ -65,6 +64,32 @@ def read_records(
     ) -> Iterable[Mapping[str, Any]]:
         """
         Yield all records from all remote files in `list_files_for_this_sync`.
+        This method acts as an adapter between the generic Stream interface and the file-based's
+        stream since file-based streams manage their own states.
+        """
+        return self.read_records_from_slice(stream_slice)
+
+    @abstractmethod
+    def read_records_from_slice(self, stream_slice: StreamSlice) -> Iterable[Mapping[str, Any]]:
```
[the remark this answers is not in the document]
girarda: Agreed. I create a new types file specific to this module so the types can evolve independently

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
girarda: Yes, it's about state management.

One slice per file would trivially work if we could always keep the history in the state, but we need an ordering guarantee since files can be removed from the history.

We can still sync files in parallel by parallelizing [this block within the read_records](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py#L45-L53). We just cannot commit them until they're all synced

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -3,37 +3,59 @@
...
-    def read_records(
-        self,
-        sync_mode: SyncMode,
-        cursor_field: List[str] = None,
-        stream_slice: Optional[StreamSlice] = None,
-        stream_state: Optional[StreamState] = None,
-    ) -> Iterable[Mapping[str, Any]]:
+    def compute_slices(self) -> Iterable[Optional[Mapping[str, Any]]]:
+        # Group all files by timestamps and return them as slices
+        files = self.list_files_for_this_sync()
```
clnoll, line 51: I believe this should be a list sorted by `(last_modified, uri)` (and then we can get rid of the `slices.sort` on line 54). I noticed that when I reordered the items in `multi_csv_same_timestamp_more_files_than_history_size_scenario` the test failed (synced the lexographically earlier file).

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -3,37 +3,59 @@
...
 from airbyte_cdk.sources.file_based.schema_validation_policies import record_passes_validation_policy
 from airbyte_cdk.sources.file_based.stream import AbstractFileBasedStream
+from airbyte_cdk.sources.file_based.stream.cursor import FileBasedCursor
+from airbyte_cdk.sources.file_based.types import StreamSlice
+from airbyte_cdk.sources.streams import IncrementalMixin
 from airbyte_cdk.sources.utils.record_helper import stream_data_to_airbyte_message
 
 
-class DefaultFileBasedStream(AbstractFileBasedStream):
+class DefaultFileBasedStream(AbstractFileBasedStream, IncrementalMixin):
```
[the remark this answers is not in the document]
girarda: I lean towards a unified flow for simplicity. Trying to optimize large full refresh syncs feels like a goose chase because we can't checkpoint.

I can take a stab at splitting the flow, but I suspect we'll still need to use the same class because the [FileBasedSource doesn't know about the sync_mode when instantiating the stream](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py#L80).

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/file_based_cursor.py
```diff
@@ -0,0 +1,31 @@
...
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+#
+
+from abc import ABC, abstractmethod
+from typing import Any, Mapping
+
+from airbyte_cdk.sources.file_based.remote_file import RemoteFile
+
+
+class FileBasedCursor(ABC):
```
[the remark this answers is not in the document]
girarda: agreed. added them.

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/default_file_based_cursor.py
```diff
@@ -0,0 +1,91 @@
...
+            oldest_file = self._compute_earliest_file_in_history()
+            del self._file_to_datetime_history[oldest_file.uri]
+
+    def get_state(self) -> StreamState:
+        state = {
+            "history": self._file_to_datetime_history,
+        }
+        return state
+
+    def is_history_partial(self) -> bool:
```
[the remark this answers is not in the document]
girarda: renamed, added a docstring, and made the method private

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/default_file_based_cursor.py
```diff
@@ -0,0 +1,91 @@
...
+from airbyte_cdk.sources.file_based.types import StreamState
+
+DATE_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
+
+
+class DefaultFileBasedCursor(FileBasedCursor):
+    def __init__(self, max_history_size: Optional[int], days_to_sync_if_history_is_full: Optional[int]):
+        self._file_to_datetime_history: Mapping[str:datetime] = {}
+        self._max_history_size = max_history_size or 10_000
+        self._time_window_if_history_is_full = timedelta(days=days_to_sync_if_history_is_full or 3)
```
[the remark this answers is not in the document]
girarda: done

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/default_file_based_cursor.py
```diff
@@ -0,0 +1,91 @@
...
+            )
+        for f in all_files:
+            if self._should_sync_file(f):
+                yield f
+
+    def get_start_time(self) -> datetime:
+        return self._start_time
+
+    def _compute_earliest_file_in_history(self) -> RemoteFile:
+        filename, last_modified = min(self._file_to_datetime_history.items(), key=lambda f: (f[1], f[0]))
```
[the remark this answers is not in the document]
girarda: yes, added a test to cover that case and fixed

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/default_file_based_cursor.py
```diff
@@ -0,0 +1,91 @@
...
+        }
+        return state
+
+    def is_history_partial(self) -> bool:
+        return len(self._file_to_datetime_history) >= self._max_history_size
+
+    def _should_sync_file(self, file: RemoteFile) -> bool:
+        if file.uri in self._file_to_datetime_history:
+            # If the file's uri is in the history, we should sync the file if it has been modified since it was synced
+            return file.last_modified > datetime.strptime(self._file_to_datetime_history.get(file.uri), DATE_TIME_FORMAT)
```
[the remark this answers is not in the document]
girarda: Added a warning.

I don't feel super strongly about whether we should sync the file if its last-modified is older than the one in the history, but it could have consequences on what else we sync if the history is full.

Assuming a max size of 3.

initial state:
```
history: {
  "a.csv": "2021-05-05",
  "b.csv": "2021-05-06",
  "c.csv": "2021-01-07",
}
```
Sync 1: we read new files
```
history: {
  "c.csv": "2021-01-07",
  "d.csv": "2021-05-08",
  "e.csv": "2021-05-09",
}
```

Sync 2: c.csv's last-modified is set to a value in the past (2021-05-04)
```
history: {
  "c.csv": "2021-01-04",
  "d.csv": "2021-05-08",
  "e.csv": "2021-05-09",
}
```

Sync 3: We need to sync a.csv and b.csv again because their last-modified is later than c.csv's

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
girarda: good catch. fixed and updated the test to cover that case

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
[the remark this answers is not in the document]
clnoll: Good point re the simplicity. Let's leave it as-is for now. If we need to do an optimization later we always can.

clnoll (approved):
Looks great @girarda!

airbyte-cdk/python/unit_tests/sources/file_based/in_memory_files_source.py
```diff
@@ -48,7 +48,7 @@ def get_matching_files(
         from_date: Optional[datetime] = None,
```
maxi297, line 48: Is there a reason we don't use `from_date` here? Shouldn't we filter files based on `last_modified`?

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/default_file_based_cursor.py
```diff
@@ -0,0 +1,120 @@
...
+            logger.warning(
+                f"The state history is full. "
+                f"This sync and future syncs won't be able to use the history to filter out duplicate files. "
+                f"It will instead use the time window of {self._time_window_if_history_is_full} to filter out files."
+            )
+        for f in all_files:
+            if self._should_sync_file(f, logger):
+                yield f
+
+    def get_start_time(self) -> datetime:
```
maxi297, line 101: Do we have an opinion as to if datetimes should be naive or not? I'm asking because the return `datetime` does not explain the extend as to which the user will be able to use the output (for example, Python can't compare a naive datetime against a aware one)

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -42,35 +61,52 @@ def read_records(
...
+    def cursor_field(self) -> Union[str, List[str]]:
+        """
+        Override to return the default cursor field used by this stream e.g: an API entity might always use created_at as the cursor field.
+        :return: The name of the field used as a cursor. If the cursor is nested, return an array consisting of the path to the cursor.
         """
-        Return the JSON Schema for a stream.
+        return self.ab_last_mod_col
 
-        If the user provided a schema, return that. Otherwise, infer it.
+    @cache
```
maxi297, line 87: Can we document why this is cached? I'm very afraid of this one because it is not caching on sometime immutable (like [Python factorial example](https://docs.python.org/3/library/functools.html)) but on something with a state. I can see a couple of surprising behavior:
* As a user of `DefaultFileBasedStream`, I can see files in the file storage that won't be outputed by `list_files`. This seems to indicate that the name is not precise enough. It can be even more confusing since this could be added to the cache because I call `get_json_schema` and not `list_files`
* `list_files_for_this_sync ` mentions that it's using the output of `list_files` but it is actually using the "uncached" version as it's redoing the same processing and not calling the method

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -42,35 +61,52 @@ def read_records(
...
+            self.ab_file_name_col: {"type": "string"},
+        }
+        schema = self.get_raw_json_schema()
+        schema["properties"] = {**extra_fields, **schema.get("properties", {})}
+        return schema
 
-        Use no more than `_discovery_policy.max_n_files_for_schema_inference` files.
-        """
+    @cache
+    def get_raw_json_schema(self) -> Mapping[str, Any]:
```
maxi297, line 98: Is this public? If so, why? The same question applies for `list_files`

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/abstract_file_based_stream.py
```diff
@@ -83,7 +108,7 @@ def list_files(self) -> List[RemoteFile]:
         ...
 
     @abstractmethod
-    def list_files_for_this_sync(self, stream_slice: Optional[StreamSlice]) -> Iterable[RemoteFile]:
+    def list_files_for_this_sync(self) -> Iterable[RemoteFile]:
```
maxi297, line 111: I find this interface a little weird. It implies the notion of sync, but it's not passed anywhere. The reason seems to be that the object has a state and can't be reused for more than one sync. Do we need to make this object stateful?

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/file_based_cursor.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @abstractmethod
+    def get_state(self) -> Mapping[str, Any]:
+        """
+        Get the state of the cursor.
+        """
+        ...
+
+    @abstractmethod
+    def get_start_time(self) -> datetime:
```
maxi297, line 42: It feels to me that we should avoid to expose internals outside of the class. Would it make more sense to have a method `should_be_synced` like we added for the declarative package?

maxi297 (approved):
LGTM! I added a couple of concerns/things I would like to understand but nothing blocking

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/file_based_cursor.py
girarda: The reason the start_time is exposed is that it's possible specific implementation of the StreamReader can leverage the start time to only list recent files (https://github.com/airbytehq/airbyte/pull/27382/files#diff-e07a7a23c8203f8a61e7c9d31cb495d6d9f8c7e1356ddff25630e704431d404cR127). A `should_be_synced` interface wouldn't allow that kind of optimization.

That being said, it doesn't seem like it's something we can do for S3 so I can remove it from the interface and we can revisit if we ever need the start date.

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
girarda: good catch. I don't think either need to be public (and we should be able to remove `list_files` from the interface)

PR #27391, opened 2023-06-15
fix trigger on format workflow

alafanechere:
`branches-ignore` can't be used in combination of `branches`.
Keep only `branches` for testing.


octavia-approvington:
It's time <br/> fine lets go

octavia-approvington (approved):

PR #27398, opened 2023-06-15
Maxi297/refactoring declarative state management

maxi297:
## What
Problems:
* It is hard to follow the lifecycle of stream_state
    * The interfaces are not explicit in what it does
        * stream_state passed as a parameter but should not be used. This is a problem since the following implementation relies on the parameter and not their internal state:
             * DatetimeBasedCursor
             * greenhouse
             * intercom
             * posthog
             * railz
             * square
             * zenloop
    * Each method should only do one thing
        * Right now, update_cursor does two things: set up the initial state, update based on a new record
* Some objects provide state while they don’t really have

Note that these problems have a somewhat broad impact on the code base and I went on a killing spree. I'm willing to review the scope and do smaller change if we think it would be wise.

## How
For problem # 1, remove argument `stream_state` from everything below DeclarativeStream.
Note: there is an exception to this and it is for interpolation where source-intercom and source-railz use interpolation for creating the request and filtering.

For problem # 2, create separate interface for StreamSlicer and Cursor where StreamSlicer does not have notions of state. If state is used to determine slices, it will rely on the internals of the class.

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/cursor.py` and `airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/cartesian_product_stream_slicer.py`
2. The rest

## 🚨 User Impact 🚨
The following are a breaking changes:
* `CustomRetriever` (no usage in repository so I'm not worried about this one)
    * `read_records` arg `stream_state` removed
    * if inheriting from SimpleRetriever, split `self.stream_slicer` into `self.stream_slicer` and `self.cursor`
* `CustomIncrementalSync`
    * `update_cursor` split into `set_initial_state` and `update_state`
    * `stream_slices` has doesn't take any parameters anymore
* `CustomPartitionRouter` (no usage in repository so I'm not worried about this one)
    * if `update_cursor` and `get_stream_state` were implemented and relevant, should be migrated to `CustomIncrementalSync`
    * `stream_slices` has doesn't take any parameters anymore
* `DatetimeBasedCursor`: Removed interpolation based on stream_state as part of. Reasons are:
    * It wasn't used
    * The state can change at any point so it's hard for the user to know which state will be used
    * I fear this might create a lot of headaches when we will try to parallelize the requests and I think we prefer parallelization to stream_state interpolation
* `SubstreamPartitionRouter` was passing state to parent_stream. This could be a problem is if parent and child stream were actually sharing the same state with the same cursor parameters (cursor_field, datetime format). As of today, it is not clear to me what is the expectation from the user in terms of incremental stream where the parent is incremental as well and I feel it is somewhat dangerous when a user configure this

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
@@ -13,6 +13,7 @@ from airbyte_cdk.models import AirbyteLogMessage, AirbyteMessage, Level, SyncMod
 from airbyte_cdk.models import Type as MessageType
 from airbyte_cdk.sources.declarative.exceptions import ReadException
 from airbyte_cdk.sources.declarative.extractors.http_selector import HttpSelector
+from airbyte_cdk.sources.declarative.incremental.cursor import Cursor
 from airbyte_cdk.sources.declarative.interpolation import InterpolatedString
 from airbyte_cdk.sources.declarative.partition_routers.single_partition_router import SinglePartitionRouter
 from airbyte_cdk.sources.declarative.requesters.error_handlers.response_action import ResponseAction
@@ -47,6 +48,7 @@ class SimpleRetriever(Retriever, HttpStream):
         record_selector (HttpSelector): The record selector
         paginator (Optional[Paginator]): The paginator
         stream_slicer (Optional[StreamSlicer]): The stream slicer
+        cursor (Optional[cursor]): The cursor
         parameters (Mapping[str, Any]): Additional runtime parameters to be used for string interpolation
     """
 
@@ -62,6 +64,7 @@ class SimpleRetriever(Retriever, HttpStream):
     _primary_key: str = field(init=False, repr=False, default="")
     paginator: Optional[Paginator] = None
     stream_slicer: Optional[StreamSlicer] = SinglePartitionRouter(parameters={})
+    cursor: Optional[Cursor] = None
     emit_connector_builder_messages: bool = False
     disable_retries: bool = False
 
@@ -384,7 +387,6 @@ class SimpleRetriever(Retriever, HttpStream):
         sync_mode: SyncMode,
         cursor_field: Optional[List[str]] = None,
         stream_slice: Optional[StreamSlice] = None,
-        stream_state: Optional[StreamState] = None,
     ) -> Iterable[StreamData]:
         # Warning: use self.state instead of the stream_state passed as argument!
         stream_slice = stream_slice or {}  # None-check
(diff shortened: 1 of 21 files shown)


airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/stream_slicer.py
```diff
@@ -23,24 +23,10 @@ class StreamSlicer(RequestOptionsProvider):
     """
 
     @abstractmethod
-    def stream_slices(self, sync_mode: SyncMode, stream_state: StreamState) -> Iterable[StreamSlice]:
+    def stream_slices(self, sync_mode: SyncMode) -> Iterable[StreamSlice]:
```
maxi297, line 26: I think we can probably remove `sync_mode` from the interface as well but we should probably answer the questions in the description in `User Impact -> SubstreamPartitionRouter`

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -62,6 +64,7 @@ class SimpleRetriever(Retriever, HttpStream):
     _primary_key: str = field(init=False, repr=False, default="")
     paginator: Optional[Paginator] = None
     stream_slicer: Optional[StreamSlicer] = SinglePartitionRouter(parameters={})
```
girarda, line 66: would it make sense to change the type of this field to `PartitionRouter` and completely decouple the cursor from the partition routers in the retriever?

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -400,7 +402,13 @@ def read_records(
         # * Why is this abstraction not on the DeclarativeStream level? DeclarativeStream does not have a notion of stream slicers and we
         #    would like to avoid exposing the stream state outside of the cursor. This case is needed as of 2023-06-14 because of
         #    interpolation.
-        slice_state = self.stream_slicer.select_state(stream_slice) if hasattr(self.stream_slicer, "select_state") else stream_state
+        if self.cursor and hasattr(self.cursor, "select_state"):
```
girarda, line 405: do we still need the hasattr if we introduce the Cursor interface?

maxi297: `select_state` still doesn't make sense as part of DatetimeBasedCursor and therefore I'm not so sure how it'll evolve as part of the cursor interface. If you think otherwise, I don't mind changing this though

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
maxi297: Not exactly as if there is a cursor, it'll be the stream slicer (see https://github.com/airbytehq/airbyte/pull/27398/files#diff-bbfbb380b441567d8e40967c149ae0397e41357a213f05add6d3b72cf92a91fdR794-R795). As of today, I think the simple retriever doesn't know about partitions. Does that sound weird to you?

airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/stream_slicer.py
maxi297: After [discussing this internally](https://airbytehq-team.slack.com/archives/C02TL38U5L7/p1686939006489719), we decided to remove the `sync_mode` from the interface of this method

PR #27429, opened 2023-06-16
cat: increase docker connection timeout to 2mn

### LINKED ISSUE #27401: connectors-ci: increase docker connection timeout in CAT
In our test pipeline we spawn a shared docker engine which is shared across connectors under test to run CAT.
This engine can be under heavy load and can take some time to respond, leading to timeout of connection to docker from CAT. We should increase the timeout value of connection to docker engine from CAT. 
Probably something to tweak here: https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/utils/connector_runner.py#L1

alafanechere:
## What
Fixes https://github.com/airbytehq/airbyte/issues/27401

## How
Increase the docker client connection timeout to 2mn in the `ConnectorRunner`

### DIFF
--- a/airbyte-integrations/bases/connector-acceptance-test/CHANGELOG.md
+++ b/airbyte-integrations/bases/connector-acceptance-test/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.10.8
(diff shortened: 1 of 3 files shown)


erohmensing (approved):
Very reasonable upper bound, lets try it!

bnchrch (approved):

PR #27441, opened 2023-06-16
airbyte-ci: validation of breaking change entries for major versions

### LINKED ISSUE #27065: Breaking change metadata validation: validate metadata existence and correctness
Add a new validators to [metadata_validator.py](https://github.com/airbytehq/airbyte/blob/master/airbyte-ci/connectors/metadata_service/lib/metadata_service/validators/metadata_validator.py) that validate:

1. If the version in the metadata file is a major version (n.0.0), the `releases` dict exists and contains a breakingChanges entry for `n.0.0`

Notes: 
2. We're relying on the new version being a major (n.0.0) - this wouldn't trigger on a major version bump that isn't n.0.0, e.g. if you for some reason bumped from 1.0.1 to 2.0.1 instead of bumping to 2.0.0. This is potentially something we could try to validate (or maybe we already do?), that the only valid version bumps are one patch, one minor, or one major. But the auto-bump version project will probably make that moot anyway. 
3. when working on auto bump version, we need to run these validations where we expect to perform a version bump, not once the bump has already happened (since by then the code is already merged and the connector dev cannot add the info - the build will just fail i guess)

erohmensing:
## What
* Validate that if the metadata file indicates the current (new) version is a breaking change, there must be a breakingChange entry for that version. 
* Close https://github.com/airbytehq/airbyte/issues/27065

## How
* Add a validation step to be run pre-publish that checks this.


erohmensing:
This PR is in draft mode because I wasn't able to run tests directly with `poetry run pytest` in `airbyte-ci/connectors/metadata_service/lib`. I'm not sure if those tests run in CI.

bnchrch (approved):
A few comments but Nothing blocking

erohmensing:
Lost all of my commits in a weird force push 🤦🏻 continuing this here https://github.com/airbytehq/airbyte/pull/27668

PR #27445, opened 2023-06-16
Maxi297/refactoring declarative state management

maxi297:
This is a duplicate of https://github.com/airbytehq/airbyte/pull/27398 but I accidentally merged so I'm re-opening it here

## What
Problems:
* It is hard to follow the lifecycle of stream_state
    * The interfaces are not explicit in what it does
        * stream_state passed as a parameter but should not be used. This is a problem since the following implementation relies on the parameter and not their internal state:
             * DatetimeBasedCursor
             * greenhouse
             * intercom
             * posthog
             * railz
             * square
             * zenloop
    * Each method should only do one thing
        * Right now, update_cursor does two things: set up the initial state, update based on a new record
* Some objects provide state while they don’t really have

Note that these problems have a somewhat broad impact on the code base and I went on a killing spree. I'm willing to review the scope and do smaller change if we think it would be wise.

## How
For problem # 1, remove argument `stream_state` from everything below DeclarativeStream.
Note: there is an exception to this and it is for interpolation where source-intercom and source-railz use interpolation for creating the request and filtering.

For problem # 2, create separate interface for StreamSlicer and Cursor where StreamSlicer does not have notions of state. If state is used to determine slices, it will rely on the internals of the class.

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/cursor.py` and `airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/cartesian_product_stream_slicer.py`
2. The rest

## 🚨 User Impact 🚨
The following are a breaking changes:
* `CustomRetriever` (no usage in repository so I'm not worried about this one)
    * `read_records` arg `stream_state` removed
    * sync_mode, cursor_field and stream_state removed from `stream_slices`
    * if inheriting from SimpleRetriever, split `self.stream_slicer` into `self.stream_slicer` and `self.cursor`
* `CustomIncrementalSync`
    * `update_cursor` split into `set_initial_state` and `update_state`
    * `stream_slices` has doesn't take any parameters anymore
* `CustomPartitionRouter` (no usage in repository so I'm not worried about this one)
    * if `update_cursor` and `get_stream_state` were implemented and relevant, should be migrated to `CustomIncrementalSync`
    * `stream_slices` has doesn't take any parameters anymore
* `DatetimeBasedCursor`: Removed interpolation based on stream_state as part of. Reasons are:
    * It wasn't used
    * The state can change at any point so it's hard for the user to know which state will be used
    * I fear this might create a lot of headaches when we will try to parallelize the requests and I think we prefer parallelization to stream_state interpolation
* `SubstreamPartitionRouter` was passing state to parent_stream. This could be a problem is if parent and child stream were actually sharing the same state with the same cursor parameters (cursor_field, datetime format). As of today, it is not clear to me what is the expectation from the user in terms of incremental stream where the parent is incremental as well and I feel it is somewhat dangerous when a user configure this


airbyte-cdk/python/airbyte_cdk/sources/declarative/stream_slicers/stream_slicer.py
```diff
@@ -23,24 +22,9 @@ class StreamSlicer(RequestOptionsProvider):
     """
 
     @abstractmethod
-    def stream_slices(self, sync_mode: SyncMode, stream_state: StreamState) -> Iterable[StreamSlice]:
+    def stream_slices(self) -> Iterable[StreamSlice]:
```
maxi297, line 25: After [discussing this internally](https://airbytehq-team.slack.com/archives/C02TL38U5L7/p1686939006489719), we decided to remove the sync_mode from the interface of this method

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
```diff
@@ -97,59 +95,51 @@ def __post_init__(self, parameters: Mapping[str, Any]):
...
-            self._cursor = max(cursor, self._cursor)
-        elif cursor:
-            self._cursor = cursor
-        if self.partition_field_end:
-            self._cursor_end = stream_slice_value_end
-
-    def stream_slices(self, sync_mode: SyncMode, stream_state: Mapping[str, Any]) -> Iterable[Mapping[str, Any]]:
+
+        possible_cursor_values = list(filter(lambda item: item, [last_record_value, self._cursor]))
+        self._cursor = max(possible_cursor_values) if possible_cursor_values else None
```
maxi297, line 118: Note that there is a bug here if the datetime format can't be ordered by alphabetical order since we compare strings. For example, datetime format `%d-%m-%Y` will not work as expected. I have created an issue for this https://github.com/airbytehq/airbyte/issues/27447

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
```diff
@@ -97,59 +95,51 @@ def __post_init__(self, parameters: Mapping[str, Any]):
...
 
         The start of the window is the minimum datetime between start_datetime - lookback_window and the stream_state's datetime
         The end of the window is the minimum datetime between the start of the window and end_datetime.
 
-        :param sync_mode:
-        :param stream_state: current stream state. If set, the start_date will be the day following the stream_state.
         :return:
         """
-        stream_state = stream_state or {}
-        kwargs = {"stream_state": stream_state}
```
maxi297, line 138: I have removed interpolation based on stream_state as part of. Reasons are:
* It wasn't used
* The state can change at any point so it's hard for the user to know which state will be used
* I fear this might create a lot of headaches when we will try to parallelize the requests and I think we prefer parallelization to stream_state interpolation

maxi297:
Build is failing because of `Wandalen/wretry.action` latest update but the PR should be good to go anyway

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -574,7 +574,6 @@ definitions:
         type: string
         interpolation_context:
           - config
-          - stream_state
```
girarda, line 577: nice

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
girarda: This is a good move. Developers shouldn't ever need to reference the state.

girarda (approved):

airbyte-cdk/python/unit_tests/sources/declarative/partition_routers/test_list_partition_router.py
```diff
@@ -37,29 +36,10 @@
...
 
-@pytest.mark.parametrize(
-    "test_name, stream_slice, last_record, expected_state",
-    [
-        ("test_update_cursor_no_state_no_record", {}, None, {}),
-        ("test_update_cursor_with_state_no_record", {"owner_resource": "customer"}, None, {"owner_resource": "customer"}),
-        ("test_update_cursor_value_not_in_list", {"owner_resource": "invalid"}, None, {}),
-    ],
-)
-def test_update_cursor(test_name, stream_slice, last_record, expected_state):
```
clnoll, line 52: Do we need to update any of these component-specific tests to replace tests for `update_cursor` with tests that hit the `Cursor.update_state` codepaths?

clnoll (commented):
Just a few minor q's for you @maxi297

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
```diff
@@ -97,59 +95,51 @@ def __post_init__(self, parameters: Mapping[str, Any]):
...
 
-    def update_cursor(self, stream_slice: StreamSlice, last_record: Optional[Record] = None):
+    def set_initial_state(self, stream_state: StreamState) -> None:
+        """
+        Cursors are not initialized with their state. As state is needed in order to function properly, this method should be called
+        before calling anything else
+
+        :param stream_state: The state of the stream as returned by get_stream_state
+        """
+        self._cursor = stream_state.get(self.cursor_field.eval(self.config)) if stream_state else None
```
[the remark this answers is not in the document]
maxi297: I don't have any strong opinion. If we were to do this:

### Benefits
* It would prevent weird error where the state initialized multiple time (behavior that we assume is the current one)

### Drawbacks
* This would change the actual behavior
* This would need to be done on every implementations

Personally, I think the real solution is that any cursor receives their state as part of `__init__` but this is very backward to the way this is done today and would imply changing the entrypoint and therefore probably affecting all the sources unfortunately...

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
```diff
@@ -97,59 +95,51 @@ def __post_init__(self, parameters: Mapping[str, Any]):
...
+
+    def update_state(self, stream_slice: StreamSlice, last_record: Record) -> None:
```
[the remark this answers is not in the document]
maxi297: My reasoning was that `cursor` is the interface of the class so `update_cursor` is redundant. `state` also makes some sense given that if I `update_state`, the expectation is that the output of `get_stream_state` will change

airbyte-cdk/python/unit_tests/sources/declarative/partition_routers/test_list_partition_router.py
maxi297: PartitionRouters do not have the `Cursor` interface anymore which means they don't have `update_state` and `get_stream_state`. This makes sense as they don't have any state of their own. Does that make sense?

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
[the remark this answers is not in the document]
clnoll: Got it, given that the current behavior doesn't do this check I'm fine not adding it in here.

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_stream.py
```diff
@@ -143,8 +146,7 @@ def stream_slices(
 
         :param sync_mode:
         :param cursor_field:
-        :param stream_state:
+        :param stream_state: we knowingly avoid using stream_state as we want cursors to manage their own state
         :return:
         """
-        # this is not passing the cursor field because it is known at init time
-        return self.retriever.stream_slices(sync_mode=sync_mode, stream_state=stream_state)
+        return self.retriever.stream_slices()
```
lazebnyi, line 152: If `retriever.stream_slices` only accepts an instance of a class, do we require any additional arguments to be awaited in the `stream_slices` method of the `DeclarativeStream` class?

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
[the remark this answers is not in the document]
clnoll: 👍 I'm good with that.

airbyte-cdk/python/unit_tests/sources/declarative/partition_routers/test_list_partition_router.py
clnoll: Yep!

clnoll (approved):

lazebnyi (approved):
LGTM! Small question above.

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_stream.py
maxi297: The `Stream` interface [requires those parameters to be there](https://github.com/airbytehq/airbyte/blob/3df819612ffa9ab34803483de69b9578cac92cf7/airbyte-cdk/python/airbyte_cdk/sources/streams/core.py#L206). This is [the contract between the AbstractSource and the Stream](https://github.com/airbytehq/airbyte/blob/3df819612ffa9ab34803483de69b9578cac92cf7/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py#L245-L248). This is why we need to keep the parameters on the `DeclarativeStream` level but we won't keep it. Does that make sense?

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -791,14 +791,17 @@ def create_simple_retriever(
             else NoPagination(parameters={})
         )
 
+        stream_slicer = stream_slicer or SinglePartitionRouter(parameters={})
+        cursor = stream_slicer if isinstance(stream_slicer, Cursor) else None
```
brianjlai, line 795: This block feels a bit weird. In the case of a `DatetimeBasedCursor`, this results in both the `stream_slicer` and the `cursor` getting set as the same component. So I like the idea of the Cursor and the StreamSlicer having distinct responsibilities, but when the Cursor is a subclass of the StreamSlicer, it does lead to this duplicative assignment. So it got me thinking about our object model a bit.

Its a little late so this could just be very stream of conscious, but I think it could be an interesting thought to even further decouple the concepts of the date ranges and the cursor state. So instead of Cursor inheriting StreamSlicer, they would be their own distinct interfaces. For example, we would have the date cursor only used to maintain state and a SinglePartitionRouter where the slice is just over the entire state -> present. And the SimpleRetriever is truly composed of one StreamSlicer and one Cursor. 

This would lead to a bigger refactor in the SimpleRetriever and DatetimeBasedCursor since they'd have to work across interface boundaries to persist state when we finish a range. Maybe we could discuss this synchronously if this is an idea worth fleshing out

brianjlai (commented):
I like the refactor and this separation of Cursor and Slicer responsibilities is a good idea. And same for with quality of life improvements to not even mention the `stream_state` parameter. 

I think the PR is in good shape to merge as is, but I did leave one comment about the component relationships and would be curious to see your take on it. And if we don't go down that route, then I'm ready to 👍

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
brianjlai: Discussed this w/ maxime offline. we think a greater refactor to truly decouple the Cursor inheritance from the StreamSlicer might be a good idea later down the line, but it is at this moment not strictly necessary to solve in the immediate for what is being implemented for per-stream

The conceptual change of decoupling cursors and stream slicers is that we can think of them as distinctive pieces of functionality in that a cursor is used to support state/incremental syncs and stream slicing handles creating date ranges or other slices. In theory you could have incremental syncs w/ one giant slice or date range slices over the entire time period. The simple retriever would then act as an orchestrator of components like it already does. If we do find a use case for this then we can tie the refactor to that need.

brianjlai (approved):
discussing a few things about the object model synchronously. we don't need to take action on it and changes look good!

PR #27465, opened 2023-06-19
fix: pin retry action version as latest is not stable

alafanechere:
## What
The GH marketplace action we're using to retry steps is not pinned to a specific version.
The latest version has a bug https://github.com/Wandalen/wretry.action/issues/93

## How
We shall pin to the last stable version (1.0.42) to get our workflow running.


octavia-approvington:

octavia-approvington (approved):

PR #27476, opened 2023-06-19
File CDK: define streams via glob list

### LINKED ISSUE #26760: File CDK: Handle glob config option

clnoll:
## What
Allows streams to be defined as a list of globs; all files matching any of the globs will be considered part of the stream.

## How
Uses the [wmatch](https://github.com/facelessuser/wcmatch) library to determine whether a glob pattern matches a filepath string. The rules for matching are here: https://facelessuser.github.io/wcmatch/glob/#split. This is the same library that's used in the existing `source_files_abstract`.

Note: I tried to find a python library with more adoption that would do something similar, but unfortunately could not. The two built-ins for unix-style pattern matching have issues that would not make this functionality flexible enough: [`fnmatch`](https://docs.python.org/3/library/fnmatch.html) doesn't support `**`, and [`glob`](https://docs.python.org/3/library/glob.html) only works on the local filesystem (due to the underlying C library that it uses).

Closes https://github.com/airbytehq/airbyte/issues/26760.


airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -54,11 +54,22 @@ def filter_files_by_globs(files: List[RemoteFile], globs: List[str]) -> List[Rem
...
+        for file in files:
+            for g in globs:
+                if globmatch(file.uri, g, flags=GLOBSTAR):
+                    matches[file.uri] = file
+
+        return list(matches.values())
 
     @staticmethod
-    def get_prefixes_from_globs(files: List[RemoteFile], globs: List[str]) -> List[str]:
+    def get_prefixes_from_globs(globs: List[str]) -> List[str]:
```
brianjlai, line 70: can we add unit tests for this method?

airbyte-cdk/python/unit_tests/sources/file_based/test_file_based_stream_reader.py
```diff
@@ -0,0 +1,71 @@
...
+    (["**/*.csv", "**/*.gz"], {"a.csv", "a.csv.gz", "a/b.csv", "a/b.csv.gz", "a/c.csv", "a/c.csv.gz", "a/b/c.csv", "a/b/c.csv.gz", "a/c/c.csv", "a/c/c.csv.gz", "a/b/c/d.csv", "a/b/c/d.csv.gz"}, set()),
+    (["*.csv", "*.gz"], {"a.csv", "a.csv.gz"}, set()),
+    (["a/*.csv", "a/*/*.csv"], {"a/b.csv", "a/c.csv", "a/b/c.csv", "a/c/c.csv"}, {"a"}),
+    (["a/*.csv", "a/b/*.csv"], {"a/b.csv", "a/c.csv", "a/b/c.csv"}, {"a", "a/b"}),
+]
+
+
+@pytest.mark.parametrize(
+    "globs,expected_matches,expected_path_prefixes",
+    cases,
```
brianjlai, line 66: nit: is the plan to reuse these cases across more test methods? If we don't plan to reuse these, it might make more sense to just declare each one as `pytest.param([], set(FILEPATHS), set(), id="test_name_whatever")` and define them within the `parametrize` block like we do other tests. I'm not sure the id loop after makes it easier to read

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -54,11 +54,22 @@ def filter_files_by_globs(files: List[RemoteFile], globs: List[str]) -> List[Rem
...
         """
-        return [file for file in files if any(fnmatch(file.uri, glob) for glob in globs)]
+        if not globs:
+            return files
+
+        matches = {}
+
+        for file in files:
+            for g in globs:
+                if globmatch(file.uri, g, flags=GLOBSTAR):
```
brianjlai, line 64: nit: can we add a comment here explaining that `GLOBSTAR` enables the recursive `**` matching? i had to look up what this did and if it saves an extra click for the next person might be worth it

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -54,11 +54,22 @@ def filter_files_by_globs(files: List[RemoteFile], globs: List[str]) -> List[Rem
...
+                    matches[file.uri] = file
+
+        return list(matches.values())
```
brianjlai, line 67: For a function with nested that could operate on a large set and potentially highly recursive, it might be worthwhile to return an `Iterable[RemoteFile]` so we can continue doing work as more files are found. The main drawbacks would be if we require logic that requires looking at the whole set of files or if it adds a lot of complexity to what is invoking this function (which hasn't been implemented yet afaict). Curious to get your take on this?

brianjlai (commented):
no major issues, just a few nits, but wanted to dig into a few aspects I was curious about

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -54,11 +54,22 @@ def filter_files_by_globs(files: List[RemoteFile], globs: List[str]) -> List[Rem
         """
         Utility method for filtering files based on globs.
         """
-        return [file for file in files if any(fnmatch(file.uri, glob) for glob in globs)]
+        if not globs:
+            return files
```
girarda, line 58: why is this defaulting to all files? I'd expect no globs to mean no files, or to result in a configuration error

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -54,11 +54,22 @@ def filter_files_by_globs(files: List[RemoteFile], globs: List[str]) -> List[Rem
...
+
+        matches = {}
+
+        for file in files:
+            for g in globs:
+                if globmatch(file.uri, g, flags=GLOBSTAR):
+                    matches[file.uri] = file
```
girarda, line 65: do we want to log something if a file matches multiple globs?

girarda (approved):

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py
clnoll: That makes sense to me and I've updated this to be a generator.

Some additional context - the originator of the call to this function will be from the Stream class's `list_files` method, which returns a List so that we can cache the results rather than hitting the source multiple times, if we needed to `list_files` more than once. (Since `list_files` returns a list and that's the only entrypoint into this `filter_files_by_globs`, there aren't perf improvements from making it a generator so I left it as a list, which felt more natural for deduping files.) I went back and forth on whether that caching should be on the stream's method versus the stream reader, which is what's actually making the calls to the source, but ultimately put it on the stream so that it's still easy to bypass it - e.g. in cases where the stream reader is paging through results - we don't necessarily want to block while paging. 

FWIW the existing [source_files_abstract code](https://github.com/airbytehq/airbyte/blob/d9a5e2d873a036a464c955aec7f5809c3760f432/airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/stream.py#L157) also works on the whole list (for caching) and AFAIK there haven't been OOM issues from it, but it is in the back of my mind as something that we should keep an eye on. I'm tempted to get rid of the caching as a protective measure, under the assumption that `list_files` calls should be rare so it should be unlikely that we ever do end up calling it more than once. WDYT?

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py
clnoll: They're rolled into the unit tests for `filter_files_by_globs`. Updated the test name to make that clearer.

airbyte-cdk/python/setup.py
```diff
@@ -62,6 +62,7 @@
         "Deprecated~=1.2",
         "Jinja2~=3.1.2",
         "cachetools",
+        "wcmatch==8.4.1",
```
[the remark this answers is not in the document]
clnoll: Good call, updated.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py
brianjlai: ah got it. Thanks for the context on the caching and how we're turning it into a list further upstream. Knowing all this and that the existing connector wasn't running into any memory issues, I don't have super strong opinions that one is a must over the other. 

And yeah framed the way you mention it, it doesn't feel like we need a cache here and even if its lightweight and just in the Python application code. So I'm on board with not having one

clnoll: Actually there's another consideration for listing files, which is that some sources (S3) don't allow us to sort by last-modified, so if we want to support incremental we'll have to sort ourselves, meaning we'll have to retrieve all the files first. So I think I'll leave the cache in for now.

PR #27412, opened 2023-06-15
Run CI Checks for #27412

marcosmarxm:
## What
*Describe what the change is solving*
*It helps to add screenshots if it affects the frontend.*

## How
*Describe the solution*

## Recommended reading order
1. `x.java`
2. `y.python`

## 🚨 User Impact 🚨
*Are there any breaking changes? What is the end result perceived by the user?*

*For connector PRs, use this section to explain which type of semantic versioning bump occurs as a result of the changes. Refer to our [Semantic Versioning for Connectors](https://docs.airbyte.com/contributing-to-airbyte/#semantic-versioning-for-connectors) guidelines for more information. **Breaking changes to connectors must be documented by an Airbyte engineer (PR author, or reviewer for community PRs) by using the [Breaking Change Release Playbook](https://docs.google.com/document/d/1VYQggHbL_PN0dDDu7rCyzBLGRtX-R3cpwXaY8QxEgzw/edit).***

*If there are breaking changes, please merge this PR with the 🚨🚨 emoji so changelog authors can further highlight this if needed.*

## Pre-merge Actions
*Expand the relevant checklist and delete the others.*

<details><summary><strong>New Connector</strong></summary>

### Community member or Airbyter

- **Community member?** Grant edit access to maintainers ([instructions](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests))
- Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run `./gradlew :airbyte-integrations:connectors:<name>:integrationTest`.
- Connector version is set to `0.0.1`
    - `Dockerfile` has version `0.0.1`
- Documentation updated
    - Connector's `README.md`
    - Connector's `bootstrap.md`. See [description and examples](https://docs.google.com/document/d/1ypdgmwmEHWv-TrO4_YOQ7pAJGVrMp5BOkEVh831N260/edit?usp=sharing)
    - `docs/integrations/<source or destination>/<name>.md` including changelog with an entry for the initial version. See changelog [example](https://docs.airbyte.io/integrations/sources/stripe#changelog)
    - `docs/integrations/README.md`

### Airbyter

If this is a community PR, the Airbyte engineer reviewing this PR is responsible for the below items.

[...]

PR #27525, opened 2023-06-20
Add any glob to filter to ensure all steps are run in CI

bnchrch:
## What
If no globs were matched in the gradle.yaml filter step then the subsequent steps would not run

this led to missing required checks in this PR
https://github.com/airbytehq/airbyte/pull/27412

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -41,9 +41,9 @@ jobs:
       build: ${{ steps.filter.outputs.build }}
       cdk: ${{ steps.filter.outputs.cdk }}
       cli: ${{ steps.filter.outputs.cli }}
-      connectors: ${{ steps.filter.outputs.connectors }}
+      connectors_base: ${{ steps.filter.outputs.connectors_base }}
       db: ${{ steps.filter.outputs.db }}
-      frontend: ${{ steps.filter.outputs.frontend }}
+      any_change: ${{ steps.filter.outputs.any_change }}
     steps:
       - name: Checkout Airbyte
         uses: actions/checkout@v3
@@ -51,6 +51,7 @@ jobs:
         id: filter
         with:
           # Note, the following glob expression within a filters are ORs.
+          # Note, if no filters match, the steps are all skipped. This is why we have the any_change filter.
           filters: |
             build:
               - '.github/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle.yml
```diff
@@ -41,16 +41,17 @@ jobs:
...
-      frontend: ${{ steps.filter.outputs.frontend }}
+      any_change: ${{ steps.filter.outputs.any_change }}
     steps:
       - name: Checkout Airbyte
         uses: actions/checkout@v3
       - uses: dorny/paths-filter@v2
         id: filter
         with:
           # Note, the following glob expression within a filters are ORs.
+          # Note, if no filters match, the steps are all skipped. This is why we have the any_change filter.
```
erohmensing, line 54: As i understand it, the jobs are skipped before even checking the filters, which is the problem? 

Now, we'll strat the `start-connectors-base-build-runner` job, the `if` will eval to false so not run the steps, and then the required check for the job still returns green? 

if that's the case, maybe some more clarity in this comment will be helpful

.github/workflows/gradle.yml
```diff
@@ -41,16 +41,17 @@ jobs:
       build: ${{ steps.filter.outputs.build }}
       cdk: ${{ steps.filter.outputs.cdk }}
       cli: ${{ steps.filter.outputs.cli }}
-      connectors: ${{ steps.filter.outputs.connectors }}
+      connectors_base: ${{ steps.filter.outputs.connectors_base }}
       db: ${{ steps.filter.outputs.db }}
-      frontend: ${{ steps.filter.outputs.frontend }}
```
erohmensing, line 46: I take it this one is not used anywhere?

erohmensing (approved):

.github/workflows/gradle.yml
bnchrch: Nope! Relic of the monorepo split

.github/workflows/gradle.yml
bnchrch: Good call out. 

Looking at the output of the failed build it was essentially because the action overwrote the `modified_files` variable used by GHA. Which caused them to be skipped without being partially ran. 

Which as I understand had them not reporting a status check back to github

will update that comment

Issue #27516, opened 2023-06-20
Source Okta: `group_members` not syncing all records properly


eschrock:
### Connector Name

source-okta

### Connector Version

0.1.14

### What step the error happened?

During the sync

### Revelant information

There is a `group_members` stream in the Okta connector. It is attempting to use this API:

https://developer.okta.com/docs/reference/api/groups/#list-group-members

We have 170+ users, 400+ groups, but only 72 rows returned in this stream. There are two issues:

- The stream doesn't propagate the slice group ID. So there is no actual way to know what users are associated with what group. The 'group_id' field should be propagated into the child stream (similar to what I've seen done with Jira issues)
- I think (?) the fact that the primary_key is declared as 'id' means there's only ever one entry per user, even though users can be part of multiple groups. If there needs to a be a primary key at all, it would be the combination of (group_id, user_id).

### Contribute

- [ ] Yes, I want to contribute

eschrock:
The stream also declares a `cusor_key` of `id`, which doesn't make much sense. For an incremental sync, the only way to do it properly is to look at the 'lastMembershipUpdated' timestamp of the parent group. Using the id field of the group member record doesn't make much sense.

eschrock:
This doesn't really make sense as an incremental stream. There is no way to request only new records, and no way to filter to only new records. You could filter the parent stream to only be groups with recently updated memberships, but you'd still have no way of detecting deleted entries. This should really be a full stream only (similar to other child streams like GroupRoles). Here's a quick and dirty diff:

`diff --git a/airbyte-integrations/connectors/source-okta/source_okta/source.py b/airbyte-integrations/connectors/source-okta/source_okta/source.py
index b80f3dbbed5..a1616a09532 100644
--- a/airbyte-integrations/connectors/source-okta/source_okta/source.py
+++ b/airbyte-integrations/connectors/source-okta/source_okta/source.py
@@ -60,12 +60,20 @@ class OktaStream(HttpStream, ABC):
             **(next_page_token or {}),
         }
 
-    def parse_response(
-        self,
-        response: requests.Response,
-        **kwargs,
-    ) -> Iterable[Mapping]:
-        yield from response.json()
+    def parse_response(self, response: requests.Response, **kwargs) -> Iterable[Mapping]:
+        response_json = response.json()
+        if isinstance(response_json, list):
+            for record in response_json:
+                yield self.transform(record=record, **kwargs)
+        else:
+            yield self.transform(record=response_json, **kwargs)
[...]

PR #27533, opened 2023-06-20
🚨🚨 Source Okta: fix group members stream

eschrock:
## What

This addresses the issues described in [27516](https://github.com/airbytehq/airbyte/issues/27516). Within the Okta connector, there is a `group_members` stream. This stream suffers from a few bugs:

1. The stream is an incremental stream, but the configuration is non-sensical in that it declares `id` as the `cursor_field`. The [Okta API](https://developer.okta.com/docs/reference/api/groups/#group-member-operations) does not support any incremental querying of group members. While there is a notion on groups of a `lastMembershipUpdated` field, this is not particularly helpful in this instance. That can help reduce the number of groups we sync, but there's no way to represent deleted members. This should really be a full stream, similar to group role assignments and other sub-streams derived from users and groups.
2. The stream does not include the group ID. While this is present in the stream slice, not having it in the records produced makes it unusable, since there's no way to correlate members to groups.
3. The primary key is declared as `id` (the user ID). Wtih (2) fixed, the primary key needs. tobhe the tuple of (`groupId` and `id`), as a user can belong to multiple groups.

While addressing these issues, I also noticed that the `user_role_assignments` and `group_role_assignments` streams suffered from the same lack of group/user ID in order to enable correlation between records. They are, however full streams and don't suffer from the same incremental problems.

## How

The fix takes a page from the Jira source connector and creates a `transform` method that can optionally be overridden within streams to change the contents of records prior to being returned. This is used in the `GroupMembers`, `GroupRoleAssignments`, and `UserRoleAssignments` streams to include the parent ID, taken from the stream slice. This required more surgery than expected in the unit tests, as creating a named `stream_slice` parameter meant that previous unit tests had to suddenly pass this parameter along to the method. While I don't love the pattern the unit tests use, I coped the model of setting up the `inputs` variable and passing as kwargs, propagating that to more places that needed it.

In addition, this changes the `GroupMembers` stream from being incremental to full-only.

## Recommended reading order
1. `source.py`
2. `test_streams.py`

### 🚨🚨 Breaking Change

This is a breaking change due to removing incremental stream capability from the `group_members` stream. From the breaking changes guide, this is a schema change that will require existing connector configurations to switch from incremental mode to full refresh mode. Given that this stream was non-functional to begin with, and this is an alpha connector, this seems like an appropriate change.

I read through the referenced semantic versioning document, but it wasn't clear to me if the semantic versioning applied to alpha releases. I bumped the version from `0.1.14` to `0.1.15`, but if this should instead go to `1.0.0` (a major version bump), please let me know.

## Pre-merge Actions

I updated the unit tests to reflect the new semantics, and was able to run them successfully:

```
Results (1.40s):
      58 passed
```

It was not clear to me how to run the integration/acceptance tests. Judging from the expected output, they are designed to connect to some predetermined Okta test instance. I tried running:

```
python -m pytest -p connector_acceptance_test.plugin
```

But it failed because it was missing some configuration files (`secrets/config_api_token.json` and `secrets/config_oauth.json`). There is no information in the `README.md` about these files, and the README appears to contain boilerplate generated commentary, not something specific to this Okta plugin.  If there are instructions about how to properly run these tests, please let me know!

I also ran this within our analytics environment, and verified that it pulled over a complete set of user group members (previously we were getting 72 group members, now we gate 5.8k+). I did some manual inspection of group membership and they all looked correct.

### Airbyter

- Build is successful


sherifnada (approved):
@eschrock looks great! thanks for submitting this. I'll merge it shortly.

eschrock:
@sherifnada Thanks so much! Do let me know if I can do anything to get the integration/acceptance tests to work. Like I said, they appear to be expecting to connect to a known instance and check for known data, but I don't have any information about what that instance is, creds for it, configuration for the tests, etc. Happy to help given some direction.

Oh and also whether the version should be bumped to a major release (e.g. `1.0.0`). The breaking changes guide suggests that should be the case, but I don't know if that's typically done with Alpha connectors.

sherifnada:
@eschrock thanks for your contribution! this should be published shortly

PR #27546, opened 2023-06-21
Datetime inferrer: Improve detected formats

flash1293:
## What

The `%z` placeholder also allows offsets in the form of `01:00` - update the documentation as well as the datetime format inferrer logic accordingly.


maxi297 (approved):
🚢

PR #27559, opened 2023-06-21
octavia-cli: deserialize selected field to a SelectedFieldInfo instance

alafanechere:
## What
octavia-cli currently cannot work with the latest version of Airbyte as its API client is not up to date

## How
Update the openapi file and handle the conversions to retrieve and store configuration from Airbyte.


terencecho (approved):
lgtm. Don't forget to update the PR description!

PR #27598, opened 2023-06-22
Update test workflow: Update to use rev-parse

bnchrch:
## What
We can run test on master via workflow as their are too many heads that match `master`

this updates the action to use rev-parse


octavia-approvington:

octavia-approvington (approved):

PR #27603, opened 2023-06-22
connectors-ci: do not cache unpinned cdk

alafanechere:
## What
Multiple python connnectors are not pinning the CDK version they use:
For example in Zenloop's `setup.py`:
```python
MAIN_REQUIREMENTS = [
    "airbyte-cdk",
]
```

If no change is made to the `setup.py` file our connector build pipeline won't re-install `airbyte-cdk` because of caching.
Not explicitly pinning CDK version is a way to always get the latest CDK installed on each new release of a connector, but this caching behavior prevents it. 
We might introduce explicit pinning + auto cdk version bump in the `setup.py` later, but to be consistent with the previous `/publish` way of building connector we should install the latest version of the CDK on each build if the version is not pinned.

## How
1. Evaluate if the current connector is pinning the CDK by checking the `egg_info` requirement file.
2. If the connector is not pinning the CDK we check the latest released version according to Pypi and run `pip install airbyte-cdk==<latest version>` at the end of the build.
3. This is not totally bursting the cache as if no new cdk version is released the `pip install` step will be cached 👍


bnchrch (changes requested):
Hey @alafanechere I really like the spirit of this PR

But I dont think this is the approach we should take.

From a design stand point our pipelines
1. Should not concern themselves with an individual dependency (even if it is our own library)
2. Should not take on define use cases of other systems. In this instance it feels like dependency management should be the concern of the individual connectors package manager and dockerfile. Ideally in terms of a pinned version or better yet a lock file.

Some other solutions for us could be
1. Enforce that all our connectors set specific dependency versions
2. Recommend that all our connectors use lockfiles (or else add a cachebuster to your dockerfile)
3. Use something like [pip list --outdated](https://pip.pypa.io/en/stable/cli/pip_list/#cmdoption-o) to update all packages

Also calling on @airbytehq/connector-operations to make sure im not living in my own echo chamber.

maxi297 (changes requested):
I agree with what @bnchrch has said. I'll add that doing a `pip install airbyte_cdk==<latest version>` can create issues in terms of transitive dependencies since it does not take into account the other dependencies. For example, if the latest version of `airbyte_cdk` is incompatible with `requests==1.0.0` but that this version was installed before using the setup.py, overriding the version of `airbyte_cdk` using `pip install airbyte_cdk==<latest version>` will make the application crash

alafanechere:
@maxi297 @bnchrch closing this PR until we find a better and holistic approach at managing connector dependencies.

PR #27609, opened 2023-06-22
File-based CDK: implement AvailabilityStrategy.check_availability

### LINKED ISSUE #27014: File CDK: Check connection
During `check`, the following should be checked:
- We can list and read files from each stream.
- If files in the stream have an extension, it agrees with the file type of the stream config.
- If the user provided a schema, parse one record (or a chunk of data from the record) and validate that the records conform to the schema.
- If no files are identified in the stream, raise an exception.

#### Acceptance Criteria
- [ ] We can call the source-specific StreamReader's `open_file` and `list_matching_files` methods without error.
- [ ] If errors are encountered opening or listing files, an appropriate error is raised.
- [ ] We can check the extension type for files and determine whether it agrees with the selected file type. This is supported for both compressed and uncompressed files (e.g. files that end in ".csv" and ".csv.gzip").
- [ ] If the extensions don't agree, an appropriate error is raised.
- [ ] For each file type, we can parse the file and verify whether the record agrees with the user-input schema. 
- [ ] If we cannot parse the file, an appropriate error is raised.
- [ ] If the record doesn't agree,, an appropriate error is raised.
- [ ] If no files are in the stream, an appropriate error is raised.

clnoll:
## What
Adds handling for the `check` command to the file-based CDK.

Updates the `FileBasedAvailabilityStrategy`'s `check_availability` method to perform the following checks per stream:
- We can list files.
- There is at least one file in the stream.
- We can read at least one file.
- If files have an extension, it agrees with the file type of the stream config.
- If the user provided a schema, we can parse one record (or a chunk of data from the record) and validate that it agrees with their validation policy. Supporting this feature required some reorganization, which will ultimately allow connector developers to pass in their own schema validation policies.

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py`
2. `airbyte-cdk/python/airbyte_cdk/sources/file_based/default_file_based_availability_strategy.py`
3. Everything else.

Closes https://github.com/airbytehq/airbyte/issues/27014.


airbyte-cdk/python/airbyte_cdk/sources/file_based/remote_file.py
```diff
@@ -16,3 +16,9 @@ def __init__(self, uri: str, last_modified: datetime, file_type: Optional[str] =
         self.uri = uri
         self.last_modified = last_modified
         self.file_type = file_type
+
+    def extension_agrees_with_file_type(self) -> bool:
+        extensions = self.uri.split(".")[1:]
+        if not extensions:
+            return True
+        return any(self.file_type.lower() in e.lower() for e in extensions)
```
brianjlai, line 24: nit: Python generally recommends `casefold()` for the purposes of comparison:
https://stackoverflow.com/questions/45745661/lower-vs-casefold-in-string-matching-and-converting-to-lowercase

airbyte-cdk/python/airbyte_cdk/sources/file_based/exceptions.py
```diff
@@ -2,22 +2,58 @@
...
     pass
 
 
-class RecordParseError(Exception):
+class SchemaInferenceError(BaseFileBasedSourceError):
     pass
 
 
-class SchemaInferenceError(Exception):
+class ConnectionCheckError(BaseFileBasedSourceError):
```
brianjlai, line 54: nit: We might want to rename this CheckAvailabilityError (or something similar). ConnectionCheck is a pretty overloaded term and it feels very general to any part of the check process (availability, config validation). But how we're using it, we are just throwing these when we call `check_availability()`.

brianjlai (approved):
just a few small comments, the code appears to test all the scenarios laid out in the PR description so looks good to me!

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -63,8 +66,10 @@ def check_connection(self, logger: logging.Logger, config: Mapping[str, Any]) ->
                     stream_is_available,
                     reason,
                 ) = stream.availability_strategy.check_availability(stream, logger, self)
-            except Exception:
-                errors.append(f"Unable to connect to stream {stream} - {traceback.format_exc()}")
+            except Exception as exc:
+                errors.append(
+                    f"Unable to connect to stream {stream} - {''.join(traceback.format_exception(etype=type(exc), value=exc, tb=exc.__traceback__))}"
```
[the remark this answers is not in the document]
clnoll: Ah good call thanks - switched to `traceback.format_exc()`.

PR #27622, opened 2023-06-22
Remove Affected connector report

bnchrch:
## What
Remove the affected connector report as all affected connectors are tested and published automagically


tools/bin/ci_check_dependency.py
evantahler, line 1: Any way to check that the rest of the CI code 'compiles' and that no one else relies on this?

evantahler (approved):

tools/bin/ci_check_dependency.py
bnchrch: I searched for any use of `ci_check_dependency` and its functions. Nothing was returned.

Ill run our gradle commands against this branch to double check

tools/bin/ci_check_dependency.py
```diff
@@ -1,253 +0,0 @@
...
-COMMENT_TEMPLATE_PATH = ".github/comment_templates/connector_dependency_template.md"
-
-
-def download_and_parse_registry_json():
-    response = requests.get(CONNECTOR_REGISTRY_URL)
-    response.raise_for_status()
-    return response.json()
-
-
-def main():
```
bnchrch, line 44: @alafanechere Just want to double check my reading here. Affected connectors seems to only look at dependant connectors through gradle (i.e. only java)

So there never was logic to find related python connectors when a base or similar changed.

so we can remove safely

alafanechere: I would explain why we only see this message on java prs :D

alafanechere (approved):

PR #27648, opened 2023-06-23
Switch to new support portal

timroes:
## What

Changes the "Support" link in the doc page to point to the new support portal.


joeykmh (approved):

timroes:
Ignoring red CI, since it seems completely unrelated (this just changes a URL in the doc system).

PR #27655, opened 2023-06-23
fix docker auth for metadata tests and orchestrator deploy

erohmensing:
## What
* Handle issues where metadata ci wasn't running because we were missing docker login info

## How
* Pass in the expected inputs to the run-dagger-pipeline action in all places which weren't yet

### DIFF
--- a/airbyte-ci/connectors/metadata_service/lib/tests/test_commands.py
+++ b/airbyte-ci/connectors/metadata_service/lib/tests/test_commands.py
@@ -18,6 +18,7 @@ def test_valid_metadata_yaml_files(valid_metadata_yaml_files):
         assert result.exit_code == 0, f"Validation failed for {file_path} with error: {result.output}"
 
 
+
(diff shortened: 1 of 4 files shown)


airbyte-ci/connectors/metadata_service/lib/tests/test_commands.py
```diff
@@ -18,6 +18,7 @@ def test_valid_metadata_yaml_files(valid_metadata_yaml_files):
         assert result.exit_code == 0, f"Validation failed for {file_path} with error: {result.output}"
 
 
+
```
erohmensing, line 21: test change to get the metadata ci running- will be reverted before merge

alafanechere (approved):
Thank you for fixing it and sorry that I forgot to update this workflow!

airbyte-ci/connectors/metadata_service/lib/tests/test_commands.py
erohmensing: lol missed this but reverting it in another pr

PR #27667, opened 2023-06-23
Docs: Remove egress notice for destination connectors

alex-gron:
## What
* Accidentally added egress alert to destinations
* This should only exist for sources


supertopher (approved):
good catch.  thanks Alex

PR #27442, opened 2023-06-16
[Source-postgres] Set default cursor value for cdc mode

nguyenaiden:
## What
[26807](https://app.zenhub.com/workspaces/db--dw-source-connectors-6333360e0a41155061efbcbd/issues/gh/airbytehq/airbyte/26807)
To prepare for Destination v2, Cdc-enabled DB sources need to have a pre-defined cursor for normalization to operate properly. 
## How
During the discover phase, `source_defined_cursor` is set to `true`, and `_ab_cdc_lsn` is selected as the predefined cursor. NOTE: This value could change if we believe it's not unique. 
image

`AirbyteStateMessage:`
```
{
  "cdc": true,
  "cdc_state": {
    "state": {
      "[\"postgres\",{\"server\":\"postgres\"}]": "{\"transaction_id\":null,\"lsn\":840386496,\"txId\":1895971,\"ts_usec\":1686867930174234}"
    }
  },
  "streams": [
    {
      "stream_name": "blah",
      "stream_namespace": "public",
      "cursor_field": [
        "_ab_cdc_lsn"
      ]
    },
    {
      "stream_name": "jsonb_test_norm",
      "stream_namespace": "public",
      "cursor_field": [
        "_ab_cdc_lsn"
      ]
    }
  ]
}
```

## Recommended reading order
1. `PostgresCatalogHelper.java`
2. `PostgresSource.java`
3.  All the test files

## 🚨 User Impact 🚨
New CDC syncs that have refreshed their source schema will see this cursor field be chosen. Customers who reset their data will continue to see old behavior until their source goes through the `discover()` phase again.

PR #27670, opened 2023-06-23
Normalization integration tests: set explicit cursor on cdc streams

edgao:
DB sources intend to set a cursor on their discovered catalogs. Update legacy normalization tests to reflect this and verify that we behave nicely.

I've already run some manual syncs (postgres->postgres, where source-postgres was built from https://github.com/airbytehq/airbyte/pull/27442) and everything looks good. This PR is just to run normalization tests on all destinations.

### DIFF
--- a/airbyte-integrations/bases/base-normalization/integration_tests/resources/test_simple_streams/data_input/catalog.json
+++ b/airbyte-integrations/bases/base-normalization/integration_tests/resources/test_simple_streams/data_input/catalog.json
@@ -124,10 +124,10 @@
         },
         "supported_sync_modes": ["full_refresh", "incremental"],
         "source_defined_cursor": true,
-        "default_cursor_field": []
+        "default_cursor_field": ["_ab_cdc_updated_at"]
       },
       "sync_mode": "incremental",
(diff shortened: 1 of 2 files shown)


airbyte-integrations/bases/base-normalization/integration_tests/resources/test_simple_streams/data_input/catalog.json
```diff
@@ -124,10 +124,10 @@
         },
         "supported_sync_modes": ["full_refresh", "incremental"],
         "source_defined_cursor": true,
-        "default_cursor_field": []
+        "default_cursor_field": ["_ab_cdc_updated_at"]
```
edgao, line 127: this stream is weird and doesn't have an actual lsn field. Use updated_at.

(I _think_ this stream isn't even used as-is; it has no records in messages.txt)

evantahler: 💪

evantahler (approved):

PR #27685, opened 2023-06-24
Bnchrch/cat/backwards fail removed prop

### LINKED ISSUE #20754: SAT: make property removal in a stream fail backward compabitility testing
We considered until today that property removal is not a breaking change as Normalization is not destructive and will not delete the removed column. 
But from a record consistency standpoint removing a field leads to record with "fewer" data.

- Detect field removal in SAT discovery backward compatibility test.
- Make the test fail if a field was removed
- 
An new assertion needs to be added to this [function](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/source-acceptance-test/source_acceptance_test/utils/backward_compatibility.py#L226).

bnchrch:
## What
Updates our backward compatibility check to test for removed properties and fail

closes #20754

### DIFF
--- a/airbyte-integrations/bases/connector-acceptance-test/CHANGELOG.md
+++ b/airbyte-integrations/bases/connector-acceptance-test/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.11.0
+Add backward_compatibility.check_if_field_removed test to check if a field has been removed from the catalog.
+
 ## 0.10.8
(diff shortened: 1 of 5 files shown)


alafanechere (approved):
👍 Please bump the CAT version, update the changelog, and publish with `/legacy-publish`. Notifying this change to connector developers would also be nice.
Shall we also document what's considered backward incompatible [here](https://docs.airbyte.com/connector-development/testing-connectors/connector-acceptance-tests-reference/)?

bnchrch:
/legacy-publish connector=bases/connector-acceptance-test auto-bump-version=false

PR #27706, opened 2023-06-26
publish pipeline: mention connector ops instead of channel

alafanechere:
## What
In case of pipeline failure we'd like to mention @dev-connector-ops instead of @channel


pedroslopez (approved):
No idea if the syntax is right, but LGTM!

PR #27710, opened 2023-06-26
Source Stripe: allow specifying API Version to use

mdibaiee:
## What
- Allow users to specify API Version to be used with Source Stripe, since different Stripe API Versions behave differently, and some users using specific features can only access their data if they use the appropriate API Version. For example, products and promotion_codes with `type: service` do not show up for API Versions before 2018-02-05.

## How
- Add a new config parameter `api_version` with default `2017-08-15` (which is the version used by default if the header is omitted. So right now the connector is using this API Version)

## Recommended reading order
1. `spec.yaml`
2. `source.py`
3. `streams.py`

## 🚨 User Impact 🚨

No, there should be no breaking change as a result of this pull-request. The change is backward-compatible since the default API version is the same as the API version the connector uses without this header.

## Pre-merge Actions

<strong>Updating a connector</strong>

### Airbyter

- Build is successful

### DIFF
--- a/airbyte-integrations/connectors/source-stripe/Dockerfile
+++ b/airbyte-integrations/connectors/source-stripe/Dockerfile
@@ -14,5 +14,5 @@ ENTRYPOINT ["python", "/airbyte/integration_code/main.py"]
 
 
 
-LABEL io.airbyte.version=3.10.0
+LABEL io.airbyte.version=3.11.0
 LABEL io.airbyte.name=airbyte/source-stripe
--- a/airbyte-integrations/connectors/source-stripe/source_stripe/source.py
+++ b/airbyte-integrations/connectors/source-stripe/source_stripe/source.py
@@ -75,6 +75,7 @@ class SourceStripe(AbstractSource):
             "account_id": config["account_id"],
             "start_date": start_date,
             "slice_range": config.get("slice_range"),
+            "version": config.get("api_version", "2017-08-15"),
         }
         incremental_args = {**args, "lookback_window_days": config.get("lookback_window_days")}
         return [
--- a/airbyte-integrations/connectors/source-stripe/source_stripe/spec.yaml
+++ b/airbyte-integrations/connectors/source-stripe/source_stripe/spec.yaml
@@ -56,3 +56,9 @@ connectionSpecification:
         the less requests will be made and faster the sync will be. On the other hand, the more seldom
         the state is persisted.
       order: 4
+    api_version:
+      type: string
+      title: Stripe API Version to use
+      default: "2017-08-15"
+      description: >-
+        The API Version to use when sending requests to Stripe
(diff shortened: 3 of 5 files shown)


sherifnada (commented):
@mdibaiee good to see you around here! 👋🏼 do you know if there are other implications of using different API versions? are the schemas of various records different or the behavior different in other ways? Parametrizing the API version feels like a big deal so I want to avoid potholes where we can

mdibaiee:
@sherifnada there can _potentially_ be all sorts of differences between different API Versions, although I don't have enough familiarity with Stripe to say _practically_ what sort of changes are more common. It is possible that stream schemas are also different between different versions.

The ideal solution is to use the latest API Version and keep the connector up-to-date with the changes of the latest API version (note of warning: Stripe updates their API pretty frequently, multiple times per week! [API Changelog](https://stripe.com/docs/changelog)), but I don't know how long that would take to do here, hence this escape hatch to allow fiddling with the API Version to find a version that works for specific use cases.

mdibaiee:
@sherifnada one potential option is having this parameter hidden, I think the way to do that is `airbyte_hidden: true`, in that case customers are not presented with this option by default but they can choose to set it as an "advanced" parameter if that makes sense.

sherifnada:
@mdibaiee looking at the Stripe docs, I think [we may have a bigger issue with Stripe versioning](https://github.com/airbytehq/airbyte/issues/27772) in that every consumer might be hitting a different Stripe API version when running the connector. My feeling is that the correct way to handle this issue is to use a consistent API version in the connector (preferably the latest version, upgraded periodically e.g once a quarter). Generally I think API version parametrization is going to be tricky to maintain. 

To this end, there may be a workaround that solves your immediate problem: it seems like you can solve the immediate user painpoint by having them set their account's default API version [as described here](https://stripe.com/docs/upgrades#how-can-i-upgrade-my-api). Do you think this is a viable option?

ysusuk:
Hello folks,

since it is reasonable to select version when calling Stripe API, it sounds to be as well reasonable to add such an option to connector as well (Source, in particular).

From experience I know that not many companies have an option to quickly update default API version for their Stripe customer accounts, because it means often require update of all downstream and/or dependent services.

sherifnada:
hi @ysusuk like mentioned above, I think parametrizing the Stripe API version means the connector needs to adequately handle every possible API version which is infeasible. We should pin to one or maybe two major supported versions but I don't think it's sustainable to do more than that. 

I think we'd accept a PR which fixed the API version used in the connector to the latest currently available version (including any schema/logic updates). Otherwise we'll make this change ourselves soon. 

@mdibaiee I'm closing this PR -- please re-open if you'd like to make the change requested above

ysusuk:
@sherifnada I understand your fear (or rather statement)... But I don't think there is a need to support all possible versions. There is a need to give users an option to specify API version if they are sure that they can handle any possible issues (know how it works behind the scene). With a reasonable warning.

In that way users will be pleased and in the same time project will not lost any of its statements of being super power for officially supported API version (last 1 or 2).

sherifnada:
@mdibaiee @ysusuk we just merged a PR to upgrade the Stripe connector to use the latest version, `2022-11-15`. Does that solve the concerns here?

PR #27719, opened 2023-06-26
Delete docs for old sentry integration

pedroslopez:
## What

This method of integrating with sentry is no longer supported. We should remove it from the docs to avoid misleading users.

## How

Delete the doc!

## 🚨 User Impact 🚨

None - this already hasn't been working for a while and was replaced with a AirbyteTraceMessage + platform-based error reporting.


bnchrch (approved):

erohmensing:
@bnchrch too speedy!!

PR #27749, opened 2023-06-27
connectors-ci: change report bucket

alafanechere:
## What
The `airbyte-ci-reports` bucket was not hosted in multi region mode. This prevented ingestion of its data as an external bigquery table.

## How
* We manually created a new bucket `airbyte-ci-reports-multi` to which we transferred the old data and set the same permissions.


octavia-approvington:

octavia-approvington (approved):

PR #27760, opened 2023-06-27
Log connector dependencies

### LINKED ISSUE #27759: Log connector dependencies in airbyte-ci
## Overview

We should log connectors published due to a dependency on a modified connector

bnchrch:
## What
Updates the airbyte ci modified connectors util to log when a connector was added due to upstream dependency

closes #27759


alafanechere (approved):
Nice readability boost, thank you! Could you please make a dummy change on `source-jdbc` and check if test on `source-postgres`are triggered? Just to make sure the change did not break anything.

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -246,21 +276,18 @@ def get_modified_connectors(modified_files: Set[Union[str, Path]]) -> dict:
     Or to tests all jdbc connectors when a change is made to source-jdbc or base-java.
     We'll consider extending the dependency resolution to Python connectors once we confirm that it's needed and feasible in term of scale.
     """
+    all_connector_dependencies = [
+        (connector, connector.get_local_dependencies_paths())
```
alafanechere, line 280: I think the conflict you face is because I renamed `get_local_dependencies_paths` to `get_local_dependency_paths`.

airbyte-integrations/connectors/source-jdbc/src/main/java/io/airbyte/integrations/source/jdbc/AbstractJdbcSource.java
```diff
@@ -141,7 +141,7 @@ protected List<CheckedConsumer<JdbcDatabase, Exception>> getCheckOperations(fina
   }
 
   /**
-   * Aggregate list of @param entries of StreamName and PrimaryKey and
+   * Aggregate list of @param entries of StreamName and PrimaryKey
```
bnchrch, line 144:
```suggestion
   * Aggregate list of @param entries of StreamName and PrimaryKey and
```

PR #27766, opened 2023-06-27
Metadata: Persist Registry entries

### LINKED ISSUE #26878: Cache RegistryEntries for OSS and Cloud connector versions
# Problem
For any OLD connector version, how do we get the registry entry that was in the catalog but no longer is?

# Solution
Lets update our registry generator to cache connector entries

For example if you are looking for the last registry entry for destination-bigquery:1.2.19 on cloud you would go to

https://connectors.airbyte.com/files/metadata/airbyte/destination-bigquery/1.2.19/cloud.json

----

# OLD CONVERSATION BELOW

---

## The Current System
1. Metadata files only allow for registry specific overrides
2. The Connector Registries can only have one connector version at a time. (OSS can have `source-s3:1.2.3` and cloud can have `source-s3:2.0.0`
3. Platform can only have one default version of a Actor Definition at a time (This directly relates to the version in the registry)
4. The Platform now allows for individual Actors to reference other Actor Definition Versions other that the default. `source-s3:eds-prerelease-123`
5. We want to extend  Actor Definition Versions so that we can allow users remain on old versions for a set window of time.
6. Populating Actor Definition Versions, at this moment, are an awkward combination of current metadata, version specific spec cache, and whatever data was ingested on that version if it ever originally appeared in the Registry.

## Questions for the team
The 6th point is the awkward turtle and we need to answer the question(s):
1. If we move a connector on [oss|cloud] from v1 to v2 with a 6 month deprecation window, how do we ensure we freeze v1s metadata related settings in time
2. If we add a connector + version to [oss|cloud] that the platform has never seen, how do we populate its metadata related values?

## Related Conversations
Normalization Overrides: https://airbytehq-team.slack.com/archives/C03VDJ4FMJB/p1685552842556829
Rollbacks: https://airbytehq-team.slack.com/archives/C03VDJ4FMJB/p1685567791734259

bnchrch:
## What
Updates the metadata service so
1. Metadata files are transformed into registry entries first
2. Those entries are written to `<metadata_bucket>/metadata/connector_folder/oss.json` `<metadata_bucket>/metadata/connector_folder/cloud.json`
3. Deletes any registry entries in the case of a disabled connector
4. Updates our registry to listen for and be generated from these new files
5. Adds a partiition sensor and asset to allow for metadata failures to not block new connector updats

closes #26878

### DIFF
--- a/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
+++ b/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
@@ -17,7 +17,7 @@ from orchestrator.assets import (
     metadata,
 )
 
-from orchestrator.jobs.registry import generate_registry_reports, generate_registry, generate_registry_entry
+from orchestrator.jobs.registry import generate_registry_reports,  generate_oss_registry, generate_cloud_registry, generate_registry_entry
 from orchestrator.jobs.connector_test_report import generate_nightly_reports, generate_connector_test_summary_reports
 from orchestrator.sensors.registry import registry_updated_sensor
 from orchestrator.sensors.gcs import new_gcs_blobs_sensor, new_gcs_blobs_partition_sensor
@@ -72,6 +72,12 @@ RESOURCES = {
     "latest_metadata_file_blobs": gcs_directory_blobs.configured(
         {"gcs_bucket": {"env": "METADATA_BUCKET"}, "prefix": METADATA_FOLDER, "match_regex": f".*latest/{METADATA_FILE_NAME}$"}
     ),
+    "latest_cloud_registry_entries_file_blobs": gcs_directory_blobs.configured(
+        {"gcs_bucket": {"env": "METADATA_BUCKET"}, "prefix": METADATA_FOLDER, "match_regex": f".*latest/cloud.json$"}
+    ),
+    "latest_oss_registry_entries_file_blobs": gcs_directory_blobs.configured(
+        {"gcs_bucket": {"env": "METADATA_BUCKET"}, "prefix": METADATA_FOLDER, "match_regex": f".*latest/oss.json$"}
+    ),
     "latest_oss_registry_gcs_blob": gcs_file_blob.configured(
         {"gcs_bucket": {"env": "METADATA_BUCKET"}, "prefix": REGISTRIES_FOLDER, "gcs_filename": "oss_registry.json"}
     ),
@@ -96,9 +102,15 @@ RESOURCES = {
 SENSORS = [
     registry_updated_sensor(job=generate_registry_reports, resources_def=RESOURCES),
     new_gcs_blobs_sensor(
-        job=generate_registry,
+        job=generate_oss_registry,
         resources_def=RESOURCES,
-        gcs_blobs_resource_key="latest_metadata_file_blobs",
+        gcs_blobs_resource_key="latest_oss_registry_entries_file_blobs",
+        interval=30,
+    ),
+    new_gcs_blobs_sensor(
+        job=generate_cloud_registry,
+        resources_def=RESOURCES,
+        gcs_blobs_resource_key="latest_cloud_registry_entries_file_blobs",
         interval=30,
     ),
     new_gcs_blobs_sensor(
@@ -118,7 +130,7 @@ SENSORS = [
 
 SCHEDULES = [ScheduleDefinition(job=generate_connector_test_summary_reports, cron_schedule="@hourly")]
 
-JOBS = [generate_registry_reports, generate_registry]
+JOBS = [generate_registry_reports, generate_oss_registry, generate_cloud_registry]
 
 """
 START HERE
--- a/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
+++ b/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
@@ -17,7 +17,7 @@ from orchestrator.assets import (
     metadata,
 )
 
-from orchestrator.jobs.registry import generate_registry_reports, generate_registry
+from orchestrator.jobs.registry import generate_registry_reports, generate_registry, generate_registry_entry
 from orchestrator.jobs.connector_test_report import generate_nightly_reports, generate_connector_test_summary_reports
 from orchestrator.sensors.registry import registry_updated_sensor
 from orchestrator.sensors.gcs import new_gcs_blobs_sensor, new_gcs_blobs_partition_sensor
@@ -107,7 +107,7 @@ SENSORS = [
         interval=(1 * 60 * 60),
     ),
     new_gcs_blobs_partition_sensor(
-        job=generate_registry,
+        job=generate_registry_entry,
         resources_def=RESOURCES,
         partitions_def=registry_entry.metadata_partitions_def,
         gcs_blobs_resource_key="latest_metadata_file_blobs",
--- a/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
+++ b/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
@@ -69,6 +69,9 @@ RESOURCES = {
     "registry_directory_manager": gcs_file_manager.configured({"gcs_bucket": {"env": "METADATA_BUCKET"}, "prefix": REGISTRIES_FOLDER}),
     "registry_report_directory_manager": gcs_file_manager.configured({"gcs_bucket": {"env": "METADATA_BUCKET"}, "prefix": REPORT_FOLDER}),
     "root_metadata_directory_manager": gcs_file_manager.configured({"gcs_bucket": {"env": "METADATA_BUCKET"}, "prefix": ""}),
+    "all_metadata_file_blobs": gcs_directory_blobs.configured(
+        {"gcs_bucket": {"env": "METADATA_BUCKET"}, "prefix": METADATA_FOLDER, "match_regex": f".*/{METADATA_FILE_NAME}$"}
+    ),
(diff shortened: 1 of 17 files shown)


airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/sensors/gcs.py
```diff
@@ -44,3 +53,66 @@ def new_gcs_blobs_sensor_definition(context: SensorEvaluationContext):
             return RunRequest(run_key=run_key)
 
     return new_gcs_blobs_sensor_definition
+
+def new_gcs_blobs_partition_sensor(
```
bnchrch, line 57: Added: A new sensor that can generate partition based jobs.

Partition jobs are how we say given an array of metadata file names, generate registry entry assets for each

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_entry.py
bnchrch, line 1: This is the new-ish concept in the code base

Registry Entry.

Previously this data model existed in registry.py 

However now we are writing it to oss.json and cloud.json files

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
```diff
@@ -93,9 +105,15 @@
 SENSORS = [
```
pedroslopez, line 105: Just to check my understanding of these updated sensors:
1. when any metadata file changes, run `generate_registry_entry` to generate the registry entry for it
2. when any "latest" cloud registry entries change, run `generate_cloud_registry` to generate the registry for cloud
3. when any "latest" cloud registry entries change, run `generate_cloud_registry` to generate the registry for OSS

That right? One thing I'm not sure I understand is where the partitions come in and how they work. Is this to run the registry generation for both cloud/oss registries when metadata changes?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_entry.py
```diff
@@ -0,0 +1,380 @@
...
+        Output[ConnectorRegistryV0]: The registry.
+    """
+    raw_entry_dict = metadata_to_registry_entry(metadata_entry, registry_name)
+    registry_entry_with_spec = apply_spec_to_registry_entry(raw_entry_dict, cached_specs)
+
+    _, ConnectorModel = get_connector_type_from_registry_entry(registry_entry_with_spec)
+
+    registry_model = ConnectorModel.parse_obj(registry_entry_with_spec)
+
+    file_handle = persist_registry_entry_to_json(registry_model, registry_name, metadata_entry, metadata_directory_manager)
```
pedroslopez, line 240: Does this persist to the versioned folder, or to latest? Or both?
Not sure I undertand how/when we save to e.g. `/1.0.0/oss.json` vs `/latest/oss.json`

pedroslopez (commented):
nice! overall looks good to me, though I'm missing some overall understanding of the system. Asked a few clarifying questions but 👍🏼

airbyte-ci/connectors/metadata_service/orchestrator/tests/test_registry.py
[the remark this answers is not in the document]
pedroslopez: If everything is properly tested then maybe tests not changing here means these changes are good overall and dont break the previous functionality - so I guess that's a good thing! Still think there may be opportunity for some added tests on the new code though.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_entry.py
pedroslopez: oh! maybe we're only writing to the versioned one actually, since "latest" would be in the full registry?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_entry.py
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    metadata_delete = {
+        f"delete_{registry_name}": MetadataValue.url(registry_url) for registry_name, registry_url in deleted_registry_entries.items()
+    }
+
+    metadata = {
+        **metadata_persist,
+        **metadata_delete,
+    }
```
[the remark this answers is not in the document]
bnchrch: General metadata 😅 

Very confusing. Will update the naming to be more clear that its operationally Dagster related

airbyte-ci/connectors/metadata_service/orchestrator/tests/test_registry.py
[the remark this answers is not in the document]
bnchrch: Functionally the core logic that transforms a metadata definition to a registry entry is the exact same

But we can always add more test :)

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
bnchrch: Exactly right 

and as far as the partitions go

Partitions is the mechanism that lets us dynamically create an `asset` for a dynamic input.

In this case the dynamic input is a file (thats referenced by etag) so that we dont process a metadata file in to its registry entry files more than once as etag is unique to the file.

I will admit the way you define a partition, a partition job, a partion sensor and a partition asset(s) in dagger is not as straightforward as you would expect.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_entry.py
bnchrch: It writes `oss.json` and/or `cloud.json` to the same folder that `metadata.yaml` was found.

Because we run the `registry_entry` asset against `all_metadata_files` then we write a registry entry to anywhere a metadata file exists.

This includes latest and any versions.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/sensors/gcs.py
```diff
@@ -44,3 +53,53 @@ def new_gcs_blobs_sensor_definition(context: SensorEvaluationContext):
...
+            ]
+            context.log.info(f"New etags found: {new_etags_found}")
+
+            if not new_etags_found:
+                return SkipReason(f"No new {gcs_blobs_resource_key} in GCS bucket")
+
+            # if there are more than the MAX_RUN_REQUEST, we need to split them into multiple runs
+            if len(new_etags_found) > MAX_RUN_REQUEST:
+                new_etags_found = new_etags_found[:MAX_RUN_REQUEST]
+                context.log.info(f"Only processing first {MAX_RUN_REQUEST} new blobs: {new_etags_found}")
```
erohmensing, line 98: What happens with the others? Do they get picked up at the next sensor interval?

erohmensing (approved):
I'm also new to the orchestrator code, but with both of us taking a look, I feel pretty good a about giving this a 👍🏻. That being said, getting a peek from Augustin who reviewed all this code before might be helpful - I'll leave that up to you though

I am interested what the general development/test workflow of figuring this stuff out in dagster is - would be happy to pair sometime on it! I think that would make me feel more confident re: knowing stuff will work when it goes in (or how to debug it otherwise)

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/sensors/gcs.py
bnchrch: Exactly!

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_entry.py
```diff
@@ -0,0 +1,380 @@
...
+
+    persisted_registry_entries = {
+        registry_name: generate_and_persist_registry_entry(metadata_entry, cached_specs, root_metadata_directory_manager, registry_name)
+        for registry_name in enabled_registries
+    }
+
+    deleted_registry_entries = {
+        registry_name: delete_registry_entry(registry_name, metadata_entry, root_metadata_directory_manager)
+        for registry_name in disabled_registries
+    }
```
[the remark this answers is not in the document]
bnchrch: Exactly.

So the sensor for generating the registry does the following:
1. Every 30s get all gcs files that match `**/latest/[oss|cloud].json`
2. Compute a cursor value from the aggregate of all etag values
3. If that cursor has changed (files added, updated, or deleted), call the generate registry jobs

And the job for generating the registry starts by
1. Getting all gcs files that match `**/latest/[oss|cloud].json`

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
```diff
@@ -104,11 +122,18 @@
...
+        partitions_def=registry_entry.metadata_partitions_def,
+        gcs_blobs_resource_key="all_metadata_file_blobs",
+        interval=30,
+    ),
 ]
 
 SCHEDULES = [ScheduleDefinition(job=generate_connector_test_summary_reports, cron_schedule="@hourly")]
 
-JOBS = [generate_registry_reports, generate_registry]
+JOBS = [generate_registry_reports, generate_oss_registry, generate_cloud_registry]
```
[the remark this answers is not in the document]
bnchrch: Ella great catch! So I think adding these jobs to a sensor also adds them as a clickable job.

I never noticed that `generate_registry_entry` report wasnt explicitly here, or `generate_nightly_reports` for that matter.

Ill add them :)

bnchrch:
@pedroslopez Requested your re-review as I added a few more test cases to round this out :)

pedroslopez (approved):
:shipit:

Issue #27022, opened 2023-06-05
File CDK: new file-based connector


clnoll:

sherifnada:
Three sources that would be useful from a product POV: 
1. GCS
2. SFTP
3. Azure Blob Storage

clnoll:
Grooming notes:
- Check in with Tobias about putting publicly available data on S3.

clnoll:
Closing this in favor of prioritizing the S3 connector.

PR #27816, opened 2023-06-28
File cdk validation policies

clnoll:
## What
Validates records against the schema provided in the configured catalog, and handles them according to a validation policy configured by the user.

The default validation policies that we are offering include:

- "emit_record": Do not validate the record against the schema, and simply emit it whether or not it conforms. The platform may filter out columns that are not in the schema, but this behavior is undefined.
- "skip_record": Skip any records that don't conform to the schema. This will result in records being skipped, so we log a message for each file where we skipped records.
- "wait_for_discover":  Stop syncing records for the stream with the nonconforming record, in the hopes that the new schema produced by the discovery cron will allow the sync to continue. 

## How
This change requires us to have the schema from the configured catalog available to the stream; it was previously not available.

The neatest way to do this would be to modify `read_records` to take the catalog's schema, however updating the interface to do this would require updating all connectors that implement the interface.

My second choice was to create a getter & setter on `Stream`. The catalog's stream is set by the source during `read`, and accessed by the file-based stream. This results in a sort of ugly/hacky situation where we are setting an attribute `_catalog_schema` outside of `__init__`. The alternative would be to add an `__init__` to `Stream`, but if any of the classes that implement it have their own `__init__`, they'd need to be updated to call `__init__` on super.

Other options -
- Have the source reread the catalog
  - Modify `FileBasedSource` to take a path to the catalog
  - During `__init__`, call `self.read_catalog()`
  - Create a mapping of stream name to schema from the catalog
  - Use the mapping to pass the schema into the `DefaultFileBasedStream` object on `__init__`.

- Use a new entrypoint pattern:
  - Instead of calling `launch`, create a new launch function, which will parse the args and create the config, catalog, and state objects
  - Instead of creating an `AirbyteEntrypoint`, creates a new `AirbyteFileBasedEntrypoint` which inherits from `AirbyteEntrypoint` but implements its own `run` method.
```
class AirbyteFileBasedEntrypoint(AirbyteEntrypoint):

    def run_file_based(self, cmd: Optional[str], debug: Optional[bool], raw_config, catalog, state) -> Iterable[str]:
        
        # 
        # NOTE: most of this is copied from AirbyteEntrypoint.run, with very minor modifications so that it takes the config, catalog, and state
        #
        
        if not cmd:
            raise Exception("No command passed")

        if debug:
            self.logger.setLevel(logging.DEBUG)
            self.logger.debug("Debug logs enabled")
        else:
            self.logger.setLevel(logging.INFO)

        source_spec: ConnectorSpecification = self.source.spec(self.logger)
        try:
            with tempfile.TemporaryDirectory() as temp_dir:
                if cmd == "spec":
                    message = AirbyteMessage(type=Type.SPEC, spec=source_spec)
                    yield from [
                        self.airbyte_message_to_string(queued_message) for queued_message in self._emit_queued_messages(self.source)
                    ]
                    yield self.airbyte_message_to_string(message)
                else:
                    config = self.source.configure(raw_config, temp_dir)

                    if cmd == "check":
                        yield from map(AirbyteEntrypoint.airbyte_message_to_string, self.check(source_spec, config))
                    elif cmd == "discover":
                        yield from map(AirbyteEntrypoint.airbyte_message_to_string, self.discover(source_spec, config))
                    elif cmd == "read":
                        config_catalog = self.source.read_catalog(catalog)
                        state = self.source.read_state(state)

                        yield from map(AirbyteEntrypoint.airbyte_message_to_string, self.read(source_spec, config, config_catalog, state))
                    else:
                        raise Exception("Unexpected command " + cmd)
        finally:
            yield from [self.airbyte_message_to_string(queued_message) for queued_message in self._emit_queued_messages(self.source)]


def launch_file_based(source: Source, args: List[str]):
    source_entrypoint = AirbyteFileBasedEntrypoint(source)
    parsed_args = source_entrypoint.parse_args(args)
    raw_config = source.read_config(parsed_args.config)
    config_catalog = source.read_catalog(parsed_args.catalog)
    state = source.read_state(parsed_args.state)
    source_entrypoint.run_file_based(parsed_args.cmd, parsed_args.debug, raw_config, config_catalog, state)
```

I'm open to opinions!

## Recommended reading order
1. `schema_helpers.py`
2. `schema_validation_policies/default_schema_validation_policies.py`
3. `stream/default_file_based_stream.py`
4. `stream/file_based_stream_config.py`
5. `sources/streams/core.py` & `sources/abstract_source.py`
6. Everything else.


airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -110,6 +110,14 @@ def read(
                         f"The requested stream {configured_stream.stream.name} was not found in the source."
                         f" Available streams: {stream_instances.keys()}"
                     )
+
+                stream_is_available, error = stream_instance.check_availability(logger, self)
+                if not stream_is_available:
+                    logger.warning(f"Skipped syncing stream '{stream_instance.name}' because it was unavailable. Error: {error}")
+                    continue
+
+                stream_instance.catalog_schema = configured_stream.stream.json_schema
```
maxi297, line 119: This feels like an initialization problem and it feels like it's not the first time we struggle with this. The [proposal for data feed](https://docs.google.com/document/d/1UpYo_kPO04KGzDOB7KxBTXUTEzmBB_sBO7X9KddXRBk/edit#heading=h.nav4gq872r1j) was expecting to have the state when we initialize the source but it did not. The consequence was that we had to propagate state everywhere on the `DeclarativeStream.state` setter but as we didn't know all the underlying components, it ended up creating a lot of `if self.cursor and isinstance(self.cursor, <whatever>)`. I fear that because we don't instantiate our component fully at the start of a run, we end-up in this situation where:
* we have multiple place where we manage the initialization
* the internal state of our classes are very hard to understand because the lifecycle isn't just "init everything" but is "init x + init y + ..."

I'm wondering if we can start pushing these up as much as possible close to the initialization of the source. I feel like there are a couple of challenges to do this:
* Not all CDK command provide the same information and therefore fields like `catalog` are not always available
* The source is created before being passed to the entrypoint which finds out some parameters that we would like for the initialization (like `config_catalog = self.source.read_catalog(parsed_args.catalog)`)

I'll think about it. I think the least we could do is put all those initialization at the same place as much as possible. The one I had to deal with was the `state` which is initialize [here](https://github.com/airbytehq/airbyte/blob/e8ddda970905ae338377d2b2a7ed5fa9ffa57ed4/airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py#L239-L242). I'm not sure this is the best place as it is scope to the stream but at least it gives us another case to consider

airbyte-cdk/python/airbyte_cdk/sources/streams/core.py
```diff
@@ -267,3 +267,14 @@ def _wrapped_primary_key(keys: Optional[Union[str, List[str], List[List[str]]]])
...
+
+    @property
+    def catalog_schema(self):
+        if hasattr(self, "_catalog_schema"):
+            return self._catalog_schema
+        return None
+
+    # noinspection PyAttributeOutsideInit
+    @catalog_schema.setter
+    def catalog_schema(self, schema: Dict[str, Any]):
```
maxi297, line 279: Should this be an object instead of a primitive type? Can we re-use `ConfiguredAirbyteCatalog`?

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -65,17 +72,49 @@ def read_records_from_slice(self, stream_slice: StreamSlice) -> Iterable[Mapping
...
+                yield AirbyteMessage(
+                    type=Type.LOG,
+                    log=AirbyteLogMessage(
+                        level=Level.ERROR,
+                        message=f"{FileBasedSourceError.ERROR_PARSING_RECORD.value} stream={self.name} file={file.uri} line_no={line_no}",
+                        stack_trace="\n".join(traceback.format_exception(etype=type(exc), value=exc, tb=exc.__traceback__)),
+                    ),
+                )
+
+            else:
```
maxi297, line 109: We are only printing the `n_skipped` AirbyteMessage if there are no errors but could there be errors and skipped records? If that's the case, do we want to send both AirbyteMessages?

maxi297 (commented):
I need to recheck this one with a cool head tomorrow to see if I can propose something interesting (right now, I was only able to raise and issue without being constructive about it)

airbyte-cdk/python/airbyte_cdk/sources/abstract_source.py
```diff
@@ -110,6 +110,14 @@ def read(
                         f"The requested stream {configured_stream.stream.name} was not found in the source."
                         f" Available streams: {stream_instances.keys()}"
                     )
+
+                stream_is_available, error = stream_instance.check_availability(logger, self)
```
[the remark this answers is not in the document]
clnoll: Strange, I don't know where these lines came from - maybe from fixing a merge conflict. The only new line in this file should be 119. Fixing this and doing a pass over the rest of the files to make sure this is the only place with unexpected code.

airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py
```diff
@@ -74,7 +90,27 @@ def conforms_to_schema(record: Mapping[str, Any], schema: Mapping[str, str]) ->
...
+
+    if not record_columns.issubset(schema_columns):
+        return False
+
+    for column, definition in schema.get("properties", {}).items():
+        expected_types = definition.get("type", [])
+        value = record.get(column)
+
+        if value is not None:
+            if "object" in expected_types if isinstance(expected_types, list) else expected_types == "object":
```
girarda, line 104: nit: these double-inlined-ifs are difficult to parse. what do you think about
`if (isinstance(expected_type, list) and "object" in expected_types) or expected_types == "object"`?

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -65,17 +72,49 @@ def read_records_from_slice(self, stream_slice: StreamSlice) -> Iterable[Mapping
...
-                        logging.warning(f"Record did not pass validation policy: {record}")
+                        n_skipped += 1
                         continue
                     record[self.ab_last_mod_col] = file_datetime_string
                     record[self.ab_file_name_col] = file.uri
                     yield stream_data_to_airbyte_message(self.name, record)
                 self._cursor.add_file(file)
+
+            except StopSyncPerValidationPolicy:
+                yield AirbyteMessage(
```
girarda, line 90: Do we need a new status for connections that are "paused"?

Setting the sync as successful feels misleading since the data wasn't replicated.

non-blocking since we have a similar issue with other sources. [We do something similar for salesforce when we hit a rate limit](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-salesforce/source_salesforce/source.py#L151-L153)

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/validation_policy_scenarios.py
```diff
@@ -0,0 +1,484 @@
...
+        }
+    )
+    .set_file_type("csv")
+    .set_expected_catalog(
+        {
+            "streams": [
+                {
+                    "json_schema": {
+                        "type": "object",
+                        "properties": {
```
girarda, line 53: shouldn't the schema include the extra fields added by DefaultFileBasedStream (_ab_source_file_last_modified and _ab_source_file_url)?

airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
```diff
@@ -178,12 +219,18 @@ def read(capsys, tmp_path, scenario):
...
         ],
     )
-    captured = capsys.readouterr()
-    return [
-        msg
-        for msg in (json.loads(line) for line in captured.out.splitlines())
-        if msg["type"] == "RECORD"
-    ]
+    captured = capsys.readouterr().out.splitlines()
+    return (
```
girarda, line 223: nit: this method might be easier to use if it returned a Mapping[AirbyteMessage.Type: List[AirbyteMessage]]

girarda (commented):
@clnoll question about this proposal:
> Have the source reread the catalog
How would we pass the catalog path to the source?

I like this approach because it's not adding another setting that may or may not set anything on the stream.

One way to do this might be to remove the argument from `Source.read_catalog` and instead set it in it's __init__. This would require modifying entrypoint.py but I think that's fair play if we mark this as a breaking change

maxi297:
> Have the source reread the catalog

What I like about this approach is that it pushes the logic in the init so at that point, the source is up and ready and we don't have to do [this](https://github.com/airbytehq/airbyte/pull/27816/files#diff-51c861828e0a614ef6ee21390f3ba17d7138549f0575051fdb24868ee05eabbbR114)

I'm wondering if we can go even further and avoid coupling the source with the file system (i.e. reading a file to get the catalog). My main concerns are:
* We might have configs from somewhere else than the file system eventually but it doesn't change the source behavior per se which makes me think it should be outside (I know right now the source expose `read_catalog` but if we agree with the previous statement, we can make sure this will be easier to change)
* It's more annoying to tests (unit and integration tests)

In both cases, this is still a bit clunky because the catalog will be optional in the __init__ (the catalog is only passed on `read` commands) so we will have to validate the catalog wherever we use it internally (like `def _safely_get_catalog(self)` but I feel we're going in the right direction overall

clnoll (commented):
Thanks for the comments @maxi297 and @girarda.

I'm still considering where exactly to set the catalog on the source or stream, but have addressed the other comments.

@girarda question on this comment:
> One way to do this might be to remove the argument from Source.read_catalog and instead set it in it's init. This would require modifying entrypoint.py but I think that's fair play if we mark this as a breaking change

Is this still a breaking change if we follow the general pattern in the code snippet from the PR description (e.g. only removing `read_catalog` from file-based sources)?

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
girarda: > do you know whether related code changes would be isolated to the CDK?

I think we would need to update [the protocol](https://github.com/airbytehq/airbyte-protocol/blob/main/protocol-models/src/main/resources/airbyte_protocol/airbyte_protocol.yaml#L250-L253) and the platform to handle the new status

> would it be sufficient to change the level to Level.Error?
The platform doesn't use the log level to determine the status. We would want to raise an `AirbyteTracedException`

airbyte-cdk/python/airbyte_cdk/sources/streams/core.py
```diff
@@ -27,6 +27,8 @@
 # AirbyteMessage: An AirbyteMessage. Could be of any type
 StreamData = Union[Mapping[str, Any], AirbyteMessage]
 
+JsonSchema = Dict[str, Any]
```
girarda, line 30: nit: this could be a `Mapping[str, Any]` since the JsonSchema should be immutable

airbyte-cdk/python/unit_tests/sources/file_based/in_memory_files_source.py
```diff
@@ -25,18 +26,20 @@ def __init__(
             file_type,
             availability_strategy: AvailabilityStrategy,
             discovery_policy: AbstractDiscoveryPolicy,
-            validation_policies: AbstractSchemaValidationPolicy,
+            validation_policies: Dict[str, AbstractSchemaValidationPolicy],
             parsers: Dict[str, FileTypeParser],
             stream_reader: AbstractFileBasedStreamReader,
+            catalog: Optional[Dict[str, Any]],
```
girarda, line 32: It's not clear how this will be initialized in a real sync. Will that be done as part of https://github.com/airbytehq/airbyte/issues/27022 ?

clnoll: Yes that's right.

girarda (approved):

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/abstract_file_based_stream.py
```diff
@@ -124,7 +122,7 @@ def get_parser(self, file_type: str) -> FileTypeParser:
             raise UndefinedParserError(FileBasedSourceError.UNDEFINED_PARSER, stream=self.name, file_type=file_type)
 
     def record_passes_validation_policy(self, record: Mapping[str, Any]) -> bool:
-        return self.validation_policy.record_passes_validation_policy(record, self.get_json_schema())
+        return self.config.validation_policy.record_passes_validation_policy(record, self._catalog_schema)
```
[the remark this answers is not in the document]
clnoll: Potentially, yes. `get_json_schema` will re-run a schema discovery if there isn't a user-input schema.

airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py
```diff
@@ -65,6 +97,12 @@ def _choose_wider_type(key: str, t1: JsonSchemaSupportedType, t2: JsonSchemaSupp
         raise SchemaInferenceError(FileBasedSourceError.UNRECOGNIZED_TYPE, key=key, detected_types=f"{t1},{t2}")
 
 
+def is_equal_or_narrower_type(value: Any, expected_types: List[str]):
```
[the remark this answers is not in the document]
clnoll: That's a good point. I actually did some substantial reworking & testing of the functions in this file in the latest commit, because there was some dummy code in there that wasn't being exercised since we're only supporting CSVs/string types currently. When adding test coverage I found and fixed a couple of bugs.

airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py
```diff
@@ -74,7 +112,30 @@ def conforms_to_schema(record: Mapping[str, Any], schema: Mapping[str, str]) ->
...
+
+        if not isinstance(expected_types, list):
+            expected_types = [expected_types]
+
+        if value is not None:
+            if "object" in expected_types:
+                if not isinstance(value, dict):
+                    return False
+            elif "array" in expected_types:
+                if not isinstance(value, list):
```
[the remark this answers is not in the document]
clnoll: While testing I realized that this one is a little more complicated since we want to check that the types in the list conform to the schema, so I updated to do so.

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
clnoll: Ah, right. The `INCOMPLETE` status feels right to me. WDYT?

Issue #26343, opened 2023-05-22
Descending-order incremental sync module


flash1293:
### Component Description

There are APIs which do not allowed to filter by time and only return data ordered by date descending (a "feed" of data).

Examples:
* Discourse: https://docs.discourse.org/#tag/Posts/operation/listPosts
* Reddit: https://www.reddit.com/dev/api/#GET_new

The current incremental sync component doesn't cover this use case.

Idea for how to represent this:
* Do not specify end date
* Introduce a "stop condition" which is evaluated for each record - if the timestamp of a record is older than the state of the start_datetime for the stream, processing stops (because we reached the newest post from the last sync)

### Proposed YAML schema

```yaml
incremental_sync:
  type: "DatetimeBasedCursor"
  start_datetime:
    datetime: "{{ config['start_date'] }}"
    datetime_format: "%Y-%m"
  datetime_format: "%Y-%m-%dT%H:%M:%S%z"
  cursor_field: "pub_date"
  is_data_feed: true
```

Given `is_data_feed` is true, we will wrap the pagination strategy with `StopConditionPaginationStrategyDecorator` which is also implementing `PaginationStrategy`.  This class will return `None` to `next_page_token` if the condition is met.

PR #27818, opened 2023-06-29
[ISSUE #26343] update close_slice to use the greater record

maxi297:
## What
Following the data feed release, we expected to update the state while closing a slice using the latest record. This doesn't work because the latest record is not necessarily the one with the state being the greatest. For example, within the same HTTP request, posthog last record has cursor field value `2021-12-10T10:21:35.003000+00:00` but there is a record before with `2021-12-10T10:21:35.208000+00:00`

## How
Compare record using the cursor and only keep the one with the highest cursor field value.

## 🚨 User Impact 🚨
With the current CDK release following the data feed feature, connectors are at risk of not updating the cursor field properly. This should fix the issue

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
@@ -361,8 +361,6 @@ class SimpleRetriever(Retriever, HttpStream):
             response=response, stream_state=self.state, stream_slice=stream_slice, next_page_token=next_page_token
         )
         self._records_from_last_response = records
-        if records:
-            self._latest_record = records[-1]
         return records
 
     @property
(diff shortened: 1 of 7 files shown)


airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -361,8 +361,6 @@ def parse_response(
             response=response, stream_state=self.state, stream_slice=stream_slice, next_page_token=next_page_token
         )
         self._records_from_last_response = records
-        if records:
-            self._latest_record = records[-1]
```
clnoll, line 365: Looks like `self._latest_record` may not be used anywhere anymore, in which case I think we can remove it from `__post_init__`.

clnoll (approved):
LGTM, just one small comment.

PR #27840, opened 2023-06-29
Fixing an issue with square as CATs only compare string and not datetime

maxi297:
## What
Issue: https://storage.googleapis.com/airbyte-ci-reports-multi/airbyte-ci/connectors/test/pull_request/maxi297-state-changes_square/1688051383/35787de568588beab80ce8565b1a8290f1709dae/source-square/1.1.1/output.html

This is only seems like an issue because CATs only supports primitive as types for states. If we could have datetime, that would be neat but this is not something that can be changed easily.

## How
Ensure we use the cursor field value from the record and not a formatted version


clnoll (approved):
LGTM!

PR #27850, opened 2023-06-29
[26763] csv config options validation and use by reader

brianjlai:
## What
This PR adds the ability to parse config options related to the formatting of the CSV file in order to parse it into records. The main configurable options are a customer `delimiter`, `quote_char`, `escape_char`, and parsing a quote as a single quote using `doublequote`. For more details about how these options affect parsing ask ChatGPT.

The structure of the config file is modeled of the existing format block of `source-s3` which looks like:
```
{
  "name": "stream2",
  "file_type": "csv",
  "globs": ["b.csv"],
  "validation_policy": "emit_record_on_schema_mismatch",
  "format": {
    "filetype": "csv",
    "delimiter": "#",
    "escape_char": "@",
    "double_quote": true,
    "newlines_in_values": false
  }
}
```

## How
I'll break this into two aspects and include what refactoring was needed

Reading and validating format configs
- I tried to leverage pydantic validations as much as possible here so most of the boilerplate we get for free and we just define the validations. Pretty straightforward since we parse the configs into the pydantic model in the flow aleady.

Applying configs to the csv reader:
- The biggest gap in the interface that existed before the csv options is that the `parse_records()` and `infer_schema()` functions did not have access to the config. Now that we need the format from the config, we also need to modify the interface and all implementers. I would also expect most if not all other parsers to also need their own custom behavior based on the config so this change feels sensible.
- The alternative to the approach mentioned is refactoring how we instantiate parsers when we initialize the source. Right now we instatiate all the parsers into the map when the source is created. At this point we don't know what the config is because we aren't processing an operation. We would need to move the parser setup into these methods. Which is possible, but in prior PRs the intent was to be able to inject non-default parsers at compile time. Unless necessary I'd prefer to keep it that way
- As for the csv specific code, there is an equivalent concept to pyarrow for the python standard library's csv package called a Dialect. There is a default Dialect automatically used, but if format is found, we can register the new dialect and use it when reading.

## Recommended reading order
1. `file_base_stream_config.py`
2. `file_type_parser.py`
3. `csv_parser.py`
4. `default_file_based_stream.py`
5. `in_memory_files_source.py`
6. all the rest of the test changes


airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
```diff
@@ -0,0 +1,76 @@
...
+    @validator("encoding")
+    def validate_encoding(cls, v):
+        try:
+            codecs.lookup(v)
+        except LookupError:
+            raise ValueError(f"invalid encoding format: {v}")
+        return v
+
+
+class FileBasedStreamConfig(BaseModel):
```
brianjlai, line 54: I had to move the FileBasedStreamConfig to its own module because now as part of `parse_records()` and `infer_schema()`, we need to pass in the Config creates a circular dependency since stream.DefaultFileBasedStream → CsvParser → stream.FileBasedStreamConfig. By moving the package from stream -> config the circular dependency is removed

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/scenario_builder.py
```diff
@@ -41,6 +41,7 @@ def __init__(
             expected_discover_error: Tuple[Optional[Type[Exception]], Optional[str]],
             expected_read_error: Tuple[Optional[Type[Exception]], Optional[str]],
             incremental_scenario_config: Optional[IncrementalScenarioConfig],
+            file_write_options: Dict[str, Any]
```
brianjlai, line 44: I had to add some new boilerplate to the scenario builder since now we need to pass some format args to the CSV writer specifically. I would imagine other file type writers might also need to do the same once we get to them

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
```diff
@@ -5,18 +5,61 @@
...
-        with stream_reader.open_file(file) as fp:
-            reader = csv.DictReader(fp)
-            yield from reader
+    def parse_records(
+        self, config: FileBasedStreamConfig, file: RemoteFile, stream_reader: AbstractFileBasedStreamReader
+    ) -> Iterable[Dict[str, Any]]:
+        config_format = config.format
+        if config_format:
+            # Formats are configured individually per-stream so a unique dialect should be registered for each stream.
+            # Wwe don't unregister the dialect because we are lazily parsing each csv file to generate records
```
brianjlai, line 48: want to call this one out a bit. the way dialects work, its not something you pass to the reader, but rather you register it with the csv package. it's not great that we leave this registered during the lifecycle of the sync, but we can't unregister it because the yield keeps going beyond the lifecycle of this method and we don't know when it finishes here.

One alternative I explored was adding interface methods like pre/post processing for the parser, that would allow us to instantiate the dialect before the start and after the end of reading. If we think its worth expanding the interface to do that we can, but I've tested this and we do isolate dialects per-stream like we should as is.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -85,7 +85,7 @@ def streams(self, config: Mapping[str, Any]) -> List[AbstractFileBasedStream]:
                 stream_config = FileBasedStreamConfig(**stream)
                 streams.append(
                     DefaultFileBasedStream(
-                        config=FileBasedStreamConfig(**stream),
+                        config=stream_config,
```
brianjlai, line 88: we're alreadying parsing into `FileBasedStreamConfig` above, so no need to do it twice

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
girarda: could we check if it's already register, and if it is, fail if the dialect is different that the one from the config?

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
```diff
@@ -0,0 +1,76 @@
...
+    double_quote: bool
+    # Noting that the existing S3 connector had a config option newlines_in_values. This was only supported by pyarrow and not
+    # the Python csv package. It has a little adoption, but long term we should ideally phase this out because of the drawbacks
+    # of using pyarrow
+
+    @validator("delimiter")
+    def validate_delimiter(cls, v):
+        if len(v) != 1:
+            raise ValueError("delimiter should only be one character")
+        if v in {"\r", "\n"}:
```
girarda, line 29: should we also check for `"\r\n"`, which I think is what Windows uses?

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class CsvFormat(BaseModel):
+    filetype: str
+    delimiter: str = ","
+    quote_char: str = '"'
+    escape_char: Optional[str]
+    encoding: Optional[str] = "utf8"
+    double_quote: bool
+    # Noting that the existing S3 connector had a config option newlines_in_values. This was only supported by pyarrow and not
```
girarda, line 21: I think this can be supported using [quoting=csv.QUOTE_ALL](https://docs.python.org/3/library/csv.html#csv.QUOTE_ALL):

```
def test_csv():
    import csv

    with open('your_file.csv', newline='') as csvfile:
        reader = csv.reader(csvfile, quotechar='"', quoting=csv.QUOTE_ALL)
        # Skip the header
        next(reader)

        for row in reader:
            assert row[0] == "hello this is a string\nwith a new line"
```

with this csv file:
```
"h1","h2"
"hello this is a string
with a new line", "world"
```

Supporting different quote modes seems reasonable to me

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
brianjlai: is that necessary? if its more than one escape char it should fail the len(v) validation

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
```diff
@@ -0,0 +1,76 @@
...
+    name: str
+    file_type: str
+    globs: Optional[List[str]]
+    validation_policy: str
+    catalog_schema: Optional[ConfiguredAirbyteCatalog]
+    input_schema: Optional[Mapping[str, Any]]
+    primary_key: PrimaryKeyType
+    max_history_size: Optional[int]
+    days_to_sync_if_history_is_full: Optional[int]
+    format: Optional[CsvFormat]  # this will eventually be a Union once we have more than one format type
```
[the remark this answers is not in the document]
brianjlai: i'm not sure about enable? but i based this on the existing S3 implementation. The `format` object is built when you expand the optional block in the connector and then whatever you specify will be included in this object. Its existence tells us to build the dialect

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
clnoll: I'm wondering about a solution that involves giving each stream its own parser object(s; in case we don't require a single file type in the future). What I'm thinking it would look like is that after the creation of the stream config, we'd  create a copy of the parser, register the dialect, and pass that into the stream on init. The non-csv parsers may not have the dialect pattern, so we'd possibly have to implement that sort of thing on their behalf. 

Alternatively we could not instantiate the parsers when they're passed into the source, and init them with *args, **kwargs, which are obtained from the config and which are used to apply the dialect.

I'm not opposed to the way you're doing it here but am thinking of a situation where future-us need to add a method to the parser and forget that we need to apply the dialect.

clnoll (commented):
Nice @brianjlai, just a few questions & comments.

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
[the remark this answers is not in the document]
brianjlai: yeah that's a good point it might set us up better for that in the future. But we are still specifying `file_type` in the config though so we're not quite at the point where we can handle multiple ones yet. I think I can make this work, but I think this would potentially impact how we would migrate existing S3 connections since format is correlated to one object. 

But simplify things and not need to worry about a migration in the immediate, how would you feel if we added a bit of logic when we parse the config into the pydantic model. If the `format` is an object that contains `filetype` (the existing format), we instead transform the mapping to `csv` -> `format` object?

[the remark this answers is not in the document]
clnoll: 👍 I like that idea.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
brianjlai: I see, yeah it doesn't feel ideal to have to put the apply/create dialect as the first step for each of the parser's actions especially if we need to create more of them. However, it does also feel a even more hands on to have to add a new `apply_format()` method to the parser interface. And in that case, we would still have to make sure to call that method to every place we get the parser (`stream.get_parser(stream.config.file_type)`). 

In my opinion, I still think this is the change that gives us the most flexibility if we want to refactor how parsers get instantiated or how the format gets applied since we're not committing to more methods. So my vote would be leave it as is and extract this out later if we find it too unwieldly

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/file_type_parser.py
```diff
@@ -19,14 +20,16 @@ class FileTypeParser(ABC):
     """
 
     @abstractmethod
-    async def infer_schema(self, file: RemoteFile, stream_reader: AbstractFileBasedStreamReader) -> Schema:
+    async def infer_schema(self, config: FileBasedStreamConfig, file: RemoteFile, stream_reader: AbstractFileBasedStreamReader) -> Schema:
```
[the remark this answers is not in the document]
brianjlai: we could do that. the main argument for is that we don't know how other formats might use the config object so this is just a little more flexible. Do you think that would be valid since you've looked more into the other formats?

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
clnoll: True, an `apply_format` method isn't the nicest approach. I'm totally fine leaving this as-is at least for now.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/file_type_parser.py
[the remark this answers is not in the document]
clnoll: It feels like they should only need info from the format-specific part of the dictionary but I'm fine with keeping it more flexible until we know for sure.

clnoll (approved):

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/file_type_parser.py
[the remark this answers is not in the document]
clnoll: Actually I'm realizing that I have a need for a different part of the config in a separate PR that I'm working on, so yeah, let's keep it as you've got it.

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
brianjlai: Ahh interesting if we were to use that option, I hadn't considered it. Although, I do worry about trying to remap the existing `newlines_in_values` to the `quoting` concept because they're not entirely 1:1. A user might not quite intend for that behavior when they choose newlines_in_values. 

wdyt if we instead just opt for a new config option that is just called `quoting` and that allows for four values: `QUOTE_ALL`, `QUOTE_MINIMAL`, `QUOTE_NONE`, `QUOTE_NONNUMERIC`? From the UI we would probably just display all, minimal, none, or nonnumeric? And maybe we still aim to deprecate the old config once we start looking into moving S3 over

girarda: I agree we shouldn't remap it. Better to introduce the new option and deprecate `newlines_in_values`

girarda (approved):

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
brianjlai: okay sounds good. I'll add the new quoting options

PR #27872, opened 2023-06-30
Recommendation: Update test_incremental to be unaware when source defines the cursor

### LINKED ISSUE #23098: connector-acceptance-test: incremental tests failures on java sources

bnchrch:
# What was done
Update the Incremental CAT test to
1. Remove any cursor comparison
2. Remove the `cursor_paths` and `threshold_days` config values
3. Remove any assertions that are incorrect according to [this comment](https://github.com/airbytehq/airbyte/pull/27872#issuecomment-1696541769)

# Why
Our incremental cat tests can no longer make cursor assumptions given that cursors are no longer always available or always parsable.

# Below is the original issue written during discovery
Disclaimer: I dont understand this system perfectly and hope that I missed something here.

## Problem
Incremental Syncs are turned off for many database sources as they are failing incremental tests for an unclear reason

### 0. Hard to reason about code
Overall the `test_two_sequential_reads` test was hard to reason about but in untangling it I think we have been making two critical errors in how we handle state and cursors.

### 1. Cursors are not always defined and/or comparable
We have been treating ALL cursors as comparable and defined.

This is not true.

Some connectors (e.g. source-postgres) uses a source-defined cursor in many cases which is opaque to our system.

As a result in these cases we cannot test if records emitted are greater than or less than the cursor value.

Also not all cursors are comparable! A connector could know how to compare two cursor values when we dont.

For example a valid cursor field could be a string which represents an concatingated field `foo` which has a value `unixdatetime:123453;someotherdata:bar`

### 2. State is not meant to be readable

We have been treating state as parseable.

We should never do this.

The connector sets and interprets its own state.

There are many cases where us as a platform will never understand the contents of a state message

## Solution

Ive updated `test_two_sequential_reads` to
1. Never parse state for a specific value
2. Only make assertions on cursors IF they are comparable and IF source-defined-cursors are not defined
4. Assert that records are different on sequential reads

## Open Questions
There are still a few things I do not uderstand
### 1. Why do we allow connectors to defined `cursor_paths` in their test config
[Example](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/connector-acceptance-test/sample_files/acceptance-test-config.yml#L25)

Why do we do this when the cursor value is already defined in the configured catalog?

### 2. What else am I missing?
Is this a big miss by our tests or a big misunderstanding by me?

## Connectors failing tests for (now) known reasons
Currently Ive discovered two issues with the tests
### Source Postgres
Source postgres is emiting the same records on both reads.

### Source Bigquery
Source big query is emitting no records on the second read

## TODO
- [ ] Discuss with connector ops what I found
- [ ] Update unit tests
- [ ] Update slices and remove old records with state
- [ ] Test all connectors and find failing GA
- [ ] Create issues for each connector

closes #23098


airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py
```diff
@@ -180,24 +245,35 @@ def test_two_sequential_reads(
             latest_state = states_1[-1].state.data
             state_input = states_1[-1].state.data
 
-        parsed_records_1 = list(records_with_state(records_1, latest_state, stream_mapping, cursor_paths))
-
-        # This catches the case of a connector that emits an invalid state that is not compatible with the schema
-        # See https://github.com/airbytehq/airbyte/issues/21863 to understand more
-        assert parsed_records_1, "At least one valid state should be produced, given a cursor path"
-
-        for record_value, state_value, stream_name in parsed_records_1:
```
bnchrch, line 189: Copied from PR description:
---
**2. State is not meant to be readable**
We have been treating state as parseable.

We should never do this.

The connector sets and interprets its own state.

There are many cases where us as a platform will never understand the contents of a state message

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py
```diff
@@ -180,24 +245,35 @@ def test_two_sequential_reads(
...
+        # READ #2
 
         output = docker_runner.call_read_with_state(connector_config, configured_catalog_for_incremental, state=state_input)
         records_2 = filter_output(output, type_=Type.RECORD)
+        records_by_stream_2 = group_records_by_stream(records_2)
+
+        assert records_2, "Second Read should produce at least one record"
 
-        for record_value, state_value, stream_name in records_with_state(records_2, latest_state, stream_mapping, cursor_paths):
-            assert compare_cursor_with_threshold(
```
bnchrch, line 198: Copied from PR Description:

---

**1. Cursors are not always defined and/or comparable**
We have been treating ALL cursors as comparable and defined.

This is not true.

Some connectors (e.g. source-postgres) uses a source-defined cursor in many cases which is opaque to our system.

As a result in these cases we cannot test if records emitted are greater than or less than the cursor value.

Also not all cursors are comparable! A connector could know how to compare two cursor values when we dont.

For example a valid cursor field could be a string which represents an concatingated field foo which has a value unixdatetime:123453;someotherdata:bar

evantahler: I agree with your statement "Also not all cursors are comparable", heck they could be a bitmap or something like that.  That said, I do think we should be testing that if we replay a sync with the same state it just gave us, only the "latest" record should emitted, or no new records.  Is that this test, or another test?

evantahler (approved):
I _think_ I agree with your understanding... but I probably want to talk it out a little more.  Yes, it's confusing!

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py
```diff
@@ -148,6 +151,64 @@ def construct_latest_state_from_messages(messages: List[AirbyteMessage]) -> Dict
...
+    meant to be opaque to the platform and should not be relied on.
+    """
+    source_defined_cursor = stream_config.stream.source_defined_cursor
+
+    if not source_defined_cursor:
+        return True
+
+    cursor_field_helper = create_cursor_field_parser(stream_config)
+    record_value = cursor_field_helper.parse(record=example_record.record.data)
+    return is_comparable(record_value)
```
alafanechere, line 209: I'm not even sure we should try to compare cursor value. As you correctly stated this in your PR description: it's the connector role to define and control the cursor logic, we've no guarantee that what looks lower or greater than from the outside is not overridden inside the connector.
The `compare_cursor_with_threshold` method looks like a best effort things that could eventually break with an exoctic connector implementation. Keeping it would mean that, in case of failure, we'd  likely investigate deeply again to circle back to the same conclusion you come with in this PR.

alafanechere (approved):
Thank you for this thorough investigation and a clear explanation of the findings. I believe these cursor comparison problems arose when we introduced java source to CAT. The python connectors, CDK powered, cursor logic is quite standardized. I think it explains why these tests were not problematic in the past.

About the `cursor_path`: If the configured catalog expose source defined cursor there's no reason to keep them in acceptance-test-config.yml. I thought it was a way to manually declared source defined cursor in the test context. If source defined cursor can be exposed in the configured catalog it's better to keep them there an create configured catalog variants for each incremental test declared in `aceptance-test-config`.

evantahler:
Grooming Notes:
* We think we can remove cursor_paths, but check with @sherifnada 
* Goal: Postgres / BQ sources pass tests

sherifnada:
test_incremental will get way more fun with parallel state management too

bnchrch:
We Have a problem.

**TLDR: I tested 8 GA connectors, 3 of them are confirmed to pass incremental sync on master, however after investigation 2 of them should be failing that test.**

This is all due to our [2 year old test logic that doesnt check if the second read is empty.
](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py#L197C62-L197C62)
This means likely when we fix this test, many more connectors will start to fail.

What im doing now is figuring out how they should be behaving and if its a simple update we can apply to all silently failing connectors, or something more involved.

Raw testing notes
1. Postgres fails incremental on master and fails this branch
2. Bigquery fails incremental on master and fails this branch
3. Snowflake fails incremental on master and fails this branch
4. Github passes on master (but only because it returns no records on second read ⚠️) and fails this branch
4. Harvest passes on master (but only because it returns no records on second read ⚠️) and fails this branch
5. Freshdesk passes on master (legitimately) and fails this branch
6. Facebook marketing seemed to work but the log limit was clipped
6. google-analytics-data-api seemed to work but the log limit was clipped

bnchrch:
More notes:
- Source-jira passes both! yay

bnchrch:
Take aways from my conversation with @evantahler 

We can **never** assert
1. anything related to cursors (they maybe unparsable, non existant, etc)
2. that the last state message when used in a second read will produce or will not produce records. 
   1. if you read page 4, and page 4 is the last page, and is only partially filled, you may receive the same state message you used (`page:4`) back as you will need to re-read page 4 in the future.
   1. if you read page 4, and page 4 is the last page, and is fully filled, you may receive a new state message (`page:5`) back as you will need to re-read page 4 in the future.
   1. More data could be added after the read

If we are **given 2 or more state messages** we can assert
1. That state messages are different
2. That the second read using the first state produces different records than the first read
3. That the second read has records

We can always assert 
1. That one state message was produced from the first read
2. That one state message was produced from the second read 
4. That records were produced from the first read
5. The second read does not error

**Important learning: We can only start seriously asserting incremental properties when we enforce that connectors emit multiple state messages**

To do this will likely have to be attached to `high strictness level` testing and a subset of important connectors as the effort to increase the robustness of all connector test sandbox accounts is too high.

Whimsical: https://whimsical.com/incremental-cat-RcEn9tfxGWduFQyvUXBSDv

bnchrch:
@evantahler @alafanechere This is ready for your eyes again.

Since your last review I
1. Removed any cursor comparison
2. Removed the cursor_paths and threshold_days config values from the acceptance test config
3. Removed any assertions that are incorrect according to https://github.com/airbytehq/airbyte/pull/27872#issuecomment-1696541769
4. Updated the unit tests

bnchrch:
rerunning failed acceptances tests on master: https://github.com/airbytehq/airbyte/actions/runs/6077799393/job/16488080590 

Will compare to see if Ive introduced new failures

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_incremental.py
```diff
@@ -212,52 +165,70 @@ async def test_read_sequential_slices(
...
-                record_value <= state_value
-            ), f"First incremental sync should produce records younger or equal to cursor value from the state. Stream: {stream_name}"
+        # To avoid spamming APIs we only test a fraction of batches (10%) and enforce a minimum of 10 tested
+        min_batches_to_test = 5
+        sample_rate = len(unique_state_messages) // min_batches_to_test
 
-        checkpoint_messages = filter_output(output, type_=Type.STATE)
+        mutating_stream_name_to_per_stream_state = dict()
+        for idx, state_message in enumerate(unique_state_messages):
+            assert state_message.type == Type.STATE
```
alafanechere, line 198: This assertion is probably redundant as `unique_state_message` is built from a filtered output of `Type.STATE`.

alafanechere (approved):
👍 Thank you for the changes! Did you spotted any regression on the connectors side?
The unit tests for CAT passed.
The /legacy-publish command is a bit broken at the moment, let me know once you want to publish this.

bnchrch:
@alafanechere I went through the regressions noticed here: https://github.com/airbytehq/airbyte/pull/27872#issuecomment-1705717934

Overall 2 were real test failures, 1 was a failure of the comprehensive test (which I disabled for the connector) and the rest were non issues / flakey.

This indicates there will be more but I think its worth merging to see what comes out of this.

**Regression notes**
**source-paypal-transaction**
Produced the same set of records on first and second read (likely a real issue)

**source-zendesk-support**
Failed basic read on master and this branch on second attempt (likely non issue / flakey test)

**source-typeform**
Produced the same set of records on first and second read (likely a real issue)

**source-google-analytics-v4**
We hit a api quota alert (likely non issue)

**source-google-ads**
Read 1 and 7 the same in the comprehensive test. Disabled that test for this connector

bnchrch:
running publish (needs a comment)

PR #27887, opened 2023-06-30
Destinations V2: change singleton log message level from warn to debug

cynthiaxyin:
Retrieving a value from the destination config singleton is done in a fair amount of places so this was polluting the sync logs. Lower the log level from warn to debug.

Verified this doesn't show up anymore in a non-1s1t sync run locally.


evantahler (approved):
🤫

PR #27896, opened 2023-06-30
Switch "request parameters" to "query parameters" in connector builder docs

lmossman:
## What
We changed the connector builder to use the term "Query Parameters" instead of "Request Parameters" since we felt that was a clearer term to use.

However, not all of the connector builder docs were updated to reflect this change

## How
Update the connector builder documentation to reflect the term change


flash1293 (approved):
I think this is a good change, however we still call it `request_parameter` in the inject into dropdown. What do you think about adjusting it there as well using the separate label I introduced in https://github.com/airbytehq/airbyte-platform-internal/pull/7485?

Screenshot 2023-07-03 at 13 19 53

lmossman:
> I think this is a good change, however we still call it request_parameter in the inject into dropdown. What do you think about adjusting it there as well using the separate label I introduced in https://github.com/airbytehq/airbyte-platform-internal/pull/7485?

That's a good idea, I will put up an airbyte-platform-internal PR for this

Issue #27867, opened 2023-06-30
connectors-ci: abusive re-use of cache


alafanechere:
We can have the impression that is some context the test pipeline running on PR is abusively caching connector images:
* Even on code change the docker image is not always rebuilt 
* CAT seems running on the wrong connector image

This happened specifically on https://github.com/airbytehq/airbyte/pull/27658
I also suspect that [this CAT failure](https://storage.cloud.google.com/airbyte-ci-reports-multi/airbyte-ci/connectors/test/nightly_builds/master/1688343674/bdbe549a90a646f2568cba02c9e04e9783477626/source-strava/0.1.4/output.html) can be related to an abusive cache / image re-use.

These problems can't be reproduced locally.

alafanechere:
I think #27858 will reduce the moving parts and improve isolation to avoid this kind of problem.

alafanechere:
This issue might be partly solved by the upgrade to Dagger 0.6.4.
But we still observe transient failures on nightly cat, hopefully it will be solved by #27858

alafanechere:
> CAT seems running on the wrong connector image

#28656[ fixed it](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1690380501524629?thread_ts=1690376508.122769&cid=C02TYE9QL9M), it was not the wrong connector image but the wrong connector secret.

PR #27919, opened 2023-07-03
connectors-ci/cat: validate image under test is the right one

alafanechere:
## What
Addresses #27867
This PR aims at validating that the current image under test in CAT is the correct one to alert us explicitly when abusive caching might be happening on our test pipelines 

## How
1. On Python connector build: make the pipeline fail if the version label on the docker image does not match the metadata.yaml version
2. When we pass the connector under test image to CAT tag it with the git revision on not `:dev` and patch acceptance-test-config to use this git revision tag.
3. Add a test in CAT that fails if the docker image version label does not match the metadata.yaml version.

### DIFF
--- a/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
+++ b/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
@@ -82,7 +82,7 @@ DATE_PATTERN = "^[0-9]{2}-[0-9]{2}-[0-9]{4}$"
 DATETIME_PATTERN = "^[0-9]{4}-[0-9]{2}-[0-9]{2}(T[0-9]{2}:[0-9]{2}:[0-9]{2})?$"
 
 
-@pytest.mark.default_timeout(10)
+@pytest.mark.default_timeout(30)
 class TestSpec(BaseTest):
 
     spec_cache: ConnectorSpecification = None
--- a/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
+++ b/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
@@ -126,13 +126,6 @@ class TestSpec(BaseTest):
         if connector_spec:
             assert actual_connector_spec == connector_spec, "Spec should be equal to the one in spec.yaml or spec.json file"
 
-    def test_docker_env(self, actual_connector_spec: ConnectorSpecification, docker_runner: ConnectorRunner):
-        """Check that connector's docker image has required envs"""
-        assert docker_runner.env_variables.get("AIRBYTE_ENTRYPOINT"), "AIRBYTE_ENTRYPOINT must be set in dockerfile"
-        assert docker_runner.env_variables.get("AIRBYTE_ENTRYPOINT") == " ".join(
-            docker_runner.entry_point
-        ), "env should be equal to space-joined entrypoint"
-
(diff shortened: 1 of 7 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -812,13 +818,16 @@ async def with_airbyte_python_connector(context: ConnectorContext, build_platfor
...
         .with_label("io.airbyte.name", context.metadata["dockerRepository"])
     )
     cdk_version = await get_cdk_version_from_python_connector(connector_container)
     if cdk_version:
         connector_container = connector_container.with_label("io.airbyte.cdk_version", cdk_version)
         context.cdk_version = cdk_version
+    if not await connector_container.label("io.airbyte.version") == context.metadata["dockerImageTag"]:
+        raise Exception(
+            "Abusive caching might be happening. The connector container should have been built with the correct version as defined in metadata.yaml"
+        )
```
evantahler, line 830: 👍

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
```diff
@@ -538,6 +531,30 @@ def test_additional_properties_is_true(self, actual_connector_spec: ConnectorSpe
...
+    # This test should not be part of TestSpec because it's testing the connector's docker image content, not the spec itself
+    # But it's cumbersome to declare a separate, non configurable, test class
+    # See https://github.com/airbytehq/airbyte/issues/15551
+    def test_image_labels(self, docker_runner: ConnectorRunner, connector_metadata: dict):
+        """Check that connector's docker image has required labels"""
+        assert docker_runner.labels.get("io.airbyte.name"), "io.airbyte.name must be set in dockerfile"
+        assert docker_runner.labels.get("io.airbyte.version"), "io.airbyte.version must be set in dockerfile"
+        assert (
+            docker_runner.labels.get("io.airbyte.name") == connector_metadata["data"]["dockerRepository"]
+        ), "io.airbyte.name must be equal to dockerRepository in metadata.yaml"
```
evantahler, line 543: One day, it would be nice to not need to modify the docker labels at all.  I'm curious if we even still need them.  I don't think the platform uses them at all...

evantahler (approved):

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
alafanechere: Agreed! It's just a good enough way for me right now to try to detect and understand the CI problem we might have with caching.
The metadata.yaml is not packaged into the connector image on the Python connectors, if we revamped as planned the build logic for python connector I think we shall package metadata.yaml in the connector image.
Wdyt about a new connector command like `metadata` that would output the connector metadata?

evantahler: I like it.  Then I can `airbybye-ci connectors metadata --name source-foo | jq` :D

alafanechere:
/legacy-test connector=connectors/source-openweather connector-acceptance-test-version=dev

airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/tests/test_core.py
alafanechere: @evantahler I meant a connector command:
`docker run airbyte/pokeapi metadata | jq`

PR #27940, opened 2023-07-04
connectors-ci: mount secrets in a late layer for CAT

alafanechere:
## What
Addresses #27867

Mount secret to the CAT container in a later layer.
I'm not sure how well the mounted secrets layer is burst on every CAT run. I suspect failure like [this one](https://storage.cloud.google.com/airbyte-ci-reports-multi/airbyte-ci/connectors/test/nightly_builds/master/1688343674/bdbe549a90a646f2568cba02c9e04e9783477626/source-strava/0.1.4/output.html) to be due to abusively cached secrets / connector config.

## How
Mount secrets later in the CAT container declaration to make sure they are re-mounted every time the cache is bursted.

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
@@ -12,7 +12,7 @@ import re
 import uuid
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


octavia-approvington:
What are we doing again? <br/> merge and squash

octavia-approvington (approved):

Issue #21014, opened 2023-01-04
Connector builder: Show all request/responses as part of the testing panel


flash1293:
Currently, the data requests for all pages are shown in the testing panel. However, in some cases there might be additional requests happening which are not presented to the user at the moment:
* Substream slicer requests
* Multi-step authentication like session token

## Acceptance criteria

* Testing panel shows all requests/responses being made as part of a read call
* Testing panel allows to switch through the requests for debugging in a suitable way

flash1293:
+1 from public Slack channel, hard to debug the oauth flow without seeing the actual requests

flash1293:
Implementation notes:
* Introduce a decorator wrapper class for the oauth authenticator class in the same way we have it for the simple retriever
* Emit a [request and response message](https://github.com/airbytehq/airbyte/blob/3f4d54b85fa6361126ca72e38db6ac019aa048d9/airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py#L485) using the new message repository - add a new property `type: "global"` and a title and description: `title: "Refresh token", description: "Obtains access token"` to it
* Adjust the message grouper to filter out these requests and responses
* Add it to the top level stream read object next to the logs:
```
    StreamRead:
      type: object
      properties:
        global_requests:
          type: array
          items:
            type: object
            properties:
               request:
                      $ref: "#/components/schemas/HttpRequest"
               response:
                      $ref: "#/components/schemas/HttpResponse"
```

In the UI, if there are global requests, add a vertical separator below the logs that contains "Supplementary requests":
<img width="423" alt="Screenshot 2023-06-22 at 13 35 14" src="https://github.com/airbytehq/airbyte/assets/1508364/ebc82059-b360-4895-9bf3-35a7acf8f00e">

This is reusing the existing component used for showing multiple partitions

PR #27973, opened 2023-07-05
Issue 21014/oauth requests

maxi297:
## What
Addresses https://github.com/airbytehq/airbyte/issues/21014 that targets specifically oauth requests. There will be another PR for SubstreamPartitionRouter.

Considered out of scope: Oauth2Authenticator as it is not used by the connector builder and therefore, those logs aren't relevant for now

## How
By adding a message repository on the AbstractOauth2Authenticator, we can emit log messages. To make the work more easily, I've done a couple of things:
* Group the request and response in one log
    * Reason: We always emitted both at the same time anyway and having this in two different logs increased the number of things to consider in the MessageGrouper. In the end, if there is a situation where we perform a request but don't have a response, the use will see it with a trace message since it'll raise a breaking exception
    * Format chosen: ECS as it already solved the problem of structuring logging for HTTP request/response but I'm open to something else
* Added explicit logging interface `log_message` to the MessageRepository. The reason is that we only want to log given a specific log level. However, if the log level is too low to be logged, we don't want do to the processing of creating the message. The `log_message` allows for lazy evaluation
* Created a NoopMessageRepository. I didn't want `message_repository` to be optional and that everywhere we need it, we need to "if self.message_repository". This improves readability

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/abstract_oauth.py`: where the logging occurs
2. `airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py`: where this log is intercepted and assembled as "global_requests"
3. `airbyte-cdk/python/airbyte_cdk/sources/http_logger.py`: new log format

## 🚨 User Impact 🚨
As of today, none since this is `ModelToComponentFactory.emit_connector_builder_messages` flag (so only enabled for Connector Builder) and there will need to be frontend change to allow the user to see it.

## To test single use (using quickbooks)

`connector_builder/main.py read --config secrets/config.json --catalog secrets/catalog.json`

<details><summary>config (update `credentials` using https://developer.intuit.com/app/developer/playground)</summary>
<p>

```
{
    "__injected_declarative_manifest" : {
        "version": "0.39.1",
        "spec": {
            "documentationUrl": "https://docs.airbyte.com/integrations/sources/quickbooks",
            "connection_specification": {
                "title": "Source QuickBooks Spec",
                "type": "object",
                "required": [
                    "credentials",
                    "start_date",
                    "sandbox"
                ],
                "additionalProperties": true,
                "properties": {
                    "credentials": {
                        "title": "Authorization Method",
                        "type": "object",
                        "order": 0,
                        "oneOf": [
                            {
                                "type": "object",
                                "title": "OAuth2.0",
                                "required": [
                                    "client_id",
                                    "client_secret",
                                    "refresh_token",
                                    "access_token",
                                    "token_expiry_date",
                                    "realm_id"
                                ],
                                "properties": {
                                    "auth_type": {
                                        "type": "string",
                                        "const": "oauth2.0"
                                    },
                                    "client_id": {
                                        "type": "string",
                                        "title": "Client ID",
                                        "description": "Identifies which app is making the request. Obtain this value from the Keys tab on the app profile via My Apps on the developer site. There are two versions of this key: development and production."
                                    },
                                    "client_secret": {
                                        "description": " Obtain this value from the Keys tab on the app profile via My Apps on the developer site. There are two versions of this key: development and production.",
                                        "title": "Client Secret",
                                        "type": "string",
                                        "airbyte_secret": true
                                    },
                                    "refresh_token": {
                                        "description": "A token used when refreshing the access token.",
                                        "title": "Refresh Token",
                                        "type": "string",
                                        "airbyte_secret": true
                                    },
                                    "access_token": {
                                        "description": "Access token fot making authenticated requests.",
                                        "title": "Access Token",
                                        "type": "string",
                                        "airbyte_secret": true
                                    },
                                    "token_expiry_date": {
                                        "type": "string",
                                        "title": "Token Expiry Date",
                                        "description": "The date-time when the access token should be refreshed.",
                                        "format": "date-time"
                                    },
                                    "realm_id": {
                                        "description": "Labeled Company ID. The Make API Calls panel is populated with the realm id and the current access token.",
                                        "title": "Realm ID",
                                        "type": "string",
                                        "airbyte_secret": true
                                    }
                                }
                            }
                        ]
                    },
                    "start_date": {
                        "order": 1,
                        "description": "The default value to use if no bookmark exists for an endpoint (rfc3339 date string). E.g, 2021-03-20T00:00:00+00:00. Any data before this date will not be replicated.",
                        "title": "Start Date",
                        "type": "string",
                        "format": "date-time",
                        "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$",
                        "examples": [
                            "2021-03-20T00:00:00+00:00"
                        ]
                    },
                    "sandbox": {
                        "order": 2,
                        "description": "Determines whether to use the sandbox or production environment.",
                        "title": "Sandbox",
                        "type": "boolean",
                        "default": false
                    }
                }
            }
        },
        "definitions": {
            "schema_loader": {
                "type": "InlineSchemaLoader",
                "schema": {
                    "$schema": "http://json-schema.org/schema#",
                    "properties": {
                        "count": {
                            "type": "integer"
                        },
                        "next": {
                            "type": "string"
                        },
                        "potato": {
                            "type": "string"
                        },
                        "previous": {
                            "type": "null"
                        },
                        "results": {
                            "items": {
                                "properties": {
                                    "name": {
                                        "type": "string"
                                    },
                                    "url": {
                                        "type": "string"
                                    }
                                },
                                "type": "object"
                            },
                            "type": "array"
                        }
                    },
                    "type": "object"
                }
            },
            "selector": {
                "type": "RecordSelector",
                "extractor": {
                    "type": "DpathExtractor",
                    "field_path": [
                        "QueryResponse",
                        "{{ parameters.table_name }}"
                    ]
                }
            },
            "requester": {
                "type": "HttpRequester",
                "url_base": "{{ 'https://sandbox-quickbooks.api.intuit.com' if config.sandbox else 'https://quickbooks.api.intuit.com' }}",
                "http_method": "GET",
                "request_parameters": {
                    "query": "SELECT * FROM {{ parameters.table_name }} WHERE Metadata.LastUpdatedTime > '2023-01-01T09:29:05-08:00' AND Metadata.LastUpdatedTime <= '2023-06-01T09:29:05-08:00' {{ parameters.additional_where }} ORDER BY Metadata.LastUpdatedTime ASC STARTPOSITION {{ (next_page_token.next_page_token or 0) + 1 }} MAXRESULTS {{ (config.max_results or '200') | int }}"
                },
                "request_headers": {
                    "Accept": "application/json",
                    "User-Agent": "airbyte-connector"
                },
                "authenticator": {
                    "type": "OAuthAuthenticator",
                    "client_id": "{{ config['credentials']['client_id'] }}",
                    "client_secret": "{{ config['credentials']['client_secret'] }}",
                    "refresh_token": "{{ config['credentials']['refresh_token'] }}",
                    "token_refresh_endpoint": "https://oauth.platform.intuit.com/oauth2/v1/tokens/bearer",
                    "refresh_token_updater": {}
                }
            },
            "retriever": {
                "type": "SimpleRetriever",
                "record_selector": {
                    "$ref": "#/definitions/selector"
                },
                "paginator": {
                    "type": "DefaultPaginator",
                    "pagination_strategy": {
                        "type": "OffsetIncrement",
                        "page_size": "{{ (config.max_results or '5') | int }}"
                    }
                },
                "requester": {
                    "$ref": "#/definitions/requester"
                }
            },
            "base_stream": {
                "primary_key": "Id",
                "retriever": {
                    "$ref": "#/definitions/retriever"
                },
                "incremental_sync": {
                    "type": "DatetimeBasedCursor",
                    "cursor_field": "airbyte_cursor",
                    "start_datetime": "2023-02-10T00:00:00.000000+00:00",
                    "end_datetime": "2023-06-02T14:51:23.005955+00:00",
                    "datetime_format": "%Y-%m-%dT%H:%M:%S.%f%z",
                    "cursor_granularity": "PT0S",
                    "step": "P30D"
                },
                "transformations": [
                    {
                        "type": "AddFields",
                        "fields": [
                            {
                                "path": [
                                    "airbyte_cursor"
                                ],
                                "value": "{{ record.MetaData.LastUpdatedTime }}"
                            }
                        ]
                    }
                ],
                "$parameters": {
                    "path": "/v3/company/{{ config.credentials.realm_id }}/query",
                    "table_name": "BillPayment"
                }
            },
            "accounts": {
                "$ref": "#/definitions/base_stream",
                "name": "accounts",
                "$parameters": {
                    "$ref": "#/definitions/base_stream/$parameters",
                    "table_name": "Account",
                    "additional_where": "AND Active IN (true, false)"
                }
            }
        },
        "streams": [
            "#/definitions/accounts",
        ],
        "check": {
            "type": "CheckStream",
            "stream_names": [
                "accounts"
            ]
        }
    },
    "__command" : "test_read",
    "credentials": {
        "auth_type": "oauth2.0",
        "client_id": <client id>,
        "client_secret": <client secret>,
        "refresh_token": <refresh token>,
        "access_token": <access token>,
        "token_expiry_date": "2022-01-01T00:00:00.000000Z",
        "realm_id": <realm id>
    },
    "start_date": "2023-02-10T00:00:00Z",
    "sandbox": true
}


```

</p>
</details>

<details><summary>catalog</summary>
<p>

```
{
  "streams": [
    {
      "stream": {
        "name": "accounts",
        "json_schema": {},
        "supported_sync_modes": ["full_refresh"]
      },
      "sync_mode": "full_refresh",
      "destination_sync_mode": "overwrite"
    }
  ]
}

```

</p>
</details>

## To test reusable refresh token (using quickbooks)
To the above but remove `refresh_token_updater` from the config

In both cases, output includes:
```
            "global_requests": [
                {
                    "request": {
                        "url": "https://oauth.platform.intuit.com/oauth2/v1/tokens/bearer",
                        "parameters": null,
                        "headers": {
                            "User-Agent": "python-requests/2.30.0",
                            "Accept-Encoding": "gzip, deflate",
                            "Accept": "*/*",
                            "Connection": "keep-alive",
                            "Content-Length": "205",
                            "Content-Type": "application/x-www-form-urlencoded"
                        },
                        "http_method": "POST",
                        "body": <redacted>
                    },
                    "response": {
                        "status": 200,
                        "body": <redacted>,
                        "headers": {
                            "Date": "Wed, 05 Jul 2023 13:35:35 GMT",
                            "Content-Type": "application/json;charset=utf-8",
                            "Content-Length": "1178",
                            "Connection": "keep-alive",
                            "intuit_tid": "1-64a571a7-2e51314f33a272710cec155e",
                            "x-spanid": "a8335f52-4dc8-4e1f-8bdc-920ea51b4a44",
                            "x-amzn-trace-id": "Root=1-64a571a7-2e51314f33a272710cec155e",
                            "X-Content-Type-Options": "nosniff",
                            "Server": "envoy",
                            "Cache-Control": "no-cache, no-store",
                            "Pragma": "no-cache",
                            "X-Frame-Options": "SAMEORIGIN",
                            "X-XSS-Protection": "1; mode=block",
                            "x-envoy-upstream-service-time": "77",
                            "Strict-Transport-Security": "max-age=31536000"
                        }
                    }
                }
            ],
```


airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
```diff
@@ -214,10 +215,12 @@ def get_access_token(self) -> str:
             self.access_token = new_access_token
             self.set_refresh_token(new_refresh_token)
             self.set_token_expiry_date(new_token_expiry_date)
-            if self._message_repository:
+            # FIXME emit_configuration_as_airbyte_control_message as been deprecated in favor of package airbyte_cdk.sources.message
```
girarda, line 218: are the messages emitted by the message repository not printed to stdout by the entrypoint?

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -410,22 +409,39 @@ def read_records(
...
+            return None
+
+    @staticmethod
+    def _extract_record(stream_data: StreamData, stream_slice: StreamSlice) -> Optional[Record]:
+        """
+        As we allow the output of _read_pages to be StreamData, it can be multiple things. Therefore, we need to filter out and normalize
+        to data to streamline the rest of the process.
+        """
+        if isinstance(stream_data, Record):
+            # Record is not part of `StreamData` but is the most common implementation of `Mapping[str, Any]` which is part of `StreamData`
```
girarda, line 438: is there a reason `Record` shouldn't be part of `StreamData`?

girarda (approved):

airbyte-cdk/python/airbyte_cdk/sources/message/repository.py
```diff
@@ -15,14 +15,43 @@ class MessageRepository(ABC):
     def emit_message(self, message: AirbyteMessage) -> None:
         raise NotImplementedError()
 
+    @abstractmethod
+    def log_message(self, level: Level, message_provider: Callable[[], AirbyteMessage]) -> None:
```
maxi297, line 19: FIXME: should be `Callable[[], str]` and as we would support any JSON, I'll probably update it to 
`JsonType = Union[dict[str, "JsonType"], list["JsonType"], str, int, float, bool, None]`

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
maxi297: This is a good question. I would tend to try to remove things from `StreamData` instead of adding some. I honestly hate the fact that we need to check the type of the return object when calling a method so I would try to avoid adding to this problem. Personally, I would love to have `HttpStream._read_pages` return records but I haven't evaluate the work required to do this. As we allow for `CustomRetriever`, it might be harder but at the same time, I don't see any in airbyte-integrations

airbyte-cdk/python/airbyte_cdk/sources/streams/http/requests_native_auth/oauth.py
maxi297: I think the issue is that we have defined a default value for `message_repository` to keep backward compatibility. If a message_repository is not provide, the behavior would change. As `NoopMessageRepository` is the default class for `message_repository`, we basically need to check "if default value, keep previous behavior". Changing this would mean breaking the config control message for all oauth sources that aren't declarative as only `ManifestDeclarativeSource` defines a message repository today. This can't be in the `AbstractSource` as it is abstract

flash1293 (commented):
Tested e2e in the connector builder and the global request for the authentication is coming through as expected - very helpful for debugging! It also works if the authentication request fails which is very important.

One question: By https://github.com/airbytehq/airbyte/issues/21014#issuecomment-1602483955 , there should be a title/description in here as well per request, do you plan to add that as part of this issue?

PR #28019, opened 2023-07-06
🐛 Source Square: fix display order of spec fields

### LINKED ISSUE #25158: Incorrect value for `order` field for `Square` source connector specification.
https://github.com/airbytehq/airbyte/blob/23c9a299e5d7e247dbff120f36869f496755a320/airbyte-integrations/connectors/source-square/source_square/spec.yaml#L70

This value should be `order: 2`?
The comma in the value, causes the `order` field to be interpreted as string instead of an integer, invalidating the type constraints.

ParthG21:
### What
resolved: #25158

### How
Removed comma from the value of order field for square source connector specification, which was causing it to be a String instead of an Integer.


sh4sh:
Nice catch! Confirmed they are showing out of order in latest, `start_date` should appear first but since it doesn't have a valid `order` field it's dropped to the end.
Screenshot 2023-07-10 at 12 54 58 PM

sh4sh (approved):

sh4sh:
Thanks for your contribution!

PR #28022, opened 2023-07-06
📝 Docs: update HubSpot source page

ChristoGrab:
## What
Updated the Hubspot docs and in-app setup guide

## How
- I reworked the general flow of the prerequisites/setup steps to match the pattern that @sherifnada mentioned liking of:

--Step 1: set up authentication method in source
--Step 2: set up additional configurations in source
--Step 3: set up the connector in Airbyte

I think this should help prevent users from getting ahead of themselves and realizing they have to go back to the source and spend more time with configuring their account, finding keys, etc.

- Updated in-app connector setup steps to more closely match the full docs.
- Removed an erroneous reference to manually configuring scopes for Airbyte Cloud OAuth.
- Specified the keys needed for Open Source OAuth and Private App setup.
- Removed a reference to leaving the start date blank (it is a required field).
- Moved the note on incremental sync from streams to sync modes section.
- Merged two sections 'Notes on engagements stream' into one section with bullet points.
- Fixed a :::note::: that was unclosed and consuming the changelog 🤣 
- Some small grammar fixes here and there.

## Recommended reading order
1. hubspot.md
2. hubspot.inapp.md


docs/integrations/sources/hubspot.inapp.md
```diff
@@ -1,17 +1,13 @@
-## Prerequisite
+## Prerequisites
 
-* Access to your HubSpot account
+### Authentication method
+You can use OAuth or a Private App to authenticate your HubSpot account.
```
sherifnada, line 4: we should keep the recommendation for oauth on cloud

docs/integrations/sources/hubspot.md
```diff
@@ -1,14 +1,43 @@
...
+## Setup guide
+
+:::note
+Airbyte Cloud users authenticating via OAuth can skip straight to 
+[setting up the connector in Airbyte](#step-3-set-up-the-hubspot-source-connector-in-airbyte).  All other users must manually set up authentication and configure scopes.
+:::
+
+### Step 1: Set up the authentication method
+
+#### Private App setup (Recommended for Airbyte Open Source)
```
sherifnada, line 23: Could you wrap oss or cloud-specific versions in `<!--env:[oss or cloud]-->` tags? See [example](https://github.com/airbytehq/airbyte/blob/master/docs/integrations/sources/airtable.md?plain=1#L21-L35) 

This makes it so that when we display these docs inside the app itself, we only display the sections between tags if they match the env e.g: `<!--env:oss-->` only shown on OSS and so on. 

this would make things like the `:::note` above redundant

docs/integrations/sources/hubspot.md
```diff
@@ -1,14 +1,43 @@
...
+This page contains the setup guide and reference information for the [HubSpot](https://www.hubspot.com/) source connector.
 
-## Prerequisite
+## Prerequisites
 
-You can use OAuth or a Private App to authenticate your HubSpot account.
+You can either use OAuth or set up a Private App to authenticate your HubSpot account.
 
-In Airbyte Cloud, we highly recommend you use OAuth and not Private App authentication as it significantly simplifies the setup process.
+For Airbyte Cloud users, we highly recommend you use OAuth rather than Private App authentication, as it significantly simplifies the setup process. Conversely, for Airbyte Open Source users we recommend Private App authentication.
```
sherifnada, line 9: just making it visually clearer for anyone quickly scanning the page

```suggestion
**For Airbyte Cloud** users, we highly recommend you use OAuth rather than Private App authentication, as it significantly simplifies the setup process. 

**For Airbyte Open Source** users we recommend Private App authentication.
```

docs/integrations/sources/hubspot.md
```diff
@@ -1,14 +1,43 @@
 # HubSpot
 
-This page guides you through setting up the HubSpot source connector.
+This page contains the setup guide and reference information for the [HubSpot](https://www.hubspot.com/) source connector.
 
-## Prerequisite
+## Prerequisites
 
-You can use OAuth or a Private App to authenticate your HubSpot account.
```
sherifnada, line 7: I think of pre-reqs as being a TL;DR of what's needed. To this end it should be something like: 

```
## Prerequisites
- If you are using **Airbyte Open Source** Private app
- If you are using **Airbyte Cloud** no additional pre-requisites are needed other than a Hubspot account
```

docs/integrations/sources/hubspot.md
```diff
@@ -100,30 +136,35 @@ The HubSpot source connector supports the following streams:
...
 - A `call` engagement has a corresponding `engagements_metadata` object with non-null values in the `toNumber`, `fromNumber`, `status`, `externalId`, `durationMilliseconds`, `externalAccountId`, `recordingUrl`, `body`, and `disposition` columns.
 - An `email` engagement has a corresponding `engagements_metadata` object with non-null values in the `subject`, `html`, and `text` columns. In addition, there will be records in four related tables, `engagements_metadata_from`, `engagements_metadata_to`, `engagements_metadata_cc`, `engagements_metadata_bcc`.
 - A `meeting` engagement has a corresponding `engagements_metadata` object with non-null values in the `body`, `startTime`, `endTime`, and `title` columns.
 - A `note` engagement has a corresponding `engagements_metadata` object with non-null values in the `body` column.
 - A `task` engagement has a corresponding `engagements_metadata` object with non-null values in the `body`, `status`, and `forObjectType` columns.
 
+2. The `engagements` stream uses two types of API:
+- **EngagementsRecent** if `start_date/last_sync_date` is less than 30 days and the API is able to return all records (<10k).
```
sherifnada, line 158: The wording here is a bit rough, we should rephrase to make it clearer. Fundamentally this section is saying: "FYI this stream might take a while to sync if it's been a while since you synced data or if there's a ton of data generated since the last sync, so we recommend scheduling frequent syncs"

sherifnada (changes requested):

docs/integrations/sources/hubspot.md
ChristoGrab: Gotcha, I added in these bullet points and moved the link for more info on authentication to the top of the setup section

docs/integrations/sources/hubspot.md
ChristoGrab: Love this feature! I've added the tags to the OAuth setup for OSS and the branching Authentication dropdown instructions during connector setup.

I left the Private App and Scope sections visible to both since Cloud users can still choose to use Private App auth. Would you prefer I made these oss specific as well? Not sure how strongly we want to veer them away from this path.

docs/integrations/sources/hubspot.inapp.md
ChristoGrab: Roger that

docs/integrations/sources/hubspot.md
ChristoGrab: Done!

docs/integrations/sources/hubspot.md
ChristoGrab: Changed this to hopefully be clearer. Now reads:

```
2. The `engagements` stream uses two different APIs based on the length of time since the last sync and the number of records.
- **EngagementsRecent** if the following two criteria are met:
    - The last sync was performed within the last 30 days
    - Fewer than 10,000 records are being synced
- **EngagementsAll** if either of these criteria are not met.

Because of this, the `engagements` stream can be slow to sync if it hasn't synced within the last 30 days and/or is generating large volumes of new data. We therefore recommend scheduling frequent syncs.
```

Let me know if it's still not quite there yet!

docs/integrations/sources/hubspot.md
```diff
@@ -36,17 +65,28 @@ If you are using either OAuth in Airbyte OSS or Private App authentication, you
...
+### Step 3: Set up the HubSpot source connector in Airbyte
+
+1. Log in to your [Airbyte Cloud](https://cloud.airbyte.com/workspaces) or Airbyte Open Source account.
+2. From the Airbyte UI, click **Sources**, then click on **+ New Source** and select **HubSpot** from the list of available sources.
+3. Enter a **Source name** of your choosing.
+4. From the **Authentication** dropdown, select your chosen authentication method:
+
+<!-- env:cloud -->
+#### For Airbyte Cloud users:
+- To authenticate using OAuth, select **OAuth** and click **Authenticate your HubSpot account** to sign in with HubSpot and authorize your account.
```
sherifnada, line 77:
```suggestion
- **Recommended:** To authenticate using OAuth, select **OAuth** and click **Authenticate your HubSpot account** to sign in with HubSpot and authorize your account.
```

docs/integrations/sources/hubspot.md
```diff
@@ -36,17 +65,28 @@ If you are using either OAuth in Airbyte OSS or Private App authentication, you
...
+- To authenticate using a Private App, select **Private App** and enter the Access Token for your HubSpot account.
```
sherifnada, line 78:
```suggestion
- **Not recommended**: To authenticate using a Private App, select **Private App** and enter the Access Token for your HubSpot account.
```

docs/integrations/sources/hubspot.md
```diff
@@ -100,30 +140,37 @@ The HubSpot source connector supports the following streams:
...
 - A `call` engagement has a corresponding `engagements_metadata` object with non-null values in the `toNumber`, `fromNumber`, `status`, `externalId`, `durationMilliseconds`, `externalAccountId`, `recordingUrl`, `body`, and `disposition` columns.
 - An `email` engagement has a corresponding `engagements_metadata` object with non-null values in the `subject`, `html`, and `text` columns. In addition, there will be records in four related tables, `engagements_metadata_from`, `engagements_metadata_to`, `engagements_metadata_cc`, `engagements_metadata_bcc`.
 - A `meeting` engagement has a corresponding `engagements_metadata` object with non-null values in the `body`, `startTime`, `endTime`, and `title` columns.
 - A `note` engagement has a corresponding `engagements_metadata` object with non-null values in the `body` column.
 - A `task` engagement has a corresponding `engagements_metadata` object with non-null values in the `body`, `status`, and `forObjectType` columns.
 
+2. The `engagements` stream uses two different APIs based on the length of time since the last sync and the number of records.
```
sherifnada, line 161:
```suggestion
2. The `engagements` stream uses two different APIs based on the length of time since the last sync and the number of records which Airbyte hasn't yet synced.
```

sherifnada (approved):

PR #28025, opened 2023-07-06
script to assign permissions to service account

c-p-b:
## What
Google Cloud does not support using wildcards to match resource names in IAM policies, including in Google Secret Manager. This means you can't directly create a policy that matches secrets with a certain prefix. In our case we need to attach a lot of secrets to a specific new service account. This bash script helps us iterate through secrets and achieve that

## How
List secrets and filter by a prefix. An example is provided

## 🚨 User Impact 🚨
N/A


cpdeethree:
> It was useful as an adhoc thing to allow dagger to read our secrets, but I don't think we have a reason to use it for other things, feel free to close

@alafanechere via chat

PR #28052, opened 2023-07-07
File-based CDK: handle user-input schema

### LINKED ISSUE #26759: File CDK: Handle user-input schema config option
Users can input a schema in [name-type mapping format](https://docs.airbyte.com/integrations/sources/s3#user-schema).

The existing `source_files_abstract` [stream._parse_user_input_schema](https://github.com/airbytehq/airbyte/blob/3aa0c284e2631a94e5b2e345fefd70220e15638d/airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/stream.py#L72) method can be used to turn user-input schema into JSON schema format.

#### Acceptance Criteria
- [ ] When a user inputs a schema in name-type mapping format, it is returned in the configured catalog as JSON schema during `discover`.
- [ ] When a user provides the schema, we validate that one record conforms to the schema during the connection check.
- [ ] Records are validated against the user-input schema during `read`; nonconforming records are handled according to error-handling option configured by the user.
- [ ] End to end test coverage for this option for all file types.

clnoll:
## What
Wires through the ability for users to input a schema, in as a name-type mapping.

## How
Allows the user to input a schema in name-type mapping format, as a comma-separated or JSON string.

We then validate the input schema and turn it into a JSON Schema at `FileBasedStreamConfig` creation time.

- During `check`, if a user has input a schema, we verify that a record from one file conforms to the schema so that they can modify it if needed.
- During `read`, if a user has input a schema, we attempt to cast the types from CSVs to the type in the schema. On error we keep the value's type a string.

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py`
2. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py`
4. `airbyte-cdk/python/airbyte_cdk/sources/file_based/default_file_based_availability_strategy.py`

Closes https://github.com/airbytehq/airbyte/issues/26759.

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
@@ -4,8 +4,9 @@
 
 import codecs
 from enum import Enum
-from typing import Any, Dict, List, Mapping, Optional, Union
+from typing import Any, List, Mapping, Optional, Union
 
+from airbyte_cdk.sources.file_based.schema_helpers import type_mapping_to_jsonschema
 from pydantic import BaseModel, validator
 
 PrimaryKeyType = Optional[Union[str, List[str], List[List[str]]]]
@@ -65,7 +66,7 @@ class FileBasedStreamConfig(BaseModel):
     file_type: str
     globs: Optional[List[str]]
     validation_policy: str
-    input_schema: Optional[Dict[str, Any]]
+    input_schema: Optional[Union[str, Mapping[str, Any]]]
     primary_key: PrimaryKeyType
     days_to_sync_if_history_is_full: int = 3
     format: Optional[Mapping[str, CsvFormat]]  # this will eventually be a Union once we have more than one format type
@@ -86,3 +87,8 @@ class FileBasedStreamConfig(BaseModel):
                     raise ValueError(f"Format filetype {file_type} is not a supported file type")
                 return {file_type: {key: val for key, val in v.items()}}
         return v
+
+    @validator("input_schema", pre=True)
+    def transform_input_schema(cls, v):
+        if v:
+            return type_mapping_to_jsonschema(v)
--- a/airbyte-cdk/python/airbyte_cdk/sources/file_based/default_file_based_availability_strategy.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/file_based/default_file_based_availability_strategy.py
@@ -10,6 +10,7 @@ from airbyte_cdk.sources import Source
 from airbyte_cdk.sources.file_based.exceptions import CheckAvailabilityError, FileBasedSourceError
 from airbyte_cdk.sources.file_based.file_based_stream_reader import AbstractFileBasedStreamReader
 from airbyte_cdk.sources.file_based.remote_file import RemoteFile
+from airbyte_cdk.sources.file_based.schema_helpers import conforms_to_schema
 from airbyte_cdk.sources.file_based.stream import AbstractFileBasedStream
 from airbyte_cdk.sources.streams.availability_strategy import AvailabilityStrategy
 
@@ -39,7 +40,7 @@ class DefaultFileBasedAvailabilityStrategy(AvailabilityStrategy):
         """
         try:
             files = self._check_list_files(stream)
-            self._check_parse_record(stream, files[0])
+            self._check_parse_record(stream, files[0], logger)
         except CheckAvailabilityError:
             return False, "".join(traceback.format_exc())
 
@@ -59,11 +60,11 @@ class DefaultFileBasedAvailabilityStrategy(AvailabilityStrategy):
 
         return files
 
-    def _check_parse_record(self, stream: AbstractFileBasedStream, file: RemoteFile):
+    def _check_parse_record(self, stream: AbstractFileBasedStream, file: RemoteFile, logger: logging.Logger):
         parser = stream.get_parser(stream.config.file_type)
 
         try:
-            record = next(iter(parser.parse_records(stream.config, file, self.stream_reader)))
+            record = next(iter(parser.parse_records(stream.config, file, self.stream_reader, logger)))
         except StopIteration:
             # The file is empty. We've verified that we can open it, so will
             # consider the connection check successful even though it means
@@ -72,11 +73,11 @@ class DefaultFileBasedAvailabilityStrategy(AvailabilityStrategy):
         except Exception as exc:
             raise CheckAvailabilityError(FileBasedSourceError.ERROR_READING_FILE, stream=stream.name, file=file.uri) from exc
 
-        if stream.config.input_schema:
-            if not stream.record_passes_validation_policy(record):
+        schema = stream.catalog_schema or stream.config.input_schema
+        if schema and stream.validation_policy.validate_schema_before_sync:
+            if not conforms_to_schema(record, schema):
                 raise CheckAvailabilityError(
                     FileBasedSourceError.ERROR_VALIDATING_RECORD,
                     stream=stream.name,
                     file=file.uri,
-                    validation_policy=stream.config.validation_policy,
                 )
--- a/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
@@ -3,12 +3,17 @@
 #
 
 import csv
-from typing import Any, Dict, Iterable
+import json
+import logging
+from distutils.util import strtobool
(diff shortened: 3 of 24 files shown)


airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
```diff
@@ -115,56 +140,58 @@ def test_discover(capsys, tmp_path, json_spec, scenario):
...
+        elif "state" in actual:
+            assert actual["state"]["data"] == expected
+
+    if scenario.expected_logs:
+        expected_logs = scenario.expected_logs.get("read", [])
+        assert len(logs) == len(expected_logs)
+        _verify_expected_logs(logs, expected_logs)
+
+
+def _verify_expected_logs(logs: List[Dict[str, Any]], expected_logs: List[Dict[str, Any]]):
```
clnoll, line 189: I probably added a bit more validation of logs than we really need. I was using it in my own manual testing, but if it feels too brittle we can take it out.

airbyte-cdk/python/airbyte_cdk/sources/file_based/default_file_based_availability_strategy.py
```diff
@@ -72,11 +73,16 @@ def _check_parse_record(self, stream: AbstractFileBasedStream, file: RemoteFile)
...
+            if not conforms_to_schema(record, stream.config.input_schema):
                 raise CheckAvailabilityError(
                     FileBasedSourceError.ERROR_VALIDATING_RECORD,
                     stream=stream.name,
                     file=file.uri,
-                    validation_policy=stream.config.validation_policy,
                 )
+
+
+def _is_read_request(stream: AbstractFileBasedStream) -> bool:
```
clnoll, line 87: Without this, we end up skipping the stream during `read` because it doesn't pass the availability check. I don't love the way that I'm handling it here though so am open to suggestions.

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
```diff
@@ -68,7 +70,7 @@ class FileBasedStreamConfig(BaseModel):
     validation_policy: Union[str, Any]
     validation_policies: Dict[str, Any]
     catalog_schema: Optional[ConfiguredAirbyteCatalog]
-    input_schema: Optional[Dict[str, Any]]
+    input_schema: Optional[Union[str, Mapping[str, Any]]]
```
girarda, line 73: shouldn't this always get set to an Optional[Mapping] by `set_input_schema`?

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
```diff
@@ -67,8 +70,75 @@ def parse_records(
...
+                yield cast_types(row, property_types)
+
+
+def cast_types(row: Dict[str, str], property_types: Dict[str, Any]) -> Dict[str, Any]:
+    """
+    Casts the values in the input 'row' dictionary according to the types defined in the JSON schema.
+
+    Array and object types are only handled if they can be deserialized as JSON.
+
+    If any errors are encountered, the value will be emitted as a string.
```
girarda, line 103: shouldn't we fail if we cannot cast the values to the expected type?

I'd at least expect a warning

airbyte-cdk/python/airbyte_cdk/sources/file_based/default_file_based_availability_strategy.py
girarda: it's not super clear to me why we don't want to verify the schema. Is it to allow us to process the file even if the first record is not conforming?

~I think it's worth adding a comment explaining why the presence of the catalog_schema means we're doing a read~
Can we move this method to AbstractFileBasedStream? The fact that the presence of a catalog_schema means we're reading seems like an implementation detail

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/user_input_schema_scenarios.py
```diff
@@ -0,0 +1,675 @@
...
+            ]
+        }
+    )
+    .set_expected_records(
+        [
+            {"data": {"col1": "val11a", "col2": 21, "_ab_source_file_last_modified": "2023-06-05T03:54:07Z",
+                      "_ab_source_file_url": "a.csv"}, "stream": "stream1"},
+            {"data": {"col1": "val12a", "col2": 22, "_ab_source_file_last_modified": "2023-06-05T03:54:07Z",
+                      "_ab_source_file_url": "a.csv"}, "stream": "stream1"},
+            # The files in b.csv are emitted despite having an invalid schema
```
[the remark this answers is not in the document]
clnoll: Yes I think as long as we're explicit that it's undefined it should be an option. We are not offering to strip individual fields out of the records for users (which would be redundant with what the platform is doing quite soon, if not already), so users may choose this if they don't want to skip entire records on schema mismatch, or if they know that they have a lot of files, or have deleted some files, in case waiting for rediscovery would not be recommended.

airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py
```diff
@@ -31,6 +30,29 @@ def __lt__(self, other):
...
+    "boolean": "boolean",
+    "float": "number",
+    "integer": "integer",
+    "number": "number",
+    "object": "object",
+    "string": "string",
+}
+
+
+JSON_TO_PYTHON_TYPE_MAPPING = {
```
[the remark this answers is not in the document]
clnoll: 👍 done.

airbyte-cdk/python/airbyte_cdk/sources/file_based/default_file_based_availability_strategy.py
girarda: The interface's doc says
> If boolean is true, then the stream is available, and no str is required. Otherwise, the stream is unavailable for some reason and the str should describe what went wrong and how to resolve the unavailability, if possible.

Can we use a different object to discard individual files during the read?

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
clnoll: Actually on second thought failing goes against the spirit of the validation policy. Logging a warning instead.

airbyte-cdk/python/airbyte_cdk/sources/file_based/default_file_based_availability_strategy.py
clnoll: To summarize the offline sync: for the `emit_records` and `skip_records` validation policies, we're not going to check a record against the input schema, even during the initial `check`, because if users have those options configured their syncs can continue even if the schema changes. However, on the `wait` policy, we will check that a record conforms to the schema during the stream availability check; if it fails, we won't start syncing that stream (given that a record didn't conform to the schema, the sync would have been halted anyway).

Regarding the `wait` case, I felt that it was more consistent to check the record regardless of whether the user input a schema, or we were using the schema from the catalog. The behavior should be consistent between these two cases.

airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_helpers.py
```diff
@@ -170,8 +183,51 @@ def conforms_to_schema(record: Mapping[str, Any], schema: Mapping[str, str]) ->
...
+        - is a key:value map
+        - all values in the map correspond to a JsonSchema datatype
     """
-    ...
+    if not input_schema:
+        return None
+
+    result_schema = {}
+
+    json_mapping = _parse_json_input(input_schema)
```
girarda, line 213: json_mapping can return `None`. We should either null check its output or have it return an empty dict instead

girarda (approved):
awesome! 🎸

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
sherifnada: Is there benefit in making this `Mapping[str, ColumnType]` where `ColumnType` is a known string enum? This would make the UI more structured when the user is inputting these mappings, so there is no chance for error

clnoll: We don't currently have validation happening right in the UI, this is just a form field with string input, with validation happening on the back end.

PR #28063, opened 2023-07-07
File-based CDK: implement schemaless option

### LINKED ISSUE #27013: File CDK: Support CSV "schemaless" option
When this option is selected, the schema emitted by `get_json_schema` will be `{"data": "object"}`, and records will be emitted in the format `{"data": {"col1": val1, …}}`.

#### Acceptance Criteria
- [ ] This option is supported for CSV file type.
- [ ] `read_records` is updated to skip record validation if this option is selected.

clnoll:
## What
Handle the "schemaless" config option.

This option allows users to emit records as JSON in a single "data" column.

## How
- Return the "schemaless" schema on `discover`.
- Once records have been parsed by the appropriate parser, wrap them in a data key - `{"data": <record>}`

Closes https://github.com/airbytehq/airbyte/issues/27013.

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
@@ -72,6 +72,7 @@ class FileBasedStreamConfig(BaseModel):
     max_history_size: Optional[int]
     days_to_sync_if_history_is_full: Optional[int]
     format: Optional[Mapping[str, CsvFormat]]  # this will eventually be a Union once we have more than one format type
+    schemaless: bool = False
 
     @validator("file_type", pre=True)
     def validate_file_type(cls, v):
--- a/airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
@@ -18,7 +18,7 @@ from airbyte_cdk.sources.file_based.exceptions import (
     StopSyncPerValidationPolicy,
 )
 from airbyte_cdk.sources.file_based.remote_file import RemoteFile
-from airbyte_cdk.sources.file_based.schema_helpers import merge_schemas
+from airbyte_cdk.sources.file_based.schema_helpers import merge_schemas, schemaless_schema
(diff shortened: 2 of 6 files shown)


airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
```diff
@@ -73,6 +73,7 @@ class FileBasedStreamConfig(BaseModel):
     max_history_size: Optional[int]
     days_to_sync_if_history_is_full: Optional[int]
     format: Optional[Mapping[str, CsvFormat]]  # this will eventually be a Union once we have more than one format type
+    schemaless: bool = False
```
girarda, line 76: Are we going to add all options for all file types to the `FileBasedStreamConfig`?

We could avoid that by adding a parsing config which could be `Union[CsvConfig, JsonConfig, ParquetConfig]`

The main benefit is that it would be clearer in which context each field is useful (schemaless isn't useful if the format is csv or parquet).

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -148,6 +149,8 @@ def get_json_schema(self) -> JsonSchema:
     def _get_raw_json_schema(self) -> JsonSchema:
         if self.config.input_schema:
             schema = self.config.input_schema
+        elif self.config.schemaless:
```
girarda, line 152: is it possible for both input_schema and schemaless to be set at the same time? It's not obvious to me that the input_schema should take precedence. Can we fail or at least log a warning that this is unexpected?

girarda (approved):
2 minor comments. lgtm!

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
brianjlai: I'm curious about this too. if this is the only case it might be fine as is, but if we are starting to see file type specific configs, we might need this file_type_name -> configs mapping. Similar to how we are preparing for the possibility of multiple file types per stream

brianjlai (approved):
same discussion point as alex on the shape of the config. but the implemention lgtm

PR #28064, opened 2023-07-07
Implement parquet parser

girarda:
## What
* Implement parser for parquet files
* The parser cannot be configured. We can add config options if needed

## How
* The parser gets the file schema by reading it's metadata (no need to read the whole file)
* When parsing records, the parser iterate over each rows, and converts the data into an appropriate format if necessary (eg. format datetime objects as iso)
* The tests use a `TemporaryParquetFilesStreamReader` to write temporary parquet files before returning their output stream. It's possible there's a way to directly create the stream in memory, but this feels safer as there's no risk of creating a corrupt parquet file.

## Do we need to read the whole file to get its schema?
No, the schema of a parquet file is in the file metadata, which can be read without reading the whole file. [This page describes the parquet file format](https://parquet.apache.org/docs/file-format/).

I've also confirmed this by looking at the source code, [pyarrow doesn't try to read the full file](https://github.com/apache/arrow/blob/main/python/pyarrow/_parquet.pyx#L1168-L1243).

And empirically by printing the schema of two files and comparing the impact on memory (there was none).
Profiler output when printing the schema of a 2.2KB file:
```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    53 117.6562 MiB 117.6562 MiB           1   @profile(precision=4)
    54                                         def foo(file_name):
    55 117.6562 MiB   0.0000 MiB           1       print("world")
    56 118.9531 MiB   1.2969 MiB           1       file = pq.ParquetFile(file_name)
    57 119.0781 MiB   0.1250 MiB           1       print(file.schema_arrow)
```

Profiler output when printing the schema of a 7.5GB file:
```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    53 115.6719 MiB 115.6719 MiB           1   @profile(precision=4)
    54                                         def foo(file_name):
    55 115.6719 MiB   0.0000 MiB           1       print("world")
    56 117.0312 MiB   1.3594 MiB           1       file = pq.ParquetFile(file_name)
    57 117.1406 MiB   0.1094 MiB           1       print(file.schema_arrow)
```

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/parquet_parser.py`
2. `airbyte-cdk/python/unit_tests/sources/file_based/file_types/test_parquet_parser.py`
3. `airbyte-cdk/python/unit_tests/sources/file_based/scenarios/parquet_scenarios.py`
4. `airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py`
5. `airbyte-cdk/python/unit_tests/sources/file_based/in_memory_files_source.py`
6. `airbyte-cdk/python/setup.py`


airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
```diff
@@ -139,19 +152,23 @@ def run_test_read_full_refresh(capsys, tmp_path, scenario):
         assert_expected_logs_match_output(logs, expected_logs)
 
 
-def assert_expected_records_match_output(output: List[Dict[str, Any]], expected_output: List[Dict[str, Any]]):
+def assert_expected_records_match_output(output: List[Mapping[str, Any]], expected_output: List[Mapping[str, Any]]) -> None:
     for actual, expected in zip(output, expected_output):
-        assert actual["record"]["data"] == expected["data"]
+        for key, value in actual["record"]["data"].items():
+            if isinstance(value, float):
+                assert math.isclose(value, expected["data"][key], abs_tol=1e-06)
```
girarda, line 159: computers are bad at comparing floats :(

airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
```diff
@@ -87,16 +95,20 @@
...
     empty_schema_inference_scenario,
+    single_parquet_scenario,
+    multi_parquet_scenario,
+    parquet_various_types_scenario,
 ]
 
 
 @pytest.mark.parametrize("scenario", discover_scenarios, ids=[s.name for s in discover_scenarios])
-def test_discover(capsys, tmp_path, json_spec, scenario):
+def test_discover(capsys: CaptureFixture[str], tmp_path: PosixPath, json_spec: Mapping[str, Any], scenario: TestScenario) -> None:
```
girarda, line 105: added these type annotations as part of https://github.com/airbytehq/airbyte/pull/27790

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/parquet_parser.py
```diff
@@ -2,21 +2,130 @@
...
+    def _is_integer(parquet_type: pa.DataType) -> bool:
+        return bool(pa.types.is_integer(parquet_type) or pa.types.is_duration(parquet_type))
+
+    @staticmethod
+    def _is_string(parquet_type: pa.DataType) -> bool:
+        return bool(
+            pa.types.is_time(parquet_type)
+            or pa.types.is_string(parquet_type)
+            or pa.types.is_large_string(parquet_type)
+            or pa.types.is_decimal(parquet_type)  # Return as a string to ensure no precision is lost
```
clnoll, line 121: Looks like this is different than what we do with the [current S3 connector](https://github.com/airbytehq/airbyte/blob/a47fcf174b8a63e80dd445369e5fd7f74924c567/airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/formats/parquet_parser.py#L26). I'm tempted to keep it consistent so we don't break the existing connector when we switch over, but also don't want to be short-sighted by failing to change something if it's an improvement, just for the sake of backwards-compatibility.

I'm leaning toward keeping it like this for now, but identifying the S3 connectors that are currently syncing parquet files with decimal types, so we know what the impact will be (this can be done as part of the ticket for adapting the new file-based CDK to handle existing S3 connector configs). Then we can either alter the behavior depending on whether the connector is using the old config or new (e.g. old configs will still cast to number), or have affected customers do a full refresh. Thoughts?

airbyte-cdk/python/unit_tests/sources/file_based/file_types/test_parquet_parser.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@pytest.mark.parametrize(
+    "parquet_type, expected_type",
+    [
+        pytest.param(pa.bool_(), {"type": "boolean"}, id="test_parquet_bool"),
+        pytest.param(pa.int8(), {"type": "integer"}, id="test_parquet_int8"),
+        pytest.param(pa.int16(), {"type": "integer"}, id="test_parquet_int16"),
+        pytest.param(pa.int32(), {"type": "integer"}, id="test_parquet_int32"),
+        pytest.param(pa.uint8(), {"type": "integer"}, id="test_parquet_uint8"),
```
clnoll, line 22: Should this and the other test include `null`, `int64` and `uint64`, `fixed_size_binary`, and/or `map`?

clnoll (approved):
Looks great @girarda! Thank you for looking into the memory usage.

Just a couple of questions about test coverage for a few parquet types, and one inconsistency with the existing S3 connector.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/parquet_parser.py
girarda: I think the current implementation is problematic. Converting decimals to floats can lead to loss of precision. I don't have a good way of checking how frequently users use this field since the schema in the config only defines the Airbyte type used (number).

I can add an optional flag to let the user configure how these fields should be represented to avoid introducing a breaking change

airbyte-cdk/python/unit_tests/sources/file_based/file_types/test_parquet_parser.py
girarda: yes, done

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/parquet_parser.py
```diff
@@ -2,21 +2,130 @@
...
+            or ParquetParser._is_binary(parquet_type)  # Best we can do is return as a string since we do not support binary
+        )
+
+    @staticmethod
+    def _is_object(parquet_type: pa.DataType) -> bool:
+        return bool(pa.types.is_dictionary(parquet_type) or pa.types.is_struct(parquet_type))
```
brianjlai, line 127: I noticed that most of these `_is...` method are all cast/wrapped into a boolean. Why do we need to cast these statements? I would think they already are booleanss

brianjlai (approved):
looks good to me, the test scenarios and coverage look great!

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/parquet_parser.py
girarda: mypy doesn't know about pyarrow so it doesn't know the method returns a bool. alternatives include
1.  add `# type: ignore` the lines so it doesn't complain
2. install stubs describing the pyarrow types

(2) sounds pretty involved for very little benefit since the pyarrow classes are only used here. I don't have a strong opinion between casting and telling mypy to ignore the "type error"

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/parquet_parser.py
girarda: @clnoll I'll add the flag in a separate PR to avoid the need to re-review this PR.

PR #28067, opened 2023-07-08
Publish metadata on prerelease

### LINKED ISSUE #27854: Publish metadata and spec on prerelease
## Problem
We have to remember to upload the spec of a prerelease when using the image in a LD connector override

## Solution
1. On prerelease Upload metadata to the metadata service bucket
2. Upload the spec to the spec cache

Once these are uploaded the orchestrator will naturally turn them into registry entries which can be used when applying LD overrides

bnchrch:
## What
Publish metadata file and spec for prerelease

closes #27854


airbyte-ci/connectors/metadata_service/lib/metadata_service/commands.py
```diff
@@ -34,10 +34,11 @@ def validate(file_path: pathlib.Path):
 @metadata_service.command(help="Upload a metadata YAML file to a GCS bucket.")
 @click.argument("metadata-file-path", type=click.Path(exists=True, path_type=pathlib.Path))
 @click.argument("bucket-name", type=click.STRING)
-def upload(metadata_file_path: pathlib.Path, bucket_name: str):
+@click.option("--prerelease", type=click.STRING, required=False, default=None, help="The prerelease tag of the connector.")
+def upload(metadata_file_path: pathlib.Path, bucket_name: str, prerelease: str):
```
evantahler, line 38: oh! now we get to pick our pre-release tag, or is this just `--prerelease` signaling that this is a prerelease action?  I think I'm reading this as 
* `--prerelease` (no value) means this is a pre-release and to make a default tag from the git sha
* `--prerelease foo` means to make a docker image tagged `dev-foo` (the dev is always there)

tools/ci_connector_ops/ci_connector_ops/pipelines/commands/groups/connectors.py
```diff
@@ -365,13 +364,6 @@ def publish(
     return all(context.state is ContextState.SUCCESSFUL for context in publish_connector_contexts)
 
 
-def validate_publish_options(pre_release: bool, context_object: Dict[str, Any]):
-    """Validate that the publish options are set correctly."""
-    for k in ["spec_cache_bucket_name", "spec_cache_gcs_credentials", "metadata_service_bucket_name", "metadata_service_gcs_credentials"]:
-        if not pre_release and context_object.get(k) is None:
-            click.Abort(f'The --{k.replace("_", "-")} option is required when running a main release publish pipeline.')
```
erohmensing, line 372: I think [this code](https://github.com/airbytehq/airbyte/pull/28067/files?diff=split&w=0#diff-956ec778a0cbdb8fbf60e4327fc8b80c3c8f06739870ad4f56c5e822203464bfR236-R263) needs to be updated to set it to required before removing this validation

airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
```diff
@@ -86,7 +90,46 @@ def upload_file_if_changed(
...
+
+def _prerelease_upload(metadata: ConnectorMetadataDefinitionV0, bucket: storage.bucket.Bucket, prerelease_tag: str) -> Tuple[bool, str]:
+    # replace any dockerImageTag references with the actual tag
+    # this includes metadata.data.dockerImageTag, metadata.data.registries[].dockerImageTag
+    # where registries is a dictionary of registry name to registry object
+    metadata_dict = to_json_sanitized_dict(metadata, exclude_none=True)
+    metadata_dict["data"]["dockerImageTag"] = prerelease_tag
+    for registry in get(metadata_dict, "data.registries", {}).values():
+        if "dockerImageTag" in registry:
+            registry["dockerImageTag"] = prerelease_tag
```
erohmensing, line 120: This feels a bit gnarly - I'm not sure about the overrides part?

Not that they matter because this isn't going into a registry anyway, but I don't think publishing a pre-release inherently makes any changes to the overrides. But you might be considering only the no overrides case and I might be considering only the overrides case

`metadata_dict["data"]["dockerImageTag"]` will definitely have to be up to date when we grab the metadata from the platform and persist the definition based off of it, if we go that route, so good catch there!

airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
```diff
@@ -86,7 +90,46 @@ def upload_file_if_changed(
...
+
+    # write metadata to yaml file in system tmp folder
+    tmp_metadata_file_path = Path("/tmp") / metadata.data.dockerRepository / prerelease_tag / METADATA_FILE_NAME
+    tmp_metadata_file_path.parent.mkdir(parents=True, exist_ok=True)
+    with open(tmp_metadata_file_path, "w") as f:
+        yaml.dump(metadata_dict, f)
+
+    prerelease_remote_path = get_metadata_remote_file_path(metadata.data.dockerRepository, prerelease_tag)
+    return upload_file_if_changed(tmp_metadata_file_path, bucket, prerelease_remote_path, disable_cache=True)
```
erohmensing, line 129: why disable_cache, and only on prereleases?

airbyte-ci/connectors/metadata_service/lib/tests/test_gcs_upload.py
```diff
@@ -155,3 +156,41 @@ def test_upload_metadata_to_gcs_invalid_docker_images(mocker, invalid_metadata_u
...
+        metadata, error = gcs_upload.validate_and_load(metadata_file_path, [])
+        expected_tmp_metadata_file_path = pathlib.Path("/tmp") / metadata.data.dockerRepository / prerelease_image_tag / METADATA_FILE_NAME
+        assert expected_tmp_metadata_file_path.exists(), f"{expected_tmp_metadata_file_path} does not exist"
+
+        # verify that the metadata is overrode
+        tmp_metadata, error = gcs_upload.validate_and_load(expected_tmp_metadata_file_path, [])
+        tmp_metadata_dict = to_json_sanitized_dict(tmp_metadata, exclude_none=True)
+        for registry in get(tmp_metadata_dict, "data.registries", {}).values():
+            if "dockerImageTag" in registry:
+                assert registry["dockerImageTag"] == prerelease_image_tag
```
erohmensing, line 196: let's validate the main `dockerImageTag` as well since that one is most likely to be used

airbyte-ci/connectors/metadata_service/lib/tests/test_gcs_upload.py
```diff
@@ -155,3 +156,41 @@ def test_upload_metadata_to_gcs_invalid_docker_images(mocker, invalid_metadata_u
...
+            metadata_file_path,
+            prerelease_image_tag,
+        )
+
+        gcs_upload._latest_upload.assert_not_called()
+
+        # Assert that _prerelease_upload is called and the third argument is prerelease_image_tag
+        # Ignore the first and second arguments (_ and __ are often used as placeholder variable names in Python when you don't care about the value)
+        _, __, prerelease_tag = gcs_upload._prerelease_upload.call_args[0]
+        assert prerelease_tag == prerelease_image_tag
```
erohmensing, line 184: Shall we add a similar "we called the right private method" assertion to the call to main upload [here](https://github.com/airbytehq/airbyte/pull/28067/files?diff=split&w=0#diff-8d2e2dfdaa33aa34583293472ad061493cf88a0ac2955e8bfe0cdc7480de10d1R100)?

erohmensing (commented):
Added some comments/questions - my only actual concern is setting the previously non-required `# Not required for pre-release pipelines, downstream validation happens for main release pipelines` to required. After that, happy to ✅

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -255,11 +254,11 @@ def create_connector_report(results: List[StepResult]) -> ConnectorReport:
 
             # If the connector image already exists, we don't need to build it, but we still need to upload the metadata file.
             # We also need to upload the spec to the spec cache bucket.
-            if check_connector_image_results.status is StepStatus.SKIPPED and not context.pre_release:
+            if check_connector_image_results.status is StepStatus.SKIPPED:
                 context.logger.info(
                     "The connector version is already published. Let's upload metadata.yaml and spec to GCS even if no version bump happened."
                 )
```
erohmensing, line 260: Out of scope of this PR - i'm not sure we want to do this for spec? The spec is versioned info, we shouldn't be changing it without versioning

airbyte-ci/connectors/metadata_service/lib/metadata_service/commands.py
bnchrch: Ella you got it!

This is just for the metadata_service cli (fun fact that this is a cli tool installed and use by the `airbyte-ci` tool)

So for the metadata service you can upload a given metadata file under a prerelease location

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
erohmensing: > a prerelease can have changes to spec

Are we checking the right "image exists"? the prerelease image shouldn't already exist, which would make us upload the spec

airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
bnchrch: This is the tricky part!

So for example if you are the platform and say you are using the LD flag to override the connector image to `airbyte/source-postgres:1.1.1-dev.123asd` you would go looking for that info at either

https://connectors.airbyte.com/files/metadata/airbyte/source-postgres/1.1.1-dev.123asd/oss.json

or

https://connectors.airbyte.com/files/metadata/airbyte/source-postgres/1.1.1-dev.123asd/cloud.json

Now if the original metadata had `registries.cloud.dockerImageTag` pinned to 1.1.0
and we didnt overwrite `registries.cloud.dockerImageTag` with the prerelease tag
then the the cloud use would still get `airbyte/source-postgres:1.1.1`
because thats what the registry entry would have

airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
bnchrch: I had disabled the cache here since I figured metadata might change more rapidly. But youre right, we should disable it all the time

tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py
```diff
@@ -525,12 +533,21 @@ def metadata_service_gcs_credentials_secret(self) -> Secret:
...
+            return metadata_tag
+
     @property
     def docker_image_name(self):
         if self.pre_release:
-            return f"{self.docker_image_from_metadata}-dev.{self.git_revision[:10]}"
+            return f"{self.docker_image}-dev.{self.git_revision[:10]}"
         else:
-            return self.docker_image_from_metadata
+            return self.docker_image
```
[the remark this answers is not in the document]
bnchrch: Ah yes, When will I learn not to rush out a PR . Will do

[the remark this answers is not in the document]
erohmensing: :D yesterday i checked in a test file, lol

airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
erohmensing: ahaaaa

that makes sense! 

Maybe a comment as to "if we request the registry entry for a prerelease version specifically, we want to make sure we do not get a different overridden version" or something

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
bnchrch: Forgot to respond here. 

Given that these run off of `context.docker_image` and docker_image is now properly receiving the [docker_image_tag override for prerelease](https://github.com/airbytehq/airbyte/blob/bnchrch/27854/metadata-prerelease/tools/ci_connector_ops/ci_connector_ops/pipelines/contexts.py#L537) this will check if the prerelease tag is published.

Generally its redundant as it should be a new one everytime as the commit is updated.

But this whole function is a bit confusing, updating that now

bnchrch: Im going to do this refactor in a different PR.

erohmensing (approved):
💯 💯

bnchrch:
Will hold off merging until pipelines settle down

PR #28097, opened 2023-07-10
Refactor config to not store validation_policies on the config itself

### LINKED ISSUE #28101: [file-based cdk] Refactor validation_policies to not be stored on the config object itself
As part of another ticket, I noticed that the way we have started storing the `validation_policies` on the config object itself seems like a pattern we want to avoid. Validation policies are not user configured behavior and are instead an aspect of the specific file based source's implementation and it should be persisted on the source and stream. Putting them on the source is misleading and confusing because now they will be reflected as part of the `spec` command which generates a schema from the `FileBasedStreamConfig`.

## Acceptance Criteria
- [ ] The `FileBasedStreamConfig` class no longer has the `validation_policies` field
- [ ] The correct validation policy implementation  is resolved from the config's `validation_policy` field and used when validating records

brianjlai:
Closes https://github.com/airbytehq/airbyte/issues/28101

## What
When I was writing up the changes to support file-based spec operations, I started to observe some undesirable behavior in our config interface that was being surfaced now that we were validating user configs against the spec json schema. A snippet from a slack convo:

> I’m wondering if we should actually remove the @ root_validator, that we had originally added in. It feels like a bit of a code smell that we have to:
manipulate the validation_policy and assign it to an instance in the pydantic validator
because we’re doing 1, we also end up loosening the types on FileBasedStreamConfig to str or Any which isn’t quite representative. From the user spec point of view, it should always be a str

## How
These changes should have no functional impact, but it refactors how we ensure the config's validation policy is valid and how we retrieve the correct policy implementation. And we do it without appending additional fields into the `FileBasedStreamConfig` which is used to generate the spec. The gist of the refactor:

* Remove the root validation that checks if the validation policy is in validation policies
* In FileBasedSource, we have an custom check that the config’s policy is in the source’s validation_policies
* We now also pass validation_policies to the stream obj.
* AbstractFileBasedStream.record_passes_validation_policy() now has to resolve the config’s validation_policy, and then invokes it

## Recommended reading order
1. `file_based_stream_config.py`
2. `file_based_source.py`
3. `file_based_stream.py`

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
@@ -7,7 +7,7 @@ from enum import Enum
 from typing import Any, Dict, List, Mapping, Optional, Union
 
 from airbyte_cdk.models import ConfiguredAirbyteCatalog
-from pydantic import BaseModel, root_validator, validator
+from pydantic import BaseModel, validator
 
 PrimaryKeyType = Optional[Union[str, List[str], List[List[str]]]]
 
@@ -66,7 +66,6 @@ class FileBasedStreamConfig(BaseModel):
     file_type: str
     globs: Optional[List[str]]
     validation_policy: Union[str, Any]
-    validation_policies: Dict[str, Any]
     catalog_schema: Optional[ConfiguredAirbyteCatalog]
     input_schema: Optional[Dict[str, Any]]
     primary_key: PrimaryKeyType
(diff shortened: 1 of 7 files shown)


airbyte-cdk/python/unit_tests/sources/file_based/scenarios/check_scenarios.py
```diff
@@ -174,8 +174,8 @@
             ],
         }
     )
-    .set_validation_policies(FailingSchemaValidationPolicy)
-    .set_expected_check_error(ConfigValidationError, FileBasedSourceError.ERROR_VALIDATING_RECORD)
+    .set_validation_policies({FailingSchemaValidationPolicy.ALWAYS_FAIL:  FailingSchemaValidationPolicy()})
+    .set_expected_check_error(None, FileBasedSourceError.ERROR_VALIDATING_RECORD)
```
brianjlai, line 178: This was actually a buggy test. Validation policies should be a mapping of name -> policy. But we passed in just the policy so this actually ended up failing the pydantic parse. When we correct the inputs, we actually want to verify that the `check` fails due to the record failing validation. And in this case its not an explicit exception, its an error message in the check response failure

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
```diff
@@ -66,31 +66,25 @@ class FileBasedStreamConfig(BaseModel):
...
+        return v
+
     @validator("format", pre=True)
     def transform_format(cls, v):
         if isinstance(v, Mapping):
             file_type = v.get("filetype", "")
-            if file_type.casefold() not in VALID_FILE_TYPES:
-                raise ValueError(f"Format filetype {file_type} is not a supported file type")
-            return {file_type: {key: val for key, val in v.items()}}
+            if file_type:
```
brianjlai, line 86: fixed a small bug. The new way we present formats would always fail since the empty string is not a valid file type. We should only perform the check if we know this is the legacy format, delineated by the `filetype` being present

airbyte-cdk/python/unit_tests/sources/file_based/config/test_file_based_stream_config.py
```diff
@@ -27,14 +26,53 @@ def test_csv_config(file_type, input_format, expected_format, expected_error):
...
             FileBasedStreamConfig(**stream_config)
     else:
         actual_config = FileBasedStreamConfig(**stream_config)
         assert not hasattr(actual_config.format[file_type], "filetype")
         for expected_format_field, expected_format_value in expected_format.items():
+            assert isinstance(actual_config.format[file_type], CsvFormat)
             assert getattr(actual_config.format[file_type], expected_format_field) == expected_format_value
+
+
+def test_legacy_format():
```
clnoll, line 45: Nice!

clnoll (approved):
Looks great @brianjlai!

PR #28139, opened 2023-07-11
[file-based cdk] file based spec boilerplate backed by pydantic models

brianjlai:
## What
Adds the boilerplate that allows for file-based sources to implement the AbstractFileBasedSpec class which provides the baseline functionality common to all connectors.

The changes to the InMemoryFilesSource acts as an example of how a new source might implement the new interface.

## How
The main thing to emphasize in the PR is that Pydantic models that we have been using to parse user configs into are oriented to also used to generate the JSON schema that is returned as part of the `spec` operation and used to perform validations at the `entrypoint.py`.

The two components that a new file-based source can implement are the `documentation_url()` method and and new Pydantic fields, both of which will be returned as part of `spec` or used to validate.

## Recommended reading order
1. `abstract_file_based_spec.py`
2. `file_based_source.py`
3. `in_memory_files_source.py`
4. `csv_scenarios.py`


airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -29,12 +31,14 @@ def __init__(
...
         catalog: Optional[ConfiguredAirbyteCatalog],
         availability_strategy: Optional[AvailabilityStrategy],
+        spec_class: Type[AbstractFileBasedSpec],
         discovery_policy: AbstractDiscoveryPolicy = DefaultDiscoveryPolicy(),
         parsers: Dict[str, FileTypeParser] = None,
         validation_policies: Dict[str, AbstractSchemaValidationPolicy] = DEFAULT_SCHEMA_VALIDATION_POLICIES,
     ):
         self.stream_reader = stream_reader
         self.availability_strategy = availability_strategy or DefaultFileBasedAvailabilityStrategy(stream_reader)
+        self.spec_class = spec_class
```
brianjlai, line 41: i made this required, there _could_ be a case for this spec object being optional, but given that every source should at least verify a documentation url and will almost always require at least one unique config option, my gut tells me this will always need to be implemented

airbyte-cdk/python/unit_tests/sources/file_based/conftest.py
```diff
@@ -1,35 +0,0 @@
...
-import pytest
-
-CONNECTION_SPECIFICATION = {
-    "type": "object",
-    "additionalProperties": True,
-}
-
-
-@pytest.fixture
-def json_spec():
```
brianjlai, line 18: Originally the conftest.py was needed to define this extra test fixture for defining the spec.json for the test scenarios. However, this is no longer necessary since the spec should always be generated from the file-based source’s implementation and not a static file.

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
```diff
@@ -0,0 +1,69 @@
...
+    Used during spec; allows the developer to configure the cloud provider specific options
+    that are needed when users configure a file-based source.
+    """
+
+    streams: List[FileBasedStreamConfig] = Field(
+        title="The list of streams to sync",
+        description="Streams defines the behavior for grouping files together that will be synced to the downstream destination. Each "
+        "stream has it own independent configuration to handle which files to sync, how files should be parsed, and the "
+        "validation of records against the schema.",
+        order=10,
```
girarda, line 26: is order=10 to ensure this is the last field?

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -34,6 +35,158 @@
...
+                        "type": "array",
+                        "items": {
+                            "title": "FileBasedStreamConfig",
+                            "type": "object",
+                            "properties": {
+                                "name": {
+                                    "title": "Name",
+                                    "type": "string"
+                                },
+                                "file_type": {
```
girarda, line 62: the file_type is part of the stream config. Does this mean the source can support multiple file types?

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -34,6 +35,158 @@
...
+                                    "type": "array",
+                                    "items": {
+                                        "type": "string"
+                                    }
+                                },
+                                "validation_policy": {
+                                    "title": "Validation Policy",
+                                    "type": "string"
+                                },
+                                "catalog_schema": {
```
girarda, line 77: why is the configured catalog part of the spec?

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -34,6 +35,158 @@
...
+                                        }
+                                    ]
+                                },
+                                "max_history_size": {
+                                    "title": "Max History Size",
+                                    "type": "integer"
+                                },
+                                "days_to_sync_if_history_is_full": {
+                                    "title": "Days To Sync If History Is Full",
+                                    "type": "integer"
```
girarda, line 124: nit: I think this should have a default value

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -34,6 +35,158 @@
...
+                                                "items": {
+                                                    "type": "string"
+                                                }
+                                            }
+                                        }
+                                    ]
+                                },
+                                "max_history_size": {
+                                    "title": "Max History Size",
+                                    "type": "integer"
```
girarda, line 120: nit: I think this should have a default value. We might even want to hide the option until it needs to be customized since this is very much an advanced option

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
```diff
@@ -0,0 +1,69 @@
...
+        str_schema = json.dumps(schema)
+        for ref_block in re.findall(r'{"\$ref": "#\/definitions\/.+?(?="})"}', str_schema):
+            ref = json.loads(ref_block)["$ref"]
+            str_schema = str_schema.replace(ref_block, json.dumps(json_schema_ref_resolver.resolve(ref)[1]))
+        pyschema: dict = json.loads(str_schema)
+        del pyschema["definitions"]
+        return pyschema
+
+    @staticmethod
+    def add_legacy_format(schema: dict) -> dict:
```
girarda, line 59: should we test this with a real s3 config?

girarda (approved):
lgtm! left a few nits but nothing blocking

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
brianjlai: yeah, it could go higher, but i just padded it with 10 similar to the existing s3 source where the abstract fields common to all sources start after the source specific ones

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
brianjlai: ah yes good point i can add another scenario in for that

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
```diff
@@ -0,0 +1,69 @@
...
+        Generates the mapping comprised of the config fields
+        """
+        schema = super().schema(*args, **kwargs)
+        resolved_schema = cls.resolve_refs(schema)
+        resolved_schema = cls.add_legacy_format(resolved_schema)
+
+        return resolved_schema
+
+    @staticmethod
+    def resolve_refs(schema: dict) -> dict:
```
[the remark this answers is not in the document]
brianjlai: ah okay yeah i didn't know we were reusing. I think that would make sense since this satisfies the common use case of pydantic models being used as source of a spec being generated. I'll add this to our common utils

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
clnoll: It makes sense to me that each stream could correspond to a different file type.

clnoll (approved):
Awesome @brianjlai! Just a couple of comments.

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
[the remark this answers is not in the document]
brianjlai: scratch that, TIL we actually already have a utility for that here, so we can releverage what we already have https://github.com/airbytehq/airbyte/blob/2e099acc52aeddc172c5ee66ed98426c69449a4a/airbyte-cdk/python/airbyte_cdk/sources/utils/schema_helpers.py#L79-L85

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
brianjlai: correct that would be the intention. Although it is still a single stream can only extract from one file type (although long term we can hopefully set us up to support multiple file types per string)

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
brianjlai: Its here because in the pydantic model for the stream configs the configured catalog has always been there, although @clnoll do you know if this is required for specific functionality. It was in the config since the beginning and never changed this specific field

clnoll: I can't think of a need for it to be in the spec.

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
brianjlai: When you mean hidden, do you mean hidden in that we don't even want it surfaced in the spec, or that we want to add `airbyte_hidden` so the UI knows not to include it?

The latter is doable, but just a bit ugly because we need to do some additional post processing (similar to us removing refs) to append it to the spec message that we return

clnoll: Is there a reason to include it in the spec if we're just going to hide it?

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
brianjlai: fair, i'll add the default value that the cursor uses if this is empty which is 3

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
girarda: I have a preference for removing the field from the config and letting connector developers configure it like Catherine proposed. We can add the field to the spec if/when there is a need for users to configure it

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
```diff
@@ -0,0 +1,69 @@
...
+import re
+from abc import abstractmethod
+from typing import Any, Dict, List
+
+from airbyte_cdk.sources.file_based.config.file_based_stream_config import FileBasedStreamConfig
+from jsonschema import RefResolver
+from pydantic import AnyUrl, BaseModel, Field
+
+
+class AbstractFileBasedSpec(BaseModel):
```
sherifnada, line 15: is there an example spec.json that you can share?

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
```diff
@@ -65,7 +65,7 @@ class FileBasedStreamConfig(BaseModel):
     name: str
     file_type: str
     globs: Optional[List[str]]
-    validation_policy: Union[str, Any]
+    validation_policy: str
     catalog_schema: Optional[ConfiguredAirbyteCatalog]
```
sherifnada, line 69: is this saying the user is expected to input a whole configured catalog?

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
brianjlai: okay sounds good, i can remove it entirely

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
clnoll: The user doesn't input the catalog - it gets set on the stream. I don't see that we're setting or using it on the config object, so we should be able to just remove it from here and the spec. Setting it here is probably just a remnant of one of my PRs, overlooked during a refactor.

airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
```diff
@@ -167,6 +167,16 @@ def run_test_read_incremental(capsys, tmp_path, scenario):
                 assert actual["state"]["data"] == expected
 
 
+spec_scenarios = [
+    csv_multi_stream_scenario,
+    csv_single_stream_scenario,
+]
+
+
+def test_spec(capsys):
```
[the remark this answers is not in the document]
brianjlai: ah yes good catch. done!

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -29,12 +31,14 @@ def __init__(
         stream_reader: AbstractFileBasedStreamReader,
         catalog: Optional[ConfiguredAirbyteCatalog],
         availability_strategy: Optional[AvailabilityStrategy],
+        spec_class: Type[AbstractFileBasedSpec],
```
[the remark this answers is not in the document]
brianjlai: Alex and I paired on this a bit, we weren't able to get something that worked with being able to hit the classmethod. in lieu of spinning our wheels will just leave it as it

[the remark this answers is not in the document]
brianjlai: Alex and I paired on this a bit, we weren't able to get something that worked with being able to hit the classmethod. in lieu of spinning our wheels will just leave it as it

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
brianjlai: An example of what it looks like is in the tests as the expected_spec in `csv_scenarios.py`

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -34,6 +35,146 @@
...
+                                    "title": "Validation Policy",
+                                    "type": "string"
+                                },
+                                "input_schema": {
+                                    "title": "Input Schema",
+                                    "type": "object"
+                                },
+                                "primary_key": {
+                                    "title": "Primary Key",
+                                    "anyOf": [
```
sherifnada, line 88: @brianjlai have you confirmed the UI can display an `anyOf` like this one? 

FWIW I think this should just be a list of strings. Nested PKs are no longer something we want to encourage, and it will likely be removed from the protocol. List of strings ensures it's just a composite PK

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -34,6 +35,146 @@
...
+                                        }
+                                    ]
+                                },
+                                "days_to_sync_if_history_is_full": {
+                                    "title": "Days To Sync If History Is Full",
+                                    "default": 3,
+                                    "type": "integer"
+                                },
+                                "format": {
+                                    "anyOf": [
```
sherifnada, line 115: same question about anyOf

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -34,6 +35,146 @@
...
+                                            "title": "Format",
+                                            "type": "object",
+                                            "additionalProperties": {
+                                                "title": "CsvFormat",
+                                                "type": "object",
+                                                "properties": {
+                                                    "delimiter": {
+                                                        "title": "Delimiter",
+                                                        "default": ",",
+                                                        "type": "string"
```
sherifnada, line 126: I think it's a missed opportunity that this spec does not inherit `description` and `title` fields from the base spec. Each connector should not have to re-explain what this field does when the base class can do a good job of it

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -34,6 +35,146 @@
...
+                                                    "double_quote": {
+                                                        "title": "Double Quote",
+                                                        "type": "boolean"
+                                                    },
+                                                    "quoting_behavior": {
+                                                        "default": "Quote Special Characters",
+                                                        "enum": ["Quote All", "Quote Special Characters", "Quote Non-numeric", "Quote None"]
+                                                    }
+                                                },
+                                                "required": ["double_quote"]
```
sherifnada, line 151: why is this required? Can't we have a default?

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
sherifnada: also in that direction, the titles in a lot of these fields are not easily understood by an end user or look like an abbreviation e.g it should be `Quote Character` instead of `Char` etc..

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -34,6 +35,146 @@
...
+            "documentationUrl": "https://docs.airbyte.com/integrations/sources/in_memory_files",
+            "connectionSpecification": {
+                "title": "InMemorySpec",
+                "description": "Used during spec; allows the developer to configure the cloud provider specific options\nthat are needed "
+                               "when users configure a file-based source.",
+                "type": "object",
+                "properties": {
+                    "streams": {
+                        "title": "The list of streams to sync",
+                        "description": "Streams defines the behavior for grouping files together that will be synced to the downstream "
```
sherifnada, line 49: I think this description of streams is hard to understand for an end user. A suggested alternate phrasing: 

```
Each instance of this configuration defines which defines a <href rel="https://docs.airbyte.com/cloud/core-concepts#stream">stream</href>. Use this to define which files belong in the stream, their format, and how they should be parsed and validated. When sending data to warehouse destination such as Snowflake or BigQuery, each stream is a separate table. 
```

sherifnada (commented):
@brianjlai I added some comments that I think could improve the maintainability and UX of the spec

PR #28145, opened 2023-07-11
Connector builder: Extend compatibility guide

flash1293:
## What

Extends the compatibility guide calling out the limitation with dynamic streams. To avoid requiring Airbyte-specific knowledge at this point, I phrased that as "fixed data endpoints". I hope the example makes it clear.


girarda (approved):

girarda:
lgtm thanks for taking care of it!

PR #28166, opened 2023-07-11
Orchestrator: Break resources into dependency trees

bnchrch:
## What
Theory: We are bundling too many resources on every sensor run, this reduces the resources we load to only what is required

### DIFF
--- a/airbyte-ci/connectors/metadata_service/orchestrator/.env.template
+++ b/airbyte-ci/connectors/metadata_service/orchestrator/.env.template
@@ -1,4 +1,7 @@
-METADATA_BUCKET="ben-ab-test-bucket"
+METADATA_BUCKET="dev-airbyte-cloud-connector-metadata-service"
 CI_REPORT_BUCKET="airbyte-ci-reports"
 GITHUB_METADATA_SERVICE_TOKEN=""
 NIGHTLY_REPORT_SLACK_WEBHOOK_URL=""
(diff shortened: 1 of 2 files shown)


evantahler (approved):
Makes sense... give it a shot!

airbyte-ci/connectors/metadata_service/orchestrator/.env.template
```diff
@@ -1,4 +1,7 @@
-METADATA_BUCKET="ben-ab-test-bucket"
+METADATA_BUCKET="dev-airbyte-cloud-connector-metadata-service"
 CI_REPORT_BUCKET="airbyte-ci-reports"
 GITHUB_METADATA_SERVICE_TOKEN=""
 NIGHTLY_REPORT_SLACK_WEBHOOK_URL=""
+# METADATA_CDN_BASE_URL="https://connectors.airbyte.com/files"
+DOCKER_HUB_USERNAME=""
+DOCKER_HUB_PASSWORD=""
```
erohmensing, line 7: Did you mean to commit these changes?

erohmensing (approved):
Reforestation! 🌲 🌲 🌲

airbyte-ci/connectors/metadata_service/orchestrator/.env.template
bnchrch: Yup!

PR #28229, opened 2023-07-12
📝 Docs: Source S3 documentation update

ChristoGrab:
## What
Updated the setup instructions for source S3, as well as the list of parameters in the S3 Provider and File Format sections.

## How
- Added step-by-step instructions for obtaining S3 credentials and configuring permissions in AWS.
- Merged OSS/Cloud setup instructions since they are functionally the same after log in.
- Added standard first steps for Airbyte UI.
- Updated order of steps to accurately reflect the flow of setup in Airbyte UI.
- Updated/clarified descriptions for all fields in S3 Provider Settings and File Format Settings and added references to any default values.
- Added a couple external links where appropriate.


docs/integrations/sources/s3.md
```diff
@@ -3,46 +3,66 @@
...
+2. In the left navigation bar, click **Sources**. In the top-right corner, click **+ New source**.
+3. Find and select **S3** from the list of available sources.
+4. Enter the **Output Stream Name**. This will be the name of the table in the destination (can contain letters, numbers and underscores).
+5. Enter the **Pattern of files to replicate**. This is a regular expression that allows Airbyte to pattern match the specific files to replicate. If you are replicating all the files within your bucket, use `**` as the pattern. For more precise pattern matching options, refer to the [Path Patterns section](#path-patterns) below.
+6. Enter the name of the **Bucket** containing your files to replicate.
+7. Toggle the **Optional fields** under the **Bucket** field to expand additional configuration options. **If you are syncing from a private bucket**, you must fill the **AWS Access Key ID** and **AWS Secret Access Key** fields with the appropriate credentials to authenticate the connection. All other fields are optional and can be left empty. Refer to the [S3 Provider Settings section](#s3-provider-settings) below for more information on each field.
```
sherifnada, line 63: could you make a PR to the S3 bucket adding the [`always_show: true` annotation ](https://docs.airbyte.com/connector-development/connector-specification-reference/#collapsing-optional-fields)? These fields are pretty common so I think they shouldn't be hidden under the optional fields section

docs/integrations/sources/s3.md
```diff
@@ -152,66 +172,105 @@ Please note, the S3 Source connector used to infer schemas from all the availabl
...
-- `verify_ssl_cert` : Skip ssl validity check in case of using custom servers with self signed certificates. Ignored in case of using Amazon service.
+- **AWS Access Key ID**: One half of the [required credentials](https://docs.aws.amazon.com/general/latest/gr/aws-sec-cred-types.html#access-keys-and-secret-access-keys) for accessing a private bucket.
+- **AWS Secret Access Key**: The other half of the [required credentials](https://docs.aws.amazon.com/general/latest/gr/aws-sec-cred-types.html#access-keys-and-secret-access-keys) for accessing a private bucket.
+- **Path Prefix**: An optional string that limits the files returned by AWS when listing files to only those starting with the specified prefix. This is different than the **Path Pattern**, as the prefix is applied directly to the API call made to S3, rather than being filtered within Airbyte. It does not accept pattern-style symbols like wildcards (`*`). We recommend using this filter if your bucket has many folders and files that are unrelated to this stream, and all the relevant files will always reside under the specified prefix.
```
sherifnada, line 177:
```suggestion
- **Path Prefix**: An optional string that limits the files returned by AWS when listing files to only those starting with the specified prefix. This is different than the **Path Pattern**, as the prefix is applied directly to the API call made to S3, rather than being filtered within Airbyte. **This is not a regular expression** and does not accept pattern-style symbols like wildcards (`*`). We recommend using this filter to improve performance if the connector if your bucket has many folders and files that are unrelated to the data you want to replicate, and all the relevant files will always reside under the specified prefix.
```

docs/integrations/sources/s3.md
```diff
@@ -152,66 +172,105 @@ Please note, the S3 Source connector used to infer schemas from all the availabl
...
+  - Together with the **Path Pattern**, there are multiple ways to specify the files to sync. For example, all the following configurations are equivalent:
+    - **Prefix** = `<empty>`, **Pattern** = `path1/path2/myFolder/**/*`
+    - **Prefix** = `path1/`, **Pattern** = `path2/myFolder/**/*.csv`
+    - **Prefix** = `path1/path2/`, **Pattern** = `myFolder/**/*.csv`
+    - **Prefix** = `path1/path2/myFolder/`, **Pattern** = `**/*.csv`.
+
+  :::note
+  The last combination is the most efficient one, because the directories are filtered earlier in the S3 API call. However, the difference in efficiency is usually negligible.
```
sherifnada, line 185: `However, the difference in efficiency is usually negligible.` This statement is confusing. So as a user, should I do this / is it impactful, or not? 😅 If it's not impactful, why is it here at all? 

if you're not sure I would just remove that statement

docs/integrations/sources/s3.md
```diff
@@ -152,66 +172,105 @@ Please note, the S3 Source connector used to infer schemas from all the availabl
...
+  :::
 
-  **File Format Settings**
+  - The ability to individually configure the prefix and pattern has been included to accommodate as many use cases as possible. If you are unsure of the best approach, you can safely leave the **Path Prefix** field empty and just [set the Path Pattern](#path-patterns) to meet your requirements.
```
sherifnada, line 188:
```suggestion
  - The ability to individually configure the prefix and pattern has been included to accommodate situations where you do not want to replicate the majority of the files in the bucket. If you are unsure of the best approach, you can safely leave the **Path Prefix** field empty and just [set the Path Pattern](#path-patterns) to meet your requirements.
```

docs/integrations/sources/s3.md
```diff
@@ -152,66 +172,105 @@ Please note, the S3 Source connector used to infer schemas from all the availabl
...
-- `block_size` : This is the number of bytes to process in memory at a time while reading files. The default value here is usually fine but if your table is particularly wide \(lots of columns / data in fields is large\) then raising this might solve failures on detecting schema. Since this defines how much data to read into memory, raising this too high could cause Out Of Memory issues so use with caution.
-- `additional_reader_options` : This allows for editing the less commonly required CSV [ConvertOptions](https://arrow.apache.org/docs/python/generated/pyarrow.csv.ConvertOptions.html#pyarrow.csv.ConvertOptions). The value must be a valid JSON string, e.g.:
-
-  ```text
-  {"timestamp_parsers": ["%m/%d/%Y %H:%M", "%Y/%m/%d %H:%M"], "strings_can_be_null": true, "null_values": ["NA", "NULL"]}
+- **Delimiter**: Even though CSV is an acronymn for Comma Separated Values, it is used more generally as a term for flat file data that may or may not be comma separated. The delimiter field lets you specify which character acts as the separator. By default, this value is set to `,`.
```
sherifnada, line 204:
```suggestion
- **Delimiter**: Even though CSV is an acronym for Comma Separated Values, it is used more generally as a term for flat file data that may or may not be comma separated. The delimiter field lets you specify which character acts as the separator. By default, this value is set to `,`.
```

sherifnada: Can you add an example of how one can for example use tab-delimiters?

docs/integrations/sources/s3.md
```diff
@@ -152,66 +172,105 @@ Please note, the S3 Source connector used to infer schemas from all the availabl
...
+- **Infer Datatypes**: This option determines whether a schema for the source should be inferred from the current data. When set to False and a custom schema is provided, the manually enforced schema takes precedence. If no custom schema is set and this option is set to False, all fields will be read as strings. Set to True by default.
+- **Quote Character**: In some cases, data values may contain instances of reserved characters \(like a comma, if that's the delimiter\). CSVs can handle this by wrapping a value in defined quote characters so that on read it can parse it correctly. By default, this is set to `"`.
+- **Escape Character**: An escape character can be used to prefix a reserved character and ensure correct parsing. Leaving this field blank (default option) will disallow escaping.
```
sherifnada, line 207: can you add an example?

docs/integrations/sources/s3.md
```diff
@@ -152,66 +172,105 @@ Please note, the S3 Source connector used to infer schemas from all the availabl
...
-  {"column_names": ["column1", "column2", "column3"]}
   ```
+  {
+    "column_names": [
+      "column1", "column2", "column3"
+    ]
+  }
+  ```
+
+- **Block Size**: This is the number of bytes to process in memory at a time while reading files. The default value of `10000` is usually suitable, but if your table is particularly wide \(lots of columns / data in fields is large\), increasing this might help avoid schema detection failures.
```
sherifnada, line 235:
```suggestion
- **Block Size**: This is the number of bytes to process in memory at a time while reading files. The default value of `10000` is usually suitable, but if your files are particularly wide \(lots of columns or the values in the columns are particularly large), increasing this might help avoid schema detection failures.
```

docs/integrations/sources/s3.md
```diff
@@ -152,66 +172,105 @@ Please note, the S3 Source connector used to infer schemas from all the availabl
...
+
+:::caution
+Be cautious when raising this value too high, as it may result in Out Of Memory issues due to increased memory usage.
+:::
 
 ### Parquet
 
-Apache Parquet file is a column-oriented data storage format of the Apache Hadoop ecosystem. It provides efficient data compression and encoding schemes with enhanced performance to handle complex data in bulk. For now, the solution involves iterating through individual files at the abstract level thus partitioned parquet datasets are unsupported. The following settings are available:
+Apache Parquet is a column-oriented data storage format of the Apache Hadoop ecosystem. It provides efficient data compression and encoding schemes with enhanced performance to handle complex data in bulk. For now, the solution involves iterating through individual files at the abstract level. Therefore, partitioned parquet datasets are unsupported. The following settings are available:
```
sherifnada, line 243: This sentence is hard to understand as a user so I just suggest removing it and saying partitioned datasets are unsupported
```suggestion
Apache Parquet is a column-oriented data storage format of the Apache Hadoop ecosystem. It provides efficient data compression and encoding schemes with enhanced performance to handle complex data in bulk. At the moment, partitioned parquet datasets are unsupported. The following settings are available:
```

docs/integrations/sources/s3.md
```diff
@@ -152,66 +172,105 @@ Please note, the S3 Source connector used to infer schemas from all the availabl
...
-The Jsonl parser uses pyarrow hence,only the line-delimited JSON format is supported.For more detailed info, please refer to the [docs](https://arrow.apache.org/docs/python/generated/pyarrow.json.read_json.html)
+- **Unexpected field behavior**: This parameter determines how any JSON fields outside of the explicit schema (if defined) are handled. Possible behaviors include:
+
+  - `ignore`: Unexpected JSON fields are ignored.
+  - `error`: Error out on unexpected JSON fields.
+  - `infer`: Unexpected JSON fields are type-inferred and included in the output.
+
+Set to `infer` by default.
+
+- **Block Size**: This sets the number of bytes to process in memory at a time while reading files. The default value of `10000` is usually suitable, but if your table is particularly wide \(lots of columns / data in fields is large\), increasing this might help avoid schema detection failures.
```
sherifnada, line 269:
```suggestion
- **Block Size**: This sets the number of bytes to process in memory at a time while reading files. The default value of `10000` is usually suitable, but if your files are particularly wide \(lots of columns or the values in the columns are particularly large), increasing this might help avoid schema detection failures.
```

sherifnada (approved):

docs/integrations/sources/s3.md
ChristoGrab: Will do!

docs/integrations/sources/s3.md
ChristoGrab: Makes sense, it is kind of self-defeating haha :) Removed

PR #28233, opened 2023-07-12
Maxi297/source updater

maxi297:
## What
First iteration to automate connector maintenance process. This includes:

### Phase: Merge into code
* Delete schemas as they are part of the manifest now
* Remove metadata from manifest
* overwrite manifest.yaml

### Phase: CATs
* update catalog if add/remove stream or change sync_modes
    * [@maxi297  to confirm] Note that some values are set as default and might have impact in CATs
* Export integration tests config from GSM

## How
Create a new project in "tools". For more information, refer to the README.md.

## 🚨 User Impact 🚨
As this isn't inserted in any value stream we already had, there is not impact except when running this script specifically

## Notes
This is outside of the CDK and therefore triggers a build for Airbyte CDK and Connector Bases. We should probably have a discussion as to where this should lie. Personally, it feels like this will eventually be workflows for the Connector Builder and therefore should lie in the airbyte-cdk/python/airbyte_cdk/connector_builder folder. The only drawback I see is that this means depending on something in the "tools" folder


tools/source_updater/updater/main.py
```diff
@@ -0,0 +1,238 @@
...
+    if args.debug:
+        logging.basicConfig(level=logging.DEBUG, force=True)
+
+    logger.info("Starting the update...")
+
+    repo = SourceRepository()
+
+    # FIXME is there a state of the local git we would like to enforce?
+
+    if not repo.exists(source_name):
```
maxi297, line 201: The following has not been unit tested yet. This will come in a follow up PR

airbyte-cdk/python/airbyte_cdk/sources/declarative/yaml_declarative_source.py
```diff
@@ -21,11 +23,17 @@ def __init__(self, path_to_yaml, debug: bool = False):
         super().__init__(source_config, debug)
 
     def _read_and_parse_yaml_file(self, path_to_yaml_file) -> ConnectionDefinition:
-        package = self.__class__.__module__.split(".")[0]
-
-        yaml_config = pkgutil.get_data(package, path_to_yaml_file)
-        decoded_yaml = yaml_config.decode()
-        return self._parse(decoded_yaml)
+        try:
```
maxi297, line 26: I didn't find a way to just pass a file path to a component in the airbyte_cdk to create a source so I had to add this. Let me know if I missed something

tools/source_updater/updater/source.py
```diff
@@ -0,0 +1,164 @@
...
+class _IndentingEmitter(yaml.emitter.Emitter):
+    def increase_indent(self, flow=False, indentless=False):
+        """Ensure that lists items are always indented."""
+        return super().increase_indent(
+            flow=False,
+            indentless=False,
+        )
+
+
+class _PrettyDumper(
```
maxi297, line 118: This is to reduce the diff between our current formatting and the default yaml lib one. There are still some diffs that are not caused by an actual change but it feel very manageable

tools/source_updater/updater/source.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def fetch_catalog(self, source_name: str) -> ConfiguredAirbyteCatalog:
+        try:
+            return ConfiguredAirbyteCatalog.parse_obj(BaseConnector.read_config(self._catalog_path(source_name)))
+        except ValidationError as error:
+            raise ValueError(error)
+
+    def update_catalog(self, source_name: str, catalog: ConfiguredAirbyteCatalog) -> None:
+        with open(self._catalog_path(source_name), "w") as catalog_file:
+            json.dump(catalog.dict(exclude_none=True), catalog_file, indent=2, default=lambda x: x.value)
```
maxi297, line 29: This is still causing some diffs as our lists are not always set up the same way. For example:
* [Inline](https://github.com/airbytehq/airbyte/blob/67a01d6be2f5064456c6c53b87301dafcd465153/airbyte-integrations/connectors/source-prestashop/integration_tests/configured_catalog.json#L7)
* [Not inline](https://github.com/airbytehq/airbyte/blob/67a01d6be2f5064456c6c53b87301dafcd465153/airbyte-integrations/connectors/source-prestashop/integration_tests/configured_catalog.json#L2)

However, the amount of diff is very minimal

tools/source_updater/updater/handler.py
```diff
@@ -0,0 +1,65 @@
...
+from airbyte_cdk.sources.declarative.manifest_declarative_source import ManifestDeclarativeSource
+from updater.catalog import CatalogMerger
+from updater.config import Config
+from updater.source import SourceRepository
+
+
+logger = logging.getLogger("handler")
+
+
+class SourceUpdaterHandler:
```
maxi297, line 14: Need to add tests around this class

tools/source_updater/README.md
```diff
@@ -0,0 +1,39 @@
+# Source update
```
girarda, line 1: idea (doesn't need to be done in this PR): Is it worth adding these new tools to airbyte-ci?

tools/source_updater/updater/catalog.py
```diff
@@ -0,0 +1,73 @@
...
+from airbyte_cdk.models import AirbyteStream, ConfiguredAirbyteCatalog, ConfiguredAirbyteStream, DestinationSyncMode, SyncMode
+from airbyte_cdk.sources import AbstractSource
+from airbyte_cdk.sources.streams import Stream
+from updater.config import Config
+from updater.source import SourceRepository
+
+logger = logging.getLogger("catalog")
+
+
+class CatalogAssembler:
```
girarda, line 14: nit: `ConfiguredCatalogAssembler`

tools/source_updater/updater/handler.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, source_repository: SourceRepository, catalog_merger: CatalogMerger):
+        self._source_repository = source_repository
+        self._catalog_merger = catalog_merger
+
+    def handle(self, source_name: str, new_manifest_source: ManifestDeclarativeSource, main_config: Config, other_configs: Set[Config]):
+        # FIXME eventually extract git related command somewhere it makes sense
+        # FIXME is there a state of the local git we would like to enforce?
+        branch_name = f"source-updater/updating-{source_name}"
```
girarda, line 22: presumably will be a command line argument?

tools/source_updater/updater/main.py
```diff
@@ -0,0 +1,80 @@
...
+                )
+            )
+
+    if not main_config:
+        raise ValueError(f"Could not find main config. Name should match {_DEFAULT_CONFIG_FILE}")
+    return main_config, other_configs
+
+
+if __name__ == "__main__":
+    import argparse
```
girarda, line 56: very nit: I recommend using click for new scripts. I find it much nice than argparse https://github.com/airbytehq/airbyte/blob/1c871756b31be55fc417e0df8f87f6ae9c842074/tools/ci_credentials/ci_credentials/main.py#L19-L22

tools/source_updater/updater/source.py
girarda: are the links pointing to the right line? I don't understand what diff you're referring to

tools/source_updater/updater/source.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def delete_schemas_folder(self, source_name: str) -> None:
+        schemas_folder_path = os.path.join(self._path_from_name(source_name), source_name.replace("-", "_"), "schemas")
+        for root, dirs, files in os.walk(schemas_folder_path, topdown=False):
```
girarda, line 33: nit: I think [`shutils.rmtree` with `ignore_errors=True` can do the traversal for you ](https://docs.python.org/3/library/shutil.html#shutil.rmtree)

tools/source_updater/README.md
maxi297: If we agree that the end goal is to have this as part of the connector builder, I would think this means it should be moved to `airbyte_cdk.conenctor_builder`. As a transition step, it can still be a script in this package that we can eventually expose through HTTP. I'll trigger the discussion tomorrow in the grooming session

tools/source_updater/updater/handler.py
maxi297: This is a good question. What are the concerns? I know there is a concern of clashes there.

I would like to avoid exposing this to the user. If the goal is to have SE use this or to be part of the Connector Builder, I don't think the user needs to know about that.

tools/source_updater/updater/source.py
maxi297: The concern is to have as few diffs as possible when updating the catalog. Using `json.dump(catalog.dict(exclude_none=True), catalog_file, indent=2, default=lambda x: x.value)` gives somewhat good results but it is not perfect because some lists are inline and others are not (see the link in the initial comment)

tools/source_updater/README.md
```diff
@@ -0,0 +1,39 @@
...
+CLI tooling to update the repository following a change in a low-code source. 
+
+## Set up
+
+### Install dependencies
+```
+cd tools/source_updated
```
brianjlai, line 8: nit: should be `cd tools/source-updater`

tools/source_updater/README.md
```diff
@@ -0,0 +1,39 @@
...
+(.venv) airbyte% python tools/source_updater/updater/main.py --help
+usage: main.py [-h] --source SOURCE --manifest MANIFEST --config CONFIG [--debug]
+
+Source updated
+
+optional arguments:
+  -h, --help           show this help message and exit
+  --source SOURCE      Name of the source. For example, 'source-jira'
+  --manifest MANIFEST  Path to the new yaml manifest file
+  --config CONFIG      Path to the config file or directory. We recommend using `airbyte-integrations/connectors/<source_name>/secrets/updated_configurations/*.json`
```
brianjlai, line 37: i'd pose to make the argument to make this location mandatory and non configurable. We should try to force them to use the `/secrets` directory just as another precaution to avoid accidentally putting up PRs or any commits that might accidentally include sensitive app secrets

tools/source_updater/updater/handler.py
```diff
@@ -0,0 +1,65 @@
...
+            logger.error(error_message)
+            raise ValueError(error_message)
+
+        if not self._source_repository.exists(source_name):
+            error_message = "Source does not exist. Please generate it as demonstrated by " \
+                            "https://docs.airbyte.com/connector-development/config-based/tutorial/create-source "
+            raise ValueError(error_message)
+
+        catalog = self._source_repository.fetch_catalog(source_name)
+        self._catalog_merger.merge_and_save(source_name, catalog, new_manifest_source, main_config)
```
brianjlai, line 37: I'm finding these interfaced to be a little bit convoluted. Right now the repository gets the catalog and passes it to the catalog_merger. And then the catalog merger uses the repository's update method. It just feels like these two objects are really dependent on each other's specific methods and implementation.

What if we were to refactor this a little bit so source_repository has a field catalog_merger. And the catalog_merger has no dependency on the repo at all. And instead of these two lines we just have one method `self._source_repository.update_catalog()`

And a rough flow of what the new`SourceRepository.update_catalog()` method look like:
1. Call to fetch the catalog (fetch_catalog)
2. Call the `catalog_merger.merge()` to perform the catalog transformation to add/remove streams (but does not save it).
3. Call to update the catalog 

Does that make sense? I was just thinking about this from the angle of how we can simplify how we invoke fetch + save behind a simpler interface.

maxi297: I think it makes a lot of sense and I'm glad you shared this with me. I've simplified this, thanks!

tools/source_updater/updater/handler.py
```diff
@@ -0,0 +1,66 @@
...
+        catalog = self._source_repository.fetch_catalog(source_name)
+        is_updated = self._catalog_merger.merge_into_catalog(source_name, catalog, new_manifest_source, main_config)
+        if is_updated:
+            self._source_repository.update_catalog(source_name, catalog)
+        self._update_manifest(source_name, new_manifest_source)
+        self._source_repository.upsert_secrets(source_name, other_configs.union({main_config}))
+
+        # CATs
+        #   Update expected records
+        pass
```
brianjlai, line 44: Doesn't pass usually go inside a if statement/method definition etc if we haven't implemented it yet to stop from throwing errors. Is this just mean to accompany the comment to make it more clear this is future work?

tools/source_updater/updater/main.py
```diff
@@ -0,0 +1,75 @@
...
+@click.option("--manifest", type=str, required=True, help="Path to the new yaml manifest file")
+@click.option("--debug", is_flag=True, default=False, help="Enable debug logs")
+def main(source: str, manifest: str, debug: bool) -> None:
+    source_name = source
+    manifest_path = manifest
+    if debug:
+        logging.basicConfig(level=logging.DEBUG, force=True)
+
+    logger.info("Starting the update...")
+    repo = SourceRepository()
```
brianjlai, line 63: nit: doesn't look like we use the repo anywhere until we pass it to `SourceUpdaterHandler`. We could just instantiate it inside there w/o the need for another variable.

Also, since all the parameters that we pass to `SourceUpdaterHandler` are the defaults, we can just get rid of all the inputs and initialize them in the `__init__` constructor. wdyt?

tools/source_updater/updater/catalog.py
```diff
@@ -0,0 +1,70 @@
...
+        logger.debug(f"Streams identified from current {source_name}: {catalog_stream_names}")
+
+        manifest_streams = new_source.streams(config.content)
+        manifest_catalog = self._catalog_assembler.assemble(manifest_streams)
+        manifest_stream_names = {stream.stream.name for stream in manifest_catalog.streams}
+        logger.debug(f"Streams identified from new implementation: {manifest_stream_names}")
+
+        streams_to_remove = catalog_stream_names - manifest_stream_names
+        if streams_to_remove:
+            logger.info(f"Removing streams {streams_to_remove}...")
```
brianjlai, line 60: nit: might be worth adding the number of streams if this gets super verbose. i.e. `logger.info(f"Removing {len(streams_to_remove)} stream(s): {streams_to_remove}...")`

tools/source_updater/updater/handler.py
brianjlai: just double checking. i see tests so we're good?

brianjlai (approved):
looks good to me, just a few small comments

tools/source_updater/updater/handler.py
maxi297: That was meant to accompany the comment but will remove it since it creates confusion

tools/source_updater/updater/handler.py
maxi297: Yes, added those last week so I'll resolve this comment!

tools/source_updater/updater/main.py
maxi297: I added them as parameters in the init so that I can just pass a mock during tests. Default arguments does not seem advisable as it would be shared between all instances. Else, we could make the param optional in the __init__ params and if None instantiate it. I don't like this too much because it means that if the interface of `SourceRepository` changes (the __init__ method specifically here), `SourceUpdaterHandler` might change as well. However, I agree that `SourceUpdaterHandler()` seems cleaner than `SourceUpdaterHandler(SourceRepository(), CatalogMerger(ConfiguredCatalogAssembler()))`. Let me know what you prefer!

brianjlai: >Default arguments does not seem advisable as it would be shared between all instances. Else, we could make the param optional in the init params and if None instantiate it.

Sorry yeah i could have been more clear, not default construct args, but just the __init__ will create a new instance of `SourceRepository()`. In that case it wouldn't be shared between instances of `SourceUpdaterHandler()` because each new creation is its own repo/merger. I would've thought? Fair point about the testing, happy to leave as is and only make the change to `SourceUpdaterHandler(SourceRepository(), CatalogMerger(ConfiguredCatalogAssembler()))`

maxi297:
Will not merge now as it would add maintenance (there is some work being done on the `tools` folder) while providing little benefits

cgardens:
We are closing this PR due to inactivity. We regularly close PRs that were created more than 6 months ago. If you'd like to keep working on it, please feel free reopen it. We'd be happy to keep working with you.

PR #28251, opened 2023-07-12
[Source-Postgres] Rollback to 2.1.1 in case 3.0.0 fails [DO NOT MERGE]

nguyenaiden:
## What
In the case that we need to rollback this [change](https://github.com/airbytehq/airbyte/pull/27442). This PR redirects version 3.0.0 to point at 2.1.1. In the case that we need to rollback, this PR will be merged, and another PR will be submitted with version 3.0.1 that removes the previous changes. Also, `postgres.md` changelog will be updated with version 3.0.1 stating the rollback.


nguyenaiden:
Closing as change is shipped and is stable

PR #27845, opened 2023-06-29
1-stream-1-table (Destinations v2) `destination-bigquery` beta testing publish PR

evantahler:
## ⚠️ This PR should never be merged ⚠️

This PR enables the secret config option to allow folks to beta-test `destination-bigquery` + 1s1t (Destinations V2).  We will be using this PR to publish dev/beta images of the connector.

### Steps for updating the Beta Image

1. Create a development branch based off of `master` and make your changes, get your PR reviewed, (generally follow the normal process for all changes)
2. After your PR from step 1 is merged into `master`, merge the newly updated `master` into `destinations-v2-beta`. An easy way to do this is clicking the `Update branch` button at the bottom of this PR.
3. Run [this GitHub Workflow](https://github.com/airbytehq/airbyte/actions/workflows/publish_connectors.yml) with the following options
    - branch: `destinations-v2-beta`
    - options: `--name=destination-bigquery`
    - leave the `--pre-release` flag set and use whatever the `runs-on` option is set to
4. Once the Github Workflow completes, Update this PR's Beta Releases Table in the description
5. Verify the image is available in [DockerHub](https://hub.docker.com/repository/docker/airbyte/destination-bigquery/general)
6. Update the feature flag to use the correct image tag variant go to the [Variations](https://app.launchdarkly.com/default/production/features/connectors.versionOverrides/variations) page, search for `Bigquery 2.0.0-destinationsv2-beta` which is currently "Variation 25" and set its "value" to the image tag, e.g. `1.5.7-dev.7e4a9dd278`
7. Also make sure OSS beta testers are informed of the new image

Feature flag things to add the new pre-release docker image to the workspaces you want to test.  You will need to change the tag on the varaint each release.  [Internal Docs](https://internal-docs.airbyte.io/How-Tos/Connector-Version-Overrides/) on how to update the feature flag

## Beta Releases

| Date | Who | GH Action | Version |
|--------|--------|--------|--------|
| 2023-06-29 | @evantahler | [Link](https://github.com/airbytehq/airbyte/actions/runs/5415948291) | [airbyte/destination-bigquery:1.5.0-dev.7a773a7766](https://hub.docker.com/r/airbyte/destination-bigquery/tags) |
| 2023-07-05 | @edgao  | [Link](https://github.com/airbytehq/airbyte/actions/runs/5469974586) | [airbyte/destination-bigquery:1.5.2-dev.27fd242fda](https://hub.docker.com/r/airbyte/destination-bigquery/tags) |
[...]

PR #27936, opened 2023-07-03
Destination bigquery: rerelease 1s1t behind gate

edgao:
readers: please see individual commits.

plan:
1. merge bugfixes into this branch
2. release to test workspace -> slow rollout
3. squash bugs as we find them

PR #28256, opened 2023-07-12
stop building normalization

edgao:
instead of building normalization, just pull from the already-published version. We're committed to phasing out legacy normalization, so it's not like we'll need to make changes here.

https://github.com/airbytehq/airbyte/pull/27845/commits/879246135b48be2c020c44741d8eb252aff6b85e does the same thing for the old airbyteDocker-style builds; I pushed it directly to the branch just to simplify some of the build investigations. This PR ports the logic to airbyte-ci.

also, this:

image


evantahler (commented):
OK hear me out: does this mean that building normalization images is no longer a dependency of the connector?  Can we remove all of [this](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/bases/base-normalization/build.gradle) gradle stuff?  You *can* still build the normalization images... but now we don't need to wait for tests and wantnot to do it.

edgao:
definitely some of it. I don't think we can kill the entire base-normalization build.gradle, but e.g. https://github.com/airbytehq/airbyte/commit/879246135b48be2c020c44741d8eb252aff6b85e#diff-0f6023601f6e522b942f3a29626e10aeece3f0565e02bf70759ab789309498dc is gone.

we should do the same for the rest of our destinations when we get to them!

evantahler:
Your base build failed, but for a reason that makes no sense:

```
2023-07-13T00:36:29.0158140Z * What went wrong:
2023-07-13T00:36:29.0159879Z Execution failed for task ':airbyte-integrations:connectors:destination-snowflake:compileIntegrationTestJavaJava'.
2023-07-13T00:36:29.0161231Z > Compilation failed; see the compiler error output for details.
```

evantahler (approved):
👍 once the tests pass... or maybe not even then

edgao:
I think the snowflake build was broken on master for a bit - merging latest should fix 🤞

edgao:
oh hm - this pr didn't actually trigger any test pipelines... so I missed the fact that it broke the ability to run tests >.>

pushed https://github.com/airbytehq/airbyte/pull/27936/commits/aa7b02943ab39ad6ff7fb38b89bb9a2069e623b3 to the 1s1t_redeploy branch, which seems to be working locally (or at least the tests are running now)

Issue #27469, opened 2023-06-19
File CDK: Support JSONL "schemaless" option


clnoll:
When this option is selected, the schema emitted by `get_json_schema` will be `{"data": "object"}`, and records will be emitted in the format `{"data": {"col1": val1, …}}`.

This will be implemented for CSV first, so should just be wired through for JSONL too.

#### Acceptance Criteria
- [ ] This option is supported for JSONL file type.
- [ ] `read_records` is updated to skip record validation if this option is selected.

PR #28259, opened 2023-07-13
File-based CDK: implement JSONL parser

### LINKED ISSUE #26754: File CDK: Implement get_json_schema and parse_records for JSONL files
#### `get_json_schema`

Return a JSON Schema for a group of records as follows:
1. Create an empty master schema.
For each file:
2. Open and read the file. For each line, until we've read 1MB of data, infer the schema using the [pyarrow library](https://arrow.apache.org/docs/python/json.html#automatic-type-inference). Once we've read 1MB, return the schema for that file even if we haven't finished reading it. Merge the file-specific schema with the master schema

The `source_files_abstract` [`jsonl_parser.get_inferred_schema`](https://github.com/airbytehq/airbyte/blob/3aa0c284e2631a94e5b2e345fefd70220e15638d/airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/formats/jsonl_parser.py#L86) method may be useful for this, however tests need to be modified to read more than one file.

#### `parse_records`
Create a parser for JSONL file types. The `source_files_abstract` [`jsonl_parser.stream_records`](https://github.com/airbytehq/airbyte/blob/3aa0c284e2631a94e5b2e345fefd70220e15638d/airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/formats/jsonl_parser.py#106) may be useful for this.

#### Acceptance Criteria
- [ ] `discover` returns a configured catalog with the master schema for streams consisting of JSONL files.
- [ ] `read` returns `AirbyteRecordMessage`s for streams consisting of JSONL files.
- [ ] Errors are handled appropriately and have test coverage; error messages are clear.

clnoll:
## What
Implements the JSONL schema inferrer and parser.

Because the schemaless option is already wired through I went ahead and wrote some tests for it, which closes out https://github.com/airbytehq/airbyte/issues/27469.

Closes https://github.com/airbytehq/airbyte/issues/26754 and https://github.com/airbytehq/airbyte/issues/27469.

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
@@ -9,9 +9,9 @@ from typing import Any, List, Mapping, Optional, Tuple, Type
 
 from airbyte_cdk.models import ConfiguredAirbyteCatalog, ConnectorSpecification
 from airbyte_cdk.sources import AbstractSource
+from airbyte_cdk.sources.file_based.availability_strategy import AbstractFileBasedAvailabilityStrategy, DefaultFileBasedAvailabilityStrategy
 from airbyte_cdk.sources.file_based.config.abstract_file_based_spec import AbstractFileBasedSpec
 from airbyte_cdk.sources.file_based.config.file_based_stream_config import FileBasedStreamConfig
-from airbyte_cdk.sources.file_based.default_file_based_availability_strategy import DefaultFileBasedAvailabilityStrategy
 from airbyte_cdk.sources.file_based.discovery_policy import AbstractDiscoveryPolicy, DefaultDiscoveryPolicy
 from airbyte_cdk.sources.file_based.exceptions import ConfigValidationError, FileBasedSourceError
 from airbyte_cdk.sources.file_based.file_based_stream_reader import AbstractFileBasedStreamReader
@@ -21,7 +21,6 @@ from airbyte_cdk.sources.file_based.schema_validation_policies import DEFAULT_SC
 from airbyte_cdk.sources.file_based.stream import AbstractFileBasedStream, DefaultFileBasedStream
 from airbyte_cdk.sources.file_based.stream.cursor.default_file_based_cursor import DefaultFileBasedCursor
 from airbyte_cdk.sources.streams import Stream
-from airbyte_cdk.sources.streams.availability_strategy import AvailabilityStrategy
 from pydantic.error_wrappers import ValidationError
 
 DEFAULT_MAX_HISTORY_SIZE = 10_000
@@ -32,7 +31,7 @@ class FileBasedSource(AbstractSource, ABC):
         self,
         stream_reader: AbstractFileBasedStreamReader,
         catalog: Optional[ConfiguredAirbyteCatalog],
-        availability_strategy: Optional[AvailabilityStrategy],
+        availability_strategy: Optional[AbstractFileBasedAvailabilityStrategy],
         spec_class: Type[AbstractFileBasedSpec],
         discovery_policy: AbstractDiscoveryPolicy = DefaultDiscoveryPolicy(),
         parsers: Mapping[str, FileTypeParser] = default_parsers,
@@ -76,7 +75,7 @@ class FileBasedSource(AbstractSource, ABC):
                 (
                     stream_is_available,
                     reason,
-                ) = stream.availability_strategy.check_availability(stream, logger, self)
+                ) = stream.availability_strategy.check_availability_and_parsability(stream, logger, self)
             except Exception:
                 errors.append(f"Unable to connect to stream {stream} - {''.join(traceback.format_exc())}")
             else:
--- a/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/jsonl_parser.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/jsonl_parser.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
 
+import json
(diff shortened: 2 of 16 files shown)


airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/jsonl_parser.py
```diff
@@ -2,24 +2,61 @@
...
 from airbyte_cdk.sources.file_based.config.file_based_stream_config import FileBasedStreamConfig
 from airbyte_cdk.sources.file_based.file_based_stream_reader import AbstractFileBasedStreamReader
 from airbyte_cdk.sources.file_based.file_types.file_type_parser import FileTypeParser
 from airbyte_cdk.sources.file_based.remote_file import RemoteFile
+from airbyte_cdk.sources.file_based.schema_helpers import PYTHON_TYPE_MAPPING, merge_schemas
 
 
 class JsonlParser(FileTypeParser):
 
     MAX_BYTES_PER_FILE_FOR_SCHEMA_INFERENCE = 1_000_000
```
sherifnada, line 18: why do we have this instead of tweaking the number of files to use for schema inference?

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/jsonl_parser.py
```diff
@@ -2,24 +2,61 @@
...
+
+        with stream_reader.open_file(file) as fp:
+            for line in fp:
+                if read_bytes < self.MAX_BYTES_PER_FILE_FOR_SCHEMA_INFERENCE:
+                    line_schema = self.infer_schema_for_record(json.loads(line))
+                    inferred_schema = merge_schemas(inferred_schema, line_schema)
+                    read_bytes += len(line)
+
+        if read_bytes > self.MAX_BYTES_PER_FILE_FOR_SCHEMA_INFERENCE:
+            logging.warning(
```
sherifnada, line 42:
```suggestion
            logger.warning(
```

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/jsonl_parser.py
clnoll: We do both. We check bytes here but not in the other parsers because we're reading many rows to get the schema for JSONL files and want to be able to cut it off at some point for very large files.

airbyte-cdk/python/unit_tests/sources/file_based/in_memory_files_source.py
```diff
@@ -78,7 +79,7 @@ def _make_file_contents(self, file_name: str):
         if self.file_type == "csv":
             return self._make_csv_file_contents(file_name)
         else:
```
girarda, line 81: this should check that the file type is jsonl and fail otherwise

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/jsonl_parser.py
```diff
@@ -22,7 +24,27 @@ async def infer_schema(
...
+
+        with stream_reader.open_file(file) as fp:
+            for line in fp:
+                if read_bytes < self.MAX_BYTES_PER_FILE_FOR_SCHEMA_INFERENCE:
+                    line_schema = self.infer_schema_for_record(json.loads(line))
+                    inferred_schema = merge_schemas(inferred_schema, line_schema)
+                    read_bytes += len(line)
+
+        if read_bytes > self.MAX_BYTES_PER_FILE_FOR_SCHEMA_INFERENCE:
+            logger.warning(
```
girarda, line 42: nit: it's possible the whole file was read since we compute the number of bytes read after reading the line

girarda (approved):

brianjlai (commented):
> schemaless option is already wired

when you say already wired, you mean it's already implemented/supported?

The changes look good, but we've had some side discussions on the `newlines_in_values` format option. Are we still concerned about using pyarrow in order to support that option and can we summarize what we're aligning in this PR?

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/jsonl_parser.py
clnoll: I've updated the parser to use pyarrow so have removed this check for now.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -74,7 +73,7 @@ def check_connection(self, logger: logging.Logger, config: Mapping[str, Any]) ->
                 (
                     stream_is_available,
                     reason,
-                ) = stream.availability_strategy.check_availability(stream, logger, self)
+                ) = stream.availability_strategy.check_availability_and_parsability(stream, logger, self)
```
clnoll, line 76: Re the conversation from the other day where I was talking about figuring out how to do a different availability check during `check` versus `read` - turns out that it didn't require access to the command.

girarda:
@clnoll 

> That means that deploying line-by-line parsing could cause the failure of previously successful syncs
Do we have a sense of whether this is something we should care about?

Would a solution that supports newlines only within quoted fields (which I think is supported by the csv library).

Not something I want to block on, but it does seem unfortunate to need to read the whole file for what seems like an edge case (unquoted newlines)

clnoll:
@girarda a solution that has newlines in quoted fields works with the previous approach, and I'm leaning toward going back to doing that. We do run the risk of breaking syncs that are currently working because right now a user can have a file with escaped non-escaped newlines and we don't error, even if the `newlines_in_values` option is not set, but if we realize we need it we can use pyarrow in a special JSONL parser implementation in the S3 connector.

PR #28263, opened 2023-07-13
[file-based cdk] spec schema improvements and fixes

brianjlai:
## What
Sherif pointed out a few UI incompatibility issues with the spec and a few opportunities to improve the readability of the spec when it is presented to the UI.

https://github.com/airbytehq/airbyte/pull/28139#pullrequestreview-1527393010

## How

There were specifically two places where we were using anyOf which (similar to allOf) is not compatible with the UI. We can instead opt to replace this with `oneOf` which is UI compatible.

This PR also adds `pydantic.Field()` to all config options which allows us to more granularly control how the pydantic models are transformed in JSON schemas for the spec. Since we basically had none in the config, I wrote all of them, but would welcome feedback if it isn't articulated quite right for the fields I didn't have as much familiarity with.

## Recommended reading order
1. `file_based_stream_config.py`
2. `abstract_file_based_spec.py`


airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py
```diff
@@ -6,9 +6,9 @@
 from enum import Enum
 from typing import Any, Dict, List, Mapping, Optional, Union
 
-from pydantic import BaseModel, validator
+from pydantic import BaseModel, Field, validator
 
-PrimaryKeyType = Optional[Union[str, List[str], List[List[str]]]]
+PrimaryKeyType = Optional[Union[str, List[str]]]
```
brianjlai, line 11: As per Sherif's comment in the other thread, ideally we'd want just `List[str]`. Looking over the existing s3 spec, I don't see any concept of a primary key, but it might be safer to allow for str as well since that is a pretty common way they are specified for other sources

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
```diff
@@ -69,7 +73,14 @@ def add_legacy_format(schema: dict) -> dict:
         config must be adjusted to support the generic mapping object. Once configs no longer adhere to the old
         format we can remove this change.
         """
+        legacy_format_options = {
```
brianjlai, line 76: Originally the `{"type": "object"}` to represent the legacy type worked when the restriction was `anyOf`. But now that we need to match exactly one with `oneOf`, we need to be more granular with this check. The legacy s3 format is always indicated with the `filetype`. We have existing tests scenarios in both formats to verify this works

sherifnada (approved):
🙏🏼🙏🏼🙏🏼🙏🏼🙏🏼

clnoll (approved):

Issue #27153, opened 2023-06-08
Low code CDK: Incremental sync with different format incoming/outgoing


flash1293:
### What area the feature impact?

None

### Revelant Information

In most APIs, dates are formatted the same way in request parameters and records in the responses. This works will with how incremental syncs are configured that just allow a single datetime format used in both directions.

However, sometimes APIs require a different format which can't be expressed using the low-code CDK right now.

For these use cases, it should be possible to define a second optional format:
```
    incremental_sync:
      cursor_field: "startDate"
      datetime_format: "%Y-%m-%d"
      cursor_granularity: "P1D"
      cursor_datetime_format: "%s"
```

flash1293:
+1 from community slack

jhsb25:
i have tried writing 2 connectors and have come across this issue both times

maxi297:
+1 from community slack

maxi297:
Closed as part of https://github.com/airbytehq/airbyte/issues/28782

davidfromtandym:
Any chance of adding a integer format for UNIX Epoch time in milliseconds? I see `%s` for an integer format in seconds, but no option for milliseconds.

davidfromtandym:
> 

Or alternatively, being able to perform simple functions on the format .. i.e., `%s * 1000`

flash1293:
@davidfromtandym That's a very valid feature request, could you open a separate issue for it?

PR #28268, opened 2023-07-13
Improve date format detection

flash1293:
## What

This PR improves the date format detection. Previously, the `Z` suffix would be mapped to the `%z` placeholder. However, while this works for parsing the date in API responses, dates formatted this way might not be accepted as a valid  format in outgoing requests as it will be rendered as `+0000`.

This PR fixes this by introducing a special case for the Z prefix to avoid the specific problem. Note that this is not a full fix, https://github.com/airbytehq/airbyte/issues/27153 has to be implement to cover all the cases.


girarda (approved):
🚢

PR #28293, opened 2023-07-13
connector builder: Set test_read_limit_reached to true if we hit the max records limit

girarda:
## What
* Set the `test_read_limit_reached` flag to True if the number of records read is >= max number of records

Note: We'll need a follow up PR to update the message show in the webapp when the flag is set to true

## How
* Count the number of records and set the flag to true if we read the maximum number of records


airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
```diff
@@ -38,7 +38,7 @@ class NoRequiredSchemaBuilder(SchemaBuilder):
 
 
 # This type is inferred from the genson lib, but there is no alias provided for it - creating it here for type safety
-InferredSchema = Dict[str, Union[str, Any, List, List[Dict[str, Union[Any, List]]]]]
```
girarda, line 41: `Any` is a subset of the `Union`

airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
```diff
@@ -254,25 +259,31 @@ def _create_request_from_log_message(json_http_message: dict) -> HttpRequest:
...
+            for page in _slice.pages:
+                record_count += len(page.records)
+                if record_count >= self._max_record_limit:
+                    return True
         return False
 
-    def _parse_slice_description(self, log_message):
-        return json.loads(log_message.replace(AbstractSource.SLICE_LOG_PREFIX, "", 1))
+    def _parse_slice_description(self, log_message: str) -> Dict[str, Any]:
+        return json.loads(log_message.replace(AbstractSource.SLICE_LOG_PREFIX, "", 1)) # type: ignore
```
girarda, line 283: we know the strings can only be strings

airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
```diff
@@ -216,7 +221,7 @@ def _close_page(current_page_request, current_page_response, current_slice_pages
             raise ValueError("Every message grouping should have at least one request and response")
 
         current_slice_pages.append(
-            StreamReadPages(request=current_page_request, response=current_page_response, records=deepcopy(current_page_records))
+            StreamReadPages(request=current_page_request, response=current_page_response, records=deepcopy(current_page_records))  # type: ignore
```
girarda, line 224: deepcopy isn't typed

airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
```diff
@@ -176,23 +178,26 @@ def _get_message_groups(
...
             at_least_one_page_in_group
             and message.type == MessageType.LOG
             and (MessageGrouper._is_page_http_request(json_message) or message.log.message.startswith("slice:"))
         )
 
     @staticmethod
-    def _is_page_http_request(json_message):
-        return MessageGrouper._is_http_log(json_message) and not MessageGrouper._is_auxiliary_http_request(json_message)
+    def _is_page_http_request(json_message: Optional[Dict[str, Any]]) -> bool:
+        if not json_message:
```
girarda, line 190: None-check

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
```diff
@@ -84,11 +84,12 @@ def _clean(self, node: InferredSchema):
                     node["type"] = [node["type"], "null"]
                     node.pop("anyOf")
             if "properties" in node:
-                for key, value in list(node["properties"].items()):
-                    if isinstance(value, dict) and value.get("type", None) == "null":
-                        node["properties"].pop(key)
-                    else:
-                        self._clean(value)
+                if isinstance(node["properties"], dict):
```
girarda, line 87: this should always be true, but mypy doesn't know. Better safe than sorry

airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
```diff
@@ -254,25 +259,31 @@ def _create_request_from_log_message(json_http_message: dict) -> HttpRequest:
...
 
     @staticmethod
-    def _create_response_from_log_message(json_http_message: dict) -> HttpResponse:
+    def _create_response_from_log_message(json_http_message: Dict[str, Any]) -> HttpResponse:
         response = json_http_message.get("http", {}).get("response", {})
         body = response.get("body", {}).get("content", "")
         return HttpResponse(status=response.get("status_code"), body=body, headers=response.get("headers"))
 
-    def _has_reached_limit(self, slices: List[StreamReadPages]):
+    def _has_reached_limit(self, slices: List[StreamReadSlices]) -> bool:
```
girarda, line 267: the actual change is in this method

airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
```diff
@@ -230,17 +235,17 @@ def _read_stream(self, source: DeclarativeSource, config: Mapping[str, Any], con
...
 
     @staticmethod
-    def _parse_json(log_message: AirbyteLogMessage):
+    def _parse_json(log_message: AirbyteLogMessage) -> Optional[Dict[str, Any]]:
         # TODO: As a temporary stopgap, the CDK emits request/response data as a log message string. Ideally this should come in the
         # form of a custom message object defined in the Airbyte protocol, but this unblocks us in the immediate while the
         # protocol change is worked on.
         try:
-            return json.loads(log_message.message)
+            return json.loads(log_message.message)  # type: ignore
```
girarda, line 243: json.loads returns `Any`

airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
```diff
@@ -83,8 +83,10 @@ def get_message_groups(
                     latest_config_update = message_group
             elif isinstance(message_group, AuxiliaryRequest):
                 auxiliary_requests.append(message_group)
-            else:
+            elif isinstance(message_group, StreamReadSlices):
                 slices.append(message_group)
+            else:
```
girarda, line 88: mostly to make mypy happy, but this is a reasonable check

maxi297: This is very important. In Java, you usually use switch/case and the compiler can emit a warning if some cases aren't managed based on the typing. I like it personally

airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
maxi297: Is this something where the typing is not provided by the lib so we need to use `Any`? Should we we restrict this to `Dict[str, JSONType]` where `JSONType = Union[str, int, float, bool, None, Dict[str, JSONType], List[JSONType]]`?

airbyte-cdk/python/airbyte_cdk/utils/schema_inferrer.py
maxi297: nit: would it be more readable to have `if "properties" in node and isinstance(node["properties"], dict):`?

maxi297 (approved):
LGTM!

Issue #28299, opened 2023-07-13
Dagger Pipeline over caching


bnchrch:
## Problem

We've observed an issue in our Dagger pipeline where certain files and steps seem to be using outdated versions of code. Here are the details:

- Our team member, Ed, noticed in a docker image (`airbyte/destination-bigquery:1.5.2-dev.c27684e40d`) that the spec was incorrect. 
- Ed then noticed in another docker image (`airbyte/destination-bigquery:1.5.2-dev.57422c1c6e`) that `run_normalization` was incorrect.
- Upon investigation, we confirmed that the spec output was indeed incorrect.
- Despite this, we were unable to reproduce the issue in various settings. New branches with minor spec changes, branching off the offending commits, even deleting the image and republishing the pre-release, did not work.
- Dagger was also unable to detect anything.
- We initially suspected that the Gradle cache might be the culprit, and disabled it in this [PR](https://github.com/airbytehq/airbyte/pull/28248).
- The next morning, Ed pointed out that `run_normalization.sh` is not handled by Gradle and remains an issue (see the related [Slack message](https://airbytehq-team.slack.com/archives/C05GA80UVRD/p1689267021475939)).
- We also noticed that we had a publish on master fail due to a metadata file being stale in the upload step (see the related [Discord message](https://discord.com/channels/707636530424053791/1072668966151004290/1129109441363591232) or [gha run](https://github.com/airbytehq/airbyte/actions/runs/5545415378/jobs/10124311400)).

These observations suggest that something might be wrong with the cache, particularly for Java connectors.

### What we need to figure out

- How did an image get old code in two cases?
- How did the metadata upload step get an old metadata file when spec was fine?

Both of these questions seem potentially related and should be investigated further. The metadata upload step, in particular, seems like a good starting point.

bnchrch:
New information:

When publishing  destination-bigquery (ss attached) and destination-gcs
GHA Run: https://github.com/airbytehq/airbyte/actions/runs/5555239244/jobs/10146110335#step:4:1588
Dagger Run: https://alpha.dagger.cloud/runs/14fbbf73-5f2f-4d2a-b11c-ab7dcf36cbe8

When publishing destination-snowflake
GHA Run: https://github.com/airbytehq/airbyte/actions/runs/5555925400/jobs/10147743126#step:4:572
Dagger Run: https://alpha.dagger.cloud/runs/c9e45d03-70ea-40fe-9b63-812c6329c856 


[...]

PR #28356, opened 2023-07-17
connectors-ci: copy metadata with `with_new_file` instead of `with_file`

alafanechere:
## What
Relates to #28299
Some publish pipeline skip metadata upload because the pipeline is using an old version of the file that was already updated:
> On investigation the logs show that the step was trying to publish the previous metadata file.

The dagger SDK API has two function to copy a file to a container: 
* [`with_file`:  _Retrieves this container plus the contents of the given file copied to the given path._](https://dagger-io.readthedocs.io/en/sdk-python-v0.5.4/api.html#dagger.api.gen.Container.with_file)
* [`with_new_file`: _Retrieves this container plus a new file written at the given path._ ](https://dagger-io.readthedocs.io/en/sdk-python-v0.5.4/api.html#dagger.api.gen.Container.with_new_file)

I [asked the dagger team ](https://discord.com/channels/707636530424053791/1129454621044252772/1130513188887220254) to deblur the behavior of these two functions.
## How
The existing metadata pipeline uses `with_file`. I suggest trying `with_new_file`. It might force a write operation to the container while `with_file` might be cached?

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
@@ -76,10 +76,10 @@ class MetadataUpload(PoetryRun):
         super().__init__(context, title, METADATA_DIR, METADATA_LIB_MODULE_PATH)
 
         # Ensure the icon file is included in the upload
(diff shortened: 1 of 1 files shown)


bnchrch (approved):
Good call 👍

PR #28358, opened 2023-07-17
connectors-ci: bust cache on java connector build

alafanechere:
## What
Relates to https://github.com/airbytehq/airbyte/issues/28299

We observe that some java connectors are occasionally built with stale code.

## How
* Bust the cache on connector container build by introducing a cache buster env var.
* Use `with_new_file` to copy the connector tar archive to the container

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
@@ -24,7 +24,7 @@ from ci_connector_ops.pipelines.consts import (
     PYPROJECT_TOML_FILE_PATH,
 )
 from ci_connector_ops.pipelines.utils import get_file_contents
-from dagger import CacheSharingMode, CacheVolume, Client, Container, DaggerError, Directory, File, Platform, Secret
+from dagger import CacheVolume, Client, Container, DaggerError, Directory, File, Platform, Secret
 from dagger.engine._version import CLI_VERSION as dagger_engine_version
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -776,14 +776,16 @@ def with_integration_base_java_and_normalization(context: PipelineContext, build
...
+    context: ConnectorContext, connector_java_tar_file: File, build_platform: Platform, bust_cache: bool = False
+) -> Container:
     application = context.connector.technical_name
-
     build_stage = (
         with_integration_base_java(context, build_platform)
         .with_workdir("/airbyte")
         .with_env_variable("APPLICATION", context.connector.technical_name)
-        .with_file(f"{application}.tar", connector_java_tar_file)
+        .with_env_variable("CACHEBUSTER", str(uuid.uuid4()) if bust_cache else "")
```
evantahler, line 787: non-blocking question - if every java build makes a new cache 'namespace' with this random UUID... do we ever clean up this chache?  It seems like it could get very large

evantahler (approved):

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
alafanechere: My understanding is that once the env var is set all the layers afterward are busted. I believe that once busted the layers are not cached anymore and the garbage collection logic is on the build kit side.

alafanechere: And, even if the layers are not garbage collected, this change won't introduce uuid namespaced cache as layers are not namespaced and are reused / shared across images, so we're not introducing an ever growing cache.

bnchrch (approved):
Ill approve now, but lets not merge this unless we know Daggers fix does not work

PR #28359, opened 2023-07-17
Ingore unit tests in mypy check

flash1293:
As discussed offline, this PR disables the CI mypy check for unit tests, as these are not as relevant to code stability as the main source code and currently contain a large number of violations.

Re-enabling this check is tracked here: https://github.com/airbytehq/airbyte-internal-issues/issues/1963


airbyte-cdk/python/bin/run-mypy-on-modified-files.sh
```diff
@@ -1,2 +1,3 @@
 set -e
-git diff --name-only --relative --diff-filter=d remotes/origin/master -- . | grep -E '\.py$' | xargs .venv/bin/python -m mypy --config-file mypy.ini --install-types --non-interactive
\ No newline at end of file
+# TODO change this to incluyde unit_tests as well once it's in a good state
```
girarda, line 2:
```suggestion
# TODO change this to include unit_tests as well once it's in a good state
```

girarda (approved):

airbyte-cdk/python/bin/run-mypy-on-modified-files.sh
```diff
@@ -1,2 +1,3 @@
 set -e
-git diff --name-only --relative --diff-filter=d remotes/origin/master -- . | grep -E '\.py$' | xargs .venv/bin/python -m mypy --config-file mypy.ini --install-types --non-interactive
\ No newline at end of file
+# TODO change this to include unit_tests as well once it's in a good state
+git diff --name-only --relative --diff-filter=d remotes/origin/master -- airbyte_cdk source_declarative_manifest | grep -E '\.py$' | xargs .venv/bin/python -m mypy --config-file mypy.ini --install-types --non-interactive
```
maxi297, line 3: We just need to make sure that whenever we will add a new package, we will have to update this. Can we work by exclusion instead? I see that `[tool.mypy] exclude` could do the trick

maxi297 (commented):
Question regarding how we can make this easier to maintain

airbyte-cdk/python/bin/run-mypy-on-modified-files.sh
flash1293: Fine with that too, feel free to update the PR @maxi297 - I can also do it tomorrow morning

flash1293: Switched it to an ignore list instead of an allow list on the git diff level

PR #28386, opened 2023-07-18
Low code CDK: Fix mypy errors

flash1293:
## What

Fixes typing issues in simple retriever and model to component factory

## How

A lot of fixes are very straight forward.

Some relevant ones to call out:
* I ignored failures in some places that would have required changes in other files which would have required changes in other files and so on - I left a comment in the code for these so we can easily identify the ignore ones we can roll back later on. A lot of stuff will be resolved by decoupling from http stream
* One construct I don't like but I didn't want to get into is how the `name` and `primary_key` properties are handled - it's picked up by dataclass so it can be passed in by the constructor, but then it's overwritten by a property getter. I hope we can revisit this later.
* In the component factory, one generic problem is that some stuff might not be defined on the model because it has default values - however, instantiating the class via `Component(optional_prop=model.optional_prop)` will cause a type error because `optional_prop` is defined as `optional_prop: str = "default_value"` in the component class so `None` can't be passed in. I think the best way to handle this is to define optional prop as `optional_prop: Optional[str] = None` and then apply the default value in post_init: `self._optional_prop = self.optional_prop or "default_value"`, however, this change would have required me to touch basically every component, that's why I repeated the default values in the factory for now. Maybe you know about a more elegant way to handle this problem.


airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -103,8 +103,9 @@ def raise_on_http_errors(self) -> bool:
     def max_retries(self) -> Union[int, None]:
         if self.disable_retries:
             return 0
-        if hasattr(self.requester.error_handler, "max_retries"):
-            return self.requester.error_handler.max_retries
+        # this will be removed once simple_retriever is decoupled from http_stream
```
girarda, line 106: <3

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -61,25 +61,25 @@ class SimpleRetriever(Retriever, HttpStream):
...
+    def __post_init__(self, parameters: Mapping[str, Any]) -> None:
+        self._paginator = self.paginator or NoPagination(parameters=parameters)
+        self._last_response: Optional[requests.Response] = None
+        self._records_from_last_response: List[Record] = []
         HttpStream.__init__(self, self.requester.get_authenticator())
-        self._last_response = None
-        self._records_from_last_response = None
         self._parameters = parameters
-        self.name = InterpolatedString(self._name, parameters=parameters)
+        self._name = InterpolatedString(self._name, parameters=parameters) if isinstance(self._name, str) else self._name
```
girarda, line 75: why not keep it as an interpolated string? it would avoid the need for the [if elses ](https://github.com/airbytehq/airbyte/pull/28386/files#diff-574e7f982441af95b199fa0d628677b9a654543eaea1534b7516c8679375a235R82)

airbyte-cdk/python/airbyte_cdk/sources/declarative/parsers/model_to_component_factory.py
```diff
@@ -119,23 +119,23 @@
...
-        message_repository: MessageRepository = None,
+        message_repository: Optional[MessageRepository] = None,
     ):
         self._init_mappings()
         self._limit_pages_fetched_per_slice = limit_pages_fetched_per_slice
         self._limit_slices_fetched = limit_slices_fetched
         self._emit_connector_builder_messages = emit_connector_builder_messages
         self._disable_retries = disable_retries
-        self._message_repository = message_repository or InMemoryMessageRepository(
+        self._message_repository = message_repository or InMemoryMessageRepository(  # type: ignore
```
girarda, line 133: curious: why is this an issue?

girarda (approved):
lgtm thanks for fixing these!

> how the name and primary_key properties are handled

I agree this is bad. It's not something we can easily rework because removing the property annotation would be a connector breaking change. However, we'll be able to remove it from the SimpleRetriever after we decouple it from HttpStream

> In the component factory, one generic problem

gosh... That problem would've been obvious if we were using mypy from the beginning...

> then apply the default value in post_init

I think this is the right approach. An alternative would be to turn the classes into normal classes since they do not need to be @dataclass anymore.

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
flash1293: This line makes sure that `_name` is always an interpolated string

PR #28406, opened 2023-07-18
Revert parquet/pyarrow changes until we have "extra" dependencies for file-based

maxi297:
## What
At first, we tried to change from alpine to slit. alpine requires us to compile the source for pyarrow if we want to import it which is much more complicated. The issue is that it would require us to update all the docker images that uses the CDK which is complex and tedious. 

Solution: We will have an extra dependency specifically for the file-based

## How
Only revert for now. There will be another PR for using extra dependencies

## 🚨 User Impact 🚨
Removing parquet support. Parquet changes will need to be re-introduced


maxi297:
Close as this was a temporary solution and we will move on to the full blown solution [here](https://github.com/airbytehq/airbyte/pull/28413)

PR #28413, opened 2023-07-18
Move pyarrow to CDK extra

girarda:
## What
* Remove pyarrow from the cdk main dependencies because connectors using the alpine python image cannot install it

## How
* Move the pyarrow dependency to the CDK dev and file-based extras
* Update the source_declarative image to use cdk v0.44.0


airbyte-cdk/python/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.9.11-alpine3.15 as base
+FROM python:3.9-slim as base
```
maxi297, line 1: Aren't we suppose to be able to avoid this as `RUN pip install --prefix=/install airbyte-cdk==<version>` would not require `pyarrow` anymore? I think that for this PR, it will fail since we don't update the version until there is an actual release but we can probably `approve-and-merge` bot this

maxi297: Else, we can set it to slim for this release and revert it to alpine afterward

airbyte-cdk/python/Dockerfile
```diff
@@ -10,7 +10,8 @@ RUN apk --no-cache upgrade \
     && apk --no-cache add tzdata build-base
 
 # install airbyte-cdk
-RUN pip install --prefix=/install airbyte-cdk==0.45.0
+# cdk is pinned to 0.44 because 0.45.0 includes parquet, which cannot be installed on alpine
+RUN pip install --prefix=/install airbyte-cdk==0.44.0
```
girarda, line 14: we'll need to update this manually since the publish workflow won't be able to automatically bump it

pedroslopez (approved):
Thanks for helping work this out 🙏🏼 

:shipit: !

airbyte-cdk/python/setup.py
```diff
@@ -13,6 +13,8 @@
 # The text of the README file
 README = (HERE / "README.md").read_text()
 
+pyarrow_dependency = "pyarrow==12.0.1"
```
pedroslopez, line 16: For my knowledge since I had this question when I was debugging this earlier, are we explicitly pinning to 12.0.1 for a reason or does 12.0.0 work just the same?

pedroslopez: There's a mix of ~= and == and I'm not sure what's intentional

girarda: I'd expect 12.0.0 to also work. Will revisit in a follow up PR

octavia-approvington:

octavia-approvington (approved):

PR #28442, opened 2023-07-18
Update documentation related to data residency

malikdiarra:
This brings two clarifications related to data residency:

 * Whitelisting US IP is no longer necessary for users that set their data residency to EU. All connection operation now run in the region chosen as data residency.
 * Airbyte users email and name are always stored in the US. The data residency setting only affects replication operation location.

### DIFF
--- a/docs/cloud/getting-started-with-airbyte-cloud.md
+++ b/docs/cloud/getting-started-with-airbyte-cloud.md
@@ -207,17 +207,6 @@ Depending on your [data residency](https://docs.airbyte.com/cloud/managing-airby
 * 34.106.225.141
 
 ### European Union
-
-#### GCP region: us-west3
-* 34.106.109.131
-* 34.106.196.165
-* 34.106.60.246
-* 34.106.229.69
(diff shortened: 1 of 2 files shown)


docs/cloud/getting-started-with-airbyte-cloud.md
```diff
@@ -208,12 +208,6 @@ Depending on your [data residency](https://docs.airbyte.com/cloud/managing-airby
 
 ### European Union
 
-:::note 
-
-Some workflows still run in the US, even when the data residency is in the EU. If you use the EU as a data residency, you must allowlist the following IP addresses from both GCP us-west3 and AWS eu-west-3.
-
-:::
-
 #### GCP region: us-west3
```
timroes, line 211: @malikdiarra could we also remove the US-west3 IP list from this section then, it should no longer be needed either?

malikdiarra: Done!

timroes (approved):

PR #28008, opened 2023-07-06
connectors-ci: upgrade to Dagger 0.6.4

alafanechere:
## What
Closes #27970
Closes https://github.com/airbytehq/airbyte/issues/28422

## How
🔴 Dagger 0.6.4 is not released yet. It should be released today. No pipeline should be run on this branch yet.

## Roll out plan:
**Once Dagger 0.6.4 is released**:
1. Deploy 0.6.4 on `dev-dagger-engine`: from https://github.com/airbytehq/airbyte-infra/pull/49: `kubectl apply manifests/dagger-daemonset-remote-cache.yml` 
3. Run [test](https://github.com/airbytehq/airbyte/actions/workflows/connectors_tests.yml) and [publish](https://github.com/airbytehq/airbyte/actions/workflows/publish_connectors.yml) workflow dispatch on java / python connectors from this branch targeting `dev-xlarge-runner`
4. If everything 🟢 : Deploy 0.6.4 on `dagger-engine`: from https://github.com/airbytehq/airbyte-infra/pull/50: `kubectl apply manifests/dagger-daemonset-remote-cache.yml`

PR #28450, opened 2023-07-19
Replace deprecated dagger APIs

helderco:
## What
This change replaces deprecated dagger usage with more current solutions since they will be removed soon, in https://github.com/dagger/dagger/issues/5479. It needs Python SDK >0.6.0 because of `ExecError`.

## How
Only two methods needed to be replaced. The following sections describe the minimum changes that need to be taken, however I took the opportunity to add some convenience functions to reduce boilerplate and modernized the error handling in a few cases.

### `HostVariable.secret`

The simplest change is to replace:
```python
client.host().env_variable("MY_SECRET").secret()
```
With:
```python
import os
client.set_secret("can-be-different", os.environ.get("MY_SECRET", ""))
```
Notice the default of `""`. Without it, if the var doesn't exist in the host, it'll default to `None`, which fails validation on the required `str` for the secret's plaintext.

### `Container.exit_code` and `stdout`/`stderr` from error

The simplest change is to use my modifications on `with_exit_code`, `with_stdout` and `with_stderr`, and replace any usage of `container.exit_code()` with `container.sync()`.

In addition to that, I recommend not to care about the exit code value if it's just to check if `exit_code != 0`. Just capture `ExecError` and handle the exception instead.

## Recommended reading order
1. `tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py` for the changes in `with_exit_code` and `with_stderr`, then the added convenience functions
2. `tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py` for a good example on replacing `if exit_code != 0:` with `try...except ExecError`

## 🚨 User Impact 🚨

There should be no breaking changes, unless one of the functions that changed is being used outside of this repo which I have no awareness of. See https://github.com/airbytehq/airbyte/pull/28450#discussion_r1267959386. 

I'm also not sure if the change from `str(e)` to `e.stderr` in `with_stderr` (same in `with_stdout`) breaks anything. It shouldn't but the error message is no longer there. See  https://github.com/airbytehq/airbyte/pull/28450#discussion_r1268023744.

Usually the error message is just `process "....." did not complete successfully: exit code: 1` or something like that. The more relevant information is usually in `stderr`, but better check.

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
@@ -555,7 +555,7 @@ async def load_image_to_docker_host(context: ConnectorContext, tar_file: File, i
     # Not tagged images only have a sha256 id the load output shares.
     if "sha256:" in image_load_output:
         image_id = image_load_output.replace("\n", "").replace("Loaded image ID: sha256:", "")
-        await docker_cli.with_exec(["docker", "tag", image_id, image_tag]).exit_code()
+        await docker_cli.with_exec(["docker", "tag", image_id, image_tag])
     image_sha = json.loads(await docker_cli.with_exec(["docker", "inspect", image_tag]).stdout())[0].get("Id")
     return image_sha
 
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
@@ -2,7 +2,6 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
 import functools
-import os
 import uuid
 from pathlib import Path
 from typing import Optional, Set
@@ -12,7 +11,7 @@ from ci_connector_ops.pipelines.actions.environments import with_pip_packages, w
 from ci_connector_ops.pipelines.bases import Report, Step, StepResult
 from ci_connector_ops.pipelines.contexts import PipelineContext
 from ci_connector_ops.pipelines.helpers.steps import run_steps
-from ci_connector_ops.pipelines.utils import DAGGER_CONFIG, METADATA_FILE_NAME, METADATA_ICON_FILE_NAME, execute_concurrently, get_secret_host_env_variable
+from ci_connector_ops.pipelines.utils import DAGGER_CONFIG, METADATA_FILE_NAME, METADATA_ICON_FILE_NAME, execute_concurrently, get_secret_host_variable
 
 METADATA_DIR = "airbyte-ci/connectors/metadata_service"
 METADATA_LIB_MODULE_PATH = "lib"
@@ -127,10 +126,7 @@ class DeployOrchestrator(Step):
         python_base = with_python_base(self.context)
         python_with_dependencies = with_pip_packages(python_base, ["dagster-cloud==1.2.6", "pydantic==1.10.6", "poetry2setup==1.1.0"])
         dagster_cloud_api_token_secret: dagger.Secret = (
-            self.context.dagger_client.set_secret(
-                "DAGSTER_CLOUD_METADATA_API_TOKEN",
-                os.environ.get("DAGSTER_CLOUD_METADATA_API_TOKEN"),
-            )
+            get_secret_host_variable(self.context.dagger_client, "DAGSTER_CLOUD_METADATA_API_TOKEN")
         )
 
         container_to_run = (
@@ -287,7 +283,7 @@ async def run_metadata_upload_pipeline(
     async with dagger.Connection(DAGGER_CONFIG) as dagger_client:
         pipeline_context.dagger_client = dagger_client.pipeline(pipeline_context.pipeline_name)
         async with pipeline_context:
-            get_secret = functools.partial(get_secret_host_env_variable, pipeline_context.dagger_client)
+            get_secret = functools.partial(get_secret_host_variable, pipeline_context.dagger_client)
             results = await execute_concurrently(
                 [
                     MetadataUpload(
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
@@ -1,6 +1,8 @@
 #
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
+import functools
+import os
 import uuid
 from pathlib import Path
 from typing import Optional, Set
@@ -10,7 +12,7 @@ from ci_connector_ops.pipelines.actions.environments import with_pip_packages, w
 from ci_connector_ops.pipelines.bases import Report, Step, StepResult
 from ci_connector_ops.pipelines.contexts import PipelineContext
 from ci_connector_ops.pipelines.helpers.steps import run_steps
-from ci_connector_ops.pipelines.utils import DAGGER_CONFIG, METADATA_FILE_NAME, METADATA_ICON_FILE_NAME, execute_concurrently
+from ci_connector_ops.pipelines.utils import DAGGER_CONFIG, METADATA_FILE_NAME, METADATA_ICON_FILE_NAME, execute_concurrently, get_secret_host_env_variable
 
 METADATA_DIR = "airbyte-ci/connectors/metadata_service"
 METADATA_LIB_MODULE_PATH = "lib"
@@ -125,7 +127,10 @@ class DeployOrchestrator(Step):
         python_base = with_python_base(self.context)
         python_with_dependencies = with_pip_packages(python_base, ["dagster-cloud==1.2.6", "pydantic==1.10.6", "poetry2setup==1.1.0"])
         dagster_cloud_api_token_secret: dagger.Secret = (
-            self.context.dagger_client.host().env_variable("DAGSTER_CLOUD_METADATA_API_TOKEN").secret()
+            self.context.dagger_client.set_secret(
+                "DAGSTER_CLOUD_METADATA_API_TOKEN",
+                os.environ.get("DAGSTER_CLOUD_METADATA_API_TOKEN"),
+            )
         )
 
         container_to_run = (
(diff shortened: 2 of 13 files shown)


alafanechere:
Thank you @helderco . I prepared https://github.com/airbytehq/airbyte/pull/28008 to update to 0.6.4 once it's released.

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -555,7 +555,7 @@ async def load_image_to_docker_host(context: ConnectorContext, tar_file: File, i
     # Not tagged images only have a sha256 id the load output shares.
     if "sha256:" in image_load_output:
         image_id = image_load_output.replace("\n", "").replace("Loaded image ID: sha256:", "")
-        await docker_cli.with_exec(["docker", "tag", image_id, image_tag]).exit_code()
+        await docker_cli.with_exec(["docker", "tag", image_id, image_tag])
```
alafanechere, line 558: Is `with_exec` now awaitable? INo need to use `exit_code` or `sync` to get the command executed?

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -57,6 +59,20 @@ async def check_path_in_workdir(container: Container, path: str) -> bool:
     else:
         return False
 
+def secret_host_variable(client: Client, name: str, default: str = ""):
```
alafanechere, line 62:
```suggestion
def secret_host_env_variable(client: Client, name: str, default: str = ""):
```

In `remote_storage.py` you import `secret_host_env_variable`

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -73,57 +89,55 @@ async def get_file_contents(container: Container, path: str) -> Optional[str]:
...
+        soon_stdout = task_group.soonify(container.stdout)
+        soon_stderr = task_group.soonify(container.stderr)
+    return soon_stdout.value, soon_stderr.value
+
+
+async def get_exec_result(container: Container) -> Tuple[int, str, str]:
+    try:
+        return 0, *(await get_container_output(container))
+    except ExecError as e:
+        return e.exit_code, e.stdout, e.stderr
```
alafanechere, line 108: ❤️

alafanechere (commented):
`ExecError` are great!!!

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
helderco: Yeah this is not a typo 🙂 It's actually `Container` that's awaitable:
```python
# execute container and return it or raise exception on error
await container

# same as
await container.sync()
```
I like the simple syntax of doing `await` on a lazy object (i.e., `Container`) to execute it. I feel the `.sync()` is redundant here, that's why I tend to drop it. It still returns the same container which allows chaining:
```python
# short circuit on error
out = await (await container).with_exec([...]).stdout()
```

This works for any type that has a `.sync()` method by the way. You can do stuff like this:
```python
# poor man's concurrency 😂
await (
    client.directory()
    .with_directory("/test", test_ctr.rootfs())
    .with_directory("/lint", lint_ctr.rootfs())
)
```

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
helderco: I went back and forth on this name 😅

tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
alafanechere: Nice! So whenever we want to run `with_exec` commands and not care about the `stdout / stderr` we can directly await the Container

tools/ci_connector_ops/ci_connector_ops/pipelines/pipelines/metadata.py
```diff
@@ -282,17 +283,14 @@ async def run_metadata_upload_pipeline(
     async with dagger.Connection(DAGGER_CONFIG) as dagger_client:
         pipeline_context.dagger_client = dagger_client.pipeline(pipeline_context.pipeline_name)
         async with pipeline_context:
-            gcs_credentials_secret: dagger.Secret = pipeline_context.dagger_client.host().env_variable("GCS_CREDENTIALS").secret()
-            docker_hub_username_secret: dagger.Secret = pipeline_context.dagger_client.host().env_variable("DOCKER_HUB_USERNAME").secret()
-            docker_hub_password_secret: dagger.Secret = pipeline_context.dagger_client.host().env_variable("DOCKER_HUB_PASSWORD").secret()
-
+            get_secret = functools.partial(get_secret_host_variable, pipeline_context.dagger_client)
```
helderco, line 286: Just a bit of sugar. `get_secret("DOCKER_HUB_USER_NAME")` is the same as:
```python
get_secret_host_variable(pipeline_context.dagger_client, "DOCKER_HUB_USERNAME")
```

tools/ci_connector_ops/setup.py
```diff
@@ -45,7 +45,7 @@ def local_pkg(name: str) -> str:
 ]
 
 PIPELINES_REQUIREMENTS = [
-    "dagger-io==0.5.4",
+    "dagger-io==0.6.4",
```
helderco, line 48: This release is coming out in just a few hours. The minimum needed for these changes is `0.6.0`.

tools/ci_connector_ops/ci_connector_ops/pipelines/publish.py
```diff
@@ -30,14 +29,13 @@ async def _run(self) -> StepResult:
...
-        if crane_ls_exit_code != 0:
-            if "NAME_UNKNOWN" in crane_ls_stderr:
+        try:
+            crane_ls_stdout = await crane_ls.stdout()
+        except ExecError as e:
+            if "NAME_UNKNOWN" in e.stderr:
                 return StepResult(self, status=StepStatus.SUCCESS, stdout=f"The docker repository {docker_repository} does not exist.")
             else:
-                return StepResult(self, status=StepStatus.FAILURE, stderr=crane_ls_stderr, stdout=crane_ls_stdout)
+                return StepResult(self, status=StepStatus.FAILURE, stderr=e.stderr, stdout=e.stdout)
```
helderco, line 38: This is another good example of replacing `if exit_code != 0`.

Notice how in a success condition, only stdout is needed here, that's why all we need is a simple `await crane_ls.stdout()` inside `try` instead of using the utils above.

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -73,57 +110,78 @@ async def get_file_contents(container: Container, path: str) -> Optional[str]:
...
+        soon_stderr = task_group.soonify(container.stderr)
+    return soon_stdout.value, soon_stderr.value
+
+
+async def get_exec_result(container: Container) -> Tuple[int, str, str]:
+    """Retrieve the exit_code along with stdout and stderr of a container by handling the ExecError.
+
+    Note: It is preferrable to not worry about the exit code value and just capture
+    ExecError to handle errors. This is offered as a convenience when the exit code
+    value is actually needed.
```
helderco, line 141: This explains why I also removed the need for `with_exit_code` in a few places. My preference would be to avoid this function and handle the `ExecError` where needed.

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -73,57 +110,78 @@ async def get_file_contents(container: Container, path: str) -> Optional[str]:
...
+    Returns:
+        Tuple[str, str]: The stdout and stderr of the container, respectively.
+
+    Raises:
+        ExecError: If the container exit code is not 0.
+    """
+    async with asyncer.create_task_group() as task_group:
+        soon_stdout = task_group.soonify(container.stdout)
+        soon_stderr = task_group.soonify(container.stderr)
+    return soon_stdout.value, soon_stderr.value
```
helderco, line 133: This convenience is for when you need both `stdout` and `stderr`. This may not be necessary in a future release if this gets solved:
- https://github.com/dagger/dagger/issues/3496

It's preferable to use `container.stdout()` directly if you don't need the other and vice versa.

Note that this doesn't handle the `ExecError` because I prefer handling it where it's needed. That's why there's `get_exec_result`, it not only returns `exit_code` but also handles `ExecError` if that's desirable.

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -73,57 +110,78 @@ async def get_file_contents(container: Container, path: str) -> Optional[str]:
...
-            exit_code = re.search(r"exit code: (\d+)", error_message)
-            if exit_code:
-                return int(exit_code.group(1))
-            else:
-                return 1
-        raise
+        await container
+    except ExecError as e:
+        return e.exit_code
     return 0
```
helderco, line 168: You can see basic usage in the feature PR:
- https://github.com/dagger/dagger/pull/5184

Note that the lone `raise` still works because this will still raise a `QueryError` on an unexpected network connection failure (for example) since we're only capturing `ExecError` here. It's the equivalent of the `if "exit code: " in error_message` block.

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -73,57 +110,78 @@ async def get_file_contents(container: Container, path: str) -> Optional[str]:
...
-# The original issue is tracked here https://github.com/dagger/dagger/issues/3192
 async def with_stderr(container: Container) -> str:
-    """Retrieve the stderr of a container and handle unexpected errors."""
+    """Retrieve the stderr of a container even on execution error."""
     try:
         return await container.stderr()
-    except QueryError as e:
-        return str(e)
+    except ExecError as e:
+        return e.stderr
```
helderco, line 176: Before 0.6.0, the error message from the API would include `Stdout:\n{stdout}` and `Stderr:\n{stderr}` lines. This was removed in the API from the error message into extension fields, which are only read by `ExecError` now. 

Additionally:
- `str(e)` on a `QueryError` would previously return the whole error message, exit code, stdout and stderr. Now it only returns the error message.
- As a convenience, if you do `str(e)` on an `ExecError` you'll get the same output as before (i.e., included "Stdout:" and "Stderr:" lines).

**Note** that I've changed from `str(e)` to `e.stderr` since that seems to be the original intent. Otherwise, when you do this:
```python
stdout = await with_stdout(ctr)
stderr = await with_stderr(ctr)
return StepResult(
    ...
    stdout=stdout,
    stderr=stderr,
)
```
You'd get the same output in both `stdout` and `stderr` instead of just what you intend (in case of error).

**However**, now you don't have the **error message** if you just return `e.stderr`. If you need it in the result for the report, I suggest adding a field for it in `StepResult`, or prepending `e.message` to `e.stderr`, whatever makes the most sense for you at the moment.

If you want to return to the old behavior, just replace `QueryError` with `ExecError` and keep `str(e)`:
```python
try:
    return await container.stderr()
except ExecError as e:
    return str(e)
```

helderco (commented):
Added some comments to help explain a few things. I'm going to work on a documentation guide on error handling to bring this home. 🙂

By the way, I haven't tested locally since at least last time I didn't have a few necessary things locally, but let me know if it's simple now for me to test locally. I'm relying on CI for now.

bnchrch (approved):
Thanks @helderco! Really appriciate the help in getting our pipelines back in working order and loved the teaching and helper functions along the way,

octavia-approvington:

octavia-approvington (approved):

tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -73,57 +110,78 @@ async def get_file_contents(container: Container, path: str) -> Optional[str]:
...
+        container (Container): The container to execute.
+
+    Returns:
+        Tuple[str, str]: The stdout and stderr of the container, respectively.
+
+    Raises:
+        ExecError: If the container exit code is not 0.
+    """
+    async with asyncer.create_task_group() as task_group:
+        soon_stdout = task_group.soonify(container.stdout)()
```
alafanechere, line 131: `ExecError` in running `container.stdout` or `container.stderr` will raise `ExceptionGroup` and not `ExecError`. 
The upstream code calling `get_container_output` is handling `ExecError` not `ExceptionGroup`.

The original `with_exit_code` , `with_stdout` etc. Captured the QueryError themselves and returned their string representation, so no `ExceptionGroup` were ever thrown in case of a `QueryError`.

Issue #28133, opened 2023-07-11
File CDK: S3 config adapter (CSV options)


clnoll:

clnoll:
Grooming notes:
- To avoid breaking OSS connectors we should handle all options instead of trying to only support the existing ones.

girarda:
End of sprint update: 
- [PR](https://github.com/airbytehq/airbyte/pull/28491) for missing options except infer_datatypes
- infer_datatypes can be tackled in a separate PR

PR #28491, opened 2023-07-19
Add CSV options to the CSV parser

girarda:
## What
* Handle the following CSV options that are supported by pyarrow
  * true_values: Define which string values convert to `True`
  * false_values: Define which string values convert to `False`
  * skip_rows (renamed to skip_rows_before_header): Number of rows to skip before header
  * skip_rows_after_names (renamed to skip_rows_after_header): Number of rows to skip after the header ([the ticket ](https://github.com/airbytehq/airbyte/issues/28133)says we can combine these two into a single config field, but pyarrow's skip_rows means "skip rows before header, not a global number of rows to skip)
  * newlines in values (only valid if the values are quoted). New lines in non-quoted strings will result in unparseable records
  * autogenerate_column_names
* Added tests for
  * strings_can_be_null (always true)

These options are not handled. I bolded the ones that would cause breaking changes:
- infer_datatypes. We'll need to write our own type detection. That can/should be done separately to make reviewing easier (this PR is already quite large)
- compression: This will need to be implemented at the source level as part of the file reader
- encoding: This will need to be implemented at the source level as part of the file reader
- include_missing_columns: This field doesn't make sense since we're always supporting schema evolution. The fields will show up as null in the destination if they are not set
- column_types: Users can pass an input schema if they want to decide on the schema
- check_utf8: This should be part of the encoding (not sure we'll need to support this though)
- **decimal_point**: Seems fairly low value to allow users to use any character as decimal point
- **quoted_strings_can_be_null**: Always setting this to true seems like a reasonable default
- include_columns: Users can use column selection to decide what to sync
- **auto_dict_encode**: I don’t know what this means in a non-pyarrow world. There is no encoded dicts type. This will be a breaking change
- **auto_dict_max_cardinality**: Same as above. There is no dict type since we're not converting to a pyarrow object
- **timestamp_parsers**: We write the timestamps as strings instead of converting them into pyarrow timestamp objects

## How
* Add config fields to the CsvFormat object: null_values, skip_rows_before_header, skip_rows_after_header, autogenerate_column_names, true_values, false_values
* Update the `FileTypeParser.parse_records` so it can return None if a row cannot be parsed
* Update the `CsvParser.parse_records` so it returns None if a row cannot be parsed
* Update `CsvParser.parse_records` to skip rows before and after the header if specified in the config
* Update CsvParse to autogenerate the header if specified in the config
* Update the cast_types method to convert bool and null values according to the config
* Update DefaultFileBasedStream to handle `None` records. These records should never pass the validation. The skip policy return False while the others raise an Exception)

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/file_based/config/csv_format.py`
2. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/file_type_parser.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py`
4. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/jsonl_parser.py`
5. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/parquet_parser.py`
6. `airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_validation_policies/abstract_schema_validation_policy.py`
7. `airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_validation_policies/default_schema_validation_policies.py`
8. `airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/abstract_file_based_stream.py`
9. `airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py`


airbyte-cdk/python/unit_tests/sources/file_based/scenarios/validation_policy_scenarios.py
```diff
@@ -231,8 +231,8 @@
             },
             {
                 "level": "WARN",
-                "message": "Records in file did not pass validation policy. stream=stream1 file=c.csv n_skipped=1 validation_policy=skip_record",
-            },
+                "message": "Records in file did not pass validation policy. stream=stream1 file=c.csv n_skipped=1 validation_policy=skip_record"
```
girarda, line 234: c.csv is also misformated

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -922,7 +941,6 @@
                             "escape_char": "@",
                             "double_quote": True,
                             "newlines_in_values": False,
-                            "quoting_behavior": "Quote All"
```
girarda, line 925: removed this option from the test since it was not impacting the output

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -84,6 +84,25 @@ def read_records_from_slice(self, stream_slice: StreamSlice) -> Iterable[Mapping
             try:
                 for record in parser.parse_records(self.config, file, self._stream_reader, self.logger):
                     line_no += 1
+                    if record is None:
```
clnoll, line 87: I have some hesitations about letting `parse_records` return `None`. If it's indicative of an invalid file, I feel like it would be more intuitive to raise an exception and stop trying to parse the file, and think that would more closely mirror what the existing S3 connector does. WDYT?

airbyte-cdk/python/airbyte_cdk/sources/file_based/schema_validation_policies/default_schema_validation_policies.py
```diff
@@ -4,30 +4,34 @@
...
         return True
 
 
 class SkipRecordPolicy(AbstractSchemaValidationPolicy):
     name = "skip_record"
 
-    def record_passes_validation_policy(self, record: Mapping[str, Any], schema: Optional[Mapping[str, Any]]) -> bool:
-        return schema is not None and conforms_to_schema(record, schema)
+    def record_passes_validation_policy(self, record: Optional[Mapping[str, Any]], schema: Optional[Mapping[str, Any]]) -> bool:
+        return schema is not None and record is not None and conforms_to_schema(record, schema)
```
clnoll, line 25: Even with the `skip_record` policy, I think that a `None` record would be invalid, if it's indicative of an issue with parsing the file.

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -1531,3 +1636,1175 @@
...
+    )
+    .set_expected_records(
+        [
+            {"data": {"col1": "val11", "col2": "val12", "_ab_source_file_last_modified": "2023-06-05T03:54:07Z",
+                      "_ab_source_file_url": "a.csv"}, "stream": "stream1"},
+        ]
+    )
+).build()
+
+csv_skip_after_header_scenario = (
```
clnoll, line 2449: Looks like this test's name is a duplicate of the test above, but its contents are a duplicate of the tests below (`csv_skip_before_and_after_header_scenario`).

airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
```diff
@@ -34,10 +34,24 @@
     success_user_provided_schema_scenario,
 )
 from unit_tests.sources.file_based.scenarios.csv_scenarios import (
+    csv_custom_bool_values_scenario,
```
clnoll, line 37:
```suggestion
    csv_autogenerate_column_names_scenario,
    csv_custom_bool_values_scenario,
```

clnoll (commented):
Nice @girarda! A couple of questions, mostly about whether we should be returning `Optional` values from `parse_records`.

airbyte-cdk/python/airbyte_cdk/sources/declarative/models/declarative_component_schema.py
```diff
@@ -823,7 +823,7 @@ class DatetimeBasedCursor(BaseModel):
     cursor_datetime_formats: Optional[List[str]] = Field(
         None,
         description="The possible formats for the cursor field",
-        title="Cursor Datetime Format",
+        title="Cursor Datetime Formats",
```
girarda, line 826: this file was modified by the autocommit step in CI

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/AirbyteInternal.yaml
```diff
@@ -23,4 +23,4 @@ properties:
       - 300
       - 400
       - 500
-      - 600
\ No newline at end of file
+      - 600
```
girarda, line 26: this file was autocomitted by CI

airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
girarda: derp. added the test

airbyte-cdk/python/unit_tests/sources/file_based/config/test_csv_format.py
```diff
@@ -0,0 +1,24 @@
...
+        pytest.param(0, True, ValidationError, id="test_no_skip_rows_before_header_and_autogenerate_column_names"),
+        pytest.param(0, False, ValidationError, id="test_no_skip_rows_before_header_and_no_autogenerate_column_names"),
+    ]
+)
+def test_csv_format(skip_rows_before_header, autogenerate_column_names, expected_error):
+    if expected_error:
+        with pytest.raises(ValidationError):
+            CsvFormat(skip_rows_before_header=1, autogenerate_column_names=True)
+    else:
+        CsvFormat(skip_rows_before_header=1, autogenerate_column_names=True)
```
clnoll, line 24:
```suggestion
        with pytest.raises(ValidationError):
            CsvFormat(skip_rows_before_header=skip_rows_before_header, autogenerate_column_names=autogenerate_column_names)
    else:
        CsvFormat(skip_rows_before_header=skip_rows_before_header, autogenerate_column_names=autogenerate_column_names)
```

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
```diff
@@ -65,54 +66,94 @@ def parse_records(
...
+        reader: csv.DictReader, schema: Optional[Mapping[str, Any]], config_format: CsvFormat, logger: logging.Logger  # type: ignore
+    ) -> Iterable[Optional[Dict[str, Any]]]:
         """
         If the user provided a schema, attempt to cast the record values to the associated type.
 
         If a column is not in the schema or cannot be cast to an appropriate python type,
         cast it to a string. Downstream, the user's validation policy will determine whether the
         record should be emitted.
         """
-        if not schema:
```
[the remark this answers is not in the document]
clnoll: With `if schema` is still skipping row skipping & validation though, right? Should we be getting rid of the `if` altogether?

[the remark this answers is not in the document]
girarda: the skip row and validation is done after the if/else.

The if/else only controls how the fields are casted (either using cast_type, or return as-is)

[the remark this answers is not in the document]
girarda: extracted to a separate function to make this clearer

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
```diff
@@ -178,5 +221,17 @@ def cast_types(row: Dict[str, str], property_types: Dict[str, Any], logger: logg
     return result
 
 
+def _value_to_bool(value: str, true_values: List[str], false_values: List[str]) -> bool:
```
maxi297, line 224: Should those be sets instead of lists for performance purposes?

maxi297, line 224: Should this check be case-insensitive? Thinking about this, I feel it provides more flexibility if not but the UX might be annoying if I want to validate "true" as case insensitive (i.e. I'll need to provide "True", "TRUE", tRue", "tRUe", etc...)

The same question probably applies to null values

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
```diff
@@ -107,16 +101,65 @@ def _read_and_cast_types(
...
+        """
+        next_line = next(fp).strip()
+        number_of_columns = len(next_line.split(config_format.delimiter))  # type: ignore
+        # Reset the file pointer to the beginning of the file so that the first row is not skipped
+        fp.seek(0)
+        return [f"f{i}" for i in range(number_of_columns)]
 
-def cast_types(row: Dict[str, str], property_types: Dict[str, Any], logger: logging.Logger) -> Dict[str, Any]:
+
+def cast_types(row: Dict[str, str], property_types: Dict[str, Any], config_format: CsvFormat, logger: logging.Logger) -> Dict[str, Any]:
```
maxi297, line 162: I know this was like this before but why is this public? It does not seem to be be used elsewhere

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
```diff
@@ -107,16 +101,65 @@ def _read_and_cast_types(
...
         record should be emitted.
         """
-        if not schema:
-            yield from reader
+        cast_fn = CsvParser._get_cast_function(schema, config_format, logger)
+        for i, row in enumerate(reader):
+            if i < config_format.skip_rows_after_header:
+                continue
+            # The row was not properly parsed if any of  the values are None
+            if any(val is None for val in row.values()):
```
maxi297, line 109: Do we have a test for this? It seems odd and I feel it would be worth documenting

maxi297 (approved):
LGTM! For my personal knowledge, do we need to cast values within a dict or a list?

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
girarda: I think they should be case-sensitve for backwards compatibility. We can add the option for case-insensitivity if needed.

Documented that the strings are case-sensitive in the spec

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py
girarda: fixed

girarda:
> LGTM! For my personal knowledge, do we need to cast values within a dict or a list?

Excellent question!

I played around with pyarrow to and conversions of List[bool] and map[str, bool] aren't supported so legacy s3 converted these to strings.

> pyarrow.lib.ArrowNotImplementedError: CSV conversion to list<item: bool> is not supported
> pyarrow.lib.ArrowNotImplementedError: CSV conversion to list<item: bool> is not supported

In general, we shouldn't have to worry about casting nested values since destination v2 won't normalize the data

PR #28500, opened 2023-07-20
[file-based cdk] Add avro parser for inferring schema and reading records

brianjlai:
## What

Adds functionality so that the `AvroParser` can now read an avro file and infer a schema or emit records.

## How

Similar to other file formats, this PR just adds implementations to getting schemas or reading records. A majority of the complexity is in how we translate Avro types into JSON schema types. As a rough summary, there are three classes of types
- Primitive: These are pretty vanilla and mostly rely on a flat mapping of avro -> json schema equivalents
- Complex: These require some bespoke logic to build a type that is representative of Avro. Also for these types we need to recursively construct types because nested fields could be other Avro types.
- Logical: Also a bit complex because JSON schema is often less expressive than Avro so we try to do a best effort at using pattern matching as a surrogate for validations. I opted not to introduce super complex type validation to err on the side of letting more records through. Happy to reverse this if we want to get really granular on some of the date patterns. This was also not supported in the original S3 destination but figured we might as well support all we can.

For more detail about the types: https://avro.apache.org/docs/1.11.1/specification/

As for reading records, this is actually more straightforward because `fastavro` is already emitting records as a dictionary so we don't have to do a lot of complex record remapping.

## Recommended reading order
1. `avro_parser.py`
2. `test_avro_parser.py`
3. `in_memory_files_source.py`
4. `avro_scenarios.py`

The PR looks absolutely massive, but a chunk of this is tests and reformatting of the `csv_scenarios.py` was mostly automation.


airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -19,7 +19,7 @@
                     "validation_policy": "emit_record",
                 }
             ],
-            "start_date": "2023-06-04T03:54:07Z"
+            "start_date": "2023-06-04T03:54:07Z",
```
brianjlai, line 22: This file looks crazy, but I ran formatting which changes a ton of stuff which might make this a bit hard to review.  To summarize the actual changes I made, I did some reordering of the file formats and adding `AvroFormat` which had to then be added to the scenario used to verify Spec. The rest remained mostly the same

airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
```diff
@@ -229,7 +239,7 @@ def _verify_read_output(output: Dict[str, Any], scenario: TestScenario) -> None:
         if "record" in actual:
             for key, value in actual["record"]["data"].items():
                 if isinstance(value, float):
-                    assert math.isclose(value, expected["data"][key], abs_tol=1e-06)
+                    assert math.isclose(value, expected["data"][key], abs_tol=1e-04)
```
brianjlai, line 242: This one was interesting. There are some rounding errors when we try to write to Avro files in the test. I don't believe this is an issue reading, but rather when we have a Python float, we have to write it to an Avro value when creating the file. The resulting value read in has around 4ish trailing "close" values. So I loosened the accuracy a bit vs what I saw in my tests

https://stackoverflow.com/questions/45827782/python-fastavro-incorrectly-converting-float

airbyte-cdk/python/unit_tests/sources/file_based/in_memory_files_source.py
```diff
@@ -139,3 +145,27 @@ def _create_file(self, file_name: str) -> bytes:
...
             return fp.read()
+
+
+class TemporaryAvroFilesStreamReader(InMemoryFilesStreamReader):
+    """
+    A file reader that writes RemoteFiles to a temporary file and then reads them back.
+    """
+
+    def open_file(self, file: RemoteFile) -> IOBase:
+        return io.BytesIO(self._make_file_contents(file.uri))  # am i returning the right type
```
girarda, line 156: looks right to me

airbyte-cdk/python/setup.py
```diff
@@ -49,8 +49,10 @@
     package_data={"airbyte_cdk": ["py.typed", "sources/declarative/declarative_component_schema.yaml"]},
     install_requires=[
         "airbyte-protocol-models==0.3.6",
+        "avro~=1.11.2",
```
girarda, line 52: can we add these dependencies to the `file-based` and `dev` and dev modules instead of the main? I don't know if we'll have the same issues we had with parquet, but there's no need to add the avro library to all connectors

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/avro_parser.py
```diff
@@ -3,13 +3,35 @@
...
 from airbyte_cdk.sources.file_based.file_types.file_type_parser import FileTypeParser
 from airbyte_cdk.sources.file_based.remote_file import RemoteFile
 
+AVRO_TYPE_TO_JSON_TYPE = {
+    "null": "null",
+    "boolean": "boolean",
+    "int": "integer",
+    "long": "integer",
+    "float": "number",
+    "double": "string",  # double -> number conversions can lose precision
```
girarda, line 20: I think this will need to be configurable to be backwards compatible with existing avro sources. 

For parquet, we added a config field that [defaults to converting to strings](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/file_based/config/parquet_format.py#L12), but [default to numbers if the config is legacy](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/file_based/config/file_based_stream_config.py#L86-L99).

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/avro_parser.py
```diff
@@ -19,7 +41,68 @@ async def infer_schema(
...
+        }
+        return json_schema
+
+    @classmethod
+    def _convert_avro_type_to_json(cls, field_name: str, avro_field: str) -> Mapping[str, Any]:
+        if isinstance(avro_field, Mapping) and avro_field["type"] == "record":
+            return {
+                "type": "object",
+                "properties": {
+                    object_field["name"]: {**AvroParser._convert_avro_type_to_json(object_field["name"], object_field["type"])}
```
girarda, line 60: why do we need the unpacking? Isn't this equivalent to

```
"properties": {
                    object_field["name"]: AvroParser._convert_avro_type_to_json(object_field["name"], object_field["type"])
                    for object_field in avro_field["fields"]
                },
```

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/avro_parser.py
```diff
@@ -19,7 +41,68 @@ async def infer_schema(
...
+            if "precision" not in avro_field:
+                raise ValueError(f"{field_name} decimal type does not have a required field precision")
+            if "scale" not in avro_field:
+                raise ValueError(f"{field_name} decimal type does not have a required field scale")
+            max_whole_number_range = avro_field["precision"] - avro_field["scale"]
+            decimal_range = avro_field["scale"]
+
+            # This regex looks like a mess, but it is validation for at least one whole number and optional fractional numbers
+            # For example: ^-?\d{1,5}(?:\.\d{1,3})?$ would accept 12345.123 and 123456.12345 would  be rejected
+            return {"type": "string", "pattern": f"^-?\\d{{{1,max_whole_number_range}}}(?:\\.\\d{1,decimal_range})?$"}
```
girarda, line 97: why do we care about encoding this in the schema? Is it to keep track of the precision?

airbyte-cdk/python/unit_tests/sources/file_based/file_types/test_avro_parser.py
```diff
@@ -0,0 +1,137 @@
...
+
+import pytest
+from airbyte_cdk.sources.file_based.file_types import AvroParser
+
+
+@pytest.mark.parametrize(
+    "avro_type, expected_json_type, expected_error",
+    [
+        # Primitive types
+        pytest.param("null", {"type": "null"}, None, id="test_null"),
```
clnoll, line 13: If we add handling for the types mentioned in my other comment we'll want to add tests.

clnoll (commented):
Just some minor comments @brianjlai. Also as @girarda mentioned, for floats I think we'll want to do something similar to what we did with parquet and continue treating them as numbers for existing syncs. Feel free to do it in a separate PR though.

airbyte-cdk/python/setup.py
brianjlai: good point will do!

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/avro_parser.py
brianjlai: I don't think it's encoding but rather because the avro's decimal logical type has the concept of precision and scale which tracks how many decimals. We can add that layer of validation with the pattern

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/avro_parser.py
```diff
@@ -3,13 +3,35 @@
...
+}
+
+AVRO_LOGICAL_TYPE_TO_JSON = {
+    "decimal": {"type": "string"},
+    "date": {"type": "string", "format": "date"},
+    "time-millis": {"type": "integer"},
+    "time-micros": {"type": "integer"},
+    "timestamp-millis": {"type": "string", "format": "date-time"},
+    "timestamp-micros": {"type": "string"},
+    # fastavro does not support duration https://fastavro.readthedocs.io/en/latest/logical_types.html
```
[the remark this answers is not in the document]
brianjlai: adding!

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/avro_parser.py
brianjlai: fair point yeah i can add a similar thing to avro as well. Sucks have the extra overhead but backwards compatibility is more important at the moment

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/avro_parser.py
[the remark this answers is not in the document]
brianjlai: also noting, I spent a decent amount of time trying to figure out why datetime objects don't work with our avro io writer since tests complaining about the type not matching on file writes. For now I'll punt on this because it seems uncommon, logical types werent supported in the legacy s3, and this was only affecting our ability to test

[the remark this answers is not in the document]
clnoll: Thanks @brianjlai. I'm tempted to hold off on supporting logical types and throw an exception if we encounter one, until we're ready to dedicate more time to it. WDYT?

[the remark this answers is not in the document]
brianjlai: On all logical types or just these two? I feel like most of the logical types everywhere else work pretty well and i've been able to test them with scenarios so I'm actually in favor leaving them in.

clnoll (approved):
Going to go ahead and approve in case you want to address https://github.com/airbytehq/airbyte/pull/28500/files#r1269937475 in a separate PR.

brianjlai:
>Going to go ahead and approve in case you want to address https://github.com/airbytehq/airbyte/pull/28500/files#r1269937475 in a separate PR.

Actually this PR contains the fix to address that as well by using an additional config option: https://github.com/airbytehq/airbyte/pull/28500/files#diff-5b1e3266091dac95c0762aa415b07bbff4ce9085898325e5c97b168bbac5d1b4R13-R17

clnoll:
Awesome, thanks for pointing that out @brianjlai. LGTM!

PR #28518, opened 2023-07-20
 Fix remove field transform

maxi297:
## What
Remove field transformation is broken because we pass a "Mapping" but dpath supports dict and not Mapping

## How
Perform transformation on the dict before casting in the Mapping

## 🚨 User Impact 🚨
This should fix the transformation "RemoveField" which is broken right now

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/transformations/transformation.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/transformations/transformation.py
@@ -4,9 +4,9 @@
 
 from abc import abstractmethod
 from dataclasses import dataclass
-from typing import Optional
+from typing import Any, Mapping, Optional
 
-from airbyte_cdk.sources.declarative.types import Config, Record, StreamSlice, StreamState
+from airbyte_cdk.sources.declarative.types import Config, StreamSlice, StreamState
 
 
 @dataclass
@@ -18,11 +18,11 @@ class RecordTransformation:
     @abstractmethod
     def transform(
         self,
-        record: Record,
+        record: Mapping[str, Any],
         config: Optional[Config] = None,
         stream_state: Optional[StreamState] = None,
         stream_slice: Optional[StreamSlice] = None,
-    ) -> Record:
+    ) -> Mapping[str, Any]:
         """
         Transform a record by adding, deleting, or mutating fields.
 
@@ -33,5 +33,5 @@ class RecordTransformation:
         :return: The transformed record
         """
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
         return other.__dict__ == self.__dict__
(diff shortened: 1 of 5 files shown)


airbyte-cdk/python/airbyte_cdk/sources/declarative/transformations/transformation.py
```diff
@@ -18,11 +18,11 @@ class RecordTransformation:
     @abstractmethod
     def transform(
         self,
-        record: Record,
+        record: Mapping[str, Any],
         config: Optional[Config] = None,
         stream_state: Optional[StreamState] = None,
         stream_slice: Optional[StreamSlice] = None,
-    ) -> Record:
+    ) -> Mapping[str, Any]:
```
maxi297, line 25: I would have updated `AddFields` as well but it meant having to fix interpolation things which brought a lot more MyPy errors so I stopped there

airbyte-cdk/python/airbyte_cdk/sources/declarative/types.py
```diff
@@ -16,43 +16,38 @@
...
     @property
-    def associated_slice(self) -> StreamSlice:
+    def associated_slice(self) -> Optional[StreamSlice]:
         return self._associated_slice
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return repr(self._data)
 
-    def __setitem__(self, key: str, value: Any):
```
maxi297, line 34: This was used to support transformation before but as this is now really considered a "Mapping" (transformation occurs on the dict and not the Record), we can remove the set

airbyte-cdk/python/airbyte_cdk/sources/declarative/types.py
```diff
@@ -16,43 +16,38 @@
...
-    def __iter__(self):
+    def __iter__(self) -> Any:
         return iter(self._data)
 
-    def __contains__(self, item: str):
+    def __contains__(self, item: object) -> bool:
         return item in self._data
 
-    def __eq__(self, other):
-        if isinstance(other, dict):
```
maxi297, line 50: Tests were passing before because we allowed the comparison with a dict while those have two very different interfaces

airbyte-cdk/python/airbyte_cdk/sources/declarative/transformations/transformation.py
```diff
@@ -18,11 +18,11 @@ class RecordTransformation:
     @abstractmethod
     def transform(
         self,
-        record: Record,
+        record: Mapping[str, Any],
```
clnoll, line 21: Are we concerned about anyone inheriting from `RecordTransformation` and implementing this method with the assumption that they'll get a `Record` (i.e. that they'll have access to the `data` and `associated_slice` methods)?

clnoll (approved):
@maxi297 have you verified that there aren't any connectors overriding, `RecordFilter` and `RecordTransformation` that might be expecting a record (and therefore its methods)? If so LGTM.

maxi297:
I can confirm. Only source-chargebee, source-facebook-pages and source-prestashop implements a RecordTransformation and they don't use the Record specific fields. I don't see any implementation of RecordFilter so this should be fine as well

Issue #21155, opened 2023-01-09
Low-code connector spec must be duplicated for SATs


girarda:
## Tell us about the problem you're trying to solve
* The spec for a low-code connector can be embedded in the connector manifest
* The SATs expect the spec to live in a separate file
* The connector spec must therefore be duplicated for the SATs to pass

## Describe the solution you’d like
* Connector developers shouldn't need to duplicate the spec for the SATs to pass

## Alternative
* Better documentation so connector developers know to duplicate the spec

## Additional context
* This causes a lot of confusion to new developers because the tutorial (rightfully) tells them to embed the spec in the manifest ([example thread](https://airbytehq-team.slack.com/archives/C027KKE4BCZ/p1672778698281269))

PR #28519, opened 2023-07-20
Allow for sources without spec.json or spec.yaml

maxi297:
## What
The extensibility team would like for the spec to be pushed completely to the manifest.yaml to allow for the Connector Builder to import sources more easily. However, CATs are failing if a spec is not provided. The only reason is that it compares the spec returned by the connector with the spec.json/spec.yaml file to ensure they are the same. We have agreed that if a connector needs this validation, it can do it as part of a unit/integration test.

Also relates to pain points users had: https://github.com/airbytehq/airbyte/issues/21155

## How
Have the connector_spec fixture return None on FileNotFoundError and use actual_connector_spec for spec validation

## 🚨 User Impact 🚨
Allows connector maintainer through the Connector Builder to delete spec.json/spec.yaml and rely only on the manifest.yaml


pedroslopez (approved):
Swapping the usages of the connector spec to the actual one rather than the file makes total sense to me! 

Also seems fine to make the assertion optional (though perhaps it should be required for high strictness level - ok to leave as is for now)

Do we need to update some docs to reflect that this is now optional?

maxi297:
Thanks for the review @pedroslopez ! I checked the documentation and don't see any place where it says that spec.json/spec.yaml is mandatory. It says that it's how the AbstractSource finds the spec but more really. Do you have a specific section I should double check?

maxi297:
> Also seems fine to make the assertion optional (though perhaps it should be required for high strictness level - ok to leave as is for now)

Which assertion are you referring to? The one where the `actual_connector_spec == connector_spec`? For no-code connector, this would never apply as spec.json/spec.yaml will not exist. Does that mean that we will never enable the high strictness level for those?

girarda (approved):

maxi297:

maxi297:
/legacy-update connector=bases/connector-acceptance-test auto-bump-version=false

maxi297:
/legacy-publish connector=bases/connector-acceptance-test auto-bump-version=false

PR #28539, opened 2023-07-20
First iteration of the maintenance through Connector Builder handbook

maxi297:
## What
The goal is to enable SE to start maintaining some connectors. The way to achieve this is by sharing the process with them so that they have some basic knowledge as to how the maintenance should be done. The information will be presented to them next Monday

## How
Added a page in the documentation.

## 🚨 User Impact 🚨
None as this is a knowledge dump

### DIFF
--- a/docs/connector-development/connector-builder-ui/maintenance.md
+++ b/docs/connector-development/connector-builder-ui/maintenance.md
@@ -1,0 +1,175 @@
+# Maintenance through Connector Builder
+
+This document will describe the process through which you can maintain a connector publicly available though our GitHub repository. This process has been split up in five broad steps:
+* Build
+* Merge into code
+* Validation (CATs)
+* Release
+* Operate/Maintain
+
+The visualization of this process is loosely maintained [here](https://whimsical.com/connector-maintenance-06-23-Rh2xAX5Byw641Xpf7P7qNg@2bsEvpTYSt1Hiz1WXyhZwMwZTwhXRsM2Sv8).
+
+## Build
+
+Goal: Modify the behavior of the connector
+
+Expected output: A manifest.yaml file that represents this new behavior
+
+All the steps mentioned below assume that you have the API available to perform requests. 
+
+### Validate Compatibility of Added Features
+Before implementing anything, we need to know if the features you want to add are supported by the Connector Builder. Please refer to the [compatibility guide](./connector-builder-compatibility.md) to confirm if you can build this change within the Connector Builder.
+
+### Config in Connector Builder
+
+Given that the connector already exists and that you want to use the account configured for CATs to iterate during the build phase, you can use [GSM](https://console.cloud.google.com/security/secret-manager?project=dataline-integration-testing) to fetch the existing configuration. Simply write the connector name in the filter bar, click on `SECRET_SOURCE-<name>__CREDS` and under `Actions`, select `View secret value` for the enabled version.
+
+### Importing in Connector Builder
+
+#### Connectors Without Manifest File
+
+This can be either a Python connector or a completely new one. Before implementing it in the Connector Builder, you should confirm if the API is suitable for the connector builder as per the [compatibility guide](./connector-builder-compatibility.md). Once this is confirmed, you are good to go! If not, please reach out to Airbyte to mention the incompatibility reason. This will help us prioritize changes and new features.
+
+#### Connectors With Manifest File
+Given a connector with a manifest.yaml file, you can attempt to import it. Go in the Connector Builder, click on "New custom connector" and select "Import a YAML". Given that the import is successful, you should be able to switch to the "UI" tab. Once this is done, there are a couple of manual validations to be performed:
+* Were the schemas properly imported? You can validate this by checking the `Declared schema` tab for a specific stream. If not, copy/paste the corresponding schema in the "Declared schema" tab. If the schemas have a "shared" folder, you will need to copy/paste the appropriate file whenever there is a `$ref` key.
+* Was the spec properly imported? If the connector has a spec and the Connector Builder does not have any testing value, the spec wasn't properly imported. 
+
+There are a couple of known cases where the import might fail:
+* The connector has custom components
+    * How to identify: Search for `class_name` in the manifest. 
+    * How to fix: Validate if the custom component can be replaced by one from the CDK (see [compatibility guide](./connector-builder-compatibility.md)). If so, replace the custom components by a dummy non-custom component in the manifest.yaml or manually remove the streams that rely on custom components and implement those as if they were new streams
+* Nested configurations 
+    * How to identify: Search for patterns like `config.<a config key>.<a nested config key>` or `config[<a config key>][<a nested config key>]`
+    * How to fix: Flatten the configuration and update the spec. An easy way to do this is to merge both keys like `config.<a config key>_<a nested config key>`
+
+If the import failed for something else, please reach out to the extensibility team with the error so that we can prioritize fixes.
+
+### Implementing the Change
+Once the current version of the connector is working in the Connector Builder, you can now implement the change! 
+
+## Merge Into Code
+
+Goal: Make change available in GitHub
+
+Expected output: The updated codebase
+
+Once you can see the behavior change through the Connector Builder, it's time to show this change to everyone. In order to do so, we need to make the updated manifest available to our airbytehq/airbyte repository.
+
+### New Source Setup
+If you are working on a new source, we need to initialize the source in our GitHub repository which requires a bit more work than to just update a source. Doing this will require you to use the terminal and Git CLI. To do this:
+* If you haven't already done it, pull the git repository using `git clone git@github.com:airbytehq/airbyte.git`. This will create a new folder where the code resides. If you already had the repository on your computer, make sure it's up-to-date by running `git checkout master && git pull`. This command might fail if you have in-flight changes. We recommend committing those changes to a specific branch or stashing them before continuing.
+* Go in the root of the repository  
+* Run `cd airbyte-integrations/connector-templates/generator && ./generate.sh` and generate a `Configuration Based Source`
+* Create a document under `docs/integrations/sources/<name>.md` that describes your source
+* Update `airbyte-integrations/connectors/source-<name>/metadata.yml`
+    * Set `registries` to reflect where this connector should be available
+    * Set `releaseDate`
+    * Set `allowedHosts.hosts`. See [Airbyte Connector Acceptance Test suite](../../connector-development/testing-connectors/connector-acceptance-tests-reference.md) for more details
+* [Optional] Add an icon at the path matching the metadata.yaml `icon` field
+
+If you intend to continue the process through the GitHub UI, you will have to make those changes available to GitHub by running: `git checkout -b source-<name>-creation && git add . && git commit -m "Creating the new source" && git push --set-upstream origin source-<name>-creation`
+
+### Make the Change Available
+
+#### Through Git CLI
+In your local repository:
+* Export the manifest from the Connector Builder, remove the `metadata` field and paste everything in `airbyte-integrations/connectors/source-<name>/source_<name>/manifest.yaml`
+* If exists, delete the `airbyte-integrations/connectors/source-<name>/source_<name>/schemas` folder
+* If exists, delete the `airbyte-integrations/connectors/source-<name>/source_<name>/spec.json` or `airbyte-integrations/connectors/source-<name>/source_<name>/spec.yaml`
+* If the connector is low-code with custom components or Python, delete every .py files under `source_<name>` except from `source.py` which should look like
+```
+#
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+#
+
+from airbyte_cdk.sources.declarative.yaml_declarative_source import YamlDeclarativeSource
+
+"""
+This file provides the necessary constructs to interpret a provided declarative YAML configuration file into
+source connector.
+
+WARNING: Do not modify this file.
+"""
+
+
+# Declarative Source
+class Source<name>(YamlDeclarativeSource):
+    def __init__(self):
+        super().__init__(**{"path_to_yaml": "manifest.yaml"})
+
+```
+
+Once this is done, commit your changes and push them to GitHub in a branch with a descriptive name.
+
+
+#### Through GitHub UI
+* Go to https://github.com/airbytehq/airbyte/branches, click on "New branch" and enter a descriptive name
+* Export the manifest from the Connector Builder, remove the `metadata` field, go to https://github.com/airbytehq/airbyte/tree/<your branch name>/airbyte-integrations/connectors/source-<name>/source_<name>/manifest.yaml, click on the pencil to "Edit this file", paste the Connector Builder manifest and commit this file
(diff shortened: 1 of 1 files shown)


docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+
+To ensure we can validate the changes on a connector, CATs are executed. A Connector Acceptance Test is a piece of software that emulates a worker (see architecture [here](../../understanding-airbyte/jobs.md)) and checks the output of commands that the connector is expected to implement. In order to configure these tests, we need to:  
+* Create tests scenarios in a sandbox environment: There is no streamlined process to do this. Each product has their own business rules that can't be captured in a unifying document. Given a new connector, this process might require to have pay for a subscription for the product in order to get credentials/workspace/etc... In order to be able to repopulate an API very easily, we suggest the use of a Postman collection or a script.
+* Configure what we want to be validated: Using `acceptance-test-config.yml`, you can toggle some of the validations that are performed. The most common cases are:
+    * If `expected_records.jsonl` to validate the records that the source returns has been added, ensure `acceptance_tests.basic_read.tests.expect_records` is populated. This might require you to create a catalog for this test suites.
```
girarda, line 144: follow up: we should provide a command to generate the expected_records. I suspect it's something like `python main.py read --config secrets/config.json --catalog integration_tests/configured_catalog.json | grep RECORD | jq -c .record` (not tested)

docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Validation
+
+Goal: Ensuring the changes work as expected and provide a safety net for the next modifications
+
+Expected output: CATs are green
+
+To ensure we can validate the changes on a connector, CATs are executed. A Connector Acceptance Test is a piece of software that emulates a worker (see architecture [here](../../understanding-airbyte/jobs.md)) and checks the output of commands that the connector is expected to implement. In order to configure these tests, we need to:  
+* Create tests scenarios in a sandbox environment: There is no streamlined process to do this. Each product has their own business rules that can't be captured in a unifying document. Given a new connector, this process might require to have pay for a subscription for the product in order to get credentials/workspace/etc... In order to be able to repopulate an API very easily, we suggest the use of a Postman collection or a script.
+* Configure what we want to be validated: Using `acceptance-test-config.yml`, you can toggle some of the validations that are performed. The most common cases are:
```
girarda, line 143: follow up: we'll want to provide some guidance on how to decide which validations to perform

docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+    * If `expected_records.jsonl` to validate the records that the source returns has been added, ensure `acceptance_tests.basic_read.tests.expect_records` is populated. This might require you to create a catalog for this test suites.
+    * If a stream was made incremental, ensure `acceptance_tests.incremental.tests.expect_records` is populated. This might require you to create a catalog for this test suites.
+* Set up the configuration: In order to run properly, CATs require a valid config. Those configs often rely on secret. You can use [GSM](https://console.cloud.google.com/security/secret-manager?project=dataline-integration-testing) to modify the existing configuration. Simply write the connector name in the filter bar, click on `SECRET_SOURCE-<name>__CREDS` and `NEW VERSION` and upload the new config.
+
+To execute CATs, there are two options.
+
+### Locally
+We recommend using `cd airbyte-integrations/connections/source-<name> && ./acceptance-test-docker.sh` to run them using docker. You might have to change the permissions to run this file. Once it's done, the results will be printed to the terminal.
```
girarda, line 151: why `acceptance-test-docker.sh` instead of `airbyte-ci`? (eg `airbyte-ci connectors --name=source-woocommerce test`)

The benefit or airbyte-ci is it's using the same process that runs on CI

docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+
+Goal: Make the connector available to end-users
+
+Expected output: https://hub.docker.com/r/airbyte/source-<name>/tags shows a recently created tag
+
+The following steps assume that you are getting
+
+* Bump version (Docker file and metadata.yaml) and update changelogs under `docs/integrations`
+* Fill the Connector Checklist in the pull request
+* If there are breaking changes, follow the process defined by [Connector Breaking Change Release Playbook](https://docs.google.com/document/d/1VYQggHbL_PN0dDDu7rCyzBLGRtX-R3cpwXaY8QxEgzw/edit#heading=h.r2xzw0kclodc). A breaking change is a change that requires the user to take action. It can be triggered by a change in the schema, in the config or in the connector behavior.
```
girarda, line 166: I'd expect any connector updates to change the connector behavior. We probably only care about the 5 types of changes listed in the playbook

docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+Goal: Make the connector available to end-users
+
+Expected output: https://hub.docker.com/r/airbyte/source-<name>/tags shows a recently created tag
+
+The following steps assume that you are getting
+
+* Bump version (Docker file and metadata.yaml) and update changelogs under `docs/integrations`
+* Fill the Connector Checklist in the pull request
+* If there are breaking changes, follow the process defined by [Connector Breaking Change Release Playbook](https://docs.google.com/document/d/1VYQggHbL_PN0dDDu7rCyzBLGRtX-R3cpwXaY8QxEgzw/edit#heading=h.r2xzw0kclodc). A breaking change is a change that requires the user to take action. It can be triggered by a change in the schema, in the config or in the connector behavior. 
+* In some specific oauth cases, there are platform changes to be done. <expand on this eventually>
```
girarda, line 167: maybe the step here is just to make sure there's a ticket with the appropriate priority?

If it's a new oauth connector, the platform changes are required before publishing it to cloud. Otherwise they should only be needed if there's a change in the scope requirements

docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+
+Expected output: https://hub.docker.com/r/airbyte/source-<name>/tags shows a recently created tag
+
+The following steps assume that you are getting
+
+* Bump version (Docker file and metadata.yaml) and update changelogs under `docs/integrations`
+* Fill the Connector Checklist in the pull request
+* If there are breaking changes, follow the process defined by [Connector Breaking Change Release Playbook](https://docs.google.com/document/d/1VYQggHbL_PN0dDDu7rCyzBLGRtX-R3cpwXaY8QxEgzw/edit#heading=h.r2xzw0kclodc). A breaking change is a change that requires the user to take action. It can be triggered by a change in the schema, in the config or in the connector behavior. 
+* In some specific oauth cases, there are platform changes to be done. <expand on this eventually> 
+* Get an approval on the pull request
```
girarda, line 168: follow up: we'll want guidance on who to ask for a review

docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+Expected output: A manifest.yaml file that represents this new behavior
+
+All the steps mentioned below assume that you have the API available to perform requests. 
+
+### Validate Compatibility of Added Features
+Before implementing anything, we need to know if the features you want to add are supported by the Connector Builder. Please refer to the [compatibility guide](./connector-builder-compatibility.md) to confirm if you can build this change within the Connector Builder.
+
+### Config in Connector Builder
+
+Given that the connector already exists and that you want to use the account configured for CATs to iterate during the build phase, you can use [GSM](https://console.cloud.google.com/security/secret-manager?project=dataline-integration-testing) to fetch the existing configuration. Simply write the connector name in the filter bar, click on `SECRET_SOURCE-<name>__CREDS` and under `Actions`, select `View secret value` for the enabled version.
```
girarda, line 25: why not use the `ci_credentials` tool? it seems like annoying and error prone than playing with GSM directly

girarda (commented):
this is great! Main question is whether this should be public or internal doc because this is referencing internal tools.

docs/connector-development/connector-builder-ui/maintenance.md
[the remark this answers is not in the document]
maxi297: The documentation for ci_credentials is public. That being said, I don't ming this being move. Let me know what you prefer

docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Importing in Connector Builder
+
+#### Connectors Without Manifest File
+
+This can be either a Python connector or a completely new one. Before implementing it in the Connector Builder, you should confirm if the API is suitable for the connector builder as per the [compatibility guide](./connector-builder-compatibility.md). Once this is confirmed, you are good to go! If not, please reach out to Airbyte to mention the incompatibility reason. This will help us prioritize changes and new features.
+
+#### Connectors With Manifest File
+Given a connector with a manifest.yaml file, you can attempt to import it. Go in the Connector Builder, click on "New custom connector" and select "Import a YAML". Given that the import is successful, you should be able to switch to the "UI" tab. Once this is done, there are a couple of manual validations to be performed:
+* Were the schemas properly imported? You can validate this by checking the `Declared schema` tab for a specific stream. If not, copy/paste the corresponding schema in the "Declared schema" tab. If the schemas have a "shared" folder, you will need to copy/paste the appropriate file whenever there is a `$ref` key.
```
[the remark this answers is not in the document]
maxi297: Because there can be a diff and I feel this is less error prone for the user to voluntarily update the schema seeing a diff than to just do a test run and without considering the current schema

docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+* If you haven't already done it, pull the git repository using `git clone git@github.com:airbytehq/airbyte.git`. This will create a new folder where the code resides. If you already had the repository on your computer, make sure it's up-to-date by running `git checkout master && git pull`. This command might fail if you have in-flight changes. We recommend committing those changes to a specific branch or stashing them before continuing.
+* Go in the root of the repository  
+* Run `cd airbyte-integrations/connector-templates/generator && ./generate.sh` and generate a `Configuration Based Source`
+* Create a document under `docs/integrations/sources/<name>.md` that describes your source
+* Update `airbyte-integrations/connectors/source-<name>/metadata.yml`
+    * Set `registries` to reflect where this connector should be available
+    * Set `releaseDate`
+    * Set `allowedHosts.hosts`. See [Airbyte Connector Acceptance Test suite](../../connector-development/testing-connectors/connector-acceptance-tests-reference.md) for more details
+* [Optional] Add an icon at the path matching the metadata.yaml `icon` field
```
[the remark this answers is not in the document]
maxi297: I'm not so sure too but we have some that are not set. For example, source-flexport

docs/connector-development/connector-builder-ui/maintenance.md
maxi297: The downside if that you have to manage two python version while the acceptance-test-docker.sh manage dependencies inside the docker which is way more simple for someone not versed in Python

docs/connector-development/connector-builder-ui/maintenance.md
maxi297: Again, this requires a user to be well versed in Python which is not the target for this document

docs/connector-development/connector-builder-ui/maintenance.md
maxi297: Yes! I think we can document these as we go as I'm personally unaware of some of those cases. GL has mentioned (which I'll add to the PR):
* Ignore some fields that are changing often
* Empty streams: sometimes in the sandbox, we have no option to populate some stream (empty_streams: ["users"])
* timeout: increase the timeout for the whole test

docs/connector-development/connector-builder-ui/maintenance.md
maxi297: Yes! Reading this, I feel like I should just refer to the document which explains what changes are breaking so that if the list is updated, we don't need to update the `maintenance.md`

docs/connector-development/connector-builder-ui/maintenance.md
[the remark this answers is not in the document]
flash1293: +1 for putting it into the internal docs - it might be confusing to users that just want to run their connector on their instance.

docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+
+If you intend to continue the process through the GitHub UI, you will have to make those changes available to GitHub by running: `git checkout -b source-<name>-creation && git add . && git commit -m "Creating the new source" && git push --set-upstream origin source-<name>-creation`
+
+### Make the Change Available
+
+#### Through Git CLI
+In your local repository:
+* Export the manifest from the Connector Builder, remove the `metadata` field and paste everything in `airbyte-integrations/connectors/source-<name>/source_<name>/manifest.yaml`
```
flash1293, line 77: Why removing the metadata field? It might store helpful settings for the builder UI and doesn't hurt.

docs/connector-development/connector-builder-ui/maintenance.md
flash1293, line 143: What about linking to https://docs.airbyte.com/connector-development/testing-connectors/connector-acceptance-tests-reference/?

docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+* If exists, delete the `airbyte-integrations/connectors/source-<name>/source_<name>/schemas` folder
```
flash1293, line 78: These points seem very specific about converting a previously python or custom-component lowcode connector to connector-builder-compatible. I think it makes sense splitting that out because it's not relevant in most cases.

docs/connector-development/connector-builder-ui/maintenance.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Release
+
+Goal: Make the connector available to end-users
+
+Expected output: https://hub.docker.com/r/airbyte/source-<name>/tags shows a recently created tag
+
+The following steps assume that you are getting
+
+* Bump version (Docker file and metadata.yaml) and update changelogs under `docs/integrations`
```
flash1293, line 164: Isn't this and the breaking changes playbook part of the checklist? Do we need to mention it here?

ryjabe:
I think this needs an alignment update. I don't think there is a world where it makes sense for a SEs org to own connector maintenance.  A proper SE org functions generally to help grow revenue.  If the AI is outside of that, it should also be outside of scope.  They work in tandem with a support function, sometimes called an SA role, or just Technical Support, who owns client support and satisfaction.  If we are shifting Connector maintenance to a non eng role, it should be that team.

I would rephrase this goal as:

> The goal is to enable TS to start maintaining some connectors, and for SEs to contribute new Connectors or improve existing ones. The way to achieve this is by sharing the process with them so that they have some basic knowledge as to how the maintenance should be done. The information will be presented to them next Monday

PR #28544, opened 2023-07-20
Metadata: Ensure you cannot add extra metadata fields outside of definition

bnchrch:
## What
Adds `additionalProperties: false` to the data field to prevent accidentally adding fields that are not used

### DIFF
--- a/airbyte-ci/connectors/metadata_service/lib/metadata_service/models/generated/ConnectorMetadataDefinitionV0.py
+++ b/airbyte-ci/connectors/metadata_service/lib/metadata_service/models/generated/ConnectorMetadataDefinitionV0.py
@@ -160,6 +160,9 @@ class Registry(BaseModel):
 
 
 class Data(BaseModel):
+    class Config:
+        extra = Extra.forbid
+
     name: str
     icon: Optional[str] = None
     definitionId: UUID
--- a/airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
+++ b/airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
@@ -25,6 +25,7 @@ properties:
       - githubIssueLabel
       - connectorSubtype
       - releaseStage
+    additionalProperties: false
     properties:
       name:
         type: string
--- a/airbyte-ci/connectors/metadata_service/lib/tests/fixtures/metadata_validate/invalid/metadata_no_extra_data.yaml
+++ b/airbyte-ci/connectors/metadata_service/lib/tests/fixtures/metadata_validate/invalid/metadata_no_extra_data.yaml
@@ -1,0 +1,15 @@
+metadataSpecVersion: 1.0
+data:
+  name: AlloyDB for PostgreSQL
+  definitionId: 1fa90628-2b9e-11ed-a261-0242ac120002
+  connectorType: source
+  dockerRepository: airbyte/image-exists-1
+  githubIssueLabel: source-alloydb-strict-encrypt
+  dockerImageTag: 0.0.1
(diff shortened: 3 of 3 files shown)


airbyte-ci/connectors/metadata_service/lib/metadata_service/models/generated/ConnectorMetadataDefinitionV0.py
```diff
@@ -160,6 +160,9 @@ class Config:
 
 
 class Data(BaseModel):
+    class Config:
+        extra = Extra.forbid
```
evantahler, line 164: The lack of good "heaven forbid" gifs is upsettings

airbyte-ci/connectors/metadata_service/lib/tests/fixtures/metadata_validate/invalid/metadata_no_extra_data.yaml
pedroslopez, line 1: So I see this new fixture, but no test code changes. Is this automatically set up to look for anything in this invalid folder?

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
```diff
@@ -25,6 +25,7 @@ properties:
       - githubIssueLabel
       - connectorSubtype
       - releaseStage
+    additionalProperties: false
```
evantahler, line 28: Double checking that the workflow we want to add new metadata properties would be to first update this file, and then add the new properties.  There's not a world in which we would want to allow new-but-as-of-yet unused metadata properties in first?

evantahler (approved):

airbyte-ci/connectors/metadata_service/lib/tests/fixtures/metadata_validate/invalid/metadata_no_extra_data.yaml
bnchrch: Thats exactly it

https://github.com/airbytehq/airbyte/blob/bnchrch/metadata/no-extra-properties/airbyte-ci/connectors/metadata_service/lib/tests/test_commands.py#L21

Bit of a quick, dirty, but kind of elegant way to define this is valid metadata and this is not

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/generated/ConnectorMetadataDefinitionV0.py
bnchrch: Forbid in general is lacking

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
bnchrch: I think if you are adding a new but not yet used you can still add it to the metadata file.

This just prevents these cases
1. You specify `public: false` thinking it will do something
2. You think you set `connectorSubtype` but in reality you set `connectorSubtyp`

airbyte-ci/connectors/metadata_service/lib/tests/fixtures/metadata_validate/invalid/metadata_no_extra_data.yaml
pedroslopez: Very nice!

pedroslopez (approved):

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
pedroslopez: > You specify public: false thinking it will do something

image

Issue #23497, opened 2023-02-27
request parameters are duplicated when using a cursor paginator


girarda:
## Current Behavior
tl;dr: request parameters are duplicated when using a cursor paginator

Some APIs return the URL of the next page of records as part of their response.

Example with a paginator setting a `limit` request parameter:

```
      paginator:
        type: DefaultPaginator
        page_token_option:
          type: RequestPath
        page_size_option:
          inject_into: request_parameter
          type: RequestOption
          field_name: limit
        pagination_strategy:
          type: CursorPagination
          cursor_value: '{{ response[''metadata''][''next''] }}'
          page_size: 1
````

part of the response will include the URL:
```
"next": "https://api.tempo.io/4/accounts?offset=5&limit=1",
```

On the next request, the `limit` request param will be duplicated if the URL returned by the API contains the request params (as is the case in the example above)

example request:
```
"next": "https://api.tempo.io/4/accounts?offset=5&limit=1&limit=1",
```

[slack thread](https://airbytehq-team.slack.com/archives/C027KKE4BCZ/p1677434219383379)

## Expected Behavior
* The simple retriever should deduplicate the request params before creating the prepared request

## Implementation details
* This can be down in the _fetch_next_page method
* urllib.urlparse can be used to extract the query params from the `path`
* This should be done in the SimpleRetriever
* `HttpStream` has the same potential for duplicate, but we should leave it out of scope to limit the blast radius of the change. We can create a follow up issue if needed.

## Open question
* How should we handle duplicate keys with different values?

## Acceptance criteria
- [ ] Request parameters are not duplicated when using the cursor pagination
- [ ] The CDK logs a warning when there are duplicate request params

girarda:
grooming notes:
- on conflict: use the value that was declared in the manifest

haithem-souala:
Hey,
Any ETA or workaround? this is blocking us from developing a new connector.

girarda:
Noting this is also a [blocker for connectors on demand](https://airbytehq-team.slack.com/archives/C03TP6W8081/p1688998668728259).

PR #28550, opened 2023-07-21
[Issue #23497] Deduplicate query parameters for declarative connectors

girarda:
## What
* Query parameters can be duplicated if they are present both in the request path and as request params.
* This happens for instance when using a CursorPagination that returns a path with query parameters since the path will often include the page. 

## How
* Before creating the PreparedRequest, parse the URL and remove the parameters from the params dict if they are also present in the URL IF they also have the same value ([passing multiple values for the same param is valid](https://stackoverflow.com/a/24728298))
* This feature is gated behind a flag that is only enabled for declarative connectors to limit the blast radius. Enabling this will break source-linkedinads and potentially more connectors.
* Fix typing issues

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/streams/http/http.py`
2. `airbyte-cdk/python/unit_tests/sources/streams/http/test_http.py`
3. `airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py`
4. `airbyte-cdk/python/airbyte_cdk/sources/message/repository.py`
5. `airbyte-cdk/python/airbyte_cdk/sources/utils/types.py`
6. `airbyte-cdk/python/airbyte_cdk/sources/streams/core.py`
7. `airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py`
8. `airbyte-cdk/python/airbyte_cdk/sources/streams/http/rate_limiting.py`

## 🚨 User Impact 🚨
* Declarative connectors will deduplicate the query parameters before sending the request

## Testing
The following CATs are failing with query param deduplication turned on. Linkedin-ads is the only GA connector that would break if we turned it on by default, so I would also be ok turning this on by default and disabling it for Linkedin

- [source-amplitude](https://github.com/airbytehq/airbyte/actions/runs/5618437957) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)
- [source-bing-ads](https://github.com/airbytehq/airbyte/actions/runs/5618438184) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)
- [source-chargebee](https://github.com/airbytehq/airbyte/actions/runs/5618438969) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)
- [source-github](https://github.com/airbytehq/airbyte/actions/runs/5618527106) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719
- [source-gitlab](https://github.com/airbytehq/airbyte/actions/runs/5618628709) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)
- [source-google-analytics-v4](https://github.com/airbytehq/airbyte/actions/runs/5618629503) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)
- [source-google-sheets](https://github.com/airbytehq/airbyte/actions/runs/5618630776) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)
- [source-greenhouse](https://github.com/airbytehq/airbyte/actions/runs/5618631036) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)
- [source-intercom](https://github.com/airbytehq/airbyte/actions/runs/5618700533) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)
- [source-klaviyo](https://github.com/airbytehq/airbyte/actions/runs/5618717950) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)
- [source-linkedin-ads](https://github.com/airbytehq/airbyte/actions/runs/5618718149) -> Fails because of this change
- [source-snapchat-marketing](https://github.com/airbytehq/airbyte/actions/runs/5618721815) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)
- [source-surveymonkey](https://github.com/airbytehq/airbyte/actions/runs/5618722187) -> [also broken in daily build](https://connectors.airbyte.com/files/generated_reports/test_summary/source-surveymonkey/index.html)
- [source-typeform](https://github.com/airbytehq/airbyte/actions/runs/5618723298) -> [also broken in daily build](https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1689836436491719)
- [source-zendesk-chat](https://github.com/airbytehq/airbyte/actions/runs/5618727980)


girarda:
/legacy-test connector=connectors/source-amazon-ads local_cdk=1

girarda:
/legacy-test connector=connectors/source-amplitude local_cdk=1

girarda:
/legacy-test connector=connectors/source-bing-ads local_cdk=1

girarda:
/legacy-test connector=connectors/source-chargebee local_cdk=1

girarda:
/legacy-test connector=connectors/source-facebook-marketing local_cdk=1

girarda:
/legacy-test connector=connectors/source-freshdesk local_cdk=1

girarda:
/legacy-test connector=connectors/source-google-ads local_cdk=1

girarda:
/legacy-test connector=connectors/source-google-analytics-data-api local_cdk=1

girarda:
/legacy-test connector=connectors/source-google-analytics-v4 local_cdk=1

girarda:
/legacy-test connector=connectors/source-google-search-console local_cdk=1

girarda:
/legacy-test connector=connectors/source-google-sheets local_cdk=1

girarda:
/legacy-test connector=connectors/source-greenhouse local_cdk=1

girarda:
/legacy-test connector=connectors/source-amazon-ads local_cdk=1

girarda:
/legacy-test connector=connectors/source-amplitude local_cdk=1

girarda:
/legacy-test connector=connectors/source-bing-ads local_cdk=1

girarda:
/legacy-test connector=connectors/source-chargebee local_cdk=1

girarda:
/legacy-test connector=connectors/source-facebook-marketing local_cdk=1

girarda:
/legacy-test connector=connectors/source-freshdesk local_cdk=1

airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
```diff
@@ -36,8 +37,6 @@
 )
 from airbyte_protocol.models.airbyte_protocol import Type as MessageType
 
-JsonType = Optional[Union[str, int, float, bool, Dict[str, "JsonType"], List["JsonType"]]]
```
girarda, line 39: moved to [types.py](airbyte-cdk/python/airbyte_cdk/sources/utils/types.py) so it can be shared

airbyte-cdk/python/airbyte_cdk/sources/utils/types.py
```diff
@@ -0,0 +1,7 @@
+#
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+#
+
+from typing import Union
+
+JsonType = Union[dict[str, "JsonType"], list["JsonType"], str, int, float, bool, None]
```
girarda, line 7: copied from [message_grouper.py](https://github.com/airbytehq/airbyte/pull/28550/files#diff-da31b6f6b71b9c35e21980b3e26f863cf1be1b2d038ddd5b65445eea0fe0d08aL39)

airbyte-cdk/python/airbyte_cdk/sources/streams/core.py
```diff
@@ -32,8 +32,11 @@
 
 def package_name_from_class(cls: object) -> str:
     """Find the package name given a class name"""
-    module: Any = inspect.getmodule(cls)
```
girarda, line 35: not sure why this is typed as `Any`. This is actually `Optional[Module]`

girarda:
/legacy-test connector=connectors/source-google-ads local_cdk=1

girarda:
/legacy-test connector=connectors/source-google-analytics-data-api local_cdk=1

girarda:
/legacy-test connector=connectors/source-google-analytics-v4 local_cdk=1

girarda:
/legacy-test connector=connectors/source-google-search-console local_cdk=1

girarda:
/legacy-test connector=connectors/source-google-sheets local_cdk=1

girarda:
/legacy-test connector=connectors/source-greenhouse local_cdk=1

airbyte-cdk/python/airbyte_cdk/sources/streams/http/rate_limiting.py
```diff
@@ -23,21 +23,27 @@
...
 
-def default_backoff_handler(max_tries: Optional[int], factor: float, **kwargs):
-    def log_retry_attempt(details):
+def default_backoff_handler(
+    max_tries: Optional[int], factor: float, **kwargs: Any
+) -> Callable[[Callable[[PreparedRequest, Mapping[str, Any]], Response]], Callable[[PreparedRequest, Mapping[str, Any]], Response]]:
+    def log_retry_attempt(details: Mapping[str, Any]) -> None:
         _, exc, _ = sys.exc_info()
-        if exc.response:
+        if isinstance(exc, RequestException) and exc.response:
```
girarda, line 34: I guess this is always the case in practice otherwise this would fail...

girarda:
/legacy-test connector=connectors/source-instagram local_cdk=1

girarda:
/legacy-test connector=connectors/source-intercom local_cdk=1

girarda:
/legacy-test connector=connectors/source-iterable local_cdk=1

girarda:
/legacy-test connector=connectors/source-linkedin-ads local_cdk=1

girarda:
/legacy-test connector=connectors/source-mailchimp local_cdk=1

girarda:
/legacy-test connector=connectors/source-mixpanel local_cdk=1

girarda:
/legacy-test connector=connectors/source-paypal-transaction local_cdk=1

girarda:
/legacy-test connector=connectors/source-pinterest local_cdk=1

girarda:
/legacy-test connector=connectors/source-recharge local_cdk=1

girarda:
/legacy-test connector=connectors/source-salesforce local_cdk=1

girarda:
/legacy-test connector=connectors/source-sendgrid local_cdk=1

girarda:
/legacy-test connector=connectors/source-snapchat-marketing local_cdk=1

girarda:
/legacy-test connector=connectors/source-surveymonkey local_cdk=1

girarda:
/legacy-test connector=connectors/source-tiktok-marketing local_cdk=1

girarda:
/legacy-test connector=connectors/source-typeform local_cdk=1

girarda:
/legacy-test connector=connectors/source-zendesk-chat local_cdk=1

girarda:
/legacy-test connector=connectors/source-zendesk-support local_cdk=1

girarda:
/legacy-test connector=connectors/source-zendesk-talk local_cdk=1

airbyte-cdk/python/airbyte_cdk/sources/streams/http/rate_limiting.py
```diff
@@ -61,9 +71,12 @@ def sleep_on_ratelimit(details):
...
             time.sleep(retry_after + 1)  # extra second to cover any fractions of second
 
-    def log_give_up(details):
+    def log_give_up(details: Mapping[str, Any]) -> None:
         _, exc, _ = sys.exc_info()
-        logger.error(f"Max retry limit reached. Request: {exc.request}, Response: {exc.response}")
+        if isinstance(exc, RequestException):
+            logger.error(f"Max retry limit reached. Request: {exc.request}, Response: {exc.response}")
+        else:
+            logger.error(f"Max retry limit reached for unknown request and response")
```
girarda, line 79: Hitting this branch would be unexpected

airbyte-cdk/python/airbyte_cdk/sources/streams/http/http.py
```diff
@@ -374,11 +395,12 @@ def parse_response_error_message(cls, response: requests.Response) -> Optional[s
...
 
         # default logic to grab error from common fields
-        def _try_get_error(value):
+        def _try_get_error(value: Optional[JsonType]) -> Optional[str]:
             if isinstance(value, str):
                 return value
             elif isinstance(value, list):
-                return ", ".join(_try_get_error(v) for v in value)
+                errors_in_value = [_try_get_error(v) for v in value]
+                return ", ".join(v for v in errors_in_value if v is not None)
```
girarda, line 403: We probably never hit a None value here since it would've crashed, but the check is still warranted

airbyte-cdk/python/airbyte_cdk/sources/streams/http/rate_limiting.py
```diff
@@ -23,21 +23,31 @@
...
         _, exc, _ = sys.exc_info()
-        if exc.response:
+        if isinstance(exc, RequestException) and exc.response:
             logger.info(f"Status code: {exc.response.status_code}, Response Content: {exc.response.content}")
         logger.info(
             f"Caught retryable error '{str(exc)}' after {details['tries']} tries. Waiting {details['wait']} seconds then retrying..."
         )
 
-    def should_give_up(exc):
+    def should_give_up(exc: Exception) -> bool:
```
girarda, line 40: I think in practice this is always a RequestException because we only use the backoff to retry HttpRequests, but the [give_up param passed to the backoff library](https://github.com/airbytehq/airbyte/pull/28550/files#diff-21fbde7917b82829112c58ad78af6ca62c2d2ba7229b0318a930176d68be6081R57)  should be a `Callable[[Exception], bool]`

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -476,10 +479,13 @@ def parse_records(
         request: requests.PreparedRequest,
         response: requests.Response,
         stream_state: Mapping[str, Any],
-        stream_slice: Mapping[str, Any],
+        stream_slice: Optional[Mapping[str, Any]],
     ) -> Iterable[StreamData]:
         yield from self.parse_response(response, stream_slice=stream_slice, stream_state=stream_state)
 
+    def must_deduplicate_query_params(self) -> bool:
+        return True
```
girarda, line 487: we can expose this in the manifest schema if needed

airbyte-cdk/python/airbyte_cdk/sources/declarative/retrievers/simple_retriever.py
```diff
@@ -476,10 +479,13 @@ def parse_records(
         request: requests.PreparedRequest,
         response: requests.Response,
         stream_state: Mapping[str, Any],
-        stream_slice: Mapping[str, Any],
+        stream_slice: Optional[Mapping[str, Any]],
     ) -> Iterable[StreamData]:
         yield from self.parse_response(response, stream_slice=stream_slice, stream_state=stream_state)
 
+    def must_deduplicate_query_params(self) -> bool:
```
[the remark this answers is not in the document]
girarda: @flash1293 this is blocking connectors on demand https://airbytehq-team.slack.com/archives/C03TP6W8081/p1688998668728259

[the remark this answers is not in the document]
flash1293: Synced offline and decided to go forward with this fix to get it resolved quickly, then remove it once we decoupled http stream from the simple retriever. @girarda do you want to flip the flag to off by default for now to keep it as focused as possible?

airbyte-cdk/python/airbyte_cdk/sources/streams/http/http.py
```diff
@@ -258,15 +262,32 @@ def error_message(self, response: requests.Response) -> str:
...
+    def must_deduplicate_query_params(self) -> bool:
+        return False
+
+    def deduplicate_query_params(self, url: str, params: Optional[Mapping[str, Any]]) -> Mapping[str, Any]:
+        if params is None:
+            params = {}
+        query_string = urllib.parse.urlparse(url).query
+        query_dict = {k: v[0] for k, v in urllib.parse.parse_qs(query_string).items()}
+
+        duplicate_keys_with_same_value = {k for k in query_dict.keys() if params.get(k) == query_dict[k]}
```
flash1293, line 274: As `urllib.parse.parse_qs` doesn't differentiate between strings and numbers (because that's not a concept for query params), this check fails if the query param in question is a number (or anything besides a string)
Screenshot 2023-07-24 at 10 35 08

In this example, q is not de-duplicated

flash1293 (commented):
Found a problem with data types (see comment in code) and one general note: This logic is only-deduplicating if there's the same key and value in path and query params, but not if there's the same key and value in the params itself or directly in the path. I guess this makes sense but it's something I didn't expect at first and is an important nuance we should probably call out somewhere

airbyte-cdk/python/airbyte_cdk/sources/streams/http/http.py
```diff
@@ -258,15 +262,32 @@ def error_message(self, response: requests.Response) -> str:
...
+        return {k: v for k, v in params.items() if k not in duplicate_keys_with_same_value}
+
     def _create_prepared_request(
```
maxi297, line 277: This method seems to be [a duplicate of the newly created class](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/http_requester.py#L353-L373). Should we update this as well?

girarda: yes, good idea

airbyte-cdk/python/airbyte_cdk/sources/streams/http/http.py
girarda: good catch! Fixed by casting the values to strings

girarda (commented):
@flash1293 I considered deduping the params from the URL too, but it adds a decent amount of complexity because it'll require parsing and rebuilding the URL.

I instead opted for describing the behavior as "not duplicating query params that are already encoded in the path" in the `RequestOption` description. Let me know what you think

airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/http_requester.py
```diff
@@ -421,7 +443,7 @@ def _send_with_retry(self, request: requests.PreparedRequest) -> requests.Respon
         if max_tries is not None:
             max_tries = max(0, max_tries) + 1
 
-        user_backoff_handler = user_defined_backoff_handler(max_tries=max_tries)(self._send)
+        user_backoff_handler = user_defined_backoff_handler(max_tries=max_tries)(self._send)  # type: ignore # we don't pass in kwargs to the backoff handler
```
flash1293, line 446: why did this one get necessary?

flash1293 (approved):
I tested this and it works as expected, however I'm not sure whether this is really fixes the problem (I think we should re-visit this as some point) - it's still possible to run into issues with this if the path emitted from the paginator overwrites a regular query param that's set explicitly for the first request or something like that. The posthog connector is such a case (and they had to use a custom component due to this): https://github.com/airbytehq/airbyte/blob/96aa72d3b2dae8dabd01dc2f052adc7bc7cfda12/airbyte-integrations/connectors/source-posthog/source_posthog/components.py#L46

The incremental sync component sets `after` to its start value, then the api returns a URL including a different `after` query param for each new page - the per-value deduplication wouldn't work properly in this case.

Doesn't need to be solved on this PR, just highlighting that the "param-overwrite" issue has some additional nuance.

airbyte-cdk/python/airbyte_cdk/sources/declarative/requesters/http_requester.py
girarda: good question. The interface of the `HttpStream._send` method is slightly different from the requester's (it accepts an extra param.

I think there are a couple of ways we can get around this
- Add an extra unused param to the requester's send_method
- Use a different backoff handler with a different callable type
- Accept that the backoff handler isn't properly typed
- Ignore the error here

None of these solutions look great to be, but I'm open to going with another one

girarda:
> I tested this and it works as expected, however I'm not sure whether this is really fixes the problem (I think we should re-visit this as some point) - it's still possible to run into issues with this if the path emitted from the paginator overwrites a regular query param that's set explicitly for the first request or something like that. The posthog connector is such a case (and they had to use a custom component due to this):
> 
> https://github.com/airbytehq/airbyte/blob/96aa72d3b2dae8dabd01dc2f052adc7bc7cfda12/airbyte-integrations/connectors/source-posthog/source_posthog/components.py#L46
> 
> The incremental sync component sets `after` to its start value, then the api returns a URL including a different `after` query param for each new page - the per-value deduplication wouldn't work properly in this case.
> 
> Doesn't need to be solved on this PR, just highlighting that the "param-overwrite" issue has some additional nuance.

Good point @flash1293 . I documented [the issue](https://github.com/airbytehq/airbyte/issues/28682)

PR #28562, opened 2023-07-21
connectors-ci: fix error handling

alafanechere:
## What
I believe https://github.com/airbytehq/airbyte/pull/28450 introduce a slight regression in our error handling logic:
When we run `.stdout` or `.stderr` in a task group and one of this task is raising an `ExecError` `anyio` wraps it in an `ExceptionGroup`. Downstream code is not handling `ExceptionGroup` but `ExecError`, which makes unhandled `ExceptionGroup` errors to be raised.

## How
Handle `ExceptionGroup` in `get_container_output` and raise an `ExecError` if the first member of the `ExceptionGroup` is an `ExecError`

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
@@ -153,7 +153,7 @@ async def get_exec_result(container: Container) -> Tuple[int, str, str]:
     """
     try:
         return 0, *(await get_container_output(container))
-    except (ExecError) as e:
+    except ExecError as e:
         return e.exit_code, e.stdout, e.stderr
 
 
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
@@ -5,6 +5,7 @@
 """This module groups util function used in pipelines."""
 from __future__ import annotations
 
+import contextlib
 import datetime
 import json
 import os
@@ -115,6 +116,17 @@ async def get_file_contents(container: Container, path: str) -> Optional[str]:
     return None
 
 
+@contextlib.contextmanager
+def catch_exec_error_group():
+    try:
+        yield
+    except anyio.ExceptionGroup as eg:
+        for e in eg.exceptions:
+            if isinstance(e, ExecError):
+                raise e
+        raise
+
+
 async def get_container_output(container: Container) -> Tuple[str, str]:
     """Retrieve both stdout and stderr of a container, concurrently.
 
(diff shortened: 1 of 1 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/utils.py
```diff
@@ -127,9 +127,14 @@ async def get_container_output(container: Container) -> Tuple[str, str]:
...
-        soon_stdout = task_group.soonify(container.stdout)()
-        soon_stderr = task_group.soonify(container.stderr)()
+    try:
+        async with asyncer.create_task_group() as task_group:
+            soon_stdout = task_group.soonify(container.stdout)()
+            soon_stderr = task_group.soonify(container.stderr)()
+    except anyio.ExceptionGroup as e:
+        if isinstance(e.exceptions[0], ExecError):
+            raise e.exceptions[0]
+        raise
```
helderco, line 137: It's not enough to check the first exception because you could have `ExceptionGroup(..., [QueryError(), ExecError()])`.

With Python 3.11, where [`ExceptionGroup`](https://peps.python.org/pep-0654/#backwards-compatibility) was introduced, you'd use this syntax:
```python
try:
    ...
except* ExecError as eg:
    # now all exceptions in `eg` are of type `ExecError`
    raise next(eg.exceptions)

except* QueryError as eg:
    # no ExecError exceptions in the group, but if there's `QueryError`s use the first 
    raise next(eg.exceptions)
```

There's a backport in earlier versions with the [exceptiongroup](https://pypi.org/project/exceptiongroup/) library.

helderco: Backport example (not tested!):
```python
from exceptiongroup import ExceptionGroup, catch

def exec_error_handler(eg: ExceptionGroup):
    raise next(eg.exceptions)

async def get_container_output(container: Container) -> Tuple[str, str]:
    with catch({ExecError: exec_error_handler):
        async with asyncer.create_task_group() as task_group:
            soon_stdout = task_group.soonify(container.stdout)()
            soon_stderr = task_group.soonify(container.stderr)()
        return soon_stdout.value, soon_stderr.value
```

alafanechere: Thank you! Very elegant syntax in Python 3.11 ! 
Will try the `exceptiongroup` lib for Python 3.10, thanks!

helderco: I just checked and `anyio` only has support for `exceptiongroup` library in [master](https://github.com/agronholm/anyio/issues/17), which will be released in version 4.0. In case the current implementation isn't compatible, I suggest you just loop until you find the first one:
```python
    except anyio.ExceptionGroup as eg:
        for e in eg.exceptions:
            if isinstance(e, ExecError):
                raise e
        raise
```

helderco: You can also but that in a context manager to reuse in other places:
```python
@contextlib.contextmanager
def catch_exec_error_group():
    try:
        yield
    except anyio.ExceptionGroup as eg:
        for e in eg.exceptions:
            if isinstance(e, ExecError):
                raise e
        raise

async def get_container_output(container: Container) -> Tuple[str, str]:
    with catch_exec_error_group():
        async with asyncer.create_task_group() as task_group:
            soon_stdout = task_group.soonify(container.stdout)()
            soon_stderr = task_group.soonify(container.stderr)()
        return soon_stdout.value, soon_stderr.value
```

alafanechere: @helderco the backport does not work, but your context manager trick is great and works!

helderco (approved):
LGTM 👍

In this task group in particular, since it's `stdout` and `stderr`, they should both have the same error (same exec) so this is totally fine. I need to remember that exceptions in a task group come in an exception group and handle accordingly. Sorry for the regression. 🙂

octavia-approvington:

octavia-approvington (approved):

PR #28000, opened 2023-07-06
cat/connectors-ci: replace docker runner with dagger runner in CAT

alafanechere:
## What
Closes https://github.com/airbytehq/airbyte/issues/27858.
In CAT, we want to run connector under test container with Dagger instead of `docker run`.
Reasons are explained in #27858.
Dagger requires Python 3.10, CAT is running Python 3.9,: we have to upgrade CAT to Python 3.10, this explain the breadth of changes in this PR.

### 1. Upgrade CAT to Python 3.10
* Remove the connector dependency to CAT (by editing setup.py and requirements.txt), because connectors run on Python 3.9. I don't think this dependency is required anymore as CAT should be run from airbyte-ci, or with acceptance-test-docker.sh, not from connectors. 
* Remove CAT dependency to airbyte-cdk, which is running Python 3.9. Replacing this dependency to airbyte-protocol.models makes more sense: CAT is not only made to test python connectors and only uses the protocol models. 
* Upgrade CAT dependency whose previous version are not working with 3.10 (`hypothesis-json-schema`) 
* Manage CAT with poetry and update its Dockerfile accordingly

### 2. Refactor the `ConnectorRunner` to use dagger
https://github.com/airbytehq/airbyte/blob/0555ca42bec2fdeddb0ed8b003def257d1b0a92b/airbyte-integrations/bases/connector-acceptance-test/connector_acceptance_test/utils/connector_runner.py#L1

We can get rid of a lot of docker related code and use Dagger API to provision the connector under test container with the required resources. The container under test image is passed from `airbyte-ci` using a tarball.

### 3. Change `airbyte-ci` to run CAT with dagger-in-dagger instead of docker-in-docker 
We can get rid of the the docker-in-docker paradigm. Binding CAT to a docker daemon sidecar container is not longer required and will reduce the number of moving parts and, hopefully, transient failures.
As CAT is now built on the fly it's also easier to make CAT changes and test these change on the same PR.

### 4. Make sure that other way of running CAT (directly via python execution or with `acceptance-test-docker.sh` is still supported
* `acceptance-test-docker.sh` still works
* **Running CAT from connector python venv does not work anymore as CAT is not a connector dependency and is using python 3.10 vs 3.9 on connectors.**
### 5. Update CAT unit test to make them pass.
* TODO
### 6. Publish CAT 1.0.0
* TODO

PR #28566, opened 2023-07-21
connectors-ci: per connector dockerd

### LINKED ISSUE #27867: connectors-ci: abusive re-use of cache

alafanechere:
## What
Closes #27867 

https://github.com/airbytehq/airbyte/pull/28000 is our best bet at reducing reliance on dockerd and its potential over caching issues. But it's currently blocked because I'm looking for advices from the Dagger team for the dagger-in-dagger usecase.

This PR is an alternative approach to try to increase isolation in CAT run:
* Create a docker daemon as a side car container for each connector
We originally had one global docker daemon shared accross containers and we suspect it to be the root cause dubious connector command output in CAT causing transient CAT failures.

Having a per-connector dockerd was not the recommended approach by the Dagger team as it creates a lot of overhead and services to manage for the engine. But we're trying this approach again to see if the original issues we had persist / have been solved in the latest Dagger version.
 
# Tests
A "nightly build" on all GA connectors is running [here](https://github.com/airbytehq/airbyte/actions/runs/5623577981)

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
@@ -23,8 +23,8 @@ from ci_connector_ops.pipelines.consts import (
     LICENSE_SHORT_FILE_PATH,
     PYPROJECT_TOML_FILE_PATH,
 )
-from ci_connector_ops.pipelines.utils import get_file_contents
-from dagger import CacheVolume, Client, Container, DaggerError, Directory, File, Platform, Secret
+from ci_connector_ops.pipelines.utils import get_file_contents, slugify
+from dagger import CacheSharingMode, CacheVolume, Container, DaggerError, Directory, File, Platform, Secret
 from dagger.engine._version import CLI_VERSION as dagger_engine_version
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py
```diff
@@ -405,12 +406,16 @@ def with_bound_docker_host(
     Returns:
         Container: The container bound to the docker host.
     """
-    dockerd = context.dockerd_service
-    docker_hostname = "global-docker-host"
+    dockerd_service = with_dockerd_service(context)
+    docker_hostname = f"{slugify(context.connector.technical_name)}-docker-host"
```
bnchrch, line 410: 👍

bnchrch (approved):
LGTM

Dock related gif

alafanechere:
Closing as unexpected Dagger / Docker errors were observed on the nightly build and https://github.com/airbytehq/airbyte/pull/28000 is showing good progress.

PR #28575, opened 2023-07-21
handle case where migrationDocumentationUrl is None (by default)

### LINKED ISSUE #28549: Migration URLs are not being generated (or their generation is not being persisted)
We validate the autogenerated migration urls in the validate step (I know this because I had to add the files to get it to pass) but they don't actually get persisted and show up in the registry.

erohmensing:
Close #28549 

By default, the object created by our pydantic models (I think) sets `migrationDocumentationUrl` to `None`. This means that if you use `releases_or_breaking_change.get("migrationDocumentationUrl", default_migration_documentation_url)`, it returns `None`, since that is the value at that key. 

Instead, return the value of the key, (or default None), and then or it with the default in order to apply the default if the value is `None`.


octavia-approvington:
Check that box <br/> TODO complete

octavia-approvington (approved):

PR #28583, opened 2023-07-21
Metadata: Ensure we run upload validate on the overrode metadata file

bnchrch:
## What
Fixes an issue where if you havent published the overrode version of the connector and go to run a prerelease then the prelease will fail because it cant find the regular docker image on dockerhub

### DIFF
--- a/airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
+++ b/airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
@@ -91,25 +91,28 @@ def upload_file_if_changed(
 
 
 def _latest_upload(metadata: ConnectorMetadataDefinitionV0, bucket: storage.bucket.Bucket, metadata_file_path: Path) -> Tuple[bool, str]:
-    version_path = get_metadata_remote_file_path(metadata.data.dockerRepository, metadata.data.dockerImageTag)
     latest_path = get_metadata_remote_file_path(metadata.data.dockerRepository, "latest")
-    latest_icon_path = get_icon_remote_file_path(metadata.data.dockerRepository, "latest")
+    return upload_file_if_changed(metadata_file_path, bucket, latest_path, disable_cache=True)
 
-    (
-        version_uploaded,
-        version_blob_id,
-    ) = upload_file_if_changed(metadata_file_path, bucket, version_path, disable_cache=True)
-    latest_uploaded, _latest_blob_id = upload_file_if_changed(metadata_file_path, bucket, latest_path, disable_cache=True)
 
-    # Replace metadata file name with icon file name
+def _version_upload(metadata: ConnectorMetadataDefinitionV0, bucket: storage.bucket.Bucket, metadata_file_path: Path) -> Tuple[bool, str]:
+    version_path = get_metadata_remote_file_path(metadata.data.dockerRepository, metadata.data.dockerImageTag)
+    return upload_file_if_changed(metadata_file_path, bucket, version_path, disable_cache=True)
+
+
+def _icon_upload(metadata: ConnectorMetadataDefinitionV0, bucket: storage.bucket.Bucket, metadata_file_path: Path) -> Tuple[bool, str]:
     local_icon_path = metadata_file_path.parent / ICON_FILE_NAME
-    if local_icon_path.exists():
-        upload_file_if_changed(local_icon_path, bucket, latest_icon_path)
+    latest_icon_path = get_icon_remote_file_path(metadata.data.dockerRepository, "latest")
+    if not local_icon_path.exists():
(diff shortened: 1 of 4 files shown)


airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
```diff
@@ -143,6 +145,12 @@ def upload_metadata_to_gcs(bucket_name: str, metadata_file_path: Path, prereleas
     Returns:
         Tuple[bool, str]: Whether the metadata file was uploaded and its blob id.
     """
+    version_uploaded = False
+    latest_uploaded = False
+
+    if prerelease:
+        metadata_file_path = create_prerelease_metadata_file(metadata_file_path, prerelease)
```
erohmensing, line 152: nice, easier to understand now - no chance of using the existing metadata file

airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
```diff
@@ -91,25 +91,28 @@ def upload_file_if_changed(
...
-        upload_file_if_changed(local_icon_path, bucket, latest_icon_path)
+    latest_icon_path = get_icon_remote_file_path(metadata.data.dockerRepository, "latest")
+    if not local_icon_path.exists():
+        return False, f"No Icon found at {local_icon_path}"
+    return upload_file_if_changed(local_icon_path, bucket, latest_icon_path)
 
-    return version_uploaded or latest_uploaded, version_blob_id
 
+def create_prerelease_metadata_file(metadata_file_path: Path, prerelease_tag: str) -> Path:
+    metadata, error = validate_and_load(metadata_file_path, [])
```
erohmensing, line 112: 👍🏻 don't validate when we load the initial file before we've overwritten it

erohmensing: this will still run it through the object validator, but not perform any extra validation (like if the docker image exists). i think that's fine

airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
```diff
@@ -91,25 +91,28 @@ def upload_file_if_changed(
...
+    if metadata is None:
+        raise ValueError(f"Metadata file {metadata_file_path} is invalid for uploading: {error}")
 
-def _prerelease_upload(metadata: ConnectorMetadataDefinitionV0, bucket: storage.bucket.Bucket, prerelease_tag: str) -> Tuple[bool, str]:
     # replace any dockerImageTag references with the actual tag
     # this includes metadata.data.dockerImageTag, metadata.data.registries[].dockerImageTag
```
erohmensing, line 117: Not requied, just a suggestion: i wonder if we should have a constant list of these somewhere instead of encoding them here, in case we add something later that alwo has docker image tags? 

or at least maybe pull out a method `overwrite_image_tags`

airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
```diff
@@ -153,7 +161,10 @@ def upload_metadata_to_gcs(bucket_name: str, metadata_file_path: Path, prereleas
...
-        return _prerelease_upload(metadata, bucket, prerelease)
-    else:
-        return _latest_upload(metadata, bucket, metadata_file_path)
+    _icon_upload(metadata, bucket, metadata_file_path)
+
+    version_uploaded, version_blob_id = _version_upload(metadata, bucket, metadata_file_path)
+    if not prerelease:
+        latest_uploaded, _latest_blob_id = _latest_upload(metadata, bucket, metadata_file_path)
+
+    return version_uploaded or latest_uploaded, version_blob_id
```
erohmensing, line 170: Will we ever see latest_uploaded here? in the case that we upload it, we'll have already uploaded the version too

erohmensing: I.e. for a non prerelease, if version upload failed, we want to know about it even it latest_uploaded succeeded, and the other way around. I guess we need separate return statements for the 2 cases?

erohmensing (approved):
Good catch! I see how we got into this sitch 

Most comments are just for my review/posterity of what's going on. Only concern is the return statement of `upload_metadata_to_gcs`.

airbyte-ci/connectors/metadata_service/lib/metadata_service/gcs_upload.py
bnchrch: Good callout.

It was bugging me to.

Updated to have a more robust return value

PR #28589, opened 2023-07-24
✨Source Instagram: upgrade api version added media type product

ext-herbert-santos:
## What
*Upgrated api facebook to =~v.17 *
*On stream media added column media_type_product*

## How
*I added a new field in the media stream, and updated the facebook api version to v.17 as it was in the very old v.11 version.*

## Recommended reading order
1. `Added new column`
2. `upgraded version facebook api`

## 🚨 User Impact 🚨
*update of the api version, being compatible with faceboool support
added new media column, with the column media_type_product it is possible to identify if the media_type video for example is from the feed or reels.*


ext-herbert-santos:
I have to fix a bug....

Issue #28617, opened 2023-07-24
Flaky destinations tests imply that secrets are sometimes not present 


evantahler:
The [nightly build history](https://connectors.airbyte.com/files/generated_reports/test_summary/destination-snowflake/index.html) for Destination Snowflake is flaky.  

Looking at the failures:

```
SnowflakeDestinationIntegrationTest > testCheckFailsWithInvalidPermissions() FAILED
    java.lang.NullPointerException: Cannot invoke "com.fasterxml.jackson.databind.JsonNode.asText()" because the return value of "com.fasterxml.jackson.databind.JsonNode.get(String)" is null
        at io.airbyte.integrations.destination.snowflake.SnowflakeDatabase.createDataSource(SnowflakeDatabase.java:63)
        at io.airbyte.integrations.destination.snowflake.SnowflakeInternalStagingDestination.getDataSource(SnowflakeInternalStagingDestination.java:94)
        at io.airbyte.integrations.destination.snowflake.SnowflakeInternalStagingDestination.check(SnowflakeInternalStagingDestination.java:50)
        at io.airbyte.integrations.destination.jdbc.copy.SwitchingDestination.check(SwitchingDestination.java:56)
        at io.airbyte.integrations.destination.snowflake.SnowflakeDestinationIntegrationTest.testCheckFailsWithInvalidPermissions(SnowflakeDestinationIntegrationTest.java:36)
```

This is the short test: https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/destination-snowflake/src/test-integration/java/io/airbyte/integrations/destination/snowflake/SnowflakeDestinationIntegrationTest.java#L29-L38

This error implies that reading the config's JSON file is failing... sometimes.  If the secret itself isn't changing, then perhaps downloading the config into the test environment is having problems.

alafanechere:
The nightly build passed today.
Will monitor for two consecutive successes before closing.

alafanechere:
[Four consecutive successes on destination-snowflake](https://github.com/airbytehq/airbyte/issues/28419), closing.
![Screen Shot 2023-07-28 at 12.14.09.png](https://images.zenhubusercontent.com/62a72f6a30e85f5981338b7c/bbf37e07-d995-4915-a6cc-539c258ebc15)

PR #28656, opened 2023-07-25
connectors-ci: unique secret name at session level

alafanechere:
## What
Relates to #27867 https://github.com/airbytehq/airbyte/issues/28617
According to the nightly tests results it appears that connectors spec validation tests often fails transiently, on different connectors.
My hypothesis is that the wrong config/secret is mounted to the CAT container [here](https://github.com/airbytehq/airbyte/blob/a29ca9b54fd3878eb79b18e7231a403007ad6d8f/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/environments.py#L468)
After investigation it's probably related to a namespace issue [when we instantiate `Secret` object after GSM download:](https://github.com/airbytehq/airbyte/blob/master/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/secrets.py#L59):
Dagger derives secret ID from their name and not their content, so far we've named secrets according to the config file name, which is not unique and often `config.json`. 
It means that when we call `context.dagger_client.set_secret(secret_filename, secret_plaintext)` in a concurrent context we can overwrite a connector secret with a wrong value as `dagger_client` is a pipeline wide object, not a context specific instance.

## How
Make secret name unique by concatenating connector name and file name:
```python
unique_secret_name = f"{context.connector.technical_name}_{secret_file}"
connector_secrets[secret_file] = context.dagger_client.set_secret(unique_secret_name, secret_plaintext)
```

## Tests
[Running a nightly build from this branch](https://github.com/airbytehq/airbyte/actions/runs/5654096657)

### DIFF
--- a/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/secrets.py
+++ b/tools/ci_connector_ops/ci_connector_ops/pipelines/actions/secrets.py
@@ -58,7 +58,9 @@ async def download(context: ConnectorContext, gcp_gsm_env_variable_name: str = "
     connector_secrets = {}
     for secret_file in await with_downloaded_secrets.directory(secrets_path).entries():
         secret_plaintext = await with_downloaded_secrets.directory(secrets_path).file(secret_file).contents()
-        connector_secrets[secret_file] = context.dagger_client.set_secret(secret_file, secret_plaintext)
+        # We have to namespace secrets as Dagger derives session wide secret ID from their name
+        unique_secret_name = f"{context.connector.technical_name}_{secret_file}"
+        connector_secrets[secret_file] = context.dagger_client.set_secret(unique_secret_name, secret_plaintext)
 
     return connector_secrets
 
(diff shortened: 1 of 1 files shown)


tools/ci_connector_ops/ci_connector_ops/pipelines/actions/secrets.py
```diff
@@ -82,8 +84,9 @@ async def upload(context: ConnectorContext, gcp_gsm_env_variable_name: str = "GC
     ci_credentials = await environments.with_ci_credentials(context, gsm_secret)
 
     return await (
-        ci_credentials.with_directory(secrets_path, context.updated_secrets_dir)
-        .with_exec(["ci_credentials", context.connector.technical_name, "update-secrets"])
+        ci_credentials.with_directory(secrets_path, context.updated_secrets_dir).with_exec(
+            ["ci_credentials", context.connector.technical_name, "update-secrets"]
+        )
```
alafanechere, line 89: formatting change

erohmensing:
Is there still the chance that 2 concurrent runs for the same connector (one which modified the spec) could have this issue? 

E.g. the new 1s1t big query branch and the regular bigquery branch?

alafanechere:
> Is there still the chance that 2 concurrent runs for the same connector (one which modified the spec) could have this issue?

@erohmensing By concurrent I mean concurrent within the same test pipeline execution. In our current paradigm a connector can't be tested twice is the same pipeline so I think we're safe.

erohmensing:
Gotcha, so this is only a problem for changes with dependencies which kick off multiple test runs in the same session. Thanks for the clarification!

alafanechere:
> Gotcha, so this is only a problem for changes with dependencies which kick off multiple test runs in the same session. Thanks for the clarification!

Or specifically nightly builds where we trigger tons of concurrent test execution in a single pipeline.

erohmensing (approved):
so long as the `connector_secrets` dict is per connector, looks good! otherwise i'd use the namespaced name as the key instead of just `secret_file`

alafanechere:
Merging as the nightly build workflow, even if not finished, look healthy so far, and I'd like to see the result on the real tomorrow's nightly build.

PR #28669, opened 2023-07-25
connectors-ci: extend nightly build timeout

alafanechere:
## What
Nightly build are running longer now that we have potentially fixed transient failure in https://github.com/airbytehq/airbyte/pull/28656 , I want to increase the time out value.


octavia-approvington:

octavia-approvington (approved):

PR #28685, opened 2023-07-25
Issue 28332/migrate spec in manifest

maxi297:
## What
Addresses https://github.com/airbytehq/airbyte-internal-issues/issues/1961

## How
Creating a script that identifies no code connectors, takes the spec.json or spec.yaml, add it to the manifest and save it

## Output
See https://github.com/airbytehq/airbyte/pull/28684

## Manually tested (imported in Connector Builder)
* source-google-pagespeed-insights
* source-ip2whois
* source-mailjet-sms
* source-metabase
* source-polygon-stock-api
* source-postmarkapp
* source-secoda
* source-sendgrid
* source-sentry
* source-smartengage
* source-tempo
* source-vantage
* source-whisky-hunter
* source-zoom

## ~~Notes on the YAML dump~~ (outdated)
Visual check done to confirm compatibility with Connector Builder generated YAML files:

Generic things that were validated equivalent in the UI:
* Breaklines will be removed
* Arrays will all be migrated to a multiline with indentation
* Comments will be removed 🚨🚨🚨

### string formatting
By default, string don't have quotes

Given that a string has a single quote, the string will be surrounded by a single quote and single quotes will be duplicated. So `{{ config['key'] }}` will become `'{{ config[''key''] }}'`

Given that a string has a double quote, no change will be done. So `{{ config["key"] }}` will stay the same

## Notes on the YAML dump
Using ruamel allows us to keep most of the formatting intact. This means:
* Breaklines will **NOT** be removed
* Arrays will **NOT** be migrated to a multiline with indentation
* Comments will **NOT** be removed

All this is fine because it keeps the changes to a minimum. Adding to that, we preserve how strings a quoted. The only change I haven't been able to prevent is some changes on the line length

This keep the change to a 

## 🚨 User Impact 🚨
By migrating the spec in the manifest, the user will now have the testing values properly created. The behavior of the connectors should not change


airbyte-cdk/python/airbyte_cdk/connector_builder/migrator/source.py
```diff
@@ -0,0 +1,166 @@
...
+        return os.path.join(self._python_package_path(source_name), "manifest.yaml")
+
+    def _python_package_path(self, source_name: str) -> str:
+        return os.path.join(self._path_from_name(source_name), source_name.replace("-", "_"))
+
+    def _path_from_name(self, source_name: str) -> str:
+        return os.path.join(self._base_path, "airbyte-integrations", "connectors", source_name)
+
+
+class _IndentingEmitter(yaml.emitter.Emitter):
```
maxi297, line 109: Using this indentation as it is what the UI uses and will therefore reduce the variability for users

airbyte-cdk/python/airbyte_cdk/connector_builder/migrator/source.py
```diff
@@ -0,0 +1,166 @@
...
+            if self._is_no_code_connector(manifest_file_path):
+                no_code_sources.append(directory)
+        return no_code_sources
+
+    @staticmethod
+    def _is_no_code_connector(manifest_file_path: str) -> bool:
+        if not os.path.exists(manifest_file_path):
+            return False
+
+        with open(manifest_file_path) as manifest_file:
```
maxi297, line 93: The code needs to be improved as some sources have the frame of no_code but do not implement `YamlDeclarativeSource` (example: sendgrid)

airbyte-cdk/python/airbyte_cdk/connector_builder/migrator/source.py
```diff
@@ -0,0 +1,166 @@
...
+        return Spec(
+            type="Spec",
+            connection_specification=spec["connectionSpecification"],
+            documentation_url=spec.get("documentationUrl", None),
+            advanced_auth=AuthFlow(**spec["advanced_auth"]) if "advanced_auth" in spec else None
+        )
+
+    def _delete_spec_outside_manifest(self, source_name: str) -> None:
+        json_spec_path = os.path.join(self._python_package_path(source_name), "spec.json")
+        yaml_spec_path = os.path.join(self._python_package_path(source_name), "spec.yaml")
```
brianjlai, line 71: just a thought: We do have some code duplication where we're checking multiple times whether json_spec or the yaml_spec exist. What we could do to simplify this a bit is:
```
def _get_spec_path(self, source_name: str) -> (Optional[str], Optional[str]):
  json_spec_path = os.path.join(self._python_package_path(source_name), "spec.json")
  yaml_spec_path = os.path.join(self._python_package_path(source_name), "spec.yaml")

  // maybe also validate there are not both?

  return (
    json_spec_path if os.path.exists(json_spec_path) else None,
    yaml_spec_path if os.path.exists(yaml_spec_path) else None
  )
```

We would do this first and pass those values into flows like the remove _fetch_spec_outside_manifest, we know that the presence of the path set implies it exists instead of reading from the OS again. Not a big deal or blocking, but that came to mind that we could figure out which spec we're dealing earlier

airbyte-cdk/python/airbyte_cdk/connector_builder/migrator/source.py
brianjlai: do you know of others where that may be the case? For context, Sendgrid is a particularly special case in that we had migrated it as low-code but we reverted it back because we weren't super confident that all the edge cases had been fixed. We left the code in as a reference point for future migrations. I'd consider this more of an in-development connector since we're not even sure it works anymore. And I don't think it is a big priority to solve for it if this is the only one

brianjlai (commented):
I had a few nits here and there, but the part that does concern me a bit is:
> Comments will be removes 🚨🚨🚨

Just a cursory glance, comments aren't in every connector, but it is used in places and there are the unfortunate part is for the times where it is used, sometimes it's really important info since the comment was necessary. Take `square` or `dremio` for example, those are really useful context.

I haven't dug into it that deeply, but have we considered other libraries like `ruamel.yaml` instead of the built in library which could better preserve some of our formatting?

https://stackoverflow.com/questions/20805418/pyyaml-dump-format

airbyte-cdk/python/airbyte_cdk/connector_builder/migrator/source.py
maxi297: I don't, no. This is the one that raised an issue on the migration. If for spec, it might be fine and we can put the spec in the manifest.yaml but this will not work for the schemas as we will delete the "schemas" folder which source-sendgrid relies on

airbyte-cdk/python/setup.py
```diff
@@ -50,6 +50,7 @@
     install_requires=[
         "airbyte-protocol-models==0.3.6",
         "backoff",
+        "click",
```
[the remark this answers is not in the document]
maxi297: No particular reason. I'll move it

airbyte-cdk/python/airbyte_cdk/connector_builder/migrator/source.py
maxi297: Good idea, will update soon!

airbyte-cdk/python/airbyte_cdk/connector_builder/migrator/source.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self, base_path: str) -> None:
+        self._base_path = base_path
+
+    def merge_spec_inside_manifest(self, source_name: str) -> None:
+        spec = self._fetch_spec_outside_manifest(source_name)
+        if spec:
+            manifest_path = self._manifest_path(source_name)
+            yaml = ruamel.yaml.YAML()
+            yaml.preserve_quotes = True
+            yaml.indent(mapping=2, sequence=4, offset=2)
```
brianjlai, line 29: are these properly indenting the resulting manifests? the one that stuck out was sequence, but going off some of our existing manifests, I would think the mapping=1, sequence=2, offset=2?

Quoted by chatgpt:
mapping: Specifies the number of spaces used for each level of indentation in mappings (dictionaries).
sequence: Specifies the number of spaces used for each level of indentation in sequences (lists).
offset: Specifies the number of spaces used for indentation when a sequence is a value in a mapping.

But also sort of trust but verify, if it indents properly for you then chatgpt might just be wrong

brianjlai (approved):
nothing blocking. nice work and thanks for moving to ruamel!

PR #28703, opened 2023-07-25
add a stub to test a java cdk workflow

c-p-b:
## What
Due to somewhat arbitrary github actions limitations, in order to iterate on a workflow in a pull request, it has to already exist as a barebones stub on the main branch, or else it won't show up in the UI and wont' be available to kick off with

## How
Add a stub

### DIFF
--- a/.github/workflows/java_cdk.yml
+++ b/.github/workflows/java_cdk.yml
@@ -1,0 +1,13 @@
+name: "Java CDK CI"
+
+on:
+  pull_request:
+    types:
+      - opened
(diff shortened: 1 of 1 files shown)


aaronsteers (approved):

aaronsteers:
I don't know if my approval alone is sufficient, but I've approved for my part all the same. 😄

cpdeethree:
It's just a stub, so assuming that we don't have weird unrelated failures I think your approval is just fine :D

PR #28717, opened 2023-07-25
Update all metadata sensor from 10 hours to 10 min

bnchrch:
## What
Updates our metadata sensor for all files to poll every 10 minutes


pedroslopez (approved):
Oh! 10 hours! That definitely explains it 😄 


Thanks!

PR #28725, opened 2023-07-26
Cdk publish flow

c-p-b:
## What
Build, test, publish pipelines for airbyte java cdk

## How
multistage it. This is probably overkill for just the cdk, but it demonstrates how it's "supposed" to work. Build all images, then copy the built jar files and test source into a new container, then test that. This of course makes the assumption that you don't need the build source to run tests. If you do, you would copy that into the container as well.

I also chose to demonstrate how to manage the CLI plugins. Here I opted to create a "cdk" plugin that holds "java" subcommands, it could theoretically also separately build the python cdk with the same plugin. It doesn't have to be this way, you could have a plugin that is just the Java CDK plugin, but I wanted to demonstrate the thought process around how plugins could/should work as an exercise


airbyte-ci/connectors/cdk/cdk/java_cdk/tasks.py
```diff
@@ -0,0 +1,104 @@
...
+    gradle_command = ["./gradlew", ":airbyte-cdk:java:airbyte-cdk:test", "--build-cache"]
+
+                
+    result = (
+                with_gradle(client, ctx, settings,bind_to_docker_host=False)
+                .with_directory("/airbyte", build_result.directory("/airbyte"), include=files_from_container) # test src and build files
+                .with_directory("/root/.m2/repository", build_result.directory("/root/.m2/repository")) # maven local jar files
+                .with_(load_settings(settings))
+                .with_exec(gradle_command + ["--scan"] if scan else gradle_command)
+                .with_exec(["rsync", "-az", "/root/.gradle/", "/root/gradle-cache"]) # sync gradle cache for future runs
```
cpdeethree, line 87: This is something I have a TODO item to make into a context manager so you don't have to care about having to remember to do this, but for now I'm putting it in here

airbyte-ci/connectors/cdk/cdk/java_cdk/ci.py
```diff
@@ -0,0 +1,81 @@
...
+from .settings import JavaCDKSettings
+from .tasks import build_java_cdk_task, test_java_cdk_task
+
+pass_pipeline_context = make_pass_decorator(PipelineContext, ensure=True)
+pass_global_settings = make_pass_decorator(JavaCDKSettings, ensure=True)
+
+java_group = ClickGroup(group_name="java", group_help="Commands for developing Airbyte Java CDK")
+
+
+class CICommand(ClickCommandMetadata):
```
cpdeethree, line 21: Curious what you think about having the classes like this. One nice thing you can do is add pydantic validators in here to validate arguments, like if you wanted to ensure that some string you passed in was valid. It does separate the command from the actual flow though, which may not be desirable.

airbyte-ci/connectors/cdk/cdk/java_cdk/ci.py
```diff
@@ -0,0 +1,81 @@
...
+from aircmd.models.utils import make_pass_decorator
+from dagger import Client, Container
+from prefect import flow
+from prefect.utilities.annotations import quote
+
+from .settings import JavaCDKSettings
+from .tasks import build_java_cdk_task, test_java_cdk_task
+
+pass_pipeline_context = make_pass_decorator(PipelineContext, ensure=True)
+pass_global_settings = make_pass_decorator(JavaCDKSettings, ensure=True)
```
cpdeethree, line 16: These are BaseSettings classes that derive information from the environment, with default values. They are singletons and intended to be immutable.

airbyte-ci/connectors/cdk/cdk/java_cdk/tasks.py
```diff
@@ -0,0 +1,104 @@
...
+                "./.env",
+                "./.env.dev",
+                ]
+    
+    repo = pygit2.Repository('.')
+    repo_root = os.path.dirname(os.path.dirname(repo.path))
+    print(repo_root)
+    mount_dir = client.host().directory(repo_root, include=files_from_host)
+    
+    gradle_command = ["./gradlew", ":airbyte-cdk:java:airbyte-cdk:assemble", "publishToMavenLocal", "--build-cache"]
```
cpdeethree, line 40: The general approach here is to only build the runtime code, then publish it to mavenlocal to make it easy to pass into new containers to do later stuff

airbyte-ci/connectors/cdk/cdk/java_cdk/tasks.py
```diff
@@ -0,0 +1,104 @@
...
+                with_gradle(client, ctx, settings,bind_to_docker_host=False)
+                .with_directory("/airbyte", build_result.directory("/airbyte"), include=files_from_container) # test src and build files
+                .with_directory("/root/.m2/repository", build_result.directory("/root/.m2/repository")) # maven local jar files
+                .with_(load_settings(settings))
+                .with_exec(gradle_command + ["--scan"] if scan else gradle_command)
+                .with_exec(["rsync", "-az", "/root/.gradle/", "/root/gradle-cache"]) # sync gradle cache for future runs
+        )
+
+    if scan:
+        await create_link_artifact(
```
cpdeethree, line 91: This is a Prefect concept, it makes the build scan show up in Prefect UI and also puts it in a nice location for us to retrieve it later to put it somewhere like a Github PR comment or Github Checks API or ship it off to metabase for querying later

airbyte-ci/connectors/cdk/cdk/java_cdk/tasks.py
```diff
@@ -0,0 +1,104 @@
...
+            link=result.file("/airbyte/scan-journal.log").contents().split(' - ')[2].strip(),
+            description="Gradle build scan",
+        )
+    
+    return result.sync()
+
+
+@task
+async def publish_java_cdk_task(build_result: Container, client: Client,settings: JavaCDKSettings, ctx: PipelineContext, scan: bool = False) -> Container:
+    # TODO: something here that will publish to maven repo from inside the build container, using the jar files
```
cpdeethree, line 102: presumably this will go in MyCloudRepo, but since I don't have the secrets part of this stuff wired in yet, I opted to leave this out for now and we can come back to it when ready

.github/workflows/java_cdk.yml
```diff
@@ -9,5 +9,41 @@ on:
...
+          DAGGER_TMP_BINDIR="/tmp/dagger_${DAGGER_CLI_COMMIT}"
+          export _EXPERIMENTAL_DAGGER_CLI_BIN="$DAGGER_TMP_BINDIR/dagger"
+          if [ ! -f  "$_EXPERIMENTAL_DAGGER_CLI_BIN" ]; then
+            mkdir -p "$DAGGER_TMP_BINDIR"
+            curl "https://dl.dagger.io/dagger/main/${DAGGER_CLI_COMMIT}/dagger_${DAGGER_CLI_COMMIT}_$(uname -s | tr A-Z a-z)_$(uname -m | sed s/x86_64/amd64/).tar.gz" | tar xvz -C "$DAGGER_TMP_BINDIR"
+          fi
+          cd airbyte-ci/connectors/cdk
+          poetry install
+          poetry run aircmd plugin install cdk --local .
+          poetry run aircmd cdk java ci --scan
```
aaronsteers, line 49: When I run `poetry run aircmd cdk java ci --scan` locally, I get this error:

```
dagger.exceptions.QueryError: lstat /tmp/buildkit-mount2440875622/scan-journal.log: no such file or directory
```

Is this expected? Do I need to run the dagger commands in order to work properly? Locally, the `cdk java ci` command works fine, just not with `--scan` suffixed.

aaronsteers: Looks like the same is triggering in CI: https://github.com/airbytehq/airbyte/actions/runs/5672355854/job/15371392305?pr=28725#step:6:963

cpdeethree: Ah, i only tested without scan. I think this is due to the location of where we are invoking it being different than the root of the project, which is where we write scan-journal as specified in settings.grádale. I’ll push a fix for this today and submit a or to your branch

PR #28746, opened 2023-07-26
Fix stream read given stream doesn't have any slice

maxi297:
## What
Resolving message grouper error where there is not request performed for the stream (outside of auxiliary requests)

## How
Remove the validation as it shows that the assumption that we would always have a request is false.

## 🚨 User Impact 🚨
Once merged in the Connector Builder project, it will stop failing if the substream partition router does not a return a slice

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
+++ b/airbyte-cdk/python/airbyte_cdk/connector_builder/message_grouper.py
@@ -182,8 +182,9 @@ class MessageGrouper:
             elif message.type == MessageType.CONTROL and message.control.type == OrchestratorType.CONNECTOR_CONFIG:
                 yield message.control
         else:
-            self._close_page(current_page_request, current_page_response, current_slice_pages, current_page_records)
-            yield StreamReadSlices(pages=current_slice_pages, slice_descriptor=current_slice_descriptor)
+            if current_page_request or current_page_response or current_page_records:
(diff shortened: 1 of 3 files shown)


girarda (approved):
:shipit: 

what's the error message we want to show here? something along the lines of "No requests were submitted. Check the configuration of the partition routers"?

maxi297:
I've asked the frontend team if they could show a warning when we have `slices == []`. If this is good, I'll merge with the fix where we return
```
{
    "type": "RECORD",
    "record": {
        "stream": "List Events",
        "data": {
            "logs": [],
            "slices": [],
            "test_read_limit_reached": false,
            "auxiliary_requests": [...],
            "inferred_schema": null,
            "inferred_datetime_formats": {},
            "latest_config_update": null
        },
        "emitted_at": 1690399116000
    }
}
```

instead of 

```
{
    "type": "RECORD",
    "record": {
        "stream": "List Events",
        "data": {
            "logs": [],
            "slices": [
                {
                    "pages": [
                        {
                            "records": [],
                            "request": null,
                            "response": null
                        }
                    ],
                    "slice_descriptor": null,
                    "state": null
                }
            ],
            "test_read_limit_reached": false,
            "auxiliary_requests": [...],
            "inferred_schema": null,
            "inferred_datetime_formats": {},
            "latest_config_update": null
        },
        "emitted_at": 1690399116000
    }
}
```

PR #28759, opened 2023-07-26
Add Slack Alert lifecycle to Dagster for Metadata publish

### LINKED ISSUE #28433: Add slack notification for new metadata file noticed by the pipeline
## Problem
Its not obvious if a connector has been added to the registry/registry entry

## Solution
Add slack notifications to the publish on merge channel for
1. Metadata being converted to registry entry
2. New registry entry published
2. New registry published

bnchrch:
# Overview
closes #28433 

1. Adds the `dagster-slack` resource
2. Adds a concept to the code base called "Logging Lifecycles"
3. Uses this new concept to log the metadata to registry life cycle across jobs and assets

## Notes for reviewer
Dagster `ops` are stupid and dont allow you to pass a basic argument to them 🤷 

## Before merge

- [x] 1. Ensure we can post to `#publish-on-merge-updates`
- [x] 2. Set `SLACK_TOKEN` in prod dagster to this bot https://api.slack.com/apps/A05K845HBE0
- [x] 3. Set `PUBLISH_UPDATE_CHANNEL` to `#publish-on-merge-updates`


erohmensing (commented):
I notice the only "FAILED" log we have is for something that might be expected anyway. Is there any extra handling we want to try to do for unexpected errors, or does our current slack reporting on dagster errors continue to cover that?

For example I wasn't sure if we'd want something [here](https://github.com/airbytehq/airbyte/pull/28759/files#diff-06628f2e19e6b503ee37e6574c4ca3ce03b15c4a082eacebfd723f130db0d5baR389)

bnchrch:
@erohmensing Thanks so much for the review.

On the note of
> I notice the only "FAILED" log we have is for something that might be expected anyway. Is there any extra handling we want to try to do for unexpected errors, or does our current slack reporting on dagster errors continue to cover that?

My current idea is that we depend on the new sentry logging in this PR for handling failures https://github.com/airbytehq/airbyte/pull/28822

In addition to dagster clouds slack errors on job failures.

One reason for this is laziness as why reinvent what these are doing.

The other is that youll notice often times to get the greater context around what an asset run means for our publish lifecycle we have to do some processing. Which means that greater context would be unavailable to any slack error try catch like mechanism that catches all failures

For example the registry_entry takes an etag (low infomation) and looks up the related metadata file (high information). And its this high information piece that will be unreachable during an out of bound error.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/ops/slack.py
```diff
@@ -18,4 +21,17 @@ def send_slack_webhook(webhook_url, report):
...
+    """
+    Send a slack message to the given channel.
+
+    Args:
+        context (OpExecutionContext): The execution context.
+        channel (str): The channel to send the message to.
+        message (str): The message to send.
+    """
+    if os.getenv("SLACK_TOKEN"):
+        context.resources.slack.get_client().chat_postMessage(channel=channel, text=message)
```
erohmensing, line 37: I used to work on a slack integration and the blast from the past that is `chat_postMessage` hits so hard

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/jobs/registry.py
```diff
@@ -19,29 +20,41 @@
...
-    context.instance.add_dynamic_partitions(partition_name, new_etags_found)
+    context.instance.add_dynamic_partitions(partition_name, etags_to_process)
+
+    # format new_files_found into a loggable string
+    new_metadata_log_string = "\n".join([f"{new_files_found[etag]} *{etag}* " for etag in etags_to_process])
+
+    PublishConnectorLifecycle.log(
+        context,
+        PublishConnectorLifecycleStage.METADATA_SENSOR,
+        StageStatus.SUCCESS,
```
erohmensing, line 55: I'd maybe make this an in progress? but not strongly opinionated on that one

erohmensing (approved):
LGTM! Let's try it out!

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/jobs/registry.py
bnchrch: At this point they are successfully queued! Next step is processing.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
```diff
@@ -139,13 +147,13 @@
         job=generate_oss_registry,
         resources_def=REGISTRY_ENTRY_RESOURCE_TREE,
         gcs_blobs_resource_key="latest_oss_registry_entries_file_blobs",
-        interval=30,
+        interval=60,
```
pedroslopez, line 150: Are we intentionally slowing things down here? What's the reasoning for doing so (just wondering)

pedroslopez (approved):
Nice - initial thought seeing this is that the channel might get quite noisy, and I wonder if it'll be obvious enough for people outside our team to know what's going on and know exactly when their connector is ready for use. 

I'm good with trying this out though and see what this looks like for real.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
bnchrch: Snuck it in to deal with a senor resource issue.

https://airbytehq-team.slack.com/archives/C056HGD1QSW/p1690931225565009

Very minor change, figured I would Save an approve-and-merge.

PR #28761, opened 2023-07-26
Downgrade orchestrator deploy back to python3.9

bnchrch:
## Overview
Im sad

I remember now why we had this set to 3.9 even though this is a new project.

essentially pendulum has prebuilt wheels for ubuntu up to python 3.9 but not 3.10 🤷 

In that case we have to build them from scratch

Which ubunutu is incapable of doing out of the box.

So here we are using python3.9 because I dont think the juice is worth the squeeze to figure out this wheel issue when eventually I assume it will be solved by pendulum/ubunutu

### Spec cache asset removed
The queueing and resourcing problem seemed to uncover a bug where the cached spec asset pickle would go missing intermittently under high load.

The underlying cause isnt well understood.

But moving it to a function call inside the registry_entry asset solved the issue.


airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/spec_cache.py
```diff
@@ -1,16 +0,0 @@
...
-GROUP_NAME = "spec_cache"
-
-
-@asset(
-    group_name=GROUP_NAME,
-    auto_materialize_policy=AutoMaterializePolicy.eager(max_materializations_per_minute=30),
-    freshness_policy=FreshnessPolicy(maximum_lag_minutes=1),
-)
-def cached_specs() -> OutputDataFrame:
-    return output_dataframe(pd.DataFrame(list_cached_specs()))
```
erohmensing, line 16: Is this change related to the "back to 3.9" change or a separate fix for specs?

airbyte-ci/connectors/pipelines/pyproject.toml
```diff
@@ -9,7 +9,7 @@ description = "Packaged maintained by the connector operations team to perform C
 authors = ["Airbyte <contact@airbyte.io>"]
 
 [tool.poetry.dependencies]
-python = "^3.10"
+python = "~3.9" # This is pinned to 3.10 as currently there is an issue when deploying via dagster-cloud where a dependency does not have a prebuild wheel file
```
erohmensing, line 12:
```suggestion
python = "~3.9" # This is pinned to 3.9 as currently there is an issue when deploying via dagster-cloud where a dependency does not have a prebuild wheel file
```

erohmensing (approved):
Seems fine, but not all changes look related to 3.9 - please flesh out the description :)

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/__init__.py
```diff
@@ -157,7 +156,7 @@
 ]
 
 SCHEDULES = [
-    ScheduleDefinition(job=add_new_metadata_partitions, cron_schedule="* * * * *"),
+    ScheduleDefinition(job=add_new_metadata_partitions, cron_schedule="*/5 * * * *", tags={"dagster/priority": HIGH_QUEUE_PRIORITY}),
```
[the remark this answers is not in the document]
bnchrch: Pretty much.

Because jobs were taking so long for spin up time during the registry entry back fill our queue became huge

Then it emptied slowly because these 1 minute sensors were pilling up to be ran.

This is a change that should reduce the number of sensors that can clog the queue.

Still not a fix to the resource problem though.

Dagster is recommending we move to hybrid to get more performance.

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/spec_cache.py
bnchrch: Updated the description to include this :)

airbyte-ci/connectors/pipelines/pyproject.toml
bnchrch: Realized I pinned the wrong pyproject here. Setting orchestrator/pyproject.toml python to ^3.9 should be sufficient to solve this

PR #28822, opened 2023-07-28
Dagster: Add sentry logging

### LINKED ISSUE #28673: Add instrumentation to Dagster

bnchrch:
# Overview
1. Adds sentry
2. Creates a sentry decorator to capture exceptions from ops and assets

closes #28673

# Notes for reviewers
1. Unfortuantely dagster eats any errors so we have to wrap the functions ourselves
2. Datadog is not an option until we move to an ec2 deployment

# Before merge
In prod dagster set
- [x] SENTRY_DSN="..."
- [x] SENTRY_ENVIRONMENT="production"
- [x] TRACES_SAMPLE_RATE=1.0


airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/logging/sentry.py
```diff
@@ -0,0 +1,107 @@
...
+    from sentry_sdk.integrations.dedupe import DedupeIntegration
+    from sentry_sdk.integrations.logging import LoggingIntegration, ignore_logger
+    from sentry_sdk.integrations.modules import ModulesIntegration
+    from sentry_sdk.integrations.stdlib import StdlibIntegration
+
+    ignore_logger("dagster")
+
+    SENTRY_DSN = os.environ.get("SENTRY_DSN")
+    SENTRY_ENVIRONMENT = os.environ.get("SENTRY_ENVIRONMENT")
+    TRACES_SAMPLE_RATE = float(os.environ.get("SENTRY_TRACES_SAMPLE_RATE", 0))
```
pedroslopez, line 28: i'm actually not sure if this is helpful for these - I have it disabled for now on `airbyte-ci` pipelines but 🤷🏼 

Did you see a use for it?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/logging/sentry.py
```diff
@@ -0,0 +1,107 @@
...
+        message=f"{context.job_name} - {context.op_def.name}",
+        level="info",
+        data={
+            "run_config": context.run_config,
+            "job_name": context.job_name,
+            "op_name": context.op_def.name,
+            "run_id": context.run_id,
+            "retry_number": context.retry_number,
+        },
+    )
```
pedroslopez, line 62: Is this information better served as a context? https://docs.sentry.io/platforms/python/enriching-events/context/

pedroslopez (commented):
Nice! Left some questions.

Also - we had said in the original issue that we wanted error rates, but I don't think this gets us that. Sentry does show some error rate stuff for transactions, but I'm not sure those will show up in the current setup. Adding sentry might still be useful though. Could you confirm what your thinking is around the error rates? Did you manage to see transactions in sentry?

pedroslopez:
Also! question: what does the release in sentry show up as when an error is caught?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/logging/sentry.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Manually sets up a bunch of the default integrations and disables logging of dagster
+    to quiet things down.
+    """
+    from sentry_sdk.integrations.argv import ArgvIntegration
+    from sentry_sdk.integrations.atexit import AtexitIntegration
+    from sentry_sdk.integrations.dedupe import DedupeIntegration
+    from sentry_sdk.integrations.logging import LoggingIntegration, ignore_logger
+    from sentry_sdk.integrations.modules import ModulesIntegration
+    from sentry_sdk.integrations.stdlib import StdlibIntegration
```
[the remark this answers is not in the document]
bnchrch: Good callout

So essentially here we had a trade off. 

If we use sentry out of the box there is one large issue
1. Dagster eats all the user code errors and doesnt let them make it to the sentry level hook

However if we work around this (like we did in this PR)

We end up in a situation where the dagster logger is reporting an exception for every asset in the graph that failed (very noisy)
Screenshot 2023-07-31 at 2 53 06 PM

Ontop of this the dagster logs dont have the sentry tags applied which makes grouping and fingerprinting a more difficult task

So this ignore is meant to reduce the noise, only log sentry errors weve caught and have applied tags to in one way or another.

The reason im comfortable enough with the leaky bucket to recommend it is we still have slack errors to fallback on

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/logging/sentry.py
```diff
@@ -0,0 +1,107 @@
...
+            "job_name": context.job_name,
+            "op_name": context.op_def.name,
+            "run_id": context.run_id,
+            "retry_number": context.retry_number,
+        },
+    )
+
+    sentry_sdk.set_tag("job_name", context.job_name)
+    sentry_sdk.set_tag("op_name", context.op_def.name)
+    sentry_sdk.set_tag("run_id", context.run_id)
```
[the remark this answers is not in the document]
bnchrch: Updated to use scopes

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/logging/sentry.py
bnchrch: The use generally is that Ill leave it set to 1.0 in production to get performance data and error rates.

We can change this to a bool, just figured Id leave it a more direct number

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/logging/sentry.py
bnchrch: I grabbed it from an example, I thought it was useful, however looking at the docs it seems like its more of a debug statement in case you forget to log context 🤔 

Will remove

bnchrch:
@pedroslopez just wanted to follow up on the error reporting comment

> Also - we had said in the original issue that we wanted error rates, but I don't think this gets us that. Sentry does show some error rate stuff for transactions, but I'm not sure those will show up in the current setup. Adding sentry might still be useful though. Could you confirm what your thinking is around the error rates? Did you manage to see transactions in sentry?

So the reason I went with Sentry here was because until we decide to move to a self hosted / hybrid dagster deployment we are unable to send logs directly to datadog (as we wont be able to spin up the datadog agent).

With this implementation though we should still get full application monitoring through sentry and get an error rate for each job

Heres one example: https://airbytehq.sentry.io/performance/summary/?project=4504087736090624&query=&referrer=performance-transaction-summary&statsPeriod=30d&transaction=generate_cloud_registry&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29

You can also get a rough waterfall view / flame graph by individual asset
https://airbytehq.sentry.io/performance/sentry-dev-test:54520cc347b5451780e9dd8a7c5f308d/?project=4504087736090624&query=&referrer=performance-transaction-summary&statsPeriod=30d&transaction=generate_cloud_registry&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29

However each asset in a job at this time still has a separate trace id which means we dont have any information directly related to average run time at the job level.

This is because sentry doesnt have an explicit/non hacky way of setting trace id at the transaction level manually yet.

> Also! question: what does the release in sentry show up as when an error is caught?
it shows as the git sha currently

https://airbytehq.sentry.io/issues/4356903801/?project=4504087736090624&query=&referrer=issue-stream&statsPeriod=14d&stream_index=10

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/logging/sentry.py
```diff
@@ -0,0 +1,226 @@
...
+def _with_sentry_op_asset_transaction(context: OpExecutionContext):
+    """
+    Start or continue a Sentry transaction for the Dagster Op/Asset
+    """
+    op_name = context.op_def.name
+    job_name = context.job_name
+
+    sentry_logger.info(f"Initializing Sentry Transaction for Dagster Op/Asset {job_name} - {op_name}")
+    transaction = sentry_sdk.Hub.current.scope.transaction
+    sentry_logger.info(f"Current Sentry Transaction: {transaction}")
```
pedroslopez, line 93: Is this not spammy? Could be a debug log level thing

pedroslopez (approved):
Thanks for explaining- Let's do it! :shipit:

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/logging/sentry.py
bnchrch: Good call!

PR #28833, opened 2023-07-28
Docs: Source Google Ads docs update

ChristoGrab:
## What
Updated the main and in-app docs for Google Ads.

### In-app file
- Added OSS prerequisites to inapp file
- Added link to main docs for OSS users to make sure they get their tokens and credentials
- Updated setup steps and added OSS-specific version

### Main file
- Slight restructure to developer token instructions
- Added step for obtaining OAuth credentials with link to Google's instructions
- Reworked/updated setup steps
- IMPORTANT!!: I updated the note about possible data duplication because of the conversion window. My understanding is that the intent was to warn users that the conversion window adds a larger range of data for incremental syncs to have to read (in case a conversion occurs within said window), which could result in data that has already been appended being duplicated.  I added a recommendation to consider using the Deduped History sync, as I believe this would mitigate the issue. That being said, I'm obviously working with a very limited understanding, so let me know if this section needs to be clarified or altered in any way!
- Removed some duplicate links
- Re-organized a couple out-of-sequence streams in the list


docs/integrations/sources/google-ads.md
```diff
@@ -82,10 +97,11 @@ The Google Ads source connector supports the following [sync modes](https://docs
 - [Incremental Sync - Append](https://docs.airbyte.com/understanding-airbyte/connections/incremental-append)
 - [Incremental Sync - Deduped History](https://docs.airbyte.com/understanding-airbyte/connections/incremental-deduped-history)
 
-**Important note**:
+:::caution
+When using the Conversion Window feature, please be aware that it may result in duplicates during Incremental Sync. Due to the possibility of conversions happening within the specified window, the connector will extend the range of the sync, which could potentially result in duplicated data.
```
sherifnada, line 101: This section is a bit vague to the user; "duplicates" isn't exactly right (and makes it sound like we're doing something wrong). This is a feature; it's more that we'd have updated  conversion data for the same campaigns/ad_groups/ad_group_ads. 

A conversion window is; 


> In digital advertising, a "conversion" generally refers to when a user who views or interacts with an ad goes on to perform a desired action. This could be anything from clicking through to the advertiser's website, signing up for a newsletter, making a purchase, and so on.
> Now, the "conversion window" is the period of time after a user sees or clicks on an ad during which their actions can still be credited to that ad.
> For example, imagine an online shoe store runs an ad and sets a conversion window of 30 days. If you click on that ad today, any purchases you make on the shoe store's site within the next 30 days will be considered conversions resulting from that ad.
> The length of the conversion window can vary depending on the goals of the advertiser and the nature of the product or service. Some businesses might set a shorter conversion window if they're promoting a limited-time offer, while others might set a longer window if they're advertising a product that consumers typically take a while to think about before buying.
> In essence, the conversion window is a tool for measuring the effectiveness of an advertising campaign. By tracking the actions users take after viewing or interacting with an ad, businesses can gain insight into how well their ads are working and adjust their strategies accordingly.


I think we can boil all this down to: every time the connector runs it will output conversion data for every campaign, ad_group, or ad_group_ad which was running less than N days ago where N is the configured conversion window. This will make it look like the same campaigns/ad_groups/ads have different conversion numbers when in reality each record tracks the number of conversions for that resource at the time of extracting the data from the Google Ads API. If you want to know the most up-to-date conversion numbers, use incremental-dedupe syncs. 

Maybe we should also have that blurb to explain conversion windows in the doc

docs/integrations/sources/google-ads.md
```diff
@@ -4,9 +4,11 @@ This page contains the setup guide and reference information for the Google Ads
 
 ## Prerequisites
 
-- A [Google Ads Account](https://support.google.com/google-ads/answer/6366720) [linked](https://support.google.com/google-ads/answer/7459601) to a [Google Ads Manager account](https://ads.google.com/home/tools/manager-accounts/)
+- A [Google Ads Account](https://support.google.com/google-ads/answer/6366720) [linked](https://support.google.com/google-ads/answer/7459601) to a Google Ads Manager account
```
sherifnada, line 7: i'm not sure if the manager account is required -- asking the team in slack

sherifnada (approved):

docs/integrations/sources/google-ads.md
ChristoGrab: Gotcha, thanks for the clarification! I added a dedicated section to this with the blurb and explanation, and removed the deduped recommendation

docs/integrations/sources/google-ads.md
sherifnada: update: it's required

PR #28845, opened 2023-07-30
Add start_date to all file-based configs

### LINKED ISSUE #28136: File CDK: add start_date config option
The new file-based CDK does not offer the `start_date` configuration option, but it seems particularly useful and will also give us backwards compatibility with the [existing S3 connector](https://github.com/airbytehq/airbyte/blob/f79aa72d64de412c2710ab806cd1148df7314133/airbyte-integrations/connectors/source-s3/source_s3/source.py#L51).

When `start_date` is set, we should only sync files whose `last_modified` is equal to or greater than that date.

brianjlai:
Closes #28136

## What

Adds support for the `start_date` config option to be included for all file-based sources. The S3 connector implement this directly, but given a start_date is a pretty common behavior that all sources should implement.

## How

Adds this to the Spec model so that it will be included in any generated spec and config validation for all sources. The `start_date` config is also used by the Cursor when we filter which files should be read or used for schema inference.

## Recommended reading order
1. `abstract_file_based_spec.py`
2. `default_file_based_cursor.py`
3. `default_file_based_stream.py`


airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -152,17 +152,18 @@ def _get_raw_json_schema(self) -> JsonSchema:
         elif self.config.schemaless:
             return schemaless_schema
         else:
-            files = self.list_files()
-            total_n_files = len(files)
+            all_files = self.list_files()
+            files_to_read = list(self._cursor.get_files_to_sync(all_files, self.logger))
```
brianjlai, line 156: This part is worth noting. The existing schema inference code never relied on the cursor (although it is defined by default). This modifies the flow a bit to also apply the cursor to filter out records for both `discover` and `read`.

A few alternative options I explored:
- Using `start_date` in `AbstractFileBasedStreamReader.filter_files_by_globs()`. This has the nice benefit of always being applied even if the source implements their own FileBasedCursor. But it has the drawback of inserting more modified comparisons across the codebase which I don't love
- We also could just add a standalone check here to filter out files here or in `self.list_files()`, but the datetime checks are meant to be done in the cursor.

Up for discussion, but I think incorporating it here feels sensible since we do plan on a stateful discover at some point which will need the cursor.

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
```diff
@@ -17,6 +17,15 @@ class AbstractFileBasedSpec(BaseModel):
     that are needed when users configure a file-based source.
     """
 
+    start_date: Optional[str] = Field(
+        title="Start Date",
+        description="UTC date and time in the format 2017-01-25T00:00:00Z. Any file modified before this date will not be replicated.",
+        examples=["2021-01-01T00:00:00Z"],
+        format="date-time",
+        pattern="^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$",
```
sherifnada, line 25: Could you use a pattern descriptor whenever you specify a pattern? https://docs.airbyte.com/connector-development/connector-specification-reference#pattern-descriptors

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -152,17 +152,18 @@ def _get_raw_json_schema(self) -> JsonSchema:
...
             if total_n_files > max_n_files_for_schema_inference:
                 # Use the most recent files for schema inference, so we pick up schema changes during discovery.
-                files = sorted(files, key=lambda x: x.last_modified, reverse=True)[:max_n_files_for_schema_inference]
+                files_to_read = sorted(files_to_read, key=lambda x: x.last_modified, reverse=True)[:max_n_files_for_schema_inference]
                 self.logger.warn(
                     msg=f"Refusing to infer schema for all {total_n_files} files; using {max_n_files_for_schema_inference} files."
                 )
 
-            inferred_schema = self.infer_schema(files)
+            inferred_schema = self.infer_schema(files_to_read)
```
clnoll, line 166: If `files_to_read` is empty, mind raising `EMPTY_STREAM` exception before we call `self.infer_schema`? It's a little more specific than the invalid schema exception.

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
clnoll: A couple of thoughts -

I think I'd prefer not to restrict schema inference to only files that are supposed to be synced, because it will potentially be a totally normal case for there to just be 0 or 1 files that need to be synced, in which case schema inference will throw an exception (for the 0 files case) or only infer from the one file, which is more error prone.

That said, it makes sense to me to not include files from before `start_date` in schema inference.

So it's feeling to me like `filter_files_by_globs` would be a good place for it (although if we do that it we'd want to rename it). Also worth noting, if no files are returned from `filter_files_by_globs`, we currently fail `check`, because the implication is that we can't do schema inference and that there's a possible misconfiguration. If we incorporate a `start_date` check in there and there are no files before `start_date`, the failure might be useful to the user (i.e. don't have to wait until a discover/sync to see that their stream is empty).

airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -1531,3 +1531,58 @@
...
+                    },
+                    "name": "stream1",
+                    "source_defined_cursor": True,
+                    "supported_sync_modes": ["full_refresh", "incremental"],
+                }
+            ]
+        }
+    )
+    .set_expected_records([])
+    .set_expected_discover_error(SchemaInferenceError, FileBasedSourceError.SCHEMA_INFERENCE_ERROR.value)
```
clnoll, line 1587: If we go the route I mentioned above where we update `filter_files_by_globs`, we'll want to add this to the `check_scenarios` and update it with the expected check failure info.

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/default_file_based_cursor.py
```diff
@@ -111,6 +114,8 @@ def _compute_earliest_file_in_history(self) -> Optional[RemoteFile]:
 
     def _compute_start_time(self) -> datetime:
         if not self._file_to_datetime_history:
+            # if self._config_start_date:
```
clnoll, line 117: Is this needed, or can it be deleted?

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
brianjlai: That's makes sense! That was the original approach that we had discussed, but this redesign is probably over-consolidating how we decide what files to pull from the file store. I will incorporate the feedback and revert back to old approach where we use `filter_files_by_globs_and_start_date()`

girarda: Moving to `filter_files_by_globs` makes sense to me. The function would effectively become a filter on the name and timestamp, which seems reasonable.

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/default_file_based_cursor.py
```diff
@@ -14,13 +14,15 @@
 class DefaultFileBasedCursor(FileBasedCursor):
     DEFAULT_DAYS_TO_SYNC_IF_HISTORY_IS_FULL = 3
     DATE_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
+    CONFIG_START_DATE_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
```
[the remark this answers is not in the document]
brianjlai: The S3 connector stored them with milli/micro, so this was to retain backwards compatibility. But actually I think we're better off keeping consistent time format and the S3 config transformer can try to support the added precision. will fix this

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/abstract_file_based_spec.py
```diff
@@ -17,6 +17,16 @@ class AbstractFileBasedSpec(BaseModel):
     that are needed when users configure a file-based source.
     """
 
+    start_date: Optional[str] = Field(
+        title="Start Date",
+        description="UTC date and time in the format 2017-01-25T00:00:00Z. Any file modified before this date will not be replicated.",
```
clnoll, line 22: This and the example should be `2017-01-25T00:00:00.000Z`, right?

clnoll (approved):
Looks great @brianjlai! Just one small comment.

girarda (approved):

PR #16172, opened 2022-08-31
SAT: Relax checking of `oneOf` common property and allow optional `default` keyword additional to `const` keyword

grubberr:
Signed-off-by: Sergey Chvalyuk <grubberr@gmail.com>

## What
We have requirement for all `oneOf` common properties to use only `const` keyword without `default` and `enum` keywords like this:

```json
"option_title": {
  "type": "string",
  "const": "OAuth Credentials"
}

"option_title": {
  "type": "string",
  "const": "PAT Credentials",
}
```

BUT new `pydentic==1.10.0` after this PR https://github.com/pydantic/pydantic/pull/4152
started to generate json-schemas with `default` keyword like this:

```json
"option_title": {
  "type": "string",
  "const": "OAuth Credentials",
  "default": "OAuth Credentials"
}

"option_title": {
  "type": "string",
  "const": "PAT Credentials",
  "default": "PAT Credentials",
}
```

## How
I propose to be compatible with `pydantic>=1.10.0` and not forbid `default` keyword in such common properties.

PR #28854, opened 2023-07-31
Python CDK: Relax pydantic version requirement

flash1293:
This PR relaxes the version requirement of pydantic for the CDK to make it easier to use libraries that don't work with 1.9.2 

It's an updated version of https://github.com/airbytehq/airbyte/pull/16172

As spec schemas are generated slightly differently, the CATs need to be adjusted. It's not allowed to specify `default` or `enum` on the common const property as long as they only repeat the `const` value. This works fine in the UI as it only checks the `const` property: https://github.com/airbytehq/airbyte-cloud/blob/18988f9e34cb245edcdf6c1f83d78e9b64b3d71d/oss/airbyte-webapp/src/core/form/schemaToFormBlock.ts#L75

and doesn't actually render any fields that have `const` set: https://github.com/airbytehq/airbyte-cloud/blob/bbb24e2f7d9525285e7a9ac73cf92b22e52ac480/oss/airbyte-webapp/src/views/Connector/ConnectorForm/components/Sections/FormSection.tsx#L33


sherifnada (approved):

flash1293:
/legacy-publish connector=bases/source-acceptance-test auto-bump-version=false

PR #28858, opened 2023-07-31
connectors-ci: increase connectors package install max duration

alafanechere:
During nightly build concurrent execution context some connectors take more than 10mn to complete their Connector Package Install step. Increasing the timeout to 20mn.
CF: https://airbytehq-team.slack.com/archives/C02TYE9QL9M/p1690797994052729?thread_ts=1690783339.381649&cid=C02TYE9QL9M

### DIFF
--- a/airbyte-ci/connectors/pipelines/README.md
+++ b/airbyte-ci/connectors/pipelines/README.md
@@ -367,9 +367,10 @@ This command runs tests for the metadata service orchestrator.
 `airbyte-ci metadata test orchestrator`
 
 ## Changelog
-| Version | PR  | Description                                                                                |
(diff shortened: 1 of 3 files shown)


pedroslopez (approved):
This change itself seems fine, but... should installs _really_ be taking 20 minutes?

Feels like we should investigate why these are taking so long.

alafanechere:
> This change itself seems fine, but... should installs really be taking 20 minutes?
Feels like we should investigate why these are taking so long.

@pedroslopez My current hypothesis, without investigation, is that the step could be long in high concurrency context were multiple steps/connectors pipelines are running in parallel and some task might be blocking an other. I increased the timeout value to check if this hypothesis is valid.

PR #28786, opened 2023-07-27
Source S3: basic structure using file-based CDK

clnoll:
## What
Adds the basic structure for using the file-based CDK for syncing data from S3:
- S3 implementation of the `AbstractFileBasedStreamReader`
- S3-specific config

This includes a few changes to the file-based CDK. 
- I've removed the `file_type` attribute on `RemoteFile` because we don't actually know the type of the `RemoteFile` when we create it. We could theoretically guess based on the stream config, but there's no need to.
- The `AbstractFileBasedStreamReader`'s `open_file` method now takes in a `mode` argument. This is passed in by the parser that's calling `open_file`, which knows its associated file read mode.

The S3 integration tests will be updated in a separate phase, which will verify that both the old and new config work.

## Recommended reading order
1. file-based CDK changes
2. S3-specific changes

PR #28862, opened 2023-07-31
File-based CDK: add read mode to stream reader interface & parsers

clnoll:
Separates out the file-based CDK portion of  https://github.com/airbytehq/airbyte/pull/28786.

These changes were made during the development of the source-S3 connector, which is the first real connector to use the file-based CDK and so identified a few changes that were needed to the interfaces.


airbyte-cdk/python/airbyte_cdk/entrypoint.py
```diff
@@ -156,57 +157,64 @@ def read(self, source_spec: ConnectorSpecification, config: TConfig, catalog: TC
...
         update_secrets(config_secrets)
 
     @staticmethod
-    def airbyte_message_to_string(airbyte_message: AirbyteMessage) -> str:
+    def airbyte_message_to_string(airbyte_message: AirbyteMessage) -> Any:
         return airbyte_message.json(exclude_unset=True)
 
-    def _emit_queued_messages(self, source) -> Iterable[AirbyteMessage]:
+    @classmethod
+    def extract_catalog(cls, args: List[str]) -> Optional[Any]:
```
girarda, line 178: Where will this method be called from? I'm not sure anything should depend on the entrypoint

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -90,6 +91,7 @@ def streams(self, config: Mapping[str, Any]) -> List[Stream]:
         """
         try:
             parsed_config = self.spec_class(**config)
+            self.stream_reader.config = parsed_config
```
girarda, line 94: why set the config as a stream_reader property instead of passing it to the methods that need it?

As it is, the stream_reader isn't fully initialized until we set its config, which is a non-obvious gotcha to keep in mind (similar to how we need to instantiate a stream's state separately from it's instantiation)

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py
```diff
@@ -46,26 +69,29 @@ def get_matching_files(
...
+        # (https://facelessuser.github.io/wcmatch/wcmatch/#globstar)
+        return any(globmatch(file.uri, g, flags=GLOBSTAR) for g in globs)
 
     @staticmethod
-    def get_prefixes_from_globs(globs: List[str]) -> List[str]:
+    def get_prefixes_from_globs(globs: List[str]) -> Set[str]:
         """
         Utility method for extracting prefixes from the globs.
         """
-        prefixes = {glob.split("*")[0].rstrip("/") for glob in globs}
```
girarda, line 70: was the rstrip just a bug? How do we know/verify this new behavior is the expected one?

airbyte-cdk/python/airbyte_cdk/sources/file_based/remote_file.py
```diff
@@ -3,24 +3,29 @@
 #
 
 from datetime import datetime
+from enum import Enum
 from typing import Optional
 
 from pydantic import BaseModel
 
 
+class FileReadMode(Enum):
```
girarda, line 12: should this enum be in `file_type_parser` since it's a property of the parser?

airbyte-cdk/python/unit_tests/sources/file_based/in_memory_files_source.py
```diff
@@ -42,48 +43,61 @@ def __init__(
         file_write_options: Mapping[str, Any],
         max_history_size: int,
     ):
+        # Attributes required for test purposes
+        self.files = files
+        self.file_type = file_type
```
girarda, line 48: will the S3 source have the same restriction of one `file_type` per source?

I assume not, but it's not clear who owns the `file_type` now that it isn't part of the `RemoteFile` anymore

girarda (commented):
just a couple of questions around the ownership of the file_type and the stream_reader instantiation then we'll be good to 🚢

airbyte-cdk/python/airbyte_cdk/sources/file_based/remote_file.py
clnoll: Makes sense to me. Moved.

airbyte-cdk/python/unit_tests/sources/file_based/in_memory_files_source.py
clnoll: `file_type` is defined per-stream.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_stream_reader.py
clnoll: In S3, if the trailing "/" is removed, the search wont' be restricted to a directory. For instance, the prefix `a/b` would match `a/bbbb.csv` instead of "everything under directory b". So allowing the "/" just allows us to be more restrictive about which files are requested.

clnoll (commented):
Thanks @girarda - responded to your questions. Since this PR was split off from one that includes [changes to source-S3](https://github.com/airbytehq/airbyte/pull/28786/files), some of the conversation is happening there.

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
brianjlai: just copying the convo over to continue here

> To give some context to other reviewers - the basic background here is that config information is needed by both the FileBasedSource and the StreamReader, but for different reasons; FileBasedSource needs it to configure streams, and the StreamReader needs it so that it knows how to, e.g. authenticate with the storage system.

In this PR, I'm allowing FileBasedSource to read the config from disk and parse it (like normal), and once parsed, the source sets the config on the StreamReader. However, the type of config accepted by FileBasedSource is AbstractFileBasedSpec, and it only cares about keys that are source-agnostic, whereas StreamReader cares about keys that are specific to the 3rd party that it's reading from. For example, S3 will be looking for a aws_access_token_id key, but FileBasedSource won't itself need that.

So that leads us to this situation, where the S3 StreamReader's config setter requires a config of type (S3) Config, but the interface config setter takes an AbstractFileBasedSpec, which doesn't have S3-specific keys. To solve this, we assert that we were given the correct Config type for our type of StreamReader.

One alternative route that I went down involved reading in the config prior to the initialization of the Source, so that it could be given to the StreamReader as an argument. Since the Source requires a StreamReader, it could get the config off of that. This is somewhat undesirable because we end up reading the config twice (because AbstractSource still reads it deep within the CDK code), and it also deviates from the pattern of letting the Source validate the config, which may have error handling behavior that we want to keep. For those two reasons I'd prefer to keep the code as-is. But I'm open to other opinions.

Thanks for adding the comment in the setter. The aspect Alex brought up makes sense, and then there's the drawback that every file based implementation will have to remember to check the instance type in their setter. But a better solution still isn't obvious to me (although I'm not sure if we would end up with a type mismatch that would still pass validation)

girarda: We could probably remove the property setter by passing the config as a parameter to `open_file`, and `get_matching_files`. That would however force us to verify the config in both methods instead of only in the setter. 

I'm fine with leaving as-is then

clnoll: Yeah I think my preference would be to leave it as-is for now. It's a bit awkward but is not going to lead to an insidious issue - if the stream reader needs the config (which the S3 connector does) there will be a [total failure](https://github.com/airbytehq/airbyte/pull/28786/files#diff-b5e8d7cf7e79880dd2e77408410825335670decbc45f4960d2cf07ac323bae86R48) if it's not set.

girarda (approved):
thanks for the additional context @clnoll  :shipit:

Issue #28016, opened 2023-07-06
POC: Python connectors as embedded langchain and llama index dataloaders 


flash1293:
## Acceptance criteria

* There is an open PR to add at least one connector as a llama index loader to the llama hub package
* There is an open PR to add at least one connector as a langchain loader to the langchain package

PR #28873, opened 2023-07-31
CDK: Embedded reader utils

flash1293:
Based on https://github.com/airbytehq/airbyte/pull/28854

Related to https://github.com/airbytehq/airbyte/issues/28016

This PR adds basic helpers to build embedded loaders in other python programs so they can leverage existing python sources.

The main part of this PR is adding the `BaseEmbeddedIntegration` class that wraps a source into a more generic interface that can easily be adapted for langchain and llama index loaders. Once it's merged, it can be used to make https://github.com/emptycrown/llama-hub/pull/428 ready for review.

Feature set:
* Load all records from any stream, just by specifying the config and the stream name
* Incrementally load new records by passing in an optional state (you can get that state from the `last_state` member)
* Easily wrap CDK-based sources (implements `AbstractSource` at the top level)

This PR is kept slim on purpose to be able to ship some loaders quickly.

Things that can be implemented in follow-up PRs:
* Add convenience functionality like checking whether the config matches the spec or not (and fail with a good error message in this case)
* Allow to wrap docker-based connectors by providing a second runner


airbyte-cdk/python/airbyte_cdk/sources/embedded/runner.py
```diff
@@ -0,0 +1,34 @@
...
+    def discover(self, config: TConfig) -> AirbyteCatalog:
+        pass
+
+    @abstractmethod
+    def read(self, config: TConfig, catalog: ConfiguredAirbyteCatalog, state: Optional[AirbyteStateMessage]) -> Iterable[AirbyteMessage]:
+        pass
+
+
+class CDKRunner(SourceRunner[TConfig]):
+    def __init__(self, source: AbstractSource, name: str):
```
girarda, line 26: why does this have to be an AbstractSource / why not accept any Source?

airbyte-cdk/python/airbyte_cdk/sources/embedded/tools.py
```diff
@@ -0,0 +1,25 @@
...
+
+from airbyte_cdk.models import AirbyteStream
+
+
+def get_first(iterable: Iterable[Any], predicate: Callable[[Any], bool] = lambda m: True) -> Optional[Any]:
+    return next(filter(predicate, iterable), None)
+
+
+def get_defined_id(stream: AirbyteStream, data: Dict[str, Any]) -> Optional[str]:
+    import dpath
```
girarda, line 15: why is dpath imported from the method instead of at the top of the file?

airbyte-cdk/python/airbyte_cdk/sources/embedded/runner.py
flash1293: It can be any source, fixed this

airbyte-cdk/python/airbyte_cdk/sources/embedded/tools.py
flash1293: Fixed

girarda (approved):

airbyte-cdk/python/airbyte_cdk/sources/embedded/base_integration.py
```diff
@@ -0,0 +1,47 @@
...
+            configured_catalog = create_configured_catalog(stream, sync_mode=SyncMode.full_refresh)
+        else:
+            configured_catalog = create_configured_catalog(stream, sync_mode=SyncMode.incremental)
+
+        for message in self.source.read(self.config, configured_catalog, state):
+            if message.type == Type.RECORD:
+                output = self._handle_record(message.record, get_defined_id(stream, message.record.data))
+                if output:
+                    yield output
+            elif message.type is Type.STATE and message.state:
```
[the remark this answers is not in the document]
flash1293: As it's just in-memory I think it's OK - if they want the guarantees from checkpointing, they can use hosted Airbyte (not trying to re-build the whole platform here :) )

PR #28885, opened 2023-07-31
File-based CDK: reorganize FileReadMode to fix circular import

clnoll:
Move `FileReadMode` to `file_based_stream_reader.py`, which is where it's ultimately required. This fixes a circular import.


brianjlai (approved):
lgtm!

PR #28891, opened 2023-08-01
Metadata: Write registry entry to docker repo override location

### LINKED ISSUE #28884: Registry entries for connector variants
## What

To look up a registry entry, we use the following url pattern `/files/metadata/[dockerRepository]/[dockerImageTag]/[registry].json`

For example, to get the cloud registry entry for github verison 1.0.2, you can go to https://connectors.airbyte.com/files/metadata/airbyte/source-github/1.0.2/cloud.json.

As of https://github.com/airbytehq/airbyte-platform-internal/pull/7965, we now use these registry entries when loading overrides to get all relevant metadata for the version.

However, this does not work for strict-encrypt. Let's take source-mysql for example. Cloud has the dockerRepostory as `source-mysql-strict-encrypt`, but the registry does not generate an entry for the strict-encrypt variant. Instead, we're handling these as registry overrides for the regular variant, `source-mysql`. This means that no registry entries exist under `/files/metadata/airbyte/source-mysql-strict-encrypt/`, and when we try getting the versioned registry entry for -strict-encrypt overrides, it 404s. 

## Workaround

For readers that need to set an override before this is solved, there's a manual workaround.

1. Find the cloud.json registry entry for the regular variant (e.g. /files/metadata/airbyte/source-mysql/2.1.0/cloud.json`, and download it.
2. Modify the `dockerImageTag` on it to the pre-release version.
3. Upload it to e.g. `/files/metadata/airbyte/source-mysql-strict-encrypt/2.1.0-dev. 16a1b9b8ab/cloud.json`

Cloud should now be able to find the relevant registry entry.

bnchrch:
## Overview
closes #28884

This writes any docker repo overrides to their cannonical version path.

## Nots for reviewer
I put this in a hacks file as the idea of having docker image overrides as a matter of process seems like a piece of legacy that we should be moving away from.

### DIFF
--- a/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_entry.py
+++ b/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_entry.py
@@ -19,7 +19,7 @@ from orchestrator.utils.object_helpers import deep_copy_params
 from orchestrator.utils.dagster_helpers import OutputDataFrame
 from orchestrator.models.metadata import MetadataDefinition, LatestMetadataEntry
 from orchestrator.config import get_public_url_for_gcs_file, VALID_REGISTRIES, MAX_METADATA_PARTITION_RUN_REQUEST
-
+import orchestrator.hacks as HACKS
 from typing import List, Optional, Tuple, Union
 
 PolymorphicRegistryEntry = Union[ConnectorRegistrySourceDefinition, ConnectorRegistryDestinationDefinition]
@@ -189,7 +189,6 @@ def get_registry_entry_write_path(metadata_entry: LatestMetadataEntry, registry_
         raise Exception(f"Metadata entry {metadata_entry} does not have a file path")
 
     metadata_folder = os.path.dirname(metadata_path)
-    print(f"metadata_folder: {metadata_folder}")
     return os.path.join(metadata_folder, registry_name)
 
 
@@ -213,6 +212,7 @@ def persist_registry_entry_to_json(
     registry_entry_write_path = get_registry_entry_write_path(metadata_entry, registry_name)
     registry_entry_json = registry_entry.json(exclude_none=True)
     file_handle = registry_directory_manager.write_data(registry_entry_json.encode("utf-8"), ext="json", key=registry_entry_write_path)
+    HACKS.write_registry_to_overrode_file_paths(registry_entry, registry_name, metadata_entry, registry_directory_manager)
     return file_handle
 
 
@@ -240,6 +240,7 @@ def generate_and_persist_registry_entry(
     registry_model = ConnectorModel.parse_obj(registry_entry_with_spec)
 
     file_handle = persist_registry_entry_to_json(registry_model, registry_name, metadata_entry, metadata_directory_manager)
+
     return file_handle.public_url
 
 
--- a/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/hacks.py
+++ b/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/hacks.py
@@ -53,7 +53,7 @@ def write_registry_to_overrode_file_paths(
         by the platform when looking for a specific registry entry. In this case, for cloud, it would be
         gs://my-bucket/metadata/source-postgres-strict-encrypt/dev.123/cloud.json
 
-        Ideally we would not have to do this, but the combonation of prereleases and common overrides
+        Ideally we would not have to do this, but the combination of prereleases and common overrides
         make this nessesary.
 
     Args:
--- a/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/hacks.py
+++ b/airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/hacks.py
@@ -1,0 +1,78 @@
+from dagster import get_dagster_logger
+from dagster_gcp.gcs.file_manager import GCSFileManager, GCSFileHandle
+
+from orchestrator.models.metadata import LatestMetadataEntry
+from metadata_service.constants import METADATA_FILE_NAME
+from metadata_service.gcs_upload import get_metadata_remote_file_path
+from metadata_service.models.generated.ConnectorRegistrySourceDefinition import ConnectorRegistrySourceDefinition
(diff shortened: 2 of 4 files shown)


airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/hacks.py
```diff
@@ -0,0 +1,78 @@
...
+
+    assumed_metadata_file_path = get_metadata_remote_file_path(docker_reposiory, docker_version)
+    registry_entry_file_path = assumed_metadata_file_path.replace(METADATA_FILE_NAME, registry_name)
+    return registry_entry_file_path
+
+def _check_for_invalid_write_path(write_path: str):
+    """Check if the write path is valid."""
+
+    if "latest" in write_path:
+        raise ValueError("Cannot write to a path that contains 'latest'. That is reserved for the latest metadata file and its direct transformations")
```
pedroslopez, line 33: 👍🏼  dont want these to go to the full registries

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/hacks.py
```diff
@@ -0,0 +1,78 @@
...
+        the same path as the metadata.yaml file that created them. This is fine for the
+        most cases, but when the docker repository is overridden, the registry entry
+        files need to be written to a different path.
+
+        For example if source-postgres:dev.123 is overridden to source-postgres-strict-encrypt:dev.123
+        then the oss.json file needs to be written to the path that would be assumed
+        by the platform when looking for a specific registry entry. In this case, for cloud, it would be
+        gs://my-bucket/metadata/source-postgres-strict-encrypt/dev.123/cloud.json
+
+        Ideally we would not have to do this, but the combonation of prereleases and common overrides
```
pedroslopez, line 56:
```suggestion
        Ideally we would not have to do this, but the combination of prereleases and common overrides
```

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/hacks.py
```diff
@@ -0,0 +1,78 @@
...
+    Underlying issue:
+        The registry entry files (oss.json and cloud.json) are traditionally written to
+        the same path as the metadata.yaml file that created them. This is fine for the
+        most cases, but when the docker repository is overridden, the registry entry
+        files need to be written to a different path.
+
+        For example if source-postgres:dev.123 is overridden to source-postgres-strict-encrypt:dev.123
+        then the oss.json file needs to be written to the path that would be assumed
+        by the platform when looking for a specific registry entry. In this case, for cloud, it would be
+        gs://my-bucket/metadata/source-postgres-strict-encrypt/dev.123/cloud.json
```
pedroslopez, line 54: Nice, thanks for writing this out!

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_entry.py
```diff
@@ -213,6 +212,7 @@ def persist_registry_entry_to_json(
     registry_entry_write_path = get_registry_entry_write_path(metadata_entry, registry_name)
     registry_entry_json = registry_entry.json(exclude_none=True)
     file_handle = registry_directory_manager.write_data(registry_entry_json.encode("utf-8"), ext="json", key=registry_entry_write_path)
+    HACKS.write_registry_to_overrode_file_paths(registry_entry, registry_name, metadata_entry, registry_directory_manager)
```
pedroslopez, line 215: Q: This means when publishing a pre-release, will we need to publish a pre-release for both the regular _and_ strict-encrypt variants?

I ask because I _think_ this reads as "when persisting a registry entry, also persist it in this other location", but in the case of pre-releases for only the strict_encrypt no registry entry is generated in the first place

pedroslopez (approved):
This looks good, but I think this means we now need to make sure to publish both the regular variant and strict-encrypt variant when doing pre-releases to generate the registry entry for it. Is that right? If so, is there any way around this?

airbyte-ci/connectors/metadata_service/orchestrator/orchestrator/assets/registry_entry.py
bnchrch: @pedroslopez nope!

This just means that if we find a registry entry file (oss.json, cloud.json) where the dockerRepository has been overriden then we make sure to also write it to
`metadata-bucket/metadata/{overridden-dockerrepo}/{version}/cloud.json`

bnchrch: But that does mean we are in a weird spot for when to and when not to publish the strict encrypt variant

bnchrch: I dont think we have a good short term way around this.

But at the same time I think its good that the organization feel a bit of pain around these strict encrypt variants

pedroslopez: @bnchrch yes, but in the case of pre-releases no registry entry file is in place for the non-strict-encrypt variant because we don't tend to publish pre-releases for both (at least not historically).

For example, see source-postgres-strict-encryppt:
image

and source-postgres

image


The pre-releases that are in there are different because people just usually dont publish both

pedroslopez: With this change, I think the process now would be:

1. publish pre-relese for strict-encrypt
2. publish pre-release for regular variant so the registry entry is made for it and copied over to strict-encrypt

Does that make sense?: Am I missing something?

bnchrch: @pedroslopez Makes sense! Looks like I published my edit at the same time you published your explanation 😅 

https://github.com/airbytehq/airbyte/pull/28891#discussion_r1280031645

bnchrch: We just have a failing test. Going to fix that up then merge

PR #28897, opened 2023-08-01
[connectors-ci] sentry: ignore error logs without exceptions

pedroslopez:
## What

We're getting a lot of logged errors that aren't actionable. We still want logger-based exception reporting though. This PR avoids sending error logs that don't have attached stacktraces.

## How

Use sentry's [before_send](https://docs.sentry.io/platforms/python/configuration/filtering/#using-platformidentifier-namebefore-send-) hook to filter out these logs.

### DIFF
--- a/airbyte-ci/connectors/pipelines/pyproject.toml
+++ b/airbyte-ci/connectors/pipelines/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "pipelines"
-version = "0.1.1"
+version = "0.2.1"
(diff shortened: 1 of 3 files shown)


airbyte-ci/connectors/pipelines/pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "pipelines"
-version = "0.1.1"
+version = "0.2.1"
```
pedroslopez, line 7: note: 0.2.0 wasn't bumped properly

evantahler (approved):
✔️

Issue #28899, opened 2023-08-01
File CDK: add partitioning testing for parquet parser


clnoll:
S3 CATs identified an issue where the parquet parser wasn't handling partitions. We should add test coverage for this case to the file-based CDK itself.

girarda:
[we added tests since this issue was written](https://github.com/airbytehq/airbyte/blob/master/airbyte-cdk/python/unit_tests/sources/file_based/scenarios/parquet_scenarios.py#L217-L372). feel free to re-open if you think we nee more and we can prioritize them

PR #28900, opened 2023-08-01
File cdk parser and cursor updates

clnoll:
## What
This PR consists of a few updates to handle some issues encountered during integration testing.

1. Updates the parquet parser so that we're returning all partitions. Unfortunately I haven't found a great way to add local test coverage for the exact but that was encountered, but it is covered by S3 CATs. I created an [issue](https://github.com/airbytehq/airbyte/issues/28899) to add it at a later time.
2. Updates the `DefaultFileBasedCursor` datetime format to agree with the format expected by CATs.
3. Returns the cursor in the state message, as required by CATs.

## Recommended reading order
1. Parquet parser
2. File-based cursor


airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/default_file_based_cursor.py
```diff
@@ -13,7 +13,7 @@
 
 class DefaultFileBasedCursor(FileBasedCursor):
     DEFAULT_DAYS_TO_SYNC_IF_HISTORY_IS_FULL = 3
-    DATE_TIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
+    DATE_TIME_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
```
brianjlai, line 16: interesting, okay i'll revert my start_date PR back to this then. I was originally converting back out to milliseconds

So do our CAT tests in general just not support anything more granular than seconds? I'm curious because I do see some configs where we use more granular precisions. Or is this specifically to retain the precision being used by S3?

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/parquet_parser.py
```diff
@@ -27,11 +29,16 @@ async def infer_schema(
...
-        parquet_file = pq.ParquetFile(stream_reader.open_file(file, self.file_read_mode, logger))
-        parquet_schema = parquet_file.schema_arrow
+        with stream_reader.open_file(file, self.file_read_mode, logger) as fp:
+            parquet_file = pq.ParquetFile(fp)
+            parquet_schema = parquet_file.schema_arrow
+
+        # Inferred non-partition schema
         schema = {field.name: ParquetParser.parquet_type_to_schema_type(field.type, parquet_format) for field in parquet_schema}
+        # Inferred partition schema
+        partition_columns = {x.split("=")[0]: {"type": "string"} for x in self._extract_partitions(file.uri)}
```
brianjlai, line 39: nit can we use a more descriptive variable name for the current partition than `x`

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/parquet_parser.py
```diff
@@ -45,13 +52,18 @@ def parse_records(
...
+            for row_group in range(reader.num_row_groups):
+                batch_dict = reader.read_row_group(row_group).to_pydict()
+                for record_values in zip(*batch_dict.values()):
+                    record = dict(zip(batch_dict.keys(), record_values))
+                    record.update(partition_columns)
+                    yield record
+
+    @staticmethod
+    def _extract_partitions(filepath: str) -> List[str]:
+        return [unquote(x) for x in filepath.split(os.sep) if "=" in x]
```
brianjlai, line 66: same here

brianjlai (commented):
Mostly non-blocking comments! 

But wanted to discuss the switch from milliseconds to seconds precision. Are we converting the entire file based CDK to the less precise type to satisfy the S3 CAT tests? Because if so it does feel a little bit backwards that we're reverting precision in the CDK to support a specific connector. Not taking CAT backwards compatibility into account, we do have the ability to add more transformations from a legacy config to the new config since we already have to do other ones.

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/default_file_based_cursor.py
brianjlai: got it. thanks for checking!

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/default_file_based_stream.py
```diff
@@ -78,7 +78,7 @@ def read_records_from_slice(self, stream_slice: StreamSlice) -> Iterable[Mapping
         parser = self.get_parser(self.config.file_type)
         for file in stream_slice["files"]:
             # only serialize the datetime once
-            file_datetime_string = file.last_modified.strftime("%Y-%m-%dT%H:%M:%SZ")
+            file_datetime_string = file.last_modified.strftime("%Y-%m-%dT%H:%M:%S.%fZ")
```
brianjlai, line 81: nit: lets put this as a constant on the class just in case we do end up using time in multiple places

brianjlai (approved):
looks good to me!

airbyte-cdk/python/airbyte_cdk/sources/file_based/stream/cursor/default_file_based_cursor.py
```diff
@@ -48,11 +48,21 @@ def add_file(self, file: RemoteFile) -> None:
                 )
 
     def get_state(self) -> StreamState:
-        state = {
-            "history": self._file_to_datetime_history,
-        }
+        state = {"history": self._file_to_datetime_history, "_ab_source_file_last_modified": self._get_cursor()}
```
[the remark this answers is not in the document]
clnoll: Hmm that isn't actually covered by the CATs. Do you know what we usually do in that situation?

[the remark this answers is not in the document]
clnoll: Actually I think I'll leave it as-is for now, since we don't actually use the cursor field for deciding what to sync; it's just being used as a tool for integration tests.

PR #28931, opened 2023-08-01
File-based CDK: fix flaky test (ignore order of oneOf in dict comparison)

clnoll:
Ignore order of oneOf in dict comparison.

### DIFF
--- a/airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
+++ b/airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
@@ -271,7 +271,27 @@ spec_scenarios = [
 
 @pytest.mark.parametrize("scenario", spec_scenarios, ids=[c.name for c in spec_scenarios])
 def test_spec(capsys: CaptureFixture[str], scenario: TestScenario) -> None:
-    assert spec(capsys, scenario) == scenario.expected_spec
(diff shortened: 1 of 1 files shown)


airbyte-cdk/python/unit_tests/sources/file_based/test_scenarios.py
```diff
@@ -271,7 +271,27 @@ def _verify_expected_logs(logs: List[Dict[str, Any]], expected_logs: Optional[Li
...
+    if not isinstance(dict1, dict) or not isinstance(dict2, dict):
+        return dict1 == dict2
+
+    if set(dict1.keys()) != set(dict2.keys()):
+        return False
+
+    for key in dict1:
+        if key == 'oneOf' and isinstance(dict1[key], list) and isinstance(dict2[key], list):
+            # If the key is 'oneOf', and the values are lists, then compare ignoring order
+            l1, l2 = sorted(dict1[key], key=lambda x: x["type"]), sorted(dict2[key], key=lambda x: x["type"])
```
maxi297, line 287: Can we have multiple entries of the same type? I can see a case where this comparison could pass even though it should fail. For example, expected could have (List[int] and List[str]) and actual could have (List[float] and List[str]). 

Can we cast the list as a set and compare the sets? This assumes that entries are hashable which I think is a fair assumption in this case

clnoll:
This issue has stopped popping up so I'm closing this PR but will leave the branch around for a while.

Issue #28782, opened 2023-07-27
Support multiple datetime format for cursor field


maxi297:
### What area the feature impact?

Connectors

### Revelant Information

Business Central API seems to return different datetime format for the cursor field: `%Y-%m-%dT%H:%M:%S.%fZ` and `%Y-%m-%dT%H:%M:%SZ`. This means that configuring an incremental sync is not possible for this API.

## Proposed solution
Allow for multiple datetime formats. If multiple `datetime_format` are provided, `start_datetime` and `end_datetime` needs to define a datetime_format

```
  incremental_sync:
    type: "DatetimeBasedCursor"
    start_datetime:
      datetime: {{ config['start_date'] }}
      datetime_format: "%Y-%m-%d %H:%M:%S"
    end_datetime:
      datetime: {{ config['end_date'] }}
      datetime_format: "%Y-%m-%d %H:%M:%S"
    datetime_format: "%Y-%m-%dT%H:%M:%SZ"
    cursor_datetime_format: 
      - "%Y-%m-%dT%H:%M:%SZ"
      - "%Y-%m-%dT%H:%M:%S.%fZ"
    cursor_field: "publishedAt"
```

## Dependency
If we allow datetime_format to be a list, it doesn't align with https://github.com/airbytehq/airbyte/issues/27153. Possible solutions:
* do this issue before so that `cursor_datetime_format` can be a list and not `datetime_format`
* have `outgoing_datetime_format` instead of `cursor_datetime_format` and keep `datetime_format` for the cursor

maxi297:
Grooming notes:
* cursor_datetime_format should be in order so that if the connector dev knows that one is more frequent than the other, he can defined it first and this will be more performant

PR #28936, opened 2023-08-01
[ISSUE #28782] support multiple cursor field datetime formats

maxi297:
## What
Addresses #28782 and allow to fix issues in prod for Amplitude

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -625,7 +625,7 @@ definitions:
           - "2020-01-1T00:00:00Z"
           - "{{ config['start_time'] }}"
       cursor_datetime_formats:
-        title: Cursor Datetime Format
+        title: Cursor Datetime Formats
         description: The possible formats for the cursor field
         type: array
         items:
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
@@ -624,6 +624,16 @@ definitions:
         examples:
           - "2020-01-1T00:00:00Z"
           - "{{ config['start_time'] }}"
+      cursor_datetime_formats:
+        title: Cursor Datetime Format
+        description: The possible formats for the cursor field
+        type: array
+        items:
+          type: string
+          examples:
+            - "%Y-%m-%dT%H:%M:%S.%f%z"
+            - "%Y-%m-%d"
+            - "%s"
       cursor_granularity:
         title: Cursor Granularity
         description:
--- a/airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
@@ -161,9 +161,6 @@ class DatetimeBasedCursor(Cursor):
     def _format_datetime(self, dt: datetime.datetime) -> str:
         return self._parser.format(dt, self.datetime_format)
 
-    def _format_cursor_datetime(self, dt: datetime.datetime) -> str:
-        return self._parser.format(dt, self.cursor_datetime_formats[0])
-
     def _partition_daterange(self, start: datetime.datetime, end: datetime.datetime, step: Union[datetime.timedelta, Duration]):
         start_field = self.partition_field_start.eval(self.config)
         end_field = self.partition_field_end.eval(self.config)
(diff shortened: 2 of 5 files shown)


maxi297:
There is the mypy situation going on but I don't know if it's worth fixing this while having an issue in production

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -624,6 +624,16 @@ definitions:
         examples:
           - "2020-01-1T00:00:00Z"
           - "{{ config['start_time'] }}"
+      cursor_datetime_formats:
+        title: Cursor Datetime Format
```
brianjlai, line 628: nit: plural

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
```diff
@@ -154,17 +158,22 @@ def _calculate_cursor_datetime_from_state(self, stream_state: Mapping[str, Any])
             return self.parse_date(stream_state[self.cursor_field.eval(self.config)])
         return datetime.datetime.min.replace(tzinfo=datetime.timezone.utc)
 
-    def _format_datetime(self, dt: datetime.datetime):
+    def _format_datetime(self, dt: datetime.datetime) -> str:
```
clnoll, line 161: Nit: looks like this is no longer used so can be deleted.

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
```diff
@@ -52,7 +52,7 @@ class DatetimeBasedCursor(Cursor):
     datetime_format: str
     config: Config
     parameters: InitVar[Mapping[str, Any]]
-    _cursor: str = field(repr=False, default=None)  # tracks current datetime
+    _cursor: Optional[str] = field(repr=False, default=None)  # tracks current datetime
```
clnoll, line 55: Was this type wrong, or are we explicitly changing it to be `Optional`?

airbyte-cdk/python/unit_tests/sources/declarative/incremental/test_datetime_based_cursor.py
```diff
@@ -524,6 +524,48 @@ def test_parse_date(test_name, input_date, date_format, date_format_granularity,
     assert expected_output_date == output_date
 
 
+@pytest.mark.parametrize(
+    "test_name, input_date, date_formats, expected_output_date",
+    [
```
clnoll, line 529: Should we have a test case that we use `datetime_format` if `date_formats` is empty?

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
```diff
@@ -154,17 +158,22 @@ def _calculate_cursor_datetime_from_state(self, stream_state: Mapping[str, Any])
             return self.parse_date(stream_state[self.cursor_field.eval(self.config)])
         return datetime.datetime.min.replace(tzinfo=datetime.timezone.utc)
 
-    def _format_datetime(self, dt: datetime.datetime):
+    def _format_datetime(self, dt: datetime.datetime) -> str:
         return self._parser.format(dt, self.datetime_format)
 
+    def _format_slice_datetime(self, dt: datetime.datetime) -> str:
+        if not self.cursor_datetime_formats:
+            raise RuntimeError("DatetimeBasedCursor can't operate without cursor_datetime_formats")
```
brianjlai, line 166: When would this happen? When we instatiate the DatetimeBasedCursor, we set it to `self.cursor_datetime_formats = [self.datetime_format]` if cursor_datetime_formats is empty. And since `datetime_format` is a required field it'll always have a value

clnoll (approved):
LGTM, just a few minor comments.

brianjlai (approved):
only non-blocking comments. lgtm

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
maxi297: This is not a nit since this is what is sent to the API and therefore, it should be used. I'll update this. My guess is that when we save the end of the slice, we should parse and format using the first `cursor_datetime_format`

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
maxi297: It was typed wrong

airbyte-cdk/python/unit_tests/sources/declarative/incremental/test_datetime_based_cursor.py
maxi297: I kept the previous tests which are doing that

airbyte-cdk/python/airbyte_cdk/sources/declarative/incremental/datetime_based_cursor.py
maxi297: It should not be an issue. I removed this code

PR #28767, opened 2023-07-27
connectors-ci: improve pytest result evaluation

alafanechere:
## What
Relates to #27425 , #28423

### Original problem
When running pytest in a dagger container we face two challenges:
* When a pytest suite is empty or all tests are skipped pytest will return a 5 exit code. exit code` >0` in Dagger raise ExecError, but we want to mark this step as skipped and not fail.
* When CAT (which is a pytest suite) is failing still need to perform after the facts operations to retrieve updated connector configurations from the container. When a container raises an ExecError we can't interact with it's filesystem without raising the same ExecError again.

#### How we overcame this problem so far
We installed a pytest plugin to always make pytest suite exit with code `0`: `pytest-custom_exit_code`. We inferred success / failure / skipping of test by parsing the stdout log. This is not robust as the stdout content can change according to the pytest version or if a CAT run outputs unexpected stuff to stdout. **This led to the false positive errors mentioned in #27425**

#### The new approach to overcome this problem
We still make CAT exec artificially exit with a `0` exit code, but we store the original exit code to a file in the container filesystem: `/exit_code`. We `get_step_result` is run it will check if such a file exist on the container and return the in file exit code value instead of the pure container exit code. This allows us to capture the actual exit code that pytest returned.   

## How
The new approach mentioned above is mainly introduced via:
- a new hack: `hacks.never_fail_exec` function:
https://github.com/airbytehq/airbyte/blob/ddb7345e008c7e3c24db77a04bc6a9c45855f528/airbyte-ci/connectors/pipelines/pipelines/hacks.py#L145
- updating the `Step.get_step_result` function to read in file exit code if `/exit_code` file is available on the container.
https://github.com/airbytehq/airbyte/blob/ddb7345e008c7e3c24db77a04bc6a9c45855f528/airbyte-ci/connectors/pipelines/pipelines/bases.py#L245

**I tried to thoroughly unit test these change to show how async test with a dagger client work:**
https://github.com/airbytehq/airbyte/blob/ddb7345e008c7e3c24db77a04bc6a9c45855f528/airbyte-ci/connectors/pipelines/tests/tests/test_common.py#L20

### Centralizing the AcceptanceTest logic
The logic to build the CAT container was in the  `environment.with_connector_acceptance_test` function. I decided to move it to `common.AcceptanceTest._build_connector_acceptance_test` as `environment` functions are originally made to be reused in multiple steps. This function is very specific to `AcceptanceTest`.

### Make all step write the `stdout` / `stderr` to a log file
[...]

PR #28938, opened 2023-08-01
connectors-ci: skipped status  when metadata upload exits with 5

alafanechere:
## What
https://github.com/airbytehq/airbyte/pull/28767 changed how we identify steps as skipped.
The exit code that leads to the "skipped" step status is now defined at the `Step` subclass level and is not global.
It introduced a regression for the MetadataUpload step - when its container exits with 5 it means the upload was skipped because the metadata yaml already exists on the bucket. 

## How
Declare in `MetadataUpload` that the skipped exit code is 5

### DIFF
--- a/airbyte-ci/connectors/pipelines/README.md
+++ b/airbyte-ci/connectors/pipelines/README.md
@@ -378,6 +378,7 @@ This command runs the Python tests for a airbyte-ci poetry package.
 
 | Version | PR                                                        | Description                                                                                  |
 |---------|-----------------------------------------------------------|----------------------------------------------------------------------------------------------|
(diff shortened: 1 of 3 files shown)


evantahler (approved):
5

evantahler:
I don't really understand why 5...

alafanechere:
@evantahler when we implemented the metadata service upload we made its process exit with `5` when the upload was skipped because the file hash is the same on the bucket. It just a way for this process run inside a container to tell airbyte-ci that the upload was skipped. We picked 5 because pytest exists with the same status code when all tests are skipped.

PR #28973, opened 2023-08-02
Update name of langchain destination

flash1293:
As discussed, updates the docs to align with the metadata information,


pedroslopez (approved):
Thanks!

There are some other references to langchain in the doc that I'm not sure if we want to change, but feel free to merge as is

PR #29024, opened 2023-08-03
Metadata: Remove leading underscore

### LINKED ISSUE #28712: Update Metadata Schema to include new fields
## Overview
We are moving to new internal and external quality levels.

We need to
1. remove `data.releaseStage`
2. add `data.supportLevel`
3. Add the new internal fields outlined in the [tech spec](https://docs.google.com/document/d/1CTr-MZJ14IQClrslpMLdC9xRdow0PCELIoT0pwsF0o0/edit)
4. and set both according to the [PRD here](https://docs.google.com/document/d/161FF__wOjwS6j-1fVyHexDD1cJCQ5RJ2LMC8FI2N6yI/edit#heading=h.ptbg5pj2cpm4)

## Notes

Ensure the new fields can be ingested on the platform side

bnchrch:
## Overview
Python doesn't mind fields with leading underscores

But [pydantic does](https://docs.pydantic.dev/latest/usage/fields/) and the way our [json schema to model generator ](https://koxudaxi.github.io/datamodel-code-generator/) works is by prepending `field_` to all the underscore fields and using the original name as an alias.

We could continue to use underscore fields `_ab_internal` by being very careful about when and how we transform our pydantic models to dicts, to json and back again.

However we would still have to reference the fields in code with the prefix e.g. `some_model["field_ab_internal"]`

And to me that was the reason not to as it would brake new devs tracing code to find out how we use metadata files

## Solution
Remove all leading prefixes

## After merge
1. Stop publish pipeline
2. Use metadata_service upload to upload all changed metadata files

closes #28712


airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
```diff
@@ -25,7 +25,7 @@ properties:
       - githubIssueLabel
       - connectorSubtype
       - releaseStage
-    additionalProperties: false
+    additionalProperties: true
```
alafanechere, line 28: Why are we allowing additionalProperties now?

alafanechere: Is it to allow `_` prefixes? I'm not sure I get why we want to continue allowing `_` prefix and I'm afraid enabling `additionalProperties` weakens the validation and can lead to sneaky errors.

airbyte-ci/connectors/metadata_service/lib/metadata_service/utils.py
```diff
@@ -14,7 +14,7 @@ def to_json_sanitized_dict(pydantic_model_obj: BaseModel, **kwargs) -> dict:
         dict: a sanitized dictionary
     """
     defalut_kwargs = {
```
alafanechere, line 16:
```suggestion
    default_kwargs = {
```

alafanechere: Isn't this function already defined in _apply_default_pydantic_kwargs?

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/generated/ConnectorMetadataDefinitionV0.py
```diff
@@ -185,7 +185,7 @@ class Config:
 
 class Data(BaseModel):
     class Config:
-        extra = Extra.forbid
+        extra = Extra.allow
```
alafanechere, line 188: Isn't allowing extra risky? Forbidding extra is a nice way to have consistency guarantees across metadata files

alafanechere (changes requested):
I'm mainly requesting changes because I believe disabling publishing on the `publish-connectors.yml` workflow should happen on this branch to not trigger a global publish pipeline on merge.
I'm also not totally sure about the purpose of allowing extra fields / additionalProperties

airbyte-ci/connectors/metadata_service/lib/metadata_service/models/src/ConnectorMetadataDefinitionV0.yaml
bnchrch: So Were in a messy middle after discovering the pydantic model generated did not like the leading underscore.

The immediate fix(s) we deployed
1. Allowed prerealease to use the alias when writing the prerelease metadata file
1. Updated the additionalProperties to true for the registry to handle a parsing error resulting from above

Now going forward we've got a few metadata files out there that have _ab_internal when they should have ab_internal
This is why im setting additionalProperties to true, just so things dont blow up when rolling these new fields out, and in the next version of the metadata we can revert it back to false

bnchrch:
> disabling publishing on the publish-connectors.yml workflow should happen on this branch to not trigger a global publish pipeline on merge.

Can we not just be quick with cancelling the publish workflow via the UI?

My worry is that if we disable it via code a connector from another develoeprs merge does not get published

alafanechere (approved):

bnchrch:
Used my admin privs to merge this one as I dont want to run ALL connector tests (required check) for a very small and currently unused metadata file change.

PR #29025, opened 2023-08-03
Fix title & description of datetime_format field

lmossman:
## What
https://github.com/airbytehq/airbyte/pull/28936 affected the semantics of the `datetime_format` field, but its title and description were never updated to reflect this.

## How
Update the title and description of `datetime_format` to reflect its new meaning as being the format for datetime values in outgoing requests to the API.

I also improved the description of the cursor_datetime_formats field since I felt it was a bit lacking


lmossman:
@maxi297 feel free to update the title or description if you think they could be worded better / more clearly

flash1293 (approved):
LGTM

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
```diff
@@ -578,9 +578,9 @@ definitions:
           - "created_at"
           - "{{ config['record_cursor'] }}"
       datetime_format:
-        title: Cursor Field Datetime Format
+        title: Outgoing Datetime Format
         description: |
-          The datetime format of the Cursor Field. Use placeholders starting with "%" to describe the format the API is using. The following placeholders are available:
+          The datetime format used to format the datetime values that are sent in outgoing requests to the API. Use placeholders starting with "%" to describe the format the API is using. The following placeholders are available:
```
maxi297, line 583: Should it be "The datetime format **is** used to format..." ?

maxi297 (approved):
Thanks for updating that! I might have cut some corners as it was developed for a P0

airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml
lmossman: I think either one would work, the way I had it was following the grammar of the original description, where the first sentence is more of a descriptor rather than a full sentence

PR #29027, opened 2023-08-03
File-based CDK: allow FileBasedSource to take a Cursor object

clnoll:
Updates the FileBasedSource so that it take an optional Cursor argument.

This will be used by source-s3 to provide an adapter, so that we can convert history from the old format to the new.

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
@@ -19,6 +19,7 @@ from airbyte_cdk.sources.file_based.file_types import default_parsers
 from airbyte_cdk.sources.file_based.file_types.file_type_parser import FileTypeParser
 from airbyte_cdk.sources.file_based.schema_validation_policies import DEFAULT_SCHEMA_VALIDATION_POLICIES, AbstractSchemaValidationPolicy
 from airbyte_cdk.sources.file_based.stream import AbstractFileBasedStream, DefaultFileBasedStream
+from airbyte_cdk.sources.file_based.stream.cursor import AbstractFileBasedCursor
 from airbyte_cdk.sources.file_based.stream.cursor.default_file_based_cursor import DefaultFileBasedCursor
 from airbyte_cdk.sources.streams import Stream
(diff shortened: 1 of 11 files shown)


airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
```diff
@@ -104,7 +107,7 @@ def streams(self, config: Mapping[str, Any]) -> List[Stream]:
                         discovery_policy=self.discovery_policy,
                         parsers=self.parsers,
                         validation_policy=self._validate_and_get_validation_policy(stream_config),
-                        cursor=DefaultFileBasedCursor(self.max_history_size, stream_config.days_to_sync_if_history_is_full),
+                        cursor=self.cursor or DefaultFileBasedCursor(self.max_history_size, stream_config.days_to_sync_if_history_is_full),
```
girarda, line 110: is it worth passing a class type instead of an instance of a cursor in the constructor so custom cursors can also depend on the stream config?

girarda (approved):
lgtm 🚀

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_based_source.py
girarda: An alternative would be for all cursors to accept `(stream_config: StreamConfig, **kwargs)`

Hmm. I don't feel too strongly either way since I don't have a good sense of how custom cursor will evolve or what information they'll need.

One thing that does feel weird, and is fine to leave out of scope for this PR, is that max_history_size and days_to_sync_if_history_is_full are related concepts so they probably both live together (either in the stream_config or not)

clnoll: Yeah I agree. 

I went ahead and tried a reorg that inits cursors with `stream_config` and does a little bit of cleanup of `max_history_size` and `days_to_sync_if_history_is_full` that I think is an improvement. LMK what you think.

girarda: I like it!

PR #29051, opened 2023-08-03
Update Cloud Docs to "Extract" and "Load"

nataliekwong:
We published a change to how we show how many records were read/written in each sync in this PR: https://github.com/airbytehq/airbyte-platform-internal/pull/8116.

This is just an update to reflect that change, and also we do 5 attempts instead of 3, so I rolled that into this update.


joeykmh (approved):
Thanks for making this change! Looks good to me

PR #29079, opened 2023-08-04
Embedded CDK: run a check before starting to load

flash1293:
This PR checks the config in embedded source integration against the spec to make sure the loader will function as intended and produce a nice error message in case the config doesn't match.


airbyte-cdk/python/airbyte_cdk/sources/embedded/runner.py
```diff
@@ -27,6 +31,9 @@ def __init__(self, source: Source, name: str):
         self._source = source
         self._logger = logging.getLogger(name)
 
+    def check(self, config: TConfig) -> AirbyteConnectionStatus:
+        return self._source.check(self._logger, config)
```
pedroslopez, line 35: So I looked through what check was actually doing, expecting it to validate things like missing required fields and such with the spec. I remembered this happened at the CDK level across the board, but was a bit surprised to find out it happens at the `AirbyteEntrypoint` level, and not in the `Source`:

https://github.com/airbytehq/airbyte/blob/642e7680b47e51d314ac8ba10e1bbf94a11c1210/airbyte-cdk/python/airbyte_cdk/entrypoint.py#L122-L123

I think this piece is the most valuable, specially as we have a raw config json that needs to be constructed by the end user and passed into the loader/connector.

We can do the same here by calling this util function:
https://github.com/airbytehq/airbyte/blob/2e099acc52aeddc172c5ee66ed98426c69449a4a/airbyte-cdk/python/airbyte_cdk/sources/utils/schema_helpers.py#L149

Though we'll need to also load spec, which _also_ seems to be a part of the AirbyteEntrypoint.

I would say just use the AirbyteEntrypoint, but that class configures things like the uncaught exception handler and the logging + works based on file paths which isn't quite what we want here.

pedroslopez: If you're worried about time added with an extra request, just doing spec validation like is done here might be enough and addresses my main worry with not doing any validation on the config.

imo it's fine to do an extra request though.

pedroslopez (changes requested):

flash1293:
@pedroslopez Good point, I adjusted the PR to only check spec compatibility which should be fairly quick. Could you take another look?

pedroslopez (approved):
Nice!

PR #29100, opened 2023-08-04
Fix: Update to use new slack op

bnchrch:
## overview
Slack webhook method was broken, likely due to a subtle version mismatch behind the scenes.

the solution here is to just use our new method.

🤠 Merging it because a report is currently broken


octavia-approvington:

octavia-approvington (approved):

PR #29164, opened 2023-08-07
[Non-Forked] ✨Source Retently: Migrate Python CDK to Low-code CDK

sajarin:
## What
Run ci checks for #29040
Migrating Source retently to Low-Code CDK
Closes https://github.com/airbytehq/airbyte/issues/28118

## How
Developed using (Configuration Based Source) low-code CDK

## Recommended reading order

1. manifest.yaml
2. schemas/*

 ##  Test
<details>
<summary>Acceptance Test</summary>
Screenshot 2023-08-03 at 6 52 13 PM
</details>

# 🚨 User Impact 🚨
**No breaking changes**, migration to low-code
<details>
<summary>Migrate Connector : Source retently</summary>

- [x] **Community Member? Grant Edit Access to Maintainers [instructions](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests)**
- [x] **Secrets in the Connector's Spec**
- [x] **Unit & integration tests added and passing. Community members, please provide proof of success locally e.g: screenshot or copy-paste unit, integration, and acceptance test output. To run acceptance tests for a Python connector, follow instructions in the README. For java connectors run ./gradlew :airbyte-integrations:connectors:<name>:integrationTest.**
- [x] **Code Reviews Completed**
  - [x] **Documentation Updated**
  - [x] **Connector's README.md**
  - [x] **Connector's bootstrap.md**
  - [x] **docs/integrations/<source or destination>/<name>.md (including changelog. See changelog example)**
  - [x] **docs/integrations/README.md**
  - [x] **airbyte-integrations/builds.md**
- [x] **PR Name follows PR Naming Conventions**
</details>

PR #29216, opened 2023-08-08
🐛 fix: correct language tag for low-code based connector

sajarin:
## What

The current low-code connector template uses the "lowcode" string for the language tag instead of "low-code" which causes QA checks to fail

## How
*Describe the solution*
Change lowcode to low-code in the metadata.yaml template file

## Recommended reading order
1. metadata.yaml.hbs

### DIFF
--- a/airbyte-integrations/connector-templates/source-configuration-based/metadata.yaml.hbs
+++ b/airbyte-integrations/connector-templates/source-configuration-based/metadata.yaml.hbs
@@ -21,5 +21,5 @@ data:
   supportLevel: community
   documentationUrl: https://docs.airbyte.com/integrations/sources/{{dashCase name}}
   tags:
-    - language:lowcode
+    - language:low-code
 metadataSpecVersion: "1.0"


sherifnada (approved):

bnchrch (changes requested):
@sajarin Do you have a link to a failure caused by `lowcode`  instead of `low-code`?

This is the code responsible for this: https://github.com/airbytehq/airbyte/blob/master/airbyte-ci/connectors/metadata_service/lib/metadata_service/validators/metadata_validator.py#L75

And im strugging to see where low-code would cause a success and lowcode a failure

sajarin:
> @sajarin Do you have a link to a failure caused by `lowcode` instead of `low-code`?
> 
> This is the code responsible for this: https://github.com/airbytehq/airbyte/blob/master/airbyte-ci/connectors/metadata_service/lib/metadata_service/validators/metadata_validator.py#L75
> 
> And im strugging to see where low-code would cause a success and lowcode a failure

Thanks for taking a look, I wasn't sure if it was actually a bug tbh. This was the PR where I was running into metadata.yaml validation errors: https://github.com/airbytehq/airbyte/pull/29164#issuecomment-1668107500

I think it's still worth merging this in for the sake of consistency with all of our other low-code connectors but maybe it's also worth broadening the scope and using this PR to add some other fields as well (like ab_internal)

Issue #28445, opened 2023-07-18
Publish Java CDK for DB-type Source connectors


aaronsteers:
Internal tech spec doc: https://docs.google.com/document/d/11VAFxQJt8UXLAsAKnxfZMjUqoXylHvkaBRxspEU4pcU/edit#heading=h.fa72b39y2m99

This would add a CDK for Java connectors. First priority will be to support Source DBs. Future scope may also expand to include destination DB connectors.

## Minimal P1 Scope

```[tasklist]
### Tasks
- [x] Publish "no-op" v0.0.1 Java CDK to Maven repository.
- [ ] https://github.com/airbytehq/airbyte/issues/28481
- [ ] https://github.com/airbytehq/airbyte/issues/28483
- [ ] https://github.com/airbytehq/airbyte/issues/28480
- [ ] https://github.com/airbytehq/airbyte/issues/28482
- [ ] https://github.com/airbytehq/airbyte/pull/28687
```

## P2 Scope (follow-ons, when time allows)

```[tasklist]
### Tasks
- [ ] Migrate all source connectors to depend on pinned CDK references.
- [ ] Create a CDK v0.2 that tackles some initial refactoring, removal of unused classes, etc.
- [ ] Migrate destination connectors to CDK if they have a dependency on included CDK classes.
- [ ] Tutorials, videos, other content creation efforts.
```

prateekmukhedkar:
@aaronsteers can we create this as an epic?

aaronsteers:
@prateekmukhedkar - Done. Link here: https://app.zenhub.com/workspaces/db--dw-source-connectors-6333360e0a41155061efbcbd/issues/gh/airbytehq/airbyte/29316

prateekmukhedkar:
This work is tracked by the epic https://github.com/airbytehq/airbyte/issues/29316

PR #29219, opened 2023-08-08
Add cdk 0.0.1 dependency to all java connectors

aaronsteers:
> **Note**
> Part of [this epic](https://github.com/airbytehq/airbyte/issues/28445).

⚠️ Note: this should only merge _after_ the related and prerequisite PR:

- #28687 

---------

This PR adds the CDK v0.0.1 dependency to all Java-based connectors.

Methodology:

1. Full text search on repo for text: `implementation project(':airbyte-db:db-lib')`
2. For each result:
    1. Ignore references under the 'bases' directories, to prevent circular dependency later on.
    2. If this is a 'normal' connector: add this text to the top of the 'dependencies' block: `implementation 'io.airbyte:airbyte-cdk:0.0.1'`.
    3. If this is a derrived connector (like the `-strict-connect` connectors), find the primary connector reference and move it to the top of the dependencies block. Add a comment so it is obvious what the primary dependency is, and so it is also clear why the CDK is not pinned directly.
3. Full text search on repo for text: `implementation project(':airbyte-integrations:bases:base-java')`
4. Repeat step 2 for these search results (unless already applied).

In total, there are approximately 85 connectors updated in this PR.


aaronsteers:
Documenting here - is this 'strictness level' expected? Presumably, they were already in this state and the CI test is triggering because I modified the connectors' files?

image

aaronsteers:
Closing. Because many connectors don't pass tests, we're taking a different approach.

PR #29226, opened 2023-08-08
java cdk built with airbyte-ci

c-p-b:
Spikes what it takes to build 

There were some obvious shortcuts taken here, just for the purpose of getting it running. But it does run right now, with `airbyte-ci cdk java build`.  The next step is to use this as a comparison vs calling `aircmd` primitives from airbyte-ci instead, which was @alafanechere 's original suggestion. I think it is possible to do that now that I am more familiar with the codebase - the click entrypoint in`airbyte-ci` is actually pretty vanilla so it might actually be possible to wire in to the existing click entrypoint by modifying the top level airbyte-ci to inject our context objects used in `aircmd`


airbyte-ci/connectors/pipelines/.python-version
```diff
@@ -0,0 +1 @@
+3.9.11
```
cpdeethree, line 1: pyenv accidentally committed this, will delete

airbyte-ci/connectors/pipelines/.tool-versions
```diff
@@ -0,0 +1 @@
+python 3.10
```
cpdeethree, line 1: same, accidental pyenv

airbyte-ci/connectors/pipelines/pipelines/bases.py
```diff
@@ -18,20 +18,30 @@
...
 from rich.console import Group
 from rich.panel import Panel
 from rich.style import Style
 from rich.table import Table
 from rich.text import Text
 from tabulate import tabulate
 
+from pipelines import sentry_utils
+from pipelines.actions import remote_storage
+from pipelines.consts import (
```
cpdeethree, line 33: i haven't run this through black/flake yet so this is just my local machine doing things that don't match your formatting settings

airbyte-ci/connectors/pipelines/pipelines/bases.py
```diff
@@ -44,6 +54,10 @@ class ConnectorWithModifiedFiles(Connector):
     def has_metadata_change(self) -> bool:
         return any(path.name == METADATA_FILE_NAME for path in self.modified_files)
 
+@dataclass(frozen=True)
+class CDKWithModifiedFiles(CDK):
```
cpdeethree, line 58: I wasn't able to hook into this without further study of the code, so modified_files is empty with all instantiations, but I left the stub here to match the existing implementation

airbyte-ci/connectors/pipelines/pipelines/contexts.py
```diff
@@ -476,6 +488,190 @@ async def __aexit__(
...
+
+    @property
+    def secrets_dir(self) -> Directory:  # noqa D102
+        return self._secrets_dir
+
+    @secrets_dir.setter
+    def secrets_dir(self, secrets_dir: Directory):  # noqa D102
+        self._secrets_dir = secrets_dir
+
+    @property
```
cpdeethree, line 573: I am pretty sure this is connector specific so could probably be deleted, but at the time I left it in because I wasn't sure

airbyte-ci/connectors/connector_ops/connector_ops/utils.py
```diff
@@ -325,6 +325,35 @@ def get_local_dependency_paths(self, with_test_dependencies: bool = True) -> Set
...
+
+    language: str
+
+    @property
+    def code_directory(self) -> Path:
+        return Path(f"./airbyte-cdk/{self.language}")
+
+
+    @property
+    def version(self) -> str:
```
alafanechere, line 342: I think it's better to set it as instantiation time instead of using a property.
If you want to keep using a dataclass I recommend using the `__post_init__` hook: https://docs.python.org/3/library/dataclasses.html#post-init-processing

airbyte-ci/connectors/connector_ops/connector_ops/utils.py
```diff
@@ -325,6 +325,35 @@ def get_local_dependency_paths(self, with_test_dependencies: bool = True) -> Set
...
+        if self.language == "java":
+            with open(self.code_directory / "resources/version.properties") as f:
+                for line in f:
+                    key, value = line.strip().split('=', 1)
+                    if key == 'version':
+                        return value
```
alafanechere, line 348: nit: I think this can be implemented in a more elegant way with a ConfigParser:
https://docs.python.org/3/library/configparser.html

airbyte-ci/connectors/pipelines/pipelines/builds/java_connectors.py
```diff
@@ -14,6 +14,39 @@
...
+            )
+        else:
+            return StepResult(
+                self,
+                StepStatus.FAILURE,
+                stderr="The distributions directory contains multiple tar files. We can't infer which one should be used. Please review and delete any unnecessary tar files.",
+            )
+
+
+
```
alafanechere, line 49: I thing this is meant to be deleted as you created `java_cdk.py` (which I think is indeed better)

airbyte-ci/connectors/pipelines/pipelines/contexts.py
alafanechere: Yes this is connector specific, for interaction with GSM. You should not need it.

airbyte-ci/connectors/pipelines/pipelines/contexts.py
```diff
@@ -476,6 +488,190 @@ async def __aexit__(
...
+    @updated_secrets_dir.setter
+    def updated_secrets_dir(self, updated_secrets_dir: Directory):  # noqa D102
+        self._updated_secrets_dir = updated_secrets_dir
+
+    @property
+    def should_save_updated_secrets(self) -> bool:  # noqa D102
+        return self.use_remote_secrets and self.updated_secrets_dir is not None
+
+
+    async def __aexit__(
```
alafanechere, line 586: I don't get why you need to override `__aexit__`

airbyte-ci/connectors/pipelines/pipelines/contexts.py
```diff
@@ -476,6 +488,190 @@ async def __aexit__(
...
+            await asyncify(send_message_to_webhook)(self.create_slack_message(), self.reporting_slack_channel, self.slack_webhook)
+
+        # Supress the exception if any
+        return True
+
+    def create_slack_message(self) -> str:
+        raise NotImplementedError
+    
+
+class PublishCDKContext(CDKContext):
```
alafanechere, line 633: Do you really need to add it in this PR as the publish flow is not implemented?

airbyte-ci/connectors/pipelines/pipelines/gradle.py
```diff
@@ -64,8 +65,8 @@ def _get_gradle_command(self, extra_options: Tuple[str] = ("--no-daemon", "--sca
         command = (
             ["./gradlew"]
             + list(extra_options)
-            + [f":airbyte-integrations:connectors:{self.context.connector.technical_name}:{self.gradle_task_name}"]
-        )
+            + [f":airbyte-integrations:connectors:{self.context.connector.technical_name}:{self.gradle_task_name}"] if hasattr(self.context, "connector") else ["./gradlew", ":airbyte-cdk:java:airbyte-cdk:build"]
+        ) # NOTE: obviously not the ideal way to do this, just wanted to get something that ran
```
alafanechere, line 69: 🙏 please override `_get_gradle_command` in your subclass before merging.

alafanechere (commented):
I think a bit of code cleaning is required, but at a high level the implementation LGTM

airbyte-ci/connectors/connector_ops/connector_ops/utils.py
```diff
@@ -325,6 +325,35 @@ def get_local_dependency_paths(self, with_test_dependencies: bool = True) -> Set
                 self.code_directory / "build.gradle", with_test_dependencies=with_test_dependencies
             )
         return sorted(list(set(dependencies_paths)))
+    
+
+@dataclass(frozen=True)
+class CDK:
+    """Utility class to gather metadata about a cdk."""
+
+    language: str
```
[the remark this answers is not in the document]
cpdeethree: I could probably use this, I guess we could also build low-code sdk with this class too!

airbyte-ci/connectors/connector_ops/connector_ops/utils.py
cpdeethree: 👍

airbyte-ci/connectors/pipelines/pipelines/bases.py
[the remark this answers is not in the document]
cpdeethree: I was just making it look like the connectors one, but we really don't. I can use the CDK class directly, it feels extraneous

airbyte-ci/connectors/pipelines/pipelines/contexts.py
cpdeethree: Probably not, if we decide to move forward with this approach I'll probably remove it to descope the pr

airbyte-ci/connectors/pipelines/pipelines/gradle.py
```diff
@@ -41,8 +42,8 @@ def build_include(self) -> List[str]:
         """
         return [
             str(dependency_directory)
-            for dependency_directory in self.context.connector.get_local_dependency_paths(with_test_dependencies=True)
-        ]
+            for dependency_directory in self.context.connector.get_local_dependency_paths(with_test_dependencies=True) 
+        ] if hasattr(self.context, "connector") else []
```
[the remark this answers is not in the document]
cpdeethree: it wouldn't run without this, I am not sure why without further investigation. This was one of the dirtier hacks I put in to get it to run

airbyte-ci/connectors/pipelines/pipelines/gradle.py
cpdeethree: For sure!

airbyte-ci/connectors/pipelines/pipelines/contexts.py
alafanechere: I think we should first try to make it use the default __aexit__ from its parent class and override if needed

airbyte-ci/connectors/pipelines/pipelines/builds/__init__.py
```diff
@@ -50,9 +52,26 @@ async def run_connector_build_pipeline(context: ConnectorContext, semaphore: any
...
+
+    Returns:
+        None
+    """
+    step_results = []
+    async with Connection(Config(log_output=None, execute_timeout=None)) as dagger_client:
+        context.dagger_client = dagger_client.pipeline("Java CDK Build")
+        build_result = await run_cdk_build(context)
+        step_results.append(build_result)
+    return None
```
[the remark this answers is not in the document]
cpdeethree: oh nice, i had thought that only ConnectorReport existed and not just Report, maybe reporting is not going to be as difficult as I had thought it would be.

postamar:
This has grown stale, closing.

PR #29229, opened 2023-08-08
update state copy info

Hesperide:
Updates state copying instructions based on Lake's mockups.


evantahler (approved):
That's a nice preview!

Issue #28727, opened 2023-07-26
Source Amazon Seller Partner: Failed Adding Source: "nonetype" object has no attribute "split"


Blochergroup:
### Connector Name

 Amazon Seller Partner

### Connector Version

1.3.0

### What step the error happened?

Configuring a new connector

### Revelant information

After authenticating my Amazon Account and clicking "Set up source" I get following Error:
"nonetype" object has no attribute "split".

I tried with and without providing the optional fields (AWS Access Key...).

### Relevant log output

```shell
"nonetype" object has no attribute "split"

2023-06-20 12:46:23 [32mINFO[m i.a.w.t.TemporalAttemptExecution(get):138 - Cloud storage job log path: /workspace/6b7890bb-c7c4-48d0-b029-1a6bae25ae47/0/logs.log
```

### Contribute

- [ ] Yes, I want to contribute

andreyAtBB:
This issue's root cause is https://github.com/airbytehq/airbyte/issues/27495
This one can be closed as duplicate IMHO

Blochergroup:
@andreyAtBB 
so it works if you just specify the Role ARN and none of the other aws credentials?

andreyAtBB:
> 

It requires Role ARN for sure and but I can't check if it works with none of the other AWS creds because we do not use AWS.

marcosmarxm:
Closed by https://github.com/airbytehq/airbyte/pull/29255

Issue #27495, opened 2023-06-20
Source Amazon Seller Partner: error when `role_arn` is not specified


lazebnyi:
### Connector Name

amazon-seller-partner

### Connector Version

1.3.0

### What step the error happened?

Configuring a new connector

## Description:
When using the source Amazon Seller Partner connector, the role_arn parameter is marked as optional. However, if this parameter is not specified, an error occurs when attempting to `split` a `None` value in the `get_sts_credentials` method.

## Expected Behavior:
The `role_arn` parameter should truly be optional, allowing users to omit it if they don't require a specific role. The code should handle the absence of the `role_arn` gracefully, without causing an error.

## Current Behavior:
If the `role_arn` parameter is not provided, an error occurs during the execution of the `get_sts_credentials` method due to an attempt to split a None value. This results in an unexpected error message and prevents the proper functioning of the Amazon Seller Partner connector.

### Steps to Reproduce:
1. Try to setup the Amazon Seller Partner connector.
2. Do not provide a value for the `role_arn` parameter.
3. Trigger an check connection.

### Contribute

- [ ] Yes, I want to contribute

andreyAtBB:
JFYI: Issue that duplicates this one https://github.com/airbytehq/airbyte/issues/28727

andreyAtBB:
This issue is fixed

marcosmarxm:
Closed by https://github.com/airbytehq/airbyte/pull/29255

PR #29255, opened 2023-08-09
🐛 Source Amazon Seller Partner: attempt to fix an issue with role_arn required with authorization is done via oauth

andreyAtBB:
## What
When user creates Amazon Seller Partner connector source on Airbyte Cloud, authenticate via Oauth and save and error appears as described in these issues:
https://github.com/airbytehq/airbyte/issues/28727
https://github.com/airbytehq/airbyte/issues/27495

## How
If role_arn is not provided (which is the case for Oauth authentication) do not attempt to split it and simply proceed.

## Recommended reading order
n/a

## 🚨 User Impact 🚨
Should not have any impact but should allow user to proceed one step further during connector set up.

## Pre-merge Actions
None


rileybrook:
@lazebnyi Hoping this small PR can be shepherded through as it blocks an Airbtye Cloud customer from onboarding new clients.

airbyte-integrations/connectors/source-amazon-seller-partner/metadata.yaml
```diff
@@ -2,7 +2,7 @@ data:
   connectorSubtype: api
   connectorType: source
   definitionId: e55879a8-0ef8-4557-abcf-ab34c53ec460
-  dockerImageTag: 1.5.0
+  dockerImageTag: 1.5.1
```
marcosmarxm, line 5: Can you bump also the doc changelog?
After that I think it's fine to merge. Right now the `role arn` is optional and if you don't specify it will break.

marcosmarxm (changes requested):

airbyte-integrations/connectors/source-amazon-seller-partner/metadata.yaml
andreyAtBB: done. could you take a look.

marcosmarxm (approved):

andreyAtBB:
@marcosmarxm, there are some workflows awaiting for approval from a maintainer blocking the ability to merge the fix in.
Is this something you can look at?

lazebnyi (approved):

marcosmarxm:
@andreyAtBB I'm running tests right now to release the change.

PR #29295, opened 2023-08-09
postgres docs update & improvements

Hesperide:
Our postgres docs are really hard to read:
* They are too long
* There's no digest to help users know what to expect
* It mixes setup with troubleshooting throughout. This means you can't read them from top to bottom (there are a lot of tangents not relevant to most users)
* It doesn't get customers excited.

This PR proposes changes which makes our Postgres documentation more than 2x shorter, and shortens the 'quick start' all customers must read to be 5x shorter than today.

To do this, I:
* Moved limitations, and advanced troubleshooting steps into a separate page. This is necessary given the complexity of this connector.
* Created a quick start guide using xmin (minimum path to value).
* Brought together all the CDC instructions into 1 section.
* Added an overview at the top to get users excited.
* Deleted a handful of oudated content (e.g. column selection not being supported)


docs/integrations/sources/postgres.md
```diff
@@ -1,221 +1,151 @@
...
 
-## When to use Postgres with CDC
-
-Configure Postgres with CDC if:
+# Postgres
 
-- You need a record of deletions
-- Your table has a primary key but doesn't have a reasonable cursor field for incremental syncing (`updated_at`). CDC allows you to sync your table incrementally
+Airbyte's certified Postgres connector offers the following features:
+* Multiple methods of keeping your data fresh, including [Change Data Capture (CDC)](./../../understanding-airbyte/cdc.md) and replication using the [xmin system column](#xmin-replication-mode).
```
prateekmukhedkar, line 7: Do you want to add recommended mode is CDC if syncing tables in the terabyte range?

Hesperide: That's outlined multiple times further down in the doc. I think this section should be more on selling the dream than starting with limitations and pointing customers down certain paths.

In the quick start section, when users configure xmin, I refer them to CDC if their tables are above a certain size.

prateekmukhedkar (approved):

Issue #29215, opened 2023-08-08
File-based CSV: Strings should not be null'd by default


girarda:
## What
The CSV parser converts values to `None` if they match any of the values in the null_values set, which is configurable as part of the CsvFormat.

To be backward compatible with the legacy S3 source, string values should be kept as-is by default, even if they match one of the values in the null_values set.

We should add an option to the CsvFormat to make the strings nullable

## Acceptance Criteria
- [ ] By default, strings are not turned to `None` even if they match a null_values
- [ ] The CsvFormat has a boolean field "strings_can_be_null". When True, strings matching a null_value are None

girarda:
Update: we'll call the field `strings_can_be_null` instead of `strings_can_be_none`

PR #29298, opened 2023-08-09
file-based CDK: Configurable strings_can_be_null

girarda:
## What
* This PR address the first part of https://github.com/airbytehq/airbyte/issues/29215 by introducing the option to make strings nullable
* The default behavior is for strings to be nullable. The S3Adapter will handle backwards compatibility separately.

## How
When converting values to null:
1. If the value is not in the null_values set -> value is not None
2. If the value is in the null_values set AND its type is not "string" -> value is None
3. If the value is in the null_values set AND its type is "string" and strings_can_be_null" is True -> value is None
4.  If the value is in the null_values set AND its type is "string" and strings_can_be_null" is False -> value is not None

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/file_based/config/csv_format.py`
5. `airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/csv_parser.py`


airbyte-cdk/python/unit_tests/sources/file_based/scenarios/csv_scenarios.py
```diff
@@ -43,6 +43,18 @@
                 "description": "Used during spec; allows the developer to configure the cloud provider specific options\nthat are needed when users configure a file-based source.",
                 "type": "object",
                 "properties": {
+                    "start_date": {
```
girarda, line 46: not 100% sure why this is being re-ordered, but properties is a map so this shouldn't matter

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/csv_format.py
```diff
@@ -64,6 +64,11 @@ class Config:
         default=[],
         description="A set of case-sensitive strings that should be interpreted as null values. For example, if the value 'NA' should be interpreted as null, enter 'NA' in this field.",
     )
+    strings_can_be_null: bool = Field(
```
clnoll, line 67: Should we have some validation on this that requires `null_values` to be set if `strings_can_be_none` is True?

clnoll (approved):
Looks great @girarda! One small question.

airbyte-cdk/python/airbyte_cdk/sources/file_based/config/csv_format.py
girarda: I don't think we want that validation. Users should be able to configure `null_values` even if `strings_can_be_none` is False to set non-string fields to `None`

clnoll: Oh right, makes sense!

PR #29313, opened 2023-08-09
Reuse existing open PR if it exists when bumping platform CDK dependency

sherifnada:
We end up with many stale PRs on the platform because of how frequently we release the CDK. This PR changes this workflow to use the same branch. This has the effect of updating an open version bump PR on the platform if it already exists, or creating a new PR if it doesn't exist


maxi297 (approved):
:shipit:

Issue #26764, opened 2023-05-30
File CDK: Implement JSONL-specific options


clnoll:
Support the "Allow newlines in values" (T/F) option, which is supported in the existing `source_files_abstract`.

The existing [`source_files_abstract` JSONL parser](https://github.com/airbytehq/airbyte/blob/776f650e9cc8abe25206e96de861c8202fd6fed6/airbyte-integrations/connectors/source-s3/source_s3/source_files_abstract/formats/jsonl_parser.py#L67) should be helpful, as it includes this option.

For details on the adoption of this option see the [appendix of the File-Based CDK Proposal](https://docs.google.com/document/d/1f9rlYK27onB2dqjS40YKXdKbCytG1Xh0q_mMUP-_Ceg/edit#heading=h.yoaiqhjxe8te).

#### Acceptance Criteria
- [ ] Thorough test coverage.
- [ ] User input is validated during the connection check. Bad input values are surfaced in appropriate error messages.

maxi297:
Follow-up
* Specifying newlines_in_values does not change the behavior i.e. having this True or False still allows for multiline JSON. It means that we would not only have to confirm if actor where configuration->'format'->>'newlines_in_values' = 'true' but all the actors with JSONL files
* The drawback block_size issues using pyarrow seems too annoying for now as it broke stream before
* We agreed on supporting multiline jsonl by brute force in the CDK and adding log to see if users actually have this. We will add a reminder to search for this log after the release. Unfortunately, we will only know this once V4 is installed though

PR #29331, opened 2023-08-10
[ISSUE #26764] support brute force multiline json objects for JSONL

maxi297:
## What
Adresses https://github.com/airbytehq/airbyte/issues/26764

As mentioned in the ticket:
* Specifying `newlines_in_values` does not change the behavior i.e. having this True or False still allows for multiline JSON and therefore, all JSONL files synced in production could have been considered "newlines_in_values"
* The drawback `block_size` issues using pyarrow seems too annoying for now as it broke stream before
* We agreed on supporting multiline jsonl by brute force in the CDK and adding log to see if users actually have this. We will add a reminder to search for this log after the release. Unfortunately, we will only know this once V4 is installed though. I've created this reminder on August 21st. Ping me if you want to be invited to the event

## How
If a line is not parsable, accumulate the content of this line and parse using the following line. Rince and repeat until JSON parsing works

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/jsonl_parser.py
+++ b/airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/jsonl_parser.py
@@ -7,6 +7,7 @@ import logging
 from typing import Any, Dict, Iterable
 
 from airbyte_cdk.sources.file_based.config.file_based_stream_config import FileBasedStreamConfig
+from airbyte_cdk.sources.file_based.exceptions import FileBasedSourceError, RecordParseError
 from airbyte_cdk.sources.file_based.file_based_stream_reader import AbstractFileBasedStreamReader, FileReadMode
 from airbyte_cdk.sources.file_based.file_types.file_type_parser import FileTypeParser
 from airbyte_cdk.sources.file_based.remote_file import RemoteFile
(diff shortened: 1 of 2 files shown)


airbyte-cdk/python/unit_tests/sources/file_based/file_types/test_jsonl_parser.py
```diff
@@ -33,3 +35,66 @@ def test_type_mapping(record: Dict[str, Any], expected_schema: Mapping[str, str]
...
+def test_given_one_json_per_line_when_parse_records_then_return_records() -> None:
+    stream_reader = MagicMock()
+    stream_reader.open_file.return_value.__enter__.return_value = JSONL_CONTENT_WITHOUT_MULTILINE_JSON_OBJECTS
+
+    records = list(JsonlParser().parse_records(Mock(), Mock(), stream_reader, Mock()))
+
+    assert records == [{"a": 1, "b": "1"}, {"a": 2, "b": "2"}]
+
+
+def test_given_multiline_json_object_when_parse_records_then_return_records() -> None:
```
girarda, line 74: this test should verify we don't log a warning

airbyte-cdk/python/airbyte_cdk/sources/file_based/file_types/jsonl_parser.py
```diff
@@ -16,6 +17,7 @@
 class JsonlParser(FileTypeParser):
 
     MAX_BYTES_PER_FILE_FOR_SCHEMA_INFERENCE = 1_000_000
+    _WITH_READ_LIMIT = True
```
brianjlai, line 20: Do you plan on using this constant more? It feels unnecessary if we're only using it in one place when infer_schema calls `infer_jsonl_entries()`

airbyte-cdk/python/unit_tests/sources/file_based/file_types/test_jsonl_parser.py
maxi297: Did you mean that we should log a warning? `test_given_multiline_json_object_when_parse_records_then_log_once_one_record_yielded` validates that. I can check that `test_given_one_json_per_line_when_parse_records_then_return_records` does not log a warning for extra security

girarda: I think I derped and commented on the wrong line. +1 for verifying we do not log a warning when only parsing single line records

girarda (approved):
:shipit:

PR #29339, opened 2023-08-10
File-CDK (Avro) Set double_as_string to false by default

girarda:
## What
* Set `double_as_string` to False by default because it is more intuitive


brianjlai (approved):
lgtm!

PR #29416, opened 2023-08-14
Docs: update Google search console docs

ChristoGrab:
## What
Updated the in-app and main docs pages for GSC source connector

## How
- Clarified use cases for authentication options. The current instructions were a bit murky particularly in regards to the nature of delegating domain-wide authority to a service account. My understanding is this step is meant for specific use cases where the service account will be accessing data from multiple users across a Google Workspace domain, and therefore needs to be granted 'carte blanche' access by a super admin. I've tried to make it clearer that this is a case-specific requirement for service accounts rather than a separate authentication option. Let me know if this still needs to be clarified further, or if I'm misunderstanding the nature of this functionality!
- Updated connector setup instructions for both pages
- Added separate sections for Custom Reports and Data State with additional context/examples. Setup steps link to these sections where appropriate

### DIFF
--- a/docs/integrations/sources/google-search-console.inapp.md
+++ b/docs/integrations/sources/google-search-console.inapp.md
@@ -1,17 +1,24 @@
 ## Prerequisite
 
-* A verified property in Google Search Console
-* Credentials to a Google Service Account (or Google Service Account with delegated Domain Wide Authority) or Google User Account
-* Enable  Google Search Console API
+- A verified property in Google Search Console
+- Enabled Google Search Console API
 ​
 ## Setup guide
-1. Enter the site URL.
-2. Enter a **Start Date** in the format YYYY-MM-DD. Data after this date will be replicated.
-2. Click **Authenticate your account** by selecting Oauth (recommended).
-   * If you prefer Service Account Key Authentication, follow the instructions in our [full documentation](https://docs.airbyte.com/integrations/sources/google-search-console/).
-3. (Optional) Set an **End Date** if you do not want data after a specific date.
-4. (Optional) Airbyte generates default reports. To add more reports, you need to add **Custom Reports** as a JSON array describing the custom reports you want to sync from Google Search Console. 
+1. For **Source name**, enter a name to help you identify this source.
+2. For **Website URL Property**, enter the specific website property in Google Seach Console with data you want to replicate.
+3. For **Start Date**, use the provided datepicker or enter a date in the format `YYYY-MM-DD`. Any data created on or after this date will be replicated.
+4. To authenticate the connection:
 
-Custom Reports can be added in the format: `{"name": "<report-name>", "dimensions": ["<dimension-name>", ...]}`
+   <!-- env:cloud -->
+   - **For Airbyte Cloud**: Select **Oauth** from the Authentication dropdown, then click **Sign in with Google** to authorize your account.
+   <!-- /env:cloud -->
+   <!-- env:oss -->
+   - **For Airbyte Open Source**: Select **Service Account Key Authorization** from the Authentication dropdown, then enter the **Admin Email** and **Service Account JSON Key**. For the key, copy and paste the JSON key you obtained during the service account setup. It should begin with `{"type": "service account", "project_id": YOUR_PROJECT_ID, "private_key_id": YOUR_PRIVATE_KEY, ...}`
+   <!-- /env:oss -->
+
+5. (Optional) For **End Date**, you may optionally provide a date in the format `YYYY-MM-DD`. Any data created between the defined Start Date and End Date will be replicated. Leaving this field blank will replicate all data created on or after the Start Date to the present.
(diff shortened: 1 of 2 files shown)


docs/integrations/sources/google-search-console.inapp.md
```diff
@@ -1,17 +1,25 @@
 ## Prerequisite
 
-* A verified property in Google Search Console
-* Credentials to a Google Service Account (or Google Service Account with delegated Domain Wide Authority) or Google User Account
-* Enable  Google Search Console API
-​
+- A verified property in Google Search Console
+- Google Search Console API enabled for your project
```
sherifnada, line 4: this is only true for OSS

docs/integrations/sources/google-search-console.inapp.md
```diff
@@ -1,17 +1,25 @@
...
+4. To authenticate the connection:
+
+   <!-- env:cloud -->
+   - **For Airbyte Cloud**: Select **Oauth** from the Authentication dropdown, then click **Sign in with Google** to authorize your account. More information on authentication methods can be found in our [full Google Search Console documentation](https://docs.airbyte.io/integrations/sources/google-search-console#setup-guide).
+   <!-- /env:cloud -->
+   <!-- env:oss -->
+   - **For Airbyte Open Source**: Select **Service Account Key Authorization** from the Authentication dropdown, then enter the **Admin Email** and **Service Account JSON Key**. For the key, copy and paste the JSON key you obtained during the service account setup. It should begin with `{"type": "service account", "project_id": YOUR_PROJECT_ID, "private_key_id": YOUR_PRIVATE_KEY, ...}`. More information on setting up a service account and generating the JSON key can be found in our [full Google Search Console documentation](https://docs.airbyte.io/integrations/sources/google-search-console#setup-guide).
```
sherifnada, line 17: for reference they can still use oauth in oss but they need to create their own oauth app. IDK if we need to list out the setup process for them, but let's at least mention that

docs/integrations/sources/google-search-console.md
```diff
@@ -1,95 +1,83 @@
...
-- Use your personal Google User Account with oauth. If you choose this option, your account must have permissions to view the Google Search Console project you choose.
-- Create the new `Service Account` credentials for your Google Project, and grant Admin Permissions to it
-- Follow the `Delegating domain-wide authority` process to obtain the necessary permissions to your google account from the administrator of Workspace
+You can authenticate using your Google Account with OAuth if you are the owner of the Google Search Console property or have view permissions. Follow [Google's instructions](https://support.google.com/webmasters/answer/7687615?sjid=11103698321670173176-NA) to ensure that your account has the necessary permissions (**Owner** or **Full User**) to view the Google Search Console property. This option is recommended for **Airbyte Cloud** users, as it significantly simplifies the setup process and allows you to authenticate the connection [directly from the Airbyte UI](#step-2-set-up-the-google-search-console-connector-in-airbyte).
 
-### Creating a Google service account
+#### II: Google service account with JSON key file (Recommended for Airbyte Open Source)
```
sherifnada, line 20: can we hide this for cloud?

docs/integrations/sources/google-search-console.md
```diff
@@ -121,6 +105,52 @@ Parameter `data_state='all'` should not be used with Incremental Sync mode as it
 - [Analytics site report by site](https://developers.google.com/webmaster-tools/search-console-api-original/v3/searchanalytics/query)
 - Analytics report by custom dimensions
 
+## Connector-specific configurations
+
+### Custom reports
```
sherifnada, line 110: nice addition!

docs/integrations/sources/google-search-console.md
```diff
@@ -121,6 +105,52 @@ Parameter `data_state='all'` should not be used with Incremental Sync mode as it
...
+   {
+      "name": "country_date",
+      "dimensions": ["country", "date"]
+   }
+]
+```
+
+You can use the [Google APIS Explorer](https://developers.google.com/webmaster-tools/v1/searchanalytics/query) to build and test the reports you want to use.
+
+### Data State
```
sherifnada, line 143: should this option and section be called data freshness instead?

sherifnada: great addition on this section btw, nice work!

sherifnada (approved):

docs/integrations/sources/google-search-console.inapp.md
ChristoGrab: Good call, I added the option to both docs pages and a link to Google's setup guide for creating the auth app

docs/integrations/sources/google-search-console.md
ChristoGrab: I've gone ahead and added the tags just in case, but I don't think they have any effect on the main page when an in-app page exists, since the latter is the one that gets displayed in the Airbyte UI

docs/integrations/sources/google-search-console.md
ChristoGrab: Thanks! Yeah I think data freshness makes sense, I've made the changes here and added the new title to the spec PR

Issue #27019, opened 2023-06-05
File CDK: user- and support-facing documentation


clnoll:
Document the options and behavior of file-based sources that use the file-based CDK, as described in the [tech spec](https://docs.google.com/document/d/12nDEOAGCoXG0LAyiBe2fCx9CZfSfGNlHGr5Xrphk4qE/edit#).

The goal is to leave no ambiguity/undefined behavior.

Note: this [PR](https://github.com/airbytehq/airbyte/pull/28491) identifies a number of potentially breaking changes. Mentioning here so we can flag them in the docs and for Support. They'll be reviewed when we test backwards compatibility.

#### Acceptance Criteria
- [ ] Documentation gives users and support a clear understanding of what is and isn't offered by sources that use the file-based CDK module.

clnoll:
Grooming notes:
This is customer-facing documentation that will replace the existing source-S3 connector documentation. 

Eventually we'll want shared file-based CDK documentation that all file-based sources can point to.

PR #29434, opened 2023-08-15
Issue 27019/doc

maxi297:
## What
Addresses https://github.com/airbytehq/airbyte/issues/27019

## How
Adding documentation, more specifically:
* Airbyte protocol level behavior
* How to implement custom file-based source
* Supported file types and options
* Schema + validation policies
* Breaking changes


airbyte-cdk/python/airbyte_cdk/sources/file_based/README.md
```diff
@@ -1,3 +1,103 @@
...
+* Skip nonconforming records: check each record to see if it conforms to the schema; keep a count of the number of records in each file that do and do not conform and emit a log message with these counts once we’re done reading the file. Do not emit records that don’t conform.
+* Emit all records: columns that don’t exist in the configured catalog may still be available in the destination’s raw tables, but this is undefined behavior.
+Only error if there are conflicting field types or malformed rows.
+* Stop the sync and wait for schema re-discovery:  if a record is encountered that does not conform to the configured catalog’s schema, we log a message and stop the whole sync.
+
+When the `schemaless` is enabled, validation will be skipped.
+
+## Breaking Changes (compared to previous S3 implementation)
+
+TODO * Different behavior if input schema maps to an object or array
```
maxi297, line 96: Still working on reproducing this one. Also not that I have not documented the breaking changes for `auto_dict_encode` and `timestamp_parsers` as I was not able to reproduce

airbyte-cdk/python/airbyte_cdk/sources/file_based/README.md
```diff
@@ -10,11 +110,11 @@ The file-based connectors supports the following [sync modes](https://docs.airby
 | Replicate Multiple Streams \(distinct tables\) | Yes        |
 | Namespaces                                     | No         |
 
-We recommend you do not 
+We recommend you do not (do not what? the suspense is unbearable!)
```
girarda, line 113:
```suggestion
We recommend you do not manually modify files that are already synced. The connector has file-level granularity, which means adding or modifying a row in a CSV file will trigger a re-sync of the content of that file.
```

airbyte-cdk/python/airbyte_cdk/sources/file_based/README.md
girarda: you can use these two commands after pulling the source-s3 secrets:

1. read: `python main.py read --config secrets/legacy_csv_user_schema_cast_complex.json --catalog integration_tests/configured_catalogs/csv.json`
2. `python main.py discover --config secrets/legacy_csv_user_schema_cast_complex.json`

What's interesting is this particular config doesn't work in legacy s3 (unless the destination is able to cast the strings to objects) because the `dict` is discovered as an object, but the it's a string in the record

airbyte-cdk/python/airbyte_cdk/sources/file_based/README.md
```diff
@@ -1,3 +1,103 @@
...
+
+## Schema
+
+Having a schema allows for the file-based CDK to take action when there is a discrepancy between a record and what are the expected types of the record fields.
+
+Schema can be either inferred or user provided.
+* If the user defines it a format using JSON types, inference will not apply. Input schemas are a key/value pair of strings describing column name and data type. Supported types are `["string", "number", "integer", "object", "array", "boolean", "null"]`. For example, `{"col1": "string", "col2": "boolean"}`.
+* If the user enables schemaless sync, schema will `{"data": "object"}` and therefore emitted records will look like `{"data": {"col1": val1, …}}`.
+* Else, the file-based CDK will infer the schema depending on the file type. Some file formats defined the schema as part of their metadata (like Parquet), some do on the record-level (like Avro) and some don't have any explicit typing (like JSON or CSV). Any file format that does not define their schema on a metadata level will require the file-based CDK to iterate to a number of records. There is a limit of bytes that will be consumed in order to infer the schema.
```
clnoll, line 82:
```suggestion
* Else, the file-based CDK will infer the schema depending on the file type. Some file formats defined the schema as part of their metadata (like Parquet), some do on the record-level (like Avro) and some don't have any explicit typing (like JSON or CSV). Note that all CSV values are inferred as strings except where we are supporting legacy configurations. Any file format that does not define their schema on a metadata level will require the file-based CDK to iterate to a number of records. There is a limit of bytes that will be consumed in order to infer the schema.
```

airbyte-cdk/python/airbyte_cdk/sources/file_based/README.md
```diff
@@ -1,3 +1,103 @@
...
+
+### Validation Policies
+Users will be required to select one of 3 different options, in the event that records are encountered that don’t conform to the schema.
+
+* Skip nonconforming records: check each record to see if it conforms to the schema; keep a count of the number of records in each file that do and do not conform and emit a log message with these counts once we’re done reading the file. Do not emit records that don’t conform.
+* Emit all records: columns that don’t exist in the configured catalog may still be available in the destination’s raw tables, but this is undefined behavior.
```
clnoll, line 88:
```suggestion
* Emit all records: emit all records, even if they do not conform to the user-provided or inferred schema. Columns that don’t exist in the configured catalog may still be available in the destination’s raw tables, but this is undefined behavior.
```

airbyte-cdk/python/airbyte_cdk/sources/file_based/README.md
```diff
@@ -1,3 +1,103 @@
...
+* If the user enables schemaless sync, schema will `{"data": "object"}` and therefore emitted records will look like `{"data": {"col1": val1, …}}`.
+* Else, the file-based CDK will infer the schema depending on the file type. Some file formats defined the schema as part of their metadata (like Parquet), some do on the record-level (like Avro) and some don't have any explicit typing (like JSON or CSV). Any file format that does not define their schema on a metadata level will require the file-based CDK to iterate to a number of records. There is a limit of bytes that will be consumed in order to infer the schema.
+
+### Validation Policies
+Users will be required to select one of 3 different options, in the event that records are encountered that don’t conform to the schema.
+
+* Skip nonconforming records: check each record to see if it conforms to the schema; keep a count of the number of records in each file that do and do not conform and emit a log message with these counts once we’re done reading the file. Do not emit records that don’t conform.
```
clnoll, line 87:
```suggestion
* Skip nonconforming records: check each record to see if it conforms to the user-input or inferred schema; skip the record if it doesn't conform. We keep a count of the number of records in each file that do and do not conform and emit a log message with these counts once we’re done reading the file.
```

airbyte-cdk/python/airbyte_cdk/sources/file_based/README.md
```diff
@@ -1,3 +1,103 @@
...
+* Emit all records: columns that don’t exist in the configured catalog may still be available in the destination’s raw tables, but this is undefined behavior.
+Only error if there are conflicting field types or malformed rows.
+* Stop the sync and wait for schema re-discovery:  if a record is encountered that does not conform to the configured catalog’s schema, we log a message and stop the whole sync.
```
clnoll, line 90:
```suggestion
* Stop the sync and wait for schema re-discovery:  if a record is encountered that does not conform to the configured catalog’s schema, we log a message and stop the whole sync. Note: this option is not recommended if the files have very different columns or datatypes, because the inferred schema may vary significantly at discover time.
```

airbyte-cdk/python/airbyte_cdk/sources/file_based/README.md
```diff
@@ -1,3 +1,103 @@
...
+Parquet is a file format defined by [Apache](https://parquet.apache.org/). Configuration options are:
+* `decimal_as_float`: Whether to convert decimal fields to floats. There is a loss of precision when converting decimals to floats, so this is not recommended.
+
+## Schema
+
+Having a schema allows for the file-based CDK to take action when there is a discrepancy between a record and what are the expected types of the record fields.
+
+Schema can be either inferred or user provided.
+* If the user defines it a format using JSON types, inference will not apply. Input schemas are a key/value pair of strings describing column name and data type. Supported types are `["string", "number", "integer", "object", "array", "boolean", "null"]`. For example, `{"col1": "string", "col2": "boolean"}`.
+* If the user enables schemaless sync, schema will `{"data": "object"}` and therefore emitted records will look like `{"data": {"col1": val1, …}}`.
```
clnoll, line 81:
```suggestion
* If the user enables schemaless sync, schema will `{"data": "object"}` and therefore emitted records will look like `{"data": {"col1": val1, …}}`. This is recommended if the contents between files in the stream vary significantly, and/or if data is very nested.
```

clnoll (approved):
Just a few suggestions but overall LGTM

airbyte-cdk/python/airbyte_cdk/sources/file_based/README.md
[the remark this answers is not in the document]
sherifnada: just don't

airbyte-cdk/python/airbyte_cdk/sources/file_based/README.md
girarda: nit: we should say that columns that don't exist in the configured catalog probably won't be available in the destination's table since that's the current behavior. (I agree it should still be considered undefined behavior)

girarda (approved):

PR #29449, opened 2023-08-15
Update high strictness to only be based off quality level

bnchrch:
# Overview
We are not ready to enable high strictness on all critical as many are alpha


octavia-approvington:

octavia-approvington (approved):

PR #29477, opened 2023-08-16
tech spec: demo of common base image for all python connectors

alafanechere:
## What
Suggest  an approach at declaring a single, versioned, base image for all python connectors.

[CF tech spec](https://docs.google.com/document/d/19XTGfqsHFnEc_LHlalUwERLw6n8JDYOnflDdPiex0OM/edit#heading=h.q0xnouw6i6xu)

### DIFF
--- a/airbyte-ci/connectors/pipelines/pipelines/builds/base_images/python.py
+++ b/airbyte-ci/connectors/pipelines/pipelines/builds/base_images/python.py
@@ -124,7 +124,7 @@ def write_changelog_file():
     markdown = markdown_table(entries).set_params(row_sep="markdown", quote=False).get_markdown()
     with open("PYTHON_BASE_IMAGES_CHANGELOG.md", "w") as f:
         f.write(
-            "Python base images used for connector built are declared [here](https://github.com/airbytehq/airbyte/blob/8328c9dd89f6417295a20f7c0f5b823a2f02ee8e/airbyte-ci/connectors/pipelines/pipelines/builds/base_images/python.py)"
+            "Python base images used for connector built are declared [here](https://github.com/airbytehq/airbyte/blob/8328c9dd89f6417295a20f7c0f5b823a2f02ee8e/airbyte-ci/connectors/pipelines/pipelines/builds/base_images/python.py)\n\n"
         )
         f.write(f"# Changelog for {AirbytePythonBase.name}\n\n")
         f.write(markdown)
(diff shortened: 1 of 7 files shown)


alafanechere:
I successfully built a connector on this branch:
`airbyte-ci connectors --name=source-openweather build`

airbyte-ci/connectors/pipelines/pipelines/builds/base_images/python.py
```diff
@@ -0,0 +1,133 @@
...
+    def container(self) -> dagger.Container:
+        return (
+            self.base.with_exec(["ln", "-snf", f"/usr/share/zoneinfo/{self.TIMEZONE}", "/etc/localtime"])
+            .with_exec(["apt-get", "update"])
+            .with_exec(["apt-get", "install", "-y", *self.apt_packages])
+            .with_exec(["pip", "install", "--upgrade", "pip"])
+        )
+
+
+class _0_0_2(_0_0_1):
```
alafanechere, line 92: I'm showcasing here how inheritance can help in patching a previous version with maximum code reuse.
Nothing in the pattern I implemented forces a new version to inherit from the previous one. In other words: the `container` property can be re-implemented from scratch if we don't want build on top of the previous version.

airbyte-ci/connectors/pipelines/pipelines/builds/base_images/python.py
```diff
@@ -0,0 +1,133 @@
...
+        return super().with_exec(["apt-get", "install", "-y", "git"])
+
+
+class _1_0_0(_0_0_2):
+    python_base_image = PythonBase.PYTHON_3_10_12
+
+    changelog = "Upgrades the base image to Python 3.10.12."
+
+
+def get_all_base_images() -> dict[str, Type[AirbytePythonBase]]:
```
alafanechere, line 109: This function is "just" sugar to automatically discover the image version declared in the module. It saves us from hardcoding the list of base images version: implementing a new class should be the only step to make a new base version available.

airbyte-ci/connectors/pipelines/pipelines/builds/python_connectors.py
```diff
@@ -15,13 +19,102 @@ class BuildConnectorImage(BuildConnectorImageBase):
...
+        connector_container = (
+            with_connector_code.with_env_variable("AIRBYTE_ENTRYPOINT", " ".join(self.entrypoint))
+            .with_entrypoint(self.entrypoint)
+            .with_label("io.airbyte.version", self.context.metadata["dockerImageTag"])
+            .with_label("io.airbyte.name", self.context.metadata["dockerRepository"])
+        )
+        return await self.finalize_build(self.context, connector_container)
+
+    @staticmethod
+    async def finalize_build(context: ConnectorContext, connector_container: Container) -> Container:
```
alafanechere, line 80: This code is currently used in the `environment` module: https://github.com/airbytehq/airbyte/blob/master/airbyte-ci/connectors/pipelines/pipelines/actions/environments.py#L893

PR #29488, opened 2023-08-16
Removing validation on skip rows and autogenerated headers

maxi297:
## What
Removing validation as it was generating a breaking change

It fixes the CATs `test_check` for all the configs and therefore, I assume it'll fix the other issues encountered during other tests


airbyte-cdk/python/airbyte_cdk/sources/file_based/config/csv_format.py
```diff
@@ -113,11 +113,3 @@ def validate_encoding(cls, v: str) -> str:
         except LookupError:
             raise ValueError(f"invalid encoding format: {v}")
         return v
-
-    @root_validator
-    def validate_option_combinations(cls, values: Mapping[str, Any]) -> Mapping[str, Any]:
```
brianjlai, line 118: just noting some build formatting failures since these imports aren't used anymore

brianjlai (commented):
I think we'll also need to fix the tests because in `test_csv_format.py` we have a test that checks that an exception thrown if skip_rows_before_header is 1 and autogenerate_column_names = true

maxi297:
Yes, there is more to do as well since jsonl parsing does not work given a StringIO. I have other CATs failing locally as well. I'll integrate all those changes in this PR

clnoll (approved):
LGTM!

Issue #29398, opened 2023-08-14
Add more datetime format options (or ability to perform operations) to CDK cursor config


davidfromtandym:
### What area the feature impact?

None

### Revelant Information

The CDK allows for flexibility in defining the format for outgoing cursor datetimes. However, there are still some limitations. For example, [this API](https://docs.rutter.com/rest/2023-02-07/timestamps) requires timestamps formatted as a UNIX timestamp in milliseconds.

The feature request is to enable additional timestamp formats, specifically related to UNIX timestamps, or to allow simple operations. For example, it should be possible to transform a UNIX seconds timestamp into a milliseconds timestamp by using `%s * 1000`.

![image](https://github.com/airbytehq/airbyte/assets/99083780/8f74a0c4-f492-404e-b08f-929ff7594cf4)

girarda:
@davidfromtandym you can use `%ms` to format the time as milliseconds since the unix epoch if you upgrade your platform to `v0.50.19`

girarda:
closing because ms are available. Feel free to open a new issue if you need more formats

PR #29504, opened 2023-08-16
low-code: Allow formatting datetimes as milliseconds since unix epoch

girarda:
## What
* Add a directive "%ms" to parse and format timestamps as unix milliseconds
* [Request to unblock](https://airbytehq-team.slack.com/archives/C02TL38U5L7/p1692221950726739) @johnwasserman and a subset of https://github.com/airbytehq/airbyte/issues/29398

## How
* Update the datetimeparser with a custom directive
* Update the DatetimeFormatInferrer to also accept valid values for ms (`range(1_000_000_000_000, 2_000_000_000_000)`)
 * I'm a little on the fence about recommending these, but unix timestamps in milliseconds are a valid format so it seems fair to recommend it 

## Recommended reading order
1. `airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py`
2. `airbyte-cdk/python/airbyte_cdk/utils/datetime_format_inferrer.py`
3. `airbyte-cdk/python/airbyte_cdk/sources/declarative/declarative_component_schema.yaml`

## 🚨 User Impact 🚨
- Low-code connectors can user cursor values formatted as milliseconds since the unix epoch


airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py
```diff
@@ -25,6 +27,8 @@ def parse(self, date: Union[str, int], format: str):
         # See https://stackoverflow.com/a/4974930
         if format == "%s":
             return datetime.datetime.fromtimestamp(int(date), tz=datetime.timezone.utc)
+        elif format == "%ms":
+            return self._UNIX_ZERO_DT + datetime.timedelta(milliseconds=int(date))
```
girarda, line 31: we can't use datetime.fromtimestamp like we do for `%s`. Instead we can add a timedelta defined in milliseconds.

johnwasserman (approved):
Approving the justification here, can't review any of the actual code

airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py
```diff
@@ -37,6 +41,9 @@ def format(self, dt: datetime.datetime, format: str) -> str:
         # See https://stackoverflow.com/a/4974930
         if format == "%s":
             return str(int(dt.timestamp()))
+        if format == "%ms":
+            # Best we can do is second precision and mulitply by 1000
+            return str(int(dt.timestamp() * 1000))
```
brianjlai, line 46: is a datetime object simply not expressive enough to retain millisecond value? looking at `timestamp()` sounds like it only returns seconds. But it just feels like a slight gotcha (although probably trivial) that we can receive milli timestamps from an API response, but end up truncating it when we save state. 

Just thinking aloud if we need to call this out in docs or something

brianjlai (approved):
The code is pretty straightforward and the use case is clear so no need to block. The losing milli precision on format does give me a little bit of pause, but I also don't see other options if the datetime obj can't express it right

airbyte-cdk/python/airbyte_cdk/sources/declarative/datetime/datetime_parser.py
girarda: good question. Looks like my comment is wrong. timstamp does give ms precision, but the value is return as a float representing the number of seconds:


```
>>> d = datetime(1970, 1,1) + timedelta(milliseconds=16712345)
>>> d
datetime.datetime(1970, 1, 1, 4, 38, 32, 345000)
>>> d.timestamp()
45512.345
>>> d.timestamp() * 1000
45512345.0
```

girarda: updated the test cases to show we retrain ms precision

PR #29514, opened 2023-08-17
Langchain destination: Clean up spec schema

flash1293:
This PR removes the `discriminator` property from the pydantic-generated json schema. The `discriminator` property is not used by the platform and references non-existing definitions as they are inlined into the places they are used (this happens to ensure compatibility with the UI)


girarda (approved):
TIL about pydantic discreminators.

:shipit:

airbyte-integrations/connectors/destination-langchain/destination_langchain/config.py
```diff
@@ -135,10 +136,16 @@ def resolve_refs(schema: dict) -> dict:
         pyschema: dict = json.loads(str_schema)
         del pyschema["definitions"]
         return pyschema
+    
+    @staticmethod
+    def remove_discriminator(schema: dict) -> None:
+        """pydantic adds "discriminator" to the schema for oneOfs, which is not treated right by the platform as we inline all references"""
```
girarda, line 142: change looks good. leaving a note that this is something we could do at the CDK level if this is a widespread issue

PR #29522, opened 2023-08-17
🐛 Destination Bigquery - ensure raw dataset created with migration

jbfbell:
If the raw dataset had not previously existed, we must create it for the migration

### DIFF
--- a/airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/typing_deduping/BigQuerySqlGenerator.java
+++ b/airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/typing_deduping/BigQuerySqlGenerator.java
@@ -604,11 +604,16 @@ public class BigQuerySqlGenerator implements SqlGenerator<TableDefinition> {
   @Override
   public String migrateFromV1toV2(final StreamId streamId, final String namespace, final String tableName) {
     return new StringSubstitutor(Map.of(
+        "raw_namespace", StringUtils.wrap(streamId.rawNamespace(), QUOTE),
+        "dataset_location", datasetLocation,
         "v2_raw_table", streamId.rawTableId(QUOTE),
         "v1_raw_table", wrapAndQuote(namespace, tableName)
     )
     ).replace(
-        """      
+        """
+            CREATE SCHEMA IF NOT EXISTS ${raw_namespace}
+            OPTIONS(location="${dataset_location}");
+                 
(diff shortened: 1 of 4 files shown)


airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/typing_deduping/BigQuerySqlGenerator.java
```diff
@@ -604,11 +604,16 @@ private String wrapAndQuote(final String namespace, final String tableName) {
...
         "v2_raw_table", streamId.rawTableId(QUOTE),
         "v1_raw_table", wrapAndQuote(namespace, tableName)
     )
     ).replace(
-        """      
+        """
+            CREATE SCHEMA IF NOT EXISTS ${raw_namespace}
+            OPTIONS(location="${dataset_location}");
+                 
             CREATE OR REPLACE TABLE ${v2_raw_table} (
```
evantahler, line 617: Does this table create statement also need `OPTIONS(location="${dataset_location}")`, or will being in the dataset take care of it?

airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/typing_deduping/BigQuerySqlGenerator.java
```diff
@@ -604,11 +604,16 @@ private String wrapAndQuote(final String namespace, final String tableName) {
...
+        "raw_namespace", StringUtils.wrap(streamId.rawNamespace(), QUOTE),
+        "dataset_location", datasetLocation,
         "v2_raw_table", streamId.rawTableId(QUOTE),
         "v1_raw_table", wrapAndQuote(namespace, tableName)
     )
     ).replace(
-        """      
+        """
+            CREATE SCHEMA IF NOT EXISTS ${raw_namespace}
+            OPTIONS(location="${dataset_location}");
```
evantahler, line 615: ✅

evantahler (approved):
Makes sense!

airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/typing_deduping/BigQuerySqlGenerator.java
jbfbell: fairly confident that bigquery doesn't let you create tables in a different location than the dataset

PR #29542, opened 2023-08-17
cat: use airbyte-protocol-model~=0.4.0

alafanechere:
CAT was not pinning its dependecy to airbyte-protocols-models. It led to the installation of an old version which is semantically higher than the most recent one.

### DIFF
--- a/airbyte-integrations/bases/connector-acceptance-test/pyproject.toml
+++ b/airbyte-integrations/bases/connector-acceptance-test/pyproject.toml
@@ -12,7 +12,7 @@ homepage = "https://github.com/airbytehq/airbyte"
 
 [tool.poetry.dependencies]
 python = "^3.10"
-airbyte-protocol-models = "*"
+airbyte-protocol-models = "<1.0.0"
 dagger-io = "==0.6.4"
 PyYAML = "~=6.0"
 icdiff = "~=1.9"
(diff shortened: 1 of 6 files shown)


airbyte-integrations/bases/connector-acceptance-test/pyproject.toml
```diff
@@ -12,7 +12,7 @@ homepage = "https://github.com/airbytehq/airbyte"
 
 [tool.poetry.dependencies]
 python = "^3.10"
-airbyte-protocol-models = "*"
+airbyte-protocol-models = ">=0.4.0,<0.5.0"
```
maxi297, line 15: Should we upper bound to a major even? I guess there would be no breaking changes until a major is changed which mean minor changes are probably fine

maxi297 (approved):
LGTM!

alafanechere:
/legacy-publish connector=bases/connector-acceptance-test run-tests=false

alafanechere:
/legacy-publish connector=bases/connector-acceptance-test run-tests=false

alafanechere:
/legacy-publish connector=bases/connector-acceptance-test run-tests=false

alafanechere:
/legacy-publish connector=bases/connector-acceptance-test run-tests=false

alafanechere:
Publishing manually in https://github.com/airbytehq/airbyte/actions/runs/5895733022 as the slash command is down.

PR #29544, opened 2023-08-17
✨ Source S3: (CATs) have schema within catalog like we would in the platform

maxi297:
## What
It does not make sense to have a catalog without json_schema. Hence we are fixing this to allow tests to pass still with source-s3 v4 as we will not infer schema if the json schema provided is empty.


maxi297:
Closing as this was to validate that V3 would support this new catalog and that there were no compatibility issues. This branch has been merged in https://github.com/airbytehq/airbyte/pull/29471

maxi297:
Re-opening as we will created separate commit in master for visibility purposes as there are too many changes to be done for S3 V4 rollout

maxi297:
This will be part of a rollout feature parity branch as there are too many changes and it would be too time consuming separating them

PR #29548, opened 2023-08-17
airbyte-ci: use GITHUB_TOKEN instead of GH_MAINTENANCE

alafanechere:
## What
Python 3.10 is installed on runners through a rate limited Github action. We exceeded this limit today.
I'm changing the token to the default `GITHUB_TOKEN`.
The other one was used originally because it has permissions on multiple repos. It's not a requirement for our existing pipelines AFAIK.


nguyenaiden (approved):
Thanks for doing this!

PR #29550, opened 2023-08-17
[skip ci]: alpha connectors should be community

alafanechere:
We mistakenly labelled a bunch of alpha connectors as certified. We should roll them back to community.


evantahler (approved):
And an alphabetical sort!

PR #29560, opened 2023-08-17
Destination bigquery v2: verify error on reserved column name prefixes

### LINKED ISSUE #28418: Determine if we need to handle table name prefix reservations in BQ
And there's some new logic to handle unusual column name prefixes https://github.com/airbytehq/airbyte/blob/fa4a278a2d496e666ddb39325b1b219c48a464e9/airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/typing_deduping/BigQuerySqlGenerator.java#L54

TODO: 
* check on Cloud if there are real streams/columns using these prefixes.
* If there are less than 5 instances, ignore this story, but confirm that BQ will throw when creating a table with these bad prefixes
* If there are more than 5 instances, reimplement the python name mangling logic.

edgao:
closes https://github.com/airbytehq/airbyte/issues/28418

as noted in https://github.com/airbytehq/airbyte/issues/28418#issuecomment-1682991494, this case is never actually happening. remove support for it + add a test to verify that we throw an error.

minor bump since this is technically behavior change, but impact is nonexistent.


evantahler (approved):
👍 
Thanks for doing the research to determine that this can throw vs something we need to handle more elegantly.

PR #29581, opened 2023-08-18
CDK: Remove list endpoint

### LINKED ISSUE #25774: Use manifest/resolve endpoint to get the list of streams instead of streams/list
The `streams/list` endpoint is currently used to get the list of streams defined in a manifest and their URL. When grooming https://github.com/airbytehq/airbyte/issues/25623, we decided to stop showing the URL because it has too many inconsistencies.

Since the URL does not need to be rendered in the UI, we can remove the `streams/list` endpoint and instead use the `manifest/resolve` to verify connector manifests.

## Acceptance criteria
- [x] The webapp does not show the stream's URL above the test read button
- [x] The webapp calls the `manifest/resolve` endpoint instead of `streams/list` to validate the manifest
- [ ] The streams/list endpoint is removed from the OpenAPI spec
- [ ] The StreamsHandler is removed from the Java connector builder server 
- [ ] The `list_streams` method is removed from the Python side of the connector builder

flash1293:
Fixes https://github.com/airbytehq/airbyte/issues/25774

The list_streams command is not required anymore, this PR is removing it.

### DIFF
--- a/airbyte-cdk/python/airbyte_cdk/connector_builder/README.md
+++ b/airbyte-cdk/python/airbyte_cdk/connector_builder/README.md
@@ -11,17 +11,17 @@ python main.py read --config path/to/config --catalog path/to/catalog
 ```
 
 Note:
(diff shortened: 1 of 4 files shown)


brianjlai (approved):
Changes look good. approving but naturally this will come after https://github.com/airbytehq/airbyte-platform-internal/pull/8403 presumably

PR #29590, opened 2023-08-18
Destination bigquery v2: Switch raw table to store airbyte_data as string to preserve number precision

edgao:
a naive `parse_json('{"foo": 67.174118}'` throws `Invalid input: Input number: 67.174118 cannot round-trip through string representation; error in PARSE_JSON expression`. This is a dv2 regression.

review this PR first, then look at https://github.com/airbytehq/airbyte/pull/29624 (which implements the migration for existing tables).

This PR:
* switches the raw tables to store `_airbyte_data` as string
* parses root-level numbers without loss of precision
* continue parsing nested JSON, which will lose precision.
* Adds a `Jsons.deserializeExact` method, which handles floats as BigDecimal internally
* Updates test cases to verify correct behavior (i.e. assert bigquery loss of precision in nested columns only; assert snowflake retains precision in all cases)

changelog message for bigquery: (minor bump)
```
Destinations v2: Changed raw table format to preserve high-precision numeric values
```


evantahler:
Related to a bug observed [here](https://cloud.airbyte.com/workspaces/5acd8e51-0e69-4dda-872d-f16ee6002c19/connections/b5e78b20-69e9-4728-982f-6426ae220862/job-history)

evantahler:
What a slack [thread](https://airbytehq-team.slack.com/archives/C03C4AVJWG4/p1692372078262529)!

Decision:
* For now, we will use  wide_number_mode=round in all JSON queries for bigquery.  This means that syncs won't fail, top-level properties in the final table will have the exact correct values, but final table columns which are JSON and contain long decimal numbers may become floats and gain some decimal values.

To do Today:
* Bigquery raw data should be stored as text so we don't loose precision
* Migrate the raw table from JSON to text for existing beta users OR inform all existing beta users they need to do a full-reset (and add a note to delete this code in 1 month)
* Add wide_number_mode=round wherever we do JSON work in BigQuery

To do Later:
* Add a bigquery option to allow users who have big decimals within nested properties to choose how we represent that data in the final tables (either JSON+round or STRING)

airbyte-integrations/bases/base-typing-deduping-test/src/main/java/io/airbyte/integrations/base/destination/typing_deduping/BaseTypingDedupingTest.java
```diff
@@ -588,7 +588,7 @@ public static List<JsonNode> readRecords(final String filename) throws IOExcepti
         .map(String::trim)
         .filter(line -> !line.isEmpty())
         .filter(line -> !line.startsWith("//"))
-        .map(Jsons::deserialize)
+        .map(Jsons::deserializeExact)
```
edgao, line 591: in the middle of a bunch of formatting changes, this line actually matters

airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/formatter/DefaultBigQueryRecordFormatter.java
```diff
@@ -34,26 +34,21 @@ public class DefaultBigQueryRecordFormatter extends BigQueryRecordFormatter {
...
       destinationV2record.put(JavaBaseConstants.COLUMN_NAME_AB_RAW_ID, UUID.randomUUID().toString());
       destinationV2record.put(JavaBaseConstants.COLUMN_NAME_AB_EXTRACTED_AT, getEmittedAtField(recordMessage));
       destinationV2record.put(JavaBaseConstants.COLUMN_NAME_AB_LOADED_AT, null);
-      if (useObjectForData()) {
-        destinationV2record.put(JavaBaseConstants.COLUMN_NAME_DATA,
-            StandardNameTransformer.formatJsonPath(recordMessage.getData()));
-      } else {
-        destinationV2record.put(JavaBaseConstants.COLUMN_NAME_DATA, getData(recordMessage));
-      }
+      destinationV2record.put(JavaBaseConstants.COLUMN_NAME_DATA, getData(recordMessage));
```
edgao, line 51: @tryangul fyi, standard inserts mode is modifying how `_airbyte_data` gets formatted (in addition to changing the column type json -> string on line 37). Hopefully nothing will break from your perspective 🤞

jbfbell (approved):

edgao:
snowflake test failures look unrelated (various legacy normalization tests, which are known to be flaky)

PR #29598, opened 2023-08-18
Source Intercom: migrate to latest cdk (0.50.2)

### LINKED ISSUE #24486: Source Intercom: Add undeclared columns to spec
## What

Intercom currently fails the updated test. Its `acceptance-test-config.yml` was edited with the

**How this will help**:
  previously being sent. Doing this allows us to enable column selection for Intercom!

## How

```
columns.txt:./source-intercom.txt-83-The admins stream has the following schema errors:
columns.txt:./source-intercom.txt:84:Additional properties are not allowed ('team_priority_level' was unexpected)
columns.txt:./source-intercom.txt-118-The conversations stream has the following schema errors:
columns.txt:./source-intercom.txt:119:Additional properties are not allowed ('topics' was unexpected)
columns.txt:./source-intercom.txt-401-The contacts stream has the following schema errors:
columns.txt:./source-intercom.txt:402:Additional properties are not allowed ('opted_out_subscription_types', 'utm_content', 'utm_campaign', 'utm_source', 'referrer', 'utm_term', 'utm_medium' were unexpected)
columns.txt:./source-intercom.txt-612-The company_attributes stream has the following schema errors:
columns.txt:./source-intercom.txt:613:Additional properties are not allowed ('id' was unexpected)
columns.txt:./source-intercom.txt-651-The contact_attributes stream has the following schema errors:
columns.txt:./source-intercom.txt:652:Additional properties are not allowed ('id' was unexpected)

```

5. Profit!

Definition of done: Intercom passes CAT without declaring `fail_on_extra_columns: false`. If the
movement.

roman-yermilov-gl:
## What
*Migrating to latest cdk 0.29.0 -> 0.50.2*
- *`HttpRequesterWithRateLimiter` is now a subclass of `HttpRequester` and it has been simplified. It mostly uses inheritance except some parts related to `request_options_provider`. Because of limitation on custom component it is not allowed to include `request_options_provider` section to custom requester*
- *`IncrementalSingleSliceCursor` and `IncrementalSubstreamSlicer` were significantly changed because of changes in Slicer interface in cdk 0.50.2. They have also been simplified by reducing the amount of code and simplifying the logic*

https://github.com/airbytehq/airbyte/issues/28214

Fixed: #24486

## How
- *Change cdk and manifest version to 0.50.2*
- *Implement new interfaces for all custom components*
- *Make refactoring and rethinking of current slicing logic*

### DIFF
--- a/airbyte-integrations/connectors/source-intercom/source_intercom/components.py
+++ b/airbyte-integrations/connectors/source-intercom/source_intercom/components.py
@@ -2,41 +2,36 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 #
 
-import os
 from dataclasses import InitVar, dataclass, field
-from functools import lru_cache, wraps
+from functools import wraps
 from time import sleep
 from typing import Any, Iterable, List, Mapping, MutableMapping, Optional, Union
 
-import dpath.util
 import requests
-from airbyte_cdk.models import AirbyteMessage, SyncMode, Type
-from airbyte_cdk.sources.declarative.auth.declarative_authenticator import DeclarativeAuthenticator, NoAuth
+from airbyte_cdk.models import SyncMode
+from airbyte_cdk.sources.declarative.incremental.cursor import Cursor
 from airbyte_cdk.sources.declarative.interpolation.interpolated_string import InterpolatedString
 from airbyte_cdk.sources.declarative.partition_routers.substream_partition_router import ParentStreamConfig
-from airbyte_cdk.sources.declarative.requesters.error_handlers.default_error_handler import DefaultErrorHandler
-from airbyte_cdk.sources.declarative.requesters.error_handlers.error_handler import ErrorHandler
 from airbyte_cdk.sources.declarative.requesters.error_handlers.response_status import ResponseStatus
+from airbyte_cdk.sources.declarative.requesters.http_requester import HttpRequester
 from airbyte_cdk.sources.declarative.requesters.request_option import RequestOptionType
 from airbyte_cdk.sources.declarative.requesters.request_options.interpolated_request_input_provider import InterpolatedRequestInputProvider
-from airbyte_cdk.sources.declarative.requesters.requester import HttpMethod, Requester
-from airbyte_cdk.sources.declarative.stream_slicers.stream_slicer import StreamSlicer
+from airbyte_cdk.sources.declarative.requesters.request_options.interpolated_nested_request_input_provider import InterpolatedNestedRequestInputProvider
 from airbyte_cdk.sources.declarative.types import Config, Record, StreamSlice, StreamState
 from airbyte_cdk.sources.streams.core import Stream
+
 
 RequestInput = Union[str, Mapping[str, str]]
 
 
 @dataclass
-class IncrementalSingleSlice(StreamSlicer):
-
+class IncrementalSingleSliceCursor(Cursor):
     cursor_field: Union[InterpolatedString, str]
     config: Config
     parameters: InitVar[Mapping[str, Any]]
-    _cursor: dict = field(default_factory=dict)
-    initial_state: dict = field(default_factory=dict)
 
     def __post_init__(self, parameters: Mapping[str, Any]):
+        self._state = {}
         self.cursor_field = InterpolatedString.create(self.cursor_field, parameters=parameters)
 
     def get_request_params(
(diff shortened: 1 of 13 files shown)


airbyte-integrations/connectors/source-intercom/source_intercom/components.py
```diff
@@ -175,186 +139,52 @@ def __post_init__(self, parameters: Mapping[str, Any]):
...
-                # Skip non-records (eg AirbyteLogMessage)
-                if isinstance(parent_record, AirbyteMessage):
-                    if parent_record.type == Type.RECORD:
-                        parent_record = parent_record.record.data
+            )
 
+            for parent_record in parent_records_gen:
                 try:
-                    substream_slice = dpath.util.get(parent_record, self.parent_field)
+                    substream_slice_value = parent_record[self.parent_field]
```
lazebnyi, line 177: can we use `get` method logic to avoid` try...except `?

docs/integrations/sources/intercom.md
```diff
@@ -49,6 +49,7 @@ The Intercom connector should not run into Intercom API limitations under normal
 
 | Version | Date       | Pull Request                                             | Subject                                                                                       |
 |:--------| :--------- | :------------------------------------------------------- | :-------------------------------------------------------------------------------------------- |
+| 0.3.0   | 2023-05-25 | [29598](https://github.com/airbytehq/airbyte/pull/29598) | Migrate to latest cdk==0.50.2                                                                 |
```
lazebnyi, line 52: Let's change subject to more human readable.

lazebnyi (changes requested):

airbyte-integrations/connectors/source-intercom/source_intercom/spec.json
```diff
@@ -20,6 +20,18 @@
         "type": "string",
         "description": "Access token for making authenticated requests. See the <a href=\"https://developers.intercom.com/building-apps/docs/authentication-types#how-to-get-your-access-token\">Intercom docs</a> for more information.",
         "airbyte_secret": true
+      },
+      "client_id": {
```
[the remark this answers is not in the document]
roman-yermilov-gl: As I understand it is a requirement now. There is even test in cdk that checks that fields declared in oauth config are also present in spec

airbyte-integrations/connectors/source-intercom/source_intercom/components.py
roman-yermilov-gl: For me the use of `try...except` block here is a little bit more clear. But in case of performance it will take longer to handle an exception block if a parent record does not contain a key. Performance should allways be on first place so I agree to switch `try...except` to `if` statement

docs/integrations/sources/intercom.md
roman-yermilov-gl: Not sure what else I can add, so added some words describing why migration even take place

lazebnyi (approved):

docs/integrations/sources/intercom.md
bazarnov: how about:
```
"Resolved conflicts after switching to the latest CDK version."
```

airbyte-integrations/connectors/source-intercom/source_intercom/manifest.yaml
```diff
@@ -23,6 +23,8 @@ definitions:
     request_headers:
       Intercom-Version: '2.5' # ATTENTION: API version change is possible here
       Accept: 'application/json'
+    error_handler:
+      type: "DefaultErrorHandler"
```
bazarnov, line 27: Is there a reason behind this addings?

docs/integrations/sources/intercom.md
roman-yermilov-gl: I have already prepared something like:
`Update custom components to make them compatible with latest cdk version 0.50.2, simplify logic, update schemas`

bazarnov (approved):
Looks good! Nice work!

docs/integrations/sources/intercom.md
bazarnov: I don't think we should include the specific version in the description, knowing that we should always use `latest` one.

roman-yermilov-gl: Agree. I will remove version

airbyte-integrations/connectors/source-intercom/source_intercom/manifest.yaml
roman-yermilov-gl: I was trying to use inheritance as much as possible. We have an error handler defined in parent `HttpRequester` component but it is `None` by default. So I added `DefaultErrorHandler` explicitly to the `manifest.yaml`

docs/integrations/sources/intercom.md
roman-yermilov-gl: Done

PR #29634, opened 2023-08-18
source-mysql : Initial load. Handle composite pk case w/multiple repeated values

### LINKED ISSUE #29574: source-mysql : Chunking logic sync bugs in case of composite primary key
Context : 
- Chunking logic for the initial CDC load was added in : https://github.com/airbytehq/airbyte/pull/29109
- Sorting logic to surface the first field in a composite primary index was implemented in https://github.com/airbytehq/airbyte/pull/29346. This is needed because our queries can only checkpoint by a single field. In addition, data is co-located by the first field in the composite primary index, so this makes the loading queries efficient. 

However, while implementing chunking, one of the fields associated with the composite key might have repeat values. 

Consider a table with the following schema with the composite pk index (state, city) : Assume it has ~1 billion records, with state = CA for all rows. 

| state | city | slogan |
| ------------- | ------------- |-------------|
| CA | San Francisco | Paris of the West |
| CA| San Jose | Urban Sprawl |
| CA | Los Angeles | Terrible Traffic |

Only a combo of state, city is guaranteed to be unique. At worst, the entire table _could_ have the same value for the first field in the primary index. e.g. state always is CA. With the current chunking implementation, we will skill syncing all of the records. Since we issue a > query  

```
Query 1 : select * from table order by state limit 1,800,000
Query 2 : select * from table where state > CA order by pk limit 1,800,000 
...End of sync..
```

There are a couple of ways to deal with this : 
1. Option 1 (short-term) : If we detect a stream with a composite primary key, we can fallback to a single query and checkpoint by pk : 
```
Query 1 : select * from table order by state
...End of sync..
```
Advantages: 
- Don't need to do any guesswork on the shape of the data associated w/first field in composite pk
- Simplest solution

Disadvantages: 
- Degraded performance 

2. Option 2 (long-term) : If we encounter a composite index, we will checkpoint by all of the fields in the index. 
```
Query 1 : select * from table order by state, city limit x
Query 1 : select * from table where state >= CA and city > 'San Francisco' order by state, CA limit x
...End of sync..
```

Notes: 
- The last field in the pk (city in this case) would have a > condition as opposed to >=
- Unknown what the efficiency of the queries are. We'll need to verify that performance is not taking a hit
- Need to change state to have a list of primary key states instead of just one. Unclear what the PII risk here is
 
Advantages : 
- If the above open questions are verified with no performance hits, this will be the most robust way to handle composite indexes, which will support chunking + checkpointing

Disadvantages:
- Need to balance out how much effort this solution will require vs how often composite indexes occur
 

Considered Options
1. Issue a >= query if it's a composite key. e.g. 

```
Query 1 : select * from table order by state limit 1,800,000
Query 2 : select * from table where state >= CA order by state limit 1,800,000 
Query 3 : select * from table where state >= CA order by state limit 1,800,000 
Query 4 : select * from table where state >= CA order by state limit 1,800,000 

...Timeout..
```

Not a fan of this. With the above scenario, we will spin queries indefinitely. We'll continue to read records so the platform won't even kill the sync. 

2. Double chunk size when we detect the same value. 

Again, doesn't completely solve the issues with Option 2

4. Check previously processed pk value. Change queries to a full table snapshot if we see the same pk value processed
```
Query 1 : select * from table order by state limit 1,800,000
Query 2 : select * from table where state >= CA order by state limit 1,800,000 
...Detecting duplicate values for state...
Query 3 : select * from table where state >= CA order by state
```
We end up re-processing the first series of chunks until we encounter a duplicated value. This solution also does not work as we may encounter some data loss at the boundary conditions

akashkulk:
Closes https://github.com/airbytehq/airbyte/issues/29574

For streams with a composite primary key, this would look like :
1. Issue a ```select * from table order by pk_field``` (no chunking).
2. For subsequent queries (from a new sync) issue a `select * from table order by pk_field where pk_field >= pk_val`

^That will account for cases with multiple repeated primary fields in the composite index.

For further discussion on options please refer to the issue. This is a stop-gap and we may revisit with a more robust solution that will allow chunking queries when there is a composite primary key


subodh1810 (approved):
Approved! As discussed over call I pushed one commit to minor refactor the computeNext method of the iterator class to avoid re-initialising the stream object

airbyte-integrations/connectors/source-mysql/src/main/java/io/airbyte/integrations/source/mysql/initialsync/MySqlInitialLoadRecordIterator.java
```diff
@@ -114,17 +123,31 @@ private PreparedStatement getPkPreparedStatement(final Connection connection) {
...
         LOGGER.info("pkLoadStatus value is : {}", pkLoadStatus.getPkVal());
         final String quotedCursorField = enquoteIdentifier(pkLoadStatus.getPkName(), quoteString);
-        // Since a pk is unique, we can issue a > query instead of a >=, as there cannot be two records with the same pk.
-        final String sql = String.format("SELECT %s FROM %s WHERE %s > ? ORDER BY %s LIMIT %s", wrappedColumnNames, fullTableName,
-            quotedCursorField, quotedCursorField, chunkSize);
+        final String sql;
+        // We cannot load in chunks for a composite key load, since each field might not have distinct values. Furthermore, we have to issue a >=
+        // query since we may not have processed all of the data associated with the last saved primary key value.
+        if (isCompositeKeyLoad) {
+          sql = String.format("SELECT %s FROM %s WHERE %s >= ? ORDER BY %s", wrappedColumnNames, fullTableName,
```
nguyenaiden, line 145: This could lead to duplicate rows for `Incremental | append` but per our conversation, that's something we don't need to optimize for since MySQL CDC customers almost exclusively use Append + deduped.

airbyte-integrations/connectors/source-mysql/src/main/java/io/airbyte/integrations/source/mysql/initialsync/MySqlInitialLoadRecordIterator.java
```diff
@@ -114,17 +123,31 @@ private PreparedStatement getPkPreparedStatement(final Connection connection) {
...
+              quotedCursorField, quotedCursorField);
+        } else {
+          sql = String.format("SELECT %s FROM %s WHERE %s > ? ORDER BY %s LIMIT %s", wrappedColumnNames, fullTableName,
```
nguyenaiden, line 148: Nit: can we save this in a static variable to explicitly identify the type of query that's used for composite vs. single column PKs?

nguyenaiden (approved):
+1. Approve to unblock. Address nits as you see fit.

octavia-approvington:

octavia-approvington (approved):

Issue #29419, opened 2023-08-14
[Postgres] Validate that no full vacuum occurred during initial snapshot (start and finish)


prateekmukhedkar:
### Connector Name

source-postgres

### Connector Version

3.1.3

### What step the error happened?

During the sync

### Revelant information

While we validate that a full vacuum happened at the _start_ of a sync job (including every attempt), we don't check it before reading every chunk. It is possible that as we are running a full snapshot, a full vacuum starts (either on demand or on schedule) and reorders rows on disk. Since every chunk is its own read transaction there is a window of time, even though small, where a full vacuum can slip in. The expected behavior is that we discard any state and start the sync from the beginning. 

There are couple of ways to handle this:

- Check for vacuum before reading every chunk
- Issue some database command that prevents a full vacuum from taking place during the time we are taking the initial snapshot
- Check for vacuum after we read all chunks, and fail the sync if a vacuum was detected.

While the first option catches a vacuum point in time I am not sure about the implementation overhead. And we don't know if there is any database command for option 2. So we want to repeat the same check that we do at the start of the sync, this time after we finish reading all chunks. And if we find that a vacuum happened we should discard the state and start reading from the beginning of the dataset. 

### Contribute

- [ ] Yes, I want to contribute

PR #29672, opened 2023-08-21
postgres-source: validate filenode before processing each ctid chunk

subodh1810:
Issue : https://github.com/airbytehq/airbyte/issues/29419

I refactored the ctid chunking logic and instead of creating multiple iterators up front, the logic is encapsulated inside a new class `InitialSyncCtidIterator` which creates the chunks dynamically, validates that the file node is still valid before each chunk and if not then creates new chunks.
This also fixes a minor bug related to usage of expired file node. Before we were querying the file node for all tables before starting syncing data and use that file node throughout the sync as part of the ctid state. But imagine if the first stream takes 10 hours to sync, then for the second stream the file node that we had was 10 hours old which might no longer be valid since in the 10 hours the stream might have gone through a VACUUM. Now we refresh the filenode for each stream when we start syncing that stream. 

There are few renaming of methods/variables for clarity as well.


airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/FileNodeHandler.java
```diff
@@ -0,0 +1,41 @@
+package io.airbyte.integrations.source.postgres.ctid;
+
+import io.airbyte.protocol.models.AirbyteStreamNameNamespacePair;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class FileNodeHandler {
```
akashkulk, line 10: nit : can you just add comments related to this class?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/InitialSyncCtidIterator.java
```diff
@@ -0,0 +1,230 @@
...
+              LOGGER.info("The latest file node {} for stream {} is equal to the last known file node {} known to Airbyte.",
+                  latestFileNode,
+                  lastKnownFileNode,
+                  airbyteStream);
+            }
+          }
+          lastKnownFileNode = latestFileNode;
+          fileNodeHandler.updateFileNode(airbyteStream, latestFileNode);
+        } else {
+          LOGGER.warn("Airbyte could not query the latest file node for stream {}. Continuing sync as usual.", airbyteStream);
```
akashkulk, line 104: Under what circumstances would we fail querying the latest file node? I do agree that if we cannot query for the file node we should continue. But if the reason this query fails is normally due to a full vacuum currently taking place then maybe this should result in a re-sync

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/InitialSyncCtidIterator.java
```diff
@@ -0,0 +1,230 @@
...
+    try {
+      if (subQueriesPlan.isEmpty() && currentIterator == null) {
+        initSubQueries();
+      }
+
+      if (currentIterator == null || !currentIterator.hasNext()) {
+        final Optional<Long> mayBeLatestFileNode = PostgresQueryUtils.fileNodeForIndividualStream(database, airbyteStream, quoteString);
+        if (mayBeLatestFileNode.isPresent()) {
+          final Long latestFileNode = mayBeLatestFileNode.get();
+          if (lastKnownFileNode != null) {
```
akashkulk, line 91: nit : can you combine if conditions for readability?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/InitialSyncCtidIterator.java
```diff
@@ -0,0 +1,230 @@
...
+    this.subQueriesPlan = new LinkedList<>();
+    this.tableName = tableName;
+    this.tableSize = tableSize;
+  }
+
+  @CheckForNull
+  @Override
+  protected RowDataWithCtid computeNext() {
+    try {
+      if (subQueriesPlan.isEmpty() && currentIterator == null) {
```
akashkulk, line 83: Can we just define a boolean isInitialized variable here?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/InitialSyncCtidIterator.java
```diff
@@ -0,0 +1,230 @@
...
+
+        final Pair<Ctid, Ctid> p = subQueriesPlan.remove();
+        currentIterator = AutoCloseableIterators.fromStream(getStream(p), airbyteStream);
+      }
+
+      if (currentIterator.hasNext()) {
+        return currentIterator.next();
+      }
+
+      assert subQueriesPlan.isEmpty();
```
akashkulk, line 123: Can we always assert that this is true? 

Consider : we have pre-computed 10 subqueries, each corresponding to a range of Ctids. I think it's completely plausible for some range in the middle (e.g. 6th query) to not return any data. In this case - we'll fail this assertion. 

In fact - we should keep going until subQueriesPlan is empty to prevent this situation

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/InitialSyncCtidIterator.java
```diff
@@ -0,0 +1,230 @@
...
+  @CheckForNull
+  @Override
+  protected RowDataWithCtid computeNext() {
+    try {
+      if (subQueriesPlan.isEmpty() && currentIterator == null) {
+        initSubQueries();
+      }
+
+      if (currentIterator == null || !currentIterator.hasNext()) {
+        final Optional<Long> mayBeLatestFileNode = PostgresQueryUtils.fileNodeForIndividualStream(database, airbyteStream, quoteString);
```
akashkulk, line 88: Is querying for file nodes adding significant performance overhead? My thoughts are that checking for the fileNode doesn't have to be done per chunk. We can have it be configurable such that it's done every N chunks to save on querying for file nodes so often. 

The tradeoff here is that we may sync more data before realizing that a full vacuum has occurred (and have to resync the entire table) but I think that that is a tradeoff we are ok with

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/InitialSyncCtidIterator.java
```diff
@@ -0,0 +1,230 @@
...
+        } else {
+          LOGGER.warn("Airbyte could not query the latest file node for stream {}. Continuing sync as usual.", airbyteStream);
+        }
+
+        if (currentIterator != null) {
+          currentIterator.close();
+        }
+
+        if (subQueriesPlan.isEmpty()) {
+          return endOfData();
```
akashkulk, line 112: Seems like we do - but just want to double check. We are issuing one last check for vacuuming before returning `endOfData()` correct?

akashkulk (changes requested):
Looks good in general - some comments on edge cases and small nits

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/InitialSyncCtidIterator.java
rodireich, line 88: why maybe?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/InitialSyncCtidIterator.java
rodireich: I wonder also. Should this be a cause for failing a sync?

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/InitialSyncCtidIterator.java
```diff
@@ -0,0 +1,230 @@
...
+    this.schemaName = schemaName;
+    this.sourceOperations = sourceOperations;
+    this.subQueriesPlan = new LinkedList<>();
+    this.tableName = tableName;
+    this.tableSize = tableSize;
+  }
+
+  @CheckForNull
+  @Override
+  protected RowDataWithCtid computeNext() {
```
rodireich, line 81: I know this is not trivial, but if we could test the logic in this `computeNext()` class since it's very stateful.
Compare to CompositeIteratorTest

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/InitialSyncCtidIterator.java
akashkulk: Discussed offline - query cost is slow and can be revisited in the future. For now, we will keep it simple

airbyte-integrations/connectors/source-postgres/src/main/java/io/airbyte/integrations/source/postgres/ctid/InitialSyncCtidIteratorConstants.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+ */
+
+package io.airbyte.integrations.source.postgres.ctid;
+
+public class InitialSyncCtidIteratorConstants {
+
+  public static final int MAX_ALLOWED_RESYNCS = 5;
+  public static final int QUERY_TARGET_SIZE = 1;
```
rodireich, line 10: nit: add back "_GB"

airbyte-integrations/connectors/source-postgres/src/test/java/io/airbyte/integrations/source/postgres/CdcPostgresSourceTest.java
```diff
@@ -947,4 +947,37 @@ protected void verifyCheckpointStatesBySeconds() throws Exception {
...
+    }
+
+    /**
+     * Setting the property to make the
+     * {@link io.airbyte.integrations.source.postgres.ctid.InitialSyncCtidIterator} use smaller page
+     * size of 8KB instead of default 1GB This allows us to make sure that the iterator logic works with
+     * multiple pages (sub queries)
+     */
+    final JsonNode config = getConfig();
+    ((ObjectNode) config).put("use_test_page_size", true);
```
rodireich, line 973: nit: this is a _chunk_ and not a page size

rodireich (approved):

rodireich:
A couple of nits but looks good otherwise

akashkulk (approved):

prateekmukhedkar (approved):
Started a sync, started an initial vacuum while the initial sync was running and observed this log indicating presence of check:
`2023-08-31 16:26:56 [44msource[0m > WARN i.a.i.s.p.c.InitialSyncCtidIterator(resetSubQueries):160 The latest file node 16572 for stream 16552 is not equal to the last known file node public_users known to Airbyte. Airbyte will sync this table from scratch again
`

octavia-approvington:

octavia-approvington (approved):

PR #29682, opened 2023-08-21
🐛 Create a connection in the acceptance test setup and close is in the tear down

benmoriceau:
## What
Update the Redshift acceptance test in order to limit the number of connections opened by the redshift acceptance tests.
We are now directly creating a connection without using a datasource. It is needed because it has been observed that:
- `datasource.close()` doesn't close the opened connection
- According to the javadoc, `datasource.getConnection()` may return a new connection.
This means that we can not ensure that a connection is being closed by using the datasource. Which explains why we switch to using a Connection directly.

## Recommended reading order
1. `ConnectionFactory.java`
2. The test files.

### DIFF
--- a/airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
+++ b/airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
@@ -1,0 +1,27 @@
+package io.airbyte.db.factory;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.Map;
+import java.util.Properties;
+
+public class ConnectionFactory {
+    public static Connection create(final String username,
+                                    final String password,
+                                    final Map<String, String> connectionProperties,
+                                    final String jdbcConnectionString) {
+        try {
+            Properties properties = new Properties();
+            properties.put("user", username);
+            properties.put("password", password);
+            connectionProperties.forEach((k, v) -> properties.put(k, v));
+
+           return DriverManager.getConnection(jdbcConnectionString,
+                    properties);
+        } catch (SQLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+}
--- a/airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
+++ b/airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
@@ -28,7 +28,7 @@ public class ConnectionFactory {
 
            return DriverManager.getConnection(jdbcConnectionString,
                     properties);
-        } catch (SQLException e) {
+        } catch (final SQLException e) {
             throw new RuntimeException(e);
         }
     }
--- a/airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
+++ b/airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
@@ -7,6 +7,15 @@ import java.util.Map;
 import java.util.Properties;
 
 public class ConnectionFactory {
+
+    /**
+     * Construct a new {@link Connection} instance using the provided configuration.
+     * @param username The username of the database user.
+     * @param password The password of the database user.
+     * @param connectionProperties The extra properties to add to the connection.
+     * @param jdbcConnectionString The JDBC connection string.
+     * @return The configured {@link Connection}
+     */
     public static Connection create(final String username,
                                     final String password,
                                     final Map<String, String> connectionProperties,
--- a/airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
+++ b/airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
@@ -1,3 +1,7 @@
+/*
+ * Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+ */
+
 package io.airbyte.db.factory;
 
 import java.sql.Connection;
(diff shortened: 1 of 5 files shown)


airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
```diff
@@ -0,0 +1,36 @@
...
+                                    final String jdbcConnectionString) {
+        try {
+            Properties properties = new Properties();
+            properties.put("user", username);
+            properties.put("password", password);
+            connectionProperties.forEach((k, v) -> properties.put(k, v));
+
+           return DriverManager.getConnection(jdbcConnectionString,
+                    properties);
+        } catch (SQLException e) {
```
tryangul, line 31: final

benmoriceau: Thanks! I wonder why my plugin didn't catch it.

airbyte-integrations/connectors/destination-redshift/src/test-integration/java/io/airbyte/integrations/destination/redshift/RedshiftStagingS3DestinationAcceptanceTest.java
```diff
@@ -120,7 +127,7 @@ public void testCheckIncorrectDataBaseFailure() throws Exception {
   public void testGetFileBufferDefault() {
     final RedshiftStagingS3Destination destination = new RedshiftStagingS3Destination();
     assertEquals(destination.getNumberOfFileBuffers(config),
-        FileBuffer.DEFAULT_MAX_CONCURRENT_STREAM_IN_BUFFER);
+            FileBuffer.DEFAULT_MAX_CONCURRENT_STREAM_IN_BUFFER);
```
davinchia, line 130: nit: I think this can be one line, not sure why the this wasn't previously the case since L137, L147 are the same.

evantahler (commented):
The tests actually pass, so 👍 from me!

I'd want another review (paging @jbfbell) to answer the question of if this is a special thing we are OK with for just redshift, of we should make this kind of change for all JDBC destination tests

airbyte-integrations/connectors/destination-redshift/src/test-integration/java/io/airbyte/integrations/destination/redshift/SshRedshiftDestinationBaseAcceptanceTest.java
```diff
@@ -201,8 +217,11 @@ protected void tearDown(final TestDestinationEnv testEnv, HashSet<String> TEST_S
         JdbcUtils.HOST_LIST_KEY,
         JdbcUtils.PORT_LIST_KEY,
         config -> {
-          getDatabaseFromConfig(config).query(ctx -> ctx.fetch(String.format("DROP USER IF EXISTS %s;", USER_WITHOUT_CREDS)));
+          getDatabase().query(ctx -> ctx.fetch(String.format("DROP USER IF EXISTS %s;", USER_WITHOUT_CREDS)));
         });
+    connection.setAutoCommit(false);
+    connection.commit();
+    connection.close();
```
davinchia, line 224: +1

evantahler: Can these 3 commands be wrapped in another method so we don't forget to do them all?

airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Construct a new {@link Connection} instance using the provided configuration.
+   *
+   * @param username The username of the database user.
+   * @param password The password of the database user.
+   * @param connectionProperties The extra properties to add to the connection.
+   * @param jdbcConnectionString The JDBC connection string.
+   * @return The configured {@link Connection}
+   */
```
davinchia, line 23: nit: can we also add to the javadoc that we need this because we run tests in parallel and the usual data source does not have sufficient connection guarantees to prevent creating too many connections

davinchia (commented):
Great stuff @benmoriceau ! 

Makes sense to me. I will defer to the Destinations team for approval merge.

airbyte-integrations/connectors/destination-redshift/src/test-integration/java/io/airbyte/integrations/destination/redshift/RedshiftStagingS3DestinationAcceptanceTest.java
benmoriceau: Done

airbyte-integrations/connectors/destination-redshift/src/test-integration/java/io/airbyte/integrations/destination/redshift/SshRedshiftDestinationBaseAcceptanceTest.java
benmoriceau: Done

benmoriceau:
> The tests actually pass, so 👍 from me!
> 
> I'd want another review (paging @jbfbell) to answer the question of if this is a special thing we are OK with for just redshift, of we should make this kind of change for all JDBC destination tests

I looked a Snowflake as well. One big difference is that Snowflake is creating one connection per test where we were creating one connection per query for redshift.

I haven't check the implementation of the redshiftJDBC connection but it could come from it as well, the javadoc says that the behavior of close on an un-commited connection is unknown.

airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
benmoriceau: Added a comment on the class itself

airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
```diff
@@ -0,0 +1,46 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.Map;
+import java.util.Properties;
+
+/**
+ * This class as been added in order to be able to save the connection in a test. It was found that
+ * the {@link javax.sql.DataSource} close method wasn't propagating the connection properly. It
+ * shouldn't be needed in our application code.
```
davinchia, line 16: What does 'shouldn't be needed in our application code' mean?

Should we replace DataSource through database code base with this?

davinchia (approved):
Great stuff!

Just one minor clarification.

airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
benmoriceau: This factory, I mean that we can use a datasource directly. Since it is created using Micronaut, it will be a singleton and thus reduce the leak risk. Changing that in our applications will make use to do unnecessary changes knowing that the datasource is not even created by use and managed through the application.yaml. It will need us to re-implement some options that we are using.

Here is an example of a datasource declaration from the worker:
```
datasources:
  config:
    connection-test-query: SELECT 1
    connection-timeout: 30000
    idle-timeout: 600000
    initialization-fail-timeout: -1 # Disable fail fast checking to avoid issues due to other pods not being started in time
    maximum-pool-size: 20
    minimum-idle: 0
    url: ${DATABASE_URL}
    driverClassName: org.postgresql.Driver
    username: ${DATABASE_USER}
    password: ${DATABASE_PASSWORD}
    pool-name: config-pool
```

jbfbell (approved):

octavia-approvington:
Lets do it! <br/> giddddy-up

octavia-approvington (approved):

airbyte-db/db-lib/src/main/java/io/airbyte/db/factory/ConnectionFactory.java
davinchia: In that case, can we clarify with something like, this does not apply to DI frameworks e.g. Micronaut?

'Application code' is confusing because there is a wide breadth of application code at Airbyte. A reader is going to get confused if they are going between Platform (Micronaut-ed) and Connectors (not-micronauted)

PR #29685, opened 2023-08-21
🐛 Destinations v2: improve behavior with special characters

edgao:
runtime code changes:
* bigquery
  * JSONPaths are now escaped correctly (single quotes, doublequotes, backslashes)
  * all string literals now use single quotes
* snowflake
  * identifiers are now escaped correctly (double quotes)
  * we skip PK validation if any PK column name contains a backtick (snowflake doesn't like having backticks inside scripts, apparently?)
  * the PK validation script now uses single quotes instead of `$$` quotes (because I couldn't figure out how to escape `$$` :shrug:)
    * in tests, we're using `\$\$` - but when running this in reality, it looks like snowflake ends up executing a literal `select "_airbyte_data"."the_column\$\$"`, which fails to fetch the actual column `the_column$$`
    * in general `$$` strings just don't behave well under weird conditions

new tests:
* `weirdColumnNames` has several new fields, and we verify that we can write all of those fields
* `noCrashOnSpecialCharacters` runs a sync with stream namespace, name, primary key, and cursor all containing various special characters
  * This test only verifies that we don't crash. Maybe it should also verify data? but I suspect normalization blows up dumbly anyway in these cases
  * This test is disabled for bigquery because bigquery keeps saying the raw table doesn't exist. E.g. this query says `Table not found: sql_generator_test_fvdrq__.raw_table`:
     ```sql
     INSERT INTO `sql_generator_test_fvdrq__`.`sql_generator_test_fvdrq__`
     (
     `sql_generator_test_fvdrq__`,
       _airbyte_meta,
       _airbyte_raw_id,
       _airbyte_extracted_at
     )
     SELECT 
       'arst',
       json'{}',
      _airbyte_raw_id,
      _airbyte_extracted_at
     FROM `sql_generator_test_fvdrq__`.`raw_table`
     ```
     but you can `select *` from that table just fine, and certainly it exists [in the UI](https://console.cloud.google.com/bigquery?ws=!1m5!1m4!4m3!1sdataline-integration-testing!2ssql_generator_test_fvdrq__!3ssql_generator_test_fvdrq__) :shrug:


airbyte-integrations/connectors/destination-bigquery/src/test-integration/java/io/airbyte/integrations/destination/bigquery/typing_deduping/BigQuerySqlGeneratorIntegrationTest.java
```diff
@@ -444,6 +445,16 @@ public void testFailureOnReservedColumnNamePrefix(final String prefix) {
...
+  /**
+   * Something about this test is borked on bigquery. It fails because the raw table doesn't exist,
+   * but you can go into the UI and see that it does exist.
+   */
+  @Override
+  @Disabled
+  public void noCrashOnSpecialCharacters(final String specialChars) throws Exception {
+    super.noCrashOnSpecialCharacters(specialChars);
+  }
+
```
evantahler, line 457: Is there a way to see the queries the UI is making, or use tablePlus or another SQL ui to see if they are doing something special to show you those reults?

edgao: it's actually dumber than that. the insert query in the PR description fails because the raw table doesn't exist, but if you pull out the `select` thing and run that on its own, it succeeds. So this works:
```sql
SELECT 
  'arst',
  json'{}',
 _airbyte_raw_id,
 _airbyte_extracted_at
FROM `sql_generator_test_fvdrq__`.`raw_table`
```

image

but the original insert query fails

image

edgao:
snowflake test failure:
```
SnowflakeS3CopyDestinationAcceptanceTest > testCheckWithKeyPairAuth() FAILED
    org.opentest4j.AssertionFailedError: expected:  but was: 
```

whatever this is, is unrelated to this PR, and specific to S3.

jbfbell (approved):

PR #29686, opened 2023-08-21
chore: commit the result of `./gradlew format`

aaronsteers:
This PR simply commits the result of `./gradlew format`. I don't think we will merge this PR, but let's use it for discussion and review of existing auto-format rules.

If too disruptive, we can recreate this PR as a set of smaller PRs based on directory structure and/or file types.


airbyte-integrations/connectors/destination-bigquery/src/main/java/io/airbyte/integrations/destination/bigquery/BigQueryRecordConsumer.java
```diff
@@ -46,13 +46,12 @@ public class BigQueryRecordConsumer extends FailureTrackingAirbyteMessageConsume
...
-      final Consumer<AirbyteMessage> outputRecordCollector,
-      final String defaultDatasetId,
-      TyperDeduper typerDeduper,
-      final ParsedCatalog catalog) {
+                                final Map<AirbyteStreamNameNamespacePair, AbstractBigQueryUploader<?>> uploaderMap,
+                                final Consumer<AirbyteMessage> outputRecordCollector,
+                                final String defaultDatasetId,
+                                TyperDeduper typerDeduper,
+                                final ParsedCatalog catalog) {
     this.bigquery = bigquery;
```
aaronsteers, line 55: Fwiw, I don't agree with this format change. I prefer the 4-character indent, as a rule, because it is kinder towards not requireing rightward-scrolling, text wrapping, and violating common-sense line length limits.

aaronsteers: As if in a meta-commentary to prove my point, GitHub's diff is almost unreadable without horizontal scrolling here. 😅 

image

aaronsteers:
Closing this. Resolved by 

- https://github.com/airbytehq/airbyte/pull/29786

Issue #29659, opened 2023-08-21
Use vector_db CDK in langchain connector


flash1293:
### What area the feature impact?

None

### Revelant Information

Based on https://github.com/airbytehq/airbyte/issues/29657

The langchain connector should be rewritten to use the provided functionality from the vector_db CDK: https://github.com/airbytehq/airbyte/pull/29442

Issue #29658, opened 2023-08-21
Use vector_db CDK in pinecone connector


flash1293:
Based on https://github.com/airbytehq/airbyte/issues/29657

The pinecone connector (not merged yet) should be rewritten to use the provided functionality from the vector_db CDK: https://github.com/airbytehq/airbyte/pull/29442

PR #29703, opened 2023-08-22
Vector based CDK

### LINKED ISSUE #29657: Vector database CDK
The langchain destination contains functionality that will be required for several vector-db-specific destinations. To avoid duplicate code and slightly deviating functionality, this functionality should be centralized in the CDK so multiple destinations can be built on top. To do so, the non-indexer-specific parts of the langchain destination should be pulled into a new `destination/vector_db_based` module in the CDK:
* Batcher
* DocumentProcessor (including config model)
* Embedder (including config model)
* Indexer interface

## Changes to the existing classes

* To avoid coupling to other libraries, the `Embedder` interface should not return a `langchain.embeddings.base import Embeddings`, but instead provide a method `embed_texts(self, texts: List[str]) -> List[List[float]]`
* Explicit metadata fields like in the pinecone connector should be supported too https://github.com/airbytehq/airbyte/pull/29442/files#diff-c4c0e0d6cc4b4c7e110f5d4ea4971e56d3e7c4380130b9b776d5eadfe896ff12R74
* Cohere embeddings like in the pinecone connector should be supported too: https://github.com/airbytehq/airbyte/pull/29442/files#diff-b94573f1964c343f457acaa673f42cacdda5e653df55dad62904d89709d73751R61

## Implementation

* The whole `destination/vector_db_based` module should be marked as experimental in the CDK as they might be subject to breaking changes.
* Like for file-based, this should be part of a package variant as there are a few additional pip dependencies that shouldn't be pulled into all connectors

flash1293:
## What

Fixes https://github.com/airbytehq/airbyte/issues/29657 by introducing helpers to make it easier to build vector-db-based destinations that can do chunking and embedding before loading data into the destination.

## How

All helpers are contained within `destinations/vector_db_based`, together with a Readme that explains the basic functionality.

The helpers are forked from the existing `destination-langchain` connector with some minor changes:
* Remove max_metadata logic and normalization for complex objects - as this is db-specific, the indexer should handle it appropriately 
* Remove all langchain types from the interfaces of the provided helpers
* Add explicit metadata field selection
* Add cohere embedder


airbyte-cdk/python/airbyte_cdk/destinations/vector_db_based/batcher.py
```diff
@@ -0,0 +1,26 @@
+#
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+#
+
+from typing import Any, Callable, List
+
+
+class Batcher:
```
aaronsteers, line 8: Can we add class-level docstrings for each class? To address:
- What is the class responsible for.
- How should it iteract with other classes.
- When it is expected to be subclassed by developers (or not).

aaronsteers (commented):
@flash1293 - Thanks for creating this. What is the first connector which will use this (langchain destination, I think?), and does it make sense to refactor one or more connectors to reference this codebase within this PR? Or would you prefer that to happen in a follow on PR?

flash1293:
@aaronsteers 

> What is the first connector which will use this (langchain destination, I think?),

Probably milvus as I'm working on this destination based on this PR already

> and does it make sense to refactor one or more connectors to reference this codebase within this PR? Or would you prefer that to happen in a follow on PR?

I planned to do the refactoring separately (https://github.com/airbytehq/airbyte/issues/29659 and https://github.com/airbytehq/airbyte/issues/29658 )

flash1293:
Thanks for your comments @aaronsteers , added a bit more documentation on how to use.

airbyte-cdk/python/airbyte_cdk/destinations/vector_db_based/config.py
```diff
@@ -0,0 +1,68 @@
...
+        description="List of fields in the record that should be stored as metadata. The field list is applied to all streams in the same way and non-existing fields are ignored. If none are defined, all fields are considered metadata fields. When specifying text fields, you can access nested fields in the record by using dot notation, e.g. `user.name` will access the `name` field in the `user` object. It's also possible to use wildcards to access all fields in an object, e.g. `users.*.name` will access all `names` fields in all entries of the `users` array. When specifying nested paths, all matching values are flattened into an array set to a field named by the path.",
+        always_show=True,
+        examples=["age", "user", "user.name"],
+    )
+
+    class Config:
+        schema_extra = {"group": "processing"}
+
+
+class OpenAIEmbeddingConfigModel(BaseModel):
```
girarda, line 41: should these embedding configs be part of the CDK or stay in the destinations they are used?

airbyte-cdk/python/airbyte_cdk/destinations/vector_db_based/document_processor.py
```diff
@@ -0,0 +1,118 @@
...
+
+    The Writer class uses the DocumentProcessor class to internally generate documents from records - in most cases you don't need to use it directly,
+    except if you want to implement a custom writer.
+
+    The config parameters specified by the ProcessingConfigModel has to be made part of the connector spec to allow the user to configure the document processor.
+    """
+
+    streams: Mapping[str, ConfiguredAirbyteStream]
+
+    def __init__(self, config: ProcessingConfigModel, catalog: ConfiguredAirbyteCatalog):
```
girarda, line 44: is the `max_metadata_size` used in t[he langchain destination](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/destination-langchain/destination_langchain/document_processor.py#L24) not needed?

Same question for the [metadata truncation](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/destination-langchain/destination_langchain/document_processor.py#L81). Is that only needed for langchain?

girarda (approved):

airbyte-cdk/python/airbyte_cdk/destinations/vector_db_based/config.py
flash1293: As I want them to be re-used by multiple destinations I think it makes sense to share them like this.

airbyte-cdk/python/airbyte_cdk/destinations/vector_db_based/document_processor.py
flash1293: Very good point, thanks for sanity-checking this :) This is what I meant by
> Remove max_metadata logic and normalization for complex objects - as this is db-specific, the indexer should handle it appropriately

in the PR description - it was a mistake to put this into the document processor as it's an indexer concern (I actually had to extend it a little bit already on the indexer level for Chroma because it doesn't support boolean values: https://github.com/airbytehq/airbyte/blob/dd170e2e9dd0aa09110ae115657da1c83761f87b/airbyte-integrations/connectors/destination-langchain/destination_langchain/indexer.py#L195 ). I don't see any patterns yet in terms of normalizing the metadata that could be pulled to the general level - if something emerges, we can refactor it down similar to the custom low-code components.

airbyte-cdk/python/airbyte_cdk/destinations/vector_db_based/README.md
```diff
@@ -0,0 +1,38 @@
...
+└┬──────────┬────────┬┘                
+┌▽────────┐┌▽──────┐┌▽────────────────┐
+│MyIndexer││Batcher││DocumentProcessor│
+└┬────────┘└───────┘└─────────────────┘
+┌▽───────┐                             
+│Embedder│                             
+└────────┘                             
+```
+
+Normally, only the `MyDestination` class and the `MyIndexer` calss has to be implemented specifically for the destination. The other classes are provided as is by the helpers.
```
nickblitz, line 38: Typo `calss` -> `class`.

PR #29710, opened 2023-08-22
✨ Source MongoDB Internal POC: Debezium target position

jdpgrailsdev:
## What
* Track target position to determine when an incremental sync should stop

## How
* Implement the `CdcTargetPosition` for MongoDB
* Track timestamp associated with most recent resume token
* Refactor resume token handling logic to common class

## Recommended reading order
1. `MongoDbCdcTargetPosition.java`
2. `MongoDbResumeTokenHelper.java`

PR is based off of https://github.com/airbytehq/airbyte/pull/29676


subodh1810 (commented):
This makes sense

jpefaur (approved):
I don't 100% understand what's going on in this PR, would be good if we could jump into a call tomorrow so you can walk me through it. Approving to not block you

edit: synced offline, looks good!

PR #29727, opened 2023-08-22
✨ Source MongoDB Internal POC: Debezium Configuration Properties

jdpgrailsdev:
## What
* Configure the Debezium MongoDB connector

## How
* Provide common configuration settings
* Override existing DebeziumPropertiesManager to work with MongoDB

## Recommended reading order
1. `DebeziumPropertiesManager.java`
2. `MongoDbDebeziumPropertiesManager.java`
3. `MongoDbCdcProperties.java`

**N.B.**:  I put both of the new classes in the `source-mongodb-internal-poc` package in order to avoid having to duplicate or move the constants to a common location.  If the manager belongs in the `debezium` package, I can do some additional refactoring to move the constants to a common location to avoid duplication.


airbyte-integrations/connectors/source-mongodb-internal-poc/src/main/java/io/airbyte/integrations/source/mongodb/internal/cdc/MongoDbDebeziumPropertiesManager.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Custom {@link DebeziumPropertiesManager} specific for the configuration of the Debezium MongoDB
+ * connector.
+ * <p />
+ * This implementation provides the specific connection properties required for the Debezium MongoDB
+ * connector. These properties differ from the general relational database connection properties
+ * used by the other Debezium connectors.
+ */
+public class MongoDbDebeziumPropertiesManager extends DebeziumPropertiesManager {
```
subodh1810, line 32: Can we keep this class in Debezium package only.

subodh1810 (commented):
We directly use the class `DebeziumPropertiesManager` in `DebeziumRecordPublisher`
 ```
 this.debeziumPropertiesManager = new DebeziumPropertiesManager(properties, config, catalog, offsetManager,  schemaHistoryManager);
 ```
 With the current change the connector will have to now send which Properties manager it wants to use which is not that bad but another thing that connector needs to take care of. The `DebeziumRecordPublisher` was created to help keep the common properties which are relevant to all the connectors at 1 place but with this change it kind of makes it difficult cause now properties are getting scattered over different places.

airbyte-integrations/connectors/source-mongodb-internal-poc/src/main/java/io/airbyte/integrations/source/mongodb/internal/cdc/MongoDbDebeziumPropertiesManager.java
jdpgrailsdev: I will see what can be done with this.  The problem is the existing one is very relational-database heavy.  The Debezium MongoDB connector does not use the same property keys for configuration that the other relational database connectors use.  We will probably need to do some refactoring to support telling the debezium record publisher which one to use.  I am going to merge this as is and clean it up as I integration with Debezium, if that is okay with you @subodh1810 .

jdpgrailsdev:
@subodh1810 I have updated the PR with the refactoring of the `DebeziumPropertyManager` that we discussed.

subodh1810 (changes requested):
There is a bug in the PR. We are using two different instances of the `offsetManager` and `schemaHistoryManager`. One is created in the connector (`MySqlInitialReadUtil`, etc.) while creating the `RelationalDbDebeziumPropertiesManager` and the other one is created in AirbyteDebeziumHandler. Thus Debezium will write state in a different file and Airbyte will try to read the state from a different file. 

Also lets run the tests for all the 3 CDC connectors for this change

subodh1810:
I think with this change we are making the connector aware of the AirbyteFileOffsetBackingStore and AirbyteSchemaHistoryStorage which I want to avoid. How about we send an enum from the connector and based on that construct the properties manager in the AirbyteHandler like we were doing before. The enums can be 
```
RELATIONAL,
MONGO
```
and something like 
`PropertiesManager p = enum == RELATIONAL ? new RelationalPropertiesManager : Mongo `

That way we wont have to create the objects of `AirbyteFileOffsetBackingStore` and `AirbyteSchemaHistoryStorage` in the connector itself.

jdpgrailsdev:
> 

@subodh1810 I think this is reasonable.  I will see how hard it is to make this change.  I may have to revert the previous change and start over to ensure all of the logic gets restored.

jdpgrailsdev:
@subodh1810 I took a stab at using an enum to select the correct property manager.  Please take a look and let me know if this removes the bug that you previously spotted.

subodh1810 (approved):
Looks good! Lets make sure the tests for mysql, mssql and postgres pass before merging

octavia-approvington:

octavia-approvington (approved):

PR #29731, opened 2023-08-22
Initial load stopping condition

### LINKED ISSUE #29722: source-mysql : Initial CDC Pk load should have stopping condition for chunking queries
For the new MySQL we keep issuing a series of queries until we get no records : 
```
select * from table order by pk limit 1000
select * from table where pk > pk_1 order by pk limit 1000
select * from table where pk > pk_2 order by pk limit 1000
select * from table where pk > pk_3 order by pk limit 1000
```

When these set of queries are run on a live database which is rapidly growing, we run the risk of the initial load never ending as data can be populated in the source MySQL DB faster than the chunking queries finish. In other words, there is always new data in the DB and this becomes a streaming solution. 

One way to get around this is to query the max (pk) at initial load time. This would be similar to the initial CDC snapshot. Then, our set of queries have a stopping condition and would look like: 
 ```
select * from table order by pk limit 1000
select * from table where pk > pk_1 **and pk <= pk_max** order by pk limit 1000
select * from table where pk > pk_2 **and pk <= pk_max** order by pk limit 1000
select * from table where pk > pk_3 **and pk <= pk_max** order by pk limit 1000
```

This would also be a slight optimization as well as we would not re-process the data that is added to the DB after the initial load started (both via CDC and initial load queries)

### PR DESCRIPTION
**Pull request by akashkulk**
Closes https://github.com/airbytehq/airbyte/issues/29722. 

Slight tweak to to what is described there. The max(pk_value) is calculated at the start of every sync. This is the stopping point for which queries stop. 

So, these series of queries : 
```
select * from table order by pk limit 1000
select * from table where pk > pk_1 order by pk limit 1000
select * from table where pk > pk_2 order by pk limit 1000
select * from table where pk > pk_3 order by pk limit 1000
```

Now look like this : 

```
select * from table order by pk limit 1000
select * from table where pk > pk_1 and pk <= pk_max order by pk limit 1000
select * from table where pk > pk_2 and pk <= pk_max order by pk limit 1000
select * from table where pk > pk_3 and pk <= pk_max order by pk limit 1000
```

Mainly so that the connector can deterministically finish an initial load

Tested locally. Bit hard to unit test this as we'd have to support modifying a database just as a connector is running


subodh1810 (approved):

octavia-approvington:
Lets do it! <br/> giddddy-up

octavia-approvington (approved):

PR #29636, opened 2023-08-18
🐛 Destination v2 general availability

jbfbell:
changelogs:

bigquery:
```
Add support for incremental/dedup on streams with no explicit cursor (e.g. source-postgres xmin)
Change raw table format to preserve high-precision numeric values
```

snowflake
```
Add support for incremental/dedup on streams with no explicit cursor (e.g. source-postgres xmin)
```

PR #29734, opened 2023-08-22
DATs: Fix spurious failures

edgao:
There are two false negatives happening in destination-snowflake tests.
1. On master - `testNamespaces` is failing, because `assertNamespaceNormalization` fails (because the namespace has a random suffix on it, but the normalized namespace does not have that suffix).
2. In https://github.com/airbytehq/airbyte/pull/29636#issuecomment-1688935644 - tearDown is failing (because the namespace contains special characters).

This PR:
1. Updates NamespaceTestCaseProvider to use the same random suffix for both namespace and normalizedNamespace
2. tweaks the teardown method to wrap namespaces in quotes. This definitely does not always work (i.e. doesn't always drop the right namespace), but at least means we're not spuriously failing tests.

I would also love to understand why https://github.com/airbytehq/airbyte/pull/29652#issuecomment-1685451850 was successful, given that it's failing on master... but that's a problem for another week.

### DIFF
--- a/airbyte-integrations/connectors/destination-snowflake/src/test-integration/java/io/airbyte/integrations/destination/snowflake/SnowflakeInsertDestinationAcceptanceTest.java
+++ b/airbyte-integrations/connectors/destination-snowflake/src/test-integration/java/io/airbyte/integrations/destination/snowflake/SnowflakeInsertDestinationAcceptanceTest.java
@@ -186,11 +186,14 @@ public class SnowflakeInsertDestinationAcceptanceTest extends DestinationAccepta
 
   @Override
   protected void tearDown(final TestDestinationEnv testEnv, final HashSet<String> TEST_SCHEMAS) throws Exception {
-    String dropSchemaQuery = String.format("DROP SCHEMA IF EXISTS %s", config.get("schema").asText());
-    database.execute(dropSchemaQuery);
-
+    TEST_SCHEMAS.add(config.get("schema").asText());
     for (final String schema : TEST_SCHEMAS) {
-      dropSchemaQuery = String.format("DROP SCHEMA IF EXISTS %s", schema);
+      // we need to wrap namespaces in quotes, but that means we have to manually upcase them.
+      // thanks, v1 destinations!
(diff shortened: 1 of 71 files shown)


evantahler (approved):

airbyte-integrations/connectors/destination-snowflake/src/test-integration/java/io/airbyte/integrations/destination/snowflake/SnowflakeInsertDestinationAcceptanceTest.java
```diff
@@ -186,11 +186,14 @@ protected void setup(final TestDestinationEnv testEnv, final HashSet<String> TES
 
   @Override
   protected void tearDown(final TestDestinationEnv testEnv, final HashSet<String> TEST_SCHEMAS) throws Exception {
```
jbfbell, line 188: unrelated nit - isn't it weird that this parameter is uppercased?

jbfbell (approved):

airbyte-integrations/connectors/destination-snowflake/src/test-integration/java/io/airbyte/integrations/destination/snowflake/SnowflakeInsertDestinationAcceptanceTest.java
edgao: ... I didn't even notice that. assumed we were using the constant directly. lemme fix it

edgao: that diff is awful, but it's purely automated intellij refactor to remove the TEST_SCHEMAS parameter. CI is going to be dumb now but at least the first run should be snowflake alone, and I'll approve-and-merge afterwards.

octavia-approvington:
All in!! <br/> all in baby

octavia-approvington (approved):

PR #29753, opened 2023-08-23
✨ Source S3: v4 rollout/feature parity

maxi297:
## What
Missing features for releasing source-s3 v4 (either on the CDK or source-s3 specifically)

Each commit represent a feature:
* (CATs) have schema within catalog like we would in the platform
* CSV to drop fields that are not in schema
* CSV null_values,  true_values and false_values default does not align with pyarrow
* ContinuationToken is not passed when fetch S3 objects hence pagination does not work 
* Support both endpoint and aws access key

Last commit is to update the version


girarda (approved):
🪨 🎸

PR #29763, opened 2023-08-23
✨ Source MongoDB Internal POC: CDC State Handling

jdpgrailsdev:
## What
* Add state handling/management for both initial snapshots and incremental syncs as part of CDC for MongoDB

## How
* Implement the `CdcStateHandler` interface so that state is updated as part of the CDC incremental sync
* Add a state manager based on the relational database state management code that can be used to keep track of state as both initial snapshots are performed and as part of incremental syncs
* Refactor existing code to use the state manager for state tracking

## Recommended reading order
1. `MongoDbStateManager.java`
2. `MongoDbCdcStateHandler.java`
3. `MongoDbCdcState.java`

This PR does not completely hook in the state management for CDC.  That will come in a later PR when we integrate all the CDC bits together to actually facilitate a CDC-based sync.  For now, I confirmed that all unit tests validation state pass, as well as acceptance tests that perform incremental syncs and validate state messages pass.  I also put some TODO's in place where we need to come back and fix once we integrate this.

I opted to not depend on the relational DB package where the current state management code lives.  That code is very coupled to relational database concepts, which do not really apply to MongoDB.

Finally, this PR is based off of https://github.com/airbytehq/airbyte/pull/29730 (you can ignore the metadata injection changes -- I needed some changes from that PR to implement this code) and supersedes https://github.com/airbytehq/airbyte/pull/29678, which will be closed in favor of this PR.


jpefaur (approved):
I _think_ I understand what's going on here. I'm going to read it again tomorrow morning, but feel free to merge

airbyte-integrations/connectors/source-mongodb-internal-poc/src/main/java/io/airbyte/integrations/source/mongodb/internal/state/MongoDbStateManager.java
```diff
@@ -0,0 +1,188 @@
...
+  public static MongoDbStateManager createStateManager(final JsonNode initialState) {
+    final MongoDbStateManager stateManager = new MongoDbStateManager();
+
+    if (initialState != null) {
+      LOGGER.info("Initial state {}", initialState);
+      final List<AirbyteStateMessage> stateMessages = deserializeState(initialState);
+      if (!stateMessages.isEmpty()) {
+        if (stateMessages.size() == 1) {
+          final AirbyteStateMessage stateMessage = stateMessages.get(0);
+          stateManager.updateCdcState(Jsons.object(stateMessage.getGlobal().getSharedState(), MongoDbCdcState.class));
```
subodh1810, line 57: Isn't the CDC state supposed to be the debezium state? How will it be de-serialised to `MongoDbCdcState`

airbyte-integrations/connectors/source-mongodb-internal-poc/src/main/java/io/airbyte/integrations/source/mongodb/internal/MongoDbStateIterator.java
```diff
@@ -60,23 +57,24 @@ class MongoDbStateIterator implements Iterator<AirbyteMessage> {
...
                        final Instant emittedAt,
                        final int checkpointInterval) {
     this.iter = iter;
+    this.stateManager = stateManager;
     this.stream = stream;
     this.checkpointInterval = checkpointInterval;
     this.emittedAt = emittedAt;
     fields = CatalogHelpers.getTopLevelFieldNames(stream).stream().toList();
-    lastId = state.map(MongodbStreamState::id).orElse(null);
+    lastId = stateManager.getStreamState(stream.getStream().getName(), stream.getStream().getNamespace()).map(MongoDbStreamState::id).orElse(null);
```
subodh1810, line 77: this.fields
this.lastId

subodh1810 (changes requested):
Few questions/concerns : 
1. Need to understand if swallowing exception in the `hasNext` method of `MongoDbStateIterator` is the right thing to do?
2. How would we identify that initial sync for a given stream is complete. I dont see any special state emitted for streams that complete initial sync to denote now we just need to process the change streams for these streams. Which leads to the next question is that I dont see any specific logic to  identify which streams only need to go through the initial sync. Right now all the streams present in the catalog go through the intial sync all the time.
3. Other DBs emit state at a frequency of every 15 minutes or once more than 10K records are emiited. May be we could do the same thing?

jdpgrailsdev:
> Few questions/concerns :
> 
> 1. Need to understand if swallowing exception in the `hasNext` method of `MongoDbStateIterator` is the right thing to do?
> 2. How would we identify that initial sync for a given stream is complete. I dont see any special state emitted for streams that complete initial sync to denote now we just need to process the change streams for these streams. Which leads to the next question is that I dont see any specific logic to  identify which streams only need to go through the initial sync. Right now all the streams present in the catalog go through the intial sync all the time.
> 3. Other DBs emit state at a frequency of every 15 minutes or once more than 10K records are emiited. May be we could do the same thing?

@subodh1810 I think the answer to #1 is that it is the correct thing to do, but I will defer to @colesnodgrass as he wrote that code.  As for #2 and #3, those are out of scope of this PR.  The intent here is to just set up the initial structure.  We will need to handle those two cases when we integrate this with the iterator, which itself will probably need to move/change to be part of the code that sets up the iterators for the CDC sync given the state.

airbyte-integrations/connectors/source-mongodb-internal-poc/src/main/java/io/airbyte/integrations/source/mongodb/internal/state/MongoDbStateManager.java
jdpgrailsdev: The `MongoDbCdcState` holds the data extracted from the Debezium offset map.  There is code to handle the translation to/from the offset map and `MongoDbCdcState` that is not in this PR, but will be in the PR that integrates all the bits together.

airbyte-integrations/connectors/source-mongodb-internal-poc/src/main/java/io/airbyte/integrations/source/mongodb/internal/cdc/MongoDbCdcConnectorMetadataInjector.java
```diff
@@ -12,8 +12,8 @@
 import io.airbyte.integrations.debezium.CdcMetadataInjector;
 import io.airbyte.integrations.debezium.internals.mongodb.MongoDbDebeziumConstants;
 import java.time.Instant;
-
 import java.util.concurrent.atomic.AtomicLong;
+import org.bson.BsonTimestamp;
```
subodh1810, line 16: Not required right?

subodh1810 (approved):
Looks good. Again lets make sure that the tests for all the other CDC connectors are green.


 Also as discussed over chat I pushed a commit to disable the following methods
```
  @Override
  public AirbyteMessage saveStateAfterCompletionOfSnapshotOfNewStreams() {
    throw new RuntimeException("Debezium is not used to carry out the snapshot of tables.");
  }

  @Override
  public boolean isCdcCheckpointEnabled() {
    return false;
  }
 ```
 
 The method `saveStateAfterCompletionOfSnapshotOfNewStreams` is not required cause we wont be using Debezium for initial snapshot of streams. 
 
 The method `isCdcCheckpointEnabled` requires the following methods in `CdcTargetPosition` to be implemented as well. So an item on the TODO list to implement these methods in the future PR and then we can set the `isCdcCheckpointEnabled` to true.
 ``` 
 default boolean isEventAheadOffset(final Map<String, String> offset, final ChangeEventWithMetadata event) {
    return false;
  }
  ```
and
```
  default boolean isSameOffset(final Map<String, String> offsetA, final Map<String, String> offsetB) {
    return true;
  }
  ```

